PR #1438, opened 2019-07-11
Use CPack in component-install mode for the libiothsm deb packages

arsing:
Otherwise the packages end up including internal cmake files and libs,
specifically ctest and testrunnerswitcher.

---

For example, the 1.0.8-rc3 libiothsm-std.deb has these extra files :

```
/usr/cmake
/usr/cmake/ctestConfig.cmake
/usr/cmake/ctestConfigVersion.cmake
/usr/cmake/ctestTargets-release.cmake
/usr/cmake/ctestTargets.cmake
/usr/cmake/testrunnerswitcherConfig.cmake
/usr/cmake/testrunnerswitcherConfigVersion.cmake
/usr/cmake/testrunnerswitcherTargets-release.cmake
/usr/cmake/testrunnerswitcherTargets.cmake
/usr/include
/usr/include/ctest.h
/usr/include/ctest_macros.h
/usr/include/ctrs_sprintf.h
/usr/include/testmutex.h
/usr/include/testrunnerswitcher.h
/usr/lib/libctest.a
/usr/lib/libtestrunnerswitcher.a
```

Ref:

- https://stackoverflow.com/questions/27619234/how-to-exclude-files-dirs-from-cpack-binary-using-cmake
- https://gitlab.kitware.com/cmake/community/wikis/doc/cpack/Component-Install-With-CPack


arsing:
The centos RPM build has `-Drun_unittests=on` and doesn't have this problem. So it might be enough to just set the deb package builds to do that as well. Currently they set it to on. (Why?)

edgelet/build/linux/debian/arm32v7/package-libiothsm.sh
```diff
@@ -42,7 +42,7 @@ run_command "cd /target/hsm/build && make package"
 
 # Old CPACK produces non-standard deb package filenames.
 # This renames them
-for f in $BUILD_DIR/$PACKAGE_NAME-*-Linux.deb ; do
+for f in $BUILD_DIR/$PACKAGE_NAME-*-Linux-libraries.deb ; do
```
mrohera: Out of curiosity, was this renamed this way for a reason, if so why in this way?

arsing: CPack appends the component name to the filename, and `libraries` is the component name I set.

mrohera:
I think this is a good point you bring up and I think needs to be cleaned up. If I recall this was added to temporarily unblock builds for a platform and then this was propagated.

myagley:
Yeah, this seems like a bug. There's no reason that switch should be on in the package build.

arsing:
Packages build succeed with #1439 so let's go with that.

PR #1439, opened 2019-07-12
Disable building libiothsm tests in deb package build

arsing:
Otherwise the packages end up including internal cmake files and libs,
specifically ctest and testrunnerswitcher.

---

This is the alternative approach to https://github.com/Azure/iotedge/pull/1438


builds/misc/packages.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
         displayName: Setup libiothsm
         inputs:
           workingDirectory: 'edgelet/hsm-sys/azure-iot-hsm-c/build'
-          cmakeArgs: '-DBUILD_SHARED=ON -Drun_unittests=ON -Duse_emulator=OFF -DCMAKE_BUILD_TYPE=Release -DCPACK_PACKAGE_VERSION=$(VERSION) -DCPACK_DEBIAN_PACKAGE_RELEASE=$(REVISION)  ..'
+          cmakeArgs: '-DBUILD_SHARED=ON -Drun_unittests=off -Duse_emulator=OFF -DCMAKE_BUILD_TYPE=Release -DCPACK_PACKAGE_VERSION=$(VERSION) -DCPACK_DEBIAN_PACKAGE_RELEASE=$(REVISION)  ..'
```
myagley: We should verify this is turned off in the release build as well. The platforms that use the `package-*` scripts should be fine but the arm32 phase might need to be tweaked.

arsing: Yeah, the Ubuntu amd64 phase uses `cmake` directly and sets `run_unittests` to `on`, so it needs to be fixed. The others use the `package-` scripts, so they're fine.

PR #1442, opened 2019-07-12
[k8s][EdgeAgent] Planners and commands for Kubernetes

darobs:
Pulled in code from POC.
Cleaned up and added tests.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/commands/KubernetesCrdCommand.cs
```diff
@@ -0,0 +1,263 @@
...
+            try
+            {
+                HttpOperationResponse<object> currentDeployment = await this.client.GetNamespacedCustomObjectWithHttpMessagesAsync(
+                    Constants.K8sCrdGroup,
+                    Constants.K8sApiVersion,
+                    this.deviceNamespace,
+                    Constants.K8sCrdPlural,
+                    resourceName,
+                    cancellationToken: token);
+                string body = JsonConvert.SerializeObject(currentDeployment.Body);
```
kkdawkins: Why are we serializing this object only to deserialize it later (line 156)?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/commands/KubernetesCrdCommand.cs
```diff
@@ -0,0 +1,263 @@
...
+            await this.UpdateImagePullSecrets(secrets, token);
+
+            var metadata = new V1ObjectMeta(name: resourceName, namespaceProperty: this.deviceNamespace);
+            // need resourceVersion for Replace.
+            activeDeployment.ForEach(deployment => metadata.ResourceVersion = deployment.Metadata.ResourceVersion);
+            var customObjectDefinition = new EdgeDeploymentDefinition<DockerConfig>(metaApiVersion, Constants.K8sCrdKind, metadata, modulesList);
+            string customObjectString = this.deploymentSerde.Serialize(customObjectDefinition);
+
+            // the dotnet client is apparently really picky about all names being camelCase,
+            object crdObject = JsonConvert.DeserializeObject(customObjectString);
```
kkdawkins: Once again, serialize followed by a deserialize...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
```diff
@@ -0,0 +1,128 @@
...
+            // Check that module names sanitize and remain unique.
+            var convertedNames = desired.Modules.Keys.Select(KubeUtils.SanitizeK8sValue).Distinct();
+            if (desired.Modules.Count != convertedNames.Count())
+            {
+                string nameList = desired.Modules.Keys.Join(",");
+                throw new InvalidIdentityException($"Deployment will cause a name collision in Kubernetes namespace, modules: [{nameList}]");
+            }
+
+            // TODO: improve this so it is generic for all potential module types.
+            if (!desired.Modules.Values.All(p => p is IModule<DockerConfig>))
```
kkdawkins: Use if (...any(...)) instead of (!all). Any will stop reflecting at the first instance whereas all will check each instance with refelection.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CombinedKubernetesConfigProvider.cs
```diff
@@ -0,0 +1,111 @@
...
+                        EndpointsConfig = endpointsConfig
+                    };
+                }
+            }
+        }
+
+        void MountSockets(IModule module, CreateContainerParameters createOptions)
+        {
+            var workloadUri = new Uri(this.configSource.Configuration.GetValue<string>(Core.Constants.EdgeletWorkloadUriVariableName));
+            if (workloadUri.Scheme == "unix")
```
kkdawkins: Any case where Scheme wouldn't be lowercase? Need to normalize here?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CombinedKubernetesConfigProvider.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        readonly IConfigSource configSource;
+
+        public CombinedKubernetesConfigProvider(IEnumerable<AuthConfig> authConfigs, IConfigSource configSource)
+            : base(authConfigs)
+        {
+            this.configSource = Preconditions.CheckNotNull(configSource, nameof(configSource));
+        }
+
+        static CreateContainerParameters CloneOrCreateParams(CreateContainerParameters createOptions) =>
```
kkdawkins: I'd like us to be more consistent with using qualifiers. Can we add them to the methods in this PR? I will make the same change in my PR.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CombinedKubernetesConfigProvider.cs
```diff
@@ -0,0 +1,111 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
```
kkdawkins: usings are inside a namespace? Outside?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesCommandFactory.cs
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
+    {
```
kkdawkins: Something is off with the tabs here

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                throw new InvalidModuleException("Kubernetes deployment currently only handles type=DockerConfig");
```
kkdawkins: If you want to make this more generic, you can use typeof(T).FullName instead of hard coding DockerConfig

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/commands/KubernetesCrdCommand.cs
```diff
@@ -0,0 +1,263 @@
...
+            string metaApiVersion = Constants.K8sApi + "/" + Constants.K8sApiVersion;
+
+            var modulesList = new List<KubernetesModule<DockerConfig>>();
+            var secrets = new Dictionary<string, ImagePullSecret>();
+            foreach (var runtime in this.runtimeInfo)
+            {
+                foreach (var m in this.modules)
+                {
+                    var combinedConfig = this.combinedConfigProvider.GetCombinedConfig(m, runtime);
+                    CombinedDockerConfig dockerConfig = combinedConfig as CombinedDockerConfig;
```
kkdawkins: Is this cast necessary? Will GetCombinedConfig take a type parameter? If it is necessary, you should be checking the cast because IIRC, as will throw if it fails.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        public async Task<Plan> PlanAsync(
+            ModuleSet desired,
+            ModuleSet current,
+            IRuntimeInfo runtimeInfo,
+            IImmutableDictionary<string, IModuleIdentity> moduleIdentities)
+        {
+            Events.LogDesired(desired);
+            Events.LogCurrent(current);
+            Events.LogIdentities(moduleIdentities);
```
kkdawkins: For my own knowledge, is it safe to log an identity? Are we leaking a secret here?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            // Check that module names sanitize and remain unique.
+            var convertedNames = desired.Modules.Keys.Select(KubeUtils.SanitizeK8sValue).Distinct();
+            if (desired.Modules.Count != convertedNames.Count())
```
kkdawkins: These lines can be simplified as 

if(desired.Modules.Keys.Select(KubeUtils.SanitizeK8sValue).GroupBy(n => n).Any(c => c.Count() > 1))

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                string nameList = desired.Modules.Keys.Join(",");
+                throw new InvalidIdentityException($"Deployment will cause a name collision in Kubernetes namespace, modules: [{nameList}]");
```
kkdawkins: Would this error message be better printing out the two that would cause a collision? You could use the groups created above to do this easily

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
```diff
@@ -0,0 +1,128 @@
...
+                // and the status of that deployment through the "edgedeployments" API.
+                var k8sModules = desired.Modules.Select(m => new KubernetesModule<DockerConfig>(m.Value as IModule<DockerConfig>));
+
+                var crdCommand = new KubernetesCrdCommand<CombinedDockerConfig>(this.deviceNamespace, this.iotHubHostname, this.deviceId, this.client, k8sModules.ToArray(), Option.Some(runtimeInfo), this.combinedConfigProvider as ICombinedConfigProvider<CombinedDockerConfig>);
+                var planCommand = await this.commandFactory.WrapAsync(crdCommand);
+                var planList = new List<ICommand>
+                {
+                    planCommand
+                };
+                Events.PlanCreated(planList);
```
kkdawkins: Should events like this live closer to relevance? In this case, should the PlanCreated event happen inside of plan right before it returns? If the plan object fails, it was never created :)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/commands/KubernetesCrdCommand.cs
```diff
@@ -0,0 +1,263 @@
...
+                    ["docker"] = typeof(CombinedDockerConfig)
+                }
+            };
+
+            this.deploymentSerde = new TypeSpecificSerDe<EdgeDeploymentDefinition<DockerConfig>>(deserializerTypesMap);
+        }
+
+        // We use the sum of the IDs of the underlying commands as the id for this group
+        // command.
+        public string Id => this.id.Value;
```
kkdawkins: This should live up with the other class properties.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CombinedKubernetesConfigProvider.cs
darobs: The standard for this project is inside the namespace.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CombinedKubernetesConfigProvider.cs
darobs: This is the applied style setting - private is not declared.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CombinedKubernetesConfigProvider.cs
darobs: It shouldn't, but you never know.  Will add case insensitive checking.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/commands/KubernetesCrdCommand.cs
darobs: I double checked, and the "as" operator returns null if conversion fails, and then I'm checking for null in the next line.  I think this could use some serious refactoring if we get different module types.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/commands/KubernetesCrdCommand.cs
darobs: We want to convert the existing CRD (which IIRC is a JSONObject returned as object), into an  EdgeDeploymentDefinition, so we can extract information from it, and then use that information to either replace or create the new EdgeDeploymentDefinition.  This is kind of a quirk with the Kubernetes client and the CustomObject functions.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/commands/KubernetesCrdCommand.cs
darobs: This is converting back from an EdgeDeploymentDefinition to a generic object that is formatted as the client wants.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
darobs: The log identities function below only reports the module names, so no secret is being leaked in the logs.  That would be a bad thing.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
darobs: Nice - will change.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
darobs: The plan itself is just populating data fields, so it's not significant behavior.   In terms of logging, you must CONFORM.
OBEY CONFORM

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
darobs: Also a good ideas - done.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
darobs: If I go by Visual Studio, this was how it preferred me to do it.  All() will also short circuit, so Any() and All() will do the same amount of work.

PR #1443, opened 2019-07-12
External provisioning changes

shantanu1singh:
1. Add error reasons.
2. Activate identity key in case the symmetric-key mode is used with the key specified in the payload itself.


edgelet/provisioning/src/provisioning.rs
```diff
@@ -1183,8 +1189,8 @@ mod tests {
             .then(|result| match result {
```
avranju: These `match` statements on the `Result` instances can be written like so:
```rust
assert_eq!(
  result.err().unwrap(),
  ErrorKind::ExternalProvisioning(ExternalProvisioningErrorReason::InvalidCredentialSource));
```

edgelet/provisioning/src/provisioning.rs
```diff
@@ -265,18 +265,24 @@ where
...
-                                        Err(Error::from(ErrorKind::Provision))
+                                        Err(Error::from(ErrorKind::ExternalProvisioning(ExternalProvisioningErrorReason::SymmetricKeyNotSpecified)))
                                     },
                                     |key| {
+                                        let decoded_key = base64::decode(&key).map_err(|_| {
+                                            Error::from(ErrorKind::ExternalProvisioning(ExternalProvisioningErrorReason::InvalidSymmetricKey))
+                                        })?;
+
+                                        key_activator
+                                            .activate_identity_key(KeyIdentity::Device, "primary".to_string(), &decoded_key)
```
avranju: So, is activation of the key something that was missed in the implementation before?

shantanu1singh: That's correct. Initially, the way we had spec'd the external provisioning mode, it was being assumed that the HSM will have the primary key and the 'activation' of the primary key won't be required as the customer would be creating their own custom HSM library wherein the key would be pre-activated.

However, later we added support for specifying the primary key in the payload itself and didn't think about the 'activation' scenario in that case .

edgelet/provisioning/src/provisioning.rs
arsing: `result.unwrap_err()`

shantanu1singh: Done

edgelet/provisioning/src/provisioning.rs
avranju: I see. Got it.

PR #1445, opened 2019-07-13
Fix Serde derive import to better interact with downstream code

kkdawkins:
React to Serde issue: https://github.com/serde-rs/serde/issues/1441


arsing:
I assume there's no actual crate that has this problem, and you're doing this defensively? I say this because none of our crates currently import serde with the feature enabled, and none of our crates are intended to be used by any third-party crates as dependencies that we would have to cater to (ie there's no "downstream" to "fix" for). Is this correct?

Anyway, I've been aware of that issue but I don't agree with the resolution. It makes it so that you end up treating an optional feature as required ("never import serde without enabling this feature"), and more importantly it breaks any other crates in the crate graph (not necessarily your ancestors or descendants, but completely unrelated sibling crates) that do have the original issue of importing `serde::Serialize` and `serde_derive::Serialize` simultaneously.

My preference is to use fully-qualified attribute paths instead (ie `#[derive(serde_derive::Deserialize)]`), which is also what I do in my own crates.

Thoughts?

kkdawkins:
@arsing - Actually, this PR was a reaction to Claire's project which uses some of our crates as a downstream dependency. That's how we discovered the issue :) 

I am not opposed to your proposed solution, I was simply using the guidance from serde.rs as a way of fixing the issue.

PR #1455, opened 2019-07-17
Fix build for snitcher & add support for UrlConnector

avranju:
This PR fixes up snitcher with some dependency upgrades and general clean-up and also adds support for the `UrlConnector` type from `edgelet` so that snitcher is able to work on Windows where the Docker daemon runs on a Windows named pipe endpoint.


tools/snitch/snitcher/src/lib.rs
```diff
@@ -451,67 +413,3 @@ where
         })
     })
 }
-
-#[cfg(test)]
-mod tests {
```
philipktlin: why all these tests are removed?

avranju: They need to be re-written I think. They are more like end-to-end tests and not unit tests. I can add them back in if you like but in order to run them one has to setup up a complete environment to run successfully.

tools/snitch/snitcher/src/lib.rs
```diff
@@ -390,38 +357,46 @@ pub fn raise_alert(
...
+                        .and_then(move |(status, body)| {
+                            if status.is_success() {
+                                Ok(())
+                            } else {
+                                Err(Error::from((status, &*body)))
+                            }
+                        })
+                });
+
+            Either::A(result.into_future())
```
avranju: `result` is already a future. Shouldn't need the `into_future` call here.

tools/snitch/snitcher/src/lib.rs
```diff
@@ -452,66 +427,51 @@ where
...
-    fn module_logs() {
-        env_logger::init();
-
-        let settings = Settings::default().merge_env().unwrap();
-        let tasks = get_module_logs(&settings)
-            .map(|_| ())
-            .map_err(|err| panic!("ERROR: {:?}", err));
-        tokio::run(tasks);
-    }
+pub trait UrlExt {
```
avranju: We should be able to add a reference to the `edgelet-core` crate instead of copying this code.

tools/snitch/snitcher/src/lib.rs
philipktlin: updated

PR #1459, opened 2019-07-19
1.0.8: Adding Experimental feature flags for the new experimental features

varunpuranik:
Adding experimentalFeature* flags to the new experimental features. To enable a feature, you need to also enable Experimental features. 

For example, to enable UploadLogs, you would set the following env vars - 
experimentalFeatures__enabled=true
experimentalFeatures__enableUploadLogs=true


varunpuranik:
> I'm nervous about the double underscore in the env var. Is it possible to make it one underscore?

This was done deliberately - everything inside experimentalFeatures__* is inside the experimentalFeatures "section" in the configuration. In the App.config file, it would look something like this - 
```
{
    "experimentalFeatures": {
        "enabled": true,
        "enableUploadLogs": true
    },
}
```
(sorry for the unfortunate capitalization, but I am just trying to be consistent here. Maybe we should change

PR #1464, opened 2019-07-22
Enable experimental flags for experimental features

varunpuranik:
Cherry pick [9e6ea0c](https://github.com/Azure/iotedge/commit/9e6ea0c7df6568554cb6508ca56a8f9ae489b07b)
Add GetLogs behind experimental flag


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/ExperimentalFeatures.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Agent.Service
+{
+    using Microsoft.Azure.Devices.Edge.Storage;
+    using Microsoft.Extensions.Configuration;
+    using Microsoft.Extensions.Logging;
+
+    public class ExperimentalFeatures
```
ancaantochi: >ExperimentalFeatures [](start = 17, length = 20)

Should this be called FeatureConfiguration or FeaturesFilter? When they are enabled after a while they are actually not experimental anymore

ancaantochi: Maybe we can use a common class for edge agent and edge hub, I see some of the features are common.

---
In reply to: [308914144](https://github.com/Azure/iotedge/pull/1464#discussion_r308914144) [](ancestors = 308914144)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/ExperimentalFeatures.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        ExperimentalFeatures(bool enabled, bool disableCloudSubscriptions, bool enableUploadLogs, bool enableGetLogs)
+        {
+            this.Enabled = enabled;
+            this.DisableCloudSubscriptions = disableCloudSubscriptions;
+            this.EnableUploadLogs = enableUploadLogs;
+            this.EnableGetLogs = enableGetLogs;
+        }
+
+        public static ExperimentalFeatures Init(IConfiguration experimentalFeaturesConfig, ILogger logger)
```
ancaantochi: >Init [](start = 43, length = 4)

nit: Create

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/ExperimentalFeatures.cs
```diff
@@ -0,0 +1,37 @@
...
+            bool enabled = experimentalFeaturesConfig.GetValue("enabled", false);
+            bool disableCloudSubscriptions = enabled && experimentalFeaturesConfig.GetValue("disableCloudSubscriptions", false);
+            bool enableUploadLogs = enabled && experimentalFeaturesConfig.GetValue("enableUploadLogs", false);
+            bool enableGetLogs = enabled && experimentalFeaturesConfig.GetValue("enableGetLogs", false);
+            var experimentalFeatures = new ExperimentalFeatures(enabled, disableCloudSubscriptions, enableUploadLogs, enableGetLogs);
+            logger.LogInformation($"Experimental features configuration: {experimentalFeatures.ToJson()}");
+            return experimentalFeatures;
+        }
+
+        public bool Enabled { get; }
```
ancaantochi: >Enabled [](start = 20, length = 7)

nit: IsEnabled, same for other features

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/ExperimentalFeatures.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Service
+{
+    using Microsoft.Extensions.Configuration;
+
+    public class ExperimentalFeatures
+    {
+        public ExperimentalFeatures(bool enabled, bool disableCloudSubscriptions, bool disableConnectivityCheck)
+        {
```
ancaantochi: private
I see this is used in the test, but a mock can be provided instead

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/ExperimentalFeatures.cs
varunpuranik: The reason it was called ExperimentalFeatures_* is to convey the intent that these are experimental features and subject to change in the future. The naming was critical for that. 
I realize that it is going to be a bit tricky to update the flags once the features are not experimental anymore. But I have some thoughts related to what we can do then...

varunpuranik: As for a common class - the common functionality is by chance, imo, and I don't see there being much overlap between EdgeHub and EdgeAgent going forward.. so slight duplication is probably okay there.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/ExperimentalFeatures.cs
varunpuranik: I was going for consistency here - most of the other enabled flags are just named Enabled (for example - MqttSettings__Enabled, AmqpSetting__Enabled).

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/ExperimentalFeatures.cs
varunpuranik: Any reason to keep this private? It is not a singleton.. 
I see the Init/Create more as a helper than a factory method here.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/ExperimentalFeatures.cs
varunpuranik: Fixed..

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/ExperimentalFeatures.cs
ancaantochi: What I meant was IsEnabled for the property name, not for the config flag

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/ExperimentalFeatures.cs
ancaantochi: At the moment this only has 3 parameters, but with new features it will be a few bool parameters which can be difficult to make sure they are provided in the right order, using Create with a configuration solves the problem.

PR #1469, opened 2019-07-23
Update Changelog for 1.0.8

varunpuranik:


myagley:
How should we handle the changelog on the master branch? Should we only show GA releases or all of the RCs. Part of me thinks we should only list GA releases in which case we should collapse all of the RC changes?

varunpuranik:
That is fair.. and makes sense. Let me collapse the changes into 1.0.8.

PR #1472, opened 2019-07-24
Update Setup-Env.ps1

SlavyMihov:
* Add option to (not) update iotedge service (switch AttemptUpdate).
* Handle potential errors when attempting to delete residual modules.

### DIFF
--- a/scripts/windows/test/Setup-Env.ps1
+++ b/scripts/windows/test/Setup-Env.ps1
@@ -89,7 +89,8 @@ If ($osEdition -eq "IoTUAP")    # Windows IoT Core - update iotedge
             }
         }
         catch {
-            Write-Host "Cleanup existing containers failed. Nothing to clean."
+            Write-Host "Cleanup existing containers failed."
+			Write-Host $_.Exception.Message
         }
 
         # Delete iotedge config file
--- a/scripts/windows/test/Setup-Env.ps1
+++ b/scripts/windows/test/Setup-Env.ps1
@@ -20,10 +20,14 @@
     .PARAMETER ArtifactImageBuildNumber
         Artifact image build number; it is used to construct path of docker images, pulling from docker registry. E.g. 20190101.1.
 
+    .PARAMETER AttemptUpdate
+        Switch controlling if an update of iotedge service should be attempted.
+
     .EXAMPLE
         .\Setup-Env.ps1
             -E2ETestFolder "C:\Data\e2etests"
             -ArtifactImageBuildNumber "20190101.1"
+            -AttemptUpdate
 
     .NOTES
         This script is to setup the environment for running E2E tests.
@@ -38,7 +42,9 @@ Param (
     [string] $E2ETestFolder = ".",
 
     [ValidateNotNullOrEmpty()]
-    [string] $ArtifactImageBuildNumber = $(Throw "Artifact image build number is required")
+    [string] $ArtifactImageBuildNumber = $(Throw "Artifact image build number is required"),
+
+    [switch] $AttemptUpdate
 )
 
 Function PrintHighlightedMessage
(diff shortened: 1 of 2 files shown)


scripts/windows/test/Setup-Env.ps1
```diff
@@ -76,7 +82,15 @@ If ($osEdition -eq "IoTUAP")    # Windows IoT Core - update iotedge
         }
         
         Write-Host "Cleanup existing containers..."
-        docker -H npipe:////./pipe/iotedge_moby_engine rm -f $(docker -H npipe:////./pipe/iotedge_moby_engine ps -aq) 2>null
+        try {
+            $residualModules = $(docker -H npipe:////./pipe/iotedge_moby_engine ps -aq)
```
philipktlin: what is the type of $residualModules? array? or string? if it is array, can docker rm command handle it? tested with more than 1 module?

scripts/windows/test/Setup-Env.ps1
```diff
@@ -76,7 +82,15 @@ If ($osEdition -eq "IoTUAP")    # Windows IoT Core - update iotedge
...
+            if($residualModules.Length -gt 0) {
+                docker -H npipe:////./pipe/iotedge_moby_engine rm -f $residualModules
+            }
+        }
+        catch {
+            Write-Host "Cleanup existing containers failed. Nothing to clean."
```
philipktlin: Remove "Nothing to clean"; and should at least print the exception caught.

scripts/windows/test/Setup-Env.ps1
SlavyMihov: _$residualModules_ is array and the docker command handles it fine.
That command for initializing _$residualModules_ fails sometimes (I am still not sure why), that's why I have the try-catch around it.
_$residualModules = $(docker -H npipe:////./pipe/iotedge_moby_engine ps -aq)_

scripts/windows/test/Setup-Env.ps1
```diff
@@ -86,19 +100,22 @@ If ($osEdition -eq "IoTUAP")    # Windows IoT Core - update iotedge
...
-        try {
-            # triggers reboot
-            Update-IoTEdge -ContainerOs Windows -OfflineInstallationPath $IoTEdgedArtifactFolder
-        }
-        catch {
-            $testExitCode = [String]::Format("0x{0:X}", $LastExitCode)
-            Write-Host "Update attempt unsuccessful (error code $testExitCode)."
-            if ($LastExitCode -eq -2145877235) {
-                Write-Host "A newer version is already installed on the device."
+        if ($AttemptUpdate) {
```
philipktlin: in what scenario,  we clean up container and config file, but don't want to update/install iotedge service?

SlavyMihov: The E2E test stage runs 10 different tests one after another. It is only meaningful to attempt update before the very first test step (e.g. TempSensor), not do updates between the tests (e.g. TempSensor and TempFilte).
The E2E test stage for Windows Pro does cleanup between tests by uninstalling/reinstalling the iotedge service - that's my understanding.
We'd like to do cleanup in between the test steps on Windows IoT Core, so that stage behaves as close as possible to Windows Pro.

scripts/windows/test/Setup-Env.ps1
SlavyMihov: Done.

scripts/windows/test/Setup-Env.ps1
philipktlin: If you use attemptUpdate option to avoid install/update of iotedge service, you should not remove config.yaml as well.

scripts/windows/test/Setup-Env.ps1
philipktlin: it will fail if docker is not installed.

SlavyMihov: I see failures on the same machine that has docker installed, which is really strange.
See attempts 10 and 11 here: https://msazure.visualstudio.com/One/_releaseProgress?releaseId=971990&environmentId=2399693&_a=release-environment-logs

scripts/windows/test/Setup-Env.ps1
SlavyMihov: On Windows IoT Core I still need to clean up _config.yaml_ file in between tests otherwise _Run-E2ETest.ps1_ fails when run with parameter _BypassEdgeInstallation_.

PR #1473, opened 2019-07-24
Reorganize package scripts

arsing:
The new packaging script uses the corresponding distros' Docker images with
their own cross compilers and libraries, instead of our own ACR-hosted images.
Our images use the linaro cross-compiler and our own compiled openssl and
have not been modified or updated since they were created.

The exceptions are Debian 8 arm32v7, CentOS 7 arm32v7 and CentOS 7 aarch64.
Neither Debian 8 nor CentOS have cross-compiler packages in their repositories.

This commit also adds support for building Debian 10 packages. It also removes
the ability to build Debian 8 aarch64 packages, since Debian 8 LTS does not
support aarch64 (there is no official repo or ports repo even for OS packages).

By virtue of using the distro-provided cross-compiled dependency packages
or the linaro-compiled dependencies, it is not necessary to use `cross`.
So the new script installs the Rust toolchain and target inside
the build container, and uses that to build.

Other changes:

- The new script also fixes the generated packages to correctly depend on
  the corresponding distro's openssl package, so that it will be installed
  automatically when installing the libiothsm-std or iotedged packages.

- The iotedge package now depends on the libiothsm-std of
  the exact same version as itself. Previously it allowed any libiothsm-std
  package > 1.0.4, which was wrong since we do add/remove/change the API
  of the HSM library. Ref #1476

- This commit changes the `iotedged` packaging to not build
  all the crates under `edgelet/` but only the `iotedge` and `iotedged` crates,
  since only those crates' outputs are used in the package. This avoids
  building the Kubernetes-specific dependencies and saves build time,
  especially because `k8s-openapi` takes a few minutes just by itself
  to compile.

---

The bulk of the change is the new `edgelet/build/linux/package.sh` file that GitHub does not expand by default.

Packages build with this change: https://msazure.visualstudio.com/One/_build/results?buildId=23746788&view=logs

TODO:

- [x] Discussed with David and there should be no problem pushing the linaro images to the public Docker Hub so that anyone can build it. Will do that and change the `edgebuilds.azurecr.io` URLs.

- [ ] Is setting `--host-arch` and `--host-type` in `DPKGFLAGS` still the correct thing to do? Or should it be `--target-arch` ? The package seems fine either way.

- [x] `iotedge` Debian packages set `Multi-Arch: foreign`. I don't believe this is correct. Need to confirm.

- [x] The packages build generates some `libiothsm-std` packages without a `-$release` suffix. This is because `CPACK_DEBIAN_PACKAGE_RELEASE` is only understood by cmake 3.6+, but Ubuntu 16.04 has cmake 3.5 and Debian 8 has 3.0.


edgelet/build/linux/package.sh
```diff
@@ -0,0 +1,499 @@
...
+                # So don't build it at all.
+                ;;
+        esac
+
+        CMAKE_ARGS="$CMAKE_ARGS '-DOPENSSL_DEPENDS_SPEC=libssl1.0.0'"
+        ;;
+
+    'debian9')
+        DOCKER_IMAGE='debian:9-slim'
+        CMAKE_ARGS="$CMAKE_ARGS '-DOPENSSL_DEPENDS_SPEC=libssl1.1'"
```
darobs: >1.1 [](start = 62, length = 3)

This is a change from the previous build.  Was this on purpose?

arsing: Yeah. I didn't think there was a reason to keep it on 1.0

edgelet/build/linux/package.sh
```diff
@@ -0,0 +1,499 @@
...
+    'aarch64')
+        CMAKE_ARGS="$CMAKE_ARGS -DCPACK_DEBIAN_PACKAGE_ARCHITECTURE=arm64"
+        CMAKE_ARGS="$CMAKE_ARGS -DCPACK_RPM_PACKAGE_ARCHITECTURE=aarch64"
+
+        RUST_TARGET='aarch64-unknown-linux-gnu'
+        ;;
+esac
+
+if [ -n "$RUST_TARGET" ]; then
+    RUST_TARGET_COMMAND="rustup target add $RUST_TARGET &&"
```
darobs: >rustup target add [](start = 25, length = 17)

no need to install rust-src?  or was that needed for cross?

arsing: It's not even needed for cross unless you need to compile for an arch that needs xargo, which is none of the arches we used it for.

arsing:
This is ready for review. Packages build: https://msazure.visualstudio.com/One/_build/results?buildId=23878186

PR #1476, opened 2019-07-25
Bump libiothsm-std package dependency to >= 1.0.8

arsing:
As discussed offline, I'll make a proper fix for 1.0.9 that takes the right version automatically in #1473

Packages build: https://msazure.visualstudio.com/One/_build/results?buildId=23769980

### DIFF
--- a/edgelet/contrib/centos/iotedge.spec
+++ b/edgelet/contrib/centos/iotedge.spec
@@ -16,7 +16,7 @@ URL:            https://github.com/azure/iotedge
 %{?systemd_requires}
 BuildRequires:  systemd
 Requires(pre):  shadow-utils
-Requires:       libiothsm-std > 1.0.4
+Requires:       libiothsm-std >= 1.0.8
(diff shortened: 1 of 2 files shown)


arsing:
Verified the ubuntu-amd64 and centos7-amd64 packages. Installing `iotedge` while having an older `libiothsm-std` forces an upgrade of `libiothsm-std`.

```
$ apt install /home/arsing/Desktop/iotedge_1.0.820190725.2-1_amd64.deb

...

The following additional packages will be installed:
  libiothsm-std libssl1.0.0
The following NEW packages will be installed:
  iotedge libssl1.0.0
The following packages will be upgraded:
  libiothsm-std
1 upgraded, 2 newly installed, 0 to remove and 9 not upgraded.
```

The Windows build failed because the manifest is malformed. Not sure if it's transient or if Windows can't handle `REVISION <> 1`, but it shouldn't matter right now anyway.

PR #1477, opened 2019-07-25
Add end-to-end tests to the build

damonbarry:
This change adds a YAML build for the new end-to-end tests. It runs jobs on three agents: Ubuntu 16.04, Windows Server 2019, and Raspberry Pi 3.

The build can be run manually by specifying the build IDs to use for the "Edgelet Packages" and "Build Images" builds--both produce artifacts that this build requires. I will also set it up to trigger automatically when either of the "Edgelet Packages" or "Build Images" builds complete.

There is one set of cross-platform tasks, but they are divided into two template files; one has all the setup tasks and the other runs the tests and does post-run tasks like publishing test results. I split them because the Windows job needs to do a couple extra steps _after_ secrets have been downloaded from Key Vault, but _before_ the tests run. The Windows job calls the setup template, then does its own setup tasks, then calls the run template.


builds/e2e/e2e.yaml
```diff
@@ -0,0 +1,87 @@
...
+jobs:
+
+################################################################################	
+  - job: linux_arm32v7
+################################################################################	
+    displayName: Linux arm32v7
+
+    pool:
+      name: Azure-IoT-Edge-Core
+      demands: rpi3-e2e-tests
```
philipktlin: refer to https://github.com/Azure/iotedge/blob/338cc3e53ddd3afed841aa00cc6cac6fd5345fde/builds/e2e/stresstest.yaml#L101. 

```suggestion
      demands:
        - Agent.OS -equals Linux
        - Agent.OSArchitecture -equals ARM
        - run-e2e-tests -equals true
        - agent-osbits -equals 32
```

builds/e2e/e2e.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+jobs:
+
+################################################################################	
+  - job: linux_arm32v7
+################################################################################	
+    displayName: Linux arm32v7
+
+    pool:
+      name: Azure-IoT-Edge-Core
```
philipktlin: define a variable in the pipeline for re-use and configurable.

```suggestion
      name: $(pool.name)
```

builds/e2e/e2e.yaml
```diff
@@ -0,0 +1,87 @@
...
+      demands: rpi3-e2e-tests
+
+    variables:
+      os: linux
```
philipktlin: is it possible to use Agent.OS instead of creating a new variable "os"?

builds/e2e/templates/e2e-run.yaml
```diff
@@ -0,0 +1,54 @@
+steps:
```
philipktlin: is it better to use template parameter to pass values into template, such as $(arch)?

builds/e2e/templates/e2e-run.yaml
```diff
@@ -0,0 +1,54 @@
...
+    if ($IsWindows)
+    {
+      # The windows installer script doesn't update the PATH in the current
+      # terminal session, so add `iotedge` and `docker` CLIs manually
+      $env:Path =
+        $env:Path,
+        $(Join-Path $env:ProgramFiles 'iotedge'),
+        $(Join-Path $env:ProgramFiles 'iotedge-moby') -join ';'
+
+      dotnet vstest $testFile --logger:trx
```
philipktlin: will it be difficult to see which test is failed since all tests are running in 1 single step in vsts pipeline? is it possible to separate each tests as steps in vsts pipeline as before?

builds/e2e/e2e.yaml
damonbarry: Good call, will do.

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -0,0 +1,140 @@
...
+    dotnet build $testDir
+
+    $binDir = Convert-Path "$testDir/bin/Debug/netcoreapp2.1"
+    Write-Output "##vso[task.setvariable variable=binDir]$binDir"
+  displayName: Build tests
+
+- pwsh: |
+    $imagePrefix = '$(cr.address)/$(cr.labelPrefix)azureiotedge'
+    $imageId = Get-Content -Encoding Utf8 `
+      '$(System.ArtifactsDirectory)/$(az.pipeline.images.artifacts)/artifactInfo.txt'
```
philipktlin: where does $(az.pipeline.images.artifacts) come from?

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -0,0 +1,140 @@
...
+    buildVersionToDownload: $(whichImageBuild)
+    branchName: refs/heads/master
+    buildId: $(imageBuildId)
+    downloadType: specific
+    itemPattern: $(az.pipeline.images.artifacts)/artifactInfo.txt
+
+- pwsh: |
+    $certsDir = '$(System.ArtifactsDirectory)/certs'
+    New-Item "$certsDir" -ItemType Directory -Force | Out-Null
+    $env:ROOT_CERT | Out-File -Encoding Utf8 "$certsDir/rsa_root_ca.cert.pem"
```
philipktlin: where does $env:ROOT_CERT and $env:ROOT_KEy come from?

builds/e2e/e2e.yaml
damonbarry: Using the `Agent.*` variables is fine, as is using the custom `agent-osbits`. But I'm intentionally avoiding the `run-e2e-tests` demand since the existing end-to-end tests use it, and I don't want the old and new systems to get mixed together. (Technically none of the Raspberry Pi agents use that demand anyway so it wouldn't conflict, but I just want to avoid confusion while there are two different end-to-end systems in place.)

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -0,0 +1,140 @@
+steps:
+- checkout: self
+  clean: true
+  fetchDepth: 100
+
+- task: AzureKeyVault@1
```
philipktlin: AFAIK when use yaml with azure key vault task to set up VSTS pipeline, it will try to verify your credentail during save to ensure you can access that key vault.  However I have problem before, it seems doesn't detect key vault task inside a template during save; therefore credential is not verified and persisted, and this will cause failure when queuing build.  Do you have this problem?

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -0,0 +1,140 @@
...
+    elseif ('$(Build.TriggeredBy.DefinitionName)' -eq '$(azure.pipeline.images)')
+    {
+      Write-Output '##vso[task.setvariable variable=whichPackageBuild]latestFromBranch'
+      Write-Output '##vso[task.setvariable variable=whichImageBuild]specific'
+      Write-Output '##vso[task.setvariable variable=imageBuildId]$(Build.TriggeredBy.BuildId)'
+    }
+    # 3) Manual build (use given args for 'Build Images' build ID and 'Edgelet Packages' build ID)
+    elseif ('$(az.pipeline.packages.buildId)' -and '$(az.pipeline.images.buildId)')
+    {
+      Write-Output '##vso[task.setvariable variable=whichPackageBuild]specific'
```
philipktlin: should both of these use "latestFromBranch"?

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -0,0 +1,140 @@
...
+    $env:ROOT_KEY | Out-File -Encoding Utf8 "$certsDir/rsa_root_ca.key.pem"
+    Write-Output "##vso[task.setvariable variable=certsDir]$certsDir"
+  displayName: Install CA keys
+  env:
+    ROOT_CERT: $(TestRootCaCertificate)
+    ROOT_KEY: $(TestRootCaKey)
+
+- pwsh: |
+    $testDir = '$(Build.SourcesDirectory)/test/Microsoft.Azure.Devices.Edge.Test'
+    dotnet build $testDir
```
philipktlin:
```suggestion
    dotnet build $testDir -c Release
```

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+    $binDir = Convert-Path "$testDir/bin/Debug/netcoreapp2.1"
```
philipktlin:
```suggestion
    $binDir = Convert-Path "$testDir/bin/Release/netcoreapp2.1"
```

builds/e2e/e2e.yaml
damonbarry: Agent.OS won't work. On Linux, the value of Agent.OS is "Linux" with a capital "L", but I need lowercase for the docker image tag (probably a solvable problem, but not worth the effort). For Windows, I need "windows" and the variable returns "Windows_NT".

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+    if ($IsWindows)
+    {
+      $context['installerPath'] = Convert-Path '$(Build.SourcesDirectory)/scripts/windows/setup'
+    }
+
+    if ('$(arch)' -eq 'arm32v7')
+    {
+      $context['optimizeForPerformance'] = 'false'
+      $context['setupTimeoutMinutes'] = 8
```
philipktlin: Should some of these be configurable by pipeline variables? for example, it is easy to update variable in vsts pipeline to increase test timeout.

builds/e2e/templates/e2e-run.yaml
```diff
@@ -0,0 +1,54 @@
...
+  inputs:
+    testResultsFormat: vstest
+    testResultsFiles: '**/*.trx'
+    searchFolder: $(Build.SourcesDirectory)/TestResults
+    testRunTitle: End-to-end tests ($(Build.BuildNumber) $(os) $(arch))
+    buildPlatform: $(arch)
+  condition: succeededOrFailed()
+
+- pwsh: |
+    $logDir = '$(Build.ArtifactStagingDirectory)/logs'
```
philipktlin: just to confirm, $(Build.ArtifactStagingDirectory) will be flushed every time before build, right?

builds/e2e/templates/e2e-run.yaml
damonbarry: Yeah there's definitely an argument to be made for moving `os`, `arch`, and `artifactDir` out of the top-level jobs, and into the templates. But it would make the YAML a lot more verbose and cluttered--you'd have to specify the arguments to the templates at all six template calls, vs just three jobs. Since the two options are functionally equivalent, I went with the simpler, cleaner one.

builds/e2e/templates/e2e-run.yaml
damonbarry: Yes, the [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables) for `Build.ArtifactStagingDirectory` say:

> This directory is purged before each new build, so you don't have to clean it up yourself.

builds/e2e/templates/e2e-setup.yaml
damonbarry: No I didn't have this problem (but I remember when you did)...I guess the fixed it. ¯\_(ツ)_/¯

builds/e2e/templates/e2e-setup.yaml
damonbarry: Why? You get better callstacks in debug mode...

builds/e2e/e2e.yaml
philipktlin: Got it.

builds/e2e/templates/e2e-setup.yaml
philipktlin: I think e2e tests should be run as close as production; therefore I think release configuration should be used.  correct me if I'm wrong.

builds/e2e/templates/e2e-setup.yaml
damonbarry: All of the `az.pipeline.*` variables are build variables defined in the UI.

builds/e2e/templates/e2e-setup.yaml
damonbarry: They certainly could be, but I don't see the need to do that here (yet). If, for example, we had different kinds of arm32v7 agent hardware--some slower and some faster--then it could be a good idea.

builds/e2e/templates/e2e-setup.yaml
damonbarry: It's just the test code that's built in debug mode. The tests get iotedged from release packages (DEB/CAB), and Edge Agent/Hub are release docker images.

builds/e2e/e2e.yaml
damonbarry: Done

PR #1479, opened 2019-07-26
Register RequestHandler in FileConfigSourceModule

varunpuranik:


myagley:
This fails to start if the file config source is used?

PR #1488, opened 2019-07-29
Add HSM version checks for all interfaces

mrohera:
Explicitly fail if an incompatible HSM lib version was found on the target Edge device.


edgelet/edgelet-core/src/error.rs
```diff
@@ -54,6 +54,9 @@ pub enum ErrorKind {
     #[fail(display = "An identity manager error occurred.")]
     IdentityManager,
 
+    #[fail(display = "An error occurred when obtaining the Hsm version")]
```
arsing: `"the HSM version"` (upper-case HSM) to match the other strings.

PR #1489, opened 2019-07-29
Temporarily disable Windows ARM32 packages build.

arsing:
It has been failing since f012c28cdff2b0ac743e167c543d371b5a3269d0 ,
though based on the changes in that commit I don't think the commit itself
is the problem.

Not having a packages build prevents other builds from running. So this commit
disables that build until we can investigate the problem.


arsing:
(The compiler errors are for missing link symbols from Rust libstd. These symbols are from newer Rust releases than the special windows-arm32-targeting Rust that the Windows ARM32 job uses. So it's possible the windows-vs2016 image has a newer pre-installed Rust and it's somehow conflicting with the special one.)

PR #1494, opened 2019-07-30
EdgeHub loses cloud subscriptions if a feedback message is received twice. 

varunpuranik:
Scenario - 
If a client sends a message completion response twice, the second call gets misinterpreted as a C2D message response, and EdgeHub tries to send to the cloud. 
This causes the SDK to throw an exception and go into a bad state. The recovery logic re-creates the connection, but the subscriptions don't get re-set correctly.

Fixes https://github.com/Azure/iotedge/issues/1485


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessor.cs
```diff
@@ -44,7 +44,8 @@ public SubscriptionProcessor(IConnectionManager connectionManager, IInvokeMethod
             this.invokeMethodHandler = Preconditions.CheckNotNull(invokeMethodHandler, nameof(invokeMethodHandler));
             this.pendingSubscriptions = new ConcurrentDictionary<string, ConcurrentQueue<(DeviceSubscription, bool)>>();
             this.processSubscriptionsBlock = new ActionBlock<string>(this.ProcessPendingSubscriptions);
-            deviceConnectivityManager.DeviceConnected += this.DeviceConnected;
+            deviceConnectivityManager.DeviceConnected += this.CloudConnectivityEstablished;
```
damonbarry: I'm stumbling on the names here...the event handler that gets called when a device connects to Edge Hub is called `CloudConnectivityEstablished`? I seem to recall that the hub<->cloud connection is entirely separate from the device <->hub connection...

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
```diff
@@ -334,11 +345,53 @@ public static void ProcessedGetTwin(string identityId)
...
         }
 
         #region IDeviceProxy
 
-        public Task SendC2DMessageAsync(IMessage message) => this.underlyingProxy.SendC2DMessageAsync(message);
+        public Task SendC2DMessageAsync(IMessage message)
+        {
+            if (!message.SystemProperties.TryGetValue(SystemProperties.LockToken, out string lockToken) || string.IsNullOrWhiteSpace(lockToken))
+            {
+                // TODO - Should we throw here instead?
```
damonbarry: Is this TODO is something we should resolve before we merge this PR?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessor.cs
varunpuranik: Yeah, sorry the names are confusing -
CloudConnectivityEstablished = Edge device connects/reconnects to IoTHub
ClientCloudConnectivityEstablished = EdgeHub connects to IoTHub on behalf of a client

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
varunpuranik: Not necessarily.. I am not sure of the right behavior here. So added a comment. 
If you or someone else has any feedback on this, I would like to hear it..

damonbarry: OK, just double-checking.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessor.cs
damonbarry: Alright. Sounds like the first is a high-level status based on some heuristics. The second is the status of specific, physical connections.

PR #1496, opened 2019-07-31
EdgeHub: Fix error if client does not send product info

varunpuranik:
If client does not send product info, EdgeHub should simply send its own user agent string.


edge-util/src/Microsoft.Azure.Devices.Edge.Storage/SerDeExtensions.cs
```diff
@@ -10,7 +10,9 @@ public static T FromJson<T>(this string json)
         {
             if (string.IsNullOrWhiteSpace(json))
             {
-                return default(T);
+                return typeof(T) == typeof(string)
```
damonbarry: What does this do?

varunpuranik: When converting from Json to object, if the json is empty string, and object is string, then return the empty string (instead of default(string) which is null). This prevents null refs down the call stack

PR #1502, opened 2019-08-01
Add feature to build iotedged in Kubernetes mode.

darobs:
Use feature flags to switch between Docker and Kuberetes module runtimes.

Provide a new target in the Makefile so we can build a release version of iotedged in k8s mode.


edgelet/iotedged/src/app.rs
```diff
@@ -54,8 +60,7 @@ fn init_common(running_as_windows_service: bool) -> Result<Settings, Error> {
         }
     }
 
-    info!("Starting Azure IoT Edge Security Daemon");
-    info!("Version - {}", edgelet_core::version_with_source_version());
+    start_message();
```
arsing: Just do `if cfg!()` instead of a new function. As a bonus you wouldn't need to duplicate the `info!("Version")` line.

edgelet/iotedged/Cargo.toml
```diff
@@ -50,3 +53,8 @@ rand = "0.5"
 tempdir = "0.3.7"
 
 edgelet-test-utils = { path = "../edgelet-test-utils" }
+
+[features]
+default = ["runtime-docker"]
+runtime-docker = []
+runtime-kubernetes = ["edgelet-kube", "kube-client", "hyper-tls"]
```
arsing: Need newline at end of file.

edgelet/iotedged/src/unix.rs
```diff
@@ -1,14 +1,30 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-use edgelet_docker::DockerModuleRuntime;
-
 use crate::app;
 use crate::error::Error;
 use crate::signal;
 
+#[cfg(feature = "runtime-docker")]
```
arsing: The only thing you do with these imports is generate the `ModuleRuntime` alias.

You could save on a lot of typing by just fully-qualifying the imports in the type alias definition.

```rust
#[cfg(feature = "runtime-docker")]
type ModuleRuntime = edgelet_docker::DockerModuleRuntime;

#[cfg(feature = "runtime-kubernetes")]
type ModuleRuntime = edgelet_kube::KubeModuleRuntime<kube_client::ValueToken, ...>;
```

Same for the other file.

kubernetes/charts/edge-kubernetes/templates/_helpers.tpl
```diff
@@ -48,13 +48,14 @@ agent:
...
     auth:
-      username: {{ .Values.registryCredentials.username | quote }}
-      password: {{ .Values.registryCredentials.password | quote }}
+      username: {{ .Values.edgeAgent.registryCredentials.username | quote }}
+      password: {{ .Values.edgeAgent.registryCredentials.password | quote }}
+      serveraddress: {{ .Values.edgeAgent.registryCredentials.serveraddress | quote }}
     {{ else }}
     auth: {}
-    {{ end }}
+    {{ end }}
```
arsing: Unnecessary trailing whitespace.

scripts/linux/buildEdgelet.sh
```diff
@@ -177,7 +177,7 @@ build_project()
     # build project with cross
     cd "$EDGELET_DIR"
 
-    execute cross build -p "$PROJECT" "$BUILD_CONFIG_OPTION" --target "$TOOLCHAIN"
+    execute cross build -p "$PROJECT" --manifest-path=iotedged/Cargo.toml --no-default-features --features runtime-kubernetes "$BUILD_CONFIG_OPTION" --target "$TOOLCHAIN"
```
arsing: Maybe rename this script to have `kube(rnetes)` in the name so that it doesn't look like it's used for the regular builds.

edgelet/iotedged/src/app.rs
```diff
@@ -75,13 +80,30 @@ fn init_common(running_as_windows_service: bool) -> Result<Settings, Error> {
...
+    info!("Starting Azure IoT Edge Security Daemon");
+    info!("Version - {}", edgelet_core::version_with_source_version());
+}
+#[cfg(feature = "runtime-kubernetes")]
+fn start_message() {
+    info!("Starting Azure IoT Edge Security Daemon - Kubernetes mode");
+    info!("Version - {}", edgelet_core::version_with_source_version());
+}
+
+pub fn init<M>() -> Result<Settings, Error>
```
arsing: What's the point of taking `M` ? You're not doing any type-level specialization with it. The actual condition (choice of start message) is done via `cfg`.

darobs: It's an artifact from a different iteration (I was originally returning Result<M::Settings, Error>).  I can remove it.

scripts/linux/buildEdgelet.sh
arsing: Actually I discovered this is also used to build the `iotedge-diagnostics` binary for the regular `azureiotedge-diagnostics` image, so I guess it's fine as it is.

edgelet/iotedged/src/unix.rs
darobs: I think that looks better, overall as well. Done.

PR #1503, opened 2019-08-02
Update SDK Version in Edge to the version with ASC support

varunpuranik:
EdgeHub needs to support the ASC message. For that we need to update to a version of the SDK that has ASC support.


damonbarry:
What is ASC?

varunpuranik:
> What is ASC?

Azure Security Center

PR #1504, opened 2019-08-02
EdgeHub: Support for limiting the max storage space used by the internal storage

varunpuranik:
Currently there is no way to limit the amount of storage space EdgeHub uses on disk. This is problematic, especially in offline scenarios where the disk can get full leading to other unwanted side effects. 
This PR address that (at least partially) - 
- It allows 2 settings - max size in bytes, or threshold percentage of total disk space
- If that is exceeded, storing to the DB will fail and EdgeHub will throw a "Storage full" exception to connected clients

Behavior - 
- Feature is being an "experimental features flag", and is off by default.
- If turned on, the "threshold check" is enabled by default, with a default threshold of 98%. Default can be configured by setting the env var "MaxDiskUsagePercentage".
- The EdgeHub twin config supports a new optional parameter "maxStorageSpaceBytes". This can be set to the max bytes the EdgeHub db can take.

Rationale behind having the 2 configs in 2 different places - 
The idea is that you would use only the maxStorageSpaceBytes in the Twin config if you want to limit the bytes used by the DB. This is similar to the TTL knob in the twin config.
If not, you can leave the default threshold percentage based checker, which will limit to 98% usage (to avoid the disk from getting full). If for some reason, this default value doesn't work, you can configure it.


edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,150 @@
...
+                        }
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                Events.ErrorGettingMatchingDrive(storageFolder, e);
+            }
+
+            if (match != null)
```
arsing: I think this would be easier to read as `if (match == null) { Events.NoMatchingDriveFound(); return Option.None<>(); } Events.FoundDrive(); return Option.Maybe();` so that the two "success" branch statements are together.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/SequentialStore.cs
```diff
@@ -97,6 +97,41 @@ public async Task<bool> RemoveFirst(Func<long, T, Task<bool>> predicate, Cancell
...
         }
 
+        public async Task<int> RemovePercentFromHead(Func<long, T, Task> callback, int percentage, CancellationToken cancellationToken)
+        {
+            using (await this.headLockObject.LockAsync(cancellationToken))
+            {
+                long itemsToDelete = 0;
+                using (await this.tailLockObject.LockAsync(cancellationToken))
+                {
+                    // Tail offset could change here, but not holding a lock for efficiency.
```
arsing: This comment doesn't apply here since you *are* taking the tail object lock, unlike `RemoveFirst`.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/FixedSizeSpaceChecker.cs
```diff
@@ -0,0 +1,76 @@
...
+            var directory = new DirectoryInfo(directoryPath);
+            return GetDirectorySize(directory);
+        }
+
+        static long GetDirectorySize(DirectoryInfo directory)
+        {
+            long size = 0;
+
+            // Get size for all files in directory
+            FileInfo[] files = directory.GetFiles();
```
shantanu1singh: >GetFiles() [](start = 41, length = 10)

I believe you would need to use GetFiles("*", SearchOption.AllDirectories) here so that you get the size of files in subdirectories too.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/IMessageStore.cs
```diff
@@ -2,6 +2,7 @@
 namespace Microsoft.Azure.Devices.Routing.Core
 {
     using System;
+    using System.Threading;
```
shantanu1singh: >using System.Threading; [](start = 4, length = 23)

Accidental?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/SequentialStore.cs
```diff
@@ -97,6 +97,41 @@ public async Task<bool> RemoveFirst(Func<long, T, Task<bool>> predicate, Cancell
             }
         }
 
+        public async Task<int> RemovePercentFromHead(Func<long, T, Task> callback, int percentage, CancellationToken cancellationToken)
```
shantanu1singh: >RemovePercentFromHead [](start = 31, length = 21)

Doesn't look like this new API is being used anywhere?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -14,6 +15,9 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Service
     using Microsoft.Azure.Devices.Edge.Hub.Core.Identity;
     using Microsoft.Azure.Devices.Edge.Hub.Http;
     using Microsoft.Azure.Devices.Edge.Hub.Mqtt;
+    using Microsoft.Azure.Devices.Edge.Storage;
+    using Microsoft.Azure.Devices.Edge.Storage.RocksDb;
+    using Microsoft.Azure.Devices.Edge.Storage.RocksDb.Disk;
```
shantanu1singh: I suppose you don't need these new using statements anymore?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/disk/IDiskSpaceChecker.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Storage.Disk
+{
+    public interface IDiskSpaceChecker
```
shantanu1singh: >IDiskSpaceChecker [](start = 21, length = 17)

Since I'm considering using the same interface and base classes for the MemorySpaceChecker, it might be better to rename these files to something like IStorageSpaceChecker or something similar? Essentially change Disk to Storage (or something else)?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/CommonModule.cs
```diff
@@ -136,6 +150,18 @@ protected override void Load(ContainerBuilder builder)
...
                 .SingleInstance();
 
+            // IDiskSpaceChecker
+            builder.Register(
+                c =>
+                {
+                    IDiskSpaceChecker dickSpaceChecker = this.experimentalFeatures.EnableDiskSpaceCheck && this.usePersistentStorage
+                        ? DiskSpaceChecker.Create(this.storagePath, this.storageLimitThresholdPercentage, this.diskSpaceCheckFrequency)
+                        : new NullDiskSpaceChecker() as IDiskSpaceChecker;
+                    return dickSpaceChecker;
```
shantanu1singh: nit: rename var? :smile:

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        public bool IsFull => this.inner.DiskStatus == DiskStatus.Full;
+
+        public static DiskSpaceChecker Create(string storageFolder, double thresholdPercentage, TimeSpan checkFrequency)
+        {
+            Option<DriveInfo> drive = GetMatchingDrive(storageFolder);
+            Option<string> driveName = drive.Map(d => d.Name);
+            DiskSpaceCheckerBase inner = driveName.Map(d => new ThresholdDiskSpaceChecker(d, thresholdPercentage, checkFrequency, Events.Log) as DiskSpaceCheckerBase)
+                .GetOrElse(() => new DummyDiskSpaceChecker(checkFrequency, Events.Log));
```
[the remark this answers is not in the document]
varunpuranik: yes, good catch. fixed

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,150 @@
...
+            var diskSpaceChecker = new DiskSpaceChecker(storageFolder, driveName, checkFrequency, inner);
+            Events.Created(storageFolder);
+            return diskSpaceChecker;
+        }
+
+        public void SetThresholdPercentage(int thresholdPercentage)
+        {
+            string drive = this.drive.Expect(() => new InvalidOperationException("Cannot set disk usage threshold percentage since drive is not known"));
```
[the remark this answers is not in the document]
varunpuranik: I could.. but seems no better/worse either ways. So keeping it as it is. Lmk if you feel otherwise.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/ThresholdDiskSpaceChecker.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                this.Logger?.LogWarning($"High disk usage detected - using {percentDiskUsed}% of a maximum of {this.thresholdPercentage}% of drive {this.drive}");
+            }
+
+            return diskStatus;
+        }
+
+        static DiskStatus GetDiskStatus(double percentDiskUsed, double thresholdPercentage)
+        {
+            double usagePercentage = percentDiskUsed * 100 / thresholdPercentage;
```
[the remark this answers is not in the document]
varunpuranik: Yeah, I can understand how it can be confusing. But it is percentage of percentage. Any suggestions on how to make it better?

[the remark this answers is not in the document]
arsing: I'd keep it as a ratio, ie remove the `* 100` and compare with `0.85`, but I don't have a strong preference either way.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,150 @@
...
+            var drives = new DriveInfo[0];
+            try
+            {
+                drives = DriveInfo.GetDrives();
+                int segmentCount = -1;
+                foreach (DriveInfo drive in drives)
+                {
+                    if (drive.IsReady)
+                    {
+                        if (storageFolder.StartsWith(drive.Name, StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
arsing: Actually, it looks like `DriveInfo.GetDrives` would not return a drive mounted at `C:\Foo` at all, so this code would end up incorrectly using the `C:` drive's disk space instead.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/CommonModule.cs
```diff
@@ -136,6 +150,18 @@ protected override void Load(ContainerBuilder builder)
                 .As<IRocksDbOptionsProvider>()
                 .SingleInstance();
 
+            // IDiskSpaceChecker
+            builder.Register(
+                c =>
+                {
+                    IDiskSpaceChecker dickSpaceChecker = this.experimentalFeatures.EnableDiskSpaceCheck && this.usePersistentStorage
```
shantanu1singh: >EnableDiskSpaceCheck [](start = 83, length = 20)

nit: should rename this to EnableStorageSpaceCheck

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/FixedSizeSpaceChecker.cs
```diff
@@ -0,0 +1,69 @@
...
+            this.storageFolder = storageFolder;
+            this.maxSizeBytes = maxSizeBytes;
+            logger?.LogInformation($"Created fixed size space checker with max capacity of {maxSizeBytes} bytes on folder {storageFolder}");
+        }
+
+        protected override StorageStatus GetDiskStatus()
+        {
+            long bytes = GetDirectorySize(this.storageFolder);
+            double usagePercentage = (double)bytes * 100 / this.maxSizeBytes;
+            StorageStatus diskStatus = GetDiskStatus(usagePercentage);
```
shantanu1singh: >diskStatus [](start = 26, length = 10)

nit: rename to storageStatus?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/StorageSpaceChecker.cs
```diff
@@ -0,0 +1,149 @@
...
+namespace Microsoft.Azure.Devices.Edge.Storage.RocksDb.Disk
+{
+    using System;
+    using System.IO;
+    using System.Linq;
+    using Microsoft.Azure.Devices.Edge.Storage.Disk;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public class StorageSpaceChecker : IDiskSpaceChecker
```
shantanu1singh: >IDiskSpaceChecker [](start = 39, length = 17)

nit: rename to IStorageSpaceChecker?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/StorageSpaceChecker.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        public static StorageSpaceChecker Create(string storageFolder, double thresholdPercentage, TimeSpan checkFrequency)
+        {
+            Option<DriveInfo> drive = GetMatchingDrive(storageFolder);
+            Option<string> driveName = drive.Map(d => d.Name);
+            StorageSpaceCheckerBase inner = driveName.Map(d => new ThresholdStorageSpaceChecker(d, thresholdPercentage, checkFrequency, Events.Log) as StorageSpaceCheckerBase)
+                .GetOrElse(() => new NullStorageSpaceChecker(checkFrequency, Events.Log));
+            var diskSpaceChecker = new StorageSpaceChecker(storageFolder, driveName, checkFrequency, inner);
+            Events.Created(storageFolder);
+            return diskSpaceChecker;
```
shantanu1singh: >diskSpaceChecker [](start = 19, length = 16)

nit: rename to storageSpaceChecker?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/StorageSpaceChecker.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            Events.FoundDrive(storageFolder, match);
+            return Option.Maybe(match);
+        }
+
+        static class Events
+        {
+            public static readonly ILogger Log = Logger.Factory.CreateLogger<StorageSpaceChecker>();
+            const int IdStart = 50000;
```
shantanu1singh: >50000 [](start = 32, length = 5)

Might want to declare this in here - edge-util\src\Microsoft.Azure.Devices.Edge.Util\UtilEventsIds.cs

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/StorageSpaceChecker.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                this.drive.ForEach(d =>
+                {
+                    Events.SetMaxPercentageUsage(thresholdPercentage, d);
+                    this.inner = new ThresholdStorageSpaceChecker(d, thresholdPercentage, this.checkFrequency, Events.Log);
+                });
+            }
+        }
+
+        public void SetMaxDiskUsageSize(long maxDiskUsageBytes)
```
shantanu1singh: >maxDiskUsageBytes [](start = 45, length = 17)

nit: rename method name and parameter to indicate storage instead of disk?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/StorageSpaceChecker.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        static class Events
+        {
+            public static readonly ILogger Log = Logger.Factory.CreateLogger<StorageSpaceChecker>();
+            const int IdStart = 50000;
+
+            enum EventIds
+            {
+                Created = IdStart,
+                SetMaxSizeDiskSpaceUsage,
```
shantanu1singh: >SetMaxSizeDiskSpaceUsage [](start = 16, length = 24)

nit: rename to SetMaxSizeStorageSpaceUsage?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/StorageSpaceChecker.cs
```diff
@@ -0,0 +1,149 @@
...
+                SetMaxPercentageUsage,
+                FoundDrive,
+                NoMatchingDriveFound,
+                ErrorGettingMatchingDrive
+            }
+
+            public static void Created(string storageFolder)
+            {
+                Log.LogInformation((int)EventIds.Created, $"Created disk space usage checker for folder {storageFolder}");
```
shantanu1singh: >disk [](start = 68, length = 4)

nit: maybe change references to the word 'disk' from the logs to 'storage' as well?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/StorageStatus.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Storage.RocksDb.Disk
```
shantanu1singh: >Disk [](start = 55, length = 4)

nit: rename namespace to use 'Storage'?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/StorageSpaceCheckerBase.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        readonly PeriodicTask storageSpaceChecker;
+        readonly object updateLock = new object();
+
+        StorageStatus diskStatus;
+
+        protected StorageSpaceCheckerBase(TimeSpan checkFrequency, ILogger logger)
+        {
+            this.Logger = Preconditions.CheckNotNull(logger);
+            this.storageSpaceChecker = new PeriodicTask(this.PeriodicTaskCallback, checkFrequency, checkFrequency, logger, "Disk space check");
```
shantanu1singh: >Disk [](start = 124, length = 4)

nit: Storage*

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/StorageSpaceCheckerBase.cs
```diff
@@ -0,0 +1,61 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public abstract class StorageSpaceCheckerBase
+    {
+        readonly PeriodicTask storageSpaceChecker;
+        readonly object updateLock = new object();
+
+        StorageStatus diskStatus;
```
shantanu1singh: >diskStatus [](start = 22, length = 10)

nit: rename to storageStatus. Might be easier to just do a string replace for disk with storage in these files as there are other places I see references to disk too

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/ThresholdStorageSpaceChecker.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    class ThresholdStorageSpaceChecker : StorageSpaceCheckerBase
+    {
+        readonly string drive;
+        readonly double thresholdPercentage;
+
+        public ThresholdStorageSpaceChecker(string drive, double thresholdPercentage, TimeSpan checkFrequency, ILogger logger)
+            : base(checkFrequency, logger)
+        {
+            Preconditions.CheckArgument(thresholdPercentage >= 0 && thresholdPercentage <= 100, $"Invalid thresholdPercentage value {thresholdPercentage}");
```
shantanu1singh: >thresholdPercentage [](start = 106, length = 19)

{nameof(thresholdPercentage)}

edge-util/test/Microsoft.Azure.Devices.Edge.Storage.RocksDb.Test/disk/DiskSpaceCheckerTest.cs
```diff
@@ -0,0 +1,95 @@
...
+{
+    using System;
+    using System.IO;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Storage.RocksDb.Disk;
+    using Microsoft.Azure.Devices.Edge.Util.Test.Common;
+    using Xunit;
+
+    [Unit]
+    public class DiskSpaceCheckerTest
```
shantanu1singh: >DiskSpaceCheckerTest [](start = 17, length = 20)

nit: rename to StorageSpaceCheckerTest

dylanbronson: Leaving this one alone since it's a specific test for DiskSpaceChecker

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/ThresholdDiskSpaceChecker.cs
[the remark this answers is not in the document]
dylanbronson: Changed it to ratio

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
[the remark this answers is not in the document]
dylanbronson: Not being used at the moment anywhere. We will decide how to move forward (perhaps using p/invoke) at a later time.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/ThresholdDiskSpaceChecker.cs
```diff
@@ -14,7 +14,7 @@ class ThresholdDiskSpaceChecker : DiskSpaceCheckerBase
         public ThresholdDiskSpaceChecker(string drive, double thresholdPercentage, TimeSpan checkFrequency, ILogger logger)
             : base(checkFrequency, logger)
         {
-            Preconditions.CheckArgument(thresholdPercentage >= 0 && thresholdPercentage <= 100, $"Invalid thresholdPercentage value {thresholdPercentage}");
+            Preconditions.CheckArgument(thresholdPercentage >= 0 && thresholdPercentage <= 1, $"Invalid thresholdPercentage value {thresholdPercentage}");
```
arsing: So just to clarify, the comment I had made about percentages vs ratios was specifically for the `double usagePercentage = percentDiskUsed * 100 / thresholdPercentage;` calculation, since that was calculating a percentage of percentages. For `thresholdPercentage` and `percentDiskUsed`, keeping them as percentage is fine.

If you do want to change them to ratios too, then you should rename the variables to not have `percent` or `percentage` in the name. But I think it'd be better to keep them as percentages in the first place.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            if (match == null)
+            {
+                Events.NoMatchingDriveFound(drives, storageFolder);
+                return Option.None<DriveInfo>();
+            }
+
+            Events.FoundDrive(storageFolder, match);
+            return Option.Maybe(match);
```
vipeller: isn't it 'some' at this point?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/ThresholdDiskSpaceChecker.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            Preconditions.CheckArgument(thresholdPercentage >= 0 && thresholdPercentage <= 100, $"Invalid thresholdPercentage value {thresholdPercentage}");
+            this.drive = Preconditions.CheckNonWhiteSpace(drive, nameof(drive));
+            this.thresholdPercentage = thresholdPercentage;
+            logger?.LogInformation($"Created threshold percentage disk space checker with threshold of {thresholdPercentage}% of drive {drive}");
+        }
+
+        protected override DiskSpaceStatus GetDiskStatus()
+        {
+            var driveInfo = new DriveInfo(this.drive);
```
vipeller: same with the exception as above

edge-util/test/Microsoft.Azure.Devices.Edge.Storage.RocksDb.Test/disk/DiskSpaceCheckerTest.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        [Fact]
```
vipeller: Won't these tests kill the 'check in' tests on the build server? I am not sure that there will be access to the drive? Also these are not unit tests, more like integration. (so the [Unit] on the class might not be the right attribute

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/FixedSizeDiskSpaceChecker.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            this.storageFolder = storageFolder;
+            this.maxSizeBytes = maxSizeBytes;
+            logger?.LogInformation($"Created fixed size space checker with max capacity of {maxSizeBytes} bytes on folder {storageFolder}");
+        }
+
+        protected override DiskSpaceStatus GetDiskStatus()
+        {
+            long bytes = GetDirectorySize(this.storageFolder);
+            double usagePercentage = (double)bytes * 100 / this.maxSizeBytes;
```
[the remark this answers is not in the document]
dylanbronson: Added a check - if maxSizeBytes is 0, usagePercentage will be set to 100.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/FixedSizeDiskSpaceChecker.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                return DiskSpaceStatus.Critical;
+            }
+
+            return DiskSpaceStatus.Full;
+        }
+
+        static long GetDirectorySize(string directoryPath)
+        {
+            var directory = new DirectoryInfo(directoryPath);
```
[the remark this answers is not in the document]
dylanbronson: CommonModule creates a DiskSpaceChecker -> FixedDiskSpaceChecker -> DiskSpaceCheckerBase that creates a new PeriodicTask that does the work.

If an exception is thrown here it should bubble up to the original process and stop it. This will shut down EdgeHub, which is expected, since it is probably the user's fault for setting the config incorrectly.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/ThresholdDiskSpaceChecker.cs
dylanbronson: See above explanation

edge-util/test/Microsoft.Azure.Devices.Edge.Storage.RocksDb.Test/disk/DiskSpaceCheckerTest.cs
dylanbronson: They don't kill the check in tests on the build server, since the dotnet checkin's are passing. I do see the merit in making them integration rather than unit if they will slow everything down.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
dylanbronson: Yeah, looks like it

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -235,13 +239,18 @@ static string GetProductInfo()
...
             bool storeAndForwardEnabled = this.configuration.GetValue<bool>("storeAndForwardEnabled");
+
+            StoreLimits storeLimits = new StoreLimits(long.MaxValue, Option.Some(120));
             if (storeAndForwardEnabled)
             {
                 IConfiguration storeAndForwardConfigurationSection = this.configuration.GetSection("storeAndForward");
                 timeToLiveSecs = storeAndForwardConfigurationSection.GetValue("timeToLiveSecs", defaultTtl);
+                IConfiguration storeLimitsConfiguration = storeAndForwardConfigurationSection.GetSection("storeLimits");
+                int checkFrequency = storeLimitsConfiguration.GetValue<int>("checkFrequency");
+                storeLimits = new StoreLimits(long.MaxValue, Option.Maybe<int>(checkFrequency));
```
shantanu1singh: > Option.Maybe<int> [](start = 60, length = 18)

Just use Option.Some here

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -235,13 +239,18 @@ static string GetProductInfo()
...
             }
 
-            var storeAndForwardConfiguration = new StoreAndForwardConfiguration(timeToLiveSecs);
+            var storeAndForwardConfiguration = new StoreAndForwardConfiguration(timeToLiveSecs, Option.Maybe(storeLimits));
```
shantanu1singh: > Option.Maybe [](start = 95, length = 13)

Just use Option.Some here

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/CommonModule.cs
```diff
@@ -386,5 +409,13 @@ async Task<IAuthenticator> GetCloudTokenAuthenticator(IComponentContext context)
 
             return tokenAuthenticator;
         }
+
+        private Option<int> ExtractCheckFrequency(StoreAndForwardConfiguration storeAndForwardConfiguration)
+        {
+            long defaultMaxStorageSize = long.MaxValue;
+            int defaultCheckFrequency = 120;
+            StoreLimits storeLimits = storeAndForwardConfiguration.StoreLimits.GetOrElse(() => new StoreLimits(defaultMaxStorageSize, Option.Some(defaultCheckFrequency)));
```
shantanu1singh: >GetOrElse [](start = 79, length = 9)

Use .Map here and within the delegate inside it, just use GetOrElse on the argument of the delegate to get the value of CheckFrequency or return the default

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/disk/NullStorageSpaceChecker.cs
```diff
@@ -0,0 +1,24 @@
...
+    using Microsoft.Azure.Devices.Edge.Util;
+    public class NullStorageSpaceChecker : IStorageSpaceChecker
+    {
+        public void SetMaxStorageSize(long maxStorageBytes, Option<int> checkFrequency)
+        {
+        }
+
+        public void SetMaxStorageSizeAndCheckFrequency(long maxStorageBytes, Option<int> checkFrequency)
+        {
+        }
```
shantanu1singh: this can be removed

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/CommonModule.cs
dylanbronson: This is irrelevant now that it has been changed - since the checker won't start until the twin updates occur, we don't need to give it any configuration from env variables.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,181 @@
...
+    using System.IO;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Storage.Disk;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public class DiskSpaceChecker : IStorageSpaceChecker
+    {
+        static readonly int checkFrequencyDefault = 3;
```
shantanu1singh: >3 [](start = 52, length = 1)

This is a pretty low default imo. I think you should make it higher considering it involves disk I/O. 60-120 seconds might be good?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,181 @@
...
+            this.inner = Preconditions.CheckNotNull(inner, nameof(inner));
+        }
+
+        public bool IsFull => this.inner.Match<bool>(b => b.DiskStatus == DiskSpaceStatus.Full, () => false);
+
+        public static DiskSpaceChecker Create(string storageFolder)
+        {
+            // Start up diskSpaceChecker unfilled - so that it will be enabled when SetMaxSizeBytes is called
+            var diskSpaceChecker = new DiskSpaceChecker(storageFolder, Option.None<DiskSpaceCheckerBase>());
+            Events.Created(storageFolder);
```
shantanu1singh: This log might be misleading as the DiskSpaceChecker hasn't really been started. Consider moving this to the place where you actually 'start' the checker. You can rename it to Started or something if you'd like

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            lock (this.updateLock)
+            {
+                this.inner = Option.Some(
+                    new FixedSizeDiskSpaceChecker(
+                        this.storageFolder,
+                        this.maxSize,
+                        TimeSpan.FromSeconds(this.checkFrequency.GetOrElse(checkFrequencyDefault)),
+                        Events.Log)
+                    as DiskSpaceCheckerBase);
```
shantanu1singh: You should add a check before this to make sure we don't try to enable/initialize the checker again if it's already created.

if(!this.inner.HasValue)

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        public void SetCheckFrequency(Option<int> checkFrequencySecs)
+        {
+            checkFrequencySecs.ForEach(updatedCheckFrequency =>
+            {
+                if (updatedCheckFrequency > 0 && !this.checkFrequency.Equals(Option.Some(updatedCheckFrequency)))
+                {
+                    lock (this.updateLock)
+                    {
+                        this.checkFrequency = Option.Some(updatedCheckFrequency);
```
shantanu1singh: >Option.Some(updatedCheckFrequency) [](start = 46, length = 34)

nit: You can pre-created this before the if condition and reuse it both in the if condition itself and here.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
dylanbronson: Changed it

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
dylanbronson: Yep - remnant from some testing I was doing. Changed back to 120

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
dylanbronson: Added the check - which also trickled into changing a few other things.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceCheckerBase.cs
```diff
@@ -0,0 +1,73 @@
...
+                if (this.diskStatus > DiskSpaceStatus.Available)
+                {
+                    this.UpdateCurrentDiskSpaceStatus();
+                }
+
+                return this.diskStatus;
+            }
+        }
+
+        public abstract void SetMaxSizeBytes(long maxSizeBytes);
```
shantanu1singh: Don't think you need this here, since it's specific to MaxSizeBytes only, it should only be in the FixedSizeDiskSpaceChecker

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceCheckerBase.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public void DisposeChecker()
```
shantanu1singh: >DisposeChecker [](start = 20, length = 14)

Simply implement IDisposable and the Dispose method instead of using DisposeChecker as a custom method

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            // Start up diskSpaceChecker unfilled - so that it will be enabled when SetMaxSizeBytes is called
+            var diskSpaceChecker = new DiskSpaceChecker(storageFolder, Option.None<DiskSpaceCheckerBase>());
+            return diskSpaceChecker;
+        }
+
+        public void SetMaxSizeBytes(long maxSizeBytes)
+        {
+            this.maxSize = maxSizeBytes;
+            this.inner.ForEach(b => b.SetMaxSizeBytes(maxSizeBytes));
```
shantanu1singh: Instead of doing this I would change it to the following as you shouldn't expect this.inner to always be a FixedSizeDiskSpaceChecker given how someone might be switching from a different checker to the fixed size one:

1. Dispose original checker -> this.inner?.Dispose(). This will take care of stopping the old checker.
2. Set this.inner to a new FixedSizeDiskSpaceChecker instance.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,169 @@
...
+    using System.IO;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Storage.Disk;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public class DiskSpaceChecker : IStorageSpaceChecker
+    {
+        static readonly int checkFrequencyDefault = 120;
```
shantanu1singh: >static readonly [](start = 8, length = 15)

Can be a const instead and rename to begin with an uppercase C if you are changing it

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -129,6 +133,7 @@ void UpdateStoreAndForwardConfig(StoreAndForwardConfiguration storeAndForwardCon
             if (storeAndForwardConfiguration != null)
             {
                 this.messageStore?.SetTimeToLive(TimeSpan.FromSeconds(storeAndForwardConfiguration.TimeToLiveSecs));
+                storeAndForwardConfiguration.StoreLimits.ForEach(b => this.storageSpaceChecker.SetMaxSizeBytes(Option.Maybe<long>(b.MaxSize)));
```
shantanu1singh: This won't disable the DiskSpaceChecker if StoreLimits is Option.None as ForEach is only called when StoreLimits has a value.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/disk/IStorageSpaceChecker.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.Azure.Devices.Edge.Storage.Disk
+{
+    using System;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+    public interface IStorageSpaceChecker
+    {
+        void SetMaxSizeBytes(Option<long> maxStorageBytes);
+
+        void DisableChecker();
```
shantanu1singh: This can be removed now.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/disk/IStorageSpaceChecker.cs
```diff
@@ -0,0 +1,17 @@
...
+
+        void SetStorageUsageComputer(Func<Task<long>> storageUsageComputer);
```
shantanu1singh: >Func<Task<long>> [](start = 37, length = 16)

I have changed this to Func<long> now. Can you update it as well?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,160 @@
...
+        public static DiskSpaceChecker Create(string storageFolder, Option<int> checkFrequency)
+        {
+            // Start up diskSpaceChecker unfilled - so that it will be enabled when SetMaxSizeBytes is called
+            var diskSpaceChecker = new DiskSpaceChecker(storageFolder, Option.None<DiskSpaceCheckerBase>(), checkFrequency.GetOrElse(CheckFrequencyDefault));
+            return diskSpaceChecker;
+        }
+
+        public void SetMaxSizeBytes(Option<long> maxSizeBytes)
+        {
+            if (maxSizeBytes.HasValue)
```
shantanu1singh: I think you should be able to use-

this.inner = maxSizeBytes.Match(...);

where 'some' creates a new Option and returns it.
'none' returns Option.None.

You will need to dispose the original PeriodicTask for either case so place that before the match maybe. Also, you do need locking to account for parallel calls to the SetMaxSizeBytes method.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -129,6 +133,16 @@ void UpdateStoreAndForwardConfig(StoreAndForwardConfiguration storeAndForwardCon
...
                 this.messageStore?.SetTimeToLive(TimeSpan.FromSeconds(storeAndForwardConfiguration.TimeToLiveSecs));
+
+                if (storeAndForwardConfiguration.StoreLimits.HasValue)
+                {
+                    storeAndForwardConfiguration.StoreLimits.ForEach(b => this.storageSpaceChecker.SetMaxSizeBytes(Option.Some<long>(b.MaxSize)));
+                }
+                else
+                {
+                    this.storageSpaceChecker.SetMaxSizeBytes(Option.None<long>());
+                }
```
shantanu1singh: You should be able to use Option.Match here as well instead of an if-else block

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/StorageSpaceCheckConfiguration.cs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Service
+{
+    using System;
+    using Microsoft.Extensions.Configuration;
+
+    public class StorageSpaceCheckConfiguration
```
shantanu1singh: >StorageSpaceCheckConfiguration [](start = 17, length = 30)

Don't think this file is needed anymore

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/StorageConfiguration.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core.Storage
+{
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Newtonsoft.Json;
+
+    public class StorageConfiguration
```
shantanu1singh: Don't think this file is needed anymore

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/StoreLimits.cs
```diff
@@ -0,0 +1,62 @@
...
+    public class StoreLimits : IEquatable<StoreLimits>
+    {
+        [JsonConstructor]
+        public StoreLimits(long maxSize)
+            : this(maxSize, Option.None<int>())
+        {
+        }
+
+        public StoreLimits(long maxSize, Option<int> checkFrequency)
+        {
```
shantanu1singh: Add a comment here stating why this constructor and the checkFrequency property exist in this class so that no one accidentally ends up consuming it for other purposes outside of test

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/DependencyManager.cs
```diff
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.E2E.Test
     using Microsoft.Azure.Devices.ProtocolGateway.Instrumentation;
     using Microsoft.Extensions.Configuration;
     using Microsoft.Extensions.Logging;
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
```
shantanu1singh: not needed?

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/TestModule.cs
```diff
@@ -53,6 +53,9 @@ public Task SetupReceiveMessageHandler(string input)
         public Task<int> SendMessagesByCountAsync(string output, int startIndex, int count, TimeSpan timeout) =>
             this.SendMessagesByCountAsync(output, startIndex, count, timeout, TimeSpan.Zero);
 
+        public Task SendMessageBySize(string output, int size) =>
+            this.SendMessageAsync(output, this.GetMessageBySize(size));
```
shantanu1singh: don't think changes in this file are needed as part of your PR?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/disk/NullStorageSpaceChecker.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Azure.Devices.Edge.Storage.Disk
+{
+    using System;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
```
shantanu1singh: add empty line after this

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/disk/IStorageSpaceChecker.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Storage.Disk
+{
+    using System;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
```
shantanu1singh: add empty line after this

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceCheckerBase.cs
```diff
@@ -0,0 +1,71 @@
...
+
+                return this.diskStatus;
+            }
+        }
+
+        public void Dispose()
+        {
+            lock (this.updateLock)
+            {
+                this.storageSpaceChecker?.Dispose();
```
shantanu1singh: >? [](start = 40, length = 1)

Can't really be null as the constructor initializes it so you can remove the ? operator

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/disk/DiskSpaceCheckerBase.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
```
shantanu1singh: I don't think you need the locking here. You're already locking in the caller DiskSpaceChecker which should suffice. I think it's okay to make this non thread-safe and let the caller handle the thread-safety.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/config/ConfigUpdaterTest.cs
```diff
@@ -86,9 +87,10 @@ public async Task TestPeriodicConfigUpdate()
                 .ReturnsAsync(Option.Some(edgeHubConfig5))
                 .ReturnsAsync(Option.Some(edgeHubConfig6));
             configProvider.Setup(c => c.SetConfigUpdatedCallback(It.IsAny<Func<EdgeHubConfig, Task>>()));
+            var diskSpaceChecker = Mock.Of<IStorageSpaceChecker>();
 
             // Act
-            var configUpdater = new ConfigUpdater(router, messageStore.Object, updateFrequency);
+            var configUpdater = new ConfigUpdater(router, messageStore.Object, updateFrequency, diskSpaceChecker);
```
shantanu1singh: I have tests for this class in my PR, just letting you know so you don't need to write them

Issue #1435, opened 2019-07-11
Windows IoT Core: The daemon could not start up successfully: Could not load setting


conekeo:
## Expected Behavior
Being able to run IoT Edge on Windows IoT Core (build 17763)

## Steps to Reproduce
My IoT Edge device is a new device with default modules only.
As Administrator log on via Powershell to your Windows IoT Core device.
Run IoT Edge command:
```
 . {Invoke-WebRequest -useb https://aka.ms/iotedge-win} | Invoke-Expression;Install-IoTEdge -Manual -SkipBatteryCheck:$True -DeviceConnectionString 'HostName=xxxx.azure-devices.net;DeviceId=xxxTestEdge;SharedAccessKey=xxxxxxx' -RestartIfNeeded
```
Installation has no error.
```
ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     0.0        IoTEdge                             {Deploy-IoTEdge, Get-IoTEdgeLog, Initialize-IoTEdge, Insta...
The container host is on supported build version 17763.
Skipping VC Runtime installation on IoT Core.
Downloading IoT Edge...
Using IoT Edge from C:\Data\Users\administrator\AppData\Local\Temp\microsoft-azure-iotedge.cab
Committing changes, this will cause a reboot on success. If this is the first time installation, run "Initialize-IoTEdge" after the reboot completes.
```

After reboot and I got the following error in the log.
The daemon could not start up successfully: Could not load settings.  C:\Data\ProgramData\iotedge\config.yaml" not found

```
7/11/2019 3:21:04 PM info: iotedged::app -- Using config file: C:\Data\ProgramData\iotedge\config.yaml
7/11/2019 3:21:04 PM error: iotedged::windows -- Error while running service. Quitting.
7/11/2019 3:21:04 PM error: edgelet_utils::logging -- The daemon could not start up successfully: Could not load
                     settings
7/11/2019 3:21:04 PM error: edgelet_utils::logging --   caused by: Could not load settings
7/11/2019 3:21:04 PM error: edgelet_utils::logging --   caused by: configuration file
                     "C:\Data\ProgramData\iotedge\config.yaml" not found
```

Did I missed something?

### Device (Host) Operating System
Windows IoT Core v.10.0.17763.253 (build 17763)

### Architecture
AMD64

### Hardware
AcePC Mini PC T1 Plus with Intel Cherry Trail Atom Z8350

ancaantochi:
[...]

PR #1505, opened 2019-08-02
Windows installer fixes for IoT Core

### LINKED ISSUE #1404: PATH is not updated on IoTCore before second reboot for additional sessions
## Expected Behavior
After running Deploy-IoTEdge and Initialize-IoTEdge on IoTCore, I should be able to run iotedge list from any new remote PS session.

## Current Behavior
After running Deploy-IoTEdge and Initialize-IoTEdge on IoTCore, I can run iotedge list only from the first session, but not from additional sessions. This is because Set-SystemPath requires reboot on IoTCore to apply the change machine wide. 

Perhaps the Set-SystemPath can be moved into Deploy-IoTEdge, which needs to reboot on IoTCore anyway and as such, would initialize the path right away. A concern could be to make sure to delete the addition in case Deploy-IoTEdge fails.

arsing:
- Use `%PROGRAMDATA%` instead of `$(runtime.programData)` in the CAB manifest.
  On IoT Core, we want the services to use `C:\Data\ProgramData` since this is
  the data partition. `%PROGRAMDATA%` points to this. `$(runtime.programData)`
  does expand to this as well, if the CAB is installed manually with
  `Deploy-IoTEdge`, but expands incorrectly to `C:\ProgramData` if the CAB is
  installed as part of imaging the device.

  Windows confirmed that this discongruity is expected, and that we should
  avoid using `$(runtime.programData)`. Since Windows does support using
  cmd-format env var references in the image path of a service, we can avoid
  the problem by using `%PROGRAMDATA%` as that is expanded at service runtime.

  Ref: #1435

- Add the IoT Edge and Moby install directories to PATH in `Deploy-IoTEdge`
  rather than in `Initialize-IoTEdge`. On IoT Core both `Deploy-` and updating
  PATH require a reboot, so it's good to combine those reboots into one.

  Fixes #1404


jiria:
Not directly related, by I think Set-SystemPath should be updated. Since edgeinstallpath is a prefix of mobyinstallpath, the detection logic in the Set-SystemPath is not working as intended.

arsing:
(Discussed offline, that `Set-SystemPath` does work as intended.)

PR #1506, opened 2019-08-02
Fix quickstart test for when devices are not explicitly created

mrohera:


smoke/IotEdgeQuickstart/details/Details.cs
```diff
@@ -585,7 +599,8 @@ string BuildImageName(string name)
 
     public class DeviceContext
     {
-        public Device Device;
+        public Option<Device> Device;
+        public string DeviceId;
```
philipktlin: Since both Device and DeviceId are public, Device.Id and DeviceId can be different.  I think it is better to make them as properties and make sure to keep them in sync.

PR #1509, opened 2019-08-03
Move log parsing functionality to edgelet core

philipktlin:
Move log parsing functionality from iotedge/logs.rs to edgelet-core/logs.rs; therefore it can be used in snitcher.


arsing:
Commit message doesn't say why. I assume you're doing this because you need it in snitcher. Then mention that in the commit message.

I think `-http-mgmt` is a better place than `-core`, since `-http-mgmt` is what produces these logs in the first place.

philipktlin:
> Commit message doesn't say why. I assume you're doing this because you need it in snitcher. Then mention that in the commit message.
> 
> I think `-http-mgmt` is a better place than `-core`, since `-http-mgmt` is what produces these logs in the first place.

Talked with Mike, he think this can be used

myagley:
@arsing I think this makes sense in core. The trait in core defines "get logs" as a stream of byte arrays. However, these aren't really opaque byte arrays. They all have structure. This was done to optimize the case where we are piping data directory from the docker engine out through the mgmt api.

in cases, where the daemon code needs access to the log lines, we need to parse this byte stream. I think it makes sense that this live next to the ModuleRuntime trait because it is really bound to this trait, and is not specific to http.

edgelet/edgelet-core/src/lib.rs
```diff
@@ -16,6 +16,7 @@ mod certificate_properties;
 pub mod crypto;
 mod error;
 mod identity;
+pub mod logs;
```
myagley: You probably don't need this `pub` if you are re-exporting the types below.

arsing:
Okay, makes sense.

PR #1516, opened 2019-08-05
Claire/logs_diag

Claire-Zheng:
Api calls for module logs and diagnostics:
- /api/modules/{module_id}/logs: Displays the entire log of the specified module
- /api/diagnostics: Displays result of `iotedge check` in json format


edge-modules/edge-dashboard/src/lib.rs
```diff
@@ -229,6 +234,75 @@ fn restart_module(
...
+    let module_id = req.match_info().get("id").unwrap_or("");
+    let api_ver = &info.api_version;
+
+    let response = context
+        .edge_config
+        .as_ref()
+        .map(move |config| {
+            let mgmt_uri = config.connect().management_uri();
+            Either::A(
+                Url::parse(&format!("{}/modules/?api-version={}", mgmt_uri, api_ver))
```
dmolokanov: This url parsing happens almost for every api handler that makes a call to iotedged. Should we make it as a parameter of Context?

edge-modules/edge-dashboard/src/lib.rs
```diff
@@ -229,6 +234,75 @@ fn restart_module(
...
+                    .map_err(ErrorInternalServerError)
+                    .and_then(|url| ModuleClient::new(&url).map_err(ErrorInternalServerError)) // can't connect to the endpoint
+                    .map(move |mod_client| {
+                        mod_client
+                            .logs(module_id, &LogOptions::new())
```
dmolokanov: I assume that `client.logs("", ...)` will return error. Shouldn't we prevent calling the client with empty module name earlier, when we take it from request params?

edge-modules/edge-dashboard/src/lib.rs
```diff
@@ -229,6 +234,75 @@ fn restart_module(
...
+        });
+
+    Box::new(response)
+}
+
+fn get_diagnostics() -> HttpResponse {
+    let output = Command::new("iotedge")
+        .args(&["check", "--output", "json"])
+        .output()
+        .expect("Failed to execute command");
```
darobs: Rather than calling expect, perhaps generate a repsonse:
```Rust
fn get_diagnostics() -> HttpResponse {
    Command::new("iotedge")
        .args(&["check", "--output", "json"])
        .output()
        .map( |o|  HttpResponse::Ok().body(o.stdout))
        .unwrap_or_else( |_| HttpResponse::ServiceUnavailable().body("Failed to execute command"))
}
```

edge-modules/edge-dashboard/src/modules.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    Box::new(response)
+}
+
+pub fn get_logs(
+    req: HttpRequest,
+    context: web::Data<Arc<Context>>,
+    info: web::Query<AuthRequest>,
+) -> Box<dyn Future<Item = HttpResponse, Error = ActixError>> {
+    let module_id = req.match_info().get("id").unwrap_or("");
```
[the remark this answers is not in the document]
dmolokanov: something like this
```
pub fn get_logs(
    req: HttpRequest,
    context: web::Data<Arc<Context>>,
    info: web::Query<AuthRequest>,
) -> Box<dyn Future<Item = HttpResponse, Error = ActixError>> {
    let api_ver = &info.api_version;
    let response = req
        .match_info()
        .get("id")
        .map(|module_id| {
            context
                .edge_config
                .as_ref()
                .map(move |config| {
                    let mgmt_uri = config.connect().management_uri();
                    Either::A(
                        Url::parse(&format!("{}/modules/?api-version={}", mgmt_uri, api_ver))
                            .map_err(ErrorInternalServerError)
                            .and_then(|url| {
                                ModuleClient::new(&url).map_err(ErrorInternalServerError)
                            }) // can't connect to the endpoint
                            .map(move |mod_client| {
                                mod_client
                                    .logs(module_id, &LogOptions::new())
                                    .map_err(ErrorInternalServerError)
                                    .and_then(|data| {
                                        data.map_err(ErrorInternalServerError)
                                            .fold(BytesMut::new(), |mut acc, chunk| {
                                                let stream = chunk.as_ref();
                                                if stream.len() >= 8 {
                                                    let (_, right) = stream.split_at(8);
                                                    acc.extend_from_slice(right);
                                                }
                                                Ok::<_, ActixError>(acc)
                                            })
                                            .and_then(|body| {
                                                let mut clone = body.freeze().to_vec().clone();
                                                clone.retain(|&byte| (byte as char).is_ascii());
                                                if let Ok(content) = String::from_utf8(clone) {
                                                    HttpResponse::Ok().body(content)
                                                } else {
                                                    HttpResponse::ServiceUnavailable()
                                                        .body("Logs unable to be displayed")
                                                }
                                            })
                                    })
                            })
                            .into_future()
                            .flatten(),
                    )
                })
                .unwrap_or_else(|err| {
                    Either::B(ok(HttpResponse::ServiceUnavailable()
                        .content_type("text/plain")
                        .body(format!("{:?}", err))))
                })
        })
        .unwrap_or_else(|| {
            Either::B(ok(
                HttpResponse::BadRequest().body("A module identifier required")
            ))
        });

    Box::new(response)
}
```

edge-modules/edge-dashboard/src/modules.rs
```diff
@@ -0,0 +1,235 @@
...
+
+use actix_web::error::ErrorInternalServerError;
+use actix_web::Error as ActixError;
+use actix_web::*;
+use bytes::BytesMut;
+use edgelet_core::{LogOptions, Module as EdgeModule, ModuleRuntime, RuntimeSettings};
+use edgelet_http_mgmt::*;
+use futures::future::{ok, Either, IntoFuture};
+use futures::stream::Stream;
+use futures::Async;
```
dmolokanov:
```suggestion
use futures::{Async, Future};
```

edge-modules/edge-dashboard/src/modules.rs
```diff
@@ -0,0 +1,235 @@
...
+use futures::Future;
```
dmolokanov:
```suggestion
```

edge-modules/edge-dashboard/src/state.rs
```diff
@@ -97,7 +97,7 @@ pub fn get_file() -> Result<String> {
     return handle_windows();
 
     #[cfg(target_os = "linux")]
```
dmolokanov:
```suggestion
    #[cfg(unix)]
```

edge-modules/edge-dashboard/src/modules.rs
```diff
@@ -0,0 +1,235 @@
...
+    return_modules(context, &info.api_version, health_response)
+}
+
+fn health_response(mods: Vec<Module>) -> HttpResponse {
+    let mut device_status = Status::new();
+    let mut edge_agent = false;
+    let mut edge_hub = false;
+    let mut other = true;
+
+    for module in mods.iter() {
```
dmolokanov: I think this loop does too many things. What about making it more single responsible?

```
fn health_response(mods: Vec<Module>) -> HttpResponse {
    let edgeagent = mods
        .iter()
        .any(|module| module.name() == "edgeAgent" && module.status() == "running");

    let edgehub = mods
        .iter()
        .any(|module| module.name() == "edgeHub" && module.status() == "running");

    let other = mods.iter().any(|module| module.status() != "running");

    let mut device_status = Status::new();
    device_status.set_iotedged();
    device_status.set_edge_agent(edgeagent);
    device_status.set_edge_hub(edgehub);
    device_status.set_other_modules(edgeagent && edgehub && other);

    let health = device_status.return_health();
    HttpResponse::Ok().body(format!(
        "Device health: {:?}\nDevice details: {:?}",
        health, device_status
    ))
}
```
some benefits:
- you don't need mutate variables anymore
- each statement is declarative and basically states what should be done not how it should be done
- a little bit shorter

edge-modules/edge-dashboard/src/status.rs
```diff
@@ -0,0 +1,69 @@
...
+use std::process::Command;
+
+use actix_web::error::ErrorInternalServerError;
+use actix_web::Error as ActixError;
+use actix_web::*;
+
+use crate::state::{return_response, Device};
+
+pub fn get_state(device: web::Data<Option<Device>>) -> HttpResponse {
+    if let Some(dev) = device.get_ref() {
```
dmolokanov: I noticed that you have multiple places where you are using `if let ... else ... ` constructs. It's not necessarily a bad thing it works fine obviously. But it looks like c code, not idiomatical rust code. Rust employs pattern matching for it.

edge-modules/edge-dashboard/Cargo.toml
```diff
@@ -14,7 +16,7 @@ os_info = "1.1.1"
 reqwest = "0.9.18"
 serde = "1.0"
 serde_derive = "1.0"
-serde_json = "1.0"
+serde_json = "1.0.40"
```
dmolokanov: Why this change?

edge-modules/edge-dashboard/src/modules.rs
```diff
@@ -0,0 +1,235 @@
...
+                Url::parse(&format!("{}/modules/?api-version={}", mgmt_uri, api_ver))
+                    .map_err(ErrorInternalServerError)
+                    .and_then(|url| ModuleClient::new(&url).map_err(ErrorInternalServerError)) // can't connect to the endpoint
+                    .map(move |mod_client| {
+                        mod_client
+                            .logs(module_id, &LogOptions::new())
+                            .map_err(ErrorInternalServerError)
+                            .and_then(|data| {
+                                data.map_err(ErrorInternalServerError)
+                                    .fold(BytesMut::new(), |mut acc, chunk| {
```
dmolokanov: Just curious why you chose `BytesMut` over `mut Vec<u8>`?

edge-modules/edge-dashboard/Cargo.toml
Claire-Zheng: Is there a drawback in having it 1.0.40 instead of 1.0? Just wondering

dmolokanov: 1.0 means any version with major part 1.0 and any build number. let's say you put 1.0.40 and another crate will use 1.0 rust will compile 2 `serde_json` crates. `cargo build` will check the latest version of `serde_json` on crates.io and for the second crate will use the latest one and for your crate still 1.0.40.

But we have Cargo.lock file that specifies all versions of all crates in a workspace. So instead of going on crates.io `cargo build` will use whatever version specified in the file.

So if you specify version without build section, all crates hopefully will use the same version of dependency.

More information in [cargo book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)

PR #1521, opened 2019-08-06
[k8s] Replace config map replace to list + create or update

dmolokanov:
k8s API doesn't `upsert` semantic for a replace config map operation. Changed replace to list config maps and create new or replace existing config map with a trust bundle if required.


arsing:
Why was this checked in when it fails with compile errors?

arsing:
Okay, I see you fixed it in #1532

PR #1524, opened 2019-08-06
[k8s] Attach trust bundle as a volume to EA

dmolokanov:
- Attach ConfigMap as a volume to pod spec
- Specify volume to trust bundle as a container spec


darobs:
Do we need to do the same in c#?

dmolokanov:
> Do we need to do the same in c#?

Yes. I've already have assigned on me tasks for it

PR #1525, opened 2019-08-07
[k8s] Edge proxy

dmolokanov:
- Edge proxy organize secure tls communication channel between modules and `iotedged` in k8s mode
- It has config.yaml file where user can specify where to proxy requests to
- Initial version support proxy from `http` to `https` endpoints
- A proxy can use certs from trust bundle as a root cas
- By default, it is aware of k8s environment and can use k8s token and supply it along with an initial request
- A proxy supports the graceful shutdown of all services on `SIGINT` and `SIGTERM`
- It exposes health endpoint for k8s to make health checks


edgelet/iotedge-proxy/docker/linux/amd64/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM alpine:3.9
```
darobs: Should we run this proxy as a less privileged user?

edgelet/iotedge-proxy/src/main.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use iotedge_proxy::{app, Error, Routine};
+use iotedge_proxy::{logging, signal};
```
darobs: Can lines 3&4 be combined?

edgelet/iotedge-proxy/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+            let mut builder = X509::builder()?;
+            builder.set_version(2)?;
+            builder.set_subject_name(&name)?;
+            builder.set_issuer_name(&name)?;
+            builder.set_not_before(Asn1Time::days_from_now(0)?.as_ref())?;
+            builder.set_not_after(Asn1Time::days_from_now(365)?.as_ref())?;
+            builder.set_pubkey(&pkey)?;
+
+            let basic_constraints = BasicConstraints::new().critical().ca().build()?;
```
kkdawkins: Can you comment some of this? Like it would be good to know what is happening here. Example - what is "critical"?

edgelet/iotedge-proxy/src/logging.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use std::env;
+
+use failure::Fail;
+use log::{log, Level, LevelFilter};
+
+pub fn init() {
+    env_logger::Builder::new()
+        .filter_level(LevelFilter::Info)
```
kkdawkins: Should our default level be higher?

edgelet/iotedge-proxy/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "iotedge-proxy"
+description = "iotedge-proxy is a HTTPS proxy server to organize secure edge modules communication"
+version = "0.1.0"
+authors = ["Azure IoT Edge Devs"]
```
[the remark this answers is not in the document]
dmolokanov: Took it from iotedge-diagnostics project

edgelet/iotedge-proxy/src/api.rs
```diff
@@ -0,0 +1,76 @@
...
+pub struct ApiService;
+
+impl ApiService {
+    pub fn new() -> Self {
+        ApiService
+    }
+
+    fn handle(&self, req: &Request<Body>) -> Result<Response<Body>, Error> {
+        match (req.method(), req.uri().path()) {
+            (&Method::GET, "/health") => Ok(Response::new(Body::empty())),
```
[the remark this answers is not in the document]
dmolokanov: Intended for k8s to make liveness probe. once it cannot get 200 response - it'll re-create pod. 
But good idea. Maybe we should ping `iotedged` instead? k8s itself will not be able to make a liveness probe because `iotedged` exposes only https endpoints and k8s doesn't have a valid cert.

@darobs

edgelet/iotedge-proxy/docker/linux/arm32v7/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine:3.10
```
darobs: From arm32v6/alpine:3.10?

edgelet/iotedge-proxy/docker/linux/arm64v8/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine:3.10
```
darobs: from arm64v8/alpine:3.10?

edgelet/iotedge-proxy/src/api.rs
```diff
@@ -0,0 +1,76 @@
...
+    type ReqBody = Body;
+    type ResBody = Body;
+    type Error = Compat<Error>;
+    type Future = Box<dyn Future<Item = Response<Self::ResBody>, Error = Self::Error> + Send>;
+
+    fn call(&mut self, req: Request<Self::ReqBody>) -> Self::Future {
+        let request = format!("{} {} {:?}", req.method(), req.uri(), req.version());
+        debug!("Starting api request {}", request);
+
+        let fut = self
```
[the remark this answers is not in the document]
dmolokanov: In my opinion, Box::new(self.....dozens of lines....) makes redundant nesting, where we can easily reduce it with a variable. More nesting - harder to read code.

edgelet/iotedge-proxy/src/app.rs
```diff
@@ -0,0 +1,46 @@
...
+        });
+
+    let settings = Settings::new(config_file)?;
+
+    Ok(settings)
+}
+
+fn create_app() -> App<'static, 'static> {
+    App::new(crate_name!())
+        .author(crate_authors!("\n"))
```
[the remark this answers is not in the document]
dmolokanov: It's not blank line, it's a separator :) It will read all authors from toml file and make a list with it

edgelet/iotedge-proxy/src/error.rs
```diff
@@ -0,0 +1,160 @@
...
+            fail = cause;
+        }
+
+        let body = json!({
+            "message": message,
+        })
+        .to_string();
+
+        Response::builder()
+            .status(StatusCode::INTERNAL_SERVER_ERROR)
```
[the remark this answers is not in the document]
dmolokanov: I wanted to return some meaningful error message when https client has an error when making a request to `iotedged`. If do nothing at all hyper simply closes the connection with HTTP client in case of error. Not sure how to improve it. Thoughts?

edgelet/iotedge-proxy/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+                    .unwrap_or(&"localhost".to_string()),
+            )?;
+            let name = name.build();
+
+            let mut builder = X509::builder()?;
+            builder.set_version(2)?;
+            builder.set_subject_name(&name)?;
+            builder.set_issuer_name(&name)?;
+            builder.set_not_before(Asn1Time::days_from_now(0)?.as_ref())?;
+            builder.set_not_after(Asn1Time::days_from_now(365)?.as_ref())?;
```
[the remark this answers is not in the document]
dmolokanov: It's a cert generation for tests. should make it 1 day instead. or even an hour

edgelet/iotedge-proxy/src/proxy/client.rs
```diff
@@ -0,0 +1,202 @@
...
+        self.config
+            .host()
+            .join(req.uri().path_and_query().map_or("", |p| p.as_str()))
+            .map_err(Error::from)
+            .and_then(|url| {
+                // set a full URL to redirect request to
+                *req.uri_mut() = url.as_str().parse()?;
+
+                // set host value in request header
+                if let Ok(host) = req.uri().host().unwrap_or_default().parse() {
```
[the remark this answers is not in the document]
dmolokanov: In the line above I assign req.uri with a new uri that I compose from whatever url set in settings. host part can be empty, and Host::default() returns empty which is a valid value for host header in the request.

edgelet/iotedge-proxy/src/proxy/client.rs
```diff
@@ -0,0 +1,202 @@
...
+    config: Config<T>,
+    client: S,
+}
+
+impl<T> Client<T, HyperHttpClient<HttpsConnector<HttpConnector>>>
+where
+    T: TokenSource,
+{
+    pub fn new(config: Config<T>) -> Self {
+        let mut http = HttpConnector::new(4);
```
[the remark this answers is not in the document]
dmolokanov: Nope, it's a default value that hyper use under the hood. Probably need to add comment

edgelet/iotedge-proxy/src/proxy/client.rs
```diff
@@ -0,0 +1,202 @@
...
+        http.enforce_http(false);
```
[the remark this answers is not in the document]
dmolokanov: Agree. Not very convenient api

edgelet/iotedge-proxy/src/logging.rs
dmolokanov: that is what iotedged itself uses right now. we can set ENV PROXY_LOG=debug as needed

edgelet/iotedge-proxy/src/app.rs
```diff
@@ -0,0 +1,57 @@
...
+        .map_or_else(
+            || {
+                let path = Path::new(settings::DEFAULT_SETTINGS_FILEPATH);
+                if path.exists() {
+                    Some(path)
+                } else {
+                    None
+                }
+            },
+            |name| Some(Path::new(name)),
```
darobs: Should you check if path exists in this as well?

edgelet/iotedge-proxy/Cargo.toml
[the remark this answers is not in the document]
darobs: We've been using "Azure IoT Edge Devs" as the author for all the crates in this project.

edgelet/iotedge-proxy/docker/linux/amd64/Dockerfile
dmolokanov: Good idea. Make the `proxy` user instead root

edgelet/iotedge-proxy/src/error.rs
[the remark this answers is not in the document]
kkdawkins: I think that we should return something more meaningful via HTTP status codes. A 500 means that the server (in this case, the proxy) has failed, but in the example you gave, iotedged failed.

IMHO, I would use the following:
502 - iotedged is down
401/403 - certificate error
500 - general proxy error unrelated to iotedged

edgelet/iotedge-proxy/src/routine.rs
```diff
@@ -0,0 +1,293 @@
...
+        .and_then(move |addr| {
+            let new_service = ApiService::new();
+
+            let server = Server::bind(&addr)
+                .serve(new_service)
+                .with_graceful_shutdown(shutdown)
+                .map_err(Error::from);
+
+            info!(
+                "Listening on {} with 1 thread for api",
```
kkdawkins: Is 1 thread enough? What if this module is high volume?

edgelet/iotedge-proxy/src/settings.rs
```diff
@@ -0,0 +1,248 @@
...
+use url::Url;
+use url_serde;
+
+use crate::{Error, ErrorKind};
+
+pub const DEFAULTS: &str = include_str!("../config/default.yaml");
+
+pub const DEFAULT_SETTINGS_FILEPATH: &str = "/etc/iotedge-proxy/config.yaml";
+
+const TOKEN_FILEPATH: &str = "/var/run/secrets/kubernetes.io/serviceaccount/token";
```
kkdawkins: Can the user configure the location of the token in Kubernetes? If so, we should take that as config.

edgelet/iotedge-proxy/src/signal.rs
```diff
@@ -0,0 +1,76 @@
...
+                SIGINT => "SIGINT",
+                SIGTERM => "SIGTERM",
+                other => return write!(f, "signal {}", other),
+            };
+            f.write_str(s)
+        }
+    }
+}
+
+#[cfg(not(unix))]
```
kkdawkins: Did you test this on Windows? I have had problems in the past when iotedge was running as a Windows Service.

edgelet/iotedge-proxy/src/app.rs
dmolokanov: It will be checked inside `Settings::new(&path)`. An error will be returned in this case.

edgelet/iotedge-proxy/src/routine.rs
dmolokanov: As far as I understand, 1 thread is going to listen to new clients to connect. Once it connected tokio will work with it in a separate task that can be scheduled on any thread from thread pool. I think it should be fine here.

edgelet/iotedge-proxy/src/settings.rs
dmolokanov: Yes. User can override it from the config file. If it is empty there, k8s location will be used.

edgelet/iotedge-proxy/src/signal.rs
dmolokanov: We are not targeting windows for k8s work so far. I added this code just to make it compile in windows along with other parts of edgelet. Should try it though

edgelet/iotedge-proxy/src/lib.rs
dmolokanov: I have plans to add integration test and I think some parts of this code will change.

edgelet/iotedge-proxy/src/error.rs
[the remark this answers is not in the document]
dmolokanov: 502 and 500 - good point. 
Certificate error will be returned by iotedged, so we can pass response as is, I think

edgelet/iotedge-proxy/src/lib.rs
kkdawkins: I still think that some comments might be useful :) Even if just talking about major chunks. What if our DRI needs to debug something off hours?

edgelet/iotedge-proxy/src/proxy/client.rs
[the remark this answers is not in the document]
kkdawkins: Is 4 the right value for us though?

kkdawkins:
I'm good with this PR @dmolokanov  :shipit:

edgelet/iotedge-proxy/src/api.rs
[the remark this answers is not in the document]
darobs: For liveness, I'm inclined to think that being able to receive a request and respond is adequate, until we know any better. It would be nice if we could check on the module container's status, but I don't think we have a consistent liveness method for a module.

Talking to iotedged seems more like a "readiness" check.

PR #1529, opened 2019-08-07
Ascsupport

nirlavi:


edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/DeviceClientMessageConverterTest.cs
```diff
@@ -156,7 +157,7 @@ public void TestFromMessage_AllSystemProperties()
             Assert.Equal("foo", clientMessage.To);
             Assert.Equal("1234", clientMessage.CorrelationId);
             Assert.Equal("m1", clientMessage.MessageId);
-            Assert.Equal(creationTime, clientMessage.CreationTimeUtc.ToString("o"));
```
varunpuranik: We probably don't want to remove this Assert..

PR #1531, opened 2019-08-07
Initialize EdgeHub before starting the Protocol heads

varunpuranik:
The periodic task to pull EdgeHub config runs asynchronously from the main thread. This means that the Protocol heads could start before EdgeHub has received its config and initialized. 
In that case there is a possibility that a client connects and sends messages before the routing engine has been initialized. This PR fixes that.


edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/config/ConfigUpdaterTest.cs
```diff
@@ -89,10 +89,15 @@ public async Task TestPeriodicConfigUpdate()
...
-            configUpdater.Init(configProvider.Object);
+            await configUpdater.Init(configProvider.Object);
 
             // Assert
-            await Task.Delay(TimeSpan.FromSeconds(8));
+            configProvider.Verify(c => c.GetConfig(), Times.Once);
+            endpointExecutorFactory.Verify(e => e.CreateAsync(It.IsAny<Endpoint>()), Times.Once);
+            messageStore.Verify(m => m.SetTimeToLive(It.IsAny<TimeSpan>()), Times.Once);
+
+            // After 5 seconds, the periodic task should not have run.
```
damonbarry: `TestPeriodicConfigUpdate` runs `Init`, verifies the config, then waits 5 seconds and verifies it again. In light of that logic, I don't understand this comment:
```
// After 5 seconds, the periodic task should not have run.
```

If the periodic task _did_ run during that 5-second sleep, would the config somehow be different the second time you check it?

damonbarry: Ah, nevermind. These Verify statements aren't checking the contents of the config, they're checking that certain functions on the mock objects were called exactly once. So this confirms that the periodic task hasn't run yet. 👍

varunpuranik: Right, it just makes sure that we are not pulling the config too often

PR #1314, opened 2019-06-08
Workaround Moby bug where unclean shutdown prevents containers from being restarted

arsing:
If Moby exits uncleanly, any containers left behind cannot be started any more
since the container VHD is help open by `System`. Specifically,
the start container API endpoint returns an error with the message:

    hcsshim::ActivateLayer failed in Win32: The process cannot access the file because it is being used by another process. (0x20)

Since this prevents the Edge Agent container from starting, the entire device
becomes unusable until a reboot.

Another way to recover such a stuck container is to delete it. This obviously
loses any data in the container's filesystem unless it was volume-mounted
from the host, but we already have such recommendations for production.

Therefore, with this change, iotedged detects this error response from
the Docker start container API and automatically deletes
the corresponding container. It then returns the original error back.

In case the stuck container was Edge Agent, the watchdog will eventually
try to start Edge Agent again, which will then succeed.

In case the stuck container was a module container, Edge Agent will
back off due to the error, and eventually retry, which will then succeed.

This change also adds a check to `iotedge check` to verify that
the storage directories of the Edge Agent and Edge Hub containers are
persisted on the host filesystem. This is required so that deleting either of
these containers does not cause data loss.

Finally, this change adds a restart-service failure-recovery action for
the Moby service.

PR #1536, opened 2019-08-07
Add checks that Edge Agent and Edge Hub storage directories are mounted from the host

arsing:
Also fix the type of the `docker::ContainerConfig::volumes` field.
The swagger spec for this field has the wrong type. It should be
a map of container directory (string) to empty object.

Upstream fixed this in
https://github.com/moby/moby/commit/dc883c0486b398eb5ad99f35aef3ff02a5a7dd29

---

Extracted from https://github.com/Azure/iotedge/pull/1314 as discussed offline.


edgelet/iotedge/src/check/mod.rs
```diff
@@ -1460,6 +1472,79 @@ fn container_engine_logrotate(check: &mut Check) -> Result<CheckResult, failure:
...
+        .map(Path::new);
+
+    if !mounted_directories
+        .chain(volume_directories)
+        .any(|container_directory| container_directory == storage_directory)
+    {
+        return Ok(CheckResult::Warning(
+            Context::new(format!(
+                "The {} module is not configured to persist its {} directory on the host filesystem.\n\
+                Data might be lost if the module is deleted or updated.",
```
varunpuranik: Should this give them a pointer on how to do it? Maybe a URL to our docs.

edgelet/iotedge/src/check/mod.rs
```diff
@@ -1460,6 +1472,79 @@ fn container_engine_logrotate(check: &mut Check) -> Result<CheckResult, failure:
...
+
+    let volume_directories = inspect_result
+        .config()
+        .and_then(docker::models::ContainerConfig::volumes)
+        .map(std::collections::HashMap::keys)
+        .into_iter()
+        .flatten()
+        .map(Path::new);
+
+    if !mounted_directories
```
varunpuranik: nice.. thx

edgelet/iotedge/src/check/mod.rs
arsing: Yeah, I was going to talk to Venkat about it when I saw him.

PR #1552, opened 2019-08-11
fix schedule to stop iotedge for stress test

philipktlin:
Currently only schedule time is passed to schedule task command; which will fail the deployment if schedule time is over midnight; the schedule time will be past since start date is not provide and uses today by default.  Therefore fix schedule script to provide both start date and time for scheduling.


philipktlin:
> f 4.5 hours from now is tomorrow, will `schtasks /create /sc once` fail to run

Yes, it will fail the deployment.  Since schedule date is not provide, it uses default (today) and the time is at a.m., but current time is at p.m. today; so it is past.  In order to fix it, I need to pass start date as well.

PR #1558, opened 2019-08-12
Add EdgeAgent identifier to product info

ancaantochi:
Similar to EdgeHub productInfo, add EdgeAgent identifier to product info


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/ModuleClientProvider.cs
```diff
@@ -61,7 +61,7 @@ public class ModuleClientProvider : IModuleClientProvider
             ISdkModuleClientProvider sdkModuleClientProvider,
             Option<UpstreamProtocol> upstreamProtocol,
             Option<IWebProxy> proxy,
-            Option<string> productInfo,
+            string productInfo,
```
philipktlin: Do you need to check not null for productInfo since type is changed from option<string> to string?

varunpuranik: +1 for null check, esp if moduleClient.SetProductInfo throws

PR #1559, opened 2019-08-13
Show certificate expiry in self-signed certs warning

arsing:
Also fix DPS identity cert expiry check to return Ignored (success)
instead of Skipped (error) if the user is not using DPS X.509 attestation.


edgelet/iotedge/src/check/mod.rs
```diff
@@ -1305,53 +1287,100 @@ fn settings_identity_certificates_expiry(check: &mut Check) -> Result<CheckResul
...
+                    Please consider using production certificates instead. See https://aka.ms/iotedge-prod-checklist-certs for best practices.",
+                    not_after,
+                ))
+                .into(),
+            ));
+        } else {
+            return Ok(CheckResult::Warning(
+                Context::new(format!(
+                    "Device is using self-signed automatically-generated development certificates.\n\
+                    They will expire at {} (in {} days) causing module-to-module and downstream device communication to fail on an active deployment.\n\
```
veyalla: Suggest wording this as: "They will expire in {x} days ({date-time})...."

edgelet/iotedge/src/check/mod.rs
```diff
@@ -1305,53 +1287,100 @@ fn settings_identity_certificates_expiry(check: &mut Check) -> Result<CheckResul
...
+            "Device CA certificate",
+            check.device_ca_cert_path.as_ref().unwrap(),
+        )?;
+
+        let now = chrono::Utc::now();
+
+        if not_after < now {
+            return Ok(CheckResult::Warning(
+                Context::new(format!(
+                    "Device is using self-signed automatically-generated development certificates.\n\
```
mrohera: Consider:  "The Edge device is ...

mrohera: Here and the next comment as well.

edgelet/iotedge/src/check/mod.rs
arsing: Done.

edgelet/iotedge/src/check/mod.rs
arsing: Done.

PR #1561, opened 2019-08-13
Update deploy-iotuap.yaml

SlavyMihov:
Fix bug when checking for Stress test.
Overwrite artifacts when copying to runner machine.


builds/e2e/longhaul.yaml
```diff
@@ -260,6 +260,7 @@ jobs:
           downloadPath: '$(Build.StagingDirectory)'
           artifactName: $(images.artifact.name.windows)
           itemPattern: |
+            $(images.artifact.name.windows)/CACertificates/*
```
philipktlin: Latest update in runE2Etests.ps1 committed to master; we should not need CACertificates files for stress or long haul test.

philipktlin: Please get the latest runE2ETests.ps1 and verify.  thanks.

PR #1566, opened 2019-08-13
Fix race between tests that try to bind to a port

arsing:
The original code would create a TcpListener and throw it away, which meant
that multiple calls to `get_unused_port` could end up with the same port.

Also, enable all the edgelet-docker tests that had been disabled on Windows
due to the original issue (even though it was not a Windows-specific problem).

However this means that the number of tests in the edgelet-docker crate running
at the same time is much higher, so it's more likely for the tests to hit
the issue described in https://github.com/rust-lang/backtrace-rs/issues/230#issuecomment-521036540
Specifically, any test that panics at the same time as another test is
using `failure` to collect a backtrace has a chance of crashing the entire
test binary with STATUS_HEAP_CORRUPTION on Windows, if the `RUST_BACKTRACE`
env var is set.

To avoid this, I've modified the tests that assert that an operation must fail
via `#[should_panic]` + `Result::unwrap` to instead use `Result::unwrap_err`
I've only done this for edgelet-docker since it has a large number of
concurrent tests so it has a very high chance of hitting this, in comparison to
the other crates that use `#[should_panic]`

### DIFF
--- a/edgelet/edgelet-docker/src/module.rs
+++ b/edgelet/edgelet-docker/src/module.rs
@@ -28,6 +28,15 @@ pub struct DockerModule<C: Connect> {
     config: DockerConfig,
 }
 
+impl<C> std::fmt::Debug for DockerModule<C>
+where
+    C: Connect,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DockerModule").finish()
+    }
+}
+
 impl<C: 'static + Connect> DockerModule<C> {
     pub fn new(client: DockerClient<C>, name: String, config: DockerConfig) -> Result<Self> {
         ensure_not_empty_with_context(&name, || ErrorKind::InvalidModuleName(name.clone()))?;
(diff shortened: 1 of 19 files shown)


arsing:
>I've only done this for edgelet-docker since it has a large number of
>concurrent tests so it has a very high chance of hitting this, in comparison to
>the other crates that use `#[should_panic]`

I ended up having to do everything else too, because I also got crashes in `edgelet-iothub` and `hsm`.

There are still three `#[should_panic]` tests left in `hsm`, but they can't be fixed in the same way because the panics there come from `Buffer::drop` when exiting the test function scope.

edgelet/edgelet-docker/src/module.rs
```diff
@@ -28,6 +28,15 @@ pub struct DockerModule<C: Connect> {
     config: DockerConfig,
 }
 
+impl<C> std::fmt::Debug for DockerModule<C>
+where
+    C: Connect,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DockerModule").finish()
```
darobs: Won't this just display "DockerModule { }" - do we care about the fields?

arsing: All these `Debug` impls are there to satisfy `Result::unwrap_err`, since that requires the `Ok` variant to impl `Debug` in case `unwrap_err` needs to panic.

I couldn't `#[derive(Debug)]` any of them because their fields aren't `Debug` either. Eventually I would have had to modify the API traits in the Docker crate to also be bound on `Debug`, at which point I gave up and created this empty impl.

It wasn't getting used before, and even now it doesn't get used outside of tests. Even if the tests panic because something returned `Ok` instead of `Err`, I don't think the internals of the struct would be particularly useful; you already know which test failed and can look it up from there.

darobs: Sounds good, thanks for the explanation.

PR #1574, opened 2019-08-15
EdgeHub: Fix receiving messages over MQTT

varunpuranik:
Issue: ProtocolGateway rejects messages if the message creation time is > Subscription time. 
This worked fine till now since the Message creation time was always set to UtcNow. But because of a change elsewhere in the codebase, this started getting set to EnqueuedTime. So in some cases, if the subscription came after the message was enqueued, the message was rejected by the PG. 
One other issue was that the EdgeHub treated rejected messages as retryable, so it kept retrying the same message. 

Fix: 
- Set message creation time for all messages sent to PG to DateTime.UtcNow
- Discard rejected messages.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
@@ -102,7 +102,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Device
                 }
                 else
                 {
-                    taskCompletionSource.SetException(new EdgeHubIOException($"Message not completed by client {this.Identity.Id}"));
+                    taskCompletionSource.SetException(new EdgeHubMessageRejectedException($"Message not completed by client {this.Identity.Id}"));
                 }
             }
             else if (this.c2dMessageTaskCompletionSources.TryRemove(messageId, out bool value) && value)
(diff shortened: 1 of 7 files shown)


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
```diff
@@ -102,7 +102,7 @@ public async Task ProcessMessageFeedbackAsync(string messageId, FeedbackStatus f
                 }
                 else
                 {
-                    taskCompletionSource.SetException(new EdgeHubIOException($"Message not completed by client {this.Identity.Id}"));
+                    taskCompletionSource.SetException(new EdgeHubMessageRejectedException($"Message not completed by client {this.Identity.Id}"));
```
philipktlin: is it possible that feedback status is aborted here? if so, why throw Rejected Exception?

varunpuranik: We don't support Aborted. Plus the idea here is to not throw an IOException

philipktlin: And if new exception is thrown here, this message will not be retried, why not retry?

varunpuranik: If the message is rejected, then it should not be retried. Else we will get into a loop of retrying the same message and it being rejected.

PR #1575, opened 2019-08-15
Fix build for new Rust stable 1.37

arsing:


arsing:
To not have to `use std::convert::TryFrom;`

Issue #1486, opened 2019-07-29
Facing Problem while installing IoT edge for windows IoT core with TPM key


arnabhetc:
Hi, I am trying to install azure IoT edge on aeon GW, running Windows IoT core 

I am using this link - https://docs.microsoft.com/en-us/azure/iot-edge/how-to-install-iot-edge-windows

there are two commands 

- one for the installation of IoT edge 

- the second one is for the DPS, with keys, 

generally while performing the second step, it asks for scope id and registration id and then built config.yaml file, 

what is happening?
It's not asking for the registration, it only asks for the scope id and then starts building the config.yaml file and gives the error that registration id is not there, but it never asked for it so, a

once the config.yaml is built with this half information, then there is no way to change it for IoT core I need to reinstall the IoT core image, I have tried twice showing the same thing 

I am attaching the logs for the details

[192.168.1.106]: PS C:\Data\Users\administrator\Documents> . {Invoke-WebRequest -useb https://aka.ms/iotedge-win} | Invoke-Expression; `
>> Initialize-IoTEdge -Dps

ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     0.0        IoTEdge                             {Deploy-IoTEdge, Get-IoTEdgeLog, Initialize-IoTEdge, Insta...

cmdlet Initialize-IoTEdge at command pipeline position 1
Supply values for the following parameters:
**ScopeId: 0ne*******F**
**The container host is on supported build version 17763.
Generating config.yaml..**.
**Initialize-IoTEdge : Exception calling "Invoke" with "1" argument(s): "RegistrationId is required for this DPS**
provisioning mode."
At line:2 char:1
+ Initialize-IoTEdge -Dps
+ ~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Initialize-IoTEdge], MethodInvocationException
    + FullyQualifiedErrorId : RuntimeException,Initialize-IoTEdge 

**I have already done this multiple times, never faced this issue, It was working with edge version- 1.0.7.1 and we are running gateway with the edge ,  but this week after the release of 1.0.8 facing this issue
**

@iotedge-ci @azure-keyvault  
<img width="960" alt="install issue" src="https://user-images.githubusercontent.com/47913360/62137780-28f69300-b304-11e9-8747-b68eb0deaf26.PNG">

arsing:
[...]

PR #1582, opened 2019-08-16
Add support for manual X.509 provisioning

mrohera:
To complement the DPS X.509 identity flow, this feature adds support to enable a user to configure a device that is manually created in IoT Hub and uses any X.509 certificate auth modes. In addition to the core feature and tests, the following was also implemented: 
* Iotedge check tool was enhanced to check the idenity certificate for expiration
* Windows installer script was modified to support installation and configuration when using manual mode of operation


edgelet/iotedged/src/lib.rs
```diff
@@ -384,16 +386,46 @@ where
...
+                        start_edgelet!(
+                            key_store,
+                            provisioning_result,
+                            root_key,
+                            force_module_reprovision,
+                            None,
+                        );
+                    }
+                    ManualAuthMethod::X509(x509) => {
+                        info!("Starting provisioning edge device via manual mode using X509 identiy certificate...");
```
shantanu1singh: >identiy [](start = 92, length = 7)

nit: identity*

edgelet/iotedged/src/lib.rs
```diff
@@ -384,16 +386,46 @@ where
...
+
+                        let id_data = device_cert_identity_data.ok_or_else(|| {
+                            ErrorKind::Initialize(InitializeErrorReason::DpsProvisioningClient)
+                        })?;
+
+                        let key_bytes = hybrid_identity_key.ok_or_else(|| {
+                            ErrorKind::Initialize(InitializeErrorReason::DpsProvisioningClient)
```
shantanu1singh: >DpsProvisioningClient [](start = 73, length = 21)

nit: This should be ManualProvisioningClient

shantanu1singh:
> const DPS_DEVICE_ID_KEY_ENV_KEY: &str = "IOTEDGE_DEVICE_IDENTITY_PK";

nit: might be worth renaming the variables now to be something more generic instead of using DPS explicitly as it's used for manual mode too?

---
Refers to: edgelet/iotedged/src/lib.rs:182 in ffa5d0f. [](commit_id = ffa5d0f93096465310a92cc710dec5e80358a052, deletion_comment = False)

edgelet/contrib/config/linux/config.yaml
```diff
@@ -16,11 +16,33 @@
...
 #     external - the device has been provisioned externally.
 #                Uses an external provisioning endpoint to get device specific information.
 #
+# Manual settings when using IoT Hub connection string
+#     device_connection_string - Required. The Edge device connection string
+#                      when using SharedAccessKey authentication mode.
+#                      Ex. "HostName=<hub-name>.azure-devices.net;DeviceId=<device-id>;SharedAccessKey=<key>
+#
+# Manual settings when using a X.509 identity certificate
+#     hub_hostname    - Required. The Azure Iot Hub hostname.
```
myagley: I think we either need to call this `iothub_hostname` or just `hostname`. The first is more explicit. From a style perspective I think we need to always refer to a hub as an iot hub. The second option is consistent with the connection string.

edgelet/contrib/config/linux/config.yaml
```diff
@@ -32,14 +54,14 @@
 #                       provisioning devices configured for symmetric key
 #                       attestation. Device specific symmetric key.
 #     identity_cert   - Optional. The Edge device identity X.509 certificate
-#                       entry should only be specified when provisioning
-#                       an Edge device configured for X.509 attestation.
+#                       entry should only be specified when an Edge device
```
myagley: This is a huge nit, but can we lowercase the `Edge`?

edgelet/contrib/config/linux/debian/config.yaml
```diff
@@ -16,11 +16,33 @@
...
 #     external - the device has been provisioned externally.
 #                Uses an external provisioning endpoint to get device specific information.
 #
+# Manual settings when using IoT Hub connection string
+#     device_connection_string - Required. The Edge device connection string
+#                      when using SharedAccessKey authentication mode.
+#                      Ex. "HostName=<hub-name>.azure-devices.net;DeviceId=<device-id>;SharedAccessKey=<key>
+#
+# Manual settings when using a X.509 identity certificate
+#     hub_hostname    - Required. The Azure Iot Hub hostname.
```
myagley: This comment makes it look like these keys are directly under `provisioning`, but the example shows them as under `provisioning.authentication`?

edgelet/iotedged/src/lib.rs
```diff
@@ -384,16 +386,46 @@ where
...
+                        })?;
+
+                        let (key_store, provisioning_result, root_key) = manual_provision_x509(
+                            x509,
+                            &mut tokio_runtime,
+                            &key_bytes,
+                            id_data.thumbprint.clone(),
+                        )?;
+                        let thumprint_op = Some(id_data.thumbprint.as_str());
```
shantanu1singh: >thumprint_op [](start = 28, length = 12)

nit: thumbprint_op*

edgelet/contrib/config/linux/config.yaml
mrohera: The actual settings calls this "iothub_hostname" and I have fixed the comments section. Thanks for pointing this discrepancy.

edgelet/contrib/config/linux/debian/config.yaml
mrohera: Take a look and let me know if this reads better.

edgelet/contrib/config/linux/config.yaml
mrohera: Resolving per our convo :).

edgelet/iotedged/src/lib.rs
mrohera: Fixed. Thanks.

mrohera:
@shantanu1singh I renamed the variable and removed the DPS_DEVICE and made this generic. Good feedback.

edgelet/contrib/config/linux/config.yaml
```diff
@@ -16,11 +16,33 @@
 # Configures the identity provisioning mode of the daemon.
 #
 # Supported modes:
-#     manual   - using an iothub connection string
+#     manual   - using an iothub connection string or a X.509 identity certificate
```
arsing: `a X.509` -> `an X.509`

edgelet/edgelet-core/src/settings.rs
```diff
@@ -61,44 +112,113 @@ impl Manual {
...
+            device_connection_string: Option<String>,
+            #[serde(skip_serializing_if = "Option::is_none")]
+            authentication: Option<ManualAuthMethod>,
+        }
+
+        let value: Inner = serde::Deserialize::deserialize(deserializer)?;
+
+        let authentication = match (value.device_connection_string, value.authentication) {
+            (Some(cs), Some(auth)) => {
+                if cs == DEFAULT_CONNECTION_STRING {
```
arsing: It's cleaner to just remove the `device_connection_string` from the `default.yaml`s instead of using it as a sentinel like this. Then you can remove this case.

edgelet/contrib/config/linux/config.yaml
```diff
@@ -49,11 +71,21 @@
 #                information such as its IoT hub connection information.
 ###############################################################################
 
-# Manual provisioning configuration
+# Manual provisioning configuration using a connection string
 provisioning:
   source: "manual"
   device_connection_string: "<ADD DEVICE CONNECTION STRING HERE>"
```
arsing: Do we want to recommend people to move to the new format for regular mode too? If so, change this too.

```
authentication:
  method: 'device_connection_string'
  device_connection_string: '...'
```

edgelet/edgelet-core/src/settings.rs
```diff
@@ -61,44 +112,113 @@ impl Manual {
...
+                    ));
+                }
+            }
+            (Some(cs), None) => {
+                ManualAuthMethod::DeviceConnectionString(ManualDeviceConnectionString::new(cs))
+            }
+            (None, Some(auth)) => auth,
+            (None, None) => {
+                return Err(serde::de::Error::custom(
+                    "Device connection string or authentication configuration should be set",
```
arsing: `"One of provisioning.device_connection_string or provisioning.authentication must be set in the config.yaml."`

edgelet/edgelet-core/src/settings.rs
```diff
@@ -61,44 +112,113 @@ impl Manual {
...
+
+        let value: Inner = serde::Deserialize::deserialize(deserializer)?;
+
+        let authentication = match (value.device_connection_string, value.authentication) {
+            (Some(cs), Some(auth)) => {
+                if cs == DEFAULT_CONNECTION_STRING {
+                    auth
+                } else if let ManualAuthMethod::DeviceConnectionString(_) = auth {
+                    return Err(serde::de::Error::custom(
+                            "Multiple connection strings specified under provisioning.device_connection_string and provisioning.authentication.device_connection_string. Please specify only one connection string in the config.yaml",
```
arsing: It doesn't really matter whether the `ManualAuthMethod` is a `DeviceConnectionString` or an `X509`. It should conflict with `device_connection_string` in either case. So don't check for what kind of `ManualAuthMethod` it is, and just return:

`"Only one of provisioning.device_connection_string and provisioning.authentication must be set in the config.yaml."`

edgelet/iotedge/src/check/mod.rs
```diff
@@ -724,10 +726,16 @@ fn settings_connection_string(check: &mut Check) -> Result<CheckResult, failure:
...
-        )?;
+        let hub = match manual.authentication_method() {
+            ManualAuthMethod::DeviceConnectionString(cs) => {
+                let (_, _, hub) = cs.parse_device_connection_string().context(
+                                "Invalid connection string format detected.\n\
+                                Please check the value of the provisioning.device_connection_string parameter.",
+                )?;
+                hub
+            }
+            ManualAuthMethod::X509(x509) => x509.iothub_hostname().to_string(),
```
arsing: `.to_string()` -> `.to_owned()`

edgelet/iotedge/src/check/mod.rs
```diff
@@ -1284,12 +1292,25 @@ fn settings_identity_certificates_expiry(check: &mut Check) -> Result<CheckResul
...
+            if let ManualAuthMethod::X509(x509) = manual.authentication_method() {
+                let path = x509.identity_cert()?;
+                return CertificateValidity::parse(
+                    "Manual authentication identity certificate",
+                    &path,
+                )?
                 .to_check_result();
+            }
         }
+        _ => (),
```
arsing: Since the remaining cases are just `Provisioning::External`, put `Provisioning::External(_)` for the pattern instead of `_`.

edgelet/iotedged/src/lib.rs
```diff
@@ -249,10 +252,12 @@ where
             .context(ErrorKind::Initialize(InitializeErrorReason::Tokio))?;
 
         if let Provisioning::Manual(ref manual) = settings.provisioning() {
-            if manual.device_connection_string() == DEFAULT_CONNECTION_STRING {
-                return Err(Error::from(ErrorKind::Initialize(
-                    InitializeErrorReason::NotConfigured,
-                )));
+            if let ManualAuthMethod::DeviceConnectionString(cs) = manual.authentication_method() {
+                if cs.device_connection_string() == DEFAULT_CONNECTION_STRING {
```
arsing: This will also become unnecessary when you remove `DEFAULT_CONNECTION_STRING` as I said in my other comment.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -143,9 +155,19 @@ function Initialize-IoTEdge {
         [Switch] $External,
 
         # The device connection string.
-        [Parameter(Mandatory = $true, ParameterSetName = 'Manual')]
```
arsing: This is going to cause more problems like https://github.com/Azure/iotedge/issues/1486

I think at this stage it's worth making several more finer-grained parameter sets. `Manual_DCS`, `Manual_X509`, `DPS_TPM`, `DPS_SymmetricKey` and `DPS_X509`. Then set `Mandatory` for the sets that need them.

shantanu1singh:
>             InitializeErrorReason::DpsProvisioningClient,

These error reasons should be updated too now that it's not just Dps using the method

---
Refers to: edgelet/iotedged/src/lib.rs:1463 in fa3dad1. [](commit_id = fa3dad1c0cd579ec93f73c71c1f1d7564979ab84, deletion_comment = False)

shantanu1singh:
>             InitializeErrorReason::DpsProvisioningClient,

These error reasons should be updated too now that it's not just Dps using the method

---
Refers to: edgelet/iotedged/src/lib.rs:1469 in fa3dad1. [](commit_id = fa3dad1c0cd579ec93f73c71c1f1d7564979ab84, deletion_comment = False)

edgelet/edgelet-core/src/settings.rs
mrohera: Cleaned up, thanks for suggesting this.

edgelet/contrib/config/linux/config.yaml
mrohera: I think there are several places that need to change including MS docs., current customer deployments, samples, tutorials and other external webpages that talk about Edge. So this might be a bit much to change this as the favored configuration at this stage to warrant for what is essentially a dev setup.

edgelet/contrib/config/linux/config.yaml
arsing: There are still five of these, some of which are also missing the `.` in `X.509`

edgelet/edgelet-docker/test/linux/bad_settings.manual.cs.3.yaml
```diff
@@ -0,0 +1,38 @@
+
```
arsing: Remove this empty first line. Other files below also have it.

edgelet/iotedged/src/lib.rs
```diff
@@ -253,7 +253,7 @@ where
 
         if let Provisioning::Manual(ref manual) = settings.provisioning() {
             if let ManualAuthMethod::DeviceConnectionString(cs) = manual.authentication_method() {
-                if cs.device_connection_string() == DEFAULT_CONNECTION_STRING {
+                if cs.device_connection_string().is_empty() {
```
arsing: The call to `manual_provision` below will already fail with `ErrorKind::ConnectionStringEmpty` as part of calling `ManualDeviceConnectionString::parse_device_connection_string`, so this extra check isn't necessary.

edgelet/contrib/config/linux/config.yaml
mrohera: Done. Thanks for keeping me honest.

edgelet/edgelet-docker/test/linux/bad_settings.manual.cs.3.yaml
mrohera: All done.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
mrohera: Thanks for this feedback and your help. I have cleaned this up per your comments and ensured backward compat as well.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -88,105 +88,136 @@ PS> Initialize-IoTEdge -Manual -DeviceConnectionString $deviceConnectionString -
...
+
+        # Specifies that the daemon will be configured manually, using an identity certificate.
+        [Parameter(ParameterSetName = 'ManualX509')]
+        [Switch] $ManualX509,
+
+        # Specified the daemon will be configured using DPS. The choice of attestation depends on the other DPS input parameters.
+        [Parameter(ParameterSetName = 'DpsTpm')]
+        [Parameter(ParameterSetName = 'DpsSymmetricKey')]
+        [Parameter(ParameterSetName = 'DpsX509')]
         [Switch] $Dps,
```
arsing: Move this one down too like for `Install-`

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -88,105 +88,136 @@ PS> Initialize-IoTEdge -Manual -DeviceConnectionString $deviceConnectionString -
...
+        [Parameter(ParameterSetName = 'DpsSymmetricKey')]
+        [Switch] $DpsSymmetricKey,
+
+        # Specifies that the daemon will be configured using DPS X509 attestation.
+        [Parameter(ParameterSetName = 'DpsX509')]
+        [Switch] $DpsX509,
 
         # The device connection string.
-        [Parameter(ParameterSetName = 'Manual')]
+        [Parameter(Mandatory=$true, ParameterSetName = 'ManualConnectionString')]
```
arsing: Spaces around the `=` like the other parameters.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1584,82 +1674,37 @@ function Validate-GatewaySettings {
 }
 
 function Get-DpsProvisioningSettings {
-    $idCertFilesProvided = $false
-    $attestationMethod = 'tpm' # default
-    if ($SymmetricKey) {
+    $attestationMethod = ''
+    if ($DpsTpm) {
```
arsing: These won't be set if the user calls `Initialize-` directly and doesn't pass any switches, just the required parameters.

One way is you pass the `$PSCmdlet.ParameterSetName` value down from `Initialize-` to here (via `Set-ProvisioningMode`) as a parameter.

The easier way is to add a `switch ($PSCmdlet.ParameterSetName)` in `Initialize-` that manually sets the corresponding switch variable to `$true`. I prefer this one.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1584,82 +1674,37 @@ function Validate-GatewaySettings {
...
 }
 
 function Get-ManualAuthSettings {
-    $idCertFilesProvided = $false
-    $authenticationMethod = 'device_connection_string' # default
+    $authenticationMethod = ''
 
-    if ($X509IdentityCertificate -or $X509IdentityPrivateKey) {
-        $idCertFilesProvided = $true
+    if ($ManualConnectionString) {
```
arsing: Same problem here as with the switches in `Get-DpsProvisioningSettings`.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
mrohera: Thanks I didn't understand this when you described this earlier but I see your point. I have fixed this now.

PR #1585, opened 2019-08-16
fix Transparent Gateway test

philipktlin:
Update e2e script to set default value for root ca cert and key file path in order generate root ca cert for transparent gateway test.


mrohera:
What does this fix? Is this more for convenience and ease of use? Also what about a default pass phrase file which could be used if a pass phrase was not provided.

philipktlin:
> What does this fix? Is this more for convenience and ease of use? Also what about a default pass phrase file which could be used if a pass phrase was not provided.

No, test is failed after I removed CA cert installation from transparent gateway test.  because it depends on that 2 variables - $EdgeE2ERootCACertRSAFile and $EdgeE2ERootCAKeyRSAFile; which is set to default value in transparent gateway before calling ca cert installation previously.  Then I need to move default value assignment from transparent gateway test to outer scope.

I manually tested and it works correctly now.

PR #1592, opened 2019-08-19
Vipeller/secur

vipeller:
Changed ASN1 encoding for Basic Constrains/CA flag for x509 certificates. Some ASN1 decoders expect DER encoded TRUE value (=0xff), and refuse validating certificates with 0x01 - which is a correct BER encoded value, and most ASN1 decoders accept it.


vipeller:
created a cleaned up version

PR #1594, opened 2019-08-19
Remove SkipArchCheck

vtam-msft:
Bring back chandde's change for removing Arch Check


arsing:
@veyalla Just confirming we can merge this.

veyalla:
I don’t think Windows ARM support has reached public preview yet. I think this check should remain in the public installer until Windows ARM bits are publicly available.

arsing:
@vtam-msft Please test it once more to be sure since you have the Windows ARM32 environment. Then we can merge it.

PR #1595, opened 2019-08-20
Drop tls connection when tls error occurred instead of bubble up error

dmolokanov:
When an error occurred on TLS handshake HTTP server erroneously reported an error and exit stopped program execution. 
Now it ignores this type of errors and continues to execute the next available requests.


edgelet/edgelet-http/src/util/incoming.rs
```diff
@@ -70,45 +71,50 @@ impl Stream for Incoming {
...
-                    .iter_mut()
-                    .map(|(fut, _)| fut.poll())
-                    .enumerate()
-                    .find(|(_, result)| match result {
-                        Ok(v) => v.is_ready(),
-                        Err(_) => true,
-                    });
+                loop {
+                    // Look through the connections list for the first connection that is either ready to be
+                    // passed to the stream selector or has failed and needs the error bubbled up.
```
arsing: "or has failed and needs the error bubbled up" is outdated now.

edgelet/edgelet-http/src/util/incoming.rs
```diff
@@ -70,45 +71,50 @@ impl Stream for Incoming {
...
-                                #[cfg(windows)]
-                                {
-                                    Async::Ready(Some(
-                                        (Box::new(StreamSelector::Tls(tls_stream), addr)),
-                                    ))
+                                // The prior block included a filter that specifically asked for is_ready state,
+                                // so this line is unreachable.
+                                Ok(_) => unreachable!(),
+                                Err(err) => {
+                                    log_failure(Level::Warn, &err);
```
arsing: Maybe add a comment like `// Ignore TLS handshake errors.`

edgelet/edgelet-http/src/util/incoming.rs
```diff
@@ -70,45 +71,50 @@ impl Stream for Incoming {
...
                                 }
-                                #[cfg(windows)]
-                                {
-                                    Async::Ready(Some(
-                                        (Box::new(StreamSelector::Tls(tls_stream), addr)),
-                                    ))
+                                // The prior block included a filter that specifically asked for is_ready state,
+                                // so this line is unreachable.
+                                Ok(_) => unreachable!(),
+                                Err(err) => {
```
kkdawkins: Is there any error filtering that we want to do or just drop connection on any/all errors?

edgelet/edgelet-http/src/util/incoming.rs
```diff
@@ -70,45 +71,50 @@ impl Stream for Incoming {
...
-                // Return a tuple containing the index and state.
-                let val = connections
-                    .iter_mut()
-                    .map(|(fut, _)| fut.poll())
-                    .enumerate()
-                    .find(|(_, result)| match result {
-                        Ok(v) => v.is_ready(),
-                        Err(_) => true,
-                    });
+                loop {
```
kkdawkins: You made this a loop?

edgelet/edgelet-http/src/util/incoming.rs
dmolokanov: It rather complicated to implement (I don't want to say impossible), but
1. This error type doesn't support ErrorKind - difficult to filter out the error
2. This is an error that comes from native_tls, so I assume that it is the type we are interested in.

During a quick discussion with @arsing we decided that it is good enough to leave like this.

edgelet/edgelet-http/src/util/incoming.rs
dmolokanov: Yes. This loop covers the situation when we have several items in a `vec` that are ready to process. Imagine, if one of them has an error, and another doesn't have. We process the first one and return `Async::NotReady`. But we have another one that has no task to signal that future is ready again. 
Other words, once we started processing inside the `poll` method we should process all futures that are ready to poll.

edgelet/edgelet-http/src/util/incoming.rs
```diff
@@ -70,45 +71,50 @@ impl Stream for Incoming {
...
+                            match result {
+                                Ok(Async::Ready(tls_stream)) => {
+                                    #[cfg(not(windows))]
+                                    {
+                                        return Ok(Async::Ready(Some((
+                                            StreamSelector::Tls(tls_stream),
+                                            addr,
+                                        ))));
+                                    }
+                                    #[cfg(windows)]
```
dmolokanov: @kkdawkins Why do we need this block for windows? I thought whole TLS thing is for [unix only](https://github.com/Azure/iotedge/blob/a11c02fb733c576853b609c387a81faeab11ab0b/edgelet/edgelet-http/src/util/incoming.rs#L55)

edgelet/edgelet-http/src/util/incoming.rs
```diff
@@ -5,8 +5,7 @@ use std::io;
 use std::sync::Mutex;
 
 use futures::{Poll, Stream};
-#[cfg(unix)]
-use tokio::io::{Error as TokioIoError, ErrorKind as TokioIoErrorKind};
+use log::Level;
```
arsing: As the CI failure says, this import and the `log_failure` one below need to be `#[cfg(unix)]`

dmolokanov: Yeah, I noticed that. Thanks!

PR #1600, opened 2019-08-20
Changed ASN1 encoding of Basic Constrains/CA. 

vipeller:
Some ASN1 decoders expect DER encoded TRUE value (=0xff), and refuse validating certificates with 0x01

### DIFF
--- a/edgelet/edgelet-http/src/certificate_manager.rs
+++ b/edgelet/edgelet-http/src/certificate_manager.rs
@@ -73,7 +73,7 @@ impl<C: CreateCertificate + Clone> CertificateManager<C> {
             X509::stack_from_pem(cert).with_context(|_| ErrorKind::CertificateConversionError)?;
 
         let mut ca_certs = Stack::new().with_context(|_| ErrorKind::CertificateConversionError)?;
-        for cert in certs.split_off(1) {
+        for cert in certs.split_off(1).into_iter().rev() {
             ca_certs
                 .push(cert)
                 .with_context(|_| ErrorKind::CertificateConversionError)?;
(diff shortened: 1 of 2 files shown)


edgelet/edgelet-http/src/certificate_manager.rs
```diff
@@ -73,7 +73,7 @@ impl<C: CreateCertificate + Clone> CertificateManager<C> {
             X509::stack_from_pem(cert).with_context(|_| ErrorKind::CertificateConversionError)?;
 
         let mut ca_certs = Stack::new().with_context(|_| ErrorKind::CertificateConversionError)?;
-        for cert in certs.split_off(1) {
+        for cert in certs.split_off(1).into_iter().rev() {
```
arsing: This change isn't covered by the commit message. Why is it needed? (And say that in the commit message.)

arsing: Ah, the second commit.

arsing: Feels wrong to reverse it here, since it could potentially be a well-formed PEM provided by the user.

If it's just the quickstart generated one that's backwards, it seems we should fix the quickstart certs generator in libiothsm.

mrohera: Yes, why is this needed? Also please see this https://github.com/Azure/iotedge/blob/50ac495e73ff7b12b47e8227cc1468877aaf9e4e/edgelet/edgelet-http/src/lib.rs#L133. Does this need changing as well?

vipeller: ok, reverted that change. The problem I saw was that the certificate order was mixed up (first one was the certificate of the tls server, then the chain up to the root was reversed). That is not valid according to TLS 1.2 (valid in TLS 1.3, though).

PR #1601, opened 2019-08-20
Split up large hsm-sys lib.rs

darobs:
The original hsm-sys/src/lib.rs file was generated and cleaned up a bit, but we've been hand-editing it since.  It's not readable in its current form, so I split off the 3 major sections (tpm, x509, and crypto) into modules to improve readability and understanding.

There is no functionality change.


darobs:
> LGTM. Did you make any changes except of moving code to its modules?

I did change a few references from "\<struct\>_TAG" to just "\<struct\>", but they are aliases of the same structure.

Otherwise, it should only be moving code into modules.

arsing:
>I did change a few references from "<struct>_TAG" to just "<struct>", but they are aliases of the same structure.

Did you? I don't see any. `HSM_CERTIFICATE_TAG`, `SIZED_BUFFER_TAG` and `HSM_CERTIFICATE_PROPS_TAG` were not moved or renamed. (`HSM_CERTIFICATE_TAG` and `HSM_CERTIFICATE_PROPS_TAG` are also misnamed since the C types are called `CERT_DATA_INFO_TAG` and `HSM_CERT_PROPS_TAG`.) `HSM_CLIENT_{TPM,X509,CRYPTO}_INTERFACE_TAG` were moved and not renamed.

If you did intend to rename them, I support it.

darobs:
I see what you mean @arsing - there was no need to have a _TAG struct or enum and then alias it, so I removed the "_TAG" from all the names and removed the aliases.

PR #1604, opened 2019-08-21
Add new certificate script commands to facilitate DPS and manual X.509 flows

mrohera:
The changes introduce new commands that are very explicit about what type of certificate is being created. Either the gateway CA certificate or and identity certificate.


tools/CACertificates/ca-certs.ps1
```diff
@@ -750,12 +755,58 @@ function New-CACertsDevice([Parameter(Mandatory=$TRUE)][string]$deviceName, [str
         CA certificate is used.
 #>
 function New-CACertsEdgeDevice([Parameter(Mandatory=$TRUE)][string]$deviceName, [string]$issuerPrefix=$_intermediatePrefix)
+{
+    if ([string]::IsNullOrWhiteSpace($deviceName))
```
damonbarry: Because `$deviceName` is a mandatory parameter, it already can't be null or empty. So that only leaves whitespace, which could be handled with the attribute `[ValidatePattern("[^\s]+")]`.

tools/CACertificates/ca-certs.ps1
```diff
@@ -750,12 +755,58 @@ function New-CACertsDevice([Parameter(Mandatory=$TRUE)][string]$deviceName, [str
...
+    New-CACertsDevice $deviceName $issuerPrefix "iot-edge-device"
+}
+
+<#
+    .SYNOPSIS
+    Generate an IoT Edge device CA certificate, private key and full certificate chain
+    using an issuer's certificate and key
+    .DESCRIPTION
+    Generate a certificate using the supplied common name and have it signed by
+    an issuer certificate and key. Issuer is spcified by its prefix and its corresponding
```
damonbarry: typo: spcified

damonbarry: (Same for the next function description)

tools/CACertificates/certGen.sh
```diff
@@ -367,7 +367,7 @@ function generate_verification_certificate()
 ###############################################################################
 # Generates a certificate for a device, chained to the intermediate.
 ###############################################################################
-function generate_device_certificate()
+function generate_identity_device_certificate()
```
damonbarry: Should this function by `generate_device_identity_certificate`?

tools/CACertificates/certGen.sh
```diff
@@ -404,16 +423,30 @@ function generate_edge_server_certificate()
 }
 
 ###############################################################################
-# Generates a certificate for a Edge device, chained to the intermediate.
+# Generates a CA certificate for a Edge device, chained to the intermediate.
 ###############################################################################
-function generate_edge_device_certificate()
+function generate_edge_device_ca_certificate()
```
damonbarry: Why did you remove parameter checking on this function? I ask because the previous function (`generate_edge_device_identity_certificate`) has parameter checking...

tools/CACertificates/ca-certs.ps1
```diff
@@ -708,18 +708,24 @@ function New-CACertsVerificationCert([Parameter(Mandatory=$TRUE)][string]$reques
     .PARAMETER isEdgeDevice
```
damonbarry: The help text needs to be updated with the new `filePrefix` parameter that replaces `isEdgeDevice`.

tools/CACertificates/README.md
```diff
@@ -78,22 +78,30 @@ On Azure IoT Hub, navigate to the "Device Explorer".  Add a new device (e.g. `my
 Note that if you're using this certificate as a DPS registration ID, the ID **must be lower case** or the server will reject it.
 
 ### **PowerShell**
-#### IoT Leaf Device
-* Run `New-CACertsDevice mydevice` to create the new device certificate.
```
damonbarry: Both old and new end-to-end tests call `New-CACertsDevice` (and `./certGen.sh create_edge_device_certificate`). It looks like those will continue to work, but should the tests be updated to use the "recommended" approach?

tools/CACertificates/ca-certs.ps1
mrohera: I prefer a friendly error message even though the null check is redundant.

tools/CACertificates/certGen.sh
mrohera: Agreed and fixed.

tools/CACertificates/certGen.sh
mrohera: The validations were not put in here and other the function. I have gone ahead and added these. Thanks.

tools/CACertificates/README.md
mrohera: Correct. Longer term the E2E tests should be migrated over to the newer commands and the older one can simply be removed.

damonbarry: Agree it should be done separately.

PR #1606, opened 2019-08-21
Add a task to build iotedged with feature=runtime-kubernetes during checkin.

darobs:
In order to make sure changes in one module runtime (Docker) don't break the other module runtime (Kubernetes), I added a step in the linux amd64 Checkin build to build iotedged with the runtime-kubernetes feature turned on.

### DIFF
--- a/edgelet/build/linux/build-k8s.sh
+++ b/edgelet/build/linux/build-k8s.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+###############################################################################
+# This script builds the project
+###############################################################################
+
+set -e
+
+###############################################################################
+# Define Environment Variables
+###############################################################################
+# Get directory of running script
+DIR=$(cd "$(dirname "$0")" && pwd)
+
+BUILD_REPOSITORY_LOCALPATH=${BUILD_REPOSITORY_LOCALPATH:-$DIR/../../..}
+PROJECT_ROOT=${BUILD_REPOSITORY_LOCALPATH}/edgelet
+IOTEDGED_MANIFEST=${PROJECT_ROOT}/iotedged/Cargo.toml
+SCRIPT_NAME=$(basename "$0")
+CARGO="${CARGO_HOME:-"$HOME/.cargo"}/bin/cargo"
+TOOLCHAIN="stable-x86_64-unknown-linux-gnu"
+RELEASE=
+
+###############################################################################
+# Print usage information pertaining to this script and exit
+###############################################################################
+usage()
+{
+    echo "$SCRIPT_NAME [options]"
+    echo ""
+    echo "options"
+    echo " -h, --help          Print this help and exit."
+    echo " -t, --toolchain     Toolchain (default: stable-x86_64-unknown-linux-gnu)"
+    echo " -r, --release       Release build? (flag, default: false)"
+    exit 1;
+}
+
+###############################################################################
+# Obtain and validate the options supported by this script
+###############################################################################
+process_args()
+{
+    save_next_arg=0
+    for arg in "$@"
+    do
+        if [ $save_next_arg -eq 1 ]; then
+            TOOLCHAIN="$arg"
+            save_next_arg=0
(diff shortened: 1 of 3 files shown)


arsing:
In that case, please change the regular `build.sh` to build with `-p ...` instead of `--all`, so that Kubernetes-specific crates are only built in the second job.

edgelet/build/linux/build-k8s.sh
```diff
@@ -0,0 +1,68 @@
...
+                * ) usage;;
+            esac
+        fi
+    done
+}
+
+process_args "$@"
+
+if [[ -z ${RELEASE} ]]; then
+    cd "$PROJECT_ROOT" && $CARGO "+$TOOLCHAIN" build --manifest-path=${IOTEDGED_MANIFEST} --no-default-features --features runtime-kubernetes
```
arsing: Just do `-p iotedged` instead of `--manifest-path`

edgelet/build/linux/build.sh
```diff
@@ -6,6 +6,11 @@
 
 set -e
 
+###############################################################################
+# These are the packages this script will build.
+###############################################################################
+packages=(iotedge iotedged iotedge-diagnostics iotedge-proxy)
```
arsing: We need `iotedge-proxy` for non-Kubernetes?

darobs: iotedge-proxy is not build with the feature "runtime-kubernetes" - it's built with the default features.

I've gone back and forth on this.  It was easier to add to the list of packages, and where it is placed does not impact build time, but it not required for non-k8s work.  If other people have strong opinions on placement, I'll gladly move it.

edgelet/build/linux/build-k8s.sh
darobs: I would but it doesn't work:
```sh
~/iotedge/edgelet$ build/linux/build-k8s.sh
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
~/iotedge/edgelet$ target/debug/iotedged -c ~/config.yaml 
<6>2019-08-23T15:18:31Z [INFO] - Starting Azure IoT Edge Security Daemon - Kubernetes mode
<6>2019-08-23T15:18:31Z [INFO] - Version - 1.0.10~dev
<6>2019-08-23T15:18:31Z [INFO] - Using config file: /home/darobs/config.yaml
<3>2019-08-23T15:18:31Z [ERR!] - The daemon could not start up successfully: Could not load settings
<3>2019-08-23T15:18:31Z [ERR!] - 	caused by: Config parsing error
<3>2019-08-23T15:18:31Z [ERR!] - 	caused by: missing field `proxy_trust_bundle_path`
~/iotedge/edgelet$ cargo build -p iotedged --no-default-features --features runtime-kubernetes 
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
~/iotedge/edgelet$ target/debug/iotedged -c ~/config.yaml 
<6>2019-08-23T15:18:47Z [INFO] - Starting Azure IoT Edge Security Daemon
<6>2019-08-23T15:18:47Z [INFO] - Version - 1.0.10~dev
```

See: https://github.com/rust-lang/cargo/issues/5015

edgelet/build/linux/build.sh
arsing: If it was meant for non-Kubernetes I would've liked to be on the code review for when it was merged. Just from a quick glance I see [it's not doing error-handling correctly.](https://github.com/Azure/iotedge/blob/master/edgelet/iotedge-proxy/src/error.rs#L127-L165)

Anyway, what features it's built with isn't the way we should differentiate. Was it planned to ship in `iotedge.deb` ? Then it's part of the regular build. If it wasn't, then it's not.

I'm fine with leaving it where it is, but in the future if you're planning on doing non-Kubernetes work, please involve the non-Kubernetes team in the code reviews.

darobs: Added a bug to our backlog to get our error handling in sync with the rest of the project.

PR #1607, opened 2019-08-21
[k8s] Add trust bundle for each module

dmolokanov:
In order to establish secure communication between module and iotedged each pod should contain volume with iotedged trust bundle attached.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
@@ -3,7 +3,6 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
 {
     using System;
     using System.Collections.Generic;
-    using System.ComponentModel;
     using System.Linq;
     using System.Threading.Tasks;
     using k8s;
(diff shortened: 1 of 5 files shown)


kkdawkins:
It looks fine, but since we are already behind on test debt, are there any UTs that you can write to cover this?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
```diff
@@ -459,130 +461,126 @@ async Task PruneServiceAccounts(List<string> accountNamesToPrune)
...
+                });
 
-            return new V1PodTemplateSpec();
+            var modulePodSpec = new V1PodSpec(containerList, volumes: volumeList, imagePullSecrets: imageSecret.OrDefault());
+            var objectMeta = new V1ObjectMeta(labels: podLabels, annotations: podAnnotations);
+            return new V1PodTemplateSpec(objectMeta, modulePodSpec);
         }
 
-        private VolumeOptions GetVolumesFromModule(IModule<AgentDocker.CombinedDockerConfig> moduleWithDockerConfig)
+        (List<V1Volume>, List<V1VolumeMount>, List<V1VolumeMount>) GetVolumesFromModule(IModule<AgentDocker.CombinedDockerConfig> moduleWithDockerConfig)
```
kkdawkins: Rather than put it back as a three tuple, why not create a VolumeOptions type?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/KubernetesModule.cs
```diff
@@ -91,32 +96,33 @@ public class KubernetesModule : Module
...
-                if (this.enableServiceCallTracing)
-                {
-                    // enable tracing of k8s requests made by the client
-                    var loggerFactory = c.Resolve<ILoggerFactory>();
-                    ILogger logger = loggerFactory.CreateLogger(typeof(Kubernetes));
-                    ServiceClientTracing.IsEnabled = true;
-                    ServiceClientTracing.AddTracingInterceptor(new DebugTracer(logger));
-                }
+            builder.Register(
+                    c =>
```
kkdawkins: Why did you move these down to their own line? It increases the amount of nesting. Can you put these back?

kkdawkins: Same comment for the instances in this file.

dmolokanov: Probably its IDE formatting that I need to configure

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
dmolokanov: We don't use option here at all. Private type for private method seems like overkill or not?

PR #1608, opened 2019-08-21
remove debug log to print connection string

philipktlin:
We should not print iot hub connection string in Quickstart for E2E testing.


mrohera:
Also can you please roll the keys as well.

PR #1609, opened 2019-08-21
Schedule task Environment Cleanup/Install to run locally

SlavyMihov:
This is to avoid remote PS session interruption when runner device updates iotedge service and reboots.


builds/e2e/templates/deploy-iotuap.yaml
```diff
@@ -42,14 +42,19 @@ steps:
       InlineScript: |
         #Requires -RunAsAdministrator
         $ArtifactInfo=convertfrom-stringdata (get-content C:\Data\e2etests\artifacts\${{ parameters['images.artifact.name'] }}\artifactInfo.txt -raw)
-        C:\Data\e2etests\artifacts\${{ parameters['images.artifact.name'] }}\scripts\windows\test\Setup-Env.ps1 -E2ETestFolder "C:\Data\e2etests" -ArtifactImageBuildNumber "$($ArtifactInfo.BuildNumber)" -AttemptUpdate
+        # schedule Cleanup/Install task - reboots the device
+        $ScheduleDatetime = [DateTime]::Now.AddMinutes(2)
+        $ScheduleDate=$ScheduleDatetime.ToString("MM/dd/yyyy")
+        $ScheduleTime=$ScheduleDatetime.ToString("HH:mm")
+        Write-Host "Schedule to stop iotedge service at $ScheduleDate $ScheduleTime"
+        SchTasks.exe /Create /RU "SYSTEM" /SC ONCE /TN “StressTestCleanup” /TR "powershell.exe -File C:\Data\e2etests\artifacts\${{ parameters['images.artifact.name'] }}\scripts\windows\test\Setup-Env.ps1 -E2ETestFolder C:\Data\e2etests -ArtifactImageBuildNumber $($ArtifactInfo.BuildNumber) -AttemptUpdate" /SD $ScheduleDate /ST $ScheduleTime /F
```
philipktlin: is it safer to use $env:ProgramData although this file is used only for iot core right now?  btw, why do you need to schedule to run setup_Env.ps1?  it is because if we run right away, it restart the machine immediately and this deployment step will be failed.  do i understand correctly?

SlavyMihov: I agree, it would be safer to use$env:ProgramData, however this is not needed as this file is used only for Windows IoT Core. Let me know if you have a strong preference here to change that.

Your understanding is correct. This task is scheduled to run on the target device and not as part of the remote PS session. This is a workaround for the reported failure when the remote PS session is terminated (by the remote device rebooting).

PR #1610, opened 2019-08-22
Refactor Kube Watcher

kkdawkins:


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesServiceBuilder.cs
```diff
@@ -0,0 +1,163 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+
+            if (portList.Count > 0)
+            {
+                // Selector: by module name and device name, also how we will label this puppy.
+                var objectMeta = new V1ObjectMeta(annotations: serviceAnnotations.GetOrElse(() => null), labels: labels, name: KubeUtils.SanitizeDNSValue(moduleIdentity.ModuleId));
```
dmolokanov: You could you `Option<T>.OrDefault()` method instead of `GetOrElse(() => default(T))`

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesServiceBuilder.cs
```diff
@@ -0,0 +1,163 @@
...
+                            exposedList =>
+                                exposedList.ForEach((item) => portList.Add(new V1ServicePort(item.Port, name: $"ExposedPort-{item.Port}-{item.Protocol.ToLower()}", protocol: item.Protocol))));
+                }
+
+                // Handle HostConfig PortBindings entries
+                if (moduleWithDockerConfig.Config?.CreateOptions?.HostConfig?.PortBindings != null)
+                {
+                    foreach (KeyValuePair<string, IList<DockerModels.PortBinding>> portBinding in moduleWithDockerConfig.Config?.CreateOptions?.HostConfig?.PortBindings)
+                    {
+                        string[] portProtocol = portBinding.Key.Split('/');
```
dmolokanov: We are using parse port and protocol in several places. Should we create a new class that represents this pair and method `Parse` to encapsulate parsing logic inside it? It should also help with testing

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesServiceBuilder.cs
```diff
@@ -0,0 +1,163 @@
...
+                            if (int.TryParse(portProtocol[0], out int port) && ProtocolExtensions.TryValidateProtocol(portProtocol[1], out string protocol))
+                            {
+                                // Entries in Docker portMap wants to expose a port on the host (hostPort) and map it to the container's port (port)
+                                // We interpret that as the pod wants the cluster to expose a port on a public IP (hostPort), and target it to the container's port (port)
+                                foreach (DockerModels.PortBinding hostBinding in portBinding.Value)
+                                {
+                                    if (int.TryParse(hostBinding.HostPort, out int hostPort))
+                                    {
+                                        // If a port entry contains the same "port", then remove it and replace with a new ServicePort that contains a target.
+                                        var duplicate = portList.SingleOrDefault(a => a.Port == hostPort);
```
dmolokanov: Probably we can use `HashSet` with custom equality comparer for storing ports. In this case, we don't need to manually iterate across list and ensure unique elements

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesServiceBuilder.cs
```diff
@@ -0,0 +1,163 @@
...
+
+                return Option.Some(new V1Service(metadata: objectMeta, spec: new V1ServiceSpec(type: serviceType, ports: portList, selector: labels)));
+            }
+            else
+            {
+                return Option.None<V1Service>();
+            }
+        }
+
+        private Option<List<(int Port, string Protocol)>> GetExposedPorts(IDictionary<string, DockerModels.EmptyStruct> exposedPorts)
```
dmolokanov: `GetExposedPorts` is duplicated in several classes

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
```diff
@@ -20,10 +20,6 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
 
     public class CrdWatcher<TConfig>
```
dmolokanov: What do you think about completely separate watcher and logic to generate k8s primitives? So all code dependable on Kubernetes client will stay in watcher but logic to determine what should be deleted, updated or created will live in a separate class. This way each class will have only one responsibility

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/watchers/CrdWatcherTests.cs
```diff
@@ -0,0 +1,121 @@
+// Copyright (c) Microsoft. All rights reserved.
```
dmolokanov: I assume you haven't started unit tests yet, right?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
kkdawkins: I agree but I want this PR to be strictly about moving the pod/service creation stuff out of the watch.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesServiceBuilder.cs
kkdawkins: I thought that I did?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesServiceBuilder.cs
kkdawkins: We could! I want this PR to simply be a refactor, then I will add back my UTs, and then make improvements a-la TDD style :)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesModelBuilder.cs
```diff
@@ -0,0 +1,97 @@
...
+                        serviceList.Add((port, protocol));
+                    }
+                    else
+                    {
+                        Events.ExposedPortValue(exposedPort.Key);
+                    }
+                }
+            }
+
+            return (serviceList.Count > 0) ? Option.Some(serviceList) : Option.None<List<(int, string)>>();
```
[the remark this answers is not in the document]
dmolokanov: Varun agreed that we should use Option for collections only when we want to separate 3 situations: no collection exist, empty collection, a collection with some items in it. So we can just return empty collection here.

PR #1278, opened 2019-05-30
Don't restart iotedged when config.yaml isn't ready

damonbarry:
On Linux (systemd) when iotedge is installed, the daemon is configured to always restart on failure. Its config.yaml file needs to be configured by the user after installation, so as soon as it starts the first time it will go into a fail/restart loop until it exceeds the start rate limit and enters a cooldown period (a failed state where manual intervention is required). This change maps a specific exit code for this condition (specifically, failure when the connection string in config.yaml hasn't been changed from its boilerplate value), and instructs systemd to prevent restart for this exit code.

This change will help with automated tests, which sometimes trip the rate limiting threshold and aren't equipped to do manual intervention.

PR #1615, opened 2019-08-22
Implement quickstart certs end-to-end test

damonbarry:
Implements `QuickstartCerts`, which is basically the same as `TransparentGateway(SasOutOfScope, Amqp)` except that it uses the default (non-production, aka "quickstart") certificates generated by iotedged instead of custom certificates.

Note I also did some refactoring of the fixtures so that the names make more sense, given that `QuickstartCerts` and `TransparentGateway` tests can't be in the same test fixture (because they have different setup requirements).


damonbarry:
Because this build always uses the latest _released_ iotedged package and agent/hub/tempSensor images, and because iotedged must have the fix from PR #1278 in order for the end-to-end tests to work reliably, we need to wait to make this change until 1.0.9 is released.

PR #1619, opened 2019-08-23
Add scheduled trigger to service deployment build

damonbarry:
The service deployment build is currently set up to run nightly through a scheduled trigger in the Azure Pipelines UI. This change adds the same schedule (nightly, 3AM UTC) to our YAML definition. Then I can remove the trigger from the UI.


kkdawkins:
:shipit:

PR #1623, opened 2019-08-23
[k8s] Add missed env variables for EA

dmolokanov:


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
```diff
@@ -248,14 +247,6 @@ private async Task UpsertDeployments(V1ServiceList currentServices, V1Deployment
                     // Get the converted pod
                     V1PodTemplateSpec v1PodSpec = kubernetesModelBuilder.GetPod();
 
-                    // if this is the edge agent's deployment then it needs to run under a specific service account
```
kkdawkins: Don't we need this if we are launching EA?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
```diff
@@ -219,6 +216,8 @@ private async Task UpsertDeployments(V1ServiceList currentServices, V1Deployment
             foreach (KubernetesModule<TConfig> module in customObject.Spec)
             {
                 var moduleId = moduleIdentities[module.Name];
+
+                string deploymentName = this.DeploymentName(moduleIdentities[module.Name].ModuleId);
```
kkdawkins: Where is deploymentName being used?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesPodBuilder.cs
```diff
@@ -68,10 +68,12 @@ public V1PodTemplateSpec GetPodFromModule(Dictionary<string, string> labels, IMo
             // Image
             string moduleImage = module.Config.Image;
 
+            var name = KubeUtils.SanitizeDNSValue(moduleIdentity.ModuleId);
```
kkdawkins: Why is this necessary? KubeUtils is a simple extension method - no need for it's own var.

dmolokanov: why sanitize data multiple time instead of one time and reuse multiple times?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
dmolokanov: nope. iotedged provide the value for EA

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
dmolokanov: yeah, should be used in labels

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesPodBuilder.cs
kkdawkins: I see. You added "name" to V1ObjectMeta. At leat name it better...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
kkdawkins: Wouldn't that have been through SA name? I am confused why this was needed before and no longer needed - can you elaborate on what changed?

dmolokanov: It was needed before because only EA had a service account with a special name. Now every module will have its own service account that equals sanitized module name. So no need to pass env variable for it anymore

kkdawkins: That makes sense!

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/CrdWatcher.cs
```diff
@@ -223,7 +223,7 @@ private async Task UpsertDeployments(V1ServiceList currentServices, V1Deployment
                     // Default labels
                     var labels = new Dictionary<string, string>
                     {
-                        [Constants.K8sEdgeModuleLabel] = KubeUtils.SanitizeLabelValue(moduleId.ModuleId),
+                        [Constants.K8sEdgeModuleLabel] = deploymentName,
```
darobs: What's the difference between module id and deployment name?

dmolokanov: moduleId is module identity. 
deployment name is module name sanitized to k8s name rules.

PR #1624, opened 2019-08-23
EdgeHub: Fix delay in module 2 module messages

varunpuranik:
EdgeHub uses prefetching logic to improve message processing efficiency. However, the in the prefetch logic, if messages were coming in too slow, then it would cause a delay where the pump waited for more messages to come in. 
This fix changes that so that every time sending messages is completed, we again check the messages store to see if there are any new messages to be processed.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
@@ -182,6 +182,7 @@ namespace Microsoft.Azure.Devices.Routing.Core.Endpoints
             readonly TimeSpan timeout;
             readonly Task populateTask;
             List<IMessage> messagesList;
+            Option<Task<IList<IMessage>>> getMessagesTask = Option.None<Task<IList<IMessage>>>();
 
             public StoreMessagesProvider(IMessageIterator iterator, TimeSpan timeout, int batchSize)
             {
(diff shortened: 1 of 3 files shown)


edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -175,65 +175,54 @@ void Dispose(bool disposing)
         // A pump is started as soon as the object is created, and it keeps the messages list populated.
         internal class StoreMessagesProvider
         {
+            static readonly IList<IMessage> EmptyList = new List<IMessage>();
```
myagley: We may want to make this an ImmutableList so that is always stays empty?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -175,65 +175,54 @@ void Dispose(bool disposing)
...
             readonly IMessageIterator iterator;
             readonly int batchSize;
             readonly AsyncLock messagesLock = new AsyncLock();
-            readonly AsyncManualResetEvent messagesResetEvent = new AsyncManualResetEvent(true);
-            readonly TimeSpan timeout;
-            readonly Task populateTask;
-            List<IMessage> messagesList;
+            Option<Task<IList<IMessage>>> getMessagesTask;
```
myagley: I don't see when this would ever be None?

varunpuranik: Good catch, initially I was prefetching the messages only when there were more to be fetched, but later changed the logic. Can remove the Option and the empty list...

PR #1648, opened 2019-08-28
Update thumbprint tests

philipktlin:
1. Update test script and leaf device to run primary and secondary cert thumbprint test in separate processes.  Tested manually.  Need to update linux tests in e2e release pipeline to run primary thumbprint test after committed.

2. Fix error code handling in RunAllTests; exit code other than zero should treat as error.


smoke/LeafDevice/Program.cs
```diff
@@ -83,6 +83,9 @@ class Program
         [Option("-ctsk|--x509-secondary-key-path", Description = "Path to a X.509 leaf certificate key file in PEM format. This is needed for thumbprint auth and used as the secondary certificate's key.")]
         public string X509SecondaryKeyPath { get; } = string.Empty;
 
+        [Option("--use-secondary-credential", Description = "Set to true if secondary certificate should be used for thumbprint test, otherwise uses primary certificate (default).")]
```
mrohera: Consider: "Set value to true if secondary credential (either certificate or SharedAccessKey) should be used for authentication, otherwise the primary credential is used by default. Note: currently this is applicable for certificates tests only."

smoke/LeafDevice/Program.cs
```diff
@@ -83,6 +83,12 @@ class Program
         [Option("-ctsk|--x509-secondary-key-path", Description = "Path to a X.509 leaf certificate key file in PEM format. This is needed for thumbprint auth and used as the secondary certificate's key.")]
         public string X509SecondaryKeyPath { get; } = string.Empty;
 
+        [Option(
+            "--use-secondary-credential",
+            Description = "Set value to true if secondary credential (either certificate or SharedAccessKey) should be used for authentication, " +
```
mrohera: Nit: Set value to true if _the_ secondary credential

PR #1667, opened 2019-09-03
Sample - X509 Authentication Amqp failure fix

and-rewsmith:
This PR will fix the X509 authentication via Amqp that currently fails in the sample. @philipktlin previously discovered a bug where Amqp fails when both of the following conditions are true:

1. There is no revocation policy in place
2. The RevocationMode.Nocheck flag is not explicitly set to false.

I have applied [his workaround](https://github.com/Azure/iotedge/blob/76bc85dc7804712b432b430ddfbc14d7d06b4ad3/smoke/LeafDevice/details/CustomCertificateValidator.cs) in the sample.

It is worth noting that I also changed the behavior of the environment variable specifying protocol. Before, if the protocol was not recognized, Mqtt was chosen by default. Now, if the protocol is not recognized then there will be an ArgumentException.


samples/dotnet/EdgeX509AuthDownstreamDevice/CustomCertificateValidator.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                Console.WriteLine("The certificate chain was not signed by the trusted root certificate.");
+                return false;
+            }
+
+            Console.WriteLine("CustomCertificateValidator.ValidateCertificate is passed.");
+            return true;
+        }
+    }
+}
```
mrohera: Super Nit: No newline as seen from the diff.

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -152,27 +153,47 @@ internal static (X509Certificate2, IEnumerable<X509Certificate2>) ParseCertifica
...
+            else
+            {
+                throw new ArgumentException("Invalid protocol");
             }
 
-            return result;
+            if (string.Compare("Amqp", protocol, StringComparison.OrdinalIgnoreCase) == 0)
+            {
+                transportSettings[0] = new AmqpTransportSettings(transportType);
+                AmqpTransportSettings amqpTransportSettings = (AmqpTransportSettings)transportSettings[0];
```
mrohera: Why not use the certificate callback this for all protocols. This makes things very consistent.

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -299,6 +322,10 @@ static void InstallCACert()
             }
         }
 
+        static X509Certificate2 getTrustedCACertFromFile(string trustedCACertPath) {
```
mrohera: What made you want to move this out? Also if you want to keep this, I would suggest renaming the case of the method to be consistent with others in the file.

and-rewsmith: I discussed this decision with @ancaantochi. We decided that it would be awkward for InstallCACert to return null in the case where there are no certificates to install. Do you think I should have InstallACert() return an X509?

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -218,17 +239,19 @@ static void Main()
                 }
             }
 
-            TransportType transport = GetTransport(ClientTransportType);
-
             InstallCACert();
 
+            X509Certificate2 trustedCACert = getTrustedCACertFromFile(TrustedCACertPath);
```
and-rewsmith: This does not need to be at this high of scope level since TrustedCACertPath is global. I will move it into the implementation of GetTransport().

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
and-rewsmith: More context:

The trusted certificate needs to be visible at a higher scope so that it can be passed into the certificate validation callback. I thought about changing the return type of InstallCACert, but decided against it for the above reason. Instead, GetTrustedCACertFromFile exists and is called in two places. One in the implementation, and one right before it is needed by the callback (this diff has it in the wrong place currently).

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -218,17 +239,19 @@ static void Main()
...
+
             Console.WriteLine("Creating device client using identity certificate...\n");
 
             var (cert, certChain) = GetClientCertificateAndChainFromFile(DeviceIdentityCertPath, DeviceIdentityPrivateKeyPath);
             InstallChainCertificates(certChain);
-            var auth = new DeviceAuthenticationWithX509Certificate(DownstreamDeviceId, cert);
 
-            DeviceClient deviceClient = DeviceClient.Create(IothubHostname, IotEdgeGatewayHostname, auth, transport);
+            ITransportSettings[] transportSettings = GetTransport(ClientTransportType, trustedCACert);
```
and-rewsmith: We will no longer need to pass in trustedCACert

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
and-rewsmith: Also, if we change the return type of InstallCACert(), then we will have to pass the root certificate back into Main() scope and then into GetTransport(). I feel it is better to not do this.

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
and-rewsmith: My logic here was to only touch the control flow where the problem occurred to reduce potential unforeseen side effects.  You make a great point about consistency though. I will change.

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
mrohera: I suggest moving the logic to obtain the trusted CA certificate(s) in GetTrustedCACertFromFile and move the error checking etc. in there. Follow that up with refactoring InstallCACert by passing the trusted cert(s) as an arg. Hopefully things should read lot better and be cleaner.

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -152,27 +154,48 @@ internal static (X509Certificate2, IEnumerable<X509Certificate2>) ParseCertifica
...
 
-        static TransportType GetTransport(string protocol)
+        static ITransportSettings[] GetTransport(string protocol)
         {
-            TransportType result = TransportType.Mqtt_Tcp_Only;
+            TransportType transportType = TransportType.Mqtt_Tcp_Only;
+            ITransportSettings[] transportSettings = new ITransportSettings[1];
 
-            if (!string.IsNullOrWhiteSpace(protocol))
+            if (string.Compare("Mqtt", protocol, StringComparison.OrdinalIgnoreCase) == 0)
```
philipktlin: nit: string.Equals

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -152,27 +154,48 @@ internal static (X509Certificate2, IEnumerable<X509Certificate2>) ParseCertifica
...
+                transportType = TransportType.Amqp_WebSocket_Only;
+            }
+            else
+            {
+                throw new ArgumentException("Invalid protocol");
+            }
+
+            X509Certificate2 trustedCACert = GetTrustedCACertFromFile(TrustedCACertPath);
+            RemoteCertificateValidationCallback certificateValidationCallback = (sender, certificate, chain, sslPolicyErrors) => CustomCertificateValidator.ValidateCertificate(trustedCACert, (X509Certificate2) certificate, chain, sslPolicyErrors);
+            if (string.Compare("Amqp", protocol, StringComparison.OrdinalIgnoreCase) == 0 || string.Compare("AmqpWs", protocol, StringComparison.OrdinalIgnoreCase) == 0)
```
philipktlin: nit: use transporType variable

philipktlin: refer to https://github.com/Azure/iotedge/blob/584041aba7e20f7b21ffdf7e706f9d71da6006dc/smoke/LeafDevice/details/CustomCertificateValidator.cs#L41

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -218,17 +241,17 @@ static void Main()
...
 
             Console.WriteLine("Creating device client using identity certificate...\n");
 
             var (cert, certChain) = GetClientCertificateAndChainFromFile(DeviceIdentityCertPath, DeviceIdentityPrivateKeyPath);
             InstallChainCertificates(certChain);
-            var auth = new DeviceAuthenticationWithX509Certificate(DownstreamDeviceId, cert);
 
-            DeviceClient deviceClient = DeviceClient.Create(IothubHostname, IotEdgeGatewayHostname, auth, transport);
+            ITransportSettings[] transportSettings = GetTransport(ClientTransportType);
+
```
philipktlin: nit: remove blank line.

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -218,17 +241,17 @@ static void Main()
...
+            var auth = new DeviceAuthenticationWithX509Certificate(DownstreamDeviceId, cert);
```
philipktlin: nit: use explicit type instead of var; as you can't tell what type is from right side.

and-rewsmith:
I'm still working on running C# style checker. Friction with Windows VM.

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
and-rewsmith: This is a good idea. Also, I overlooked the fact that this should work when providing no certificate. The callback needs a X509Certificate object in order to work. Currently brainstorming a way to make this object when not directly specified in env vars.

and-rewsmith: I haven't found an easy way to do this. If the certificate validation callback is necessary for Amqp, and this requires a trusted certificate explicitly defined in order to work, this implies that  IOTEDGE_TRUSTED_CA_CERTIFICATE_PEM_PATH should be a required field in the launchSettings env vars. However, this seems to break the concept that one shouldn't need to know what is in one's trusted certificate store.

 If there was a way to programatically get a trusted certificate path from a hostname then I think this is the way to go.

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
and-rewsmith: My understanding was that var was ok since auth is of type DeviceAuthenticationWithX509Certificate. Am I misunderstanding something?

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
and-rewsmith: I spoke with @mrohera and we decided it would be best to make the trusted certificate path required

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -273,30 +295,23 @@ static void InstallChainCertificates(IEnumerable<X509Certificate2> certificateCh
         /// </summary>
         static void InstallCACert()
```
mrohera: Can I suggest passing X509Certificate2 trustedCACert as an arg?

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
```diff
@@ -271,32 +299,26 @@ static void InstallChainCertificates(IEnumerable<X509Certificate2> certificateCh
...
-                    Console.WriteLine("Attempting to install CA certificate: {0}", certPath);
-                    X509Store store = new X509Store(StoreName.Root, StoreLocation.CurrentUser);
-                    store.Open(OpenFlags.ReadWrite);
-                    store.Add(new X509Certificate2(System.Security.Cryptography.X509Certificates.X509Certificate.CreateFromCertFile(certPath)));
-                    Console.WriteLine("Successfully added certificate: {0}", certPath);
-                    store.Close();
-                }
-            }
-            else
+            Console.WriteLine("User configured CA certificate path: {0}", TrustedCACertPath);
```
mrohera: Shouldn't this be trustedCACert?

mrohera: Here and in the function

and-rewsmith: We add trustedCACert to the store @ line 308. trustedCACert is the certificate object, not the string representation.

mrohera: Yep. Okay makes sense.

samples/dotnet/EdgeX509AuthDownstreamDevice/Program.cs
mrohera: Yeah I am not seeing an issue here.

PR #1668, opened 2019-09-04
Honor configured stop timeout of modules (Fixes #1618)

### LINKED ISSUE #1618: StopTimeout parameter is discarded
# Expected Beahviour
Setting a StopTimeout parameter as Container Create Option for a custom module the runtime would wait this amount of seconds before killing the module when stopping it.   

## Current Behavior
When stopping a module the module is killed after 10 seconds - the default value, no matter of the provided options.

## Steps to Reproduce

1. In your edge device in the portal click on **Set Modules**
2. Click **Configure** for the selected slowly stopping custom module
3. Add **"StopTimeout": 180** in the Container Create Options 
4. Submit the module changes.
5. Wait for the module to be fully updated.
6. Click on **Set Modules**
7. Click on **Configure** for the same module
8. On desired state drop down select **stopped**
9. Wait 10 seconds and verify the module is stopped with error 137

### Output of `iotedge check`
 sudo iotedge check
error: Found argument 'check' which wasn't expected, or isn't valid in this context

USAGE:
    iotedge [OPTIONS] <SUBCOMMAND>

For more information try --help

### Device (Host) Operating System
Raspbian Stretch

### Architecture
arm32

### Container Operating System
Linux containers

#### iotedged
iotedge 1.0.6.1 (3fa6cbef8b7fc3c55a49a622735eb1021b8a5963)

#### Edge Agent
1.0

#### Edge Hub
1.0

#### Docker

 Version:           3.0.2
 API version:       1.40
 Go version:        go1.11.1
 Git commit:        e042b58f
 Built:             Tue Dec  4 17:16:17 2018
 OS/Arch:           linux/arm
 Experimental:      false

## Additional Information
A custom module runs a shell script with the exec form entrypoint. This script runs a Java application using the exec command. The Java application stops for about a minute to be able to cleanup all its resources, but it gets killed before it is able to finalize this operation.

Looking at the **StopCommand** in here https://github.com/Azure/iotedge/blob/master/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/commands/StopCommand.cs it seems it initializes a 10 seconds timeout and uses this all the time, no matter of any other options. 

Not sure about the usage of the edgelets, but it seems the edglet stop command sends an empty POST to the docker stop url and I presume it should take the **StopTimeout** container parameter in account.

shantanu1singh:
Honor configured stop timeout of modules.

### DIFF
--- a/edgelet/docker-rs/src/apis/container_api.rs
+++ b/edgelet/docker-rs/src/apis/container_api.rs
@@ -142,7 +142,7 @@ pub trait ContainerApi: Send + Sync {
     fn container_restart(
         &self,
         id: &str,
-        t: Option<i32>,
+        t: Option<&i32>,
     ) -> Box<dyn Future<Item = (), Error = Error<serde_json::Value>> + Send>;
     fn container_start(
         &self,
@@ -157,7 +157,7 @@ pub trait ContainerApi: Send + Sync {
     fn container_stop(
         &self,
         id: &str,
-        t: Option<i32>,
+        t: Option<&i32>,
     ) -> Box<dyn Future<Item = (), Error = Error<serde_json::Value>> + Send>;
     fn container_top(
         &self,
@@ -1077,7 +1077,7 @@ where
     fn container_restart(
         &self,
         id: &str,
-        t: Option<i32>,
+        t: Option<&i32>,
     ) -> Box<dyn Future<Item = (), Error = Error<serde_json::Value>> + Send> {
         let configuration: &configuration::Configuration<C> = self.configuration.borrow();
 
@@ -1243,7 +1243,7 @@ where
     fn container_stop(
         &self,
         id: &str,
-        t: Option<i32>,
+        t: Option<&i32>,
     ) -> Box<dyn Future<Item = (), Error = Error<serde_json::Value>> + Send> {
         let configuration: &configuration::Configuration<C> = self.configuration.borrow();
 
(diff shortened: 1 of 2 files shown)


edgelet/docker-rs/src/apis/container_api.rs
```diff
@@ -142,7 +142,7 @@ pub trait ContainerApi: Send + Sync {
     fn container_restart(
         &self,
         id: &str,
-        t: i32,
+        t: Option<&i32>,
```
arsing: `Option<i32>`

edgelet/docker-rs/src/apis/container_api.rs
```diff
@@ -1077,16 +1077,26 @@ where
...
     ) -> Box<dyn Future<Item = (), Error = Error<serde_json::Value>> + Send> {
         let configuration: &configuration::Configuration<C> = self.configuration.borrow();
 
         let method = hyper::Method::POST;
+        let query = if let Some(ref timeout) = t {
```
arsing: ... and `if let Some(timeout) = t`

edgelet/docker-rs/src/apis/container_api.rs
```diff
@@ -1077,16 +1077,26 @@ where
...
+            None
+        };
 
-        let query = ::url::form_urlencoded::Serializer::new(String::new())
-            .append_pair("t", &t.to_string())
-            .finish();
-        let uri_str = format!("/containers/{id}/restart?{}", query, id = id);
+        let uri_str = format!(
+            "/containers/{id}/restart?{}",
+            query.unwrap_or_else(|| "".to_string()),
```
arsing: If `query` is going to become `""` anyway, you might as well do

```rust
let query = t.map_or(std::borrow::Cow::Borrowed(""), |t| std::borrow::Cow::Owned(::url:: ... .finish()));
```

PR #1672, opened 2019-09-04
Change Content-Length to be the UTF8 Byte Count

CloudyEMS:
This PR is supposed to fix the current behaviour when a message payload contains umlauts.

### Current behaviour
When a payload that contains umlauts (e.g. `ü, ö, ä`) is sent, an unhandled exception is thrown with the following message `System.InvalidOperationException: Response Content-Length mismatch: too many bytes written`

This happens because the Content-Length HTTP header is set to be the length of the string, and not the byte count. 

### New behaviour
The Content-Length is correctly set to the UTF8 byte count and no exception is thrown. 
The UTF8 encoding was used, as it is already being used in other sections of the code.

### DIFF
--- a/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
+++ b/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
@@ -107,7 +107,15 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Http.Test
             var validator = new Mock<IValidator<MethodRequest>>();
             validator.Setup(v => v.Validate(It.IsAny<MethodRequest>()));
 
-            var testController = new TwinsController(Task.FromResult(edgeHub.Object), validator.Object);
+            var httpContext = new DefaultHttpContext();
+            var controllerContext = new ControllerContext()
+            {
+                HttpContext = httpContext,
+            };
+            var testController = new TwinsController(Task.FromResult(edgeHub.Object), validator.Object)
+            {
+                ControllerContext = controllerContext
+            };
             testController.OnActionExecuting(actionExecutingContext);
 
             string toDeviceId = "device1";
@@ -124,6 +132,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Http.Test
             Assert.NotNull(methodResult);
             Assert.Equal(200, methodResult.Status);
             Assert.Equal(new JRaw(responsePayload), methodResult.Payload);
+            long methodResultByteCount = Encoding.UTF8.GetByteCount(JsonConvert.SerializeObject(methodResult));
+            Assert.Equal(methodResultByteCount, testController.Response.ContentLength);
             Assert.Equal(objectResult.StatusCode, (int)HttpStatusCode.OK);
         }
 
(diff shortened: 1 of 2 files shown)


edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
```diff
@@ -163,6 +163,190 @@ public async Task TestInvokeMethodOnModuleWithResponsePayload()
...
+            Assert.NotNull(methodResult);
+            Assert.Equal(200, methodResult.Status);
+            Assert.Equal(new JRaw(responsePayload), methodResult.Payload);
+            long methodResultByteCount = Encoding.UTF8.GetByteCount(JsonConvert.SerializeObject(methodResult));
+            Assert.Equal(methodResultByteCount, testController.Response.ContentLength);
+            Assert.Equal(objectResult.StatusCode, (int)HttpStatusCode.OK);
+        }
+
+        [Fact]
+        public async Task TestInvokeMethodOnModuleWithResponsePayloadWithNoUmlautAndResponseContentLengthHeaderIsEqualToByteCount()
```
SebastianRoether: Not sure if a test specifically for "no umlauts" is necessary (when we already have one including umlauts)

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
```diff
@@ -163,6 +163,190 @@ public async Task TestInvokeMethodOnModuleWithResponsePayload()
...
+            IActionResult actionResult = await testController.InvokeModuleMethodAsync(WebUtility.UrlEncode(toDeviceId), WebUtility.UrlEncode(toModuleId), methodRequest);
+
+            Assert.NotNull(actionResult);
+            var objectResult = actionResult as ObjectResult;
+            Assert.NotNull(objectResult);
+            var methodResult = objectResult.Value as MethodResult;
+            Assert.NotNull(methodResult);
+            Assert.Equal(200, methodResult.Status);
+            Assert.Equal(new JRaw(responsePayload), methodResult.Payload);
+            Assert.Equal(new JRaw(responsePayload), methodResult.Payload);
```
SebastianRoether: Duplicate line

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
```diff
@@ -163,6 +163,190 @@ public async Task TestInvokeMethodOnModuleWithResponsePayload()
...
+            long methodResultByteCount = Encoding.UTF8.GetByteCount(JsonConvert.SerializeObject(methodResult));
+            Assert.Equal(methodResultByteCount, testController.Response.ContentLength);
+            Assert.Equal(objectResult.StatusCode, (int)HttpStatusCode.OK);
```
SebastianRoether: Shouldn't expected be first argument?

CloudyEMS: The test should check if the `Content-Length` header (actual) has the same byte count as the text being sent (expected).

arsing: SebastianRoether's comment is for line 347, where the parameters are indeed backwards.

arsing:
The general fix seems right to me.

The duplicate assert and the backwards expected/actual for the status code asserts that SebastianRoether pointed out do need to be fixed.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
```diff
@@ -163,6 +163,190 @@ public async Task TestInvokeMethodOnModuleWithResponsePayload()
             Assert.Equal(objectResult.StatusCode, (int)HttpStatusCode.OK);
         }
 
+        [Fact]
+        public async Task TestInvokeMethodWithResponsePayloadWithNoUmlautAndResponseContentLengthIsEqualToByteCount()
```
arsing: You can add the content-length assers into the existing `TestInvokeMethod*WithResponsePayload` tests. I don't think they need to be separate tests.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
```diff
@@ -163,6 +163,190 @@ public async Task TestInvokeMethodOnModuleWithResponsePayload()
...
+            Assert.Equal(200, methodResult.Status);
+            Assert.Equal(new JRaw(responsePayload), methodResult.Payload);
+            Assert.Equal(new JRaw(responsePayload), methodResult.Payload);
+            long methodResultByteCount = Encoding.UTF8.GetByteCount(JsonConvert.SerializeObject(methodResult));
+            Assert.Equal(methodResultByteCount, testController.Response.ContentLength);
+            Assert.Equal(objectResult.StatusCode, (int)HttpStatusCode.OK);
+        }
+
+        [Fact]
+        public async Task TestInvokeMethodWithResponsePayloadWithUmlautAndResponseContentLengthIsEqualToByteCount()
```
arsing: `TestInvokeMethodWithResponsePayloadWithMultiByteCharacters`

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
```diff
@@ -163,6 +163,190 @@ public async Task TestInvokeMethodOnModuleWithResponsePayload()
...
+            Assert.NotNull(methodResult);
+            Assert.Equal(200, methodResult.Status);
+            Assert.Equal(new JRaw(responsePayload), methodResult.Payload);
+            long methodResultByteCount = Encoding.UTF8.GetByteCount(JsonConvert.SerializeObject(methodResult));
+            Assert.Equal(methodResultByteCount, testController.Response.ContentLength);
+            Assert.Equal(objectResult.StatusCode, (int)HttpStatusCode.OK);
+        }
+
+        [Fact]
+        public async Task TestInvokeMethodOnModuleWithResponsePayloadWithUmlautAndResponseContentLengthHeaderIsEqualToByteCount()
```
arsing: `TestInvokeMethodOnModuleWithResponsePayloadWithMultiByteCharacters`

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
CloudyEMS: You are right. I will edit them in the other tests as well.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/TwinsControllerTest.cs
CloudyEMS: You are right.

SebastianRoether:
Is there anything we can do to speed up the process? This bug is a major annoyance in out edge framework application

vipeller:
We are investigating a more general fix to this problem.

vipeller:
@SebastianRoether @CloudyEMS thank you for the fix, it showed a general inconsistency in the code. We are working on a fix for the broader problem that would be merged in a couple of days. Also, sorry for the delay. I will let you know when it is there.

vipeller:
The related change is now merged, which should solve the UTF8 problem.

CloudyEMS:
Thanks for the fix. When should we expect this to be pushed to the Microsoft container registry?

PR #1673, opened 2019-09-04
Update E2E test yaml

philipktlin:
Avoid git clean step fail if it is first time and git repository doesn't exist yet.

Tested in E2E vsts pipeline and it get through git clean step.
https://dev.azure.com/msazure/One/_build/results?buildId=24749467

### DIFF
--- a/builds/e2e/e2e.yaml
+++ b/builds/e2e/e2e.yaml
@@ -27,7 +27,7 @@ jobs:
         # We use a self-hosted agent for this job, so we need to handle
         # cleanup of root-owned files ourselves (anything created during
         # `sudo dotnet vstest ...`)
-        sudo git clean -ffdx .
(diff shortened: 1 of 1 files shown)


builds/e2e/e2e.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
         # We use a self-hosted agent for this job, so we need to handle
         # cleanup of root-owned files ourselves (anything created during
         # `sudo dotnet vstest ...`)
-        sudo git clean -ffdx .
+        sudo git clean -ffdx . | true
```
damonbarry: I believe you want to use logical OR `||` here, not pipe `|`. We use logical OR for this pattern in our other scripts. Also, the pipe will swallow stdout from the `git clean` command...I think we'd like to see the command's output.

PR #1684, opened 2019-09-06
Update the Kubernetes agent system info

richma-ms:
Update the Kubernetes runtime system info to return the following:
OS = "Kubernetes"
version = K8s API version
arch = list of architectures in the cluster (and the node count for each)

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs
@@ -153,9 +153,9 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
             }
 
             // Construct the accumulated architecture string
-            foreach (var entry in archList)
+            foreach (var (arch, count) in archList)
             {
-                accumulatedArch += entry.arch + "(" + entry.count + ")\t";
+                accumulatedArch += arch + "(" + count + ")\t";
             }
 
 
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs
@@ -133,7 +133,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
                         if (archList[i].arch == node.Status.NodeInfo.Architecture)
                         {
                             // Just increase the node count
-                            archList[i] = (archList[i].arch, archList[i].count+1);
+                            archList[i] = (archList[i].arch, archList[i].count + 1);
                             haveSeen = true;
                             break;
                         }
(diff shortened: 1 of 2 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs
```diff
@@ -110,31 +110,55 @@ public async Task<Stream> GetModuleLogs(string module, bool follow, Option<int>
...
+
+            // Use "Kubernetes" and the current API version as analogy to device OS/version
             string osType = "Kubernetes";
-            string arch = "Kubernetes";
-            string version = "Kubernetes";
-            return Task.FromResult(new SystemInfo(osType, arch, version));
+            string version = k8sNodes.ApiVersion;
+
+            // Enumerate each node and count the unique architectures
+            var archList = new List<(string arch, int count)>();
```
dmolokanov: I think it's a very C/C++ way of doing this stuff. In a C# we have a very powerful way to work with collections called LINQ. 
This code can be re-written in a more declarative way:
```
		var achitectures = k8sNodes.Items
			.Where(node => node?.Status?.NodeInfo != null)
			.Select(node => node.Status.Status.NodeInfo.Architecture)
			.GroupBy(architecture => architecture)
			.Select(architectures => $"{architectures.Key} ({architectures.Count})";
		
		var arch = string.Join("\t", achitectures);
		return new SystemInfo(osType, arch, version);
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs
```diff
@@ -110,31 +110,55 @@ public async Task<Stream> GetModuleLogs(string module, bool follow, Option<int>
...
-            // }
+            V1NodeList k8sNodes = await this.client.ListNodeAsync();
+            string accumulatedArch = string.Empty;
+
+            // Use "Kubernetes" and the current API version as analogy to device OS/version
             string osType = "Kubernetes";
-            string arch = "Kubernetes";
-            string version = "Kubernetes";
-            return Task.FromResult(new SystemInfo(osType, arch, version));
+            string version = k8sNodes.ApiVersion;
```
dmolokanov: I think we need to take a version of k8s api server by making http request to `/version` api. I believe the c# client should have this method implemented

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs
```diff
@@ -110,31 +110,55 @@ public async Task<Stream> GetModuleLogs(string module, bool follow, Option<int>
...
+                {
+                    Events.NullNodeInfoResponse(node?.Metadata?.Name ?? "UNKNOWN");
+                    break;
+                }
+            }
+
+            // Construct the accumulated architecture string
+            foreach (var (arch, count) in archList)
+            {
+                accumulatedArch += arch + "(" + count + ")\t";
```
dmolokanov: @darobs Should we preserve the same structure as Docker mode or we can make it different? For example like this: 
```
{
  OperatingSystemType: "Kubernetes",
  Version: "1.12",
  Architectures: [
    { Name: "x86", NodesCount: 10 },
    { Name: "arm64", NodesCount: 1}
  ]
}
```

richma-ms:
Abandoning this PR as listing nodes requires a ClusterRole, and we don't want the agent to have a ClusterRole.  Going to try a different approach.

PR #1697, opened 2019-09-11
Do not deleted service accounts if deloyment is updated, fix hubname in

darobs:
iotedged.


edgelet/edgelet-kube/src/convert/mod.rs
```diff
@@ -12,6 +12,21 @@ pub use self::to_k8s::{
     trust_bundle_to_config_map,
 };
 
+const DNS_DOMAIN_MAX_SIZE: usize = 253;
+
+pub fn sanitize_dns_domain(domain: &str) -> Result<String> {
+    let sanitized = domain
+        .split('.')
```
dmolokanov:
```suggestion
        .split(".")
```

PR #1699, opened 2019-09-11
support manufacturing flows

seanmcd-msft:
support manufacturing flows that deploy the .cab into the image with different mechanism then the script but still need the script for the remaining initialization.

### DIFF
--- a/scripts/windows/setup/IotEdgeSecurityDaemon.ps1
+++ b/scripts/windows/setup/IotEdgeSecurityDaemon.ps1
@@ -336,6 +336,7 @@ function Initialize-IoTEdge {
     }
 
     New-Sockets $EdgeDataDirectory
+    Set-SystemPath
 
     # config.yaml
     Write-Host 'Generating config.yaml...'
--- a/scripts/windows/setup/IotEdgeSecurityDaemon.ps1
+++ b/scripts/windows/setup/IotEdgeSecurityDaemon.ps1
@@ -65,6 +65,23 @@ enum ContainerOs {
     Windows
 }
 
+function New-Sockets([string] $EdgeDataDirectory) {
+    foreach ($name in 'mgmt', 'workload') {
+        # We can't bind socket files directly in Windows, so create a folder
+        # and bind to that. The folder needs to give Modify rights to a
+        # well-known group that will exist in any container so that
+        # non-privileged modules can access it.
+        $path = "$EdgeDataDirectory\$name"
+        New-Item $Path -ItemType Directory -Force | Out-Null
+        $sid = New-Object System.Security.Principal.SecurityIdentifier 'S-1-5-11' # NT AUTHORITY\Authenticated Users
+        $rule = New-Object -TypeName System.Security.AccessControl.FileSystemAccessRule(`
+            $sid, 'Modify', 'ObjectInherit', 'InheritOnly', 'Allow')
+        $acl = Get-Acl -Path $path
+        $acl.AddAccessRule($rule)
+        Set-Acl -Path $path -AclObject $acl
+    }
+}
+
 <#
 .SYNOPSIS
 
@@ -252,7 +269,8 @@ function Initialize-IoTEdge {
         [String] $Username,
 
         # Password used to access the container registry and pull the IoT Edge Agent image.
-        [SecureString] $Password
+        [SecureString] $Password,
+        [switch] $Force=$false
     )
 
     switch ($PSCmdlet.ParameterSetName) {
@@ -309,13 +327,15 @@ function Initialize-IoTEdge {
     Setup-Environment -ContainerOs $ContainerOs -SkipArchCheck -SkipBatteryCheck
 
     $configPath = Join-Path -Path $EdgeDataDirectory -ChildPath 'config.yaml'
-    if (Test-Path $configPath) {
+    if ((-not $Force) -and (Test-Path $configPath)) {
         Write-HostRed
         Write-HostRed "$configPath already exists."
         Write-HostRed ('Delete it using "Uninstall-IoTEdge -Force" and then ' +
             're-run "Deploy-IoTEdge" and "Initialize-IoTEdge"')
         throw
     }
+
+    New-Sockets $EdgeDataDirectory
 
     # config.yaml
     Write-Host 'Generating config.yaml...'
@@ -1291,7 +1311,6 @@ function Try-StopService([string] $Name) {
         }
     }
 }
-
 function Get-IoTEdge([ref] $RestartNeeded, [bool] $Update) {
     try {
         # If we create the archive ourselves, then delete it when we're done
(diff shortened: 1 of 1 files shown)


scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -252,7 +269,8 @@ function Initialize-IoTEdge {
         [String] $Username,
 
         # Password used to access the container registry and pull the IoT Edge Agent image.
-        [SecureString] $Password
+        [SecureString] $Password,
+        [switch] $Force=$false
```
philipktlin:
```suggestion
        # flag to force installation
        [switch] $Force=$false
```

philipktlin: if $Force set to true, but IoT Edge is already installed, what happened? should it work?

arsing:
Rebase it on master and get rid of all the unnecessary commits.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
arsing: Why is this parameter added? Needs to be in the commit message as well as the helptext.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -309,14 +327,17 @@ function Initialize-IoTEdge {
...
+    if ((-not $Force) -and (Test-Path $configPath)) {
         Write-HostRed
         Write-HostRed "$configPath already exists."
         Write-HostRed ('Delete it using "Uninstall-IoTEdge -Force" and then ' +
             're-run "Deploy-IoTEdge" and "Initialize-IoTEdge"')
         throw
     }
 
+    New-Sockets $EdgeDataDirectory
+    Set-SystemPath
```
arsing: This was removed from here intentionally. Why are you adding it back?

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -65,6 +65,23 @@ enum ContainerOs {
     Windows
 }
 
+function New-Sockets([string] $EdgeDataDirectory) {
```
arsing: AFAICT it would've been sufficient to move this code from `Deploy-` code path to `Initialize-`. I don't see why you need to extract it into a function and then call it from two different places.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: the issue here is that iot-core can't uninstall a .cab.  so, to upgrade you need to upgrade the cab and re-run initialize.  initialize should be idempotent anyway imho.  so, that should be ok.  rather then add a -force I can add an iotcore check to skip the error @line 331-336.  would that be better?

seanmcd-msft: answered in previous comment above.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: multiple customers are reporting this to us as a bug.    'iotedge check' and docker cli don't work after install.  why was this removed?

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: since everything else was already moved from deploy to initialize i assumed it was being done in deploy for some scenario i didn't understand.  if you want me to just move it then i can do that.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
arsing: https://github.com/Azure/iotedge/commit/87957c9a92b4df2a7fb5e77951e60184bc94e6a2

>Add the IoT Edge and Moby install directories to PATH in `Deploy-IoTEdge`
>rather than in `Initialize-IoTEdge`. On IoT Core both `Deploy-` and updating
>PATH require a reboot, so it's good to combine those reboots into one.
>
>Fixes #1404

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
arsing: No, there's no reason it needs to be in `Deploy-`

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: but, for manufacturing flows that don't use deploy in the script then they don't get the path set ever.  presumably we could check in initialize and only add to the path if it isn't already there. but, i'd rather treat that as a separate new work item and get this PR completed.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
arsing: Why does `Initialize-` need to be re-run?

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
arsing: Okay, if "manufacturing flows" don't use `Deploy-` then yes it needs to be in both.

>presumably we could check in initialize and only add to the path if it isn't already there

`Set-SystemPath` does that already.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: in my case, it was because I was iterating on script changes.  but, in theory if a newer version of the cab has additional or different setup requirements you can't uninstall and re-install on iot core. you need to upgrade the .cab and then rerun init to get the additional settings or whatever other relevant changes there may be.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: ok, i left it in the new function but am only calling it from init.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
arsing: We'll never require running `Initialize-` again. It's impossible for non-Windows because updates happen transparently there and `/etc/iotedge/config.yaml` is never re-generated. So you don't have to worry about that.

seanmcd-msft: what about when we switched to UDS and had to add functionality to create the new socket folders?  that's the kind of thing i'm talking about.  currently the script expects you to uninstall and re-install for that sort of case.  but, on iotcore you can't do that. if you can't force a re-init then you have to re-flash the device.

arsing: That happened before GA.

If that happens again, we can worry about it then.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
arsing: You're calling it in `Get-IoTEdge` too which gets called from `Deploy-`.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -252,7 +269,7 @@ function Initialize-IoTEdge {
         [String] $Username,
 
         # Password used to access the container registry and pull the IoT Edge Agent image.
-        [SecureString] $Password
+        [SecureString] $Password,
```
arsing: This trailing comma is a syntax error.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1291,7 +1316,6 @@ function Try-StopService([string] $Name) {
         }
     }
 }
-
 function Get-IoTEdge([ref] $RestartNeeded, [bool] $Update) {
```
arsing: Restore the empty line you deleted here.

PR #1704, opened 2019-09-12
Set max total wal size for RocksDB to 512M

ancaantochi:
The update of rocksdb is causing WAL (write ahead log) files to keep growing to the default number of 1000 and each file has default size 64Mb. This limits the size of all WAL files to 512Mb by default, it can also be configured from an Env variable.


edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbOptionsProvider.cs
```diff
@@ -40,13 +41,21 @@ public class RocksDbOptionsProvider : IRocksDbOptionsProvider
...
+        readonly ulong maxTotalWalSize = 512 * 1024 * 1024;
+
         readonly ISystemEnvironment env;
         readonly bool optimizeForPerformance;
 
         public RocksDbOptionsProvider(ISystemEnvironment env, bool optimizeForPerformance)
         {
             this.env = Preconditions.CheckNotNull(env);
             this.optimizeForPerformance = optimizeForPerformance;
+            if (ulong.TryParse(Environment.GetEnvironmentVariable("RocksDB_MaxTotalWalSize"), out ulong envMaxTotalWalSize))
```
varunpuranik: Nit: For consistency, we should get the value from DependencyManager.cs and pass it down. That is how we do it everywhere else in the code.

and-rewsmith:
Not ready to be merged yet. I still need to fix windows environment to mount iotedge/storage.

and-rewsmith:
Reopening, however we still need to fix behavior with failing Windows AMD64 environment

scripts/windows/test/Run-E2ETest.ps1
```diff
@@ -1317,6 +1317,15 @@ Function SetEnvironmentVariable
     $env:Path="$env:Path;C:\Program Files\iotedge-moby;C:\Program Files\iotedge"
 }
 
+Function SetupMountedStorage
+{
+    If (!(Test-Path "C:\data\edgehub"))
```
and-rewsmith: I will make var for this path when I make the same changes to master, then port it back.

scripts/windows/test/Setup-Env.ps1
```diff
@@ -68,6 +68,11 @@ $IoTEdgedArtifactFolder = Join-Path $E2ETestFolder "artifacts\iotedged-windows"
 $InstallationScriptPath = Join-Path $E2ETestFolder "artifacts\core-windows\scripts\windows\setup\IotEdgeSecurityDaemon.ps1"
 Invoke-Expression $InstallationScriptPath
 
+If ((Test-Path "C:\data\edgehub"))
```
and-rewsmith: I will make var for this path when I make the same changes to master, then port it back. We likely don't have time to rebuild the images.

and-rewsmith:
This should now be merged so we can start long haul for 1.0.9 following 1.0.8.3 release.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbOptionsProvider.cs
ancaantochi: Changed to read from config in DependencyManager

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
```diff
@@ -294,6 +295,21 @@ static string GetLocalConfigFilePath(IConfiguration configuration, ILogger logge
             return localConfigPath;
         }
 
+        static Option<ulong> GetStorageMaxTotalWalSizeIfExists(IConfiguration configuration)
+        {
+            ulong storageMaxTotalWalSize = 0;
+            try
+            {
+                storageMaxTotalWalSize = configuration.GetValue<ulong>(Constants.StorageMaxTotalWalSize);
```
myagley: There's no way to test for the presence of the config value without having to catch an exception? Like a `TryGetValue`?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
```diff
@@ -294,6 +295,21 @@ static string GetLocalConfigFilePath(IConfiguration configuration, ILogger logge
             return localConfigPath;
         }
 
+        static Option<ulong> GetStorageMaxTotalWalSizeIfExists(IConfiguration configuration)
```
myagley: This code looks like it is in the DependencyManager too?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
ancaantochi: I doesn't have trygetvalue, this only throws when the value can not be converted to ulong. If the value is not set in the env variable it returns 0. For other settings edgeHub initialization fails it the value can't be converted to what is expected, so I can remove the try catch and have the same behavior

PR #1708, opened 2019-09-12
[k8s] Refactor CrdWatcher

dmolokanov:
- Renamed CrdWatchOperator to EdgeDeploymentOperator
- Renamed CrdWatcher to EdgeDeploymentController
- EdgeDeploymentOperator only 
    1. watch CRD and delegates modules deployment on k8s to EdgeDeploymentController
    2. stores current modules
- EdgeDeploymentController make all checks against k8s api and deploy modules with services and service accounts


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/EdgeDeploymentOperator.cs
```diff
@@ -0,0 +1,166 @@
...
+                            await this.HandleEdgeDeploymentsChangedAsync(type, item);
+                        }
+                        catch (Exception ex) when (!ex.IsFatal())
+                        {
+                            Events.ExceptionInCustomResourceWatch(ex);
+                        }
+                    },
+                    onClosed: () =>
+                    {
+                        Events.CrdWatchClosed();
```
darobs: A really minor comment: If we've renamed everything else, we should also update the Events name...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/EdgeDeploymentOperator.cs
```diff
@@ -0,0 +1,166 @@
...
+                ExceptionInCustomResourceWatch = IdStart,
+                DeploymentStatus,
+                DeploymentError,
+                DeploymentNameMismatch,
+                CrdWatchClosed,
+            }
+
+            public static void ExceptionInCustomResourceWatch(Exception ex)
+            {
+                Log.LogError((int)EventIds.ExceptionInCustomResourceWatch, ex, "Exception caught in Custom Resource Watch task.");
```
darobs: ...and the debug messages.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/EdgeDeploymentOperator.cs
dmolokanov: Good. Will do it in the next PR

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/EdgeDeploymentOperator.cs
dmolokanov: sure

PR #1712, opened 2019-09-13
Set max total wal size to 512M

myagley:


edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbOptionsProvider.cs
```diff
@@ -40,13 +41,21 @@ public class RocksDbOptionsProvider : IRocksDbOptionsProvider
...
+        readonly ulong maxTotalWalSize = 512 * 1024 * 1024;
+
         readonly ISystemEnvironment env;
         readonly bool optimizeForPerformance;
 
         public RocksDbOptionsProvider(ISystemEnvironment env, bool optimizeForPerformance)
         {
             this.env = Preconditions.CheckNotNull(env);
             this.optimizeForPerformance = optimizeForPerformance;
+            if (ulong.TryParse(Environment.GetEnvironmentVariable("RocksDB_MaxTotalWalSize"), out ulong envMaxTotalWalSize))
```
varunpuranik: Nit: For consistency, we should get the value from DependencyManager.cs and pass it down. That is how we do it everywhere else in the code.

myagley: I cherry picked this from master. Can we merge as is and then fix master?

PR #1714, opened 2019-09-14
[k8s] Refactor runtime info provider

dmolokanov:
- Move pod list watch routine to KubernetesEnvironmentOperator
- Add ResourseName object instead of pair hostname and deviceId
- Fix tets


edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/commands/KubernetesCrdCommandTest.cs
```diff
@@ -44,12 +40,7 @@ public void CrdCommandCreateValidation()
             IModule m1 = new DockerModule("module1", "v1", ModuleStatus.Running, Core.RestartPolicy.Always, Config1, ImagePullPolicy.OnCreate, DefaultConfigurationInfo, EnvVars);
             KubernetesModule km1 = new KubernetesModule(m1 as IModule<DockerConfig>, config);
             KubernetesModule[] modules = { km1 };
-            Assert.Throws<ArgumentException>(() => new KubernetesCrdCommand(null, Hostname, DeviceId, DefaultClient, modules, Runtime, DefaultConfigProvider));
-            Assert.Throws<ArgumentException>(() => new KubernetesCrdCommand(Ns, null, DeviceId, DefaultClient, modules, Runtime, DefaultConfigProvider));
-            Assert.Throws<ArgumentException>(() => new KubernetesCrdCommand(Ns, Hostname, null, DefaultClient, modules, Runtime, DefaultConfigProvider));
-            Assert.Throws<ArgumentNullException>(() => new KubernetesCrdCommand(Ns, Hostname, DeviceId, null, modules, Runtime, DefaultConfigProvider));
```
darobs: Why did you drop the tests for null client, modules, and config provider?  Why didn't I have a test for null runtime?

dmolokanov: I added ResourseName object instead of pair hostname and deviceId. Inside this object that is used everywhere, values cannot be invalid. Test case to create invalid ResourseName moved to appropriate tests. And I deleted those cases from everywhere else.

PR #1717, opened 2019-09-16
Get k8s systeminfo from edgelet

richma-ms:
This changes makes it so that the edgelet provides systeminfo (OS, arch, version) for the k8s agent.

UTs for the new functionality in edgelet will be coming in a later iteration of this PR.

### DIFF
--- a/edgelet/edgelet-kube/src/runtime.rs
+++ b/edgelet/edgelet-kube/src/runtime.rs
@@ -249,7 +249,7 @@ where
             .expect("Unexpected lock error")
             .borrow_mut()
             .list_nodes()
-            .map_err(|err| Error::from(err.context(ErrorKind::SystemInfo)))
+            .map_err(|err| Error::from(err.context(ErrorKind::RuntimeOperation(RuntimeOperation::SystemInfo))))
             .map(|nodes| {
                 // Accumulate the architectures and their node counts into a map
                 let architectures = nodes
@@ -432,7 +432,7 @@ mod tests {
         let mut runtime = Runtime::new().unwrap();
         let info = runtime.block_on(task).unwrap();
 
-        assert_eq!(info.architecture(), "[{\"name\":\"amd64\",\"nodes_count\":2},{\"name\":\"arm32\",\"nodes_count\":1}]");
+        assert_eq!(info.architecture(), "[{\"name\":\"amd64\",\"nodes_count\":2}]");
     }
 
     fn list_node_handler() -> impl Fn(Request<Body>) -> ResponseFuture + Clone {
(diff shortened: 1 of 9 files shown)


edgelet/edgelet-kube/src/runtime.rs
```diff
@@ -236,11 +237,39 @@ where
...
+            .map_err(|err| {
+                Error::from(err.context(ErrorKind::RuntimeOperation(RuntimeOperation::ListNodes)))
+            })
+            .map(|nodes| {
+                // Accumulate the architectures and their node counts into a hashmap
+                let mut arch_map: HashMap<String, u32> = HashMap::new();
+                for node in nodes.items.into_iter() {
+                    *(arch_map
+                        .entry(node.status.unwrap().node_info.unwrap().architecture)
+                        .or_insert(1)) += 1;
```
dmolokanov: We should insert 0 at the start and then add 1 for each item.

edgelet/edgelet-kube/src/runtime.rs
```diff
@@ -236,11 +237,39 @@ where
...
+                }
+
+                // Flatten the hashmap into a formatted string
+                let mut arch_string = String::from("[\n");
```
dmolokanov: Making the string as JSON array does not seem like the nicest solution. Initially, I had an idea  we add `SystemInfo` type that will live in `edgelete-kube` only. As opposite to `SystemInfo` in `edgelet-core` it will have collection `architectures: Vec<Architecture>`.

But what I figured out that this approach is rather complicated that it should be. I can describe what approach I have in mind in person if you are interested.

For now we can just all a new struct `Architecture` compose a list of architectures and serialize it to a json string.

```
#[derive(Debug, serde_derive::Serialize)]
pub struct Architecture{
    name: String,
    nodes_count: u32
}
```

edgelet/edgelet-kube/src/runtime.rs
```diff
@@ -236,11 +237,39 @@ where
     }
 
     fn system_info(&self) -> Self::SystemInfoFuture {
```
dmolokanov: I propose a little concise solution to the code below
```
    fn system_info(&self) -> Self::SystemInfoFuture {
        let fut = self
            .client
            .lock()
            .expect("Unexpected lock error")
            .borrow_mut()
            .list_nodes()
            .map_err(|err| {
                Error::from(err.context(ErrorKind::RuntimeOperation(RuntimeOperation::SystemInfo)))
            })
            .map(|nodes| {
                // Accumulate the architectures and their node counts into a map
                let architectures = nodes
                    .items
                    .into_iter()
                    .filter_map(|node| {
                        node.status
                            .and_then(|status| status.node_info.map(|info| info.architecture))
                    })
                    .fold(HashMap::new(), |mut architectures, current_arch| {
                        let count = architectures.entry(current_arch).or_insert(0);
                        *count += 1;
                        architectures
                    });

                // Convert a map to a list of architectures
                let architectures = architectures.into_iter().map(|(name, count)| Architecture {
                    name,
                    nodes_count: count,
                }).collect::<Vec<Architecture>>();

                SystemInfo::new("Kubernetes".to_string(), serde_json::to_string(&architectures).unwrap())
            });

        Box::new(fut)
```

edgelet/edgelet-kube/src/runtime.rs
```diff
@@ -236,11 +237,50 @@ where
...
+            nodes_count: u32,
+        };
+
+        let fut = self
+            .client
+            .lock()
+            .expect("Unexpected lock error")
+            .borrow_mut()
+            .list_nodes()
+            .map_err(|err| Error::from(err.context(ErrorKind::SystemInfo)))
```
dmolokanov:
```suggestion
            .map_err(|err| Error::from(err.context(ErrorKind::RuntimeOperation(RuntimeOperation::SystemInfo)))
```
And no need in `ErrorKind::SystemInfo` anymore

PR #1722, opened 2019-09-17
Single Quote Trimming in Routes

yophilav:
Bug Fix: 
1. Correct the parsing mechanism for the use of `'` in a route `BrokeredEndpoint()`
2. Add a unit test to verify the parsing of `'` in a route `BrokeredEndpoint()`


myagley:
Hey Bear, can you please add a description that discusses the bug that is being fixed real quick?

arsing:
Also, when you merge the PR, remember to modify the default commit message to be the text that you put in the PR description (the default will be the empty message in your commit).

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/routeFactory/RouteFactoryTest.cs
```diff
@@ -64,6 +64,10 @@ public void TestParseRouteWithSystemEndpoint(string routeString, IMessageSource
         [InlineData(@"FORM /messages/modules/adapter INTO brokeredEndpoint(""/modules/alertLogic/inputs/in"")")]
         [InlineData(@"SELECT FROM /messages/modules/adapter INTO brokeredEndpoint(""/modules/alertLogic/inputs/in"")")]
         [InlineData(@"FROM INTO brokeredEndpoint(""/modules/alertLogic/inputs/in"")")]
+        [InlineData(@"FORM /* INTO brokeredEndpoint(""/modules/alertLogic/inputs/in')")]
```
arsing: `FORM` -> `FROM` (the intention is for the parse to fail because of the mismatched quotes, not the misspelling), and `/messages` -> `/*` for consistency.

yophilav: Done

arsing:
As you can see from the CI failures, some of those negative tests actually end up parsing successfully. So please make them positive tests, and also verify that they're *supposed* to parse successfully.

(I guess you did not run the tests yourself before you committed and pushed them? :) )

Related, please also see what happens with `@"FROM /* INTO brokeredEndpoint(""/modules/alertLogic/inputs/in'"")"` and `@"FROM /* INTO brokeredEndpoint('/modules/alertLogic/inputs/in""')"`. Specifically I suspect the former will parse successfully but it'll parse differently with and without your change, so it'll be backward-incompatible.

yophilav:
> As you can see from the CI failures, some of those negative tests actually end up parsing successfully. So please make them positive tests, and also verify that they're supposed to parse successfully.
> (I guess you did not run the tests yourself before you committed and pushed them? :) )
> Related, please also see what happens with @"FROM /* INTO brokeredEndpoint(""/modules/alertLogic/inputs/in'"")" and @"FROM /* INTO brokeredEndpoint('/modules/alertLogic/inputs/in""')". Specifically I suspect the former will parse successfully but it'll parse differently with and without your change, so it'll be backward-incompatible.

I need to break my bad habit of ` let's commit, push, then test to realize that it failed locally`. I need some more of your input before I make any changes of what format is support and not-supported route strings.

PR #1727, opened 2019-09-18
Update end-to-end build YAML for new tests

damonbarry:
Two new end-to-end tests have been added recently, but they both fail in automated builds because they rely on new context.json parameters and secrets that haven't been added to the YAML build. This change fixes that.

Note this change also makes the context.json parameter `caCertScriptPath` optional because it's only required for the `TransparentGateway` tests.

### DIFF
--- a/builds/e2e/templates/e2e-run.yaml
+++ b/builds/e2e/templates/e2e-run.yaml
@@ -19,6 +19,7 @@ steps:
     }
   displayName: Run tests
   env:
+    E2E_DPS_GROUP_KEY: $(TestDpsGroupKeySymmetric)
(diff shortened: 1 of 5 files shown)


builds/e2e/templates/e2e-run.yaml
```diff
@@ -19,7 +19,7 @@ steps:
     }
   displayName: Run tests
   env:
-    E2E_DPS_GROUP_KEY: $(TestDpsGroupKeySymmetric)
+    E2E_DPS_GROUP_KEY: '$(TestDpsGroupKeySymmetric)'
```
yophilav: Is there any other variable that can potentially be a base64 that also require this change?

PR #1729, opened 2019-09-19
Cleanup filesystem

SlavyMihov:
Add a cleanup step that removes residual files and links that docker creates with every image and eventually clog up the filesystem.
If not cleaned, the device will eventually hit the maximum allowed links supported by the NTFS filesystem and will start failing.


scripts/windows/test/Setup-Env.ps1
```diff
@@ -87,6 +87,7 @@ If ($osEdition -eq "IoTUAP")    # Windows IoT Core - update iotedge
             if($residualModules.Length -gt 0) {
                 docker -H npipe:////./pipe/iotedge_moby_engine rm -f $residualModules
             }
+	    docker -H npipe:////./pipe/iotedge_moby_engine system prune -a -f
```
philipktlin:
```suggestion
	    
	      docker -H npipe:////./pipe/iotedge_moby_engine system prune -a -f
```

philipktlin: just add spacing and add a new line after if-block.

SlavyMihov: Done.

PR #1733, opened 2019-09-19
[k8s] Refactor docker to k8s primitives conversion

dmolokanov:
- Refactor Kubernetes module to k8s primitives
- Simplified code for managing services, deployments and service accounts


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/Set.cs
```diff
@@ -0,0 +1,42 @@
...
+                .Select(key => this.items[key]);
+
+            // build list of items that are available in "this" set but not available in "other" set;
+            // this represents items that may need to be removed
+            IEnumerable<string> removed = other.items.Keys
+                .Except(this.items.Keys);
+
+            // build list of items that are available in both sets;
+            // this represents items that may be updated
+            IEnumerable<T> updated = this.items.Keys
```
darobs: I don't like calling this "updated" if there's no indication that there's a difference between set[key] and other[key]. Should T bet restricted to `IEquatable<T>` so a comparison may be made?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/DiffCollectionExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    public static class DiffCollectionExtensions
+    {
+        public static Diff<T> Diff<T>(this IEnumerable<T> desired, IEnumerable<T> existing, Func<T, string> selector)
+        {
+            var either = new Set<T>(desired.ToDictionary(selector));
```
darobs: It's a minor thing, but can we keep the name consistent with "desired" and "existing"?  "new" and "old" maybe, or desiredSet and existingSet?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentCommand.cs
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Commands
+namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment
```
darobs: Do we want to keep this in "Commands" to be consistent with edgelet and docker projects?

dmolokanov: I don't think we should dedicate the whole folder for an only command... But I open to suggestions

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -0,0 +1,372 @@
...
+            var diff = desired.Diff(existing.Items, service => service.Metadata.Name);
+
+            // Update only those services if configurations have not matched
+            var updatingTask = diff.Updated
+                .Where(service => service.Metadata.Annotations[KubernetesConstants.CreationString] != creationStrings[service.Metadata.Name])
+                .Select(
+                    service =>
+                    {
+                        Events.UpdateService(service);
+                        return this.client.ReplaceNamespacedServiceAsync(service, service.Metadata.Name, this.deviceNamespace);
```
darobs: We haven't had much luck with "ReplaceNamespacedServiceAsync" in the past.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -0,0 +1,372 @@
...
+        async Task ManageDeployments(V1DeploymentList existing, IEnumerable<V1Deployment> desired)
+        {
+            // Compose creation strings from existing items
+            Dictionary<string, string> creationStrings = GetDeploymentConfig(existing);
+
+            // Update only deployments if configurations have not matched or it's an EdgeAgent with the same version
+            bool ShouldUpdate(V1Deployment deployment)
+            {
+                // current module is to update an existing one
+                if (deployment.Metadata.Annotations[KubernetesConstants.CreationString] != creationStrings[deployment.Metadata.Name])
```
darobs: This logic probably needs to move lower in this function.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
dmolokanov: Replace operations require `resourceVersion` to set from an existing resource.
In addition Service wants `ClusterIP` to be set.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentCommand.cs
darobs: We can always move it back if it confuses others.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/Set.cs
dmolokanov: Made `IEqualityComparer` for those types

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
dmolokanov: Got rid of this method

PR #1736, opened 2019-09-19
E2E Introduce Temperature Filter (Azure) Function

yophilav:
Introduce an end-to-end test for temperature filter function. 
This is an end-to-end test mimic the steps in "Deploy Azure functions as IoT Edge modules" tutorial: https://docs.microsoft.com/en-us/azure/iot-edge/tutorial-deploy-function.


edge-modules/functions/README.md
```diff
@@ -1,38 +1,43 @@
 # Function Sample
```
damonbarry: Looking at this file more carefully, I think I agree with your first impression...we might just want to remove it. The doc you referenced (https://docs.microsoft.com/en-us/azure/iot-edge/tutorial-deploy-function) already describes what you need to do, in a more generalized and user-friendly way.  :)

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -12,20 +12,24 @@ namespace Microsoft.Azure.Devices.Edge.Test
...
     public class Module : ManualProvisioningFixture
     {
-        private const string DefaultSensorImage = "mcr.microsoft.com/azureiotedge-simulated-temperature-sensor:1.0";
+        private string GetTempSensorImage()
+        {
+            const string DefaultSensorImage = "mcr.microsoft.com/azureiotedge-simulated-temperature-sensor:1.0";
+            return Context.Current.TempSensorImage.GetOrElse(DefaultSensorImage);
+        }
 
+        [Category("TempSensor")]
```
damonbarry: Why add this category?

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -12,20 +12,24 @@ namespace Microsoft.Azure.Devices.Edge.Test
 
     public class Module : ManualProvisioningFixture
     {
-        private const string DefaultSensorImage = "mcr.microsoft.com/azureiotedge-simulated-temperature-sensor:1.0";
+        private string GetTempSensorImage()
```
damonbarry: I'd have this function return a `(string name, string image)` tuple so that tests can do:

```
var tempSensor = GetTempSensor();
// ...
    builder => { builder.AddModule(tempSensor.name, tempSensor.image); },
// ...
EdgeModule sensor = deployment.Modules[tempSensor.name];
```

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -56,33 +60,72 @@ public async Task TempSensor()
...
 
+        [Category("TempSensor")]
         [Test]
         public async Task TempFilter()
         {
-            string sensorImage = Context.Current.TempSensorImage.GetOrElse(DefaultSensorImage);
-            string filterImage = Context.Current.TempFilterImage.Expect(() => new ArgumentException());
+            string filterImage = Context.Current.TempFilterImage.Expect(() => new ArgumentException("tempFilterImage parameter is required for TempFilter test"));
+
+            const string filterModName = "tempFilter";
```
damonbarry: "Mod" is not a common abbreviation for module so I wouldn't recommend using it in variable names.

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
damonbarry: I get that you might want to run all the tempSensor-related tests together, but I think you could do that with `--filter Name~Temp` (or `--filter FullyQualifiedName~Microsoft.Azure.Devices.Edge.Test.Module.Temp`).

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -56,33 +60,72 @@ public async Task TempSensor()
...
+
+            const string filterFuncModuleName = "tempFilterFunctions";
+            const string tempSensorModName = "tempSensor";
+
+            CancellationToken token = this.TestToken;
+
+            EdgeDeployment deployment = await this.runtime.DeployConfigurationAsync(
+                builder =>
+                {
+                             builder.AddModule(tempSensorModName, this.GetTempSensorImage());
```
damonbarry: Can you fix the indentation here?

test/Microsoft.Azure.Devices.Edge.Test/helpers/Context.cs
```diff
@@ -97,8 +97,9 @@ public Context()
             this.RootCaKeys = GetAndValidateRootCaKeys();
             this.SetupTimeout = TimeSpan.FromMinutes(context.GetValue("setupTimeoutMinutes", 5));
             this.TeardownTimeout = TimeSpan.FromMinutes(context.GetValue("teardownTimeoutMinutes", 2));
-            this.TempSensorImage = Option.Maybe(Get("tempSensorImage"));
+            this.TempFilterFunc = Option.Maybe(Get("tempFilterFunc"));
```
damonbarry: It pains me to make the variable longer, but this one should end in "Image" like the others (at least the context.json parameter, and probably the C# property as well. 😢 If we want to keep it shorter, I don't mind "tempFunctionImage".

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
yophilav: Side question: How do you know when to ` return by tuplex ` or using say ` GetTempSensor( out tempSensor ) ` ?

damonbarry: I think you should generally prefer tuples. From the `out` [docs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out-parameter-modifier): "Declaring a method with out arguments is a classic workaround to return multiple values. Beginning with C# 7.0, consider tuples for similar scenarios."

yophilav: I know you asked for a tuple but I'm giving a `tempSensor` class.

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
yophilav: ` filterModName ` has leveled up to be ` filterModuleName `

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
yophilav: Oh my, how did it even get this bad. 😬

test/Microsoft.Azure.Devices.Edge.Test/helpers/Context.cs
yophilav: That's very reasonable. The member variable is now `TempFilterFuncImage` for the config JSON `tempFilterFuncImage`

yophilav:
I'm not removing `DeviceId` yet. I'm thinking it would be more related to the change regarding module uniqueness. In the last 2 commits are solely addressing the comments from this PR.

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
yophilav: Fact check: I'm giving you a tempSensor tuple as requested.

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -12,24 +12,25 @@ namespace Microsoft.Azure.Devices.Edge.Test
 
     public class Module : ManualProvisioningFixture
     {
-        private string GetTempSensorImage()
+        private (string Name, string Image) GetTempSensorImage()
```
damonbarry: Name and Image should be lowercased (technically camelCase).

damonbarry: ...but actually, we have a problem here where the name "tempSensor" will be used for three different modules on a single device (when all of the tests are run together), which IoT Hub doesn't like (at least, not within such a short period of time) and which might cause test failures. So this method probably needs to go away, replaced by a class-level const for the tempSensor image, and then unique names for the tempSensor identity in each test.

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -12,6 +12,35 @@ namespace Microsoft.Azure.Devices.Edge.Test
...
     public class Module : ManualProvisioningFixture
     {
+        private sealed class TempSensorModule
+        {
+            public string Name { get; }
+            public string Image { get; }
+
+            private const string DefaultSensorImage = "mcr.microsoft.com/azureiotedge-simulated-temperature-sensor:1.0";
+            // Make sure it's thread-safe
+            private static readonly object padlock = new object();
```
damonbarry: I wouldn't introduce a lock for thread safety unless there's a demonstrated need for it.

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -12,6 +12,35 @@ namespace Microsoft.Azure.Devices.Edge.Test
...
+            private static int instanceCount = 0;
+
+            private TempSensorModule()
```
damonbarry: Did StyleCop warn about this? I'd remove it since it's empty, it's never used, and using it would put the object in a bad state (`Name` and `Image` would be `null` I think).

PR #1739, opened 2019-09-20
Allow hostnames to contain upper-case letters

arsing:
This change removes the additional restriction that hostnames must not contain
upper-case A-Z. The reason for the restriction was to be consistent with
Kubernetes services, but we've received feedback from non-Kubernetes users
that they'd like to use upper-case.


edgelet/iotedge/src/check/mod.rs
```diff
@@ -1964,22 +1964,26 @@ fn is_rfc_1035_valid(name: &str) -> bool {
             Some(c) => c,
             None => return false,
         };
-        if first_char < 'a' || first_char > 'z' {
+        let first_char_invalid =
+            (first_char < 'a' || first_char > 'z') && (first_char < 'A' || first_char > 'Z');
```
darobs: Could this be replaced by  `is_ascii_alphabetic()`?

arsing: Oh yeah, good point. Done.

edgelet/iotedge/src/check/mod.rs
```diff
@@ -1964,22 +1964,26 @@ fn is_rfc_1035_valid(name: &str) -> bool {
...
+        if first_char_invalid {
             return false;
         }
 
-        if label
-            .chars()
-            .any(|c| (c < 'a' || c > 'z') && (c < '0' || c > '9') && c != '-')
-        {
+        let any_char_invalid =
+            |c| (c < 'a' || c > 'z') && (c < 'A' || c > 'Z') && (c < '0' || c > '9') && c != '-';
```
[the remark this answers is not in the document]
arsing: Done.

arsing:
Done.

PR #1743, opened 2019-09-23
#5321638: RuntimeInfoProvider.GetModules is racy

vipeller:
causing parallel unit tests failing. Modified RuntimeInfoProvider to handle exceptions caused by requesting information about removed modules.

### DIFF
--- a/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/RuntimeInfoProviderTest.cs
+++ b/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/RuntimeInfoProviderTest.cs
@@ -340,7 +340,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker.Test
             var missingContainerName = "container2";
             var containersToReturn = new[] { "container1", "container2", "container3" };
 
-            var dockerClient = SetupDockerClient(missingContainerName, containersToReturn);
+            var dockerClient = this.SetupDockerClient(missingContainerName, containersToReturn);
             var provider = await RuntimeInfoProvider.CreateAsync(dockerClient);
 
             var moduleInfo = await provider.GetModules(CancellationToken.None);
@@ -350,8 +350,8 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker.Test
 
             var expectedContainerNames = containersToReturn
                                             .Where(c => !c.Equals(missingContainerName))
-                                            .Select(c => c.Substring(1))    // cuts the first letter
-                                            .Concat(new[] { "dgeAgent" });  // added anyway
+                                            .Select(c => c.Substring(1))
+                                            .Concat(new[] { "dgeAgent" });
 
             Assert.Equal(expectedContainerNames, actualContainerNames);
         }
@@ -363,7 +363,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker.Test
             var missingContainerName = "edgeAgent";
             var containersToReturn = new[] { "container1", "container2", "container3" };
 
-            var dockerClient = SetupDockerClient(missingContainerName, containersToReturn);
+            var dockerClient = this.SetupDockerClient(missingContainerName, containersToReturn);
             var provider = await RuntimeInfoProvider.CreateAsync(dockerClient);
 
             var moduleInfo = await provider.GetModules(CancellationToken.None);
@@ -372,7 +372,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker.Test
                                             .Select(m => m.Name);
 
             var expectedContainerNames = containersToReturn
-                                            .Select(c => c.Substring(1));    // cuts the first letter
+                                            .Select(c => c.Substring(1));
 
             Assert.Equal(expectedContainerNames, actualContainerNames);
         }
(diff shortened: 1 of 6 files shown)


edge-util/src/Microsoft.Azure.Devices.Edge.Util/Option.cs
```diff
@@ -202,6 +202,25 @@ public Option<T> Filter(Func<T, bool> predicate)
 
     public static class Option
     {
+        public static IEnumerable<T> FilterMap<T>(this IEnumerable<Option<T>> source, Func<T, bool> predicate)
+        {
+            Preconditions.CheckNotNull(source, nameof(source));
+            Preconditions.CheckNotNull(predicate, nameof(predicate));
+
+            foreach (var item in source)
```
philipktlin: please run stylecop

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/RuntimeInfoProviderTest.cs
```diff
@@ -332,5 +332,91 @@ public async Task GetModuleLogsTest()
...
+            var provider = await RuntimeInfoProvider.CreateAsync(dockerClient);
+
+            var moduleInfo = await provider.GetModules(CancellationToken.None);
+
+            var actualContainerNames = moduleInfo
+                                            .Select(m => m.Name);
+
+            var expectedContainerNames = containersToReturn
+                                            .Where(c => !c.Equals(missingContainerName))
+                                            .Select(c => c.Substring(1))    // cuts the first letter
```
philipktlin: why needs to cut the first letter?

edge-util/src/Microsoft.Azure.Devices.Edge.Util/Option.cs
```diff
@@ -202,6 +202,25 @@ public Option<T> Filter(Func<T, bool> predicate)
...
+                    yield return item.OrDefault();
+        }
+
+        public static IEnumerable<T> FilterMap<T>(this IEnumerable<Option<T>> source)
+        {
+            Preconditions.CheckNotNull(source, nameof(source));
+
+            foreach (var item in source)
+                if (item.HasValue)
+                    yield return item.OrDefault();
```
philipktlin:
```suggestion
                    yield return item.Value;
```

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/RuntimeInfoProviderTest.cs
vipeller: that is what RuntimeInfoProvider class does in the ExtractModuleRuntimeState() method. It would trim the module name, so I do the same in the test with the passed name strings. I suppose originally it trims a special character returned by docker api.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/Option.cs
vipeller: 'Value' property is not accessible by the extension method, because it is private. The two options are OrDefault() or Expect(), where Expect() needs a function creating an exception when there is no value. As I check and ensure that always there is a value (previous 'if'), I figured that using OrDefault() is safe and seemed more simple.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/RuntimeInfoProviderTest.cs
philipktlin: I figured out container name returned from docker.DotNet NuGet will contains "/" prefix.  Therefore we remove 1st characters.

Therefore I think expectedContainersNames should be the same as containersToReturn; You need to update containerList variable in SetupDockerClient to returns each container name with "/" prefix.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/RuntimeInfoProviderTest.cs
```diff
@@ -332,5 +332,91 @@ public async Task GetModuleLogsTest()
...
+
+            Assert.Equal(expectedContainerNames, actualContainerNames);
+        }
+
+        private IDockerClient SetupDockerClient(string nonExistingContainer, params string[] containers)
+        {
+            var dockerClient = new Mock<IDockerClient>();
+            var dockerContainer = new Mock<IContainerOperations>();
+            var dockerSystem = new Mock<ISystemOperations>();
+            var containerList = containers.Select(c => new ContainerListResponse { ID = c }).ToList();
```
philipktlin:
```suggestion
            var containerList = containers.Select(c => new ContainerListResponse { ID = $"/{s}" }).ToList();
```

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/RuntimeInfoProviderTest.cs
vipeller: ok, now I am adding the prefix in the mocked docker api object, so the test method doesn't need to trim the first letter

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/RuntimeInfoProviderTest.cs
vipeller: adding the trailing '/' a few lines below

PR #1748, opened 2019-09-24
[k8s] Lifecycle manager for kubernetes needs to return all active modules.

darobs:


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleIdentityLifecycleManager.cs
```diff
@@ -44,7 +37,19 @@ public ModuleIdentityLifecycleManager(IIdentityManager identityManager, ModuleId
             }
         }
 
-        async Task<IImmutableDictionary<string, IModuleIdentity>> GetModuleIdentitiesAsync(Diff diff)
+        protected virtual async Task<IImmutableDictionary<string, IModuleIdentity>> GetModuleIdentitiesWorkAsync(ModuleSet desired, ModuleSet current)
```
kkdawkins: We should think about this design better. It's not clear what the goal is unless you read through both implementations. Perhaps create a virtual method that "manages" empty diff sets. Then, for Kube, you can have that be a noop (return true or something).

Just GetModuleIdentitiesWork is confusing because you're clearly abstracting away **one** behavior (empty diff handling) but that's not clear from the design of this virtual method.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/KubernetesModule.cs
```diff
@@ -189,7 +189,8 @@ protected override void Load(ContainerBuilder builder)
             // KubernetesRuntimeInfoProvider
             builder.Register(c => new KubernetesRuntimeInfoProvider(this.deviceNamespace, c.Resolve<IKubernetes>(), c.Resolve<IModuleManager>()))
                 .As<IRuntimeInfoProvider>()
-                .As<IRuntimeInfoSource>();
+                .As<IRuntimeInfoSource>()
+                .SingleInstance();
```
kkdawkins: Did you mean to add SingleInstance?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleIdentityLifecycleManager.cs
kkdawkins: Also, scoping down the virtual behavior to something like empty diff handler will scope down the surface here. I could see someone in the future not realize what's happening and add generic behavior to one but not the other.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/KubernetesModule.cs
kkdawkins: Checking b/c it seems unrelated to PR :)

darobs: Yes, this was a bug discovered testing this.  Can't prove I've fixed the problem without this change.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleIdentityLifecycleManager.cs
darobs: Yes, that was @dmolokanov's suggestion, too.  I hated how it looked - essentially in one implementation, it's returning an empty dictionary, and in the other implementation, it's duplicating the rest of the function.  I'll try it again...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesModuleIdentityLifecycleManager.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    using System;
+    using System.Collections.Immutable;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Agent.Core;
+    using Microsoft.Azure.Devices.Edge.Agent.Edgelet;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public class KubernetesModuleIdentityLifecycleManager : Edgelet.ModuleIdentityLifecycleManager
```
kkdawkins: I wonder if we should do this "correctly" and rename ModuleIdentityLifecycleManger -> BaseIdentityLifeCycleManager and have a DockerModuleIdentityLifecycleManager? Thoughts?

^ Not required for approval but food for thought.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleIdentityLifecycleManager.cs
```diff
@@ -17,6 +17,8 @@ public class ModuleIdentityLifecycleManager : IModuleIdentityLifecycleManager
         readonly ModuleIdentityProviderServiceBuilder identityProviderServiceBuilder;
         readonly Uri workloadUri;
 
+        protected virtual bool GetRequiresIdentities() => false;
```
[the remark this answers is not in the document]
dmolokanov: IsUnchangedIdentitiesRequested?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleIdentityLifecycleManager.cs
```diff
@@ -27,7 +29,7 @@ public ModuleIdentityLifecycleManager(IIdentityManager identityManager, ModuleId
         public async Task<IImmutableDictionary<string, IModuleIdentity>> GetModuleIdentitiesAsync(ModuleSet desired, ModuleSet current)
```
dmolokanov: IMO method that returns empty list when desired and current doesn't have any differences is somewhat unexpected behavior of the method with a name `GetModuleIdentitiesAsync`. Shouldn't we find a root cause of why it was done in this manner?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleIdentityLifecycleManager.cs
```diff
@@ -44,7 +46,7 @@ public ModuleIdentityLifecycleManager(IIdentityManager identityManager, ModuleId
             }
         }
 
-        async Task<IImmutableDictionary<string, IModuleIdentity>> GetModuleIdentitiesAsync(Diff diff)
+        protected async Task<IImmutableDictionary<string, IModuleIdentity>> GetModuleIdentitiesAsync(Diff diff)
```
dmolokanov: Can be `private` again, right?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleIdentityLifecycleManager.cs
darobs: The internal function calls edgelet to list identities every time it is called.  This code was put in so we could shortcut this, because if there's no diff, there will be no changes to modules.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleIdentityLifecycleManager.cs
darobs: Yes it can.

PR #1753, opened 2019-09-25
[k8s] Spec: Kubernetes experimental extensions for createOptions

dmolokanov:
Add spec describing Kubernetes experimental extensions for createOptions:
- volumes
- resources
- nodeSelector

### DIFF
--- a/kubernetes/doc/create-options.md
+++ b/kubernetes/doc/create-options.md
@@ -1,0 +1,144 @@
+# Kubernetes createOptions Extensions
+
+IoT Edge on Kubernetes supports several experimental features to allow a user to assign additional parameters to pods that represents IoT Edge Modules.
+
+## Experimental features
+
+All features described in this document are experimental. To enable them, the following environment variable need to be set for the edgeAgent (make note of the double underscores):
+
+| Environment Variable Name                  | value  |
+|--------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`            | `true` |
+
+## Create Options
+
+We added CreateOptions for experimental features on Kubernetes. These options will be ignores until both `ExperimentalFeatures__Enabled` and `ExperimentalFeatures__Enabled{FeatureName}` are not set to `true`.
+
+```json
+{
+  "k8s-experimental": {
+    "volumes": [{...}],
+    "resources": [{...}],
+    "nodeSelector": {...}
+  }
+}
+```
+
+## Volumes
+
+EdgeAgent allows to mount existing volumes in the namespace from different sources e.g. `ConfigMap` as a pod `Volume`.
+
+### Enabling this feature
+To enable this feature, the following environment variables need to be set for the edgeAgent (make note of the double underscores):
+
+| Environment Variable Name                | value  |
+|------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`          | `true` |
+| `ExperimentalFeatures__K8SEnableVolumes` | `true` |
+
+### Create Options
+
+A `volumes` section of config used to describe how a `ConfigMap` existing in a namespace will be mounted into module container. The value of this config section is an array of pairs `volume` and `volumeMount`. A `volume` part describes how a specified volume source is mounted to `pod` and `volumeMount` describes a mounting of a `volume` within a container. 
+
+`EdgeAgent` doesn't do any translations or interpretations of values but simply assign values to the `pod` and `container` specs. The description of exact structure can be found here for [Volume](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#volume-v1-core) and for [VolumeMount](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#volumemount-v1-core).
+
+```json
+{
+  ...
+  "k8s-experimental": {
+    "volumes": [
+      {
+        "volume": {
+          "name": "ModuleA",
+          "configMap": {
+            "optional": "true",
+            "defaultMode": 420,
+            "items": [{
+                "key": "config-file",
+                "path": "config.yaml",
+                "mode": 420
+            }],
+            "name": "module-config"
+          }
+        },
+        "volumeMount": {
+          "name": "module-config",
+          "mountPath": "/etc/module/config.yaml",
+          "mountPropagation": "None",
+          "readOnly": "true",
+          "subPath": "" 
+        }
+      }
+    ]
+  }
+}
+```
+
+## CPU and Memory Resources
+
+EdgeAgent allows to specify Compute resources for a module container.
+
+### Enabling this feature
+To enable this feature, the following environment variables need to be set for the edgeAgent (make note of the double underscores):
+
+| Environment Variable Name                   | value  |
+|---------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`             | `true` |
+| `ExperimentalFeatures__K8SEnableResources`  | `true` |
+
+### Create Options
+
+A `resources` section of config used to specify compute resources for an IoT Edge Module container. The value of this section is the same description Kubernetes uses for [container spec](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/). 
+
+`EdgeAgent` doesn't do any translations or interpretations of values but simply assign value from module deployment to `resources` parameter of container spec. The description of exact structure can be found [here](hhttps://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#resourcerequirements-v1-core).
+
+```json
+{
+  ...
+  "k8s-experimental": {
+    "resources": {
+      "limits": {
+        "memory": "128Mi",
+        "cpu": "500m"
+      },
+      "requests": {
+        "memory": "64Mi",
+        "cpu": "250m"
+      }
+    }
+  }
+}
+```
+
+## Assigning Modules to Nodes
+
+EdgeAgent allows to constrain an IoT Edge Module to only be able to run on particular node(s), or to prefer to run on particular nodes.
+
+### Enabling this feature
+To enable this feature, the following environment variables need to be set for the edgeAgent (make note of the double underscores):
+
+| Environment Variable Name                | value  |
+|------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`          | `true` |
+| `ExperimentalFeatures__K8SNodeSelector`  | `true` |
+
+### CreateOptions
+
+A `nodeSelector` section of config used for node selection constrain. It specifies a map of key-value pairs that corresponds to a key-value pair of node labels. To know more please refer to a [corresponding docs](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector).
+
+`EdgeAgent` doesn't do any translations or interpretations of values but simply assign value from module deployment to `nodeSelector` parameter of a pod spec.
+
+```json
+{
+  ...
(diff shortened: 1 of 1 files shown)


kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,146 @@
...
+
+All features described in this document are experimental. To enable them, the following environment variable need to be set for the edgeAgent (make note of the double underscores):
+
+| Environment Variable Name                  | value  |
+|--------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`            | `true` |
+
+## Create Options
+
+We added CreateOptions for experimental features on Kubernetes. These options will be ignores until both `ExperimentalFeatures__Enabled` and `ExperimentalFeatures__Enabled{FeatureName}` are not set to `true`.
```
darobs: `ExperimentalFeatures__{FeatureName}` - remove the extra "Enabled"

darobs: Do we need to have each extension enabled?  Does it make more sense to have a "k8sextensions"?

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,146 @@
...
+To enable this feature, the following environment variables need to be set for the edgeAgent (make note of the double underscores):
+
+| Environment Variable Name                | value  |
+|------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`          | `true` |
+| `ExperimentalFeatures__K8SEnableVolumes` | `true` |
+
+### Create Options
+
+A `volumes` section of config used to describe how a `ConfigMap` existing in a namespace will be mounted into module container. The value of this config section is an array of pairs `volume` and `volumeMount`. A `volume` part describes how a specified volume source is mounted to `pod` and `volumeMount` describes a mounting of a `volume` within a container.
```
darobs: Are we only mounting the volume in the primary (module) container?  Should we make that explicit in the document?

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,146 @@
...
+            "defaultMode": 420,
+            "items": [{
+                "key": "config-file",
+                "path": "config.yaml",
+                "mode": 420
+            }],
+            "name": "module-config"
+          }
+        },
+        "volumeMount": {
```
darobs: Do we need to make volumeMounts an array of VolumeMounts?  You can theoretically mount the same volume multiple times.

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,146 @@
...
+| Environment Variable Name                | value  |
+|------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`          | `true` |
+| `ExperimentalFeatures__K8SEnableVolumes` | `true` |
+
+### Create Options
+
+A `volumes` section of config used to describe how a `ConfigMap` existing in a namespace will be mounted into module container. The value of this config section is an array of pairs `volume` and `volumeMount`. A `volume` part describes how a specified volume source is mounted to `pod` and `volumeMount` describes a mounting of a `volume` within a container. 
+
+`EdgeAgent` doesn't do any translations or interpretations of values but simply assign values to the `pod` and `container` specs. The description of exact structure can be found here for [Volume](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#volume-v1-core) and for [VolumeMount](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#volumemount-v1-core).
```
darobs: Was there a particular reason to point to v1.12?

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,146 @@
...
+| Environment Variable Name                   | value  |
+|---------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`             | `true` |
+| `ExperimentalFeatures__K8SEnableResources`  | `true` |
+
+### Create Options
+
+A `resources` section of config used to specify compute resources for an IoT Edge Module container. The value of this section is the same description Kubernetes uses for [container spec](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/). In addition to CPU and Memory EdgeAgent operator can specify [device](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) limits for a module.
+
+`EdgeAgent` doesn't do any translations or interpretations of values but simply assign value from module deployment to `resources` parameter of container spec. The description of exact structure can be found [here](hhttps://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#resourcerequirements-v1-core).
```
darobs: Typo: extra 'h' in https

kubernetes/doc/create-options.md
dmolokanov: That what I thought initially. But @veyalla suggested flag per section instead.

kubernetes/doc/create-options.md
dmolokanov: Do you think it is a real-world scenario?

kubernetes/doc/create-options.md
dmolokanov: A minimal version of k8s that we support. New k8s versions can add additional functionalities that edge may not support yet.

kubernetes/doc/create-options.md
dmolokanov: Good idea

veyalla:
> My biggest question is, should we enable each section as a feature, or simply allow all k8s extensions in one experimental feature?

I think I miscommunicated, agree all extensions behind one experimental feature flag is better UX.

kubernetes/doc/create-options.md
dmolokanov: Made it array

darobs: I don't see it as a common situation, but it's possible to do in the podspec.

kubernetes/doc/create-options.md
darobs: I added a Backlog task to settle on an API version for GA.  We should probably worry about the exact version in this doc at that time.

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,122 @@
...
+All features described in this document are experimental. To enable them, the following environment variable need to be set for the edgeAgent (make note of the double underscores):
+
+| Environment Variable Name                   | value  |
+|---------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`             | `true` |
+| `ExperimentalFeatures__EnableK8SExtensions` | `true` |
+
+## Create Options
+
+We added CreateOptions for experimental features on Kubernetes. These options will be ignores until both `ExperimentalFeatures__Enabled` and `ExperimentalFeatures__Enabled{FeatureName}` are not set to `true`.
```
darobs: typo: ignored, , and I think these options will be ignored until both are set to true.

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,122 @@
...
+}
+```
+
+## Volumes
+
+EdgeAgent allows to mount existing volumes in the namespace from different sources e.g. `ConfigMap` as a pod `Volume`.
+
+### Create Options
+
+A `volumes` section of config used to describe how a `ConfigMap` or any other volume source existing in a namespace will be mounted into module container. The value of this config section is an array of pairs `volume` and `volumeMount`. A `volume` part describes how a specified volume source is mounted to `pod` and `volumeMounts` describes a mounting of a `volume` within a module main container.
```
darobs: Maybe add at the end, 
"See [Volumes](https://kubernetes.io/docs/concepts/storage/volumes/) and [PodSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#podspec-v1-core) reference for more information."

darobs: Here or at the end of the 2nd paragraph....

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,122 @@
+# Kubernetes createOptions Extensions
+
+IoT Edge on Kubernetes supports several experimental features to allow a user to assign additional parameters to pods that represents IoT Edge Modules.
```
massand: >represents [](start = 123, length = 10)

Consider rephrasing this to clarify the pods represent modules, not the parameters.

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,122 @@
...
+All features described in this document are experimental. To enable them, the following environment variable need to be set for the edgeAgent (make note of the double underscores):
+
+| Environment Variable Name                   | value  |
+|---------------------------------------------|--------|
+| `ExperimentalFeatures__Enabled`             | `true` |
+| `ExperimentalFeatures__EnableK8SExtensions` | `true` |
+
+## Create Options
+
+We added CreateOptions for experimental features on Kubernetes. These options will be ignored until both `ExperimentalFeatures__Enabled` and `ExperimentalFeatures__EnableK8SExtensions` are not set to `true`.
```
massand: >ignored [](start = 86, length = 7)

Nit: Please change double negative to something like "options will take effect if both variable names are set to true"

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,122 @@
...
+      }
+    ]
+  }
+}
+```
+
+## CPU, Memory and Device Resources
+
+EdgeAgent allows to specify Compute resources for a module container.
+
```
massand: Worth adding the point about what EdgeAgent does with HostConfig CPU and memory limits. Because the CPU memory limits are pretty much passthrough k8s via limits below

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,122 @@
...
+        "hardware-vendor.example/foo": 2
+      }
+    }
+  }
+}
+```
+
+## Assigning Modules to Nodes
+
+EdgeAgent allows to constrain an IoT Edge Module to only be able to run on particular node(s), or to prefer to run on particular nodes.
```
massand: > or to prefer to run on particular nodes [](start = 94, length = 40)

redundant. can remove ", or to prefer..."

kubernetes/doc/create-options.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Volumes
+
+EdgeAgent allows to mount existing volumes in the namespace from different sources e.g. `ConfigMap` as a pod `Volume`.
+
+### Create Options
+
+A `volumes` section of config used to describe how a `ConfigMap` or any other volume source existing in a namespace will be mounted into module container. The value of this config section is an array of pairs `volume` and `volumeMount`. A `volume` part describes how a specified volume source is mounted to `pod` and `volumeMounts` describes a mounting of a `volume` within a module main container. See [Volumes](https://kubernetes.io/docs/concepts/storage/volumes/) and [PodSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#podspec-v1-core) reference for more information.
+
+`EdgeAgent` doesn't do any translations or interpretations of values but simply assigns values to the `pod` and `container` specs. The description of exact structure can be found here for [Volume](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#volume-v1-core) and for [VolumeMount](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#volumemount-v1-core).
```
massand: >exact structure can be found here for [Volume] [](start = 150, length = 46)

We need to be more specific about the major K8s API and structure version we plan on supporting / e2e testing with this release. Currently, the versions are sprinkled within links.

PR #1754, opened 2019-09-25
Configurable 'Consumer Group' for MessageAnalyzer

vipeller:
Allows injecting 'Consumer Group' name for MessageAnalyzer to use when processing messages.

Usage is: runE2ETest.sh -eventHubConsumerGroup groupname

If not specified, it will inject $Default, which results in the current working, as MessageAnalyzer uses $Default consumer today.

Before specifying any consumer group name, that group name must be created manually in IotHub to make it existing.

### DIFF
--- a/scripts/linux/runE2ETest.sh
+++ b/scripts/linux/runE2ETest.sh
@@ -945,7 +945,7 @@ function usage() {
     echo ' -containerRegistryPassword      Password of given username for container registory.'
     echo ' -iotHubConnectionString         IoT hub connection string for creating edge device.'
     echo ' -eventHubConnectionString       Event hub connection string for receive D2C messages.'
-	echo ' -eventHubConsumerGroup          An existing consumer group id for D2C messages.'
+    echo ' -eventHubConsumerGroup          An existing consumer group id for D2C messages.'
     echo ' -loadGenMessageFrequency        Frequency to send messages in LoadGen module for long haul and stress test. Default is 00.00.01 for long haul and 00:00:00.03 for stress test.'
     echo ' -snitchAlertUrl                 Alert Url pointing to Azure Logic App for email preparation and sending for long haul and stress test.'
     echo ' -snitchBuildNumber              Build number for snitcher docker image for long haul and stress test. Default is 1.1.'
--- a/scripts/linux/runE2ETest.sh
+++ b/scripts/linux/runE2ETest.sh
@@ -170,6 +170,7 @@ function prepare_test_from_artifacts() {
                 local escapedSnitchAlertUrl
                 local escapedBuildId
                 sed -i -e "s@<Analyzer.EventHubConnectionString>@$EVENTHUB_CONNECTION_STRING@g" "$deployment_working_file"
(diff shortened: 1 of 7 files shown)


philipktlin:
you need to update LH and stress deployment json files for Windows as well.

scripts/linux/runE2ETest.sh
```diff
@@ -940,6 +946,7 @@ function usage() {
     echo ' -containerRegistryPassword      Password of given username for container registory.'
     echo ' -iotHubConnectionString         IoT hub connection string for creating edge device.'
     echo ' -eventHubConnectionString       Event hub connection string for receive D2C messages.'
+	echo ' -eventHubConsumerGroup          An existing consumer group id for D2C messages.'
```
philipktlin: nit: spacing.

scripts/linux/runE2ETest.sh
```diff
@@ -347,6 +347,9 @@ function process_args() {
             saveNextArg=0
         elif [ $saveNextArg -eq 28 ]; then
             DPS_MASTER_SYMMETRIC_KEY="$arg"
+            saveNextArg=0
+		elif [ $saveNextArg -eq 29 ]; then
```
philipktlin: nit: spacing; you can run shellcheck to check style of sh script.

scripts/linux/runE2ETest.sh
```diff
@@ -617,6 +621,7 @@ function run_longhaul_test() {
     local device_id="$RELEASE_LABEL-Linux-$image_architecture_label-longhaul"
 
     sed -i -e "s@<Analyzer.DeviceID>@$device_id@g" "$deployment_working_file"
+	sed -i -e "s@<Analyzer.ConsumerGroupId>@${EVENT_HUB_CONSUMER_GROUP:-\$Default}@g" "$deployment_working_file"
```
philipktlin: please move this change to prepare_test_from_artifacts function.

scripts/linux/runE2ETest.sh
```diff
@@ -170,6 +170,7 @@ function prepare_test_from_artifacts() {
                 local escapedSnitchAlertUrl
                 local escapedBuildId
                 sed -i -e "s@<Analyzer.EventHubConnectionString>@$EVENTHUB_CONNECTION_STRING@g" "$deployment_working_file"
+                sed -i -e "s@<Analyzer.ConsumerGroupId>@${EVENT_HUB_CONSUMER_GROUP:-\$Default}@g" "$deployment_working_file"
```
philipktlin: Suggest to move default value assignment below process_args call.

philipktlin:
Please update runE2ETest.ps1 as well; Also, are you going to update stress.yaml and longhaul.yaml to pass different consumer group in this PR or in a separate PR?

vipeller:
I think that should be a separate step (configuring *.yaml). The current change would not cause any differences in behavior, just gives the infrastructure. Then we need to create the appropriate consumer groups in IoTHub and then we can configure it to be used.

vipeller:
merging this now then will do the same with the powershell version in a separate PR to avoid getting obsolate

PR #1759, opened 2019-09-26
Add support for backup and restore of in-memory db stores.

shantanu1singh:
EdgeAgent and EdgeHub currently support an in-memory only store which can be enabled by setting the `UsePersistentStorage` configuration flag to `false` in the module’s configuration.

With this change, this in-memory store will be backed up on graceful shutdown and restored from on the next startup. This will help customers leverage the benefits of an in-memory only store (thus limiting disk writes) and not lose messages when a module is killed or the device restarts.

The feature can be enabled by setting the `EnableNonPersistentStorageBackup` environment variable to `true`. 

A host disk-mapped folder can be mounted into the container to be used for storing backups and the path to it will have to be specified using the `BackupFolder` environment variable with its value equal to the path of the backup folder within the container (as specified in the volume mount).

If a backup folder is not explicitly specified, the backups will be created on the container's file system instead.


edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
```diff
@@ -1,18 +1,100 @@
...
+            if (useBackupAndRestore)
+            {
+                string backupPathValue = this.backupPath.Expect(() => new ArgumentException($"The value of {nameof(backupPath)} needs to be specified if backup and restore is enabled."));
+                Preconditions.CheckNonWhiteSpace(backupPathValue, nameof(backupPath));
+
+                this.RestoreDb(backupPathValue);
+            }
+        }
+
+        private void RestoreDb(string backupPath)
```
varunpuranik: async?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
```diff
@@ -1,18 +1,100 @@
...
+
+            if (!File.Exists(backupMetadataFilePath))
+            {
+                Events.NoBackupsForRestore();
+                return;
+            }
+
+            try
+            {
+                using (StreamReader file = File.OpenText(backupMetadataFilePath))
```
varunpuranik: Nit: feels like we should make it async, or just use ReadAllText (should be small, right?)

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
```diff
@@ -1,18 +1,100 @@
...
+                {
+                    JsonSerializer serializer = new JsonSerializer();
+                    BackupMetadataList backupMetadataList = (BackupMetadataList)serializer.Deserialize(file, typeof(BackupMetadataList));
```
varunpuranik: Why do you have a list of metadata here? Shouldn't it be only one? (per your design doc)

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
```diff
@@ -24,9 +106,225 @@ public void RemoveDbStore(string partitionName)
             this.partitionDbStoreDictionary.TryRemove(partitionName, out IDbStore _);
         }
 
+        public void Close()
+        {
+            this.CloseAsync().Wait();
```
varunpuranik: Don't do Wait().. 
Can you simply not have a sync Close() and only have CloseAsync()

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
```diff
@@ -1,18 +1,100 @@
...
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Threading.Tasks;
     using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Converters;
 
     public class InMemoryDbStoreProvider : IDbStoreProvider
```
varunpuranik: As an overall approach, I think we should have a separate class that takes care of BackupAndRestore - 
Takes in a IDictionary or KeyedCollection and backs it up, or reads from the backup and returns a list of Dictionaries or KeyedCollections. 
These can then be used to populate and backup the DB. 
That will allow us to separate the logic, and test/maintain/replace it more easily. 
Note that we should still use decorator pattern - but it will simply call the Backup and Resotre apis on this interface.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -22,7 +23,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Service
 
     public class Program
     {
-        static readonly TimeSpan ShutdownWaitPeriod = TimeSpan.FromSeconds(20);
+        static readonly TimeSpan ShutdownWaitPeriod = TimeSpan.FromSeconds(60);
```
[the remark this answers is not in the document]
shantanu1singh: We can. It'll only be used when somehow EdgeHub is taking greater than 60 seconds to clean up for some customer. Is that a good enough reason?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -102,8 +103,10 @@ static async Task<int> MainAsync(IConfigurationRoot configuration)
                 await protocolHead.StartAsync();
                 await Task.WhenAny(cts.Token.WhenCanceled(), renewal.Token.WhenCanceled());
                 logger.LogInformation("Stopping the protocol heads...");
-                await Task.WhenAny(protocolHead.CloseAsync(CancellationToken.None), Task.Delay(TimeSpan.FromSeconds(10), CancellationToken.None));
+                await protocolHead.CloseAsync(CancellationToken.None);
```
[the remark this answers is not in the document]
shantanu1singh: Yeah, I removed it because we were already waiting for 60 seconds before shutting the app down in ShutdownHandler so another artificial wait here doesn't make much sense? 

I also wanted the Db store provider to be closed only after the protocol heads are closed so that backup and restore occurs only after the protocol heads are closed to prevent cases where we don't backup some messages.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -102,8 +103,10 @@ static async Task<int> MainAsync(IConfigurationRoot configuration)
...
                 logger.LogInformation("Protocol heads stopped.");
+
+                CloseDbStoreProvider(container);
```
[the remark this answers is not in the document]
shantanu1singh: Not sure what you mean by implicitly? We would need to make a call to Close somewhere and I thought we had decided on calling Close directly after the protocol heads were closed?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStore.cs
```diff
@@ -17,12 +21,51 @@ class InMemoryDbStore : IDbStore
...
 
-        public InMemoryDbStore()
+        public InMemoryDbStore(string dbName)
         {
+            Preconditions.CheckNonWhiteSpace(dbName, nameof(dbName));
+            this.dbName = dbName;
             this.keyValues = new ItemKeyedCollection(new ByteArrayComparer());
         }
 
+        public InMemoryDbStore(string dbName, string backupPath)
```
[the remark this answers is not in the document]
shantanu1singh: Sure, I'll change it

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStore.cs
```diff
@@ -17,12 +21,51 @@ class InMemoryDbStore : IDbStore
...
+            string backupFileName = HttpUtility.UrlEncode(dbName);
+            string dbBackupPath = Path.Combine(backupPath, $"{backupFileName}.bin");
+            if (!File.Exists(dbBackupPath))
+            {
+                throw new IOException($"The backup data for database {dbName} doesn't exist.");
+            }
+
+            try
+            {
+                using (FileStream file = File.OpenRead(dbBackupPath))
```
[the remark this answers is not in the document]
shantanu1singh: All backups are cleaned up after restore in the `RestoreDb` method of `InMemoryDbStoreProvider`

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
shantanu1singh: I have a list here so as to support multiple backups in the future with ease. It wasn't adding a whole lot of bloat to the design so I decided on doing that. The design doc does indicate that we are keeping a list

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
shantanu1singh: Sure, I'll take a stab at it and update the PR

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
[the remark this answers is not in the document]
shantanu1singh: Made it configurable now.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
shantanu1singh: Yep should be small. Changed to ReadAllText.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
shantanu1singh: Done, made all of it async

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStore.cs
```diff
@@ -186,6 +262,8 @@ public ItemKeyedCollection(IEqualityComparer<byte[]> keyEqualityComparer)
                 .Select(i => (i.Key, i.Value))
                 .ToList();
 
+            internal IList<Item> AllItems => this.Items;
```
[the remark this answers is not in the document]
shantanu1singh: Exposing the items through this property so that it can be used for serialization using ProtoBuf when backups are created. Added a comment to state the same.

`ItemList` creates another list that has a tuple type instead of using `Item` directly which makes it tricky to add ProtoBuf annotations to the type.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
```diff
@@ -1,18 +1,100 @@
...
         readonly ConcurrentDictionary<string, IDbStore> partitionDbStoreDictionary = new ConcurrentDictionary<string, IDbStore>();
+        readonly Option<string> backupPath;
+        readonly bool useBackupAndRestore;
+
+        public InMemoryDbStoreProvider()
+            : this(Option.None<string>(), false)
+        {
+        }
+
+        public InMemoryDbStoreProvider(Option<string> backupPath, bool useBackupAndRestore)
```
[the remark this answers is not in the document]
shantanu1singh: Created `InMemoryDbStoreWithBackupRestore` to follow the decorator pattern

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStore.cs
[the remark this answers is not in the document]
shantanu1singh: Created `InMemoryDbStoreWithBackupRestore` to follow the decorator pattern

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
shantanu1singh: Updated the PR using the decorator pattern for DbStoreProvider.

Abstracted the backup and restore of both DBStores and the data specifically using the IDbStoreBackupRestore and IBackupRestore interfaces. This way the individual components can be replaced as needed.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStore.cs
```diff
@@ -17,12 +21,51 @@ class InMemoryDbStore : IDbStore
...
+        {
+            Preconditions.CheckNonWhiteSpace(dbName, nameof(dbName));
+            Preconditions.CheckNonWhiteSpace(backupPath, nameof(backupPath));
+            this.dbName = dbName;
+            this.keyValues = new ItemKeyedCollection(new ByteArrayComparer());
+
+            this.RestoreDb(dbName, backupPath);
+        }
+
+        private void RestoreDb(string dbName, string backupPath)
```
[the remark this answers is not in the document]
shantanu1singh: I have made this async but the ProtoBuf serialization libraries currently don't have async methods that we can leverage.

That being said, we could use Task.Run() to asynchronize the deserialization operation and wait for it later but I wasn't sure how useful that would be. Thoughts?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStore.cs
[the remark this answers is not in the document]
shantanu1singh: Not needed anymore with the new changes on the PR

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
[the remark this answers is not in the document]
shantanu1singh: Using a decorator pattern now for all `IDbStoreProvider` instances in general.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -259,6 +264,19 @@ string GetStoragePath()
...
         }
 
+        string GetStorageBackupPath()
+        {
+            string baseBackupStoragePath = this.configuration.GetValue<string>("backupFolder");
+            if (string.IsNullOrWhiteSpace(baseBackupStoragePath) || !Directory.Exists(baseBackupStoragePath))
+            {
+                baseBackupStoragePath = Path.GetTempPath();
+            }
+
```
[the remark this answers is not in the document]
shantanu1singh: The logger factory isn't registered on the container by the time this code block is executed so it'll be a little tricky to add logging. I'll keep it this way for now given we don't have logging for when the 'Temp' path is used for 'StorageFolder' either

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
```diff
@@ -24,9 +106,225 @@ public void RemoveDbStore(string partitionName)
...
+                catch (IOException exception)
+                {
+                    Events.BackupFailure($"The backup operation failed with error ${exception}.");
+
+                    // Clean up any artifacts of the attempted backup.
+                    CleanupKnownBackups(backupPathValue, backupMetadataList);
+                }
+            }
+        }
+
```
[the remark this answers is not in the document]
shantanu1singh: This code isn't there anymore but I have moved the backup specific code in the `CloseAsync` implementation of `DbStoreProviderWithBackupRestore` to a separate `BackupAsync` method instead

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -102,8 +103,10 @@ static async Task<int> MainAsync(IConfigurationRoot configuration)
                 await protocolHead.StartAsync();
                 await Task.WhenAny(cts.Token.WhenCanceled(), renewal.Token.WhenCanceled());
                 logger.LogInformation("Stopping the protocol heads...");
-                await Task.WhenAny(protocolHead.CloseAsync(CancellationToken.None), Task.Delay(TimeSpan.FromSeconds(10), CancellationToken.None));
```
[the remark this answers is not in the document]
shantanu1singh: Yeah, I removed it because we were already waiting for 60 seconds before shutting the app down in ShutdownHandler so another artificial wait here doesn't make much sense? 

I also wanted the Db store provider to be closed only after the protocol heads are closed so that backup and restore occurs only after the protocol heads are closed to prevent cases where we don't backup some messages

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/DbStoreProvider.cs
```diff
@@ -71,6 +74,11 @@ public void RemoveDbStore(string partitionName)
             }
         }
 
+        public void Close()
+        {
+            // No-op.
```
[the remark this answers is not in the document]
shantanu1singh: The reason behind adding a CloseAsync method is to execute the backup process as part of shutting down gracefully. 

We didn't want to put it in the Dispose() method because that is traditionally used to *just* dispose resources (such as unmanaged handles) instead of executing product code such as backup. Introductions of new bugs that prevent the execution of the Dispose() method of DbStoreProvider(s) are likely as the disposal execution chain might not be executed due to an upstream consumer not calling it explicitly.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
```diff
@@ -0,0 +1,353 @@
...
+                // Clean up any restored state in the dictionary if the backup fails midway.
+                foreach (string store in this.dbStores.ToList())
+                {
+                    this.RemoveDbStore(store);
+                }
+
+                this.RemoveDbStore();
+
+                // Delete all backups as the last backup itself is corrupt.
+                this.CleanupAllBackups(this.backupPath);
```
vipeller: Maybe it is valid to clean up after a non IO/JSonException, too, and the CleanupAllBackups() could go into finaly?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
```diff
@@ -0,0 +1,353 @@
...
+            }
+            catch (Exception exception) when (
+            exception is IOException
+            || exception is JsonException)
+            {
+                this.events.BackupFailure($"The backup operation failed with error ${exception}.");
+
+                // Clean up any artifacts of the attempted backup.
+                this.CleanupKnownBackups(this.backupPath, backupMetadataList);
+            }
```
vipeller: Similar as above, does it matter what sort of exception it is? (even if nothing else can come in theory)

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
shantanu1singh: Yeah, sounds good. I'll change it

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
shantanu1singh: Yeah, sounds good. I'll change it

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
shantanu1singh: Yeah, sounds good. I'll change it

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/AgentModule.cs
```diff
@@ -134,9 +140,22 @@ protected override void Load(ContainerBuilder builder)
...
+                builder.Register(c => new ProtoBufBackupRestore())
+                    .As<IBackupRestore>()
+                    .SingleInstance();
+
+                // Backup and restore of IDbStore.
+                builder.Register(c => new DbStoreBackupRestore(c.Resolve<IBackupRestore>()))
+                    .As<IDbStoreBackupRestore>()
+                    .SingleInstance();
+            }
+
```
vipeller: Does it cause any problem if these get registered with feature off? (don't change the code if not, just asking)

shantanu1singh: Nope, it won't cause any issues. Just that an instance will be created and never be used.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
```diff
@@ -0,0 +1,351 @@
...
+        }
+
+        private void CleanupAllBackups(string backupPath)
+        {
+            DirectoryInfo backupDirInfo = new DirectoryInfo(backupPath);
+            foreach (FileInfo file in backupDirInfo.GetFiles())
+            {
+                file.Delete();
+            }
```
vipeller: maybe catch all exceptions and continue the loop? I can imagine somehow a restricted file or directory gets there accidentally and then the backup files never will be deleted because this loop exist with an exception?

shantanu1singh: Yep, sounds good. We can log when we fail to delete some of the files in there. I'll make that change.

edge-util/test/Microsoft.Azure.Devices.Edge.Storage.Test/DbStoreBackupRestoreTest.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        ~DbStoreBackupRestoreTest()
+        {
+            if (!string.IsNullOrWhiteSpace(this.backupFolder) && Directory.Exists(this.backupFolder))
+            {
+                Directory.Delete(this.backupFolder, true);
+            }
+        }
+
```
vipeller: is creating a backup folder here a forgotten copy/paste code? I don't see it being used below. Also, I don't see the tests in this source super-useful, it doesn't check higher level behavior, but very low-level implementation specific things like call forwarding and parameter checking.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
```diff
@@ -0,0 +1,351 @@
...
+                // Delete all other backups as we've either:
+                // 1. Restored successfully.
+                // 2. Failed during restore which would indicate a bad backup.
+                this.CleanupAllBackups(this.backupPath);
+            }
+        }
+
+        public override IDbStore GetDbStore(string partitionName)
+        {
+            this.dbStores.Add(partitionName);
```
ancaantochi: This should be a concurrent collection

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
```diff
@@ -0,0 +1,351 @@
...
+                        {
+                            dbStore = this.GetDbStore(store);
+                        }
+                        else
+                        {
+                            dbStore = this.GetDbStore();
+                        }
+
+                        await this.dbStoreBackupRestore.RestoreAsync(store, dbStore, latestBackupDirPath);
+                    }
```
ancaantochi: I think it should not pass the dbStore and add items to it in the RestoreAsync, but return the items that were restored and the code that iterates over the items to add them to be moved here

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
```diff
@@ -0,0 +1,351 @@
...
+            if (!File.Exists(backupMetadataFilePath))
+            {
+                this.events.NoBackupsForRestore();
+                return;
+            }
+
+            try
+            {
+                string fileText = File.ReadAllText(backupMetadataFilePath);
+                BackupMetadataList backupMetadataList = JsonConvert.DeserializeObject<BackupMetadataList>(fileText);
```
ancaantochi: Code that is related to IO and deserialization I think it should be moved to dbStoreBackupStore

edge-util/test/Microsoft.Azure.Devices.Edge.Storage.Test/DbStoreBackupRestoreTest.cs
shantanu1singh: Ah yes, looks like I missed removing the code blocks in the constructor and dispose.

Yes, I added those tests to provide code coverage. The code in the target source class itself is minimal and I wanted the unit test to only be localized to testing what only the source class itself is doing.

Tests for other classes cover high-level behavior.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/IBackupRestore.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Storage
+{
+    using System.Threading.Tasks;
+
+    public interface IBackupRestore
```
ancaantochi: >IBackupRestore [](start = 21, length = 14)

I think it should be a distinction in the name for IBackupRestore and IDbBackupRestore. IBackupRestore it seems the it contains the serialization algorithm, maybe add something like that to the name

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/BackupRestoreUtil.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft. All rights reserved.
```
ancaantochi: Is this needed? It seesm that IBackupRestore implementation could have a property for the SerializationFormat

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
shantanu1singh: Changed to ConcurrentDictionary with a single dummy byte as value

edge-util/test/Microsoft.Azure.Devices.Edge.Storage.Test/DbStoreBackupRestoreTest.cs
shantanu1singh: Removed this test class and its target source completely

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/BackupRestoreUtil.cs
shantanu1singh: Good point, changed IBackupRestore or rather IDataBackupRestore (new name) to have a format property that returns this instead.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/IBackupRestore.cs
shantanu1singh: Renamed IBackupRestore to IDataBackupRestore and removed IDbBackupRestore and IDbStoreBackupRestore

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
shantanu1singh: DbStoreBackupRestore has been removed as per our conversation and all that code has been moved to DbStoreProviderWithBackupRestore

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreProviderWithBackupRestore.cs
shantanu1singh: Removed DbStoreBackupRestore and moved it's code to DbStoreProviderWithBackupRestore as per our discussion

PR #1762, opened 2019-09-26
Add ASC support

myagley:


philipktlin:
Do we need to upgrade Microsoft.Azure.Devices.Client in edge-modules\functions\binding\src\Microsoft.Azure.WebJobs.Extensions.EdgeHub\Microsoft.Azure.WebJobs.Extensions.EdgeHub.csproj?

PR #1763, opened 2019-09-26
fix deployment status update

philipktlin:
Update deployment status to success even there is no deployment change.  We got IcM incident that we find when any exception happened during edge agent reconciliation, it will report 500 failed deployment status.   If it is a transient exception (timeout exception when calling to management endpoint to get container list) and there is no change in deployment, the next reconciliation doesn't detect any change and doesn't assign any deployment status.  Therefore it doesn't update deployment status back to 200 OK.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Agent.cs
```diff
@@ -145,6 +145,10 @@ public async Task ReconcileAsync(CancellationToken token)
                                 throw;
                             }
                         }
+                        else
```
arsing: Maybe flip the `if` and `else` around to simplify the condition. `if (plan.IsEmpty) { ... } else { ... }`

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Agent.cs
```diff
@@ -145,6 +145,10 @@ public async Task ReconcileAsync(CancellationToken token)
...
+                        {
+                            status = Option.Some(DeploymentStatus.Success);
```
arsing: Since `status` is now a `Some` in every path, why does it need to be `Option` at all? Maybe have it be `DeploymentStatus.Success` by default and only override it inside the `catch` block?

philipktlin:
rebase problem; will re-create a new PR.

PR #1764, opened 2019-09-26
fix deployment status update

philipktlin:
Update deployment status to success even there is no deployment change. We got IcM incident that we find when any exception happened during edge agent reconciliation, it will report 500 failed deployment status. If it is a transient exception (timeout exception when calling to management endpoint to get container list) and there is no change in deployment, the next reconciliation doesn't detect any change and doesn't assign any deployment status. Therefore it doesn't update deployment status back to 200 OK.


philipktlin:
Will re-create a new PR based on the commit from master.

PR #1765, opened 2019-09-27
Add support for limiting in-memory store space usage.

shantanu1singh:
EdgeHub currently supports an in-memory only message store which can be enabled by setting the `UsePersistentStorage` configuration flag to `false` in the module’s configuration.

With this change, the size of the in-memory store can be limited by specifying a maximum value (in total number of bytes). This will help customers leverage the benefits of an in-memory only store (thus limiting disk writes) while also preventing the in-memory stores from consuming an indefinite amount of memory.

The maximum storage size limit can be specified using a new configuration section `StoreLimits` in the `StoreAndForwardConfiguration` of the `EdgeHub` twin. `StoreLimits` has a property `MaxSizeBytes` that can be used to specify the maximum storage size (in bytes) that the customer would like. If a value is specified in the twin, `EdgeHub` would keep track of the total number of bytes being consumed by the in-memory store and if that limit is exceeded at any point, a `StorageFullException` will be thrown.


edge-util/src/Microsoft.Azure.Devices.Edge.Storage/IStorageSpaceChecker.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Storage
+{
+    using System;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public interface IStorageSpaceChecker
+    {
+        void SetMaxSize(long maxSize, Option<long> checkFrequency);
```
dylanbronson: nit: rename method since the method not only sets maxSize but also checkFrequency

shantanu1singh: Not applicable anymore as we're not setting checkFrequency through this method.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/EdgeHubConfigTest.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Test
     using System.Collections.Generic;
     using System.Collections.ObjectModel;
     using Microsoft.Azure.Devices.Edge.Hub.Core.Config;
+    using Microsoft.Azure.Devices.Edge.Util;
```
shantanu1singh: can be removed

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
```diff
@@ -28,5 +45,19 @@ public void Dispose()
         {
             // No-op
         }
+
+        long GetTotalMemoryUsage()
+        {
+            long dbSizeInBytes = 0;
+            foreach (var inMemoryDbStore in this.partitionDbStoreDictionary)
```
ancaantochi: I think this needs to take a snapshot because the dictionary might be changed while it iterates and this might throw

shantanu1singh: Iterating over a ConcurrentDictionary is thread-safe and won't throw exceptions as per: https://stackoverflow.com/questions/17776422/is-this-the-proper-way-to-iterate-over-concurrentdictionary-in-c-sharp

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/MemoryUsageAwareInMemoryDbStore.cs
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft. All rights reserved.
```
ancaantochi: I would wrap the InMemoryDbStore instead of extending

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/MemorySpaceChecker.cs
```diff
@@ -0,0 +1,135 @@
...
+            this.memoryUsageStatus = MemoryUsageStatus.Unknown;
+        }
+
+        public void SetMaxSizeBytes(Option<long> maxSizeBytes)
+        {
+            this.maxSize = maxSizeBytes.Match(
+                s =>
+                {
+                    this.isEnabled = true;
+                    Events.SetMaxMemorySpaceUsage(s);
```
ancaantochi: This gets set when there is a twin update with a config change, right? I think it should be thread safe

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
```diff
@@ -9,11 +9,28 @@ public class InMemoryDbStoreProvider : IDbStoreProvider
     {
         const string DefaultPartitionName = "$Default";
         readonly ConcurrentDictionary<string, IDbStore> partitionDbStoreDictionary = new ConcurrentDictionary<string, IDbStore>();
+        readonly Option<IStorageSpaceChecker> memoryStorageSpaceChecker;
+
+        public InMemoryDbStoreProvider()
+            : this(Option.None<IStorageSpaceChecker>())
+        {
```
ancaantochi: What's the case when we want to set the checker to None? I see if the limit is not set this means is disabled, isn't that enough?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/MemorySpaceChecker.cs
shantanu1singh: I don't think this needs to be thread-safe given how ConfigUpdater takes locks before calling these methods which should suffice. I can add a comment on the method to indicate that the method isn't supposed to be thread-safe.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/MemoryUsageAwareInMemoryDbStore.cs
shantanu1singh: Sure, will change to a decorator pattern

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/InMemoryDbStoreProvider.cs
shantanu1singh: Yeah, I just wanted to keep the parameterless constructor for the sake of Edge Agent for now. We can remove it later. Edge Agent currently doesn't use any of the checker stuff and just uses the default parameterless constructor

shantanu1singh: Yeah, I kept it only for the sake of Edge Agent where we don't need to create and register an IStorageSpaceChecker that's passed to the InMemoryDbStoreProvider. We can later remove the Option and just use IStorageSpaceChecker directly

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/MemorySpaceChecker.cs
```diff
@@ -0,0 +1,140 @@
...
+    using System;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// Provides functionality to check whether total memory space consumption hasn't exceeded configured limits.
+    /// NOTE: The methods in this class aren't thread safe. The expectation is for callers to employ thread-safe
+    /// practices when using methods of this class.
+    /// </summary>
+    public class MemorySpaceChecker : IStorageSpaceChecker
```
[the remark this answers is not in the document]
shantanu1singh: Sure, I can move stuff around and make that change once Dylan has his PR checked in.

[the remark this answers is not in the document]
shantanu1singh: As we spoke offline, I'll reorganize stuff once Dylan's change also goes in

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/DbStoreDecorator.cs
```diff
@@ -0,0 +1,106 @@
...
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    /// <summary>
+    /// A decorating wrapper over a Db store to provide ease of extending DB store behavior.
+    /// </summary>
+    public abstract class DbStoreDecorator : IDbStore
```
dylanbronson: Should we be using this on our other DB stores as well? (ColumnFamilyDbStore, DiskSpaceAware once it goes in)

shantanu1singh: Yeah, I think you should for DiskSpaceAware

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/EdgeHubFixtureCollection.cs
```diff
@@ -0,0 +1,33 @@
...
+            {
+                // Before creating a new fixture, we need to ensure that other ProtocolHeadFixture
+                // instances are closed. This is because a ProtocolHeadFixture encapsulates a
+                // socket connection over an address and only one usage of this address is allowed
+                // at a time.
+                foreach (ProtocolHeadFixture fixture in Fixtures)
+                {
+                    if (!fixture.IsClosed)
+                    {
+                        fixture.CloseAsync().Wait();
```
varunpuranik: Is there a reason we maintain this list? Why not just have a current Protocol Head?
Also, what if 2 tests ask for this at the same time?

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/EdgeHubFixtureCollection.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.E2E.Test
+{
+    using System.Collections.Generic;
+
+    public static class EdgeHubFixtureCollection
```
varunpuranik: Where was the existing behavior being a problem?

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/EdgeHubFixtureCollection.cs
shantanu1singh: Thinking about it, you're right, we can just close the current ProtocolHeadFixture and create a new one as current instead. I'll try it out.

Yeah, parallel tests won't work. Currently, all integration tests running sequentially.

The reason all these changes were needed to the ProtocolHeadFixture (and the new code added in #1759 ) was to have a way to close the DbStoreProvider which in turn will trigger a backup. To test restores, I had to recreate the EdgeHub app (running in the same test process) so that the in-memory store gets re-created and we can test whether restore worked.

The problem I was facing with simply disposing ProtocolHeadFixture and recreating a new one was that:
1. You can't have multiple ProtocolHeadFixtures due to the following reason - a ProtocolHeadFixture encapsulates a socket connection over an address and only one usage of this address is allowed at a time.
2. If you however try to only recreate a single ProtocolHeadFixture after disposing the existing one, I was getting errors because somehow disposing the fixture also leads to the disposal of the singleton Logger instance (created by Logger.Factory) which causes any attempts at recreation to fail.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/EdgeHubFixtureCollection.cs
shantanu1singh: Just responded on the other comment. The main reason I needed all this code was the backup and restore PR which required creating the EdgeHub app again in the same process.

shantanu1singh: The changes in this specific PR were trying to make things simpler and solve another problem with test sequence failure.

PR #1772, opened 2019-09-30
Iotedge can bundle logs

lfitchett:
This adds the cssbundle command to the iotedge tool. cssbundle will take all of the logs from a specific date and bundle them into 1 zip file. 

Usage: ```iotedge cssbundle --location "~/my_directory" --since "1 day"```


edgelet/iotedge/src/bundle.rs
```diff
@@ -0,0 +1,233 @@
...
+use futures::{Future, Stream};
+use tokio::prelude::*;
+
+use crate::error::{Error, ErrorKind};
+use failure::Fail;
+
+use crate::logs::pull_logs;
+use crate::Command;
+
+use edgelet_core::{LogOptions, LogTail::Num, Module, ModuleRuntime};
```
arsing: Our existing style is to not import individual enum variants like `LogTail::Num`. Instead import `LogTail` and write `LogTail::Num(0)`

edgelet/iotedge/src/bundle.rs
```diff
@@ -0,0 +1,233 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use chrono::{DateTime, Local, NaiveDateTime, Utc};
```
arsing: Order of imports is:

- libstd `use std::`
- 3rd party crates `use chrono::`
- Our crates `use edgelet_core::`
- Current crate `use crate::`
- Super imports `use super::`

Each group is internally sorted alphabetically and separated by an empty line.

edgelet/iotedge/src/bundle.rs
```diff
@@ -0,0 +1,233 @@
...
+        let result = future::result(self.make_state())
+            .and_then(Bundle::write_all_logs)
+            .map(drop)
+            .map(|_| println!("Wrote all logs to file"));
+
+        Box::new(result)
+    }
+}
+
+impl<M> Bundle<M>
```
arsing: Merge this into the first `impl` block for `Bundle<M>`

edgelet/iotedge/src/bundle.rs
```diff
@@ -0,0 +1,233 @@
...
+                        zip_writer: zw,
+                    }
+                })
+            })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use edgelet_core::{MakeModuleRuntime, ModuleRuntimeState};
```
arsing: Same comment about order of imports applies here.

edgelet/iotedge/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-#![deny(rust_2018_idioms, warnings)]
+// #![deny(rust_2018_idioms, warnings)]
```
arsing: Uncomment this.

edgelet/iotedge/src/logs.rs
```diff
@@ -32,27 +32,41 @@ where
...
 }
+
+pub fn pull_logs<M, W>(
+    runtime: &M,
+    id: &str,
+    options: &LogOptions,
+    writer: W,
+) -> impl Future<Item = W, Error = Error> + Send
+where
+    M: 'static + ModuleRuntime + Clone,
```
arsing: Do you need `M` to be `Clone` ? (I know it was `Clone` in the original code, but I don't think it was necessary there either.)

edgelet/iotedge/src/main.rs
```diff
@@ -14,6 +14,10 @@ use failure::{Fail, ResultExt};
 use futures::Future;
 use url::Url;
 
+use chrono::{DateTime, Duration, Local};
+use parse_duration;
```
arsing: You don't need this `use parse_duration;`

edgelet/iotedge/src/main.rs
```diff
@@ -237,6 +241,34 @@ fn run() -> Result<(), Error> {
...
+                .arg(
+                    Arg::with_name("location")
+                        .help("Directory to store bundle")
+                        .long("location")
+                        .takes_value(true)
+                        .value_name("DIRECTORY")
+                        .default_value("."),
+                )
+                .arg(
+                    Arg::with_name("tail")
```
arsing: Keep parameters in alphabetical order, so define `tail` after `since`.

arsing: Why do we want a `tail` option for this command? Shouldn't the `since` date be the only way?

edgelet/iotedge/src/main.rs
```diff
@@ -293,7 +325,7 @@ fn run() -> Result<(), Error> {
                 args.is_present("verbose"),
                 args.is_present("warnings-as-errors"),
             )
-            .and_then(|mut check| check.execute()),
+            .and_then(iotedge::Command::execute),
```
arsing: Just `Command::execute`. There's already a `use iotedge::*;` in scope.

edgelet/iotedge/src/main.rs
```diff
@@ -314,7 +346,7 @@ fn run() -> Result<(), Error> {
                 .unwrap_or_default();
             let since = args
                 .value_of("since")
-                .and_then(|a| a.parse::<i32>().ok())
+                .and_then(|s| parse_since(s).try_into().ok())
```
arsing: This will silently swallow an invalid value and use the default. We should print an error instead.

Same for `tail` above, and for `tail` and `since` in the new subcommand below.

edgelet/iotedge/src/main.rs
```diff
@@ -323,6 +355,76 @@ fn run() -> Result<(), Error> {
...
+    if let Ok(datetime) = DateTime::parse_from_rfc3339(since) {
+        datetime.timestamp()
+    } else if let Ok(epoch) = since.parse() {
+        epoch
+    } else if let Ok(duration) = parse_duration::parse(since) {
+        (Local::now() - Duration::nanoseconds(duration.as_nanos().try_into().unwrap_or_default()))
+            .timestamp()
+    } else {
+        /* Default to 1 day ago */
+        println!("Flag since ({}) not recognized. Defaulting to 1 day", since);
```
arsing: This should return a fatal error instead of defaulting.

edgelet/iotedge/src/main.rs
```diff
@@ -323,6 +355,76 @@ fn run() -> Result<(), Error> {
...
+                .with_follow(false)
+                .with_tail(tail)
+                .with_since(since);
+            tokio_runtime.block_on(Bundle::new(options, location, runtime()?).execute())
+        }
         (command, _) => tokio_runtime.block_on(Unknown::new(command.to_string()).execute()),
     }
 }
+
+fn parse_since(since: &str) -> i64 {
```
arsing: Since the underlying API takes `i32`, this should perform the fallible conversion to `i32` and return that, instead of `i64`

edgelet/iotedge/Cargo.toml
```diff
@@ -10,7 +10,7 @@ edition = "2018"
 [dependencies]
 atty = "0.2"
 bytes = "0.4"
-chrono = { version = "0.4", features = ["serde"] }
+chrono = { version = "0.4.7", features = ["serde"] }
```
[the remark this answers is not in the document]
lfitchett: I don't, removed

edgelet/iotedge/src/bundle.rs
```diff
@@ -0,0 +1,233 @@
...
+
+use crate::error::{Error, ErrorKind};
+use failure::Fail;
+
+use crate::logs::pull_logs;
+use crate::Command;
+
+use edgelet_core::{LogOptions, LogTail::Num, Module, ModuleRuntime};
+
+pub struct Bundle<M> {
```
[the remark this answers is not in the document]
lfitchett: decided on support-bundle

edgelet/iotedge/src/logs.rs
lfitchett: nope, removed

edgelet/iotedge/src/main.rs
lfitchett: yea

lfitchett: yea

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -0,0 +1,230 @@
...
+            "Writing all logs {}since {} (local time {}) to {}",
+            max_lines, since_time, since_local, self.location
+        );
+
+        /* Make state */
+        let file_options =
+            zip::write::FileOptions::default().compression_method(zip::CompressionMethod::Deflated);
+
+        let zip_writer = zip::ZipWriter::new(
+            File::create(format!("{}/css_bundle.zip", self.location.to_owned()))
```
myagley: Let's not include the `css` in this file name. What about `iotedge_logs.zip` or `iotedge_bundle.zip` or `bundle.zip`?

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -0,0 +1,230 @@
...
+                .map_err(|err| Error::from(err.context(ErrorKind::WriteToStdout)))?,
```
myagley: We should probably add a different `ErrorKind` than `WriteToStdout`. This is going to possibly print an error message that doesn't make sense because this is not writing to stdout.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -0,0 +1,230 @@
...
+        } = state;
+
+        let file_name = format!("logs/{}_log.txt", module_name);
+        zip_writer
+            .start_file_from_path(&Path::new(&file_name), file_options)
+            .into_future()
+            .map_err(|err| Error::from(err.context(ErrorKind::WriteToStdout)))
+            .and_then(move |_| {
+                pull_logs(&runtime, &module_name, &log_options, zip_writer).map(move |zw| {
+                    println!("Wrote {} to file", module_name);
```
myagley: We are probably going to need to do a quick "strings" review on this for all of the text written to stdout. We'll have to loop in whoever is PM'ing this.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -0,0 +1,234 @@
...
+use std::path::Path;
+
+use chrono::{DateTime, Local, NaiveDateTime, Utc};
+use failure::Fail;
+use futures::{Future, Stream};
+use tokio::prelude::*;
+use zip;
+
+use crate::error::{Error, ErrorKind};
+use edgelet_core::{LogOptions, LogTail, Module, ModuleRuntime};
```
arsing: Imports are in the wrong order.

edgelet/iotedge/src/main.rs
```diff
@@ -14,6 +15,8 @@ use failure::{Fail, ResultExt};
 use futures::Future;
 use url::Url;
 
+use chrono::{DateTime, Duration, Local};
```
arsing: Imports are in the wrong order.

edgelet/iotedge/src/main.rs
```diff
@@ -237,6 +240,26 @@ fn run() -> Result<(), Error> {
                 ),
         )
         .subcommand(SubCommand::with_name("version").about("Show the version information"))
+        .subcommand(
+            SubCommand::with_name("support-bundle")
+                .about("Bundles troubleshooting information")
+                .arg(
+                    Arg::with_name("location")
```
arsing: Shouldn't it be a filename rather than a directory? The command only produces one archive file, right?

edgelet/iotedge/src/main.rs
```diff
@@ -323,6 +351,96 @@ fn run() -> Result<(), Error> {
             tokio_runtime.block_on(Logs::new(id, options, runtime()?).execute())
         }
         ("version", _) => tokio_runtime.block_on(Version::new().execute()),
+        ("support-bundle", Some(args)) => {
+            let location = args.value_of("location").unwrap_or_default();
```
arsing: We'd talked about using `expect` like the `check` subcommand does, since the args have a default value so `value_of` won't return `None`. `unwrap_or_default` is not what you want - it's empty string in this case, 0 for the integer parameters, neither of which make sense given the way they're used.

arsing: File / directory paths should use `value_of_os`, not `value_of`

edgelet/iotedge/src/main.rs
```diff
@@ -323,6 +351,96 @@ fn run() -> Result<(), Error> {
...
+    }
+
+    #[test]
+    fn parse_unix() {
+        assert_eq!(parse_since("1569600000").unwrap(), 1_569_600_000);
+    }
+
+    #[test]
+    fn parse_default() {
+        assert!(parse_since("asdfasdf").is_err());
```
arsing: `let _ = parse_since("asdfasdf").unwrap_err();` will give you a better panic message like `called unwrap_err on an Ok value <>` than `expected true, got false`.

THIRDPARTYNOTICES
```diff
@@ -5503,3 +5503,51 @@ Permission is hereby granted, free of charge, to any person obtaining a copy of
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
arsing: Missing newline at end of file.

edgelet/Cargo.lock
```diff
@@ -1,5 +1,10 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
+[[package]]
```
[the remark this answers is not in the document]
arsing: Not yet resolved. You need to add all the new deps that have been added to `Cargo.lock` to the `THIRDPARTYNOTICES` file if they're not already there (and most of them aren't).

edgelet/iotedge/src/main.rs
```diff
@@ -237,6 +239,33 @@ fn run() -> Result<(), Error> {
...
+        .subcommand(
+            SubCommand::with_name("support-bundle")
+                .about("Bundles troubleshooting information")
+                .arg(
+                    Arg::with_name("output")
+                        .help("output file")
+                        .long("output")
+                        .short("o")
+                        .takes_value(true)
+                        .value_name("DIRECTORY")
```
arsing: `.value_name("FILENAME")`

edgelet/iotedge/src/main.rs
```diff
@@ -237,6 +239,33 @@ fn run() -> Result<(), Error> {
                 ),
         )
         .subcommand(SubCommand::with_name("version").about("Show the version information"))
+        .subcommand(
+            SubCommand::with_name("support-bundle")
+                .about("Bundles troubleshooting information")
+                .arg(
+                    Arg::with_name("output")
+                        .help("output file")
```
arsing: `.help("Path of output file")`

edgelet/iotedge/src/main.rs
```diff
@@ -237,6 +239,33 @@ fn run() -> Result<(), Error> {
...
+                    Arg::with_name("since")
+                        .help("Only return logs since this time, as a duration (1d, 90m, 2h30m), rfc3339 timestamp, or UNIX timestamp")
+                        .long("since")
+                        .takes_value(true)
+                        .value_name("DURATION or TIMESTAMP")
+                        .default_value("1 day"),
+                )
+                .arg(
+                    Arg::with_name("include-ms-only")
+                        .help("Follow output log")
```
arsing: `.help("Only include logs from Microsoft-owned Edge modules")`

edgelet/iotedge/src/main.rs
```diff
@@ -310,19 +339,117 @@ fn run() -> Result<(), Error> {
...
+                .transpose()?
+                .expect("arg has a default value");
             let options = LogOptions::new()
                 .with_follow(follow)
                 .with_tail(tail)
                 .with_since(since);
             tokio_runtime.block_on(Logs::new(id, options, runtime()?).execute())
         }
         ("version", _) => tokio_runtime.block_on(Version::new().execute()),
+        ("support-bundle", Some(args)) => {
```
arsing: Keep the arms alphabetically sorted. So this goes above `"version"`

edgelet/iotedge/src/support_bundle.rs
arsing: Not yet resolved.

```rust
use tokio::prelude::*;
use zip;

use edgelet_core:: ...

use crate:: ...
```

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -0,0 +1,290 @@
...
+    fn write_all_logs(s1: BundleState<M>) -> impl Future<Item = BundleState<M>, Error = Error> {
+        SupportBundle::get_modules(s1).and_then(|(names, s2)| {
+            stream::iter_ok(names).fold(s2, SupportBundle::write_log_to_file)
+        })
+    }
+
+    fn get_modules(
+        state: BundleState<M>,
+    ) -> impl Future<Item = (Vec<String>, BundleState<M>), Error = Error> {
+        let ms_modules = &["edgeHub", "edgeAgent"];
```
arsing:
```rust
const MS_MODULES: &[&str] = &["edgeAgent", "edgeHub"];
```

edgelet/iotedge/src/main.rs
```diff
@@ -261,11 +260,12 @@ fn run() -> Result<(), Error> {
                 )
                 .arg(
                     Arg::with_name("include-ms-only")
```
veyalla: We should call this "include-edge-runtime-only" and change the corresponding help text as well. Also suggest have a `.short("e")` alias.

MS owned modules can mean other first party modules like streamanalytics, blobstorage etc.

edgelet/iotedge/src/main.rs
```diff
@@ -310,19 +340,117 @@ fn run() -> Result<(), Error> {
...
+        assert_eq!(
+            parse_since("2019-09-27T16:00:00+00:00").unwrap(),
+            1_569_600_000
+        );
+    }
+
+    #[test]
+    fn parse_english() {
+        assert_near(
+            parse_since("1 hour").unwrap(),
```
myagley: Does the customer need to put this duration in quotes on the command line or is clap smart enough to parse it?

lfitchett: if there's spaces they will, but 1d or 2d6h works as well

PR #1006, opened 2019-03-26
EdgeHub: Upstream perf improvements

varunpuranik:
This PR includes the following -
- Prefetching messages from store for processing (should benefit all messaging endpoints)
- Batching messages while sending upstream
- Processing upstream messages for each client in parallel
- Increasing batch size to process based on endpoint fan-out factor

With these changes, a perf gain of 6-8x was observed in testing.

PR #1781, opened 2019-09-30
Revert "EdgeHub: Upstream perf improvements (#1006)"

richma-ms:
This reverts commit 864b33d6a038596aa6656a58f2f3d28ae4358cc4 for 1.0.8 release.


arsing:
@richma-ms I believe we don't need this any more, right?

PR #1783, opened 2019-10-01
[k8s] Service mapper generates incorrect name for exposed ports

dmolokanov:
Refactored KubernetesServiceMapper
Fixed bug with incorrect ServicePort name value. Should be in alphanumeric with lower case.


edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesServiceMapperTest.cs
```diff
@@ -92,47 +96,173 @@ public void UnknownProtocolDoesNotCreateService()
...
+
+            var service = mapper.CreateService(moduleId, module, moduleLabels).OrDefault();
+
+            Assert.Equal(1, service.Spec.Ports.Count);
+            AssertPort(new V1ServicePort(10, "hostport-10-tcp", null, "TCP", 10), service.Spec.Ports.First());
+            Assert.Equal("ClusterIP", service.Spec.Type);
+        }
+
+        [Fact]
+        public void ExposingPortsCreatesAServiceWithPorts()
```
massand: Nit on ordering this test before the PortBindingOverrideExposedPort for readability - since it appears that test is a prerequisite for this test :)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/service/PortExtensions.cs
```diff
@@ -2,56 +2,50 @@
...
+
+        static IEnumerable<V1ServicePort> ExtractHostPorts(string name, IList<PortBinding> bindings)
+            =>
+                PortAndProtocol.Parse(name)
+                    .Map(
+                        portAndProtocol =>
+                            bindings.Select(
+                                    hostBinding =>
+                                    {
+                                        if (int.TryParse(hostBinding.HostPort, out int hostPort))
```
massand: >HostPort [](start = 69, length = 8)

Apparently the TryValidateProtocol was implicitly defaulting to "TCP. Does this behavior still hold true? If yes, add a UT for this behavior?

dmolokanov: Behavior is not changed. It was just a bit confusing before. It use to be `bool TryValidateProtocol(key, out protocol)` method that returned false if user specified unsupported protocol. In this case TCP was set to TCP, but return value is `false`. 
Code that used TryValidateProtocol checked return value and didn't apply protocol value.

Now intention declare more explicitly if protocol is not supported it returns `None`

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesServiceMapperTest.cs
dmolokanov: Sure. Will put this change in another PR

PR #1785, opened 2019-10-01
Direct method calls to support multibyte utf8 encoded characters in payload

vipeller:
The HTTP protocol head previously was extended to send back Content Length in HTTP header. The length was calculated from the length of the .NET string of the json payload, which gave incorrect value for strings containing characters with multibyte utf8 coding.

Although that is not clear why the Content Length is being sent back to the client, removing it could break some existing users relying on it. On the other hand, we cannot guarantee that the MVC framework will use the exact same serialization when it writes the response stream than we do when we calculate the content length. 

To ensure this consistency, I modified to code to write the response stream directly (vs let MVC serialize)

Note, that during testing, it turned out that the error path was not implemented correctly, and incorrect payload was sent back to the client. Talking to the platform team they gave me the payload format for errors. I've put a basic implementation in place, but that might not be adequate for the client SDK-s to correctly interpret errors.

I need to get in sync with the SDK teams about what is the minimum content they need from the payload - not all properties have corresponding information in edge. This will be another commit in the future.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Http/MethodResult.cs
```diff
@@ -13,23 +14,31 @@ public MethodResult(int status, JRaw payload)
...
         [JsonProperty("status")]
-        public int Status { get; }
+        public virtual int Status { get; }
 
         [JsonProperty("payload")]
-        public JRaw Payload { get; }
+        public virtual JRaw Payload { get; }
     }
 
     public class MethodErrorResult : MethodResult
```
philipktlin: would you explain why we need to hide Status and Payload in json for Error case? What is the difference between Status (int) and StatusCode (HttpStatusCode)?  should we just implement MethodErrorResult without inheriting from MethodResult?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/TwinsController.cs
```diff
@@ -16,6 +16,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Http.Controllers
 
     public class TwinsController : Controller
     {
+        static readonly string supportedContentType = "application/json; charset=utf-8";
```
philipktlin: nit: is the spacing after semi-colon is required?

philipktlin:
would you please remind me, is there any bug reported from our customers, waiting for this fix?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Http/MethodResult.cs
philipktlin: is it because MethodResult is used for return type?  if so, should we have a base class MethodResult without any property; then we have 2 classes inherited from it - MethodErrorResult and MethodSuccessResult?  What do you think?

arsing:
@philipktlin https://github.com/Azure/iotedge/pull/1672

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Http/MethodResult.cs
vipeller: status vs status code: in normal case, the http response has its usual http status code (e.g. 200-ok). the returned http content, which is a json structure, also has a status property, which needs to be carried.
The suggestion about the inheritance (create an empty base class and two derived for the two case) makes sense.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/TwinsController.cs
vipeller: the white space is optional according to RFC7231. Most of the examples contain space thought (I copied this one from a google result)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/TwinsController.cs
```diff
@@ -16,6 +15,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Http.Controllers
...
```
varunpuranik: Nit: PascalCase for static readonly fields I believe

PR #1786, opened 2019-10-01
Fix Twin property 4K length validation

philipktlin:
Update Twin property value validation to apply 4K length limit on JValue type only.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
```diff
@@ -265,21 +265,26 @@ static void ValidateTwinCollectionSize(TwinCollection collection)
...
             {
                 ValidatePropertyNameAndLength(kvp.Name);
 
                 ValidateValueType(kvp.Name, kvp.Value);
 
-                string s = kvp.Value.ToString();
-                ValidatePropertyValueLength(kvp.Name, s);
-
-                if ((kvp.Value is JValue) && (kvp.Value.Type is JTokenType.Integer))
+                if (kvp.Value is JValue)
```
myagley: What happens here if `kvp.Value` is null? Does it evaluate to false here?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
```diff
@@ -265,21 +265,26 @@ static void ValidateTwinCollectionSize(TwinCollection collection)
...
                 {
-                    ValidateIntegerValue(kvp.Name, (long)kvp.Value);
+                    if (kvp.Value.Type is JTokenType.Integer)
```
myagley: How many different types are their? Is it only integer and string? Where do we reject arrays?

myagley: Arrays get rejected at line 272.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/TwinManagerTest.cs
```diff
@@ -1511,39 +1511,58 @@ public async void ConnectionReestablishedDoesNotSyncReportedPropertiesWhenEmptyS
         }
 
         [Fact]
-        public void ValidateTwinPropertiesSuccess()
+        public void ValidateTwinPropertiesWithLongName()
```
myagley: Do we have a test case that models the reported issue?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
philipktlin: I added a test case with null value; when validating, (kvp.Value is JValue) return true; and kvp.Value.ToString() will return empty string.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
philipktlin: For types other than integer, it will be treated as string (that what we have previously).  You can find all supported types from JsonEx.ValidDiffTypes.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/TwinManagerTest.cs
philipktlin: yes, it is covered by ValidateTwinPropertiesWithLongValue; and the assertion checks the message is complaint to the expected property, not from the json root.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
myagley: I guess I'm worried about a null ref exception here. If this evaluates to true will the `if (kvp.Value.Type is JTokenType.Integer)` throw an NRE?

philipktlin: this is JTokenType for your reference.

  public enum JTokenType
  {
    /// <summary>No token type has been set.</summary>
    None,
    /// <summary>A JSON object.</summary>
    Object,
    /// <summary>A JSON array.</summary>
    Array,
    /// <summary>A JSON constructor.</summary>
    Constructor,
    /// <summary>A JSON object property.</summary>
    Property,
    /// <summary>A comment.</summary>
    Comment,
    /// <summary>An integer value.</summary>
    Integer,
    /// <summary>A float value.</summary>
    Float,
    /// <summary>A string value.</summary>
    String,
    /// <summary>A boolean value.</summary>
    Boolean,
    /// <summary>A null value.</summary>
    Null,
    /// <summary>An undefined value.</summary>
    Undefined,
    /// <summary>A date value.</summary>
    Date,
    /// <summary>A raw JSON value.</summary>
    Raw,
    /// <summary>A collection of bytes value.</summary>
    Bytes,
    /// <summary>A Guid value.</summary>
    Guid,
    /// <summary>A Uri value.</summary>
    Uri,
    /// <summary>A TimeSpan value.</summary>
    TimeSpan,
  }

philipktlin: It will return "JTokenType.Null"; therefore no NRE will be thrown.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
```diff
@@ -103,6 +103,36 @@ public async Task UpdateReportedPropertiesAsync(string id, IMessage reportedProp
 
         internal static void ValidateTwinProperties(JToken properties) => ValidateTwinProperties(properties, 1);
 
+        // TODO: Move to a Twin helper class (along with Twin manager update).
```
varunpuranik: Nit: why was this method moved? VS?

philipktlin: it moves up as static method has higher order than class method.  Resharper formatting rule.

philipktlin: As I changed this file, I just ran format code by Resharper, which move this method.

PR #1792, opened 2019-10-02
Allow the user to specify auto generated certificate expiry

kkdawkins:
* Specify Auto-Generated expiry
* Default to 90 days if not specified


edgelet/edgelet-core/src/settings.rs
```diff
@@ -442,6 +442,13 @@ impl Listen {
 
 #[derive(Clone, Debug, serde_derive::Deserialize, serde_derive::Serialize)]
 pub struct Certificates {
+    #[serde(flatten)]
+    device_cert: Option<DeviceCertificate>,
+    auto_generated_ca_lifetime: u16,
```
ttll13: Should this be u64? But then below you multiply this by 86,400... so you should check overflow.

kkdawkins: Discussed offline and we decided that u16 is fine since the input from the config.yaml is in days.

edgelet/contrib/config/linux/config.yaml
```diff
@@ -150,6 +150,7 @@ provisioning:
 #   device_ca_cert: "<ADD URI TO DEVICE CA CERTIFICATE HERE>"
 #   device_ca_pk: "<ADD URI TO DEVICE CA PRIVATE KEY HERE>"
 #   trusted_ca_certs: "<ADD URI TO TRUSTED CA CERTIFICATES HERE>"
+#   auto_generated_ca_lifetime_days: "<Default is 90 days>"
```
arsing: `"<Default is 90 days>"` implies the value is a string `"90"` or even `"90 days"`, rather than a number `90`.

But anyway, it's probably better to mention it in the comment block at lines 134-141:

```
#     auto_generated_ca_lifetime_days - The lifetime of the auto-generated workload CA certificate.
                                        If device_ca_cert and device_ca_pk have not been set,
                                        then this also applies to the auto-generated device CA certificate.
                                        Defaults to 90 days.
```

and change this one to `# auto_generated_ca_lifetime_days: <value>`

Same for `windows/config.yaml`

edgelet/edgelet-docker/config/unix/default.yaml
```diff
@@ -24,3 +24,6 @@ homedir: "/var/lib/iotedge"
 moby_runtime:
   uri: "unix:///var/run/docker.sock"
   network: "azure-iot-edge"
+
+certificates:
+  auto_generated_ca_lifetime_days: 90
```
arsing: Missing newline at end of file.

edgelet/edgelet-docker/config/windows/default.yaml
```diff
@@ -24,3 +24,6 @@ homedir: "C:\\ProgramData\\iotedge"
 moby_runtime:
   uri: "npipe://./pipe/iotedge_moby_engine"
   network: "nat"
+
+certificates:
+  auto_generated_ca_lifetime_days: 90
```
arsing: Missing newline at end of file.

edgelet/edgelet-core/src/settings.rs
```diff
@@ -634,8 +652,15 @@ where
...
     }
 
-    fn certificates(&self) -> Option<&Certificates> {
-        self.certificates.as_ref()
+    // Certificates is left as an option for backward compat
+    fn certificates(&self) -> &Certificates {
+        match &self.certificates {
+            None => &Certificates {
+                device_cert: None,
+                auto_generated_ca_lifetime_days: 90,
```
arsing: Make this a `pub const DEFAULT_AUTO_GENERATED_CA_LIFETIME_DAYS: u16 = 90;`

edgelet/iotedged/src/lib.rs
```diff
@@ -2237,6 +2245,30 @@ mod tests {
...
     }
 
+    #[test]
+    fn settings_without_cert_life_uses_default() {
+        let _guard = LOCK.lock().unwrap();
+
+        let settings = Settings::new(Path::new(GOOD_SETTINGS1)).unwrap();
+        // Default is 90 days so check for that in seconds
+        assert_eq!(
+            7_776_000,
```
arsing: `DEFAULT_AUTO_GENERATED_CA_LIFETIME_DAYS * 86_400` (and then you won't need the comment).

arsing:
@kkdawkins Please also bump `AZURE_IOT_HSM_VERSION` in `hsm_client_data.h` too, so that we don't forget to do it when we branch the release.

PR #1795, opened 2019-10-02
Fix deployment status

philipktlin:
* Update deployment status to success if there is no deployment change
* use DeploymentStatus type instead of option<DeploymentStatus> as deployment status should be set in all code paths.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Agent.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Agent.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Core
 
         public async Task ReconcileAsync(CancellationToken token)
         {
-            Option<DeploymentStatus> status = Option.None<DeploymentStatus>();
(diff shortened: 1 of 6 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Agent.cs
```diff
@@ -148,30 +152,30 @@ public async Task ReconcileAsync(CancellationToken token)
                     }
                     else
                     {
-                        status = Option.Some(DeploymentStatus.Success);
+                        status = DeploymentStatus.Success;
```
darobs: Should we invert the logic here, so this is closer to the reason why it's Success?  Plus I think it looks better:
```
if (deploymentConfig == DeploymentConfig.Empty)
{
    status = DeploymentStatus.Success;
}
else
{
// [...]
    if (plan.IsEmpty)
    {
        status = DeploymentStatus.Success;
    }
```

philipktlin: will do it in master.

PR #1806, opened 2019-10-03
[k8s] CRD separation and helm chart version bump

massand:
This change removes the edge-kubernetes-crd requirement from edge-kubernetes helm chart. This allows flexibility in redeploying edge-kubernetes without manually cleaning up the CRD, which happened before when edge-kubernetes-crd was a requirement for edge-kubernetes. 
Also, version bumped helm chart. 

How built: 
pushd kubernetes/charts
mkdir dist && pushd dist
helm serve &
helm package ../edge-kubernetes ../edge-kubernetes-crd/

helm install --name k8s-edge1 local/edge-kubernetes-crd --namespace=ms-iot --debug
helm install --name k8s-edge2 local/edge-kubernetes --namespace=ms-iot --debug -f override.yaml


kubernetes/charts/edge-kubernetes/values.yaml
```diff
@@ -85,9 +85,13 @@ edgeAgent:
 
 # Optional registry credentials for module images
 # registryCredentials:
-#   <SERVER ADDRESS>:
-#     username: '<USERNAME>'
-#     password: '<PASSWORD>'
+#   serveraddress: <SERVER ADDRESS>:
```
dmolokanov: This change seems suspicious.
use to be
```
registryCredentials: 
  mcr.io: 
    username: user
    password: pass
```
now
```
registryCredentials: 
  serveraddress: mcr.io
  username: user
  password: pass
```

kubernetes/charts/edge-kubernetes/requirements.yaml
```diff
@@ -1,4 +1,5 @@
 dependencies:
   - name: edge-kubernetes-crd
-    version: 0.1.6
-    repository: "file://../edge-kubernetes-crd" #TODO change to real repository once uploaded
\ No newline at end of file
+    version: 0.1.7
+    repository: "https://edgek8s.blob.core.windows.net/helm/" #TODO change to real repository once uploaded
+    condition: edge-kubernetes-crd.enabled
```
dmolokanov: neat!

kubernetes/charts/edge-kubernetes-crd/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for installing CRD for Azure IoT Edge on Kubernetes
 name: edge-kubernetes-crd
-version: 0.1.6
\ No newline at end of file
+version: 0.1.7
```
dmolokanov: should we sync helm chart version with iotedge version?

massand: Let us task this for GA, since we need to figure out the right conventions for Edge appVersion vs helm chart version. And we also need to tackle user expectations about deploying multiple releases in the same namespace (which isn't possible yet I believe).

kubernetes/charts/edge-kubernetes/values.yaml
massand: You are right! Thanks for pointing out that we are trying to consolidate our registry creds. I will revert this now since we already have a task to tackle reg creds for iotedged-proxy as well.

PR #1808, opened 2019-10-04
Introducing E2E for X509 DPS for an Edge Device

yophilav:
Introducing E2E for X509 DPS for an Edge Device on a new end-to-end test framework. 
New Test function: DpsX509 in ` /home/bear/Desktop/iotedge/test/Microsoft.Azure.Devices.Edge.Test/Provisioning.cs `
Original Test: `run_dps_provisioning_test "X509"` in ` runE2ETest.sh `


test/Microsoft.Azure.Devices.Edge.Test.Common/certs/Certificates.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Test.Common.Certs
+{
+    public class Certificates
```
damonbarry: Previously we had "edge certificates" and "leaf certificates". Edge certs were really CA certs, and leaf certs were really "identity" (aka end-entity) certs. With your new test, this distinction isn't enough because now we have edge CA certs and edge identity certs (as well as leaf identity certs). Maybe the names for the two cert classes should be `CaCertificate` and `IdentityCertificate`.

`CertificateAuthority` would return these certificate objects from `GenerateCaCertificate()` and `GenerateIdentityCertificate()` respectively.

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/Certificates.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        private string[] GetFileLocation(string deviceId)
+        {
+            return new[]
+            {
+                $"certs/iot-device-{deviceId}-full-chain.cert.pem",
+                $"private/iot-device-{deviceId}.key.pem"
+            };
```
damonbarry: The code today consolidates all hard-coded certificate paths into one file: Common/OsPlatform.cs. It would be nice to preserve that. In this PR, there are hard-coded cert paths in at least three files.

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/EdgeCertificates.cs
```diff
@@ -20,7 +16,28 @@ public class EdgeCertificates
                 new X509Certificate2(X509Certificate.CreateFromCertFile(this.TrustedCertificatesPath))
             };
 
+        private string[] GetEdgeCertFileLocation(string deviceId)
```
damonbarry: See my comment in `Certificates.GetFileLocation()`...

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/Certificates.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public string CertificatePath { get; protected set; }
+
+        public string KeyPath { get; protected set; }
+
+        protected Certificates()
```
damonbarry: No need for an empty constructor...

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/EdgeCertificates.cs
```diff
@@ -20,7 +16,28 @@ public class EdgeCertificates
...
+        {
+            var location = this.GetEdgeCertFileLocation(deviceId);
+            var files = OsPlatform.NormalizeFiles(location, scriptPath);
+            this.CertificatePath = files[0];
+            this.KeyPath = files[1];
+            this.TrustedCertificatesPath = files[2];
+        }
+
         public EdgeCertificates(string certificatePath, string keyPath, string trustedCertsPath)
+            : base()
```
damonbarry: `certificatePath` and `keyPath` should be passed to the base constructor and set there. That way you don't need to mark the base class properties as `protected set;`.

test/Microsoft.Azure.Devices.Edge.Test/Provisioning.cs
```diff
@@ -67,5 +74,61 @@ public async Task DpsSymmetricKey()
...
+                        Context.Current.CaCertScriptPath.Expect(() => new InvalidOperationException("Missing CA cert script path"));
+            string idScope = Context.Current.DpsIdScope.Expect(() => new InvalidOperationException("Missing DPS ID scope"));
+            string registrationId = this.GetRegistrationId();
+
+            Assert.IsTrue(File.Exists(Path.Combine(caCertScriptPath, "certGen.sh")), "CA Certification script does not exist. Path: " + $"{caCertScriptPath}" );
+
+            CertificateAuthority ca;
+            CancellationToken token = this.TestToken;
+
+            using (var cts = new CancellationTokenSource(Context.Current.SetupTimeout))
```
damonbarry: We already have a `CancellationTokenSource` for this test; it's created in `BaseFixture` and it's token is already set to local variable `token` above. You can remove this `using` statement and the braces, and replace instances of `caToken` with `token`.

test/Microsoft.Azure.Devices.Edge.Test/Provisioning.cs
```diff
@@ -67,5 +74,61 @@ public async Task DpsSymmetricKey()
...
+            {
+                DateTime startTime = DateTime.Now;
```
damonbarry: Remove unused variable `startTime`.

test/Microsoft.Azure.Devices.Edge.Test.Common/DaemonConfiguration.cs
```diff
@@ -39,13 +40,28 @@ public void SetDpsSymmetricKey(string idScope, string registrationId, string dev
...
             this.config.ReplaceOrAdd("provisioning.source", "dps");
-            this.config.ReplaceOrAdd("provisioning.global_endpoint", "https://global.azure-devices-provisioning.net");
+            this.config.ReplaceOrAdd("provisioning.global_endpoint", GlobalEndPoint);
             this.config.ReplaceOrAdd("provisioning.scope_id", idScope);
             this.config.ReplaceOrAdd("provisioning.attestation.method", "symmetric_key");
             this.config.ReplaceOrAdd("provisioning.attestation.registration_id", registrationId);
             this.config.ReplaceOrAdd("provisioning.attestation.symmetric_key", deviceKey);
         }
 
+        public void SetDpsX509(string idScope, string registrationId, CertificateAuthority ca)
```
damonbarry: This method should take a certificate object directly instead of `CertificateAuthority`.

test/Microsoft.Azure.Devices.Edge.Test.Common/DaemonConfiguration.cs
```diff
@@ -39,13 +40,28 @@ public void SetDpsSymmetricKey(string idScope, string registrationId, string dev
...
+        {
+            Uri certUri = new Uri(ca.Certificate.CertificatePath, UriKind.Absolute);
+            Uri pKeyUri = new Uri(ca.Certificate.KeyPath, UriKind.Absolute);
+
+            this.config.RemoveIfExists("provisioning");
+            this.config.ReplaceOrAdd("provisioning.source", "dps");
+            this.config.ReplaceOrAdd("provisioning.global_endpoint", GlobalEndPoint);
+            this.config.ReplaceOrAdd("provisioning.scope_id", idScope);
+            this.config.ReplaceOrAdd("provisioning.attestation.method", "x509");
```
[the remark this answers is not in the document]
yophilav: ` SetBasicDpsProvisioning(1) ` was introduced

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/CertificateAuthority.cs
```diff
@@ -2,16 +2,20 @@
...
     using RootCaKeys = System.ValueTuple<string, string, string>;
 
     public class CertificateAuthority
     {
         readonly Option<string> scriptPath;
 
-        public EdgeCertificates Certificates { get; }
+        public EdgeCertificates EdgeCertificate { get; }
+
+        public Certificates Certificate { get; }
```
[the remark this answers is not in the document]
yophilav: - - - - - - - - - - - - -
`CreateDpsX509CaAsync()` is not equivalent to 
```
CertificateAuthority.CreateAsync()
ca.GenerateDeviceCertificatesAsync()
```
`CreateDpsX509CaAsync()` does the following: 
1. Generate & Install Root cert (if not existed)
2. Generate & Install Device cert given the Root cert
while the proposed changes would
1. Generate & Install Root cert (always)
2. Generate & Install Edge (server) cert
3. Generate & Install Device cert
Hence, the proposed change would introduce another set of certs into the test. 
 - - - - - - - - - - - - -
As of `CertificateAuthority` class, my understanding of the term is an entity that manages all the certificates being use. That's why I decided it to be a composition class composes of `EdgeCertificate` and `Certificate`
 - - - - - - - - - - - - -
`GenerateLeafCertificatesAsync(2)` is basically `GenerateDeviceCertificatesAsync(3)` with two input arguments instead of 3. I didn't change the `GenerateLeafCertificatesAsync(2)` name to limit the scope of change.

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/CertificateAuthority.cs
```diff
@@ -21,6 +25,28 @@ public static async Task<CertificateAuthority> CreateAsync(string deviceId, Root
             return new CertificateAuthority(certs, scriptPath);
         }
 
+        public static async Task<CertificateAuthority> CreateDpsX509CaAsync(string deviceId, RootCaKeys rootCa, string scriptPath, CancellationToken token)
+        {
+            if (!File.Exists(Path.Combine(scriptPath, "certs", "azure-iot-test-only.root.ca.cert.pem")))
```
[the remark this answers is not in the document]
yophilav: **It's unfortunate that we'd have to reference a hard-coded certificate path in this class. Previously all hard-coded paths were confined to the common OsPlatform class** 
This cert path is hard-coded in `certGen.sh` and `ca-certs.ps1`. The cleanest way to reduce the duplication of this hard code would be to create system environment variables that share between the script and the e2e framework. I currently do not want to make the change right now. 
 - - - - - - - - - - - 
**Does OsPlatform.Current.InstallRootCertificateAsync() fail otherwise?**
The `OsPlatform.Current.InstallRootCertificateAsync()` does NOT fail. However every time it runs a test requiring root cert, it would overwrite and existing one.  In a customer scenario, the root cert should only be installed once anyway, hence I decide to leave the root cert alone if it was installed. In the old E2E scripts only installs the root cert once per device and re-use them across all the Dps test as well.

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/Certificates.cs
yophilav: Oh I see what you mean. I'll work on this suggestion.

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/CertificateAuthority.cs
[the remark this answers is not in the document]
yophilav: Oh I see what you mean. I'll address your comment: 
```
Previously we had "edge certificates" and "leaf certificates". Edge certs were really CA certs, and leaf certs were really "identity" (aka end-entity) certs. With your new test, this distinction isn't enough because now we have edge CA certs and edge identity certs (as well as leaf identity certs). Maybe the names for the two cert classes should be CaCertificate and IdentityCertificate.
CertificateAuthority would return these certificate objects from GenerateCaCertificate() and GenerateIdentityCertificate() respectively.
```

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/Certificates.cs
yophilav: ` Certificates ` -> `IdCertificates`
`EdgeCertificate` -> `CaCertificates`

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/Certificates.cs
yophilav: The hard-coded value have been centralized to `FixedPaths.cs`

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/EdgeCertificates.cs
yophilav: The hard-coded value have been centralized to FixedPaths.cs

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/Certificates.cs
yophilav: An empty constructor is necessary for the code to compile for the derived `CaCertificate` class to set its params

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/EdgeCertificates.cs
yophilav: There is an unfortunate identical constructor signature of `IdCertificates(string, string)`. I created a task for myself to refactor the X509 Cert in test framework.

test/Microsoft.Azure.Devices.Edge.Test/Provisioning.cs
yophilav: remove `caToken`

test/Microsoft.Azure.Devices.Edge.Test/Provisioning.cs
yophilav: removed `startTime`

test/Microsoft.Azure.Devices.Edge.Test.Common/IOsPlatform.cs
```diff
@@ -19,16 +19,16 @@ public interface IOsPlatform
         // After calling this function, the following files will be available under {scriptPath}:
         //  certs/iot-device-{deviceId}-full-chain.cert.pem
         //  private/iot-device-{deviceId}.key.pem
-        Task<Certificates> CreateDeviceCertificatesAsync(string deviceId, string scriptPath, CancellationToken token);
+        Task<IdCertificates> GenerateIdentiyCertificatesAsync(string deviceId, string scriptPath, CancellationToken token);
```
damonbarry: Typo in method name: Identiy

test/Microsoft.Azure.Devices.Edge.Test.Common/OsPlatform.cs
```diff
@@ -57,13 +57,9 @@ protected void InstallTrustedCertificates(IEnumerable<X509Certificate2> certs, S
...
             CancellationToken token)
         {
-            await Profiler.Run(
-                async () =>
-                {
-                    string[] output = await Process.RunAsync(command.name, command.args, token);
-                    Log.Verbose(string.Join("\n", output));
-                },
-                "Executed: " + command.name + ' ' + command.args );
+            Log.Verbose("Executed: " + command.name + ' ' + command.args);
```
damonbarry: Since this Log call happens before the command is executed, "Executed" should be "Execute" or "Executing"...

test/Microsoft.Azure.Devices.Edge.Test/Provisioning.cs
```diff
@@ -90,28 +90,22 @@ public async Task DpsX509()
...
-                    {
-                        config.SetDpsX509(idScope, registrationId, ca);
-                        config.Update();
-                        return Task.FromResult((
-                            "with DPS X509 attestation for '{Identity}'",
-                            new object[] { registrationId }));
-                    },
-                    caToken);
-            }
+            ca = await CertificateAuthority.CreateDpsX509CaAsync(
```
damonbarry: Pls declare and assign on the same line

test/Microsoft.Azure.Devices.Edge.Test.Common/DaemonConfiguration.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.Devices.Edge.Test.Common
 
     public class DaemonConfiguration
     {
+        private const string GlobalEndPoint = "https://global.azure-devices-provisioning.net";
```
damonbarry: Remove `private`

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/CaCertificates.cs
```diff
@@ -20,7 +16,28 @@ public class EdgeCertificates
                 new X509Certificate2(X509Certificate.CreateFromCertFile(this.TrustedCertificatesPath))
             };
 
-        public EdgeCertificates(string certificatePath, string keyPath, string trustedCertsPath)
+        private string[] GetEdgeCertFileLocation(string deviceId)
```
damonbarry: Remove `private`

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/IdCertificates.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Test.Common.Certs
+{
+    public class IdCertificates
+    {
+        private string[] GetFileLocation(string deviceId)
```
damonbarry: Remove `private`

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/CertificateAuthority.cs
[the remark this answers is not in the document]
damonbarry: > my understanding of the term is an entity that manages all the certificates being use ...

A certificate authority does not _manage_ certs; it _issues_ certs.

> CreateDpsX509CaAsync() is not equivalent to ...

I guess the part I left out of my original comment was that `CertificateAuthority` would need an additional (non-static) method that generates a CA certificate based on the root cert.

So the setup fixture for the `TransparentGateway` test would call:
```
CertificateAuthority.CreateAsync()
ca.GenerateCaCertificateAsync()
```

...while the `DpsX509Test` test would call:
```
CertificateAuthority.CreateAsync()
ca.GenerateDeviceCertificatesAsync()
```

test/Microsoft.Azure.Devices.Edge.Test/Provisioning.cs
```diff
@@ -67,5 +74,55 @@ public async Task DpsSymmetricKey()
...
+        public async Task DpsX509()
+        {
+            (string, string, string) rootCa =
+                        Context.Current.RootCaKeys.Expect(() => new InvalidOperationException("Missing root CA keys"));
+            string caCertScriptPath =
+                        Context.Current.CaCertScriptPath.Expect(() => new InvalidOperationException("Missing CA cert script path"));
+            string idScope = Context.Current.DpsIdScope.Expect(() => new InvalidOperationException("Missing DPS ID scope"));
+            string registrationId = this.GetRegistrationId();
+
+            Assert.IsTrue(File.Exists(Path.Combine(caCertScriptPath, "certGen.sh")), "CA Certification script does not exist. Path: " + $"{caCertScriptPath}" );
```
damonbarry: This assert probably won't fail on Windows (because windows needs `ca-certs.ps1` which happens to live in the same directory), but it also doesn't make sense for Windows. The test shouldn't really care if that script exists because it's an implementation detail of `CertificateAuthority` and `LeafDevice`. If the script doesn't exist we should know it from the output of `Process.RunAsync`, so I'd remove just this.

test/Microsoft.Azure.Devices.Edge.Test.Common/IOsPlatform.cs
yophilav: `GenerateIdentiyCertificatesAsync()` -> `GenerateIdentityCertificatesAsync()`

test/Microsoft.Azure.Devices.Edge.Test.Common/OsPlatform.cs
yophilav: EnGiRsh Is HaRd. 
`Executed` -> `Executing`

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/CertificateAuthority.cs
```diff
@@ -2,48 +2,69 @@
...
+    using Serilog;
     using RootCaKeys = System.ValueTuple<string, string, string>;
 
     public class CertificateAuthority
     {
         readonly Option<string> scriptPath;
 
-        public EdgeCertificates Certificates { get; }
+        // TODO: EdgeCertificate needs to be moved into certificate types.
+        public CaCertificates EdgeCertificate { get; set; }
```
damonbarry: Property name should be plural to match type (EdgeCertificates)

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/CertificateAuthority.cs
```diff
@@ -2,48 +2,69 @@
...
-        CertificateAuthority(EdgeCertificates certs)
+        CertificateAuthority(string scriptPath)
         {
-            this.Certificates = certs;
-            this.scriptPath = Option.None<string>();
+            this.scriptPath = Option.Maybe(scriptPath);
         }
 
-        CertificateAuthority(EdgeCertificates certs, string scriptPath)
+        CertificateAuthority(CaCertificates certs)
```
damonbarry: `this.scriptPath` needs to be set to `Option.None<string>()` explicitly in this constructor, otherwise methods `GenerateIdentityCertificatesAsync` and `GenerateCaCertificatesAsync` will throw a null reference exception.

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/CertificateAuthority.cs
```diff
@@ -2,48 +2,69 @@
...
 
         public static async Task<CertificateAuthority> CreateAsync(string deviceId, RootCaKeys rootCa, string scriptPath, CancellationToken token)
         {
-            (string rootCertificate, string rootPrivateKey, string rootPassword) = rootCa;
-            await OsPlatform.Current.InstallRootCertificateAsync(rootCertificate, rootPrivateKey, rootPassword, scriptPath, token);
-            EdgeCertificates certs = await OsPlatform.Current.GenerateEdgeCertificatesAsync(deviceId, scriptPath, token);
-            return new CertificateAuthority(certs, scriptPath);
+            if (!File.Exists(Path.Combine(scriptPath, FixedPaths.RootCaCert.Cert)))
+            {
+                // [9/26/2019] Setup all the environment for the certificates.
```
damonbarry: Why does this comment have a date?

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/CertificateAuthority.cs
```diff
@@ -2,48 +2,69 @@
...
+                // More info: /iotedge/scripts/linux/runE2ETest.sh : run_test()
```
damonbarry: runE2ETest.sh will go away once these tests roll out, so we probably don't want to reference it here.

test/Microsoft.Azure.Devices.Edge.Test.Common/certs/Certificates.cs
damonbarry: Let's try to find a way to fix this in the future...

PR #1810, opened 2019-10-05
VSTS to Azure DevOps

SychevIgor:
VSTS was re-branded a year ago


kkdawkins:
Thanks for the update! I don't have an issue with this unless @damonbarry does?

damonbarry:
@kkdawkins This is fine...

PR #1817, opened 2019-10-07
Bundle check and inspect

lfitchett:
This adds to the support-bundle command, adding the json output from iotedge check and docker inspect output for each module.


edgelet/iotedge/src/error.rs
```diff
@@ -44,6 +44,12 @@ pub enum ErrorKind {
 
     #[fail(display = "Could not write to file")]
     WriteToFile,
+
+    #[fail(display = "Unable to bundle iotedge check")]
```
arsing: Put enum variants in alphabetical order.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -157,6 +165,134 @@ where
...
+        println!("Getting system logs for iotedge");
+        let since_time: DateTime<Utc> = DateTime::from_utc(
+            NaiveDateTime::from_timestamp(state.log_options.since().into(), 0),
+            Utc,
+        );
+        let since = since_time.format("%F %T").to_string();
+
+        #[cfg(unix)]
+        let inspect = ShellCommand::new("journalctl")
+            .args(&["-u", "iotedge"])
```
arsing: Also pass `-a`

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -157,6 +165,134 @@ where
...
+            .args(&["-S", &since])
+            .arg("--no-pager")
+            .output();
+
+        #[cfg(windows)]
+        let inspect = ShellCommand::new("Get-WinEvent")
```
arsing: Are you sure this works? `Get-WinEvent` is a PS commandlet. You'd need to spawn `powershell.exe` with argv `["-NoProfile", "-Command", "Get-WinEvent -ea ..."]`

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -157,6 +165,134 @@ where
...
+            .map_err(|err| Error::from(err.context(ErrorKind::WriteToFile)))?;
+
+        Ok(state)
+    }
+
+    fn write_check_to_file(mut state: BundleState<M>) -> Result<BundleState<M>, Error> {
+        let iotedge = env::args().into_iter().nth(0).unwrap();
+        println!("Calling iotedge check");
+        let check = ShellCommand::new(iotedge)
+            .arg("check")
```
arsing: You also need to pass in `--iothub-hostname` for DPS.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -157,6 +165,134 @@ where
...
+
+    fn write_inspect_to_file(
+        mut state: BundleState<M>,
+        module_name: String,
+    ) -> Result<BundleState<M>, Error> {
+        println!("Running docker inspect for {}", module_name);
+        let mut inspect = ShellCommand::new("docker");
+
+        #[cfg(windows)]
+        inspect.args(&["-H", "npipe://./pipe/iotedge_moby_engine"]);
```
arsing: This is only the case if the user is using Windows containers.

Admittedly it's hard at this layer to read the config.yaml and figure out what the real URI is. And it's likely that anyone using `iotedge support-bundle` is in a production scenario, so they're not using Linux containers.

So if you want to leave it like this, it's fine, but mention this justification in a comment.

edgelet/iotedge/src/error.rs
arsing: I think all of these (including `WriteToFile`) could be merged into `#[fail(display = "Could not generate support bundle")] SupportBundle`. Are the inner errors good enough to figure out what substep failed?

lfitchett: They are readable. Plus the errors for the individual components (logs, check, inspect) don't fail the command, just note the error in the console and output file. So if inspect fails, you'll still get the logs and check output.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -176,7 +183,7 @@ where
 
         #[cfg(unix)]
         let inspect = ShellCommand::new("journalctl")
-            .args(&["-u", "iotedge"])
+            .args(&["-u", "-a", "iotedge"])
```
arsing: This won't work. It needs to be `"-a", "-u", "iotedge"`

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -257,6 +269,12 @@ where
         println!("Running docker inspect for {}", module_name);
         let mut inspect = ShellCommand::new("docker");
 
+        /***
+         * Note: this assumes using windows containers on a windows machine.
+         * This is the expected production scenario.
+         * Since the bundle command does not read the config.yaml, it will fail if Linux containers are used.
```
arsing: `Since the bundle command does not read the config.yaml, it cannot use the `moby.runtime_uri` from there.`

edgelet/iotedge/src/support_bundle.rs
arsing: Hasn't been resolved. (Make sure you don't resolve unless you've actually made the requested change, or added a comment why you're not making the requested change).

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -130,33 +143,163 @@ where
...
+        let since_time: DateTime<Utc> = DateTime::from_utc(
+            NaiveDateTime::from_timestamp(state.log_options.since().into(), 0),
+            Utc,
+        );
+        let since = since_time.format("%F %T").to_string();
+
+        #[cfg(unix)]
+        let inspect = ShellCommand::new("journalctl")
+            .args(&["-u", "-a", "iotedge"])
+            .args(&["-S", &since])
```
arsing: You don't need two calls to `args` and one call to `arg`. A single `args` would cover all of that.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -130,33 +143,163 @@ where
...
+            .arg("--no-pager")
+            .output();
+
+        #[cfg(windows)]
+         let inspect = ShellCommand::new("powershell.exe")
+            .arg("-Command")
```
arsing: You should have `-NoProfile`. Why did you remove it?

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -130,33 +143,163 @@ where
...
+
+        Ok(state)
+    }
+
+    fn write_check_to_file(mut state: BundleState<M>) -> Result<BundleState<M>, Error> {
+        let iotedge = env::args().nth(0).unwrap();
+        println!("Calling iotedge check");
+
+        let mut check = ShellCommand::new(iotedge);
+        check.arg("check").args(&["-o", "json"]);
```
arsing: Same here. There's no reason to call `arg` and `args` separately when a single `args` will suffice.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -130,33 +143,163 @@ where
...
+            .arg("-Command")
+            .arg(&format!(r"Get-WinEvent -ea SilentlyContinue -FilterHashtable @{{ProviderName='iotedged';LogName='application';StartTime='{}'}} |
+                            Select TimeCreated, Message |
+                            Sort-Object @{{Expression='TimeCreated';Descending=$false}} |
+                            Format-Table -AutoSize -Wrap", since))
+            .output();
+
+        let (file_name, output) = if let Ok(result) = inspect {
+            if result.status.success() {
+                ("edgelet.txt", result.stdout)
```
arsing: `edgelet.txt` -> `iotedged.txt`

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -130,33 +143,163 @@ where
...
+            } else {
+                ("edgelet_err.txt", result.stderr)
```
arsing: `edgelet_err.txt` -> `iotedged_err.txt`

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -40,8 +45,11 @@ where
     fn execute(self) -> Self::Future {
         let result = future::result(self.make_state())
             .and_then(SupportBundle::write_all_logs)
+            .and_then(SupportBundle::write_edgelet_log_to_file)
+            .and_then(SupportBundle::write_check_to_file)
+            .and_then(SupportBundle::write_all_inspects)
             .map(drop)
```
arsing: The next `map(|_|)` will already drop it. Why do you need to `drop` explicitly?

lfitchett: I was just being explicit. I'll remove it

edgelet/iotedge/src/support_bundle.rs
lfitchett: multiple lines for organization

edgelet/iotedge/src/support_bundle.rs
lfitchett: added back in

PR #1828, opened 2019-10-09
5473107 [k8s] don't update edgeAgent in Public Preview.

CindyXing:
This is to disable edgeagent updating itself for the kubernetes case by not creating a deployment. 
It also kind of laying out the long term updateability work.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
@@ -63,6 +63,12 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Planners
                 throw new InvalidModuleException($"Kubernetes deployment currently only handles type={typeof(DockerConfig).FullName}");
             }
 
+            // This is a workaround for K8s public preview refresh
+            // TODO: mark edgeagent module updateable when merging back to main release
+            if (desired.Modules.ContainsKey(Constants.EdgeAgentModuleName)) {
+                desired.Modules[Constants.EdgeAgentModuleName].IsUpdateable = false;
+            }
+
             Diff moduleDifference = desired.Diff(current);
 
             Plan plan;
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/IModule.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/IModule.cs
@@ -119,6 +119,9 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Core
         IDictionary<string, EnvVal> Env { get; }
 
         bool IsOnlyModuleStatusChanged(IModule other);
+
+        [JsonIgnore]
+        bool IsUpdateable { get; set; }
     }
 
     public interface IModule<TConfig> : IModule, IEquatable<IModule<TConfig>>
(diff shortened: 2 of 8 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
```diff
@@ -63,6 +63,12 @@ public class KubernetesPlanner : IPlanner
                 throw new InvalidModuleException($"Kubernetes deployment currently only handles type={typeof(DockerConfig).FullName}");
             }
 
+            // This is a workaround for K8s public preview refresh
+            // TODO: mark edgeagent module updateable when merging back to main release
+            if (desired.Modules.ContainsKey(Constants.EdgeAgentModuleName)) {
```
dmolokanov: I suggest we just exclude EdgeAgent module from the list of modules here instead of mark modules to be updatable.
```
// This is a workaround for K8s public preview refresh
// TODO: mark edgeagent module updateable when merging back to main release
desired = new ModulesSet(desired.Remove(Constants.EdgeAgentModuleName));
current = new ModulesSet(current.Remove(Constants.EdgeAgentModuleName));
```

It will prevent EdgeAgent to generate commands to Add/Update/Remove EdgeAgent module

CindyXing:
> This was intended as a simple fix for Kubernetes mode only. But the code changes spread across multiple projects. Can we possibly localize it in EdgeAgent Kubernetes project?
> For GA we have to revert these changed and it will be much easier to do when there is minimal impact on a codebase. Any thoughts?

Agree. The issue is that the modules inside the moduleset is immutabledictionary. I tried what you  listed below to remove edgeagent module.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/planners/KubernetesPlanner.cs
CindyXing: modules is an immutable dictionary in moduleset that we can't change.

CindyXing: That's why I add the getter/setter in the module interface. And the getter/setter can be used for long term.

dmolokanov: ImmutableDictionary preserves its inside state right but when someone tries to modify it, operation returns a new instance of a dictionary that contains required change.
This is why I suggested creating a new instance of ModuleSets.
Here is some more info https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.iimmutabledictionary-2.remove?view=netcore-3.0#System_Collections_Immutable_IImmutableDictionary_2_Remove__0_

CindyXing: I see. Thanks Dennis. 

If edgeagent status is changed, what's going to happen? Is it taking care of here?

dmolokanov: KubernetesRuntimeOperator watches pods inside namespaces and generates RuntimeInfo based on information about pods marked with a special label. Then it reports the status of the pod to iothub.
Other words, it is different path of EdgeAgent that should take care of status change for all modules (EdgeAgent included). We should be fine there.

But I agree, in the long run we may need this functionality in place. By that time we will have to come up with holistic approach though.

CindyXing: Sounds good. Agree that it is risky making a core change for public preview refresh at this point.

PR #1833, opened 2019-10-10
Include test runtime data in check json output

lfitchett:
This breaks each test into its own file and gives it a corresponding struct. These structs are filled with copies of the data as the test runs and included in the json output.

It also adds system info like remaining disk space to additional info.

It adds a verbose flag. 

Here is a sample output json: https://gist.github.com/lfitchett/bb08656b8fcebef19315cff6645e1d01


lfitchett:
Waiting until this has a clear spec and we know when it will be released

edgelet/iotedge/src/check/additional_info.rs
```diff
@@ -184,3 +195,71 @@ pub(super) fn os_version() -> Result<
...
+            file_system: str::from_utf8(disk.get_file_system()).unwrap().to_owned(),
+            file_type: format!("{:?}", disk.get_type()),
+        }
+    }
+}
+
+fn pretty_kbyte(bytes: u64) -> String {
+    match Byte::from_unit(bytes as f64, ByteUnit::KiB) {
+        Ok(b) => b.get_appropriate_unit(true).format(2),
+        Err(_) => "Could not find bytes".to_owned(),
```
arsing: "Could not find bytes" doesn't make sense. And don't swallow the error.

`Err(err) => format!("could not parse bytes value: {}", err),`

(As it happens, the only way this can fail is if `bytes` is negative, which it can't be since it started out as a `u64`. But might as well handle the error in case it changes in the future.)

edgelet/iotedge/src/check/checks/certificates_quickstart.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use super::identity_certificate_expiry::CertificateValidity;
+use crate::check::{checker::Checker, Check, CheckResult};
+
+#[derive(Default, serde_derive::Serialize)]
+pub struct CertificatesQuickstart {
+    device_ca_cert_path: Option<PathBuf>,
+    certificate_info: Option<CertificateValidity>,
+}
+impl Checker for CertificatesQuickstart {
```
arsing: Add an empty line between the end of the struct and the impl block.

edgelet/iotedge/src/check/checks/certificates_quickstart.rs
```diff
@@ -0,0 +1,108 @@
...
+        "production readiness: certificates"
+    }
+    fn result(&mut self, check: &mut Check) -> CheckResult {
+        self.execute(check).unwrap_or_else(CheckResult::Failed)
+    }
+    fn get_json(&self) -> serde_json::Value {
+        serde_json::to_value(self).unwrap()
+    }
+}
+impl CertificatesQuickstart {
```
arsing: Add an empty line between the end of the previous impl block and this one.

edgelet/iotedge/src/check/checks/container_engine_installed.rs
```diff
@@ -0,0 +1,138 @@
...
+        check.additional_info.docker_version = check.docker_server_version.clone();
+
+        self.docker_host_arg = check.docker_host_arg.clone();
+        self.docker_server_version = check.docker_server_version.clone();
+
+        Ok(CheckResult::Ok)
+    }
+}
+
+pub fn docker<I>(
```
arsing: This gets used by a lot of checks, so put this in `checks/mod.rs`

edgelet/iotedge/src/check/checks/container_connect_iothub.rs
```diff
@@ -0,0 +1,155 @@
+use crate::check::{
+    checker::Checker, upstream_protocol_port::UpstreamProtocolPort, Check, CheckResult,
+};
+
+pub fn get_host_container_iothub_tests() -> Vec<Box<dyn Checker>> {
```
arsing: I think just needs to be `pub(crate)`

edgelet/iotedge/src/check/checks/certificates_quickstart.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use failure::{self, Context, ResultExt};
+
+use edgelet_core::RuntimeSettings;
+
+use super::identity_certificate_expiry::CertificateValidity;
+use crate::check::{checker::Checker, Check, CheckResult};
+
+#[derive(Default, serde_derive::Serialize)]
+pub struct CertificatesQuickstart {
```
arsing: I don't think these need to be `pub`, just `pub(crate)`.

edgelet/iotedge/src/check/checks/container_connect_iothub.rs
```diff
@@ -0,0 +1,155 @@
...
+                    port,
+                ))
+                .into());
+        }
+
+        Ok(CheckResult::Ok)
+    }
+}
+
+fn make_box(
```
arsing: `make_box` -> `make_check`

edgelet/iotedge/src/check/checks/container_engine_is_moby.rs
```diff
@@ -0,0 +1,81 @@
+use failure::{self, Context, Fail};
+
+use crate::check::{checker::Checker, Check, CheckResult};
+
+#[derive(Default, serde_derive::Serialize)]
+pub struct ContainerEngineIsMoby {
+    docker_server_version: Option<String>,
+    #[cfg(windows)]
```
arsing: Don't make this `#[cfg(windows)]`. Just let it be `None` so that the shape of the JSON is the same regardless of OS.

edgelet/iotedge/src/check/checks/host_connect_iothub.rs
```diff
@@ -0,0 +1,80 @@
...
+            &(&**iothub_hostname, self.port_number),
+            iothub_hostname,
+            &format!("{}:{}", iothub_hostname, self.port_number),
+        )?;
+
+        Ok(CheckResult::Ok)
+    }
+}
+
+fn make_box(
```
arsing: `make_box` -> `make_check`

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
```diff
@@ -0,0 +1,150 @@
...
+
+use failure::{self, Context, ResultExt};
+
+use edgelet_core::{self, AttestationMethod, ManualAuthMethod, Provisioning, RuntimeSettings};
+
+use crate::check::{checker::Checker, Check, CheckResult};
+
+#[derive(Default, serde_derive::Serialize)]
+pub struct IdentityCertificateExpiry {
+    // provisioning: Option<Provisioning>, /* Note: contains shared keys */
```
arsing: We do want to know the kind of provisioning, so add a `provisioning_mode: Option<&'static str>` and set it to `Some("manual")` or `Some("dps")` or `Some("external")`.

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+impl IdentityCertificateExpiry {
+    fn execute(&mut self, check: &mut Check) -> Result<CheckResult, failure::Error> {
+        let settings = if let Some(settings) = &check.settings {
+            settings
+        } else {
+            return Ok(CheckResult::Skipped);
+        };
+        // self.provisioning = Some(settings.provisioning().to_owned());
```
arsing: Remove this commented-out line.

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
```diff
@@ -0,0 +1,150 @@
...
+            }
+            Provisioning::External(_) => (),
+        }
+
+        Ok(CheckResult::Ignored)
+    }
+}
+
+#[derive(Debug, serde_derive::Serialize, Clone)]
+pub struct CertificateValidity {
```
arsing: `pub(crate)`

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        Ok(CertificateValidity {
+            cert_name,
+            cert_path,
+            not_after,
+            not_before,
+        })
+    }
+
+    pub(crate) fn to_check_result(&self) -> Result<CheckResult, failure::Error> {
```
arsing: This seems like it could be private.

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        Ok(CheckResult::Ignored)
+    }
+}
+
+#[derive(Debug, serde_derive::Serialize, Clone)]
+pub struct CertificateValidity {
+    cert_name: String,
+    cert_path: PathBuf,
+    pub not_after: chrono::DateTime<chrono::Utc>,
```
arsing: This seems like it could be private.

edgelet/iotedge/src/check/checks/iotedged_version.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use failure::{self, Context, Fail, ResultExt};
+use regex::Regex;
+
+use crate::check::{checker::Checker, Check, CheckResult};
+
+#[derive(Default, serde_derive::Serialize)]
+pub struct IotedgedVersion {
+    version: Option<String>,
+    //TODO: find if 1.* or specified
```
arsing: What does this TODO mean?

edgelet/iotedge/src/check/checks/host_connect_iothub.rs
```diff
@@ -0,0 +1,80 @@
+use crate::check::{
+    checker::Checker, upstream_protocol_port::UpstreamProtocolPort, Check, CheckResult,
+};
+
+pub fn get_host_connect_iothub_tests() -> Vec<Box<dyn Checker>> {
```
arsing: `pub(crate)`

edgelet/iotedge/src/check/mod.rs
```diff
@@ -375,25 +217,62 @@ impl Check {
...
+            .map(|check| check.id())
+            .collect();
+
+        result.into_iter()
     }
 
     pub fn print_list() -> Result<(), Error> {
         // All our text is ASCII, so we can measure text width in bytes rather than using unicode-segmentation to count graphemes.
-        let widest_section_name_len = CHECKS
+        let widest_section_name_len = Check::checks()
```
arsing: Just get the checks once with `let checks = Check::checks();` , then use this same `checks` for `widest_section_name_len`, `widest_check_id_len` and the actual for-loop.

edgelet/iotedge/src/check/checker.rs
```diff
@@ -0,0 +1,8 @@
+use crate::check::{Check, CheckResult};
+
+pub trait Checker {
+    fn id(&self) -> &'static str;
+    fn description(&self) -> &'static str;
+    fn result(&mut self, check: &mut Check) -> CheckResult;
```
arsing: Should probably call this `execute`

edgelet/iotedge/src/check/mod.rs
```diff
@@ -227,7 +66,7 @@ pub enum OutputFormat {
 ///
 /// Check functions return `Result<CheckResult, failure::Error>` where `Err` represents the check failed.
 #[derive(Debug)]
-enum CheckResult {
+pub enum CheckResult {
```
arsing: `pub(crate)`

edgelet/iotedge/src/main.rs
```diff
@@ -270,6 +270,12 @@ fn run() -> Result<(), Error> {
                         .value_name("IOTHUB_HOSTNAME")
                         .help("Sets the hostname of the Azure IoT Hub that this device would connect to. If using manual provisioning, this does not need to be specified.")
                         .takes_value(true),
+                ).arg(
+                    Arg::with_name("verbose")
```
arsing: Is there a reason the user would ever not want verbose output?

arsing:
Thank you. This is great.

edgelet/iotedge/src/main.rs
arsing: Talked about it with Venkat. I don't think it helps much, but it's consistent with other things.

edgelet/iotedge/src/check/checks/container_connect_iothub.rs
```diff
@@ -0,0 +1,155 @@
...
+            "container on the IoT Edge module network can connect to IoT Hub MQTT port",
+            UpstreamProtocolPort::Mqtt,
+            true,
+        ),
+    ]
+}
+
+#[derive(serde_derive::Serialize)]
+pub struct ContainerConnectIotHub {
+    upstream_protocol_port: UpstreamProtocolPort,
```
arsing: This can be `serde(skip)` - it's redundant with the port number.

edgelet/iotedge/src/check/checks/host_connect_iothub.rs
```diff
@@ -0,0 +1,80 @@
...
+            "host-connect-iothub-mqtt",
+            "host can connect to and perform TLS handshake with IoT Hub MQTT port",
+            UpstreamProtocolPort::Mqtt,
+        ),
+    ]
+}
+
+#[derive(serde_derive::Serialize)]
+pub struct HostConnectIotHub {
+    upstream_protocol_port: UpstreamProtocolPort,
```
arsing: Same here - this can be `serde(skip)`

edgelet/iotedge/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-#![deny(rust_2018_idioms, warnings)]
-#![deny(clippy::all, clippy::pedantic)]
+// #![deny(rust_2018_idioms, warnings)]
```
arsing: Uncomment these.

edgelet/iotedge/src/main.rs
```diff
@@ -270,6 +270,12 @@ fn run() -> Result<(), Error> {
                         .value_name("IOTHUB_HOSTNAME")
```
arsing: At line 177, change the helptext to `"Output format. Note that JSON output contains some additional information like OS name, OS version, disk space, etc."`

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
lfitchett: it's used in certificate_quickstart.rs

let CertificateValidity { not_after, .. } = CertificateValidity::parse(

edgelet/iotedge/src/check/checks/iotedged_version.rs
lfitchett: Css wants to know if the customer specified a specific version (1.0.8) or simply 1.*

edgelet/iotedge/src/main.rs
lfitchett: venkat asked for it as a flag. talked to him, it might get changed to --quiet or just removed

edgelet/iotedge/src/check/checks/iotedged_version.rs
arsing: That has nothing to do with iotedged version. iotedged version is the version of the iotedged binary, which you can see is obtained by `iotedged --version`. The thing CSS is talking about the versions of edge agent and hub in the deployment template.

THIRDPARTYNOTICES
```diff
@@ -6033,3 +5969,29 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
arsing: Missing newline at end of file.

edgelet/iotedge/src/check/additional_info.rs
```diff
@@ -260,6 +260,6 @@ impl DiskInfo {
 fn pretty_kbyte(bytes: u64) -> String {
     match Byte::from_unit(bytes as f64, ByteUnit::KiB) {
         Ok(b) => b.get_appropriate_unit(true).format(2),
-        Err(_) => "Could not find bytes".to_owned(),
+        Err(err) => format!("could not parse bytes value: {:?}", err),
```
arsing: `{}`, not `{:?}`

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
```diff
@@ -10,7 +10,7 @@ use crate::check::{checker::Checker, Check, CheckResult};
 
 #[derive(Default, serde_derive::Serialize)]
 pub struct IdentityCertificateExpiry {
-    // provisioning: Option<Provisioning>, /* Note: contains shared keys */
+    provisioning: Option<String>,
```
arsing: `provisioning_mode`, not `provisioning`

arsing: and `&'static str`, not `String`

edgelet/edgelet-core/src/settings.rs
```diff
@@ -404,6 +404,11 @@ pub enum Provisioning {
     Dps(Box<Dps>),
     External(External),
 }
+impl Provisioning {
+    pub fn to_string(&self) -> String {
+        format!("{:?}", self)
```
arsing: This is unnecessary, and leaks the secrets that we don't want to leak.

edgelet/iotedge/src/check/checks/mod.rs
```diff
@@ -1,42 +1,42 @@
...
-pub mod host_connect_iothub;
-pub mod host_local_time;
-pub mod hostname;
-pub mod identity_certificate_expiry;
-pub mod iotedged_version;
-pub mod storage_mounted_from_host;
-pub mod well_formed_config;
-pub mod well_formed_connection_string;
-pub mod windows_host_version;
+pub(crate) mod certificates_quickstart;
```
arsing: Why do the mods need to be `pub(crate)` instead of private?

edgelet/iotedge/src/check/additional_info.rs
lfitchett: byte_unit::ByteError doesn't impl display, only debug

edgelet/iotedge/src/check/checks/certificates_quickstart.rs
```diff
@@ -0,0 +1,111 @@
...
+                        "Could not find device CA certificate under {}",
+                        certs_dir.display(),
+                    ))
+                })?
+            }
+        });
+        self.device_ca_cert_path = check.device_ca_cert_path.to_owned();
+
+        if settings.certificates().device_cert().is_none() {
+            let CertificateValidity { not_after, .. } = CertificateValidity::parse(
```
arsing: The original `CertificateValidity` value also needs to be assigned to `self.certificate_info`

edgelet/iotedge/src/check/checks/container_engine_dns.rs
```diff
@@ -0,0 +1,79 @@
...
+        };
+        let daemon_config: DaemonConfig = serde_json::from_reader(daemon_config_file)
+            .with_context(|_| {
+                format!(
+                    "Could not parse container engine config file {}",
+                    check.container_engine_config_path.display(),
+                )
+            })
+            .context(MESSAGE)?;
+        self.dns = daemon_config.dns.to_owned();
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/additional_info.rs
```diff
@@ -184,3 +195,74 @@ pub(super) fn os_version() -> Result<
...
+        DiskInfo {
+            name: disk.get_name().to_str().unwrap().to_owned(),
+            percent_free,
+            available_space: Byte::from_bytes(u128::from(available_space))
+                .get_appropriate_unit(true)
+                .format(2),
+            total_space: Byte::from_bytes(u128::from(total_space))
+                .get_appropriate_unit(true)
+                .format(2),
+            file_system: str::from_utf8(disk.get_file_system()).unwrap().to_owned(),
```
arsing: This will panic if the result of `get_file_system` is not UTF-8.

`file_system: String::from_utf8_lossy(disk.get_file_system()).into_owned(),`

edgelet/iotedge/src/check/checks/certificates_quickstart.rs
```diff
@@ -0,0 +1,111 @@
...
+
+                device_ca_cert_path.ok_or_else(|| {
+                    Context::new(format!(
+                        "Could not find device CA certificate under {}",
+                        certs_dir.display(),
+                    ))
+                })?
+            }
+        });
+        self.device_ca_cert_path = check.device_ca_cert_path.to_owned();
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/checks/container_connect_iothub.rs
```diff
@@ -0,0 +1,155 @@
...
+        args.extend(&[
+            &check.diagnostics_image_name,
+            "/iotedge-diagnostics",
+            "iothub",
+            "--hostname",
+            iothub_hostname,
+            "--port",
+            &port,
+        ]);
+        self.diagnostics_image_name = Some(check.diagnostics_image_name.to_owned());
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/checks/container_engine_installed.rs
```diff
@@ -0,0 +1,101 @@
...
+                    }
+                }
+
+                return Err(err.context(error_message).into());
+            }
+        };
+
+        check.docker_host_arg = Some(docker_host_arg);
+
+        check.docker_server_version = Some(String::from_utf8_lossy(&output).trim().to_owned());
```
arsing: `.to_owned()` -> `.into_owned()`

`.to_owned()` uses `str::to_owned` so it always clones, whereas `Cow::into_owned` can potentially reuse itself if it's already the `Owned` variant.

edgelet/iotedge/src/check/checks/container_engine_is_moby.rs
```diff
@@ -0,0 +1,83 @@
...
+        #[cfg(windows)]
+        {
+            let settings = if let Some(settings) = &check.settings {
+                settings
+            } else {
+                return Ok(CheckResult::Skipped);
+            };
+
+            let moby_runtime_uri = settings.moby_runtime().uri().to_string();
+            self.moby_runtime_uri = Some(moby_runtime_uri.to_owned());
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/checks/container_engine_logrotate.rs
```diff
@@ -0,0 +1,95 @@
...
+        };
+        let daemon_config: DaemonConfig = serde_json::from_reader(daemon_config_file)
+            .with_context(|_| {
+                format!(
+                    "Could not parse container engine config file {}",
+                    check.container_engine_config_path.display(),
+                )
+            })
+            .context(MESSAGE)?;
+        self.daemon_config = Some(daemon_config.to_owned());
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/checks/container_engine_ipv6.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+impl ContainerEngineIPv6 {
+    fn inner_execute(&mut self, check: &mut Check) -> Result<CheckResult, failure::Error> {
+        const MESSAGE: &str =
+        "Container engine is not configured for IPv6 communication.\n\
+         Please see https://aka.ms/iotedge-docker-ipv6 for a guide on how to enable IPv6 support.";
+
+        #[derive(serde_derive::Deserialize)]
```
arsing: Move this out to the top level like it is for `ContainerEngineLogRotate`

edgelet/iotedge/src/check/checks/container_engine_dns.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+impl ContainerEngineDns {
+    fn inner_execute(&mut self, check: &mut Check) -> Result<CheckResult, failure::Error> {
+        const MESSAGE: &str =
+        "Container engine is not configured with DNS server setting, which may impact connectivity to IoT Hub.\n\
+         Please see https://aka.ms/iotedge-prod-checklist-dns for best practices.\n\
+         You can ignore this warning if you are setting DNS server per module in the Edge deployment.";
+
+        #[derive(serde_derive::Deserialize)]
```
arsing: Move this out to the top level like it is for `ContainerEngineLogRotate`

edgelet/iotedge/src/check/checks/container_local_time.rs
```diff
@@ -0,0 +1,71 @@
...
+        let docker_host_arg = if let Some(docker_host_arg) = &check.docker_host_arg {
+            docker_host_arg
+        } else {
+            return Ok(CheckResult::Skipped);
+        };
+
+        let expected_duration = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .context("Could not query local time of host")?;
+        self.expected_duration = Some(expected_duration.to_owned());
```
arsing: `self.expected_duration = Some(expected_duration);`

`std::time::Duration` is `Copy`.

edgelet/iotedge/src/check/checks/hostname.rs
```diff
@@ -0,0 +1,231 @@
...
+                    "Could not get hostname: gethostname did not return NUL-terminated string",
+                )
+            })?;
+
+            CStr::from_bytes_with_nul_unchecked(&result[..=nul_index])
+                .to_str()
+                .context("Could not get hostname: gethostname returned non-ASCII string")?
+                .to_owned()
+        };
+        self.machine_hostname = Some(machine_hostname.to_owned());
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/checks/host_connect_iothub.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+impl HostConnectIotHub {
+    fn inner_execute(&mut self, check: &mut Check) -> Result<CheckResult, failure::Error> {
+        let iothub_hostname = if let Some(iothub_hostname) = &check.iothub_hostname {
+            iothub_hostname
+        } else {
+            return Ok(CheckResult::Skipped);
+        };
+        self.iothub_hostname = Some(iothub_hostname.to_owned());
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/checks/well_formed_connection_string.rs
```diff
@@ -0,0 +1,57 @@
...
+                    let (_, device, hub) = cs.parse_device_connection_string().context(
+                                "Invalid connection string format detected.\n\
+                                Please check the value of the provisioning.device_connection_string parameter.",
+                )?;
+                    self.device_id = Some(device);
+                    hub
+                }
+                ManualAuthMethod::X509(x509) => x509.iothub_hostname().to_owned(),
+            };
+            check.iothub_hostname = Some(hub.to_owned());
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/checks/windows_host_version.rs
```diff
@@ -0,0 +1,82 @@
...
+        #[cfg(windows)]
+        {
+            let settings = if let Some(settings) = &check.settings {
+                settings
+            } else {
+                return Ok(CheckResult::Skipped);
+            };
+
+            let moby_runtime_uri = settings.moby_runtime().uri().to_string();
+            self.moby_runtime_uri = Some(moby_runtime_uri.to_owned());
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/additional_info.rs
```diff
@@ -184,3 +195,74 @@ pub(super) fn os_version() -> Result<
...
+        let available_space = disk.get_available_space();
+        let total_space = disk.get_total_space();
+        #[allow(clippy::cast_precision_loss)]
+        let percent_free = format!(
+            "{:.1}%",
+            available_space as f64 / total_space as f64 * 100.0
+        );
+
+        DiskInfo {
+            name: disk.get_name().to_str().unwrap().to_owned(),
```
arsing: This will panic if the result of `get_name` is not UTF-8.

`name: String::from_utf8_lossy(disk.get_name()).into_owned(),`

edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs
```diff
@@ -0,0 +1,132 @@
...
+use crate::check::{checker::Checker, Check, CheckResult};
+
+#[derive(Default, serde_derive::Serialize)]
+pub(crate) struct EdgeAgentStorageMounted {}
+impl Checker for EdgeAgentStorageMounted {
+    fn id(&self) -> &'static str {
+        "edge-agent-storage-mounted-from-host"
+    }
+    fn description(&self) -> &'static str {
+        // Note: Keep in sync with Microsoft.Azure.Devices.Edge.Agent.Service.Program.GetStoragePath
```
arsing: Make this and the one below a doc comment of the module.

```
//! Note: Keep this in sync with Microsoft.Azure.Devices.Edge.Agent.Service.Program.GetStoragePath and Microsoft.Azure.Devices.Edge.Hub.Service.DependencyManager.GetStoragePath
```

arsing: Also update the two `GetStoragePath` functions - they each have a comment pointing to `edgelet/iotedge/src/check/mod.rs` that needs to be updated to `edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs`

edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs
```diff
@@ -0,0 +1,132 @@
+use std::path::Path;
+
+use failure::{self, Context, ResultExt};
+use regex::Regex;
+
+use crate::check::{checker::Checker, Check, CheckResult};
+
+#[derive(Default, serde_derive::Serialize)]
+pub(crate) struct EdgeAgentStorageMounted {}
+impl Checker for EdgeAgentStorageMounted {
```
arsing: Missing empty line between `struct` and `impl` block.

edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    let volume_directories = inspect_result
+        .config()
+        .and_then(docker::models::ContainerConfig::volumes)
+        .map(std::collections::HashMap::keys)
+        .into_iter()
+        .flatten()
+        .map(Path::new);
+
+    if !mounted_directories
```
arsing: Add the result of `mounted_directories.chain(volume_directories)` to the additional info. (It's an iterator so you'll need to `collect()` it into a `Vec` first.)

edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs
```diff
@@ -0,0 +1,132 @@
...
+            // Hard-code the value here rather than using the tempfile crate. It needs to match .Net Core's implementation,
+            // and needs to be in the context of the container user instead of the host running `iotedge check`.
+            if cfg!(windows) {
+                r"C:\Windows\Temp"
+            } else {
+                "/tmp"
+            },
+        );
+
+    let storage_directory = Path::new(&*temp_dir).join(storage_directory_name);
```
arsing: Add `storage_directory` to the additional info.

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
```diff
@@ -0,0 +1,156 @@
...
+    fn inner_execute(&mut self, check: &mut Check) -> Result<CheckResult, failure::Error> {
+        let settings = if let Some(settings) = &check.settings {
+            settings
+        } else {
+            return Ok(CheckResult::Skipped);
+        };
+
+        match settings.provisioning() {
+            Provisioning::Dps(dps) => {
+                self.provisioning_mode = Some("Dps");
```
arsing: `"dps"` (lower-case).

Actually, we do want to know what kind of DPS provisioning it is too, so set it to `dps-x509` inside the block at line 44 and `dps-other` otherwise.

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
```diff
@@ -0,0 +1,156 @@
...
+                    let path = x509_info.identity_cert()?;
+
+                    let result =
+                        CertificateValidity::parse("DPS identity certificate".to_owned(), path)?;
+                    self.certificate_info = Some(result.clone());
+                    return result.to_check_result();
+                }
+            }
+            Provisioning::Manual(manual) => {
+                self.provisioning_mode = Some("Manual");
```
arsing: Just like DPS, we want to differentiate between `"manual-x509"` and `"manual-other"` (and both lower-case).

edgelet/iotedge/src/check/checks/identity_certificate_expiry.rs
```diff
@@ -0,0 +1,156 @@
...
+                    let result = CertificateValidity::parse(
+                        "Manual authentication identity certificate".to_owned(),
+                        path,
+                    )?;
+                    self.certificate_info = Some(result.clone());
+                    return result.to_check_result();
+                }
+            }
+            Provisioning::External(_) => {
+                self.provisioning_mode = Some("External");
```
arsing: `"external"` (lower-case).

edgelet/iotedge/src/check/checks/iotedged_version.rs
```diff
@@ -0,0 +1,95 @@
...
+            })?;
+        let version = captures
+            .get(1)
+            .expect("unreachable: regex defines one capturing group")
+            .as_str();
+        self.version = Some(version.to_owned());
+
+        check.additional_info.iotedged_version = Some(version.to_owned());
+
+        if version != latest_versions.iotedged {
```
arsing: Also add `expected_version` to additional info with the value of `latest_versions.iotedged`

edgelet/iotedge/src/check/checks/iotedged_version.rs
arsing: (Still needs to be removed.)

edgelet/iotedge/src/check/checks/iotedged_version.rs
```diff
@@ -0,0 +1,95 @@
+use std::process::Command;
+
+use failure::{self, Context, Fail, ResultExt};
+use regex::Regex;
+
+use crate::check::{checker::Checker, Check, CheckResult};
+
+#[derive(Default, serde_derive::Serialize)]
+pub(crate) struct IotedgedVersion {
+    version: Option<String>,
```
arsing: Rename this to `actual_version` when you add `expected_version` per my comment below.

arsing:
The general intention behind my `.to_owned()` vs `.clone()` comments is:

- For converting `&BorrowType` -> `OwnedType`, use `.to_owned()`. Eg for `&str` to `String`.
- For converting `&OwnedType` -> `OwnedType`, use `.clone()`. Eg for `&String` -> `String`.

`.to_owned()` happens to work when `OwnedType` impls `Deref<Target = BorrowType` because `&OwnedType` auto-derefs to `&BorrowType`, but it helps to differentiate why a particular clone is being made when reading the code later.

edgelet/iotedge/src/check/checks/container_engine_installed.rs
lfitchett: This is a String, not a Cow

edgelet/iotedge/src/check/additional_info.rs
lfitchett: used 
```disk.get_name().to_str().unwrap_or("unrecognized_name").to_owned()```
instead, since from_utf8_lossy doesn't work on OsStr

File: edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs (hunk 48 above)
Comment (lfitchett) [reply edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs]: wasn't sure about a good name for the field. I used other_directories, any suggestions?

File: edgelet/iotedge/src/check/checks/container_engine_installed.rs (hunk 35 above)
Comment (arsing) [reply edgelet/iotedge/src/check/checks/container_engine_installed.rs]: And `.into_owned()` returns a `String`, and is preferable to `.to_owned()` for the reason in my comment.

File: edgelet/iotedge/src/check/additional_info.rs (hunk 45 above)
Comment (arsing) [reply edgelet/iotedge/src/check/additional_info.rs]: `disk.get_name().to_string_lossy().into_owned()` then. Lossy conversion is better than throwing away the value completely.

File: edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs (hunk 48 above)
Comment (arsing) [reply edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs]: `container_directories` (the individual element is called `container_directory` in the `.any()` callback).

File: edgelet/iotedge/src/check/checks/certificates_quickstart.rs [hunk 55]
```diff
@@ -0,0 +1,113 @@
...
+            }
+        });
+        self.device_ca_cert_path = check.device_ca_cert_path.clone();
+
+        if settings.certificates().device_cert().is_none() {
+            let certificate_info = CertificateValidity::parse(
+                "Device CA certificate".to_owned(),
+                check.device_ca_cert_path.to_owned().unwrap(),
+            )?;
+            let not_after = certificate_info.not_after.to_owned();
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/checks/certificates_quickstart.rs
```diff
@@ -0,0 +1,113 @@
...
+                    ))
+                })?
+            }
+        });
+        self.device_ca_cert_path = check.device_ca_cert_path.clone();
+
+        if settings.certificates().device_cert().is_none() {
+            let certificate_info = CertificateValidity::parse(
+                "Device CA certificate".to_owned(),
+                check.device_ca_cert_path.to_owned().unwrap(),
```
arsing: `.to_owned()` -> `.clone()`

edgelet/iotedge/src/check/checks/container_engine_dns.rs
```diff
@@ -0,0 +1,79 @@
...
+        const MESSAGE: &str =
+        "Container engine is not configured with DNS server setting, which may impact connectivity to IoT Hub.\n\
+         Please see https://aka.ms/iotedge-prod-checklist-dns for best practices.\n\
+         You can ignore this warning if you are setting DNS server per module in the Edge deployment.";
+
+        self.container_engine_config_path = Some(
+            check
+                .container_engine_config_path
+                .to_str()
+                .unwrap_or_else(|| "No path found")
```
arsing: `"No path found"` is not right - `to_str()` failing means `check.container_engine_config_path` is not UTF-8.

Anyway this should be `Some(check.container_engine_config_path.to_str_lossy().into_owned())`

edgelet/iotedge/src/check/checks/container_engine_installed.rs
arsing: That said, I missed the `.trim()`, so it's a `&str` and does need `.to_owned()`

edgelet/iotedge/src/check/additional_info.rs
arsing: (Commenting to bring it on top.)

edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs
arsing: (Commenting to bring it on top.)

PR #1836, opened 2019-10-11
k8s-openapi version bump to 0.5.1 + kube-client delete APIs

massand:
- Bumped crate version to 0.5.1 for k8s-openapi and did minor refactor for breaking change
- Added delete kube-client APIs needed for edgelet-kube cleanup logic

### DIFF
--- a/edgelet/kube-client/src/client.rs
+++ b/edgelet/kube-client/src/client.rs
@@ -17,7 +17,7 @@ use k8s_openapi::api::authentication::v1 as api_auth;
 use k8s_openapi::api::core::v1 as api_core;
 use k8s_openapi::api::rbac::v1 as api_rbac;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1 as api_meta;
-use k8s_openapi::{http, Response as K8sResponse, ResponseBody, ListOptional, DeleteOptional};
+use k8s_openapi::{http, DeleteOptional, ListOptional, Response as K8sResponse, ResponseBody};
 use log::debug;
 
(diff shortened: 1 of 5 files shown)


edgelet/kube-client/src/client.rs
```diff
@@ -293,14 +293,42 @@ where
         .flatten()
     }
 
+    pub fn delete_deployment(
+        &mut self,
+        namespace: &str,
+        name: &str,
```
darobs: You may want to add `propagation_policy: Option<&'a str>` as a parameter.  You may also want to add `grace_period_seconds: Option<i64>`  if you think you want to give Agent time to do its work.

massand: We should definitely have more fine tuning done when we validate this E2E, compared to the docker runtime shutdown/termination behavior. Currently, there is no grace period defined for replace deployment/service account either. So, I will refrain from poking the beast that is default values. Besides, I think there is a deployment container configuration in the deployment spec itself called terminationGracePeriodSeconds. I think we should figure out the appropriate values there maybe?

darobs: Yes - I was thinking you are almost certainly going to want the propagation policy, but the grace period may or may not be needed.  There's no requirement to do it now.

massand: This deserves a separate PR because we don't propagate in the replace case as well. 
We should review the [link for cascade deletion](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion) and spin off some work on this.

PR #1837, opened 2019-10-11
new test attribute and basic test set to setup a nightly scenario test pipeline

vipeller:
The important class is 'HighMessageRateDeliveryTests', the rest is just the plumbing to make the test cases concise. Also ScenarioDiscoverer/ScenarioAttribute is for the new test pipeline.

There are scenarios that is hard to test with unit tests, also with end to end tests. In these cases we need to build up some collaborating classes and drive a multistep process, or poke them with a massive amount of data.

These tests may run long, also history shows that they tend to generate false red - so we cannot add them to the normal test run (e.g. check in tests). Also these tests don't require E2E infrastructure, so stress tests are not a good place for them.

Created a new test attribute [Scenario], and a basic test set (to be improved and extended), so I could set up a new pipeline for nightly tests. Currently the tests have [RunnableInDebugOnly] attribute to avoid that the build server accidentally  picks them up and kill the normal test runs.


philipktlin:
I think it is better to have 1 page of high level design and class relationships; which will definitely helps for review.

vipeller:
sure, others also wanted something similar. I sketch up something to show how it works

vipeller:
### Extra high-level overview
The test classes leverage on the ‘builder design pattern’ used many times in testing. That is because these classes are so versatile that it would be inconvenient to setup every little piece on them. The classes have a ‘good default’ and the rest can be configured. A common way to design a ‘builder class’ interface is to use methods starting with the word “with”. Also, these methods return themselves as a return value and this way the ‘with’ calls can be chained:

`Circle.Create().WithColor(red).WithRadius(30).Build()`

We can see two different styles using the builders in the Hub.Services.Test project:
-	For classes part of the Edge Hub infrastructure (implemented in those projects), there is a separate builder class for a given class. For example, RoutingEdgeHub is part of the Edge Hub infrastructure, and it has its own builder class in the Hub.Services.Test project, called EdgeHubBuilder.
-	For classes implemented solely for testing, and defined in the Hub.Services.Test project, they implement the builder pattern on ‘themselves’. This means that the class itself has a functionality and the interface to set it up using the ‘with’ calls. This is not really a builder pattern, but inherits its style.

The reason for this distinct style is to keep the number of classes low. For classes already implemented in a separate project, there is no other way, only to create a new separate builder class. Classes created only for test from zero had the option to embed the builder methods. This causes the asymmetry that separate builder classes need a .Build() call after setting up using the ‘with’ method calls, to create an actual instance, see EdgeHubBuilder class.

Classes embedding the builder methods don’t need a separate Build() class, because the ‘with’ calls set the state of the actual instance, see LinearTimingStrategy class. I was considering to add a transparent, dummy Build() method (doing nothing but returning itself) just to preserve symmetry, but then it gives the fake illusion that a separate instance gets created with the dummy Build() call.

### Using the builder classes

The purpose of the builder classes to make test subject creation easy. To achieve that, builder classes have “good default” behavior. In theory we don’t even need to specify anything, and it just works. Let’s see the following code:

```
public void Test()
{
var device = DeviceBuilder.Create().Build();
_ = device.ProcessDeviceMessageAsync(SimpleDeviceMessageGeneratingStrategy.Create().Next());
}
```

These few lines actually send through a message involving many important parts of edge hub. The message goes through DeviceMessageHandler, RoutingEdgeHub, Router, which has a routing set up with a cloud endpoint, and a dummy cloud proxy which acknowledges the message.

We also can see the asymmetry of the builder classes. DeviceBuilder (which is a separate builder class, creating a DeviceMessageHandler instance) needs a build() call. The SimpleDeviceMessageGeneratingStrategy also implements the ‘With’ calls (we don’t use any of those in this example, though), but it doesn’t need the Build() call, because it sets up itself – and we can call the “Next()” method on the ready to use instance.

### Test Subjects

Although it looks promising that a single line can setup a whole test infrastructure, it is not that convenient than it might seems at first. Firstly, it is hard to find good default values, so most of the times we need to specify more details. Secondly, these default values, and the built infrastructure is ‘good enough’ only for certain test scenarios. What it means is that these test objects were being developed and evolved during the test of specific scenarios, and the default values, the convenient ‘with’ methods, hooks, etc are built to support those specific tests.

Creating a device object as above doesn’t make much sense, it is impossible to check whether the message went through or not. We could do that starting to specify other parameters, hooks, etc, but the device object is not the optimal starting point, because so far no tests were explicitly focused on testing the underlying DeviceMessageHandler class.

There are two major versions of test setups in the existing test-scenarios, and every test object is streamlined to support those. If a new kind of test is needed, the test classes, builder interfaces, default values might need some adjustment to make everything round and comfortable to use. These objects are not designed to solve all problems right now, they are half ready and will go through several refactors later.

The two major scenarios mentioned above are tests from ‘router’ to ‘cloud proxy’ (see HighMessageRateDeliveryRouterToCloudTests.cs for example), and test from DeviceMessageHandler to IClient (see IdleTimeoutRoutingHubToCloudProxyTests for example). For some history, the first set was the narrower router to cloud proxy test, then in a second round a wider ‘device’ to ‘IClient’ tests, involving more collaborating units.

Both test-sets have its own ‘central’ test subject. For Router to Cloud/Module tests, the central object is the Router. For the broader Device to IClient tests, the central object is EdgeHub. The term ‘central’ means, that setting up a test starts from a RouterBuilder or EdgeHubBuilder instance. Look at the following code (from HighMessageRateDeliveryRouterToCloud.cs)

 
image

Although setting up the RouterBuilder doesn’t seem as a ‘central’ point at first look, that is because it is optimized to have ‘good defaults’ for minimal setup. In the background it will create a route with an endpoint, and it can inject the logic to use a “FlakyCloudProxy” for that endpoint. If we had started from an ‘Endpoint’ setup (that we cannot even see in this test, because it is configured and created in the background with default), we would have had a hard time to connect the rest of the test objects to that Endpoint instance, because that is not designed to be a ‘central’ test subject and doesn’t have the convenient default mechanisms and ‘With’ methods. Once somebody wants more configurable Endpoint behavior, because they need to write ‘Endpoint focused’ tests, then they need to add the necessary logic to EndpointBuilder to make it smarter and a good ‘central’ object.

### Deliverables

The example above starts setting up a ‘deliverable’ instance. Most tests want to ‘deliver’ something then check if it has arrived. It can be a device message or a twin update, etc. The ‘Deliverables’ family used to be a simple class, however adding more and more tests, it needed to deliver more and more kind of things and because of that, it got generalized. The main class is in ‘StandardDeliverable.cs’. The base idea is that a test class needs to send ‘something’. It will send this ‘something’ using a specific ‘medium’. Then this ‘something’ gets transformed by the time it arrives into a ‘different something’ (see different implementations of the Message class).

For these three variables, we’ve got TSend, TRcv and TMedium template parameters in StandardDeliverable. TSend can be e.g. an IMessage derived type. We want to send it using something implementing IDeviceListener, which is currently the DeviceMessageHandler class. And by the time the message arrives, it becomes a client SDK ‘Message’ instance. This example is implemented by the DeviceMessageDeliverable class.

Other than the type differences e.g. for sending a device message and sending a twin update, the sending logic is always the same, and the majority of this logic is implemented in StandardDeliverable.cs. The small differences are caught by the IDeliveryStrategy interface (where the ‘Strategy’ word comes for the Strategy design pattern, and not because there are heavy lifting decision making logics implemented behind the interface). Today it bridges four differences: how to identify a deliverable to be sent (GetDeliverableId(TSend)) how to identify a deliverable received (GetDeliverableId(TRcv)), how to ‘tell’ the medium to transfer something (SendDeliverable), and how to generate the deliverable to be sent (DefaultDeliverableGeneratingStrategy, where the ‘default’ word is there because this is configurable with a ‘With’ method on the builder interfaces.

Once a deliverable instance is set up, it can start delivering (using the StartDeliveringAsync() method). It is basically a loop with two major variables: it wants to wait between two sends, and it wants to know if there are more deliverables. These two variables can be affected by helper classes, which pattern goes through almost all test classes in the project.

### Strategy classes

Opening the project in visual studio, there are a whole lot of strategy classes in Hub.Services.Test. They have their own directory in the project and subdivided by other directories. Using strategy patterns is a key ingredient to achieve highly customizable classes, which is a necessity to create a flexible test framework. Looking at the StandardDeliverable implementation, it uses two strategy classes.

When writing a test which sends messages, the timing of the messages can have different nature. Maybe the scenario needs accurate 1 msg / 30 second. Or 10 messages/ minute average, but with some randomness. Or maybe it doesn’t depend on time at all, but on some event. StandardDeliverable class accepts a ‘TimingStrategy’ that is used to decide when to send the next message. In the code example above, it uses ‘NoWaitTimingStrategy’, which means that it will not wait any, but keeps sending messages as fast as it can.

We could use EventBasedTimingStrategy, which is hooked to something like an AutoResetEvent, sending messages in the exact moment we want, e.g when the test code has set up a disconnected scenario and wants to send messages only after disconnected.

The other strategy class-based decision is that how many messages should be sent. It can be a specific number, e.g. 1000. Or it can be connected to other events, e.g. send device messages for five minutes after disconnection. In the former case ‘ConstantSizeVolumeStrategy’ can be used, otherwise ‘ConditionBasedVolumeStrategy’.

For very common scenarios, a test class can use short-cut builder methods. For example the code snippet above writes the following:

```
var deliverable = RoutingMessageDeliverable
                                .Create()
                                .WithPackSize(BigPack)
```

It is clear that it means that this setup will send over a certain number of messages, depending on the ‘BigPack’ constant, which can be e.g. 1000. No ‘VolumeStrategy’ is defined on the deliverable class though, and that is because it uses a short-cut method, which is ‘WithPackSize()’. What it means is actually:

```
var deliverable = RoutingMessageDeliverable
                    .Create()
                    .WithVolumeStrategy<ConstantSizeVolumeStrategy>(
                            volumeStrategy => volumeStrategy.WithPackSize(BigPack))
```

This latter code is the ‘pure’ solution, which sets up a ‘volume strategy’ type, and then configures that with a pack size. In a more complex test setup we can write something similar:

```
var allTwinUpdatesSent = new TestMilestone();
var sendingMessagesTask = DeviceMessageDeliverable
                            .Create()
                            .WithVolumeStrategy<ConditionBasedVolumeStrategy>(
                                    volumeStrategy => volumeStrategy.WithCondition(() => allTwinUpdatesSent.IsPassed))
```

This setup keeps sending device messages until an event, which in case is a ‘TestMilestone’ signaling that all twin updates has been sent through.

### Builder Interface and plumbing

The builder interface gives a convenient interface to setup test scenarios, but it also means a lot of plumbing in the builder classes. A massive amount of code is nothing but a 2-3 liner to provide ‘With’ implementations. There are different kind of ‘With’ implementations, some of them just gets a concrete instance/value and stores it in an internal variable. 

Maybe the most commonly used ‘With’ implementation is the one in the example above, which lets specifying a type by template parameters (e.g. <ConditionBasedVolumeStrategy>) then let that class setup with its own builder interface. 

In the example above, the .WithVolumeStrategy() is a method of the DeviceMessageDeliverable class builder interface, but the volumeStrategy.WithCondition() is a method on the builder interface of the ConditionBasedVolumeStrategy class.

This syntax is made possible by a ‘With’ implementation as the following, which can be seen through the whole code base:
 
image

The first two ‘With’ implementation is not really complex. The trick of the third one is that it receives a function which has the same input parameter type as the result. The type of the input/result is the same as the strategy we want to use. What happens in this code is that it creates a fresh instance according to the type parameter, then it gives this new instance to the function passed as ‘decorator’. This decorator function can be defined as a lambda in the ‘WithThrowTimingStrategy’ method call (or in the WithVolumeStrategy call above in the example), and because its parameter has the type of the strategy we want to set up, now it is possible to call its builder methods.

Note, that the implementation is not consistent across the builder class implementations. It wasn’t a rule to implement every combination, only those needed (or might be useful and it was simple to write that 2 lines real-quick). Also, there are special snowflakes (e.g the decorator function cannot be called immediately, because the strategy class – or test class – would be instantiated later.

### Test doubles

These classes are mostly simpler/smarter imitations of the “frontier”, e.g. CloudProxy. The AllGoodxx versions just confirm back or create a minimum result when get called. The Flakyxxx versions are able to inject exceptions or delays. The classes are not thoroughly implemented. They have the functionality/hooks to support the currently implemented test scenarios. Also, some classes has some example/minimum implementation for scenarios not used today, but might be implemented in the future.

The most complex test double maybe the AllGoodClient, which is a test IClient implementation maintaining some state information (opened/closed). 

Another complex class in this folder is the GenericClientBuilder. This class makes it possible that although IClient instances are created dynamically in many test scenarions, it is still possible to set them up using the Builder pattern. What happens is that GenericClientBuilder works with a similar function than a ‘strategyDecorator’ in the TimingStrategy example above, but instead of using it on a just created instance, it stores the decorator. When ConnectionManager later requests a new connection (and GenericClientProvider is injected to do that, which happens through CloudConnectionProviderBuilder, which is used by ConnectionManagerBuilder), then GenericClientBuilder creates an IClient instance (using its generic type parameter), and then executes the builderDecorator on it, setting up timings, exceptions to throw, etc.

### Scenarios

The actual test-runs are in the folder ‘Scenarios’. These tests are built by the building blocks we have been discussing so far. Although some of the tests are look a bit complex, the general idea is that once a pattern is repeating, then a shortcut ‘with’ method (or another construct) must be built to compact the test setup. These tests are designed to run for several minutes/hours.

### Other

Not all builders are created as general. Especially the complex ones (EdgeHubBuilder), they have ‘hardcoded’ setup, which is not configurable today. Once a test setup requires that, these classes can be changed to a more general version.

philipktlin:
@vipeller, is this PR still valid?  I even can't find HighMessageRateDeliveryRouterToCloud.cs anymore.

PR #1841, opened 2019-10-11
Fix "storage directories are mounted from host" checks to account for mounts of parent directories

arsing:
Previously the checks required that the storage directory itself should be
mounted in the container. However it's possible that a parent directory of
the storage directory was mounted instead. This is common for users that
set the `storageFolder` env var on Edge Agent and/or Edge Hub and then mount
this directory from the host.

For example, the user could set `storageFolder` to `/iotedge/storage` and then
mount the host directory `/var/iotedge/storage` to `/iotedge/storage` in
the container. So the storage directories are `/iotedge/storage/edgeAgent`
and `iotedge/storage/edgeHub` but the mount is for `/iotedge/storage`.


kkdawkins:
:shipit:

PR #1842, opened 2019-10-11
Support dynamic reprovisioning of devices in the external provisioning mode

shantanu1singh:
Support dynamic reprovisioning of devices.

If the dynamic reprovisioning feature is enabled, `iotedged` will simply shut down when a possible reprovisioning event is signaled by `EdgeAgent` (through a call to a new API on the `mgmt` endpoint `/device/reprovision`). The reason behind this is so as to allow a subsequent restart of the `iotedged` to trigger the device provisioning flow again (based on the appropriate provisioning mode) and if the device is provisioned on another hub, `iotedged` will be configured using the new provisioning information of the device.

For the external provisioning mode specifically, the `iotedged` will also notify the external endpoint about a 'possible' device reprovisioning event by calling a new `/device/reprovision` of the external endpoint. 

On receiving such a notification, the external endpoint can check whether the device has indeed been de-provisioned from it's original IoT Hub and provisioned on another IoT Hub instead.
The external endpoint can then take appropriate actions to reconfigure the device such as cleaning up any cached local state and restarting IoT Edge.

More details about the feature can be found in the updates made to the *ExternalProvisioning.md* file.


edgelet/external-provisioning/src/apis/external_provisioning_api.rs
```diff
@@ -110,4 +115,59 @@ where
...
+        req.method(method).uri(uri.unwrap());
+        if let Some(ref user_agent) = configuration.user_agent {
+            req.header(http::header::USER_AGENT, &**user_agent);
+        }
+
+        let req = req
+            .body(hyper::Body::empty())
+            .expect("could not build hyper::Request");
+
+        //        if let Some(ref sas_token) = configuration.sas_token {
```
arsing: Delete this commented-out code.

(And also delete it from `get_device_provisioning_information`)

edgelet/hsm-rs/src/tpm.rs
```diff
@@ -171,7 +171,7 @@ impl SignWithTpm for Tpm {
 
 /// When buffer data is returned from TPM interface, it is placed in this struct.
 /// This is a buffer allocated by the C library.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
arsing: This should not be `Clone`. It does not clone the buffer so there'll be a double-free.

edgelet/iotedged/src/lib.rs
```diff
@@ -1442,25 +1494,64 @@ where
...
+            Ok(_) => Err(None),
+            Err(_) => Err(Some(Error::from(ErrorKind::ManagementService))),
+        })
+        .for_each(move |_x: Option<Error>| Ok(()))
+        .then(|res| match res {
+            Ok(_) | Err(None) => Ok(None),
+            Err(Some(e)) => Err(Some(e)),
+        });
+
+    #[allow(clippy::type_complexity)]
```
arsing: Just `allow` it on the whole crate like the other crates do. It's a silly lint.

edgelet/provisioning/src/provisioning.rs
```diff
@@ -116,19 +126,53 @@ impl From<&str> for ReprovisioningStatus {
...
+    fn from(s: &str) -> ProvisioningStatus {
+        match s {
+            "assigned" => ProvisioningStatus::Assigned,
+            "assigning" => ProvisioningStatus::Assigning,
+            "disabled" => ProvisioningStatus::Disabled,
+            "failed" => ProvisioningStatus::Failed,
+            "unassigned" => ProvisioningStatus::Unassigned,
+            _ => {
+                debug!("Provisioning result status {}", s);
+                ProvisioningStatus::Assigned
```
arsing: Is `Assigned` a safe default? I feel we should fail instead.

edgelet/provisioning/src/provisioning.rs
```diff
@@ -116,19 +126,53 @@ impl From<&str> for ReprovisioningStatus {
...
+                debug!("Provisioning result substatus {}", s);
                 ReprovisioningStatus::InitialAssignment
             }
         }
     }
 }
 
+impl Display for ReprovisioningStatus {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
```
arsing: Why do you want to impl `Display` that forwards to `Debug` instead of using `Debug` directly?

edgelet/iotedged/src/lib.rs
```diff
@@ -392,14 +402,21 @@ where
...
                     ManualAuthMethod::DeviceConnectionString(cs) => {
                         info!("Starting provisioning edge device via manual mode using a device connection string...");
+                        let (key, device_id, hub) = cs
+                            .parse_device_connection_string()
+                            .context(ErrorKind::Initialize(InitializeErrorReason::LoadSettings))?;
+                        let manual = ManualProvisioning::new(key, device_id, hub);
+
                         let (key_store, provisioning_result, root_key) =
-                            manual_provision_connection_string(&cs, &mut tokio_runtime)?;
+                            manual_provision_connection_string(manual.clone(), &mut tokio_runtime)?;
```
arsing: I thought we'd discussed that you could change `Provision::provision` to take `&self` so that you didn't have to `clone()` the `Provision` impls. Did that not work out?

edgelet/provisioning/src/provisioning.rs
shantanu1singh: Changed it to return an error by default

edgelet/iotedged/src/lib.rs
shantanu1singh: Had forgotten about it, changed it now.

edgelet/edgelet-core/src/settings.rs
```diff
@@ -396,10 +396,33 @@ impl External {
...
+    )]
+    dynamic_reprovisioning: Option<bool>,
+}
+
+impl Provisioning {
+    pub fn provisioning_type(&self) -> &ProvisioningType {
+        &self.provisioning
+    }
+
+    pub fn dynamic_reprovisioning(&self) -> Option<&bool> {
```
arsing: `-> Option<bool>` and `{ self.dynamic_provisioning }` since `bool` is `Copy`

edgelet/edgelet-core/src/settings.rs
```diff
@@ -396,10 +396,33 @@ impl External {
...
+#[serde(rename_all = "lowercase")]
+pub struct Provisioning {
+    #[serde(flatten)]
+    provisioning: ProvisioningType,
+
+    #[serde(
+        rename = "dynamic_reprovisioning",
+        skip_serializing_if = "Option::is_none"
+    )]
+    dynamic_reprovisioning: Option<bool>,
```
arsing: Since `None` is treated the same as `Some(false)`, you could do:

```rust
#[serde(default)]
dynamic_reprovisioning: bool,
```

instead.

edgelet/iotedged/src/lib.rs
```diff
@@ -1442,25 +1493,70 @@ where
...
+        .then(|res| match res {
+            Ok(_) | Err(None) => Ok(None),
+            Err(Some(e)) => Err(Some(e)),
+        });
+
+    let mgmt_stop_and_reprovision_signaled =
+        if let Some(&true) = settings.provisioning().dynamic_reprovisioning() {
+            futures::future::Either::B(mgmt_stop_and_reprovision_signaled)
+        } else {
+            futures::future::Either::A(future::empty().map(|_: ()| None))
```
arsing: `.map(|_: ()| None)` -> `.map(|()| None)` , ie use `()` as a pattern.

edgelet/edgelet-http-external-provisioning/src/client/external_provisioning.rs
```diff
@@ -24,13 +24,18 @@ pub trait ExternalProvisioningInterface {
...
+
+    fn reprovision_device(&self) -> Self::ReprovisionDeviceFuture;
 }
 
-pub trait GetApi {
+pub trait GetApi: Send + Sync {
     fn get_api(&self) -> &dyn ExternalProvisioningApi;
 }
 
+#[derive(Clone)]
```
arsing: Does this still need `Clone` ?

edgelet/iotedged/src/lib.rs
arsing: Actually, do you really need the `.map` at all? I would assume the type parameter of `empty()` would infer to `Option` already. Even if it doesn't, you can turbofish it.

edgelet/edgelet-core/src/settings.rs
shantanu1singh: Sweet, thanks for the suggestion!

edgelet/iotedged/src/lib.rs
shantanu1singh: You're right, removed the map.

edgelet/edgelet-http-external-provisioning/src/client/external_provisioning.rs
shantanu1singh: Nope, just missed removing it.

edgelet/provisioning/src/provisioning.rs
```diff
@@ -207,31 +234,29 @@ impl Provision for ManualProvisioning {
...
                 reconfigure: ReprovisioningStatus::DeviceDataNotUpdated,
                 sha256_thumbprint: None,
                 credentials: None,
             })
             .map_err(|err| Error::from(err.context(ErrorKind::Provision)));
         Box::new(result.into_future())
     }
+
+    fn reprovision(&self) -> Box<dyn Future<Item = (), Error = Error> + Send> {
+        Box::new(future::ok(()))
```
arsing: Add a `// No reprovision action is needed for the manual provisioning mode.` comment like you have for DPS.

arsing:
Two minor nits, then I can sign off for the Rust side.

For the C# side, you currently have a TODO that can only be resolved by updating the SDK. So currently Edge Agent won't actually ever invoke the `/device/reprovision`, right? Do you have an ETA for that? (If it's soon enough then we can do it as part of this PR instead of a separate one.)

shantanu1singh:
> For the C# side, you currently have a TODO that can only be resolved by updating the SDK. So currently Edge Agent won't actually ever invoke the `/device/reprovision`, right? Do you have an ETA for that? (If it's soon enough then we can do it as part of this PR instead of a separate one.)

Yes, I'm waiting on [this PR ](https://github.com/Azure/azure-iot-sdk-csharp/pull/1074) to get merged and once this change is available in the 1.21.2 release of the C# SDK, I will uncomment the changes commented out right now.
As of now, the `/device/reprovision` API will never be called.
I was hoping to get this change in first so that I don't need to deal with prospective merge conflicts later and the change later is minimal (involving just the SDK dependency + uncommenting this code)

edgelet/api/managementVersion_2019_01_30.yaml
```diff
@@ -436,7 +436,24 @@ paths:
         default:
           description: Error
           schema:
-            $ref: '#/definitions/ErrorResponse'            
+            $ref: '#/definitions/ErrorResponse'
+
+  '/device/reprovision':
```
ancaantochi: Because this is adding a new API, I think there should be a bump on the version.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/version_2019_01_30/ModuleManagementHttpClient.cs
```diff
@@ -186,6 +186,15 @@ public override async Task PrepareUpdateAsync(ModuleSpec moduleSpec)
             }
         }
 
+        public override async Task ReprovisionDeviceAsync()
```
ancaantochi: Same comment as I added to the managementVersion swagger, it needs a version bump and new generated code for the new version

shantanu1singh: Bumped the version now

edgelet/api/managementVersion_2019_01_30.yaml
shantanu1singh: Bumped the version now

edgelet/edgelet-http-mgmt/src/server/mod.rs
```diff
@@ -64,6 +72,8 @@ impl ManagementService {
             delete  Version2018_06_28 runtime Policy::Module(&*AGENT_NAME)  => "/identities/(?P<name>[^/]+)"        => DeleteIdentity::new(identity.clone()),
 
             get     Version2018_06_28 runtime Policy::Anonymous             => "/systeminfo"                        => GetSystemInfo::new(runtime.clone()),
+
+            post    Version2019_01_30 runtime Policy::Module(&*AGENT_NAME)  => "/device/reprovision"                => ReprovisionDevice::new(initiate_shutdown_and_reprovision),
```
arsing: This needs to use the new `Version`

shantanu1singh: Good catch, updated!

PR #1845, opened 2019-10-11
Changed CloudEndpoint to fail subsequent batches after a transient error

vipeller:
The scenario it fixes is that when multiple batches are being sent out, and one fails with retriable error, then that batch will be sent later again. If we don't fail subsequent batches being sent in the same time, then the failing batch will be sent after those subsequent batches, changing the original message order.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/routing/CloudEndpoint.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/routing/CloudEndpoint.cs
@@ -100,7 +100,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Routing
             public Task CloseAsync(CancellationToken token) => Task.CompletedTask;
 
             internal static int GetBatchSize(int batchSize, long messageSize) =>
-                Math.Min((int)(Constants.MaxMessageSize / messageSize), batchSize);
+                Math.Min((int)(Constants.MaxMessageSize / Math.Max(1, messageSize)), batchSize);
 
             static bool IsRetryable(Exception ex) => ex != null && RetryableExceptions.Any(re => re.IsInstanceOfType(ex));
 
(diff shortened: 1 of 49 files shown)


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/routing/CloudEndpoint.cs
```diff
@@ -100,7 +100,7 @@ public Task<ISinkResult> ProcessAsync(ICollection<IRoutingMessage> routingMessag
             public Task CloseAsync(CancellationToken token) => Task.CompletedTask;
 
             internal static int GetBatchSize(int batchSize, long messageSize) =>
-                Math.Min((int)(Constants.MaxMessageSize / messageSize), batchSize);
+                Math.Min((int)(Constants.MaxMessageSize / Math.Max(1, messageSize)), batchSize);
```
myagley: Oooh, good catch.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/routing/CloudEndpoint.cs
```diff
@@ -136,103 +136,54 @@ static ISinkResult GetSyncResultForInvalidMessages(Exception ex, List<IRoutingMe
...
-                        failed.AddRange(res.Failed);
-                        invalid.AddRange(res.InvalidDetailsList);
-                        // Different branches could have different results, but only the most significant will be reported
-                        if (IsMoreSignificant(sendFailureDetails, res.SendFailureDetails))
-                        {
-                            sendFailureDetails = res.SendFailureDetails;
-                        }
-                    }
+
+                    foreach (var res in sinkResults)
```
myagley: We should probably put braces around the body of this for loop.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/SinkResult.cs
```diff
@@ -44,8 +45,62 @@ public SinkResult(ICollection<T> succeeded, ICollection<T> failed, ICollection<I
...
+        }
+
+        public void Merge(ISinkResult<T> other)
+        {
+            this.Succeeded.AddRange(other.Succeeded);
+            this.Failed.AddRange(other.Failed);
+            this.InvalidDetailsList.AddRange(other.InvalidDetailsList);
+
+            if (IsMoreSignificant(this.SendFailureDetails, other.SendFailureDetails))
+                this.SendFailureDetails = other.SendFailureDetails;
```
myagley: We should probably add braces around this if statement.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/CollectionEx.cs
```diff
@@ -143,6 +143,9 @@ public static void AddIfNonEmpty(this IDictionary<string, string> dictionary, st
 
         public static IEnumerable<IEnumerable<T>> Batch<T>(this IEnumerable<T> list, int batchSize)
         {
+            if (batchSize == 0)
+                throw new ArgumentException("BatchSize cannot be zero", nameof(batchSize));
```
myagley: Ditto on braces.

PR #1848, opened 2019-10-14
port automated mount for stress / longhaul

and-rewsmith:


varunpuranik:
Does this PR always mount the volume into the containers? Should we just provide a option to do that instead? So that it can only be enabled for the e2e tests where it makes sense?

scripts/windows/test/Run-E2ETest.ps1
```diff
@@ -1324,6 +1324,15 @@ Function SetEnvironmentVariable
     $env:Path="$env:Path;C:\Program Files\iotedge-moby;C:\Program Files\iotedge"
 }
 
+Function SetupMountedStorage
```
varunpuranik: Here you don't seem to delete the mount folder if it exists.

scripts/linux/runE2ETest.sh
```diff
@@ -19,6 +19,9 @@ function clean_up() {
     rm -rf /var/run/iotedge/
     rm -rf /etc/iotedge/config.yaml
 
+    echo 'Remove mounted storage folder'
+    rm -rf $mounted_storage_path
```
varunpuranik: Suggest moving this to setup_mounted_storage

scripts/windows/test/Run-E2ETest.ps1
and-rewsmith: The Setup-Env.ps1 should handle that part

and-rewsmith: But we should probably keep all the logic in one place, so I will move

scripts/linux/runE2ETest.sh
```diff
@@ -852,9 +855,17 @@ function run_test()
     exit $ret
 }
 
+function setup_mounted_storage() {
```
and-rewsmith: need to fix this brace

and-rewsmith: Actually, same line is the standard here. run_test() is the weird one

and-rewsmith:
The mount will only happen for stress and long haul, as it is specified as a bind in the deployment file through CreateOptions. The option is a good idea so other e2e tests can take advantage of the mounts. Their deployment files will need to change though if they want to do so.

and-rewsmith:
Closing as we should make configurable.

PR #1854, opened 2019-10-15
[k8s] Allow deployments to delete dependencies in the background.

darobs:
Prevent EdgeDeploymentController from removing EdgeAgent's stuff.

E2E test.

1. install edge runtime into a k8s cluster
N=<some namespace>
helm install --name edgy --namespace $N --set "deviceConnectionString=<connection string>" -f ~/yaml/override.yaml ./edge-kubernetes

2. Update the device's deployment to add a module.

3. kubectl get all -n $N

4. Update the device's deployment to remove a module.

5. kubectl get all -n $N
- confirm that the deployment, replicaset, pod, service (if any) related to the deleted module are deleted.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -180,11 +198,11 @@ async Task ManageDeployments(V1DeploymentList existing, IEnumerable<V1Deployment
                     name =>
                     {
                         Events.DeleteDeployment(name);
-                        return this.client.DeleteNamespacedDeployment1Async(
+                        return this.client.DeleteNamespacedDeploymentAsync(
                             name,
                             this.deviceNamespace,
-                            propagationPolicy: "Foreground",
-                            body: new V1DeleteOptions(propagationPolicy: "Foreground"));
+                            propagationPolicy: KubernetesConstants.DefaultPropagationPolicy,
```
CindyXing: Are we changing all unsuccessful deployment to be deleted in the background? If so, can you help understand why?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -357,11 +380,11 @@ public async Task PurgeModulesAsync()
             V1DeploymentList deployments = await this.client.ListNamespacedDeploymentAsync(this.deviceNamespace, labelSelector: this.deploymentSelector);
             var deploymentTasks = deployments.Items
                 .Select(
-                    deployment => this.client.DeleteNamespacedDeployment1Async(
+                    deployment => this.client.DeleteNamespacedDeploymentAsync(
                         deployment.Metadata.Name,
                         this.deviceNamespace,
-                        new V1DeleteOptions(propagationPolicy: "Foreground"),
-                        propagationPolicy: "Foreground"));
+                        new V1DeleteOptions(propagationPolicy: KubernetesConstants.DefaultPropagationPolicy),
```
CindyXing: Would like to understand the background on the change.

massand:
I just went ahead and built the agent and container image myself to test this. Here is what I would expect you did (using either local or CI build) - 
scripts/linux/buildBranch.sh -c Debug --no-rocksdb-bin
scripts/linux/buildImage.sh -P Microsoft.Azure.Devices.Edge.Agent.Service  -i azureiotedge-agent -v 0.5 -r <redact>.azurecr.io --bin-dir "$(pwd)/iotedge-build" -u <redact> -p <redact>
helm install --name k8s-edge1 local/edge-kubernetes-crd --debug -n ms-iot
helm install --name k8s-edge2 local/edge-kubernetes --debug -f override_with_new_agent_image.yaml --namespace ms-iot

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -151,11 +153,27 @@ async Task ManageServices(V1ServiceList existing, IEnumerable<V1Service> desired
         static Diff<V1Service> FindServiceDiff(IEnumerable<V1Service> desired, IEnumerable<V1Service> existing)
         {
             var desiredSet = new Set<V1Service>(desired.ToDictionary(service => service.Metadata.Name));
-            var existingSet = new Set<V1Service>(existing.ToDictionary(service => service.Metadata.Name));
+            // Remove Agent-owned assets from existing list
```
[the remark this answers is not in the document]
darobs: Oh - this comment needs to be removed...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
darobs: These would be deployments being deleted because the module was removed from the Edge Deployment, or the entire Edge Deployment .  When we delete the module we're considering switching from "Foreground" delete propagation policy to "Background" for Deployment deletions.  Both should have the same net effect, but with background deletion, the API should respond more quickly.

We were a bit concerned that if the API call times out (modules with volumes seems to take a long time to delete), we might leave orphaned objects.  Also, we aren't waiting as long on deployments, so the operation takes less time.

darobs:
> I just went ahead and built the agent and container image myself to test this. Here is what I would expect you did (using either local or CI build) -

I added a description of the E2E test at the top.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -357,11 +380,11 @@ public async Task PurgeModulesAsync()
...
+                    deployment => this.client.DeleteNamespacedDeploymentAsync(
                         deployment.Metadata.Name,
                         this.deviceNamespace,
-                        new V1DeleteOptions(propagationPolicy: "Foreground"),
-                        propagationPolicy: "Foreground"));
+                        new V1DeleteOptions(propagationPolicy: KubernetesConstants.DefaultPropagationPolicy),
+                        propagationPolicy: KubernetesConstants.DefaultPropagationPolicy));
             await Task.WhenAll(deploymentTasks);
 
             V1PersistentVolumeClaimList pvcs = await this.client.ListNamespacedPersistentVolumeClaimAsync(this.deviceNamespace, labelSelector: this.deploymentSelector);
```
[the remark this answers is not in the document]
darobs: Handling Edge Agent was moved to a different PR (and implemented differently.)

PR #1855, opened 2019-10-15
[k8s] Prevent EdgeDeploymentController from removing EdgeAgent's stuff.

darobs:
Prevent EdgeDeploymentController from removing EdgeAgent's stuff.


darobs:
Dropped this in favor of another approach.

PR #1857, opened 2019-10-15
[k8s] Treat empty auth config section as no section at all

dmolokanov:
-  Remove unused use_pvc config parameter
- Treat empty auth config section as no section at all
- Refactored big to_k8s module to make image pull secret conversion clean and simple
- Remove edge agent `Mode` env duplication


edgelet/edgelet-kube/src/registry/image_pull_secret.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    #[test]
+    fn it_returns_none_secret_name_when_registry_missing() {
+        let image_pull_secret = ImagePullSecret::default().with_username("USER");
+
+        let name = image_pull_secret.name();
+
+        assert_eq!(name, None);
+    }
```
darobs: Duplicate of test above?

dmolokanov: yep

PR #1866, opened 2019-10-17
1.0.8 - Port support of additional consumer groups for test environments

and-rewsmith:
**This should not be merged prior to 1.0.8.3 release.**


and-rewsmith:
1.0.8.3 has been released, so this should now be merged.

and-rewsmith:
FYI there are no iotuap deployments set up for 108, so the consumer group fix we need to make in master does not apply here.

philipktlin:
Please also bring latest changes from iotedge\builds\e2e\templates\deploy-iotuap.yaml

PR #1871, opened 2019-10-18
fix additional minor bugs found while testing

seanmcd-msft:
several folks have been confused by vcruntime being missing in iot enterprise.  
I have seen a weird very intermittent issue with uninstall failing on enterprise so I added more diagnostic info to try to get something actionable if I see it again.


scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -921,7 +933,7 @@ function Install-Packages(
 
         if ((Test-MobyNeedsToBeMoved) -or (Test-LegacyInstaller)) {
             Write-HostRed
-            Write-HostRed ('IoT Edge is installed in an invalid location. ' + $ReinstallMessage)
+            Write-HostRed ('IoT Edge or the IoT Edge Moby Engine is installed in an invalid location. Probably there is an old preview install present and it needs to be uninstalled or the system should be re-imaged.' + $ReinstallMessage)
```
arsing: Space is missing from the end of the string literal.

`re-imaged.' + $ReinstallMessage` -> `re-imaged. ' + $ReinstallMessage`

Same for the other instances of this.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1564,13 +1575,16 @@ function Reset-SystemPath {
     Write-Verbose 'Removed IoT Edge directories from system PATH'
 }
 
+function Test-VcRuntimePresent {
+    return Test-Path 'C:\Windows\System32\vcruntime140.dll'
+}
```
arsing: Missing empty line between the end of this function and the start of the next.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1629,10 +1643,10 @@ function Uninstall-Services([ref] $RestartNeeded, [bool] $LegacyInstaller) {
         Stop-Service -NoWait -ErrorAction SilentlyContinue -ErrorVariable cmdErr $MobyServiceName
         if ($?) {
             Start-Sleep -Seconds 7
-            Write-Verbose 'Stopped the IoT Edge Moby Engine service'
+            Write-Verbose "'Stopped the IoT Edge Moby Engine service $MobyServiceName"
```
arsing: `"'` -> `"`

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1614,12 +1628,12 @@ function Uninstall-Services([ref] $RestartNeeded, [bool] $LegacyInstaller) {
     if (Get-Service $EdgeServiceName -ErrorAction SilentlyContinue) {
         Set-Service -StartupType Disabled $EdgeServiceName -ErrorAction SilentlyContinue
         Stop-Service -NoWait -ErrorAction SilentlyContinue -ErrorVariable cmdErr $EdgeServiceName
-        if ($?) {
+        if ($? -and $LASTEXITCODE -eq 0) {
```
arsing: `$LASTEXITCODE` is not set by PS commandlets. This will read an arbitrary stale value so it can fail spuriously.

```powershell
$ false; gci | Out-Null; echo $LASTEXITCODE

1
```

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1614,12 +1628,12 @@ function Uninstall-Services([ref] $RestartNeeded, [bool] $LegacyInstaller) {
...
             Start-Sleep -Seconds 7
-            Write-Verbose 'Stopped the IoT Edge service'
+            Write-Verbose "Stopped the IoT Edge service $EdgeServiceName"
         }
         else {
-            Write-Verbose "$cmdErr"
+            Write-Verbose "stopping IoT Edge service $EdgeServiceName failed.  exitcode: $LASTEXITCODE error: $cmdErr"
```
arsing: Same here, `$LASTEXITCODE` is meaningless for `Stop-Service`

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1629,10 +1643,10 @@ function Uninstall-Services([ref] $RestartNeeded, [bool] $LegacyInstaller) {
         Stop-Service -NoWait -ErrorAction SilentlyContinue -ErrorVariable cmdErr $MobyServiceName
         if ($?) {
             Start-Sleep -Seconds 7
-            Write-Verbose 'Stopped the IoT Edge Moby Engine service'
+            Write-Verbose "'Stopped the IoT Edge Moby Engine service $MobyServiceName"
         }
         else {
-            Write-Verbose "$cmdErr"
+            Write-Verbose "stopping IoT Edge Moby Engine service $MobyServiceName failed.  exitcode: $LASTEXITCODE error: $cmdErr"
```
arsing: Same here, `$LASTEXITCODE` is meaningless for `Stop-Service`

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1649,6 +1663,7 @@ function Uninstall-Services([ref] $RestartNeeded, [bool] $LegacyInstaller) {
         }
     }
     else {
+        Write-Verbose "uninstalling iot edge package"
```
arsing: `Uninstalling IoT Edge package`

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -307,9 +307,15 @@ function Initialize-IoTEdge {
         throw
     }
 
+    if (-not (Test-VcRuntimePresent)) {
+        Write-HostRed 'VC Runtime must be installed before IoT Edge can be initialized.'
+        throw
+    }
+
```
arsing: Change two empty lines to one.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -307,9 +307,15 @@ function Initialize-IoTEdge {
...
+
     if ((Test-MobyNeedsToBeMoved) -or (Test-LegacyInstaller)) {
         Write-HostRed
-        Write-HostRed ('IoT Edge is installed in an invalid location. ' + $ReinstallMessage)
+        Write-HostRed ('IoT Edge or the IoT Edge Moby Engine is installed in an invalid location. Probably there is an old preview install present and it needs to be uninstalled or the system should be re-imaged.' + $ReinstallMessage)
```
arsing: `There may be an old preview install of IoT Edge still present. Please run Uninstall-IoTEdge first or re-image the device.`

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: fixed

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: fixed

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: fixed

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: right. for some reason I was thinking this script was using an internal function that invoked sc.exe. i'm still puzzled by why I've seen at least 3 times the stop-service succeed but the service was not actually stopped or even in 'stopping' state.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: fixed per above.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: fixed per above

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: fixed

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: fixed

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
seanmcd-msft: fixed

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
arsing: One reason I know of is that if the service crashes, it gets restarted due to the failure recovery policy, and this takes precedence over any pending stop request. I've seen it happen on my VM a bunch of times. I work around it by doing a `Set-Service -StartupType Disabled` first.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
arsing: Not yet resolved.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1575,13 +1585,17 @@ function Reset-SystemPath {
     Write-Verbose 'Removed IoT Edge directories from system PATH'
 }
 
+function Test-VcRuntimePresent {
+    return Test-Path 'C:\Windows\System32\vcruntime140.dll'
```
arsing: Is this correct when running in 32-bit PS? Won't it redirect to `C:\Windows\SysWoW64` ?

seanmcd-msft: I suppose.  that could only occur on Enterprise and i'm not sure why anybody would run 32-bit PS to install a 64 bit product.  but, I guess it's possible.  however, that path test was what was there before I just moved it into a separate named function to make the intent of the path clear.

arsing: >i'm not sure why anybody would run 32-bit PS to install a 64 bit product. but, I guess it's possible.

Me neither, but apparently there was enough reason that we added support for it intentionally https://github.com/Azure/iotedge/pull/1293 . I can't remember why, though. Oh well.


>however, that path test was what was there before I just moved it into a separate named function to make the intent of the path clear.

Oh yes, I missed that it was already using System32 before. Never mind then.

PR #1876, opened 2019-10-18
Update crates hyper, tokio, h2

shantanu1singh:
Update crates hyper, tokio, h2.

The older version of hyper was missing a bugfix for IPv6 hostname parsing when the address was enclosed within the `[]` characters. Has been fixed in 0.12.22

### DIFF
--- a/THIRDPARTYNOTICES
+++ b/THIRDPARTYNOTICES
@@ -620,54 +620,33 @@ DEALINGS IN THE SOFTWARE.
 ***
 
 tokio-rs/tokio 
-Copyright (c) 2018 Tokio Contributors
 
+MIT License
 
+Copyright (c) 2019 Tokio Contributors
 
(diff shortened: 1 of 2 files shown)


THIRDPARTYNOTICES
```diff
@@ -4282,26 +4261,6 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 ***
 
-crossbeam-rs/crossbeam-epoch
```
arsing: Another version of each of these two is still in the lockfile, so don't remove them from TPN.

arsing:
Clarify the reason in the commit message please ("The older version of hyper is missing a bugfix for ... that was fixed in 0.12.22")

PR #1888, opened 2019-10-22
Bundle docker network inspect and docker daemon logs

lfitchett:
Adds the docker network inspect for each network and docker daemon logs


edgelet/iotedge/src/support_bundle.rs
```diff
@@ -234,6 +248,61 @@ where
...
+            .arg("--no-pager")
+            .output();
+
+        /* from https://docs.microsoft.com/en-us/virtualization/windowscontainers/troubleshooting#finding-logs */
+        #[cfg(windows)]
+        let inspect = ShellCommand::new("powershell.exe")
+            .arg("-NoProfile")
+            .arg("-Command")
+            .arg(&format!(
+                r#"Get-EventLog -LogName Application -Source Docker -After "{}" |
```
arsing: Are you sure the source is `Docker` ? The URL in your comment is about Docker CE. For IoT Edge we install our own Moby engine and its service name is `iotedge-moby`

Also this uses `Get-EventLog` while `write_edgelet_log_to_file` uses `Get-WinEvent`. Is there a reason not to use `Get-WinEvent` ?

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -313,6 +382,89 @@ where
...
+            .start_file_from_path(&Path::new(&file_name), state.file_options)
+            .map_err(|err| Error::from(err.context(ErrorKind::SupportBundle)))?;
+
+        state
+            .zip_writer
+            .write(&output)
+            .map_err(|err| Error::from(err.context(ErrorKind::SupportBundle)))?;
+
+        state.print_verbose(&format!("Got docker network inspect for {}", network_name));
+        drop(network_name);
```
arsing: Why do you need the `drop` ?

Remove it from `write_inspect_to_file` too.

edgelet/iotedge/src/support_bundle.rs
lfitchett: I tested this with moby in a vm. It collects the logs, although they aren't very useful unless docker is set to debug level logging. 

I couldn't get Get-WinEvent to pull the logs. I tried ```Get-WinEvent -FilterHashTable @{ProviderName=Docker}```, but it doesn't recognize docker as an event provider

edgelet/iotedge/src/support_bundle.rs
lfitchett: A while ago clippy complained about not consuming the String. It doesn't now, so I'll remove it

edgelet/iotedge/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-#![deny(rust_2018_idioms, warnings)]
-#![deny(clippy::all, clippy::pedantic)]
+// #![deny(rust_2018_idioms, warnings)]
```
arsing: Revert.

PR #1903, opened 2019-10-24
Add location to support bundle output

lfitchett:
Changing output to include location per venkat's request

### DIFF
--- a/edgelet/iotedge/src/support_bundle.rs
+++ b/edgelet/iotedge/src/support_bundle.rs
@@ -54,7 +54,12 @@ where
             .and_then(SupportBundle::write_check_to_file)
             .and_then(SupportBundle::write_all_inspects)
             .and_then(SupportBundle::write_all_network_inspects)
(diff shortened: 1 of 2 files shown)


edgelet/iotedge/src/support_bundle.rs
```diff
@@ -45,14 +46,15 @@ where
...
+        println!("Making support bundle");
         let result = future::result(self.make_state())
             .and_then(SupportBundle::write_all_logs)
             .and_then(SupportBundle::write_edgelet_log_to_file)
             .and_then(SupportBundle::write_docker_log_to_file)
             .and_then(SupportBundle::write_check_to_file)
             .and_then(SupportBundle::write_all_inspects)
             .and_then(SupportBundle::write_all_network_inspects)
-            .map(|state| state.print_verbose("Created support bundle"));
+            .map(|state| println!("Created support bundle at {:#?}", state.location));
```
arsing: `println!("Created support bundle at {}", state.location.to_string_lossy())`

PR #1904, opened 2019-10-24
Fix test results

philipktlin:
Update script to output to different test result files; which will be published later on.


scripts/linux/runTests.sh
```diff
@@ -73,9 +73,12 @@ then
   testCommandPrefix+=" /TestCaseFilter:"$testFilterValue""
 fi
 
+count=0
```
dmolokanov: Could we possibly use assembly name instead counter?

PR #1909, opened 2019-10-24
Adding ForEach overload with None method to Option class

dylanbronson:
Adding MatchNoReturn method to Option class. 
I recently encountered a need to use the Match method, but without it returning anything. There was no perfect method for this. This PR adds that method. 

It also adds tests for the method and tests that were missing for the Match method.


edge-util/src/Microsoft.Azure.Devices.Edge.Util/Option.cs
```diff
@@ -127,6 +127,19 @@ public T Expect<TException>(Func<TException> exception)
         [Pure]
         public TResult Match<TResult>(Func<T, TResult> some, Func<TResult> none) => this.HasValue ? some(this.Value) : none();
 
+        [Pure]
+        public void MatchNoReturn(Action<T> some, Action none)
```
darobs: I wish this was more like Rust, and we could just do `option.map(some).or_else(none)`.
Would this alternate make sense to people?
```
        public void ForEach(Action<T> action, Action none = null)
        {
            if (this.HasValue)
            {
                action(this.Value);
            }
            else
            {
                none?.Invoke();
            }
        }
```

darobs: `ForEachAsync`  could be:
```
public Task ForEachAsync(Func<T, Task> action, Func<Task> none = null) => this.HasValue ? action(this.Value) : none?.Invoke();
```

varunpuranik: I also like adding an overload to ForEach better...

dylanbronson: Updated to overload ForEach

edge-util/test/Microsoft.Azure.Devices.Edge.Util.Test/OptionTest.cs
```diff
@@ -260,6 +279,41 @@ public async void TestForEachAsync()
...
+                return Task.CompletedTask;
+            },
+            () =>
+            {
+                val = -2;
+                return Task.CompletedTask;
+            });
+            Assert.Equal(-2, val);
+        }
+
```
vipeller: In theory the actions run on a different thread (e.g. val += b) that Assert.Equal() so the variables should be written/read using some synchronization mechanism (e.g volatile.read/write) Now it works only because 'await' optimizes and when it sees that the returned task is completed, it continues on the same thread. (also for the same reason, I'd put an 'await Task.Yield()' to the beginning of the actions to make it more realistic and test an actual async scenario.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/Option.cs
```diff
@@ -140,8 +140,22 @@ public void ForEach(Action<T> action)
...
 
+        public void ForEach(Action<T> action, Action none = null)
+        {
+            if (this.HasValue)
+            {
+                action(this.Value);
+            }
+            else
+            {
+                none?.Invoke();
```
varunpuranik: Hmm, the idea of Option is to avoid nulls.. so maybe we should avoid one here. :) Separate overload maybe?

darobs: Yes, after re-reviewing this my suggestion is a bit ironic. I like `ForEach` syntax better, but if an overload function works, that's how it should be implemented.

edge-util/test/Microsoft.Azure.Devices.Edge.Util.Test/OptionTest.cs
dylanbronson: Viktor ran these tests a large number of times, and there were no failures, so I will not change.

PR #1915, opened 2019-10-25
Long Haul Module Restarter

and-rewsmith:
This PR adds randomized module restarting to our long haul tests.

You can configure through the vsts pipeline:
- restart interval
- desired modules to restart


e2e_deployment_files/long_haul_deployment.template.json
```diff
@@ -279,6 +279,29 @@
...
             }
+          },
+          "moduleRestarter": {
+            "version": "1.0",
+            "type": "docker",
+            "status": "running",
+            "restartPolicy": "always",
+            "env": {
+              "ServiceClientConnectionString": {
+                "value": "<ServiceClientConnectionString>"
```
philipktlin: I believe you can parse service client connection string to get device Id, please check.

e2e_deployment_files/long_haul_deployment.template.windows.json
```diff
@@ -282,6 +282,29 @@
               "image": "<Container_Registry>/microsoft/azureiotedge-direct-method-receiver:<Build.BuildNumber>-windows-<Architecture>",
               "createOptions": ""
             }
+          },
+          "moduleRestarter": {
+            "version": "1.12",
```
philipktlin: difference from long_haul_deployment.template.json; should it be 1.0?
```suggestion
            "version": "1.0",
```

e2e_deployment_files/long_haul_deployment.template.json
and-rewsmith: I'm going to use the env var IOTEDGE_DEVICEID. I'm guessing we should make this change to the analyzer as well?

edge-modules/ModuleRestarter/ModuleRestarter.csproj
```diff
@@ -0,0 +1,65 @@
...
+
+  <PropertyGroup Condition="'$(OS)|$(DotNet_Runtime)' == 'Unix|netcoreapp3.0'">
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+    <Configurations>Debug;Release;CodeCoverage;CheckInBuild</Configurations>
+    <HighEntropyVA>true</HighEntropyVA>
```
philipktlin: I remember Varun used this for a project (forget which project); but I forget why.  Do we really need this?

edge-modules/ModuleRestarter/ModuleRestarter.csproj
```diff
@@ -0,0 +1,65 @@
...
+    <OutputPath>bin\CodeCoverage</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include="docker*/**/*.*" CopyToPublishDirectory="Always" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.*" />
```
philipktlin: you should use an explicit version and same version as other projects.

edge-modules/ModuleRestarter/ModuleRestarter.csproj
and-rewsmith: I am not sure. I will try removing it, but it exists in all other csproj files under edge-modules as well.

edge-modules/ModuleRestarter/docker/linux/arm32v7/base/Dockerfile
```diff
@@ -0,0 +1,6 @@
+ARG base_tag=2.1.10-bionic-arm32v7
```
philipktlin: Remove this file; this Linux ARM32 base image is same across all the modules; which you can find the docker file under  edge-modules\SimulatedTemperatureSensor\docker\linux\arm32v7\base folder.

edge-modules/ModuleRestarter/docker/linux/arm64v8/base/Dockerfile
```diff
@@ -0,0 +1,6 @@
+ARG base_tag=3.0.0-preview5-bionic-arm64v8
```
philipktlin: Remove this file; same reason as ARM32 base image docker file.

edge-modules/ModuleRestarter/docker/windows/arm32v7/base/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Base image for arm32v7 dotnet-runtime container
```
philipktlin: remove this file; same reason as linux arm32 docker base image file.

edge-modules/ModuleRestarter/ModuleRestarter.csproj
and-rewsmith: [The documentation](https://docs.microsoft.com/en-us/cpp/build/reference/highentropyva?view=vs-2019) says that this indicates support for ASLR with 64-bit addresses, which I'm guessing we support or it would not be elsewhere. Guessing we should keep it?

and-rewsmith: [This documentation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/highentropyva-compiler-option) says:

The highentropyva compiler option tells the Windows kernel whether a particular executable supports high entropy Address Space Layout Randomization (ASLR).

If it is everywhere else I'm guessing we should keep it.

e2e_deployment_files/long_haul_deployment.template.windows.json
and-rewsmith: Good catch!

edge-modules/ModuleRestarter/ModuleRestarter.csproj
philipktlin: Please check with Anca if she knows what this parameter is used for.  And btw you are right we used this parameter in many projects except loadGen and MessageAnalyzer, please make it consistent if needed.  thanks.

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+        public static int Main() => MainAsync().Result;
+
+        static async Task<int> MainAsync()
+        {
+            Log.LogInformation($"Starting module restarter with the following settings:\r\n{Settings.Current}");
+
+            if (string.IsNullOrEmpty(DesiredModulesToRestartCSV))
+            {
+                Log.LogInformation("No modules specified to restart. Stopping.");
+                return 0;
```
philipktlin: if the program exits, I believe the docker container will exit as well. Will this be forced to restart by edgeAgent as restart policy is set to always?

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+        static readonly string DeviceId = Settings.Current.DeviceId;
+        static readonly string DesiredModulesToRestartCSV = Settings.Current.DesiredModulesToRestartCSV;
+        static readonly int RandomRestartIntervalInMins = Settings.Current.RandomRestartIntervalInMins;
+
+        public static int Main() => MainAsync().Result;
+
+        static async Task<int> MainAsync()
+        {
+            Log.LogInformation($"Starting module restarter with the following settings:\r\n{Settings.Current}");
+
```
philipktlin: validate ServiceClientConnectionString is not empty?

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (string.IsNullOrEmpty(DesiredModulesToRestartCSV))
```
philipktlin:
```suggestion
            if (string.IsNullOrWhiteSpace(DesiredModulesToRestartCSV))
```

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                Log.LogInformation("No modules specified to restart. Stopping.");
+                return 0;
+            }
+
+            (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromMinutes(2 * RandomRestartIntervalInMins), null);
```
philipktlin: why shutdown wait period is set to 2 * RandomRestartIntervalInMins?  and why not provide logger (2nd argument)?

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            await RestartModules(cts, RandomRestartIntervalInMins);
+            completed.Set();
+
+            return 0;
```
philipktlin: you don't need this line, by default it will return 0.

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                Log.LogInformation("No modules specified to restart. Stopping.");
+                return 0;
+            }
+
+            (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromMinutes(2 * RandomRestartIntervalInMins), null);
+
+            await RestartModules(cts, RandomRestartIntervalInMins);
+            completed.Set();
+
```
philipktlin: you need to call below method for cleanup, please check out other modules, e.g. DirectMethodCloudSender.
```suggestion
handler.ForEach(h => GC.KeepAlive(h));
```

edge-modules/ModuleRestarter/src/Program.cs
philipktlin: Even safer, you should parse here to ensure that you got non-empty list of module names (what happen if ",,," is given).

edge-modules/ModuleRestarter/src/Program.cs
philipktlin: You need to think about what behavior you expected.  For example, you can still run and check every X period if the environment variable may be updated.

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+            while (!cts.Token.IsCancellationRequested)
+            {
+                string payload = "{{ \"SchemaVersion\": \"1.0\", \"Id\": \"{0}\" }}";
+                payload = string.Format(payload, moduleNames[random.Next(0, moduleNames.Length)]);
+                Log.LogInformation("RestartModule Method Payload: {0}", payload);
+                c2dMethod.SetPayloadJson(payload);
+
+                await iotHubServiceClient.InvokeDeviceMethodAsync(DeviceId, "$edgeAgent", c2dMethod);
+
+                Thread.Sleep(RandomRestartIntervalInMins * 60 * 1000);
```
philipktlin: use "await Task.Delay"

edge-modules/ModuleRestarter/src/Program.cs
philipktlin: is it ok to restart edgeAgent?

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+            Random random = new Random();
+
+            while (!cts.Token.IsCancellationRequested)
+            {
+                string payload = "{{ \"SchemaVersion\": \"1.0\", \"Id\": \"{0}\" }}";
+                payload = string.Format(payload, moduleNames[random.Next(0, moduleNames.Length)]);
+                Log.LogInformation("RestartModule Method Payload: {0}", payload);
+                c2dMethod.SetPayloadJson(payload);
+
+                await iotHubServiceClient.InvokeDeviceMethodAsync(DeviceId, "$edgeAgent", c2dMethod);
```
philipktlin: please handle exception.

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    class Program
+    {
+        static readonly ILogger Log = Logger.Factory.CreateLogger<Program>();
+        static readonly string ServiceClientConnectionString = Settings.Current.ServiceClientConnectionString;
+        static readonly string DeviceId = Settings.Current.DeviceId;
+        static readonly string DesiredModulesToRestartCSV = Settings.Current.DesiredModulesToRestartCSV;
+        static readonly int RandomRestartIntervalInMins = Settings.Current.RandomRestartIntervalInMins;
```
philipktlin: It is not random for restart interval; it is random for which module to be restarted.  Therefore I think this parameter should be named as "Restart interval In Mins".  However, do we want to restart with random intervals?

scripts/windows/test/Run-E2ETest.ps1
```diff
@@ -932,7 +944,7 @@ Function RunLongHaulTest
 
     $testStartAt = Get-Date
     $deviceId = "${ReleaseLabel}-Windows-${Architecture}-longHaul"
-    (Get-Content $DeploymentWorkingFilePath).replace('<Analyzer.DeviceID>',$deviceId) | Set-Content $DeploymentWorkingFilePath
+    (Get-Content $DeploymentWorkingFilePath).replace('<DeviceID>',$deviceId) | Set-Content $DeploymentWorkingFilePath
```
philipktlin: Need to roll back?

and-rewsmith: Yes, we could also get this param from the environment in the analyzer.

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: I think that restarting on fixed intervals is fine. I agree that a name change would help clarity here.

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: I think we should not block this. If the user wants to restart edgeAgent, they should be able to.

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: Forgot to change when debugging how this was used. Will make changes.

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: I did not know the container will stop. EdgeAgent will definitely restart so this absolutely needs to change.

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: No exception gets thrown for repeated commas, so I will do the split() first then remove all empty strings. This edge case would happen by accident, so it's better not to halt the module restarting.

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: I don't think it's a good idea to wait for new environment variables. Wouldn't the only way this could happen is from another deployment? In that case, the module is restarting anyway. I feel like the cleaner solution is to use Preconditions.Check() to throw an exception in the case of bad args, which will not restart the module.

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: I added precondition checks to all the environment parameters

edge-modules/ModuleRestarter/src/Settings.cs
```diff
@@ -0,0 +1,56 @@
...
+                    configuration.GetValue<int>("RestartIntervalInMins", 10));
+            });
+
+        Settings(
+            string serviceClientConnectionString,
+            string deviceId,
+            string desiredModulesToRestartCSV,
+            int restartIntervalInMins)
+        {
+            this.ServiceClientConnectionString = serviceClientConnectionString;
```
philipktlin: it should be better to move preconditions calls to here.

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    class Program
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger("ModuleRestarter");
+        static readonly string ServiceClientConnectionString = Preconditions.CheckNonWhiteSpace(Settings.Current.ServiceClientConnectionString, "ServiceClientConnectionString");
```
philipktlin: you don't need to define these static readonly variables as you can directly call though Settings.Current static variable.

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+                throw new ArgumentException("No module names specified");
+            }
+
+            ServiceClient iotHubServiceClient = ServiceClient.CreateFromConnectionString(ServiceClientConnectionString);
+            CloudToDeviceMethod c2dMethod = new CloudToDeviceMethod("RestartModule");
+            Random random = new Random();
+
+            while (!cts.Token.IsCancellationRequested)
+            {
+                string payload = "{{ \"SchemaVersion\": \"1.0\", \"Id\": \"{0}\" }}";
```
philipktlin: move this line up and out from this while block.

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+                payload = string.Format(payload, moduleNames[random.Next(0, moduleNames.Count)]);
+                Logger.LogInformation("RestartModule Method Payload: {0}", payload);
+                c2dMethod.SetPayloadJson(payload);
```
philipktlin: Just to confirm CloudToDeviceMethod instance can be reused and calls SetPayloadJson multiple times.

scripts/linux/runE2ETest.sh
```diff
@@ -979,7 +987,9 @@ LOADGEN2_TRANSPORT_TYPE="${LOADGEN2_TRANSPORT_TYPE:-amqp}"
 LOADGEN3_TRANSPORT_TYPE="${LOADGEN3_TRANSPORT_TYPE:-mqtt}"
 LOADGEN4_TRANSPORT_TYPE="${LOADGEN4_TRANSPORT_TYPE:-mqtt}"
 if [[ "${TEST_NAME,,}" == "longhaul" ]]; then
+    DESIRED_MODULES_TO_RESTART_CSV="${DESIRED_MODULES_TO_RESTART_CSV:-moduleRestarter}"
```
philipktlin: why default value is "moduleRestarter"?

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: Yes

scripts/linux/runE2ETest.sh
and-rewsmith: I did not want to pick any modules by default, as that could lead to unexpected behavior. I thought the cleanest solution was to restart itself so that we are still testing the the edgeAgent device method without having to restart any module involved in message/method/twin operations.

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: I thought this was a good pattern for two reasons:
- if you have to reuse the variable, you won't have to preface with Settings.Current each time
- provides a convenient place to perform preconditions.check (cleaner than in the main method)

I see Direct Method Cloud Sender does not use static readonly, so it might be a better idea to change my implementation for consistency. Do you think I should change for this reason?

edge-modules/ModuleRestarter/src/Settings.cs
```diff
@@ -0,0 +1,83 @@
...
+            this.RestartIntervalInMins = Preconditions.CheckRange(restartIntervalInMins, 0);
+        }
+
+        public static Settings Current => DefaultSettings.Value;
+
+        public string ServiceClientConnectionString { get; }
+
+        public string DeviceId { get; }
+
+        private string DesiredModulesToRestartCSV { get; }
```
and-rewsmith: I chose to keep this string field and make private to avoid using new in constructor to initialize a list.

edge-modules/ModuleRestarter/src/Settings.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                throw new ArgumentException("No module names specified");
+            }
+
+            return moduleNames;
+        }
+
+        public override string ToString()
+        {
+            Dictionary<string, string> state = new Dictionary<string, string>();
```
philipktlin: why need to create dictionary for json serialization?  I think you can serialize the object with default json serializer.  Also if this method is used for logging, you can use concatenate those fields with new line characters.

edge-modules/ModuleRestarter/src/Settings.cs
philipktlin: why do you need this field anymore? you can just add List<string> DesiredModulesToRestart {get;}.

edge-modules/ModuleRestarter/src/Settings.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Configuration;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Serialization;
+
+    [JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
```
philipktlin: why do you need this?

edge-modules/ModuleRestarter/src/Settings.cs
and-rewsmith: Settings takes in a string so we could use a list for this field, however would have to initialize it in the constructor via new(), which I thought was a bad practice.

edge-modules/ModuleRestarter/src/Settings.cs
and-rewsmith: So you can serialize the object, however this will not show the module names the way this is implemented, as the field is private. I created the dictionary to store the list of parsed module names as well.

This is all a byproduct of not wanting to use new() in the constructor for Settings.

edge-modules/ModuleRestarter/src/Settings.cs
and-rewsmith: You need this if you want to serialize the object for logging. If we are going to create a dictionary and then serialize the dict, then we can remove this header.

edge-modules/ModuleRestarter/src/Settings.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                if (!string.IsNullOrWhiteSpace(name))
+                {
+                    moduleNames.Add(name);
+                }
+            }
+
+            if (moduleNames.Count == 0)
+            {
+                throw new ArgumentException("No module names specified");
```
and-rewsmith: Having an ArgumentException here seems awkward to me in a method that takes no arguments. I initially chose as it describes a problem with the arguments to the module. I'm wondering if we should just use new() in constructor to clean the serialization / helper method up?

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+            Logger.LogInformation("ModuleRestarter Main() finished.");
+            return 0;
+        }
+
+        /// <summary>
+        /// Restarts random modules periodically (with default restart occurrence once every 10 minutes)
+        /// </summary>
+        static async Task RestartModules(CancellationTokenSource cts)
+        {
+            ServiceClient iotHubServiceClient = ServiceClient.CreateFromConnectionString(Settings.Current.ServiceClientConnectionString);
```
philipktlin: this line may throw exception; please handle it.

edge-modules/ModuleRestarter/src/Settings.cs
```diff
@@ -0,0 +1,68 @@
...
+            string desiredModulesToRestartCSV,
+            int restartIntervalInMins)
+        {
+            this.ServiceClientConnectionString = Preconditions.CheckNonWhiteSpace(serviceClientConnectionString, "ServiceClientConnectionString");
+            this.DeviceId = Preconditions.CheckNonWhiteSpace(deviceId, "DeviceId");
+            this.RestartIntervalInMins = Preconditions.CheckRange(restartIntervalInMins, 0);
+
+            // mitigate unintended repeated commas
+            this.DesiredModulesToRestart = new List<string>();
+            foreach (string name in desiredModulesToRestartCSV.Split(","))
```
philipktlin: You don't need to use foreach here; you can directly call split method with SplitOptions to remove empty entries.  And I think you can return array instead of list.

edge-modules/ModuleRestarter/src/Settings.cs
```diff
@@ -0,0 +1,68 @@
...
+        public string DeviceId { get; }
+
+        public List<string> DesiredModulesToRestart { get; }
+
+        public int RestartIntervalInMins { get; }
+
+
+        public override string ToString()
+        {
+            return JsonConvert.SerializeObject(this, Formatting.Indented);
```
philipktlin: why not just construct all properties delimited by new line?

scripts/linux/runE2ETest.sh
```diff
@@ -979,7 +987,9 @@ LOADGEN2_TRANSPORT_TYPE="${LOADGEN2_TRANSPORT_TYPE:-amqp}"
 LOADGEN3_TRANSPORT_TYPE="${LOADGEN3_TRANSPORT_TYPE:-mqtt}"
 LOADGEN4_TRANSPORT_TYPE="${LOADGEN4_TRANSPORT_TYPE:-mqtt}"
 if [[ "${TEST_NAME,,}" == "longhaul" ]]; then
+    DESIRED_MODULES_TO_RESTART_CSV="${DESIRED_MODULES_TO_RESTART_CSV:-,}"
```
philipktlin: did you try empty string as default value?

edge-modules/ModuleRestarter/src/Settings.cs
and-rewsmith: I thought this was cleaner than attempting manually. If a new field is added, you do not have to edit the ToString() method. Also, loadgen currently uses the same JsonConvert approach.

philipktlin: ServiceClientConnectionString/any secrets should not be printed out without mask.  It is convenient but it may not be exactly your intent (when someone added a new field later).  Also, I prefer simple concatenation as it should be faster than object serialization.

scripts/linux/runE2ETest.sh
and-rewsmith: yes

edge-modules/ModuleRestarter/src/Settings.cs
and-rewsmith: Unfortunately, SplitOptions.RemoveEmptyEntries does not remove whitespace. A solution is outlined [here](https://stackoverflow.com/questions/10682301/stringsplitoptions-removeemptyentries-doesnt-work-as-advertised/10682341) however dotnet 2.0 does not support this.

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: I looked at the implementation within DirectMethodCloudSender. Two things I would like to confirm:
- Since this is a cloud operation, we need to retry in case of transient failure. DMCS exits the application, however their restartPolicy is set to 'always' which ensures a retry. We are going to have to manually retry this.
- DMCS nests try catches which I feel like we should not do. I am going to try catch this separately.

edge-modules/ModuleRestarter/ModuleRestarter.csproj
and-rewsmith: This cpp Microsoft Docs [link](https://github.com/MicrosoftDocs/cpp-docs/issues/282#issuecomment-396361257) gives more insight into the option (purpose is same as for c#). I think we should keep it for security reasons so I will add the option where needed.

edge-modules/ModuleRestarter/src/Program.cs
```diff
@@ -0,0 +1,85 @@
...
+        public static int Main() => MainAsync().Result;
+
+        static async Task<int> MainAsync()
+        {
+            Logger.LogInformation($"Starting module restarter with the following settings:\r\n{Settings.Current}");
+
+            if (Settings.Current.DesiredModulesToRestart.Count == 0)
+            {
+                Logger.LogInformation("No modules names found in input. Stopping.");
+                return 0;
```
ancaantochi: If the module is stopped when no modules were set, but is deployed to be running. edgeAgent will start it.

edge-modules/ModuleRestarter/src/Settings.cs
ancaantochi: You can add [JsonIgnore] attribute to the ServiceClientConnectionString to avoid to be serialized

edge-modules/ModuleRestarter/src/Program.cs
and-rewsmith: Restart policy is "on-failure"

and-rewsmith: However, someone could deploy the module differently from long-haul settings, which could lead to unexpected behavior.

edge-modules/ModuleRestarter/src/Settings.cs
and-rewsmith: I liked the output format with output serialization, as well as the maintainability. Since we can add the tag [JsonIgnore], I prefer the serialization solution.

PR #1917, opened 2019-10-26
Fix test result path

philipktlin:
fix test result path to remove test dll source folder.


philipktlin:
Verified with checkin build that test results in Microsoft.Azure.Devices.Edge.Agent.Docker.Test are shown up.

PR #1920, opened 2019-10-28
[k8s] k8s module needs to find IDeviceManager, IEntityStore.

darobs:
ModuleHttpManagement implements a new interface, IDeviceManager,  needed to be added to autofac in K8s mode.
IEntityStore was changed to resolved as a Task, k8s module was missed in change.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/KubernetesModule.cs
```diff
@@ -287,9 +289,10 @@ protected override void Load(ContainerBuilder builder)
             builder.Register(
                     async c =>
                     {
-                        var moduleStateStore = c.Resolve<IEntityStore<string, ModuleState>>();
+                        var moduleStateStoreTask = c.Resolve<Task<IEntityStore<string, ModuleState>>>();
```
massand: This code isn't as clean as just awaiting on moduleStateStore after resolving IRestartPolicyManager and IRunInfoProvider. I recommend we consolidate to one variable again.

darobs: Thanks, I changed it on the chance this was part of the problem.  I'll change it back and test it out.

PR #1921, opened 2019-10-28
K8s RuntimeInfo provider needs to fails faster if it can't get SystemInfo

darobs:
Rather than waiting for 2 minutes, we should let agent fail faster if iotedged (or its proxy) is not ready when agent comes up.

User will still see agent reset, but the time to wait will be shorter.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs
```diff
@@ -17,17 +18,20 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
...
         readonly IModuleManager moduleManager;
+        readonly CancellationTokenSource tokenSource;
 
         public KubernetesRuntimeInfoProvider(string deviceNamespace, IKubernetes client, IModuleManager moduleManager)
         {
             this.deviceNamespace = Preconditions.CheckNonWhiteSpace(deviceNamespace, nameof(deviceNamespace));
             this.client = Preconditions.CheckNotNull(client, nameof(client));
             this.moduleManager = Preconditions.CheckNotNull(moduleManager, nameof(moduleManager));
             this.moduleRuntimeInfo = new ConcurrentDictionary<string, ModuleRuntimeInfo>();
+            this.tokenSource = new CancellationTokenSource(SystemInfoTimeout);
```
CindyXing: As discussed, this is due to the sequence of sidecarproxy and edgeagent running time. It is similar to envoy sidecar with applicaiton container. Long term, we can try the pod readiness-gate https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs
```diff
@@ -53,9 +57,6 @@ public bool RemovePodInfo(string podName)
                 sinceSeconds: since.Map(sec => (int?)sec).OrDefault(),
                 cancellationToken: cancellationToken);
 
-        public Task<SystemInfo> GetSystemInfo()
-        {
-            return this.moduleManager.GetSystemInfoAsync();
-        }
+        public Task<SystemInfo> GetSystemInfo() => this.moduleManager.GetSystemInfoAsync(this.tokenSource.Token);
```
dmolokanov: Usually, the caller passes the cancellation token into the method. What will happen if the token will be canceled due to a timeout and caller execute this method?

dmolokanov: all other methods from the IRuntimeInfoProvider interface have cancellation token as a parameter. we should add it here as well and move cancellation token to the higher level

darobs: I'm always leery of changing interfaces that haven't changed in a while, but I agree.  It's going to a lot more code changes, though/

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/KubernetesModule.cs
```diff
@@ -111,6 +114,7 @@ public class KubernetesModule : Module
             this.closeOnIdleTimeout = closeOnIdleTimeout;
             this.idleTimeout = idleTimeout;
             this.experimentalFeatures = experimentalFeatures;
+            this.tokenSource = new CancellationTokenSource(SystemInfoTimeout);
```
dmolokanov: lets make it local variable and move it closer to usage. in future if someone will try to resolve other instance of a DockerEnvironmentProvider it will be a new token for it

PR #1922, opened 2019-10-28
Fix update reported properties

ancaantochi:
Changed to use SemaphoreSlim because with AsyncAutoResetEvent it loses the set if there is no thread waiting. Issue caught by E2E tests.


varunpuranik:
Can we add unit tests to catch this issue?

PR #1931, opened 2019-10-31
[k8s] Modify edgedeployment cleanup in Kubernetes

massand:
For Kubernetes edge cleanup, we are adding the following pieces to achieve automatic garbage collection by the Kubernetes environment: 

1) attach an iotedged deployment ownerreference to the edgeagent deployment at creation time in iotedged
2) attach an iotedge deployment ownerreference to the edgedeployment crd object
3) attach an edgedeployment crd object ownerReference in custom module deployments/service accounts/services/pvc 

How validated:
helm install 
helm delete

This PR addreses item 1 and 3 above.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/IKubernetesDeploymentMapper.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment
 
     public interface IKubernetesDeploymentMapper
     {
-        V1Deployment CreateDeployment(IModuleIdentity identity, KubernetesModule module, IDictionary<string, string> labels);
+        V1Deployment CreateDeployment(IModuleIdentity identity, KubernetesModule module, IDictionary<string, string> labels, EdgeDeploymentDefinition edgeDeploymentDefinition);
```
dmolokanov: I don't think it is a good idea to pass the whole EdgeDeploymentDefinition in each method for mappers. 
I think maybe we can create a context for each method `Create{Resource}()` and this context will contain data to make a conversion to `{Resource}`. It will help us:
1. pass additional data into methods in the future.
2. we can even make a class hierarchy for these contexts

dmolokanov: Or we can make this owner information to be a part of KubernetesModule. And in the long run this KubernetesModule can become a CRD object to operate on.
```csharp
    public class KubernetesModule
    {
        ...
        // TODO initialize via constructor
        public KubernetesModuleOwner Owner { get; }
        ...
    }

    public class KubernetesModuleOwner
    {
        // TODO initialize via constructor
        public string ApiVersion { get; }
        public string Kind { get; }
        public string Name { get; }
        public string Namespace { get; }
        public string Uid { get; }
    }
```

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesDeploymentMapperTest.cs
```diff
@@ -37,6 +39,10 @@ public class KubernetesDeploymentMapperTest
             }
         };
 
+        static readonly ResourceName ResourceName = new ResourceName("hostname", "deviceId");
+
+        static readonly EdgeDeploymentDefinition EdgeDeploymentDefinition = new EdgeDeploymentDefinition("v1", "EdgeDeployment", new k8s.Models.V1ObjectMeta(name: ResourceName), new List<KubernetesModule>(), null);
```
dmolokanov: We have a default in the constructor so you don't need to specify the last parameter. Right IDE usually highlights this :)
```suggestion
        static readonly EdgeDeploymentDefinition EdgeDeploymentDefinition = new EdgeDeploymentDefinition("v1", "EdgeDeployment", new k8s.Models.V1ObjectMeta(name: ResourceName), new List<KubernetesModule>());
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/IKubernetesDeploymentMapper.cs
massand: The KubernetesModule and KubernetesConfig contains fields that are specified by IoTHub. They are pretty much clones of DockerModule and DockerConfig. However, CRD objects (aka edgeDeploymentDefinition) represent Kubernetes objects created at deployment time only. I am reluctant to mix the two at the moment.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -80,16 +80,25 @@ public void UpdateDeployment(V1Deployment to, V1Deployment from)
...
             var selector = new V1LabelSelector(matchLabels: labels);
             var deploymentSpec = new V1DeploymentSpec(replicas: 1, selector: selector, template: podSpec);
+            var ownerReferences = new List<V1OwnerReference>
+            {
+                new V1OwnerReference(
+                    apiVersion: edgeDeploymentDefinition.ApiVersion,
+                    kind: edgeDeploymentDefinition.Kind,
+                    name: edgeDeploymentDefinition.Metadata.Name,
+                    uid: edgeDeploymentDefinition.Metadata.Uid,
+                    blockOwnerDeletion: true)
```
darobs: Is this required?  We are by default doing background deletion, so this shouldn't apply.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentOperator.cs
```diff
@@ -114,14 +114,14 @@ async Task HandleEdgeDeploymentChangedAsync(WatchEventType type, EdgeDeploymentD
...
-                    var status = await this.controller.DeployModulesAsync(desiredModules, this.currentModules);
+                    var status = await this.controller.DeployModulesAsync(desiredModules, this.currentModules, edgeDeploymentDefinition);
                     await this.ReportEdgeDeploymentStatus(edgeDeploymentDefinition, status);
                     this.currentModules = desiredModules;
                     this.currentStatus = status;
                     break;
 
                 case WatchEventType.Deleted:
-                    await this.controller.PurgeModulesAsync();
+                    // Kubernetes garbage collection will handle cleanup of deployment artifacts
```
darobs: Can we remove IEdgeDeploymentController/EdgeDeploymentController::PurgeModulesAsync?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/IKubernetesDeploymentMapper.cs
massand: Good suggestion. Please review latest changes!

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentOperator.cs
massand: Done

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
massand: You are right. I will remove this after testing.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
```diff
@@ -41,7 +41,7 @@ bool ShouldCreatePvc(Mount mount)
             return this.storageClassName.HasValue || this.persistentVolumeName.HasValue;
         }
 
-        V1PersistentVolumeClaim ExtractPvc(Mount mount, IDictionary<string, string> labels)
+        V1PersistentVolumeClaim ExtractPvc(Mount mount, IDictionary<string, string> labels, KubernetesModule module)
```
darobs: Which should be the preferred order "mount, label, module," or "module, mount, label?"(@CindyXing 's PR has it in the latter order)

massand: Thanks. I'll handle on merge after that PR goes in.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentOperator.cs
```diff
@@ -99,12 +99,12 @@ internal async Task EdgeDeploymentOnEventHandlerAsync(WatchEventType type, EdgeD
             }
         }
 
-        async Task HandleEdgeDeploymentChangedAsync(WatchEventType type, EdgeDeploymentDefinition edgeDeploymentDefinition)
+        async Task HandleEdgeDeploymentChangedAsync(WatchEventType type, EdgeDeploymentDefinition edgeDeploymentCrdObject)
```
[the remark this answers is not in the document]
massand: It clarifies to reader that the source of the event is the CRD object, and not the definition. Not the CRD.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
```diff
@@ -67,7 +67,16 @@ V1PersistentVolumeClaim ExtractPvc(KubernetesModule module, Mount mount, IDictio
                 persistentVolumeClaimSpec.StorageClassName = this.storageClassName.OrDefault();
             }
 
-            return new V1PersistentVolumeClaim(metadata: new V1ObjectMeta(name: volumeName, labels: labels), spec: persistentVolumeClaimSpec);
+            var ownerReferences = new List<V1OwnerReference>
```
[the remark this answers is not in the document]
kkdawkins: Like the caller of this class creates the list once?

edgelet/edgelet-kube/src/module/create.rs
```diff
@@ -51,9 +71,51 @@ where
...
+    S::ResBody: Stream,
+    Body: From<S::ResBody>,
+    S::Error: Fail,
+    S::Future: Send,
+{
+    let module_name = module_name.to_string();
+    runtime
+        .client()
+        .lock()
+        .expect("Unexpected lock error")
```
kkdawkins: Be more specific on this error message. What lock is failing?

edgelet/edgelet-kube/src/convert/to_k8s.rs
```diff
@@ -430,12 +463,22 @@ pub fn spec_to_role_binding(
     let mut annotations = BTreeMap::new();
     annotations.insert(EDGE_ORIGINAL_MODULEID.to_string(), spec.name().to_string());
 
+    let owner_reference = api_meta::OwnerReference {
```
kkdawkins: Once again, is it possible to DRY out this conversion code?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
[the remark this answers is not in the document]
massand: Good point about redundant code (something I will task myself to refactor in near future). Unfortunately, the K8s object creation code in Agent is at the end of the call stack. And we are forced to recreate the same metadata objects for each K8s object type, even though they all share the same labels, owners, etc.

edgelet/edgelet-kube/src/module/create.rs
massand: It is the global client lock. Staying within error message convention for now.

edgelet/edgelet-kube/src/convert/to_k8s.rs
massand: Added this to my task list for further refactoring when we have figured out all the right  objects produced by iotedged and edgeAgent and the K8s object associations to each other.

dmolokanov: lets put it on a separate method in this module for now.

edgelet/edgelet-kube/src/module/create.rs
```diff
@@ -29,18 +30,37 @@ where
...
     let module_for_sa = module.clone();
 
+    let runtime_for_rb = runtime.clone();
+    let module_for_rb = module.clone();
+
     let runtime_for_deployment = runtime.clone();
     let module_for_deployment = module.clone();
 
     let module_name = module.name().to_string();
+    let iotedged_module_name = "iotedged";
```
dmolokanov: lets inline this variable into method call

edgelet/edgelet-kube/src/module/create.rs
```diff
@@ -29,18 +30,37 @@ where
...
 
-    create_or_update_service_account(&runtime, &module)
-        .and_then(move |_| create_or_update_role_binding(&runtime_for_sa, &module_for_sa))
-        .and_then(move |_| {
-            create_or_update_deployment(&runtime_for_deployment, &module_for_deployment)
+    get_module_owner(&runtime_for_owner, &iotedged_module_name)
+        .and_then(move |module_owner| {
+            let module_owner1 = module_owner.clone();
```
dmolokanov: I think another approach here, you can wrap module_owner into Arc and place cloning references when call methods.

edgelet/edgelet-kube/src/module/create.rs
dmolokanov: normally it should not happen in a way how we use kube-client, but rust requires us to handle this explicitly

edgelet/edgelet-kube/src/module/create.rs
```diff
@@ -51,9 +71,51 @@ where
...
+            deployments
+                .items
+                .into_iter()
+                .find(|deployment| {
+                    deployment.metadata.as_ref().map_or(false, |meta| {
+                        meta.name.as_ref().map_or(false, |n| *n == module_name)
+                    })
+                })
+                .ok_or(Error::from(ErrorKind::MissingMetadata))
+                .and_then(|this_deployment| KubeModuleOwner::try_from(this_deployment))
```
dmolokanov:
```suggestion
                .and_then(|deployment| KubeModuleOwner::try_from(deployment))
```

edgelet/edgelet-kube/src/lib.rs
```diff
@@ -17,11 +17,77 @@ mod registry;
...
+        let metadata = deployment
+            .metadata
+            .as_ref()
+            .ok_or(ErrorKind::MissingMetadata)?;
+        Ok(Self {
+            name: metadata
+                .name
+                .as_ref()
+                .map(|name| name.to_string())
+                .ok_or(ErrorKind::MissingMetadata)?,
```
dmolokanov: now this `ErrorKind::MissingMetadata` is used in bunch of different contexts. lets make it more descriptive:
```rust
.ok_or(ErrorKind::MissingMetadata(MissingMetadataReason::Name))?
```

edgelet/edgelet-kube/src/lib.rs
```diff
@@ -17,11 +17,77 @@ mod registry;
...
+            api_version: Deployment::api_version().to_string(),
+            kind: Deployment::kind().to_string(),
+            uid: metadata
+                .uid
+                .as_ref()
+                .map(|uid| uid.to_string())
+                .ok_or(ErrorKind::MissingMetadata)?,
```
dmolokanov:
```suggestion
                .ok_or(ErrorKind::MissingMetadata(MissingMetadataReason::Uid))?,
```

edgelet/edgelet-kube/src/module/create.rs
```diff
@@ -51,9 +71,51 @@ where
...
+        .map(move |deployments| {
+            deployments
+                .items
+                .into_iter()
+                .find(|deployment| {
+                    deployment.metadata.as_ref().map_or(false, |meta| {
+                        meta.name.as_ref().map_or(false, |n| *n == module_name)
+                    })
+                })
+                .ok_or(Error::from(ErrorKind::MissingMetadata))
```
dmolokanov:
```suggestion
                .ok_or(Error::from(ErrorKind::MissingMetadata(MissingMetadataReason::Owner))
```

edgelet/edgelet-kube/src/module/create.rs
kkdawkins: @massand  - Okay, but why can't the message be "Global client lock failure" :) Would help with debugging if it ever were to happen, no?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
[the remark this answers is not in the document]
kkdawkins: Agreed about not taking too much time, but you could very quickly write a function or an extension method to at least not contribute to the redundant code...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesModule.cs
```diff
@@ -38,7 +39,8 @@ public KubernetesModule(IModule module, KubernetesConfig config)
             ConfigurationInfo configurationInfo,
             IDictionary<string, EnvVal> env,
             KubernetesConfig settings,
-            ImagePullPolicy imagePullPolicy)
+            ImagePullPolicy imagePullPolicy,
+            KubernetesModuleOwner owner = null)
```
kkdawkins: Is this parameter truly optional? Why not use Option? Also, did you need to add this to the GetHashCode call below?

kkdawkins: I ask about it truly being optional b/c I see a lot of checks against it being null later on which makes me think that it is a required property...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -78,22 +78,22 @@ public async Task<EdgeDeploymentStatus> DeployModulesAsync(ModuleSet desiredModu
                 };
 
                 var desiredServices = desiredModules.Modules
-                    .Select(module => this.serviceMapper.CreateService(moduleIdentities[module.Key], module.Value as KubernetesModule, labels[module.Key]))
+                    .Select(module => this.serviceMapper.CreateService(moduleIdentities[module.Key], (KubernetesModule)module.Value, labels[module.Key]))
```
kkdawkins: Why move from as to cast? Cast will throw an exception - is that what you want? As allows us to handle the error w/o an excpetion by checking against null.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentCommand.cs
```diff
@@ -79,7 +81,15 @@ async Task UpdateImagePullSecrets(IEnumerable<ImagePullSecret> imagePullSecrets,
         {
             foreach (var imagePullSecret in imagePullSecrets)
             {
-                var secretMeta = new V1ObjectMeta(name: imagePullSecret.Name, namespaceProperty: this.deviceNamespace);
+                var ownerReferences = new List<V1OwnerReference>
```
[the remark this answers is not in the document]
massand: Fixed. Thanks!

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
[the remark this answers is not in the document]
massand: Makes sense. It isn't an extension method however. It will just be within a separate utility class that does the one way conversion from KubernetesModuleOwner->V1ObjectReference.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
massand: I don't have a major opinion on this item yet - maybe @dmolokanov does. The code downstream is going to NullException anyway, should 'as' operator assign null. I wanted to failfast sooner here.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesModule.cs
massand: Fixed

edgelet/edgelet-kube/src/module/create.rs
massand: The stack trace is plenty for identifying one of 8 places this lock error could occur in this module.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesModuleOwner.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
+{
+    using System;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Serialization;
+
+    public class KubernetesModuleOwner
+    {
+        [JsonConstructor]
+        public KubernetesModuleOwner(string apiVersion, string kind, string name, string uid)
```
darobs: Do we care if all or some are null?

massand: The checks are happening during KubernetesModule (the Service one) initialize. The next point this will failfast is V1ObjectMeta creation, if there were any gaps. I think this is considered an internal structure.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
dmolokanov: Yes, if a cast is failed here - it means that we failed to maintain only KubernetesModule here on a call stack before this place. It's better to fail fast here :)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentUtils.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment
+{
+    using System.Collections.Generic;
+    using k8s.Models;
+
+    public class EdgeDeploymentUtils
+    {
+        public static List<V1OwnerReference> KubeModuleOwnerToOwnerReferences(KubernetesModuleOwner kubeModuleOwner) =>
```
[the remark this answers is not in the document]
massand: Fixed

PR #1932, opened 2019-10-31
Fixing a bug, changing values to match IoTHub, and adding missing tests

dylanbronson:
IoT Hub bumped up their limits for Twins. The following limits need to be updated in the Edge Hub to match.

TwinPropertyMaxDepth = 10;
TwinPropertyNameMaxLength = 1024;
TwinPropertyValueMaxLength = 4096;
TwinPropertyDocMaxLength = 32 * 1024;     // 32kb

This change sets those values and fixes a bug where TwinPropertyNameMaxLength and TwinPropertyValueMaxLength were sharing the same value.


ancaantochi:
>                 {

levels up to 10 should be added here to validate that it works

---
Refers to: edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/twin/ReportedPropertiesValidatorTest.cs:22 in 8f5b4a5. [](commit_id = 8f5b4a512a8509f1643b6e512c92eae0777d8172, deletion_comment = False)

PR #1934, opened 2019-10-31
Bug 5588683: If two modules reference the same volume name, edge agent gets into a crash loop

CindyXing:
Currently customers can specify volume mount through the createoption; and can also add k8s style PV/storageclass in the extended feature section. The -v and k8s can both be specified or either of them. 

For above, we create PVCs and then let k8s to schedule the right PV. PV will be created in case of stroageclass or the use what customer specified. 

The root cause of the crash is that customer specify two modules mouting to the same volume. Our code creates PVC using the mount.source. When adding to the dictionary of the PVC list, the name crashes. 

The change is to 
- Include module name in PVC to prevent conflict
- Make sure using the PV name customer specified otherwise use mount.source
- Allow volume share
- mount to the name customer specified.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesModule.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesModule.cs
@@ -5,6 +5,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
     using System.Collections.Generic;
     using System.Collections.Immutable;
     using Microsoft.Azure.Devices.Edge.Agent.Core;
+    using Microsoft.Azure.Devices.Edge.Agent.Docker.Models;
     using Microsoft.Azure.Devices.Edge.Util;
     using Newtonsoft.Json;
 
@@ -81,6 +82,11 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
         public virtual bool Equals(IModule other) => this.Equals(other as KubernetesModule);
 
         public bool Equals(IModule<KubernetesConfig> other) => this.Equals(other as KubernetesModule);
+
+        public static string PvcName(KubernetesModule module, Mount mount)
+        {
+            return KubeUtils.SanitizeK8sValue($"{module.Name}-{mount.Source}");
+        }
 
         public bool Equals(KubernetesModule other)
         {
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
@@ -310,7 +310,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment.Deploymen
             // Volume name will be customer defined name or modulename + mount.source
             if (this.persistentVolumeName.HasValue)
             {
-                this.persistentVolumeName.ForEach(pvVolumeName => volumeName = pvVolumeName);
+                volumeName = this.persistentVolumeName.OrDefault();
                 return new V1Volume(volumeName, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(pvcName, mount.ReadOnly));
             }
             else if (this.storageClassName.HasValue)
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
@@ -235,13 +235,13 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment.Deploymen
                 .FlatMap(config => Option.Maybe(config.Mounts))
                 .Map(mounts => mounts.Where(mount => mount.Type.Equals("volume", StringComparison.InvariantCultureIgnoreCase)).ToList());
 
-            volumeMounts.Map(mounts => mounts.Select(this.GetVolume))
+            volumeMounts.Map(mounts => mounts.Select(mount => this.GetVolume(module, mount)))
                 .ForEach(volumes => volumeList.AddRange(volumes));
 
             volumeMounts.Map(mounts => mounts.Select(mount => new V1VolumeMount(mount.Target, KubeUtils.SanitizeDNSValue(mount.Source), readOnlyProperty: mount.ReadOnly)))
                 .ForEach(mounts => volumeMountList.AddRange(mounts));
 
-            // collect volume and volume mounts from CreateOption.Volumes section
+            // collect volume and volume mounts from CreateOption.Volumes section @kubernetes extended feature
             module.Config.CreateOptions.Volumes
                 .Map(volumes => volumes.Select(volume => volume.Volume).FilterMap())
                 .ForEach(volumes => volumeList.AddRange(volumes));
@@ -301,14 +301,27 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment.Deploymen
             return proxyLevel;
         }
 
-        V1Volume GetVolume(Mount mount)
+        V1Volume GetVolume(KubernetesModule module, Mount mount)
         {
-            string name = KubeUtils.SanitizeDNSValue(mount.Source);
-            return this.ShouldUsePvc()
-                ? new V1Volume(name, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(name, mount.ReadOnly))
-                : new V1Volume(name, emptyDir: new V1EmptyDirVolumeSource());
-        }
+            string name = KubeUtils.SanitizeK8sValue(mount.Source);
+            string pvcName = KubeUtils.SanitizeK8sValue(module.Name) + name;
 
-        bool ShouldUsePvc() => this.persistentVolumeName.HasValue || this.storageClassName.HasValue;
+            // PVC name will be modulename + volume name
+            // Volume name will be customer defined name or modulename + mount.source
+            if (this.persistentVolumeName.HasValue)
+            {
+                this.persistentVolumeName.ForEach(volumeName => name = volumeName);
+                return new V1Volume(name, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(pvcName, mount.ReadOnly));
+            }
+            else if (this.storageClassName.HasValue)
+            {
+                return new V1Volume(pvcName, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(pvcName, mount.ReadOnly));
+            }
+            else
+            {
+                return new V1Volume(name, emptyDir: new V1EmptyDirVolumeSource());
+            }
+           
+        }
     }
 }
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
@@ -303,15 +303,15 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment.Deploymen
 
         V1Volume GetVolume(KubernetesModule module, Mount mount)
         {
-            string name = KubeUtils.SanitizeK8sValue(mount.Source);
+            string volumeName = KubeUtils.SanitizeK8sValue(mount.Source);
             string pvcName = KubernetesModule.PvcName(module, mount);
 
             // PVC name will be modulename + volume name
             // Volume name will be customer defined name or modulename + mount.source
             if (this.persistentVolumeName.HasValue)
             {
-                this.persistentVolumeName.ForEach(volumeName => name = volumeName);
-                return new V1Volume(name, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(pvcName, mount.ReadOnly));
+                this.persistentVolumeName.ForEach(pvVolumeName => volumeName = pvVolumeName);
+                return new V1Volume(volumeName, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(pvcName, mount.ReadOnly));
             }
             else if (this.storageClassName.HasValue)
             {
@@ -319,7 +319,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment.Deploymen
             }
             else
             {
-                return new V1Volume(name, emptyDir: new V1EmptyDirVolumeSource());
+                return new V1Volume(volumeName, emptyDir: new V1EmptyDirVolumeSource());
             }
         }
     }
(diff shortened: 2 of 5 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -301,14 +301,27 @@ static string ToProxyLogLevel(LogEventLevel level)
...
-        }
+            string name = KubeUtils.SanitizeK8sValue(mount.Source);
+            string pvcName = KubeUtils.SanitizeK8sValue(module.Name) + name;
 
-        bool ShouldUsePvc() => this.persistentVolumeName.HasValue || this.storageClassName.HasValue;
+            // PVC name will be modulename + volume name
+            // Volume name will be customer defined name or modulename + mount.source
+            if (this.persistentVolumeName.HasValue)
+            {
+                this.persistentVolumeName.ForEach(volumeName => name = volumeName);
```
dmolokanov: This re-assigning name is a bit confusing. It's can be easier in this way:
```csharp
return V1Volume(this.presistentVolumeName.GetOrElse(name), ....
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -301,14 +301,27 @@ static string ToProxyLogLevel(LogEventLevel level)
...
+                return new V1Volume(name, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(pvcName, mount.ReadOnly));
+            }
+            else if (this.storageClassName.HasValue)
+            {
+                return new V1Volume(pvcName, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(pvcName, mount.ReadOnly));
```
dmolokanov: It use to be `name` that was used for volume name, now it is changed to `pvcName`. Could you please explain what is the reason for this change?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -301,14 +301,27 @@ static string ToProxyLogLevel(LogEventLevel level)
...
+            }
+            else
+            {
+                return new V1Volume(name, emptyDir: new V1EmptyDirVolumeSource());
+            }
+
```
dmolokanov: nit: redundant blank line

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -301,14 +301,27 @@ static string ToProxyLogLevel(LogEventLevel level)
...
-        V1Volume GetVolume(Mount mount)
+        V1Volume GetVolume(KubernetesModule module, Mount mount)
         {
-            string name = KubeUtils.SanitizeDNSValue(mount.Source);
-            return this.ShouldUsePvc()
-                ? new V1Volume(name, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(name, mount.ReadOnly))
-                : new V1Volume(name, emptyDir: new V1EmptyDirVolumeSource());
-        }
+            string name = KubeUtils.SanitizeK8sValue(mount.Source);
+            string pvcName = KubeUtils.SanitizeK8sValue(module.Name) + name;
```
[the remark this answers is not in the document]
CindyXing: Like the idea. Thankd Denis. 

My understanding volume share occurs only when the volume (PV) exists, i.e. customers have the volume created manually. For that, this PR didn't change the behavior. 

In case of storageclass, first we crate PVC and then PV, where PVC and PV will have the same name. If customer doesn't specify k8s attributes, the emptydir will be created. So there is no volume to share.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
```diff
@@ -41,9 +41,9 @@ bool ShouldCreatePvc(Mount mount)
             return this.storageClassName.HasValue || this.persistentVolumeName.HasValue;
         }
 
-        V1PersistentVolumeClaim ExtractPvc(Mount mount, IDictionary<string, string> labels)
+        V1PersistentVolumeClaim ExtractPvc(string moduleName, Mount mount, IDictionary<string, string> labels)
         {
-            string name = KubeUtils.SanitizeK8sValue(mount.Source);
+            string name = KubeUtils.SanitizeK8sValue(moduleName) + KubeUtils.SanitizeK8sValue(mount.Source);
```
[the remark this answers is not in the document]
CindyXing: ok

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
CindyXing: When storageclass is used, there is no PV. In this case, PVC and PV are having the same name.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
CindyXing: It is to ensure the volume name be what customer specified and not the mount source. As they can be different.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
massand: @CindyXing  Do we need to change logic in the CollectModuleVolumes to use pvcname when querying volumes?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
```diff
@@ -41,9 +41,9 @@ bool ShouldCreatePvc(Mount mount)
             return this.storageClassName.HasValue || this.persistentVolumeName.HasValue;
         }
 
-        V1PersistentVolumeClaim ExtractPvc(Mount mount, IDictionary<string, string> labels)
+        V1PersistentVolumeClaim ExtractPvc(KubernetesModule module, Mount mount, IDictionary<string, string> labels)
         {
-            string name = KubeUtils.SanitizeK8sValue(mount.Source);
+            string name = KubernetesModule.PvcName(module, mount);
```
massand: Nit readability: I would recommend changing variable name because name is super overloaded
```suggestion
            string pvcName = KubernetesModule.PvcName(module, mount);
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
CindyXing: @massand, CollectModuleVolumes did take the change in GetVolume.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
CindyXing: Sounds good.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -301,14 +301,26 @@ static string ToProxyLogLevel(LogEventLevel level)
...
-        V1Volume GetVolume(Mount mount)
+        V1Volume GetVolume(KubernetesModule module, Mount mount)
         {
-            string name = KubeUtils.SanitizeDNSValue(mount.Source);
-            return this.ShouldUsePvc()
-                ? new V1Volume(name, persistentVolumeClaim: new V1PersistentVolumeClaimVolumeSource(name, mount.ReadOnly))
-                : new V1Volume(name, emptyDir: new V1EmptyDirVolumeSource());
-        }
+            string name = KubeUtils.SanitizeK8sValue(mount.Source);
+            string pvcName = KubernetesModule.PvcName(module, mount);
```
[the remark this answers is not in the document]
CindyXing: The static function is KubernetesModule specific. It is coherent to be in the same file.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesDeploymentMapperTest.cs
```diff
@@ -166,7 +166,8 @@ public void PvcMappingForVolumeNameVolume()
             Assert.True(pod != null);
             var podVolume = pod.Spec.Volumes.Single(v => v.Name == "elephant");
             Assert.NotNull(podVolume.PersistentVolumeClaim);
-            Assert.Equal(module.Name + "a-volume", podVolume.PersistentVolumeClaim.ClaimName);
+
+            Assert.Equal(module.Name + "-a-volume", podVolume.PersistentVolumeClaim.ClaimName);
```
dmolokanov: Usually, it is recommended to put some precomputed value to increase the readability of the tests. And if it fails on this line it will be obvious to see what's wrong.
I'll leave it only here but can be applied to all changes below
```suggestion 
            Assert.Equal("module1-a-volume", podVolume.PersistentVolumeClaim.ClaimName);
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesModule.cs
```diff
@@ -82,6 +83,11 @@ public KubernetesModule(IModule module, KubernetesConfig config)
 
         public bool Equals(IModule<KubernetesConfig> other) => this.Equals(other as KubernetesModule);
 
+        public static string PvcName(KubernetesModule module, Mount mount)
```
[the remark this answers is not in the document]
CindyXing: PVC belongs to a module and mount doesn't always need a PVC, right?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesDeploymentMapperTest.cs
CindyXing: En, good point. It seems that the module.Name is a common pattern in the existing test cases. Let's have a work item to improve readability.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -301,14 +301,26 @@ static string ToProxyLogLevel(LogEventLevel level)
...
-        }
+            string volumeName = KubeUtils.SanitizeK8sValue(mount.Source);
+            string pvcName = KubernetesModule.PvcName(module, mount);
 
-        bool ShouldUsePvc() => this.persistentVolumeName.HasValue || this.storageClassName.HasValue;
+            // PVC name will be modulename + volume name
+            // Volume name will be customer defined name or modulename + mount.source
+            if (this.persistentVolumeName.HasValue)
+            {
+                this.persistentVolumeName.ForEach(pvVolumeName => volumeName = pvVolumeName);
```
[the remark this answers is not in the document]
CindyXing: Sounds good. Good to see we pick up the RUST best practices.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
```diff
@@ -60,14 +60,14 @@ V1PersistentVolumeClaim ExtractPvc(KubernetesModule module, Mount mount, IDictio
             // prefer persistent volume name to storage class name, if both are set.
             if (this.persistentVolumeName.HasValue)
             {
-                this.persistentVolumeName.ForEach(volumeName => persistentVolumeClaimSpec.VolumeName = volumeName);
+                persistentVolumeClaimSpec.VolumeName = this.persistentVolumeName.OrDefault();
             }
             else if (this.storageClassName.HasValue)
             {
                 this.storageClassName.ForEach(storageClass => persistentVolumeClaimSpec.StorageClassName = storageClass);
```
darobs: If we're doing persistentVolumeName.OrDefault(), we should do that for storageClassName.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
```diff
@@ -56,17 +56,18 @@ V1PersistentVolumeClaim ExtractPvc(Mount mount, IDictionary<string, string> labe
                     Requests = new Dictionary<string, ResourceQuantity>() { { "storage", new ResourceQuantity($"{this.persistentVolumeClaimSizeMb}Mi") } }
                 },
             };
+
             // prefer persistent volume name to storage class name, if both are set.
             if (this.persistentVolumeName.HasValue)
             {
-                this.persistentVolumeName.ForEach(volumeName => persistentVolumeClaimSpec.VolumeName = volumeName);
+                persistentVolumeClaimSpec.VolumeName = this.persistentVolumeName.OrDefault();
```
philipktlin: nit: I think you can just call OrDefault without if block.

CindyXing: The if block is needed as the overwrite depends on the condition. We don't want the volumename to be none when no value exists.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
massand: I cannot see this change in the file at the moment. Hope you can accept suggestion prior to merge.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
philipktlin: The default value of "persistentVolumeClaimSpec.VolumeName" is null; that's why you can directly call this.persistentVolumeName.OrDefault(); or you should directly call this.persistentVolumeName.Value inside if block.

CindyXing: We prefer none than null. :)

philipktlin: persistentVolumeClaimSpec.VolumeName is in third-party lib, not our code.

PR #1938, opened 2019-11-01
Use PodStatus instead of ContainerStatus to create ModuleStatus

richma-ms:
ContainerStatus doesn't have a concept of "pending", as that happens at the pod level.  So pending pods end up with an unknown module status, which is not well-handled throughout the rest of the codebase.

This change makes it so that we base the module status off the pod phase, which is arguably more correct conceptually.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
@@ -6,7 +6,6 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment
     using k8s.Models;
     using Microsoft.Azure.Devices.Edge.Agent.Core;
     using Microsoft.Azure.Devices.Edge.Util;
-    using Org.BouncyCastle.Security;
     using AgentDocker = Microsoft.Azure.Devices.Edge.Agent.Docker;
     using KubernetesConstants = Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Constants;
 
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment
         public static ModuleRuntimeInfo ConvertToRuntime(this V1Pod pod, string name)
         {
             Option<V1ContainerStatus> containerStatus = GetContainerByName(name, pod);
-            ReportedModuleStatus moduleStatus = ConvertPodStatusToModuleStatus(containerStatus);
+            ReportedModuleStatus moduleStatus = ConvertPodStatusToModuleStatus(pod.Status);
             RuntimeData runtimeData = GetRuntimeData(containerStatus.OrDefault());
 
             string moduleName = string.Empty;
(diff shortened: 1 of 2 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
```diff
@@ -43,31 +43,35 @@ static Option<V1ContainerStatus> GetContainerByName(string name, V1Pod pod)
             return Option.Maybe(status);
         }
 
-        static ReportedModuleStatus ConvertPodStatusToModuleStatus(Option<V1ContainerStatus> podStatus)
+        static ReportedModuleStatus ConvertPodStatusToModuleStatus(V1PodStatus podStatus)
         {
-            return podStatus.Map(
-                pod =>
+            ModuleStatus moduleStatus = ModuleStatus.Failed;
```
dmolokanov: in which cases this default values will be used? now it is a little bit confusing

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
```diff
@@ -43,31 +43,35 @@ static Option<V1ContainerStatus> GetContainerByName(string name, V1Pod pod)
...
+            string description = "Invalid pod status";
+
+            if (podStatus?.Phase != null)
```
dmolokanov: Mike was talking about that we should reduce the usage of null in our codebase. Can you please check that we always have a podStatus and Phase in it. If yes, you can remove redundant check that only adds lines of code without bringing clarity; or change it to use `Option` type and deal with `None` explicitly.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
```diff
@@ -43,31 +43,35 @@ static Option<V1ContainerStatus> GetContainerByName(string name, V1Pod pod)
...
-                    return new ReportedModuleStatus(ModuleStatus.Unknown, "Unknown");
-                }).GetOrElse(() => new ReportedModuleStatus(ModuleStatus.Unknown, "Unknown"));
+                    case "Running":
+                        moduleStatus = ModuleStatus.Running;
+                        description = $"Started at {podStatus.StartTime}";
+                        break;
+                    case "Failed":
+                    case "Pending":
+                    case "Unknown":
+                        moduleStatus = ModuleStatus.Failed;
```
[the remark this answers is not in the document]
richma-ms: Logging is gonna be a little more involved than what I have here, as we don't instantiate logging providers in extension implementations, so any logging we want will need to happen at the calling layer.  I'm going to leave that to the bigger effort to remap ModuleStatus to k8s pod+container status that we have coming up.

---
In reply to: [341684198](https://github.com/Azure/iotedge/pull/1938#discussion_r341684198) [](ancestors = 341684198)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
```diff
@@ -5,7 +5,8 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment
     using System.Linq;
     using k8s.Models;
     using Microsoft.Azure.Devices.Edge.Agent.Core;
-    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Org.BouncyCastle.Security;
```
dmolokanov: Seems suspicious...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
```diff
@@ -43,35 +44,25 @@ static Option<V1ContainerStatus> GetContainerByName(string name, V1Pod pod)
...
+                        case "Failed":
+                        case "Pending":
+                        case "Unknown":
+                            return new ReportedModuleStatus(ModuleStatus.Failed, status.Reason);
+                        case "Succeeded":
+                            return new ReportedModuleStatus(ModuleStatus.Stopped, status.Reason);
+                        default:
+                            throw new InvalidOperationException($"Invalid pod status {status.Phase}");
+                    }
+                }).GetOrElse(() => throw new InvalidParameterException("No pod status"));
```
dmolokanov: Not sure that throwing an exception when an option is none is a very good idea. Basically, an exception describes impossible to handle a situation, right? So if you think podStatus can be None in some cases, we should handle this case in a way that it doesn't lead to a crash.

richma-ms: talked offline, changed to returning a failed status with a custom description.

---
In reply to: [341776800](https://github.com/Azure/iotedge/pull/1938#discussion_r341776800) [](ancestors = 341776800)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
richma-ms: I think VS added this, another import got added in my new iteration without me doing anything (that I known of).  

---
In reply to: [341775711](https://github.com/Azure/iotedge/pull/1938#discussion_r341775711) [](ancestors = 341775711)

richma-ms: omg it came back wtf, I removed this locally!

---
In reply to: [341783173](https://github.com/Azure/iotedge/pull/1938#discussion_r341783173) [](ancestors = 341783173,341775711)

PR #1950, opened 2019-11-04
[k8s] iotedged: Generate image pull secret for both proxy and edgeagent

dmolokanov:
Initially, we assume that proxy will live in the public container registry that maybe not the case for some customers. So edge should be able to provide all credentials for k8s to pull and run proxy as a sidecar container.
- Refactor `config.yaml` file structure to support proxy settings
- Generate image pull secret for both EdgeAgent and Proxy
- Pass proxy image pull secret to EdgeAgent in order for other modules to have proxy pulled

### DIFF
--- a/edgelet/edgelet-kube/src/registry/pull.rs
+++ b/edgelet/edgelet-kube/src/registry/pull.rs
@@ -1,0 +1,113 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use std::collections::hash_map::Values;
+use std::collections::HashMap;
+use std::convert::TryFrom;
+
+use failure::Fail;
+use futures::future::Either;
+use futures::prelude::*;
+use futures::{future, Future, Stream};
+use hyper::service::Service;
+use hyper::Body;
+
+use edgelet_docker::DockerConfig;
+use kube_client::TokenSource;
+
+use crate::convert::NamedSecret;
+use crate::error::{Error, ErrorKind};
+use crate::registry::ImagePullSecret;
+use crate::KubeModuleRuntime;
+
+pub fn create_image_pull_secrets<T, S>(
+    runtime: &KubeModuleRuntime<T, S>,
+    config: &DockerConfig,
+) -> impl Future<Item = (), Error = Error>
+where
+    T: TokenSource + Send + 'static,
+    S: Service + Send + 'static,
+    S::ReqBody: From<Vec<u8>>,
+    S::ResBody: Stream,
+    Body: From<S::ResBody>,
+    S::Error: Fail,
+    S::Future: Send,
+{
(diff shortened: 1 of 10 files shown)


edgelet/edgelet-kube/src/registry/pull.rs
```diff
@@ -0,0 +1,362 @@
...
+{
+    let image_pull_secrets = vec![config.auth(), runtime.settings().proxy().auth()]
+        .into_iter()
+        .filter_map(|auth| {
+            auth.and_then(|auth| ImagePullSecret::from_auth(&auth))
+                .map(|secret| (secret.name(), secret))
+        })
+        .collect::<HashMap<_, _>>()
+        .drain()
+        .map(|(_, secret)| secret)
```
darobs: Possible suggestion: 
```
use itertools::Itertools;

let image_pull_secrets = vec![config.auth(), runtime.settings().proxy().auth()]
        .into_iter()
        .filter_map(|auth| {
            auth.and_then(|auth| ImagePullSecret::from_auth(&auth))
        })
        .unique_by(|secret| secret.name())
        .collect::<Vec<_>>();
```

kubernetes/charts/edge-kubernetes/templates/_helpers.tpl
```diff
@@ -78,13 +78,22 @@ listen:
...
-proxy_config_path: "/etc/iotedge-proxy"
-proxy_config_map_name: "iotedged-proxy-config"
-proxy_trust_bundle_path: "/etc/trust-bundle"
-proxy_trust_bundle_config_map_name: "iotedged-proxy-trust-bundle"
-image_pull_policy: {{ .Values.iotedgedProxy.image.pullPolicy | quote }}
 device_hub_selector: ""
+proxy:
+  image: "{{.Values.iotedgedProxy.image.repository}}:{{.Values.iotedgedProxy.image.tag}}"
+  image_pull_policy: {{ .Values.iotedgedProxy.image.pullPolicy | quote }}
+  {{- if .Values.edgeAgent.registryCredentials }}
```
darobs: We are assuming that the proxy registry==agent registry?

edgelet/edgelet-kube/src/registry/pull.rs
dmolokanov: I like this idea, but do we want to add additional crate?

kubernetes/charts/edge-kubernetes/templates/_helpers.tpl
dmolokanov: nope. mistake here

edgelet/edgelet-kube/src/registry/pull.rs
darobs: It's not a required change, I just like the fact I found a more terse way to do the same thing. :)

PR #1952, opened 2019-11-04
[k8s] EA: Generate image pull secret for both proxy and edgeagent

dmolokanov:
Initially, we assume that proxy will live in the public container registry that maybe not the case for some customers. So edge should be able to provide all credentials for k8s to pull and run proxy as a sidecar container.
- If image pull secret is set via ENV apply it to the pod spec

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
@@ -175,8 +175,8 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Service
                         apiVersion = configuration.GetValue<string>(Constants.EdgeletApiVersionVariableName);
                         iothubHostname = configuration.GetValue<string>(Constants.IotHubHostnameVariableName);
                         deviceId = configuration.GetValue<string>(Constants.DeviceIdVariableName);
-                        string networkId = configuration.GetValue<string>(Constants.NetworkIdKey);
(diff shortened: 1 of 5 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
```diff
@@ -175,8 +175,8 @@ public static async Task<int> MainAsync(IConfiguration configuration)
                         apiVersion = configuration.GetValue<string>(Constants.EdgeletApiVersionVariableName);
                         iothubHostname = configuration.GetValue<string>(Constants.IotHubHostnameVariableName);
                         deviceId = configuration.GetValue<string>(Constants.DeviceIdVariableName);
-                        string networkId = configuration.GetValue<string>(Constants.NetworkIdKey);
                         string proxyImage = configuration.GetValue<string>(K8sConstants.ProxyImageEnvKey);
+                        Option<string> proxyImagePullSecretName = configuration.GetValue<Option<string>>(K8sConstants.ProxyImagePullSecretNameEnvKey);
```
darobs: Will this work the same as `Option.Maybe(configuration.GetValue<string>(K8sConstants.ProxyImagePullSecretNameEnvKey)`?

dmolokanov: Yes, it should. If there is no value found it will return default(Option<string>) which is None

PR #1953, opened 2019-11-04
Update dotnetcore to 2.1.13

ancaantochi:
dotnet core updated to 2.1.13 on all platforms except Linux arm64 where it was updated to 3.0.0


edge-agent/docker/linux/arm64v8/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.0-preview014-linux-arm64v8
+ARG base_tag=1.0.0-preview015-linux-arm64v8
```
varunpuranik: We should settle on a better and consistent tagging..

PR #1963, opened 2019-11-06
Update dotnetcore version

ancaantochi:
dotnet core updated to 2.1.13 for all platforms except arm64 where is 3.0.0


varunpuranik:
Is this a cherry-pick of the 1.0.8 PR, or anything different in this PR that we need to be aware of?

PR #1967, opened 2019-11-07
Fix formatting and clippy for new Rust 1.39 stable.

yophilav:
- Fix formatting and clippy for new Rust 1.39 stable.
- Update `rust.toolchain` to `nightly-2019-11-06` because the clippy in `nightly-2019-07-05` mysteriously disappeared.


arsing:
When you merge, please remember to mention in the commit message the reason for bumping the nightly used by the code coverage job. (Something like "Also bump the nightly used by the code coverage job because the previous one's clippy component inexplicably disappeared from rust-lang's dist server.")

PR #1968, opened 2019-11-07
Get disk space for arm32 and print support bundle full location

lfitchett:


arsing:
(Build break being fixed in #1967 )

PR #1973, opened 2019-11-08
Analyzer - Fix device id default to device1 if not set 

and-rewsmith:
If device id was not set, the analyzer would take it's value from the settings.json. 

The environment variable IOTEDGE_DEVICEID will always be set within a module, so we don't have to look at the settings config.

### DIFF
--- a/edge-modules/MessagesAnalyzer/Settings.cs
+++ b/edge-modules/MessagesAnalyzer/Settings.cs
@@ -10,14 +10,13 @@ namespace MessagesAnalyzer
     {
         const string ExcludeModulesIdsPropertyName = "ExcludeModules:Ids";
         const string EventHubConnectionStringPropertyValue = "eventHubConnectionString";
-        const string DeviceIdPropertyName = "DeviceId";
+        const string DeviceIdPropertyName = "IOTEDGE_DEVICEID";
         const string ConsumerGroupIdPropertyName = "ConsumerGroupId";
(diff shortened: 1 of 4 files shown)


edge-modules/MessagesAnalyzer/Settings.cs
```diff
@@ -32,21 +32,21 @@ class Settings
...
                 return new Settings(
                     configuration.GetValue<string>(EventHubConnectionStringPropertyValue),
-                    configuration.GetValue(ConsumerGroupIdPropertyName, DefaultConsumerGroupId),
-                    configuration.GetValue(DeviceIdPropertyName, DefaultDeviceId),
+                    configuration.GetValue<string>(ConsumerGroupIdPropertyName, DefaultConsumerGroupId),
+                    configuration.GetValue<string>(DeviceIdPropertyName),
                     excludedModules,
-                    configuration.GetValue(WebhostPortPropertyName, DefaultWebhostPort),
-                    configuration.GetValue(ToleranceInMillisecondsPropertyName, DefaultToleranceInMilliseconds));
+                    configuration.GetValue<string>(WebhostPortPropertyName, DefaultWebhostPort),
```
philipktlin: legacy issue. And even better if we validate it between 0 and 65535.  Btw I approved this PR, your call if you want to fix it this PR or not.
```suggestion
                    configuration.GetValue<int>(WebhostPortPropertyName, DefaultWebhostPort),
```

PR #1977, opened 2019-11-08
Dashboard: update Build Management API

philipktlin:
Update Management API to support retrieving more than 1 build results by build definition Ids.


tools/IoTEdgeDevOps/IoTEdgeDashboard/Startup.cs
```diff
@@ -57,7 +57,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                 {
                     routes.MapRoute(
                         name: "default",
-                        template: "{controller=Agent}/{action=Index}/{id?}");
+                        template: "{controller=Home}/{action=Index}");
```
yophilav: Should we consider using AspNet session feature? 
https://www.c-sharpcorner.com/UploadFile/484ad3/session-state-in-Asp-Net/

philipktlin: Please specify what do you want to use AspNet session for?

yophilav: Per the document, I can foresee using a session as a cache as we enable more components to the dashboard. The session is not relevant for our application at the moment.

philipktlin: Ok, for user session, I think this may not be something applicable to use at least for now.  For caching, we can think about it later on and handle in another PBI.

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
```diff
@@ -10,9 +12,9 @@ namespace DevOpsLib
 
     public class BuildManagement
```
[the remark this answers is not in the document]
philipktlin: I will add API endpoint online doc link here.

PR #1980, opened 2019-11-12
[k8s] Start EdgeAgent container as non root by default

dmolokanov:
Make sure that EdgeAgent container runs a user with id=1000 and indicate it is non-root user by default.


edgelet/edgelet-kube/src/convert/named_secret.rs
```diff
@@ -25,10 +25,10 @@ impl NamedSecret {
     }
 }
 
-impl<'a> TryFrom<(&str, ImagePullSecret<'a>)> for NamedSecret {
+impl TryFrom<(String, ImagePullSecret)> for NamedSecret {
```
darobs: What's the reason behind moving from a &str to String?

dmolokanov: In the first commit, I pay a technical debt to not following rustlang guidelines. `From<T>`/`TryFrom<T>` implementation should consume `T` and "deconstruct" object. This way we delegate a caller a decision to get owned version of a `T` instead of making cloning parts of implementation of `T` inside `From<T>`/`TryFrom<T>`

kubernetes/charts/edge-kubernetes/templates/iotedged-deployment.yaml
```diff
@@ -63,6 +63,9 @@ spec:
       imagePullSecrets:
         - name: regcreds
       {{ end }}
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
```
darobs: :(

edgelet/edgelet-kube/src/registry/pull.rs
```diff
@@ -64,7 +64,7 @@ where
 
     let named_secrets: Vec<_> = image_pull_secrets
         .into_iter()
-        .map(|secret| NamedSecret::try_from((runtime.settings().namespace(), secret)))
+        .map(|secret| NamedSecret::try_from((runtime.settings().namespace().to_string(), secret)))
```
darobs: Chose not to do into() here?

PR #1983, opened 2019-11-12
Enable dynamic reprovisioning in EdgeAgent based on connection state

shantanu1singh:
This PR if a follow up from an earlier pull-request #1842  where the setup for supporting dynamic reprovisioning in IoT Edge was done.

With this change, the feature will be activated as EdgeAgent will be able to trigger the device reprovisioning flow based on the appropriate `ConnectionStatusChangeReason` reported by IoT hub indicating that the device has been deleted in IoT hub and could have been reprovisioned.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -139,15 +139,25 @@ async void OnConnectionStatusChanged(ConnectionStatus status, ConnectionStatusCh
...
-                //    await this.deviceManager.ReprovisionDeviceAsync();
-                // }
+                // Notify the IoT Edge daemon that a device has been deprovisioned and it should check
+                // if the device has been provisioned to a different IoT hub instead.
+                // When the Amqp or AmqpWs protocol is used, the SDK returns a connection status change reason of
+                // Device_Disabled when a device is either disabled or deleted in IoT hub.
+                // For the Mqtt and MqttWs protocol however, the SDK returns a Bad_Credential status as it's not
+                // possible for IoT hub to distinguish between 'device does not exist', 'device is disabled' and
+                // 'device exists but wrong credentials were supplied' cases.
+                if ((reason == ConnectionStatusChangeReason.Device_Disabled &&
```
varunpuranik: Do we really need to do this? Why not simply check for reason Bad_Credential or Device_Disabled? When does AMQP throw Bad_Credential? Wouldn't we want to re-provision in that case?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -139,15 +139,25 @@ async void OnConnectionStatusChanged(ConnectionStatus status, ConnectionStatusCh
...
+                    (protocol == UpstreamProtocol.Amqp || protocol == UpstreamProtocol.AmqpWs)) ||
+                    (reason == ConnectionStatusChangeReason.Bad_Credential &&
+                    (protocol == UpstreamProtocol.Mqtt || protocol == UpstreamProtocol.MqttWs)))
+                {
+                    await this.deviceManager.ReprovisionDeviceAsync();
```
varunpuranik: Where is DeviceManager implemented?

shantanu1singh: In `ModuleManagementHttpClient`. That work was done in an earlier PR

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
shantanu1singh: The reason I did that was to be specific about the cases where we do call Reprovision.

I'm not entirely sure when Amqp can return a Bad_Credential error. Will need to dig into the SDK or check with the SDK team.

shantanu1singh: Took a cursory glance at the SDK and IoT hub and it seems like Hub will return UnauthorizedAccessException(s) for a bunch of cases such as SAS token expiry, cert auth failure (not really applicable here). Do you think it's a safer option to only look at Bad_Credential for Mqtt only in that case (since there's no way of Hub to distinguish between device disable/deletion/bad sas token for Mqtt only).

I can change the current condition for Device_Disabled to ignore the protocol check (for Amqp) though.

PR #1992, opened 2019-11-14
Increase delay/retry for flaky test in CI build

philipktlin:
Recently there are some flaky tests reported failure randomly.  Therefore increase delay/retry for short term fix; if this happens again, we should move these tests to e2e tests or take them out from CI build.

### DIFF
--- a/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
+++ b/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test
     {
         static readonly TimeSpan ClockSkew = TimeSpan.FromMinutes(5);
 
-        static readonly int EventHubMessageReceivedRetry = 5;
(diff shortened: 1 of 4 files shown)


edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
```diff
@@ -222,7 +222,7 @@ static async Task CheckMessageInEventHub(IList<IMessage> sentMessages, DateTime
             var eventHubReceiver = new EventHubReceiver(eventHubConnectionString);
             var cloudMessages = new List<EventData>();
             bool messagesFound = false;
-            // Add retry mechanism to make sure all the messages sent reached Event Hub. Retry 3 times.
+            // Add retry mechanism to make sure all the messages sent reached Event Hub.
```
and-rewsmith: Is there a reason why this is a retry, and not a Task.Delay?

philipktlin: eventHubReceiver.GetMessagesFromAllPartitions already did Task.Delay and it also in Line 235.  Therefore original dev want to retry after some time (20s) again.  This is my best guess why original dev did this.

PR #1999, opened 2019-11-15
Set the backup path to a writeable location.

darobs:
changed the location of iotedged

This change will prevent edgeAgent from complaining about not being able to write to /app/backup.json

### DIFF
--- a/kubernetes/charts/edge-kubernetes/values.yaml
+++ b/kubernetes/charts/edge-kubernetes/values.yaml
@@ -1,7 +1,7 @@
 # iotedged configuration
 iotedged:
   image:
-    repository: azureiotedge/iotedged
+    repository: azureiotedge/azureiotedge-iotedged
     tag: 0.1.0-beta1
     pullPolicy: Always
   service:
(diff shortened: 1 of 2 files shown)


kubernetes/charts/edge-kubernetes/values.yaml
```diff
@@ -59,6 +59,8 @@ edgeAgent:
     # Set this to one of "LoadBalancer", "NodePort", or "ClusterIP" to tell the
     # IoT Edge runtime how you want to expose mapped ports as Services.
     portMappingServiceType: 'ClusterIP'
+    # Set the location of the backup configuration file.
```
dmolokanov: Nit: new line befor line 62 is missing

kubernetes/charts/edge-kubernetes/values.yaml
```diff
@@ -1,7 +1,7 @@
 # iotedged configuration
 iotedged:
   image:
-    repository: azureiotedge/iotedged
+    repository: azureiotedge/azureiotedge-iotedged
     tag: 0.1.0-beta1
```
ggjjj: 👍

kubernetes/charts/edge-kubernetes/values.yaml
darobs: Done

PR #2001, opened 2019-11-16
Introduce LogAnalytics and AzureActiveDirectory for IoT Edge Dashboard

yophilav:
- Introduce Log Analytics and Azure Active Directory for IoT Edge Dashboard


tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+    public sealed class AzureActiveDirectory
+    {
+        static readonly AzureActiveDirectory instance = new AzureActiveDirectory();
+
+        static string azureActiveDirTenant = null;
+        static string azureActiveDirClientId = null;
+        static string azureActiveDirClientSecret = null;
+        static string azureResource = null;
+        static string accessToken = null;
+        static DateTime accessTokenExpiration = new DateTime(1970, 01, 01, 0, 0, 0, 0, DateTimeKind.Utc);
```
philipktlin: why need to use this specific value, not DateTime.MinValue?

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+        // not to mark type as beforefieldinit
+        static AzureActiveDirectory()
+        {
+        }
+
+        AzureActiveDirectory()
+        {
+        }
+
+        public static AzureActiveDirectory Instance
```
philipktlin: nit: Expression-bodied member

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+                    var responseJson = JObject.Parse(responseMsg);
+                    AzureActiveDirectory.accessTokenExpiration = DateTime.UtcNow.AddSeconds((double)responseJson["expires_on"]);
+                    AzureActiveDirectory.accessToken = (string)responseJson["access_token"];
+                }
+
+                return AzureActiveDirectory.accessToken;
+            }
+            catch (Exception e)
+            {
+                throw e;
```
philipktlin: why catch and throw again without doing anything?

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+    using System.Net.Http;
+    using System.Net.Http.Headers;
+    using System.Threading.Tasks;
+    using Newtonsoft.Json.Linq;
+
+    public sealed class AzureActiveDirectory
+    {
+        static readonly AzureActiveDirectory instance = new AzureActiveDirectory();
+
+        static string azureActiveDirTenant = null;
```
philipktlin: You don't need these fields; please remove them.

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+            string azureResource)
+        {
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirTenant, nameof(azureActiveDirTenant));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientId, nameof(azureActiveDirClientId));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientSecret, nameof(azureActiveDirClientSecret));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureResource, nameof(azureResource));
+
+            try
+            {
+                if ((DateTime.Compare(DateTime.UtcNow, AzureActiveDirectory.accessTokenExpiration) >= 0) ||
```
philipktlin: you need to have lock statement to avoid race condition.

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                    var requestBody = new List<KeyValuePair<string, string>>();
+                    requestBody.Add(new KeyValuePair<string, string>("client_id", azureActiveDirClientId));
+                    requestBody.Add(new KeyValuePair<string, string>("client_secret", azureActiveDirClientSecret));
+                    requestBody.Add(new KeyValuePair<string, string>("grant_type", grantType));
+                    requestBody.Add(new KeyValuePair<string, string>("resource", azureResource));
+
+                    HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, "");
+                    // By default, if FormUrlEncodedContent() is used, the "Content-Type" is set to "application/x-www-form-urlencoded"
+                    // request.Content.Headers = new MediaTypeHeaderValue("application/x-www-form-urlencoded");
```
philipktlin: do you need this line? if not, remove it.

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                if ((DateTime.Compare(DateTime.UtcNow, AzureActiveDirectory.accessTokenExpiration) >= 0) ||
+                   (AzureActiveDirectory.accessToken == null) ||
+                   (AzureActiveDirectory.azureResource != azureResource))
+                {
+                    string requestUri = $"https://login.microsoftonline.com/{azureActiveDirTenant}/oauth2/token";
+                    const string grantType = "client_credentials";
+
+                    var client = new HttpClient();
+                    client.BaseAddress = new Uri(requestUri);
```
philipktlin: why use client.BaseAddress, but not pass requestUri in HttpRequestMessage constructor?

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                    var requestBody = new List<KeyValuePair<string, string>>();
+                    requestBody.Add(new KeyValuePair<string, string>("client_id", azureActiveDirClientId));
+                    requestBody.Add(new KeyValuePair<string, string>("client_secret", azureActiveDirClientSecret));
+                    requestBody.Add(new KeyValuePair<string, string>("grant_type", grantType));
+                    requestBody.Add(new KeyValuePair<string, string>("resource", azureResource));
+
+                    HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, "");
```
philipktlin: can you use HttpRequestMessage contructor with single argument since 2nd parameter is empty string?

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirTenant, nameof(azureActiveDirTenant));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientId, nameof(azureActiveDirClientId));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientSecret, nameof(azureActiveDirClientSecret));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureResource, nameof(azureResource));
+
+            try
+            {
+                if ((DateTime.Compare(DateTime.UtcNow, AzureActiveDirectory.accessTokenExpiration) >= 0) ||
+                   (AzureActiveDirectory.accessToken == null) ||
+                   (AzureActiveDirectory.azureResource != azureResource))
```
philipktlin: use string.Equals with string comparison option. https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=netframework-4.8#System_String_Equals_System_String_System_String_System_StringComparison_

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+                    requestBody.Add(new KeyValuePair<string, string>("client_secret", azureActiveDirClientSecret));
+                    requestBody.Add(new KeyValuePair<string, string>("grant_type", grantType));
+                    requestBody.Add(new KeyValuePair<string, string>("resource", azureResource));
+
+                    HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, "");
+                    // By default, if FormUrlEncodedContent() is used, the "Content-Type" is set to "application/x-www-form-urlencoded"
+                    // request.Content.Headers = new MediaTypeHeaderValue("application/x-www-form-urlencoded");
+                    request.Content = new FormUrlEncodedContent(requestBody);
+
+                    var response = await client.SendAsync(request).ConfigureAwait(false);
```
philipktlin: nit: use var when you can determine the type when looking at right side.  For example, var x = new object();

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+                    response.EnsureSuccessStatusCode();
+                    var responseMsg = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+
+                    AzureActiveDirectory.azureActiveDirTenant = azureActiveDirTenant;
```
philipktlin:
```suggestion
                    this.azureActiveDirTenant = azureActiveDirTenant;
```

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                return AzureActiveDirectory.accessToken;
+            }
+            catch (Exception e)
+            {
+                throw e;
+            }
+        }
+
+        public Task<string> GetAccessToken() =>
```
philipktlin: Why do you need this method?

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace DevOpsLib
+{
+    using System;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+
+    public sealed class AzureLogAnalytics
```
philipktlin: I like "sealed"

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        static readonly AzureLogAnalytics instance = new AzureLogAnalytics();
+        static readonly string apiVersion = "v1";
```
philipktlin: use const

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            get
+            {
+                return instance;
+            }
+        }
+
+        // Use get-request to do a Kusto query
+        // API reference: https://dev.loganalytics.io/documentation/Using-the-API/RequestFormat
+        public async Task<string> GetKqlQuery(
```
philipktlin:
```suggestion
        public async Task<string> RunQuery(
```

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,57 @@
...
+            string logAnalyticsWorkspaceId,
```
philipktlin: validate inputs

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace DevOpsLib
+{
+    using System;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+
+    public sealed class AzureLogAnalytics
+    {
+        static readonly AzureLogAnalytics instance = new AzureLogAnalytics();
```
philipktlin: why do you want to make this as singleton?

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
philipktlin: I think you may need to add buffer in this checking? otherwise, access token may be expired right after this checking.  And the caller use this expired token, will get exception.

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            try
+            {
+                string requestUri = $"https://api.loganalytics.io/{AzureLogAnalytics.apiVersion}/workspaces/{logAnalyticsWorkspaceId}/query?query=";
+                string accessToken = await AzureActiveDirectory.Instance.GetAccessToken();
+
+                var client = new HttpClient();
+                client.DefaultRequestHeaders.Add("Authorization", $"Bearer {accessToken}");
+
+                var response = await client.GetAsync(requestUri + kqlQuery).ConfigureAwait(false);
```
philipktlin: use string interpolation
```suggestion
                var response = await client.GetAsync($"{requestUri}{kqlQuery}").ConfigureAwait(false);
```

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
yophilav: It is actually doing the same thing. 
But I'd prefer to do 
` client.BaseAddress = new Uri(requestUri); ` then later follow by `HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, "\some\path\endpoint");`  I think it's more readable that you're sending a `BaseAddress `with `\some\path\endpoint `

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
yophilav: Nope, it's either no argument (create an empty request, not sure what method) or specific a method with endpoint. In my case, the endpoint is just `/`

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
yophilav: This is a comment to clarify of what I'm doing with the content header. (Maybe I should update the comment to be more explicit)

HttpClient is weird, when you do `FormUrlEncodedContent()` the default type of it is `"application/x-www-form-urlencoded"` which is correct. But when you explicitly setting it `request.Content.Headers = new MediaTypeHeaderValue("application/x-www-form-urlencoded");` you will get a header error. This is a known issue in their repo.

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
yophilav: Ohh I see. I'll update that. Thank you!

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
yophilav: I'd very much prefer to have the ADD config setup once.

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
yophilav: Just for the completeness that you can just get the token once it is set.

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
yophilav: Thank you

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
yophilav: I could.  Previously it was couple with ADD which SHOULD really be a singleton.
I can make change to this so that it is not a singleton.

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
yophilav: `const`

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
yophilav: There are multiple ways to run query using Log Analytic endpoint API. 
You can also do `POST` kusto query.  Currently, we are using `GET`; hence, GetKqlQuery()

yophilav:
I was hoping to use `ConcurrentDictionary<azureResource, (accessToken,expiration)>` to keep track of all tokens under the ADD. But I ran into a concurrency issue with checking and await for response. 

Say I have the following code:
```
ConcurrentDictionary<string, AccessToken> tokenStorage;
if((tokenStorage.IsEmpty) || (!this.IsAccessTokenValid(azureResource)))
{
   ...
HttpResponseMessage response = await client.SendAsync(request).ConfigureAwait(false);
var responseJson = JObject.Parse(responseMsg);
DateTime accessTokenExpiration = DateTime.UtcNow.AddSeconds((double)responseJson["expires_on"] - 1);
string accessToken = (string)responseJson["access_token"];
this.tokenStorage.AddOrUpdate(azureResource,
    new AccessToken(accessToken, accessTokenExpiration),
    (k, v) => new AccessToken(accessToken, accessTokenExpiration));
}

public bool IsAccessTokenValid(string azureResource)
{
    AccessToken accessToken;
    if(this.tokenStorage.TryGetValue(azureResource, out accessToken) &&
        (accessToken != null))
    {
        //Console.WriteLine($"-------------------------IsVAlid: {accessToken.accessToken[0]}  {accessToken.accessTokenExpiration} ");
        return DateTime.Compare(DateTime.UtcNow, accessToken.accessTokenExpiration) >= 0;
    }
    Console.WriteLine($"-------------------------IsVAlid: {this.tokenStorage.TryGetValue(azureResource, out accessToken)}  {(accessToken != null)} ");
    return false;
}
```
Often the `IsAccessTokenValid(string azureResource)` would think that the `azureResource` does not have an accessKey, because the previous request on the token for the same azureResource was not yet responded. I don't have a good idea on how to solve the problem yet.

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,92 @@
...
+        string azureActiveDirTenant = null;
+        string azureActiveDirClientId = null;
+        string azureActiveDirClientSecret = null;
+
+        string azureResource = null;
+        string accessToken = null;
+        DateTime accessTokenExpiration = new DateTime(DateTime.MinValue.Ticks);
+        readonly object locker = new object();
+
+        public AzureActiveDirectory(string azureActiveDirTenant,
```
philipktlin:
```suggestion
        public AzureActiveDirectory(
            string azureActiveDirTenant,
```

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace DevOpsLib
+{
+    using System;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+
+    public sealed class AzureLogAnalytics
+    {
+        public static string AzureResource = "https://api.loganalytics.io";
```
philipktlin:
```suggestion
        const string AzureResource = "https://api.loganalytics.io";
```

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+        string accessToken = null;
+        DateTime accessTokenExpiration = new DateTime(DateTime.MinValue.Ticks);
+        readonly object locker = new object();
+
+        public AzureActiveDirectory(
+            string azureActiveDirTenant,
+            string azureActiveDirClientId,
+            string azureActiveDirClientSecret)
+        {
+            this.azureActiveDirTenant = azureActiveDirTenant;
```
philipktlin: validation inputs

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    public sealed class AzureActiveDirectory
+    {
+        string azureActiveDirTenant = null;
+        string azureActiveDirClientId = null;
+        string azureActiveDirClientSecret = null;
+
+        string azureResource = null;
+        string accessToken = null;
+        DateTime accessTokenExpiration = new DateTime(DateTime.MinValue.Ticks);
```
philipktlin: why not?
```suggestion
        DateTime accessTokenExpiration = DateTime.MinValue;
```

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+            string azureActiveDirTenant,
+            string azureActiveDirClientId,
+            string azureActiveDirClientSecret)
+        {
+            this.azureActiveDirTenant = azureActiveDirTenant;
+            this.azureActiveDirClientId = azureActiveDirClientId;
+            this.azureActiveDirClientSecret = azureActiveDirClientSecret;
+        }
+
+        // GetAccessToken(4) is required to be called before invoking GetAccessToken(0), or GetAccessToken(1)
```
philipktlin: what does (4) mean?

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+        // Trigger Azure Active Directory (AAD) for an OAuth2 client credential for an azure resource access.
+        // API reference: https://dev.loganalytics.io/documentation/Authorization/OAuth2
+        public async Task<string> GetAccessToken(string azureResource)
+        {
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirTenant, nameof(azureActiveDirTenant));
```
philipktlin: move first 3 lines to constructor

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+    using System.Collections.Generic;
+    using System.Net.Http;
+    using System.Net.Http.Headers;
+    using System.Reflection.Metadata.Ecma335;
+    using System.Threading.Tasks;
+    using Newtonsoft.Json.Linq;
+
+    public sealed class AzureActiveDirectory
+    {
+        string azureActiveDirTenant = null;
```
philipktlin: mark as readonly

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirTenant, nameof(azureActiveDirTenant));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientId, nameof(azureActiveDirClientId));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientSecret, nameof(azureActiveDirClientSecret));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureResource, nameof(azureResource));
+
+            try
+            {
+                
+                if (this.IsAccessTokenExpired() ||
```
philipktlin: nit: wrap all these condition in 1 method named "RequireAccessToken" or "RefreshAccessToken".

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+                    (this.accessToken == null) ||
+                    (!this.azureResource.Equals(azureResource, StringComparison.OrdinalIgnoreCase)))
```
philipktlin: are you going to cache more than 1 resource? if so, please put todo here.

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,42 @@
...
+        // Use get-request to do a Kusto query
+        // API reference: https://dev.loganalytics.io/documentation/Using-the-API/RequestFormat
+        public async Task<string> GetKqlQuery(
+            AzureActiveDirectory azureActiveDirectory,
+            string logAnalyticsWorkspaceId,
+            string kqlQuery)
+        {
+            ValidationUtil.ThrowIfNullOrWhiteSpace(logAnalyticsWorkspaceId, nameof(logAnalyticsWorkspaceId));
+
+            try
```
philipktlin: nit: remove try-catch

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+        // Trigger Azure Active Directory (AAD) for an OAuth2 client credential for an azure resource access.
+        // API reference: https://dev.loganalytics.io/documentation/Authorization/OAuth2
+        public async Task<string> GetAccessToken(string azureResource)
+        {
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirTenant, nameof(azureActiveDirTenant));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientId, nameof(azureActiveDirClientId));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientSecret, nameof(azureActiveDirClientSecret));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(azureResource, nameof(azureResource));
+
+            try
```
philipktlin: nit: remove try-catch

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+            catch (Exception e)
+            {
+                throw e;
+            }
+        }
+
+        public bool IsAccessTokenExpired()
+        {
+            return DateTime.Compare(DateTime.UtcNow, this.accessTokenExpiration) >= 0;
```
philipktlin: Add buffer to this checking.

philipktlin: I think it is easier to read (with buffer, you can make buffer timespan as a readonly variable).

e.g. (((DateTime.UtcNow.AddDays(3) - DateTime.UtcNow) - new TimeSpan(0,0,0,5)).Ticks > 0)

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+                    HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, "");
+                    // By default, if FormUrlEncodedContent() is used, the "Content-Type" is set to "application/x-www-form-urlencoded"
+                    // which can be explicitly written : request.Content.Headers = new MediaTypeHeaderValue("application/x-www-form-urlencoded");
+                    request.Content = new FormUrlEncodedContent(requestBody);
+
+                    HttpResponseMessage response = await client.SendAsync(request).ConfigureAwait(false);
+                    response.EnsureSuccessStatusCode();
+                    string responseMsg = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+
+                    lock (locker)
```
philipktlin: you need to lock before if block

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+                    {
+                        var responseJson = JObject.Parse(responseMsg);
```
philipktlin: specify type on left side.

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,93 @@
...
+                        this.azureResource = azureResource;
+                        this.accessTokenExpiration = DateTime.UtcNow.AddSeconds((double)responseJson["expires_on"] - 1);
```
philipktlin: why -1?

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -11,83 +11,86 @@ namespace DevOpsLib
...
 
-        // GetAccessToken(4) is required to be called before invoking GetAccessToken(0), or GetAccessToken(1)
         // Trigger Azure Active Directory (AAD) for an OAuth2 client credential for an azure resource access.
         // API reference: https://dev.loganalytics.io/documentation/Authorization/OAuth2
         public async Task<string> GetAccessToken(string azureResource)
         {
-            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirTenant, nameof(azureActiveDirTenant));
-            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientId, nameof(azureActiveDirClientId));
-            ValidationUtil.ThrowIfNullOrWhiteSpace(azureActiveDirClientSecret, nameof(azureActiveDirClientSecret));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(this.azureActiveDirTenant, nameof(this.azureActiveDirTenant));
```
philipktlin: remove these validation since they are already done in constructor.

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -11,83 +11,86 @@ namespace DevOpsLib
...
+                string responseMsg = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
 
-                    lock (locker)
-                    {
-                        var responseJson = JObject.Parse(responseMsg);
-                        this.azureResource = azureResource;
-                        this.accessTokenExpiration = DateTime.UtcNow.AddSeconds((double)responseJson["expires_on"] - 1);
-                        this.accessToken = (string)responseJson["access_token"];
-                    }
+                lock (locker)
```
philipktlin: Need to move this lock statement above if block (line 47); and you need to Semaphore. https://blog.cdemi.io/async-waiting-inside-c-sharp-locks/

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public static string AzureResource = "https://api.loganalytics.io";
+        const string apiVersion = "v1";
+
+        // Use get-request to do a Kusto query
+        // API reference: https://dev.loganalytics.io/documentation/Using-the-API/RequestFormat
+        public async Task<string> GetKqlQuery(
+            AzureActiveDirectory azureActiveDirectory,
+            string logAnalyticsWorkspaceId,
+            string kqlQuery)
```
philipktlin: validate kqlQuery

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,99 @@
...
+            finally
+            {
+                mutexSlim.Release();
+            }
+        }
+
+        public bool IsAccessTokenExpired()
+        {
+            // Add 2 seconds as a buffer so the token does not immediately expire after the function checked.
+            return DateTime.Compare(DateTime.UtcNow.AddSeconds(2), this.accessTokenExpiration) >= 0;
```
philipktlin: nit: you can use minus operation, i think it is easier to read.
```suggestion
            return DateTime.UtcNow.AddSeconds(2) > this.accessTokenExpiration;
```

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,99 @@
...
+
+                    HttpResponseMessage response = await client.SendAsync(request).ConfigureAwait(false);
+                    response.EnsureSuccessStatusCode();
+                    string responseMsg = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+
+                    JObject responseJson = JObject.Parse(responseMsg);
+                    this.azureResource = azureResource;
+                    // (-1) second on the expiration time to prevent the access token immediately expired
+                    // after this function is returned.
+                    this.accessTokenExpiration = epochStart.AddSeconds((double)responseJson["expires_on"]);
```
philipktlin: https://docs.microsoft.com/en-us/dotnet/api/system.datetime.-ctor?view=netframework-4.8#System_DateTime__ctor_System_Int64_System_DateTimeKind_

tools/IoTEdgeDevOps/DevOpsLib/AzureActiveDirectory.cs
```diff
@@ -0,0 +1,99 @@
...
+                    // which can be explicitly written : request.Content.Headers = new MediaTypeHeaderValue("application/x-www-form-urlencoded");
+                    request.Content = new FormUrlEncodedContent(requestBody);
+
+                    HttpResponseMessage response = await client.SendAsync(request).ConfigureAwait(false);
+                    response.EnsureSuccessStatusCode();
+                    string responseMsg = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+
+                    JObject responseJson = JObject.Parse(responseMsg);
+                    this.azureResource = azureResource;
+                    // (-1) second on the expiration time to prevent the access token immediately expired
```
philipktlin: remove this comment

tools/IoTEdgeDevOps/DevOpsLib/AzureLogAnalytics.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            ValidationUtil.ThrowIfNull(azureActiveDirectory, nameof(azureActiveDirectory));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(logAnalyticsWorkspaceId, nameof(logAnalyticsWorkspaceId));
+
+            if (string.IsNullOrWhiteSpace(kqlQuery))
+            {
+                return "";
```
philipktlin: move this if at the top of this function; and use "return string.Empty".

PR #2004, opened 2019-11-18
Roll SDK version with pipeline fixes

and-rewsmith:
Pipeline fix: port ca-certs.ps1 from master


and-rewsmith:
SDK changes not going into 109rc3 so I am closing.

PR #2008, opened 2019-11-18
[k8s] Add pod security context to create options

dmolokanov:
Provide an option for user to specify pod security context.
To verify changes:
```
1. install edge on Kubernetes
2. deploy a module with default parameters and default image.
3. verify the pod is up and running
4. apply the pod security context as a createOptions section on a portal
5. wait until agent reconcile changes
6. check new pod for the deployment is in a failed state.
```


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -132,9 +132,10 @@ V1PodTemplateSpec GetPod(string name, IModuleIdentity identity, KubernetesModule
                 .Select(pullSecretName => new V1LocalObjectReference(pullSecretName))
                 .ToList();
 
-            V1PodSecurityContext securityContext = this.runAsNonRoot
-                ? new V1PodSecurityContext { RunAsNonRoot = true, RunAsUser = 1000 }
-                : null;
+            V1PodSecurityContext securityContext = module.Config.CreateOptions.SecurityContext.GetOrElse(
+                () => this.runAsNonRoot
```
massand: I am struggling to see the value of having us store a default for runAsNonRoot in every V1PodSecurityContext and ignore it when module CreateOptions don't set any default runAsNonRoot. Our agent will not error out or warn anybody about this happening, making troubleshooting a pain for operators / module developers. Can we stay consistent with policy always and override the module CreateOption runAsNonRoot setting, when the helm chart runAsNonRoot policy is true?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -132,9 +132,10 @@ V1PodTemplateSpec GetPod(string name, IModuleIdentity identity, KubernetesModule
                 .Select(pullSecretName => new V1LocalObjectReference(pullSecretName))
                 .ToList();
 
-            V1PodSecurityContext securityContext = this.runAsNonRoot
-                ? new V1PodSecurityContext { RunAsNonRoot = true, RunAsUser = 1000 }
-                : null;
+            V1PodSecurityContext securityContext = module.Config.CreateOptions.SecurityContext.GetOrElse(
```
massand: PodSecurityContext can be overriden by a SecurityContext set on the Pod. We should go for the higher precedence object.

dmolokanov: For the sake of consistency (edge module == k8s pod), I put SecurityContext to a Pod, not a module container. So if a user wants to override this for some modules it can be done. I didn't quite understand your suggestion here.

massand: Also, we are setting SecurityContexts on all containers in our V1PodTemplateSpec with privileged = true. This implies we are setting runAsNonRoot=true (from PodSecurityContext) and privileged=true (from SecurityContext). This means we run as a privileged non-root user?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
dmolokanov: The main idea we discussed: we don't want to dictate the user how to create their modules. We should allow a user to run privileged containers if it what he wants. If he opt-in `runAsNonRoot: true` in helm chart we should put this policy. If somehting is specified in CreateOptions section it will override whatever defaults we have.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
dmolokanov: Yes, some containers can have privileged option coming from DockerConfig options. SecurityContext from container section overrides PodSecurityContext from pod spec per this [doc](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#podsecuritycontext-v1-core). 
Yes, in this case container will run is some [priviledged](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#securitycontext-v1-core) user

massand: Can we set just SecurityContext in the PodTemplate? It has 'runAsNonRoot' as well. This way the policy merge will be more intuitive for customers. The default SecurityContext can be runAsNonRoot=true and runAsUser=1000. Since everything SecurityContext overrides PodSecurityContext, we don't have to worry about backwards compat if we decide to support PodSecurityContext in the future.

dmolokanov: @darobs @veyalla Any thoughts?

veyalla: Agree with @massand that setting everything at pod level is more intuitive, **however**, I think the flexibility that translating docker privileged flag at container level provides flexibility that some users might want. 

I don't see a case where the proxy would ever need to run privileged and not having a way to prevent that from happening while requiring the workload co-located in the pod to run privileged might be off putting.

massand: @veyalla to be clear, I suggest setting everything at the container level AND pod level. This is  ensure any future (say ephemeral containers) added to those pods have good defaults too. The container spec is where all Edge module createoptions go. This should be no exception to that.

darobs: @massand has some good points. Here's what I think:
We called this the "Pod Security Context" - which I think is correct term for this, injected sidecars without a security policy will need to inherit the PodSecurityContext.  

Running as privileged but non-root is possible. but weird.  I don't entirely understand what that means, 

How certain are we that "Some fields are present in both SecurityContext and PodSecurityContext. When both are set, the values in SecurityContext take precedence." is an inviolate rule in k8s? If there is any doubt, or any uncertainty how this is interpreted or used, then we should be explicit and apply the PodSecurityContext to all the pod container's SecurityContext.  This leaves the least room for doubt how to interpret its use on the module.

darobs: _"Running as privileged but non-root is possible. but weird. I don't entirely understand what that means,"_ 
To make this clear, I think this is something that the user sets a "run as non-root" policy, and then sets "privileged" for a module, then we should set the module to do that.

CindyXing: I think exposing both docker and k8s in the spec is a problem. Same for volume. 
Personally I prefer allowing customers picking one. If both are specified, use the k8s one. 
Resolving potential conflict or defining the behavior can be a pain.

darobs: As I've been thinking about it, the feature, as it is, is clear: add a pod security context.  This shouldn't be confusing to a novice or expert user, and and expert user will understand the subtleties of this.  I'm also more convinced that a user setting the Pod Security Context will not be messing with the Docker createOptions security context, and vice-versa.

PR #2014, opened 2019-11-19
Add metrics collector to build pipeline

and-rewsmith:
This PR adds the metrics collector module (created by Varun and Venkat) into the build pipeline. I made some modifications to their implementation to take a dependency on the diagnostics module. I also converted the string field of tags in the diagnostic module to a dictionary as this was useful for both Lee and I. This change accounts for the JsonConvert. SerializeObject diffs. 

There is a PR on top of this to add the metrics collector module to the test pipeline: 
https://github.com/and-rewsmith/iotedge/pull/4

In retrospect, I should have opened a new PR after I added a dependency on the diagnostics module instead of closing and reopening the same one.


e2e_deployment_files/long_haul_deployment.template.json
```diff
@@ -203,10 +203,10 @@
                 "value": "<Analyzer.LogAnalyticEnabled>"
               },
               "LogAnalyticWorkspaceId": {
-                "value": "<Analyzer.LogAnalyticWorkspaceId>"
```
and-rewsmith: Unsure why this appears as a diff. See [here](it is the same).

edge-modules/MetricsCollector/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+            Console.WriteLine($"Using message identifier {identifier}");
+
+            var messageFormatter = new MessageFormatter(configuration.MetricsFormat, identifier);
+            var scraper = new Scraper(configuration.Endpoints.Values.ToList());
+
+            IMetricsSync metricsSync;
+            if (configuration.SyncTarget == SyncTarget.AzureLogAnalytics)
+            {
+                string workspaceId = Environment.GetEnvironmentVariable("AzMonWorkspaceId") ??
+                    Environment.GetEnvironmentVariable("azMonWorkspaceId") ?? // Workaround for IoT Edge k8s bug
```
and-rewsmith: @varunpuranik I'm refactoring this to follow the edge-modules convention for environment variables. Do you know if we still need this workaround?

edge-modules/MetricsCollector/Configuration.cs
```diff
@@ -0,0 +1,51 @@
...
+        Prometheus,
+        Json
+    }
+
+    public enum SyncTarget
+    {
+        IoTHub,
+        AzureLogAnalytics
+    }
+}
```
and-rewsmith: We should fix these newlines

edge-modules/MetricsCollector/LogAnalyticsMetricsSync.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            try
+            {
+                IEnumerable<string> scrapedMetrics = await this.scraper.Scrape();
+                foreach (string scrape in scrapedMetrics)
+                {
+                    string workspaceId = Settings.Current.AzMonWorkspaceId;
+                    string workspaceKey = Settings.Current.AzMonWorkspaceKey;
+                    string customLogName = Settings.Current.AzMonCustomLogName;
+                    AzureLogAnalytics.Instance.PostAsync(workspaceId, workspaceKey, this.messageFormatter.BuildJSON(scrape), customLogName);
```
and-rewsmith: Using shared log analytics client in Edge Util

Microsoft.Azure.Devices.Edge.sln
```diff
@@ -191,6 +191,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Devices.Edg
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Azure.Devices.Edge.Agent.Kubernetes.IntegrationTest", "edge-agent\test\Microsoft.Azure.Devices.Edge.Agent.Kubernetes.IntegrationTest\Microsoft.Azure.Devices.Edge.Agent.Kubernetes.IntegrationTest.csproj", "{79540E35-45EA-4568-9AB7-C1CC0BFF73EE}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "edge-modules", "edge-modules", "{F17982D5-947A-4E96-8F2A-0E45F1ED7A30}"
```
philipktlin: why "edge-modules" added?

and-rewsmith: When I use the dotnet command to add projects to the solution, it automatically adds edge-modules. I could be doing something wrong? If you think this shouldn't be happening I can try to add again.

edge-modules/MetricsCollector/Configuration.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace MetricsCollector
```
and-rewsmith: I'm going to explore merging this with Settings.cs

edge-modules/MetricsCollector/IMetricsSync.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+namespace MetricsCollector
+{
+    using System.Threading.Tasks;
+
+    interface IMetricsSync
```
philipktlin: should it be IMetricSubscriber?

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        readonly MessageFormatter messageFormatter;
+        readonly Scraper scraper;
+        readonly ModuleClient moduleClient;
+
+        public IoTHubMetricsSync(MessageFormatter messageFormatter, Scraper scraper, ModuleClient moduleClient)
+        {
+            this.messageFormatter = messageFormatter ?? throw new ArgumentNullException(nameof(messageFormatter));
+            this.scraper = scraper ?? throw new ArgumentNullException(nameof(scraper));
+            this.moduleClient = moduleClient;
```
philipktlin: check moduleClient for not null

edge-modules/MetricsCollector/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+                },
+                tcs);
+            return tcs.Task;
+        }
+
+        /// <summary>
+        ///     Initializes the ModuleClient and sets up the callback to receive
+        ///     messages containing temperature information
+        /// </summary>
+        private static async Task Init()
```
philipktlin:
```suggestion
        private static async Task InitAsync()
```

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
and-rewsmith: I'll also switch to Preconditions for consistency.

edge-modules/MetricsCollector/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+            Console.CancelKeyPress += (sender, cpe) => cts.Cancel();
+            await WhenCancelled(cts.Token, completed, handler);
+            Logger.LogInformation("MetricsCollector Main() finished.");
+            return 0;
+        }
+
+        /// <summary>
+        ///     Handles cleanup operations when app is cancelled or unloads
+        /// </summary>
+        public static Task WhenCancelled(CancellationToken cancellationToken, ManualResetEventSlim completed, Option<object> handler)
```
philipktlin:
```suggestion
        public static Task WhenCancelledAsync(CancellationToken cancellationToken, ManualResetEventSlim completed, Option<object> handler)
```

edge-modules/MetricsCollector/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+            Logger.LogInformation($"Starting metrics collector with the following settings:\r\n{Settings.Current}");
+
+            await Init();
+
+            (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromSeconds(5), Logger);
+
+            // Wait until the app unloads or is cancelled
+            AssemblyLoadContext.Default.Unloading += ctx => cts.Cancel();
+            Console.CancelKeyPress += (sender, cpe) => cts.Cancel();
+            await WhenCancelled(cts.Token, completed, handler);
```
philipktlin:
```suggestion
            await WhenCancelled(cts.Token, completed, handler).ConfigureAwait(false);
```

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public async Task ScrapeAndSync()
+        {
+            try
+            {
+                IEnumerable<string> scrapedMetrics = await this.scraper.Scrape();
+                IList<Message> messages =
+                        scrapedMetrics.SelectMany(prometheusMetrics => this.messageFormatter.Build(prometheusMetrics)).ToList();
+                await this.moduleClient.SendEventBatchAsync(messages);
+                Console.WriteLine($"Sent metrics as {messages.Count} messages");
```
and-rewsmith: I'm going to switch all console.Writes to logger.Logs

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
```diff
@@ -0,0 +1,43 @@
...
+        readonly ModuleClient moduleClient;
+
+        public IoTHubMetricsSync(MessageFormatter messageFormatter, Scraper scraper, ModuleClient moduleClient)
+        {
+            this.messageFormatter = Preconditions.CheckNotNull(messageFormatter);
+            this.scraper = Preconditions.CheckNotNull(scraper);
+            this.moduleClient = Preconditions.CheckNotNull(moduleClient);
+        }
+
+        public async Task ScrapeAndSync()
```
and-rewsmith: If we re-name this with "async" postfix then the name gets messy. Should I rename anyway?

edge-modules/MetricsCollector/IMetricsSync.cs
and-rewsmith: It's not really subscribing to metrics updates, but scraping and syncing them itself.

edge-modules/MetricsCollector/Program.cs
and-rewsmith: We shouldn't need this since metrics collector is running on dotnet core, right?

and-rewsmith: Adding as it should still be compatible with .NET standard

builds/misc/images.yaml
```diff
@@ -354,6 +368,13 @@ jobs:
           imageName: azureiotedge-module-restarter
           project: ModuleRestarter
 
+     # Metrics Collector
+      - template: templates/image-windows.yaml
+        parameters:
+          name: Metrics Collector
+          imageName: azureiotedge-metrics-collector
+          project: MetricsCollector
```
yophilav: Are we not publishing the docker binary for Metric Collector?

edge-modules/MetricsCollector/Configuration.cs
yophilav: 1.Merging with `Settings.cs`  does not quite make sense to me. 
This is more assimilate with `ModuleMessagesReport` and `ResponseOrientedReport` since they are simply just another report type (soon-to-be LogType).

2.  Can we consolidate `\iotedge\edge-modules\MessagesAnalyzer\ModuleMessagesReport.cs` and `\iotedge\edge-modules\MessagesAnalyzer\ModuleReport.cs`?

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
yophilav: How about we rename it to `ScrapeAndUploadPrometheusMetricsAsync()`?

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                IEnumerable<string> scrapedMetrics = await this.scraper.Scrape();
+                IList<Message> messages =
+                        scrapedMetrics.SelectMany(prometheusMetrics => this.messageFormatter.Build(prometheusMetrics)).ToList();
+                await this.moduleClient.SendEventBatchAsync(messages);
+                Logger.LogInformation($"Sent metrics as {messages.Count} messages");
+            }
+            catch (Exception e)
+            {
+                Logger.LogError($"Error scraping and syncing metrics - {e}");
```
yophilav: It looks like `SendEventBatchAsync()` does not retry the upload if the message failed. 
Should we do a retry here for a not-so-critical http (503, 429, 301) here? 
See Mozilla standard: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
and-rewsmith: I like that name considering its an async method here, however this method is implemented following the interface IMetricsSync, which has a definition of: 
```
Task ScrapeAndSync
```
It feels wrong to rename the method to async considering that we would also have to update the interface, which allows for non-async implementations.

I think its fine as is. Do you agree?

and-rewsmith: Followed your suggestion in Program.cs

edge-modules/MetricsCollector/Program.cs
and-rewsmith: This method runs synchronously though, right?

edge-modules/MetricsCollector/Configuration.cs
and-rewsmith: 1 - Aren't these just settings? Settings.cs sets configuration based on env vars. This Configuration.cs class sets configuration based on the module twin's desired properties. I agree that merging is not the best strategy, so I was thinking of renaming Configuration.cs to something more descriptive that indicates it sources settings from the module twin.

2 - I couldn't find ModuleReport. Are you sure this is the correct name, or am I looking in the wrong place?

builds/misc/images.yaml
and-rewsmith: I feel we should discuss publishing via releases after we have verified the metrics collector is stable (through building the dashboard).

edge-modules/MetricsCollector/Configuration.cs
yophilav: 1. You're right now that I read the rest of the code. I don't think you should merge it though b/c you still want to be able to deploy MetricCollector without Analyzer. 
2. Interesting. Maybe my git is merging some branch together that I didn't notice.

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
yophilav: I was thinking that you can have the implementation of the code as `ScrapeAndSync ` but then when you have the alias `ScrapeAndUploadPrometheusMetricsAsync() => ScrapeAndSync ()` and use `ScrapeAndUploadPrometheusMetricsAsync()` everywhere else in the code. What do you think?

edge-modules/MetricsCollector/Configuration.cs
and-rewsmith: 1 - The metrics collector also has a Settings.cs, which is what I was referring to

edge-modules/MetricsCollector/Configuration.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Converters;
+
+    public class Configuration
```
philipktlin: be specific. for example.
```suggestion
    public class MetricsConfiguration
```

edge-modules/MetricsCollector/IMetricsSync.cs
```diff
@@ -0,0 +1,10 @@
...
+    {
+        Task ScrapeAndSync();
```
philipktlin:
```suggestion
        Task ScrapeAndSyncAsync();
```

edge-modules/MetricsCollector/Program.cs
yophilav: https://stackoverflow.com/questions/56349117/configureawait-in-net-core-console-worker-not-asp-net-core

I think you're right Andrew.

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
philipktlin: First Async suffix should apply to any method returns Task or Task<T>.  If you don't like method named as "ScrapeAndSyncAsync", you can rename to anything with Async suffix, e.g. ScrapeAndReportAsync, etc.

edge-modules/MetricsCollector/Scraper.cs
```diff
@@ -0,0 +1,86 @@
...
+    using System.Linq;
+    using System.Net;
+    using System.Net.Http;
+    using System.Text.RegularExpressions;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public class Scraper
```
philipktlin: I found that you are implementing something very similar to what Lee is doing.  I think you need to sync with him and see how to reuse in both places.

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -0,0 +1,58 @@
...
+        public string AzMonWorkspaceId { get; }
+
+        [JsonIgnore]
+        public string AzMonWorkspaceKey { get; }
+
+        public string AzMonLogType { get; }
+
+        public override string ToString()
+        {
+            return JsonConvert.SerializeObject(this, Formatting.Indented);
```
philipktlin: nit: personally I don't like this; because if someone add a new property (maybe a secret), it will automatically show up in ToString().  This may leak it unintentionally.

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
and-rewsmith: I thought that you used async suffix if the method was async. If the Task is synchronous isn't this misleading?

and-rewsmith: `I'm going to make the interface an async method and rename`per your suggestions, but I'm still curious about this async suffix rule.

edge-modules/MetricsCollector/IoTHubMetricsSync.cs
and-rewsmith: Can edgeHub return a 503? I'm pretty sure it wouldn't return a 429 or 301.

edge-modules/MetricsCollector/Configuration.cs
and-rewsmith: Bear and I discussed above, but I think the name should indicate that these metrics configurations are sourced from the twin. I have decided to go with MetricsConfigFromTwin

edge-modules/MetricsCollector/IMetricsSync.cs
and-rewsmith: I went with scrapeAndSyncMetricsAsync

edge-modules/MetricsCollector/MessageFormatter.cs
```diff
@@ -0,0 +1,221 @@
...
+            Message message = new Message(messageBytes);
+            message.Properties[IdentifierPropertyName] = this.identifier;
+            message.MessageSchema = this.format.ToString();
+            message.ContentEncoding = "UTF-8";
+            message.ContentType = "application/json";
+
+            return new List<Message> { message };
+        }
+
+        IList<MetricsData> GetMetricsDataList(string prometheusMessage)
```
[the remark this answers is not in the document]
and-rewsmith: BuildMessagesInJsonFormat2 has no callers so we should remove.

edge-modules/MetricsCollector/Configuration.cs
philipktlin: If I define default metrics configuration in config file and use it in case these config cannot get from Twin; I should reuse this class, right?  If so, this name is not fit.

edge-modules/MetricsCollector/LogAnalyticsMetricsSync.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            try
+            {
+                IEnumerable<string> scrapedMetrics = await this.scraper.Scrape();
+                foreach (string scrape in scrapedMetrics)
+                {
+                    string workspaceId = Settings.Current.AzMonWorkspaceId;
+                    string workspaceKey = Settings.Current.AzMonWorkspaceKey;
+                    string logType = Settings.Current.AzMonLogType;
+                    AzureLogAnalytics.Instance.PostAsync(workspaceId, workspaceKey, this.messageFormatter.BuildJSON(scrape), logType);
```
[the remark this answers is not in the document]
and-rewsmith: 1 - This will be helpful, so I'll find one.

2 - Assuming you are talking about JSON version for Log Analytics Post() body. Is this something formally defined by log analytics? If so, it would be helpful.

3 - I thought it was an anti-pattern to await things that don't need to be awaited?

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: Yeah, this is a good point. I will change.

edge-modules/MetricsCollector/LogAnalyticsMetricsSync.cs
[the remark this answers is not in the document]
yophilav: 1. See my other comment. I think it is that.

2. I'm thinking if there is a version associate with `this.messageFormatter.BuildJSON(scrape)`. It looks like the answer is yes, he has a `schemaVersion` to track which Json format.

3. After talking with Damon, if you don't `await` it would be just fired and forget about it. But you want to send it in order you want to do `await`.  He also has a concern about how the `PostAsync()` is used in `foreach()` which means you actually send out a bunch of POST request.  Would it be better to formulate all the json then send a POST request once?  (Also my bad on the `public async void PostAsync(string workspaceId, string sharedKey, string content, string logType)` should really be `public async Task PostAsync(string workspaceId, string sharedKey, string content, string logType)` Do you want me to fix this?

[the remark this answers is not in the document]
and-rewsmith: 1 - Sounds good

2 - I'll investigate

3 - Awaiting the call would provide consistent ordering for log analytics, so I'll change. I'll also make the change from void to task. For the post requests, you'll only be making a few post requests (one for each endpoint you are scraping, which is by default 1), so I think it is fine to send multiple. Sending multiple requests also allows you to send in the case where BuildJson fails for one or more endpoints. I'm going to modify the try / catch logic to support this.

edge-modules/MetricsCollector/Configuration.cs
and-rewsmith: I'm going to merge this configuration class with Settings.cs. The way this is implemented is counter-intuitive to me. 

I think the only reason it is implemented with two settings/configuration classes is to allow for the endpoints as a dictionary via the twin's desired properties. I had assumed that the GetConfigurationAsync() read from the twin on a timer to refresh the configuration, but that is not the case. The configuration is retrieved once from the twin and never again. It seems counter-intuitive to update the twin and have nothing happen, so I am going to have everything read from the environment.

and-rewsmith:
Closing. I'll re-open when I add the dependency on the diagnostics module.

and-rewsmith:
Reopening with dependency on the diagnostic module.

and-rewsmith:
Re-opened for testing. Once Lee's PR gets merged the diff will be manageable to review.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/prometheus.net/MetricsProvider.cs
```diff
@@ -13,16 +13,20 @@ public class MetricsProvider : IMetricsProvider
...
-        public MetricsProvider(string namePrefix, string iotHubName, string deviceId)
+        public MetricsProvider(string namePrefix, string iotHubName, string deviceId, string storagePath)
         {
             this.namePrefix = Preconditions.CheckNonWhiteSpace(namePrefix, nameof(namePrefix));
+
+
             Preconditions.CheckNonWhiteSpace(iotHubName, nameof(iotHubName));
             Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
+            Preconditions.CheckNonWhiteSpace(storagePath, nameof(storagePath));
+            Lazy<string> instanceNumber = new Lazy<string>(() => GetInstanceNumber(storagePath));
```
and-rewsmith: I preserved the pattern of being lazy, but I don't think it needs to be. Can I simply return a string? I'm assuming I am missing the point here and Lazy was intentional.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public LogAnalyticsUpload(string workspaceId, string workspaceKey, string logType, Guid guid)
+        {
+            this.workspaceId = Preconditions.CheckNonWhiteSpace(workspaceId, nameof(workspaceId));
+            this.workspaceKey = Preconditions.CheckNonWhiteSpace(workspaceKey, nameof(workspaceKey));
+            this.logType = Preconditions.CheckNonWhiteSpace(logType, nameof(logType));
+            this.guid = Preconditions.CheckNotNull(guid);
+        }
+
+        public IEnumerable<Metric> GetGuidTaggedMetrics(IEnumerable<Metric> originalMetrics)
```
lfitchett: Does this need to be public?

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public void Start(TimeSpan scrapeAndUploadInterval)
+        {
+            this.periodicScrapeAndUpload = new PeriodicTask(this.ScrapeAndUploadPrometheusMetricsAsync, scrapeAndUploadInterval, scrapeAndUploadInterval, Logger, "Scrape and Upload Metrics");
+        }
+
+        public void Dispose()
+        {
+           this.periodicScrapeAndUpload.Dispose();
```
lfitchett:
```suggestion
           this.periodicScrapeAndUpload?.Dispose();
```

this will prevent it from erroring if disposed before being started

edge-modules/MetricsCollector/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+            }
+            else
+            {
+                MqttTransportSettings mqttSetting = new MqttTransportSettings(TransportType.Mqtt_Tcp_Only);
+                ITransportSettings[] transportSettings = { mqttSetting };
+                ModuleClient moduleClient = await ModuleClient.CreateFromEnvironmentAsync(transportSettings);
+                publisher = new IoTHubMetricsUpload(moduleClient);
+            }
+
+            MetricsScrapeAndUpload metricsScrapeAndUpload = new MetricsScrapeAndUpload(scraper, publisher);
```
lfitchett: since this is an IDisposable, it should be disposed before shutdown. I recommend a ```using``` statement

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -0,0 +1,87 @@
...
+            fields.Add(nameof(this.AzMonWorkspaceId), this.AzMonWorkspaceId);
+            fields.Add(nameof(this.AzMonLogType), this.AzMonLogType);
+            fields.Add(nameof(this.Endpoints), JsonConvert.SerializeObject(this.Endpoints, Formatting.Indented));
+            fields.Add(nameof(this.ScrapeFrequencySecs), this.ScrapeFrequencySecs.ToString());
+            fields.Add(nameof(this.UploadTarget), Enum.GetName(typeof(UploadTarget), this.UploadTarget));
+            return JsonConvert.SerializeObject(fields, Formatting.Indented);
+        }
+    }
+
+    public enum UploadTarget
```
lfitchett: you shouldn't need this anymore, since this only uploads directly to azure log analytics

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -0,0 +1,87 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Configuration;
+    using Newtonsoft.Json;
+
+    public class Settings
+    {
+        static readonly Lazy<Settings> DefaultSettings = new Lazy<Settings>(
```
lfitchett: why is this lazy and static?

edge-modules/MetricsCollector/config/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "AzMonWorkspaceId": "",
+  "AzMonWorkspaceKey": "",
+  "AzMonLogType": "edgeHubMetrics",
+  "MetricsEndpoints": "[\"http://edgeHub:9600/metrics\"]",
```
lfitchett: you need to add edgeAgent here as well

edge-modules/MetricsCollector/docker/windows/amd64/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG base_tag=2.1.13-nanoserver-1809
+FROM mcr.microsoft.com/dotnet/core/runtime:${base_tag}
+
+ARG EXE_DIR=.
+
```
lfitchett: I'm pretty sure you must be container admin to scrape metrics on windows. @varunpuranik can you confirm?

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/prometheus.net/MetricsProvider.cs
```diff
@@ -13,17 +13,19 @@ public class MetricsProvider : IMetricsProvider
...
 
-        public MetricsProvider(string namePrefix, string iotHubName, string deviceId)
+        public MetricsProvider(string namePrefix, string iotHubName, string deviceId, string storagePath)
         {
             this.namePrefix = Preconditions.CheckNonWhiteSpace(namePrefix, nameof(namePrefix));
+
             Preconditions.CheckNonWhiteSpace(iotHubName, nameof(iotHubName));
             Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
-            this.defaultLabelNames = new List<string> { iotHubName, deviceId, instanceNumber.Value };
+            Preconditions.CheckNonWhiteSpace(storagePath, nameof(storagePath));
```
lfitchett: you should probably put this file in its own directory, similar to the AvailabilityMetrics class.

edge-modules/MetricsCollector/docker/windows/amd64/Dockerfile
and-rewsmith: I confirmed this yesterday and will make this change.

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: It also allows you to publish to IoTHub

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: It is static so it can be accessed without an instance (i.e. Settings.Current.<var>). It is lazy so if the user does not make calls to Settings, it does not read from the environment vars.

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: I am going to add the eventhub IoTHub Publisher as that is an unsupported scenario between veyalla/metricsCollector and this module. I will keep the compression flow if we conclude that this will only be a test module, as it doesn't really make sense if it is intended for customer use.

lfitchett: Andrew will add both

edge-modules/MetricsCollector/Settings.cs
lfitchett: It is immediately called by ```public static Settings Current```, so it will resolve anyway.

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -23,7 +23,7 @@ public class Settings
                     configuration.GetValue<string>("AzMonWorkspaceId"),
                     configuration.GetValue<string>("AzMonWorkspaceKey"),
                     configuration.GetValue<string>("AzMonLogType", "edgeHubMetrics"),
-                    configuration.GetValue<string>("MetricsEndpointsCSV", "http://edgeHub:9600/metrics"),
+                    configuration.GetValue<string>("MetricsEndpointsCSV", "http://edgeHub:9600/metrics,http://edgeAgent:9600/metrics"),
```
lfitchett: Should this default be edgehub and edgeagent or just empty? The settings json specifies the endpoints to scrape

and-rewsmith: I tried earlier to just use empty because it should take its values from the settings.json, however they did not get read. I'm planning on asking Philip about this.

and-rewsmith: The convention in other modules though is to specify in both places, which I disagree with as it is unnecessary.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EdgeRuntimeDiagnosticsUpload.cs
```diff
@@ -9,11 +9,11 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Publisher
     using Microsoft.Azure.Devices.Client;
     using Microsoft.Azure.Devices.Edge.Util;
 
-    public sealed class IoTHubMetricsUpload : IMetricsPublisher
+    public sealed class EventHubMetricUpload : IMetricsPublisher
```
philipktlin: why filename and class out of sync?  And there is another file with same namespace and class name.

and-rewsmith: I just changed this. Forgot to change the name. Will fix now

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EventHubMetricsUpload.cs
```diff
@@ -0,0 +1,39 @@
...
+        public EventHubMetricsUpload(ModuleClient moduleClient)
+        {
+            this.moduleClient = Preconditions.CheckNotNull(moduleClient, nameof(moduleClient));
+        }
+
+        public async Task PublishAsync(IEnumerable<Metric> metrics, CancellationToken cancellationToken)
+        {
+            try
+            {
+                Message metricsMessage = new Message(Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(metrics)));
```
philipktlin: what happens if metrics is null, passing to JsonConvert.SerializeObject?

Microsoft.Azure.Devices.Edge.sln
```diff
@@ -203,6 +203,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ModuleRestarter", "edge-mod
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Relayer", "edge-modules\Relayer\Relayer.csproj", "{C1FDA6CF-7EF6-4C3F-81E7-4C13EE399BF1}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MetricsCollector", "edge-modules\MetricsCollector\MetricsCollector.csproj", "{C01DF161-76F7-47F2-9025-B0D99DB4B814}"
```
philipktlin: New project for testing purpose only should move under <repo>/test/modules? is it used for testing?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public LogAnalyticsUpload(string workspaceId, string workspaceKey, string logType, Guid guid)
+        {
+            this.workspaceId = Preconditions.CheckNonWhiteSpace(workspaceId, nameof(workspaceId));
+            this.workspaceKey = Preconditions.CheckNonWhiteSpace(workspaceKey, nameof(workspaceKey));
+            this.logType = Preconditions.CheckNonWhiteSpace(logType, nameof(logType));
+            this.guid = Preconditions.CheckNotNull(guid);
+        }
+
+        IEnumerable<Metric> GetGuidTaggedMetrics(IEnumerable<Metric> originalMetrics)
```
philipktlin: I think public methods should be put above internal and private methods

Microsoft.Azure.Devices.Edge.sln
and-rewsmith: So this might replace the metrics collector currently living in veyalla/metricsCollector. I need to sync with Varun on this though. My only concern is that I have changed the EventHub upload format, so it is not backwards compatible if we decide to deprecate the old module.

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -0,0 +1,87 @@
...
+            fields.Add(nameof(this.Endpoints), JsonConvert.SerializeObject(this.Endpoints, Formatting.Indented));
+            fields.Add(nameof(this.ScrapeFrequencySecs), this.ScrapeFrequencySecs.ToString());
+            fields.Add(nameof(this.UploadTarget), Enum.GetName(typeof(UploadTarget), this.UploadTarget));
+            return JsonConvert.SerializeObject(fields, Formatting.Indented);
+        }
+    }
+
+    public enum UploadTarget
+    {
+        IoTHub,
```
and-rewsmith: This should change to EventHub.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EventHubMetricsUpload.cs
```diff
@@ -0,0 +1,39 @@
...
+    using System.Collections.Generic;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Client;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    public class EventHubMetricsUpload : IMetricsPublisher
```
and-rewsmith: Since we are now supporting event hub upload, it should also tag metrics with a guid as we do in the log analytics upload.

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: It shouldn't be lazy then. @philipktlin  do you agree?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EventHubMetricsUpload.cs
and-rewsmith: I've added a preconditions null check, but what do you think we do if the IEnumerable contains no metrics? I feel like there should be a check for this. Do you agree?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public LogAnalyticsUpload(string workspaceId, string workspaceKey, string logType, Guid guid)
+        {
+            this.workspaceId = Preconditions.CheckNonWhiteSpace(workspaceId, nameof(workspaceId));
+            this.workspaceKey = Preconditions.CheckNonWhiteSpace(workspaceKey, nameof(workspaceKey));
+            this.logType = Preconditions.CheckNonWhiteSpace(logType, nameof(logType));
+            this.guid = Preconditions.CheckNotNull(guid);
+        }
+
+        public async Task PublishAsync(IEnumerable<Metric> metrics, CancellationToken cancellationToken)
```
and-rewsmith: I've also addressed possibility of null metrics here

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: Resolving as we should remove lazy across all test modules in one PR.

Microsoft.Azure.Devices.Edge.sln
philipktlin: If it is not for test purpose, this folder should be fine.  Please confirm with Varun.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
```diff
@@ -33,7 +34,7 @@ public int GetMetricKey()
             // when upgraded to .net standard 2.1: https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=netstandard-2.1
             int hash = 17;
             hash = hash * 31 + this.Name.GetHashCode();
-            hash = hash * 31 + this.Tags.GetHashCode();
+            hash = hash * 31 + JsonConvert.SerializeObject(this.Tags).GetHashCode();
```
philipktlin: why do json convert? just concatenate tags instead.

philipktlin: Add how can you ensure the order of tags is same all the time when calling GetHashCode

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
```diff
@@ -43,7 +44,7 @@ public bool Equals(Metric other)
             return this.TimeGeneratedUtc == other.TimeGeneratedUtc &&
                 this.Name == other.Name &&
                 this.Value == other.Value &&
-                this.Tags == other.Tags;
+                JsonConvert.SerializeObject(this.Tags) == JsonConvert.SerializeObject(other.Tags);
```
philipktlin: don't use Json; just use dictionary count and except; or implement an extension method.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
and-rewsmith: I didn't think JsonConvert.SerializeObject added enough overhead to be worth implementing manually. I could be wrong about this though. 

For the tag ordering, the original implementation would yield different hash codes for different tag orders, which I agree can be improved. I should be calling GetHashCode() on the dictionary object anyways though which I will change.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
```diff
@@ -6,15 +6,16 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics
...
     using Microsoft.Azure.Devices.Edge.Util;
+    using Newtonsoft.Json;
 
     public sealed class Metric : IEquatable<Metric>
     {
         public DateTime TimeGeneratedUtc { get; }
         public string Name { get; }
         public double Value { get; }
-        public string Tags { get; }
+        public Dictionary<string, string> Tags { get; }
```
philipktlin:
```suggestion
        public IReadonlyDictionary<string, string> Tags { get; }
```

philipktlin: To be more safe, you can copy items from given dictionary in constructor.  To ensure, these values cannot be changed.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsSerializer.cs
```diff
@@ -22,13 +23,13 @@ public static IEnumerable<byte> MetricsToBytes(IEnumerable<Metric> metrics)
                 x.Select(m => new RawMetricValue(m.TimeGeneratedUtc, m.Value))));
         }
 
-        static IEnumerable<byte> MetricGroupsToBytes(string name, string tags, IEnumerable<RawMetricValue> rawValues)
+        static IEnumerable<byte> MetricGroupsToBytes(string name, Dictionary<string, string> tags, IEnumerable<RawMetricValue> rawValues)
```
philipktlin:
```suggestion
        static IEnumerable<byte> MetricGroupsToBytes(string name, IReadonlyDictionary<string, string> tags, IEnumerable<RawMetricValue> rawValues)
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsSerializer.cs
```diff
@@ -22,13 +23,13 @@ public static IEnumerable<byte> MetricsToBytes(IEnumerable<Metric> metrics)
...
         {
             byte[] nameArray = Encoding.UTF8.GetBytes(name);
             byte[] nameLength = BitConverter.GetBytes(name.Length);
 
-            byte[] tagsArray = Encoding.UTF8.GetBytes(tags);
-            byte[] tagsLength = BitConverter.GetBytes(tags.Length);
+            byte[] tagsArray = Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(tags));
```
philipktlin: I am not sure SerializeObject will always serialize dictionary items in the same order; please check.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsSerializer.cs
and-rewsmith: This is intentionally a readonly dictionary as Lee and I decided to allow for custom tags to be added in the case where the scraper is used elsewhere. For example in the metrics collector I edit the tags to add a guid.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsSerializer.cs
```diff
@@ -22,13 +23,13 @@ public static IEnumerable<byte> MetricsToBytes(IEnumerable<Metric> metrics)
...
+            byte[] tagsLength = BitConverter.GetBytes(JsonConvert.SerializeObject(tags).Length);
```
philipktlin: call JsonConvert.SerializeObject(tags) once and reuse it.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/PrometheusMetricsParser.cs
```diff
@@ -52,7 +51,7 @@ public static IEnumerable<Metric> ParseMetrics(DateTime timeGeneratedUtc, string
                             timeGeneratedUtc,
                             metricName,
                             metricValue,
-                            JsonConvert.SerializeObject(tags));
```
philipktlin: why change this?

and-rewsmith: Tags is now a dictionary. I could make this clearer by explicitly stating the type on line 47

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EventHubMetricsUpload.cs
```diff
@@ -0,0 +1,43 @@
...
+                byte[] metricsData = Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(metrics));
+                Message metricsMessage = new Message(metricsData);
+
+                await this.moduleClient.SendEventAsync(metricsMessage);
+
+                Log.LogInformation($"Successfully sent metrics to Event Hub via IoT Hub");
+            }
+            catch (Exception e)
+            {
+                Log.LogError($"Error uploading metrics to Event Hub via IoTHub: {e}");
```
philipktlin:
```suggestion
                Log.LogError(e, $"Error uploading metrics to Event Hub via IoTHub");
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
```diff
@@ -0,0 +1,43 @@
...
+        readonly string workspaceKey;
+        readonly string logType;
+        readonly Guid guid;
+
+        public LogAnalyticsUpload(string workspaceId, string workspaceKey, string logType, Guid guid)
+        {
+            this.workspaceId = Preconditions.CheckNonWhiteSpace(workspaceId, nameof(workspaceId));
+            this.workspaceKey = Preconditions.CheckNonWhiteSpace(workspaceKey, nameof(workspaceKey));
+            this.logType = Preconditions.CheckNonWhiteSpace(logType, nameof(logType));
+            this.guid = Preconditions.CheckNotNull(guid);
```
philipktlin: Guid cannot be null; only need to check it is not equal to Guid.Empty.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EventHubMetricsUpload.cs
and-rewsmith: I will change here and elsewhere

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            try
+            {
+                Preconditions.CheckNotNull(metrics, nameof(metrics));
+                await AzureLogAnalytics.Instance.PostAsync(this.workspaceId, this.workspaceKey, JsonConvert.SerializeObject(metrics), this.logType);
+                Log.LogInformation($"Successfully sent metrics to LogAnalytics");
+            }
+            catch (Exception e)
+            {
+                Log.LogError($"Error uploading metrics to LogAnalytics: {e}");
```
philipktlin:
```suggestion
                Log.LogError(e, "Error uploading metrics to LogAnalytics");
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
and-rewsmith: I think I'm missing something. Can't any input be null?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/MetricsModule.cs
```diff
@@ -18,18 +18,20 @@ public sealed class MetricsModule : Module
...
         string deviceId;
+        string edgeAgentStorageFolder;
 
-        public MetricsModule(MetricsConfig metricsConfig, string iothubHostname, string deviceId)
+        public MetricsModule(MetricsConfig metricsConfig, string iothubHostname, string deviceId, string edgeAgentStorageFolder)
         {
             this.metricsConfig = Preconditions.CheckNotNull(metricsConfig, nameof(metricsConfig));
             this.iothubHostname = Preconditions.CheckNotNull(iothubHostname, nameof(iothubHostname));
             this.deviceId = Preconditions.CheckNotNull(deviceId, nameof(deviceId));
+            this.edgeAgentStorageFolder = Preconditions.CheckNotNull(edgeAgentStorageFolder, nameof(edgeAgentStorageFolder));
```
philipktlin: what happens if this path doesn't exist?

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -0,0 +1,61 @@
...
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    class MetricsScrapeAndUpload : IDisposable
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger("MetricsCollector");
+        readonly MetricsScraper scraper;
+        readonly IMetricsPublisher publisher;
+        PeriodicTask periodicScrapeAndUpload;
+
```
philipktlin: nit: remove blank line
```suggestion
```

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -0,0 +1,61 @@
...
+        Guid guid;
+
+        public MetricsScrapeAndUpload(MetricsScraper scraper, IMetricsPublisher publisher, Guid guid)
```
philipktlin: what is this guid used for? please give a more meaningful name.

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            this.scraper = Preconditions.CheckNotNull(scraper);
+            this.publisher = Preconditions.CheckNotNull(publisher);
+            this.guid = Preconditions.CheckNotNull(guid);
```
philipktlin: Guid is struct, just check not equal to Guid.Empty.

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
philipktlin: is there any interface for MetricsScraper?

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            this.periodicScrapeAndUpload = new PeriodicTask(this.ScrapeAndUploadPrometheusMetricsAsync, scrapeAndUploadInterval, scrapeAndUploadInterval, Logger, "Scrape and Upload Metrics");
+        }
+
+        public void Dispose()
+        {
+           this.periodicScrapeAndUpload?.Dispose();
+        }
+
+        async Task ScrapeAndUploadPrometheusMetricsAsync(CancellationToken cancellationToken)
```
philipktlin:
```suggestion
        async Task ScrapeAndUploadMetricsAsync(CancellationToken cancellationToken)
```

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -0,0 +1,61 @@
...
+                    metric.Tags.Add("guid", this.guid.ToString());
+                }
+
+                await this.publisher.PublishAsync(metrics, cancellationToken);
+
+                Logger.LogInformation("Successfully scraped and uploaded metrics");
+            }
+            catch (Exception e)
+            {
+                Logger.LogError($"Error scraping and uploading metrics: {e}");
```
philipktlin:
```suggestion
                Logger.LogError(e, "Error scraping and uploading metrics");
```

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public List<string> Endpoints { get; }
+
+        public int ScrapeFrequencySecs { get; }
+
+        public UploadTarget UploadTarget { get; }
+
+        public override string ToString()
+        {
+            Dictionary<string, string> fields = new Dictionary<string, string>();
```
philipktlin: use same implementation in load-gen Settings class.

philipktlin: you should make it as util function; make it in another PR.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/prometheus.net/MetricsProvider.cs
```diff
@@ -8,22 +8,40 @@ namespace Microsoft.Azure.Devices.Edge.Util.Metrics.Prometheus.Net
...
+            {
+                Directory.CreateDirectory(storageDirectory);
+                this.instanceFileAbsolutePath = Path.Combine(storageDirectory, InstanceFileName);
+            }
+            catch (Exception ex)
+            {
+                this.logger.LogError(ex, "Could not create metrics directory");
+            }
+
+            string instanceNumber = this.GetInstanceNumber();
```
philipktlin: What happens if GetInstanceNumber returns zero?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsSerializer.cs
and-rewsmith: Since equals will no longer check for order, does this still matter?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/MetricsModule.cs
and-rewsmith: I believe this would cause a problem downstream when attempting to create the inner storage directory (not the same one) in MetricsProvider.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
and-rewsmith: I am using the following:
```
this.Tags.Count == other.Tags.Count && !this.Tags.Except(other.Tags).Any()
```

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/prometheus.net/MetricsProvider.cs
and-rewsmith: My understanding is that it will continue the normal flow thereby signalling a failed read of the instance file.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsSerializer.cs
and-rewsmith: Philip and I discussed and we will make the Metric's dictionary readonly. In the metrics collector I will have a custom metric class with custom tags field.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
```diff
@@ -4,17 +4,17 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics
...
+    public class Metric : IEquatable<Metric>
     {
         public DateTime TimeGeneratedUtc { get; }
         public string Name { get; }
         public double Value { get; }
-        public string Tags { get; }
+        public IReadOnlyDictionary<string, string> Tags { get; }
 
-        public Metric(DateTime timeGeneratedUtc, string name, double value, string tags)
+        public Metric(DateTime timeGeneratedUtc, string name, double value, IReadOnlyDictionary<string, string> tags)
```
philipktlin: Since tags are used to calculate metric key and IReadOnlyDictionary doesn't protect changes to underlying dictionary.  If underlying dictionary is changed, this.Tags.GetHashCode() will be changed as well.  Does this cause any problem as metric key will be different?

philipktlin: I think this is a problem; even with same key-value pair in a dictionary, GetHashCode can be different.  You can't use GetHashCode from a dictionary to calculate metric key.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public sealed class LogAnalyticsUpload : IMetricsPublisher
+    {
+        static readonly ILogger Log = Logger.Factory.CreateLogger<LogAnalyticsUpload>();
+        readonly string workspaceId;
+        readonly string workspaceKey;
+        readonly string logType;
+        readonly Guid guid;
+
+        public LogAnalyticsUpload(string workspaceId, string workspaceKey, string logType, Guid guid)
```
philipktlin: please give a more meaningful name for guid argument

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/MetricsModule.cs
```diff
@@ -18,18 +19,25 @@ public sealed class MetricsModule : Module
...
         string deviceId;
+        string edgeAgentStorageFolder;
 
-        public MetricsModule(MetricsConfig metricsConfig, string iothubHostname, string deviceId)
+        public MetricsModule(MetricsConfig metricsConfig, string iothubHostname, string deviceId, string edgeAgentStorageFolder)
         {
             this.metricsConfig = Preconditions.CheckNotNull(metricsConfig, nameof(metricsConfig));
             this.iothubHostname = Preconditions.CheckNotNull(iothubHostname, nameof(iothubHostname));
             this.deviceId = Preconditions.CheckNotNull(deviceId, nameof(deviceId));
+            if (!Directory.Exists(Preconditions.CheckNotNull(edgeAgentStorageFolder, nameof(edgeAgentStorageFolder))))
```
philipktlin: nit: split this into 2 lines

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
and-rewsmith: I need to remove this guid

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/MetricsModule.cs
philipktlin: this is existing code issue; iothuhHostname and deviceId should call Preconditions.CheckNullOrWhiteSpace.

edge-modules/MetricsCollector/MetricWithCustomTags.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics
+{
+    using System;
+    using System.Collections.Generic;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public sealed class MetricWithCustomTags : Metric, IEquatable<Metric>
```
philipktlin: Metric class already implement IEquatable<Metric>, why do you need to explicitly implement it here?
```suggestion
    public sealed class MetricWithCustomTags : Metric
```

edge-modules/MetricsCollector/MetricWithCustomTags.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public IReadOnlyDictionary<string, string> CustomTags { get; }
+        public MetricWithCustomTags(Metric metric, IReadOnlyDictionary<string, string> customTags)
```
philipktlin: why need Metric? you can directly pass TimeGeneratedUtc, Name, Value, Tags with custom tags.

and-rewsmith: I thought this was more desirable since the use case is to extend a pre-existing metric.

edge-modules/MetricsCollector/MetricWithCustomTags.cs
philipktlin: if we are 1 more property in this class, we may need to rename this class.  You need to rename this class, e.g. "ExtendedMetric".

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    class MetricsScrapeAndUpload : IDisposable
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger("MetricsCollector");
+        readonly IMetricsScraper scraper;
+        readonly IMetricsPublisher publisher;
+        PeriodicTask periodicScrapeAndUpload;
+        Guid testId; // used for differentiating between test runs on the same device
+
+        public MetricsScrapeAndUpload(IMetricsScraper scraper, IMetricsPublisher publisher, Guid testId)
```
philipktlin: why call testId? this unique information can only be used for testing?

edge-modules/MetricsCollector/Program.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+            else
+            {
+                MqttTransportSettings mqttSetting = new MqttTransportSettings(TransportType.Mqtt_Tcp_Only);
+                ITransportSettings[] transportSettings = { mqttSetting };
+                ModuleClient moduleClient = await ModuleClient.CreateFromEnvironmentAsync(transportSettings);
+                publisher = new EventHubMetricsUpload(moduleClient);
+            }
+
+            (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromSeconds(5), Logger);
```
philipktlin: why this line is called between metrics publisher and metrics scrap & upload?  should it put before metrics publisher?

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
and-rewsmith: Good point, as others might use this module we should call it something like metricsCollectorRuntimeId

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -0,0 +1,91 @@
...
+            string endpoints,
+            int scrapeFrequencySecs,
+            UploadTarget uploadTarget)
+        {
+            this.AzMonWorkspaceId = Preconditions.CheckNonWhiteSpace(azMonWorkspaceId, nameof(azMonWorkspaceId));
+            this.AzMonWorkspaceKey = Preconditions.CheckNonWhiteSpace(azMonWorkspaceKey, nameof(azMonWorkspaceKey));
+            this.AzMonLogType = Preconditions.CheckNonWhiteSpace(azMonLogType, nameof(azMonLogType));
+
+            this.Endpoints = new List<string>();
+            foreach (string endpoint in endpoints.Split(","))
```
philipktlin: what happens if this.Endpoints is empty? is it valid?

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                if (!string.IsNullOrWhiteSpace(endpoint))
+                {
+                    this.Endpoints.Add(endpoint);
+                }
+            }
+
+            this.ScrapeFrequencySecs = Preconditions.CheckRange(scrapeFrequencySecs, 0);
```
philipktlin: does this allows zero? if zero is given, scrape will continue to run all the time?

edge-modules/MetricsCollector/Program.cs
and-rewsmith: I wanted to keep it as close as possible to where it is needed. It is not needed before this point because we don't use the cts or other objects until the body of the using statement and after.

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: It splits into an empty array.

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: I don't think we should allow 0. I thought this made sure the first arg was > the second arg, but that is not the case.

and-rewsmith: But since its a periodic task I think it would run all the time.

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -0,0 +1,91 @@
...
+            this.UploadTarget = Preconditions.CheckNotNull(uploadTarget);
```
philipktlin: no null is needed for enum

edge-modules/MetricsCollector/docker/windows/arm32v7/Dockerfile
```diff
@@ -0,0 +1,11 @@
+ARG base_tag=1.0.3-windows-arm32v7
+ARG base_registry
+FROM ${base_registry}/azureiotedge/azureiotedge-module-base:${base_tag}
+
+ARG EXE_DIR=.
```
philipktlin: why no MODULE_NAME here?

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: Are there any checks you recommend here?

edge-modules/MetricsCollector/docker/windows/arm32v7/Dockerfile
and-rewsmith: I'm not sure what MODULE_NAME means in this context. The dockerfile here follows the same format as in load-gen.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
and-rewsmith: https://stackoverflow.com/questions/980390/consistent-hashcodes-for-dictionaries-in-c-sharp

I read this post and it says that for reference types, the expectation is that different references will have different hash codes even if equals() returns true. Thus, for the same reason Dictionary.GetHashCode() returns different hashes for two dicts with the same keys/vals, the Metric should also return different hashes for two Metrics with the same data.

and-rewsmith: Also this method should actually be an override of GetHashCode. The way it is used elsewhere in the code uses a dictionary with this custom function, whereas if we overrode GetHashCode then the dictionary could be a set instead. We can discuss tomorrow.

edge-modules/MetricsCollector/MetricWithCustomTags.cs
lfitchett: Why make this class instead of making metrics have a regular dictionary? Now you have to copy every metric just to add a tag.

edge-modules/MetricsCollector/Settings.cs
lfitchett: Periodic task already does this. 
```
Preconditions.CheckArgument(frequency > TimeSpan.Zero, "Frequency should be > 0");
Preconditions.CheckArgument(startAfter >= TimeSpan.Zero, "startAfter should be >= 0");
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
```diff
@@ -4,17 +4,17 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics
...
     using Microsoft.Azure.Devices.Edge.Util;
 
-    public sealed class Metric : IEquatable<Metric>
+    public class Metric : IEquatable<Metric>
     {
         public DateTime TimeGeneratedUtc { get; }
         public string Name { get; }
         public double Value { get; }
-        public string Tags { get; }
+        public IReadOnlyDictionary<string, string> Tags { get; }
```
lfitchett: Why read only?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsSerializer.cs
```diff
@@ -54,7 +57,7 @@ public static IEnumerable<Metric> BytesToMetrics(byte[] bytes)
 
                     int tagsLength = BitConverter.ToInt32(bytes, index);
                     index = checked(index + sizeof(int));
-                    tags = Encoding.UTF8.GetString(bytes, index, tagsLength);
+                    tags = JsonConvert.DeserializeObject<Dictionary<string, string>>(Encoding.UTF8.GetString(bytes, index, tagsLength));
```
lfitchett: You need to catch possible deserialization errors

edge-modules/MetricsCollector/MetricWithCustomTags.cs
and-rewsmith: Lee, Philip, and I discussed in person. We did not want to make the original metric mutable since the tags are used to in the GetMetricKey() function. For the base metrics class, we settled on reverting the conversion to the tags dictionary in favor of keeping the original string type.

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: I don't think we should depend on periodic task to do input validation.

edge-modules/MetricsCollector/docker/windows/arm32v7/Dockerfile
philipktlin: Please help to follow up with Varun to remove these from linux docker files or add it to Windows Docker file.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/prometheus.net/MetricsProvider.cs
```diff
@@ -13,17 +13,15 @@ public class MetricsProvider : IMetricsProvider
...
         readonly List<string> defaultLabelNames;
-        static readonly Lazy<string> instanceNumber = new Lazy<string>(GetInstanceNumber);
 
         public MetricsProvider(string namePrefix, string iotHubName, string deviceId)
         {
             this.namePrefix = Preconditions.CheckNonWhiteSpace(namePrefix, nameof(namePrefix));
             Preconditions.CheckNonWhiteSpace(iotHubName, nameof(iotHubName));
             Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
-            this.defaultLabelNames = new List<string> { iotHubName, deviceId, instanceNumber.Value };
+            this.defaultLabelNames = new List<string> { iotHubName, deviceId, Guid.NewGuid().ToString() };
```
and-rewsmith: I was deciding whether or not to have this injected into the constructor, but decided against it.  Let me know if you want me to change this.

PR #2021, opened 2019-11-20
Remove call to edgelet GetSystemInfo() and stubbed the result

darobs:
...to remove dependence on ClusterRole in Helm chart.


darobs:
Closing because this has been fixed and we don't need this change.

PR #2022, opened 2019-11-20
Port e2e test changes from master to 1.0.9

and-rewsmith:


smoke/IotEdgeQuickstart/details/Iotedgectl.cs
```diff
@@ -40,7 +40,7 @@ public async Task VerifyNotActive()
 
         public async Task VerifyModuleIsRunning(string name)
         {
-            using (var cts = new CancellationTokenSource(TimeSpan.FromMinutes(2)))
+            using (var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10)))
```
damonbarry: This change isn't in master

smoke/LeafDevice/LeafDevice.cs
```diff
@@ -158,4 +160,4 @@ public LeafDevice Build()
             }
         }
     }
-}
```
and-rewsmith: will fix this

smoke/IotEdgeQuickstart/details/Iotedgectl.cs
and-rewsmith: I opened a PR for master here (currently in draft mode): https://github.com/Azure/iotedge/pull/2024.

I synced with Philip and he said there will be other pipeline changes I need to make, so let's hold the review for after I confirm that we have everything we need. After that, I'll also port everything to the master PR.

varunpuranik: If it is not a port, let's not call it that

varunpuranik:
Can you please add description to this PR?

smoke/IotEdgeQuickstart/details/Iotedgectl.cs
philipktlin: The problem Andrew tried to solve is temp-sensor-func docker image takes too long to download, especially on ARM32. 
 Therefore the workaround for now is to increase this timeout for checking module running; this will not cause other tests to run longer since other tests can finish in 2 minutes previously.  Only temp-sensor-function will run longer.

and-rewsmith: @varunpuranik Just to clarify, everything except for this one line change is already in master. I still agree with you though. Calling it a port doesn't make sense if there are extra changes.

PR #2026, opened 2019-11-21
Update rolebindings for Agent

richma-ms:
The Agent was missing a few rolebindings for managing services and PVCs.

Also added a catch-all log for module deployment k8s exceptions.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -415,9 +404,9 @@ public static void UpdatePvc(V1PersistentVolumeClaim pvc)
                 Log.LogInformation((int)EventIds.UpdatePvc, $"Update PVC {pvc.Metadata.Name}");
             }
 
-            public static void PvcException(Exception ex)
+            public static void DeployModuleException(Exception ex)
```
massand: nit: Can we change accessibility of all exception logger methods to internal static void? It is reasonable code cleanup because we are introducing a potential conflicting name "DeployModule" here.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -369,10 +358,10 @@ enum EventIds
                 CreatePvc,
                 DeletePvc,
                 UpdatePvc,
-                PvcException,
                 CreateServiceAccount,
                 DeleteServiceAccount,
-                UpdateServiceAccount
+                UpdateServiceAccount,
+                DeployModuleException
```
massand:
```suggestion
                DeployModulesException
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -113,6 +113,7 @@ public async Task<EdgeDeploymentStatus> DeployModulesAsync(ModuleSet desiredModu
             }
             catch (HttpOperationException e)
             {
+                Events.DeployModuleException(e);
```
massand:
```suggestion
                Events.DeployModulesException(e);
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -232,37 +233,25 @@ async Task ManagePvcs(V1PersistentVolumeClaimList existing, IEnumerable<V1Persis
...
-                var addingTasks = diff.Added
-                    .Select(
-                        pvc =>
-                        {
-                            Events.CreatePvc(pvc);
-                            return this.client.CreateNamespacedPersistentVolumeClaimAsync(pvc, this.deviceNamespace);
-                        });
-                await Task.WhenAll(addingTasks);
-            }
-            catch (HttpOperationException ex)
```
darobs: Have you tested this w/storage class PVC requests?  The comment in the removed code may still apply.

dmolokanov: Yeah... seems suspicious that try-catch here was added here intentionally. Or the error described in comment section is caused by lack of permissions?

darobs: I may have misinterpreted the error I received while testing, but I'd like confirmation.

richma-ms:
yes, PVC updates and deletes now both work.

---
In reply to: [321194429](https://github.com/Azure/iotedge/pull/2026#issuecomment-321194429) [](ancestors = 321194429)

PR #2027, opened 2019-11-21
Add Release Management API

philipktlin:
Add release management API to get result of IoT Edge e2e test release pipeline results.


tools/IoTEdgeDevOps/DevOpsLib/ReleaseManagement.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <summary>
+        /// This method is used to get latest release result of given release definition Id and branch name with descending order.
+        /// Reference: https://docs.microsoft.com/en-us/rest/api/azure/devops/release/releases/list?view=azure-devops-rest-5.1
+        /// </summary>
+        /// <param name="definitionId">release definition Ids</param>
+        /// <param name="branchName">github repository branch name</param>
+        /// <returns>List of IoT Edge releases</returns>
+        public async Task<List<IoTEdgeRelease>> GetReleasesAsync(ReleaseDefinitionId definitionId, string branchName, int top = 5)
+        {
+            ValidationUtil.ThrowIfNullOrWhiteSpace(branchName, nameof(branchName));
```
yophilav: Is there any reason the `definitionId` is not being checked here?

tools/IoTEdgeDevOps/DevOpsLib/ValidationUtil.cs
```diff
@@ -22,6 +22,14 @@ public static void ThrowIfNonPositive(int value, string field)
                 throw new ArgumentOutOfRangeException(field, "Cannot be less than or equal to zero.");
             }
         }
+        public static void ThrowIfNegative(int value, string field)
```
yophilav: Should this function also check for Null?

tools/IoTEdgeDevOps/DevOpsLib/ValidationUtil.cs
```diff
@@ -35,7 +43,7 @@ public static void ThrowIfNull(object value, string field)
         {
             if (value == null)
             {
-                throw new ArgumentNullException("Cannot be null.", field);
+                throw new ArgumentNullException( field, "Cannot be null.");
```
yophilav:
``` suggestion
throw new ArgumentNullException( field, "Cannot be null or an empty set.");
```

tools/IoTEdgeDevOps/DevOpsLib/VstsModels/VstsEnvironmentStatus.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace DevOpsLib.VstsModels
+{
+    public enum VstsEnvironmentStatus
```
yophilav: Please add a ref to these status value.

tools/IoTEdgeDevOps/DevOpsLibTest/IoTEdgeReleaseEnvironmentTest.cs
```diff
@@ -0,0 +1,100 @@
...
+                    new IoTEdgeReleaseEnvironment(
+                        id,
+                        343406,
+                        VstsEnvironmentStatus.Succeeded));
+            Assert.True(ex.Message.StartsWith("Cannot be negative."));
+            Assert.AreEqual("id", ex.ParamName);
+        }
+
+        [Test]
+        public void TestConstructorWithInvalidDefinitionId([Values(-1, 0)] int definitionId)
```
yophilav: What about defintionId that is positive but not defined in the set that we care about? 
Do we lock those up as well?

tools/IoTEdgeDevOps/DevOpsLibTest/ReleaseManagementTest.cs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            this.httpTest.Dispose();
+        }
+
+        [Test]
+        public async Task TestGetReleasesAsyncWithEmptyBranchNameAsync()
```
yophilav: What about the case when the `branchName` contains JSON escape characters?

tools/IoTEdgeDevOps/DevOpsLibTest/ReleaseManagementTest.cs
```diff
@@ -0,0 +1,196 @@
...
+            string branch = "refs/heads/master";
+            List<IoTEdgeRelease> releases = await this.releaseManagement.GetReleasesAsync(ReleaseDefinitionId.E2ETest, branch).ConfigureAwait(false);
+
+            string requestUri = $"https://vsrm.dev.azure.com/{DevOpsAccessSetting.AzureOrganization}/{DevOpsAccessSetting.AzureProject}/_apis/release/releases?definitionId={ReleaseDefinitionId.E2ETest.IdString()}&queryOrder=descending&$expand=environments&statusFilter=active&$top=5&api-version=5.1&sourceBranchFilter={Url.Encode(branch)}";
+            this.httpTest.ShouldHaveCalled(requestUri)
+                .WithVerb(HttpMethod.Get)
+                .WithBasicAuth(string.Empty, PersonalAccessToken)
+                .Times(1);
+
+            Assert.AreEqual(0, releases.Count);
```
yophilav: Do we want to check the HTTP status as well?

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
```diff
@@ -52,11 +52,11 @@ public async Task<IList<VstsBuild>> GetLatestBuildsAsync(HashSet<BuildDefinition
 
             if (!result.ContainsKey("count") || (int) result["count"] <= 0)
             {
-                return buildDefinitionIds.Select(i => VstsBuild.GetBuildWithNoResult(i, branchName)).ToList();
+                return buildDefinitionIds.Select(i => VstsBuild.CreateBuildWithNoResult(i, branchName)).ToList();
```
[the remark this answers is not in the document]
philipktlin: buildDefinitioniId is coming from buildDefinitionIds which is already validated.  What is sourceBranch?  btw vstsbuild will be refactored into vstsbuild and iotedgebuild in another PR.

tools/IoTEdgeDevOps/DevOpsLib/DevOpsAccessSetting.cs
```diff
@@ -4,6 +4,7 @@ namespace DevOpsLib
     public class DevOpsAccessSetting
     {
         public const string BaseUrl = "https://dev.azure.com";
```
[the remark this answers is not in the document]
philipktlin: BaseUrl is used not just for build, but also workitem and query.  Therefore it is not appropriate to rename as BuildManagementBaseUrl.  Currently I just know that vsrm.dev.azure.com is used for release management (even based on the url vsrm => visual studio release management).  Let me know if vsrm.dev.azure.com is used for other purpose, then we can think about what better name should be.

tools/IoTEdgeDevOps/DevOpsLib/ReleaseEnvironmentId.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            { 10073, "Linux AMD64 Docker" },
+            { 10538, "Linux AMD64 Moby" },
+            { 10075, "RBPi ARM32 Moby" },
+            { 28576, "RBPi ARM64 Docker" },
+            { 13148, "VM Proxy" },
+            { 15664, "WinPro-x64" },
+            { 37326, "WinSvr-x64" },
+            { 50386, "WinIoT-x64" },
+            { 51994, "WinIoT-ARM32" }
```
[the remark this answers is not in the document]
philipktlin: It is different that build definition id; although environment id is also unique in devops, but it doesn't has a hierarchy under a specific release definition.  I am thinking to create a specific class and use enum; will put a TODO to follow up in another PR.

tools/IoTEdgeDevOps/DevOpsLib/ReleaseManagement.cs
philipktlin: it is an enum; don't need to check.

tools/IoTEdgeDevOps/DevOpsLib/ValidationUtil.cs
philipktlin: int type cannot be null.  but i will use string.empty if field is null.

tools/IoTEdgeDevOps/DevOpsLib/ValidationUtil.cs
philipktlin: we didn't check if it is empty set; why change the message?

tools/IoTEdgeDevOps/DevOpsLibTest/IoTEdgeReleaseEnvironmentTest.cs
philipktlin: this is not what this class does.  This cannot be happened in ReleaseManagement as it use releasedefinitionid enum.

tools/IoTEdgeDevOps/DevOpsLibTest/ReleaseManagementTest.cs
philipktlin: This test is not testing this; and I don't think JSON escape characters are problem when de-serialization.  If you have concerns, please add suggestion to one of release management test to change branch name with escape character in test data.

tools/IoTEdgeDevOps/DevOpsLibTest/ReleaseManagementTest.cs
philipktlin: This test should not test logic / state inside GetReleasesAsync method.

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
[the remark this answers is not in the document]
yophilav: What do you have in mind for `iotedgebuild`? 
To me, it makes sense right now that we have `vstsbuild` that represent the Json returning from VSTS build pipeline service API.

tools/IoTEdgeDevOps/DevOpsLib/DevOpsAccessSetting.cs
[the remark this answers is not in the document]
yophilav: Yeah.. after digging around on the link (see above), it seems like they have multiple so-call "Service endpoints" which for DevOps has all the `Build`, and `WIT (Work Item Tracking` in it.  Later they provided what's called "Extension Service Endpoints" which includes `Release` APIs. 
The "Service Endpoints" uses `"https://dev.azure.com"` while "Extension Service Endpoints" uses `<extension>.dev.azure.come`

tools/IoTEdgeDevOps/DevOpsLib/ReleaseEnvironmentId.cs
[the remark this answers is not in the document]
yophilav: Sounds good. Thank you!

tools/IoTEdgeDevOps/DevOpsLibTest/ReleaseManagementTest.cs
yophilav: The key is I'm not sure how Flurl deal with a GET/POST parameter that has an escaped JSON.  Right now we kinda assume that it would handle the escape properly.

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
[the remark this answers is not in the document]
philipktlin: similar to vstsagent and iotedgeagent; vstsagent is for de-serilaize result from vsts api; and iotedgeagent is domain entity and contains business logic specific to iot edge.

tools/IoTEdgeDevOps/DevOpsLibTest/ReleaseManagementTest.cs
philipktlin: We can test whether Flurl with escaped JSON; I think all data will be uri-encoded therefore it should works exactly the same.  Also what escape character do you think it will happen in our scenario.  Can you give me an example?

PR #2028, opened 2019-11-21
Set max total wal size for RocksDB to 512M (#1704)

ancaantochi:
* Set max total wal size to 512M port from 1.0.8 fix


viptickets:
packages/blockchain-wallet-v4-frontend/src/services/WhatsNewService/WhatsNewContent/index.js

varunpuranik:
This is cherry-picked from 1.0.8 branch, right?

ancaantochi:
it is from 1.0.9. 1.0.8 reads directly from env variable

---
In reply to: [565231333](https://github.com/Azure/iotedge/pull/2028#issuecomment-565231333) [](ancestors = 565231333)

PR #2039, opened 2019-11-25
Add a by-value comparer for service accounts...

darobs:
...so we don't update service accounts unless they actually change.

To test in k8s,  create a deployment and:
1. Watch the secrets in the namespace
2. Watch the serviceaccount of a module.

You should only see one service account token for each module.  When the system is stable, the service account for the module should not change, and should only change when the module is added or removed.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/serviceaccount/KubernetesServiceAccountByValueEqualityComparer.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                return false;
+            }
+
+            if (x.GetType() != y.GetType())
+            {
+                return false;
+            }
+
+            return x.Metadata.Name == y.Metadata.Name &&
```
dmolokanov: Should we care about namespace here? I mean if a resource from another namespace accidently got here...

darobs: It's not set in KubernetesServiceAccountMapper , so we don't always have it here.

PR #2031, opened 2019-11-21
Adding Relayer initial logic for connectivity tests

dylanbronson:
Adding initial logic for Relayer. This is just baseline logic. This module will be added to later once we finalize the design of the connectivity tests.

The Relayer simply receives a message and relays it onward. It is very similar to TemperatureFilter with the temperature logic removed.

PR #2041, opened 2019-11-26
Modifying load gen to use PeriodicTasks instead of Timers

dylanbronson:
Using PeriodicTask instead of Timers. Timers will execute off regardless of whether a call has finished or not. PeriodicTasks wait a certain time period after completing the task and then perform the task again.

This will make LoadGen more predictable and stable, which is necessary for our new connectivity tests.

One side effect is that jitter goes away. But I think that jitter hasn't illuminated any problems so far, and I don't exactly see how it would anyway.


edge-modules/load-gen/Program.cs
```diff
@@ -30,42 +32,45 @@ static async Task Main()
...
+                // setup the message PeriodicTask
+                messageTask = Option.Some(new PeriodicTask(
+                    () => GenerateMessageAsync(moduleClient, batchId),
+                    Settings.Current.MessageFrequency,
+                    TimeSpan.FromSeconds(Settings.Current.StartDelay),
+                    Logger,
+                    "Generate message"));
+
+                // setup the twin update PeriodicTask
+                twinUpdateTask = Option.Some(new PeriodicTask(
```
and-rewsmith: This twin update is never checked, so do we need it?

edge-modules/load-gen/Program.cs
```diff
@@ -22,6 +22,8 @@ static async Task Main()
         {
             Logger.LogInformation($"Starting load gen with the following settings:\r\n{Settings.Current}");
 
+            Option<PeriodicTask> messageTask = Option.None<PeriodicTask>();
+            Option<PeriodicTask> twinUpdateTask = Option.None<PeriodicTask>();
```
and-rewsmith: Why option?

edge-modules/load-gen/Program.cs
```diff
@@ -30,42 +32,45 @@ static async Task Main()
...
                 Logger.LogError($"Error occurred during load gen.\r\n{ex}");
             }
+            finally
+            {
+                messageTask.ForEach(x => x.Dispose());
+                twinUpdateTask.ForEach(x => x.Dispose());
+            }
         }
 
         static async Task GenerateMessageAsync(ModuleClient client, Guid batchId)
```
and-rewsmith: This function no longer needs Interlocked if it is a periodic task. It can also use messageIdCounter directly.

edge-modules/load-gen/Settings.cs
```diff
@@ -23,35 +23,33 @@ public class Settings
...
+            string outputName,
+            int startDelay)
         {
             this.MessageFrequency = messageFrequency;
-            this.JitterFactor = jitterFactor;
             this.TwinUpdateFrequency = twinUpdateFrequency;
             this.MessageSizeInBytes = messageSizeInBytes;
             this.TransportType = transportType;
             this.OutputName = outputName;
+            this.StartDelay = startDelay;
```
and-rewsmith: Recommend adding preconditions here

edge-modules/load-gen/Settings.cs
```diff
@@ -61,6 +59,8 @@ public class Settings
 
         public string OutputName { get; }
 
+        public int StartDelay { get; }
+
         public override string ToString()
         {
             return JsonConvert.SerializeObject(this, Formatting.Indented);
```
and-rewsmith: nit: Philip wants to move away from the Json serialization pattern as it makes it easy to accidentally leak secrets.

edge-modules/load-gen/Program.cs
```diff
@@ -30,42 +32,45 @@ static async Task Main()
...
-
-                    completed.Set();
-                    handler.ForEach(h => GC.KeepAlive(h));
-                    Logger.LogInformation("Load Gen complete. Exiting.");
-                }
+                Guid batchId = Guid.NewGuid();
+                Logger.LogInformation($"Batch Id={batchId}");
+
+                // setup the message PeriodicTask
+                messageTask = Option.Some(new PeriodicTask(
```
and-rewsmith: I don't understand how periodic task works w.r.t its cancellation token. I do understand that the ShutdownHandler creates a cancellation token and hooks cancellation into typical "shutdown" actions. It seems that PeriodicTask has it's cancellation token configured to cancel when the object is disposed. My guess is that this works because the underlying task runs async, allowing for the main() to exit then trigger disposal?

edge-modules/load-gen/config/settings.json
```diff
@@ -1,8 +1,8 @@
 {
   "messageFrequency": "00:00:00.02",
   "twinUpdateFrequency": "00:00:05",
-  "jitterFactor": 0.5,
   "transportType": "amqp",
   "messageSizeInBytes": 1024,
-  "outputName": "output1"
+  "outputName": "output1",
+  "startDelay": 2
```
philipktlin: What does it means? 2 mins or 2 seconds?  Please use TimeSpan like messageFrequency.

philipktlin:
1. Since you mentioned this PR updates loadGen for connectivity tests, would you tell me what is the relationship between loadGen and relayer module (in this PR, https://github.com/Azure/iotedge/pull/2031)?

2. Message frequency: I originally think it is the parameter to specify what fast messages send to Edgehub/iot hub.  Do we want to fix this as well as current implementation doesn't reflect this?

edge-modules/load-gen/Settings.cs
varunpuranik: Typically we only add preconditions to public methods.

edge-modules/load-gen/config/settings.json
varunpuranik: +1 on timespan.. or rename it to startDelaySecs

edge-modules/load-gen/Settings.cs
philipktlin: I think for this case, validation is needed for inputs since these inputs are passed directly from configuration file; which will validate the format; but not the value.  For example if startDelay is negative; this doesn't catch when getting from configuration.  We should at least validate messageSizeInBytes greater than 0; and change startDelay to uint.

edge-modules/load-gen/Program.cs
varunpuranik: wrap the tasks around the Shutdown handler cancellation token await.

edge-modules/load-gen/Program.cs
varunpuranik: Same with this

edge-modules/load-gen/Program.cs
```diff
@@ -30,42 +32,45 @@ static async Task Main()
                     ModuleUtil.DefaultTransientRetryStrategy,
                     Logger);
 
-                using (var timers = new Timers())
```
philipktlin: are you going to remove Timers class since no usage anymore?

edge-modules/load-gen/Program.cs
dylanbronson: Can you explain more what you mean @varunpuranik ?

edge-modules/load-gen/Program.cs
dylanbronson: Yes I can do this if we don't think they will be necessary for anything else in the future

dylanbronson:
@philipktlin To answer your questions
1. This would be "Mod1" in your Connectivity Test write up. We can reuse LoadGen, since it will be doing the same thing that it is doing in the other tests - just creating messages and sending them.

2. Do you mean because now the messageFrequency means that the PeriodicTask stops for that long, which won't exactly equal the frequency of messages? I think it will be basically the same, but if we wanted to be more clear we could name it messageDelay or timeBetweenMessages? But I think that messageFrequency still makes the most sense.

philipktlin:
1. If LoadGen is "Mod1", are you going to have another PR to allow loadGen to be configurable to send messages only, but no twin update; as message processor scenario should only focus on messages only.
2. I just want to clarify what do we want to achieve.  If this is what we want for now, it is ok for me.

edge-modules/load-gen/Program.cs
philipktlin: please remove it.

edge-modules/load-gen/Program.cs
philipktlin: Please remove it.

dylanbronson: Removed all twin update logic

edge-modules/load-gen/Program.cs
dylanbronson: Removed option. Disposing correctly now.

edge-modules/load-gen/Settings.cs
dylanbronson: Changed it and added comment explaining why

edge-modules/load-gen/Program.cs
dylanbronson: Put the PeriodicTask in a using statement with an await for the cancellation token cancel

philipktlin:
LGTM; @ancaantochi please help to review.

edge-modules/load-gen/Settings.cs
```diff
@@ -23,47 +25,52 @@ public class Settings
...
 
         public string OutputName { get; }
 
+        public TimeSpan StartDelay { get; }
+
         public override string ToString()
         {
-            return JsonConvert.SerializeObject(this, Formatting.Indented);
+            // serializing in this pattern so that secrets don't accidentally get added anywhere in the future
+            var fields = new Dictionary<string, string>()
```
ancaantochi: nit: I think it is not necessary to serialize as json, just build a string with the fields

PR #2046, opened 2019-11-26
Host and Docker Metrics

lfitchett:
This adds metrics to edge agent exposing both host and docker metrics. A function was added to the edgelet runtime that collects host metrics such as disk space and cpu usage. It also queries the docker api to collect similar metrics for each docker container. This data is passed to edgeAgent through the management api. 

edgeAgent includes these metrics with its existing metrics. There is no distinction made between agent and host/docker metrics. 

The docker api response is not parsed by the daemon, instead edgeAgent parses it when it when computing the metrics. This allows our management api to remain consistent even if docker's api changes.


edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -590,6 +597,95 @@ impl ModuleRuntime for DockerModuleRuntime {
...
+            .list() // Get all modules
+            .and_then(|modules: Vec<Self::Module>| { // Get iterable of stats
+                future::join_all(
+                    modules
+                        .into_iter()
+                        .map(|module| module.name().to_owned())
+                        .map(move |name| {
+                            client
+                                .container_api()
+                                .container_stats(&name, false)
```
arsing: This will fail if the container happens to be deleted after the list call that returned it but before it can be `container_stats`ed. We shouldn't fail the whole `system_resources` API in that case, just skip that container.

Regular `list` had the same problem, which is why we added `list_with_details` that handles the container-not-found error. You should do the same thing here.

edgelet/edgelet-docker/Cargo.toml
```diff
@@ -17,7 +17,9 @@ log = "0.4"
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
+sysinfo = "0.9.5"
```
arsing: Make this `0.9.6` so that it's clear we have to use the version with the bugfix, and so that nobody can accidentally downgrade it later.

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -1,9 +1,11 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 use std::collections::HashMap;
-use std::convert::From;
+use std::convert::{From, TryInto};
```
arsing: Just remove `From`. It's already in prelude so it shouldn't have been explicitly `use`d anyway.

edgelet/edgelet-core/src/module.rs
```diff
@@ -379,6 +379,66 @@ impl SystemInfo {
...
+            used_ram,
+            total_ram,
+            disks,
+            docker_stats,
+        }
+    }
+}
+
+#[derive(Debug, serde_derive::Serialize)]
+pub struct DiskInfo {
```
arsing: Since you've added this to `-core`, can you remove it from `iotedge/src/check/additional_info.rs` ?

lfitchett: I was planning on making additional info use this path instead of doing it itself, but it doesn't make sense in this pr. I'll make a separate issue to do it, but probably mark it as low priority

arsing: What do you mean by "use this path" ? If you mean you were going to have `iotedge check` invoke the `/resources` management endpoint to get the system resources info, then you should definitely not do that. Otherwise `iotedge check` won't be able to collect system resources information unless `iotedged` is up and running.

lfitchett: I mean having check call runtime.system_resources() in check. It won't go through the http endpoint.

lfitchett: Although, now that I look at it the endpoint gets a lot more info, like docker and cpu use. check shouldn't get this, so it should probably be kept separate. 

To address your original comment, the two disk info classes are different, since the check diskinfo stores human readable info (10.6MB) instead of just numbers.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/metrics/SystemResources.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public string FileSystem { get; }
+
+        public string FileType { get; }
+    }
+
+#pragma warning disable SA1300 // Element should begin with upper-case letter
+    public class DockerStats
+    {
+        public string name { get; set; }
```
arsing: I think we would want these to follow C# naming conventions for the fields, and use attributes to override the name in the serialized format. Check with Varun.

edgelet/edgelet-docker/Cargo.toml
```diff
@@ -17,7 +17,9 @@ log = "0.4"
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
+sysinfo = "0.9.6"
 tokio = "0.1.11"
+uptime_lib = "0.1.0"
```
arsing: Needs to be added to THIRDPARTYNOTICES

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -590,6 +596,91 @@ impl ModuleRuntime for DockerModuleRuntime {
...
+
+        let uptime: u64 = uptime_lib::get()
+            .map(|u| u.num_seconds())
+            .unwrap_or_default()
+            .try_into()
+            .unwrap_or_default();
+
+        let current_time = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
```
arsing: `Err` here is unlikely, but please handle it instead of `unwrap()`ping: `.duration_since(...).map(Duration::as_secs).unwrap_or_default()` so that `current_time` just becomes 0 instead.

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -590,6 +596,91 @@ impl ModuleRuntime for DockerModuleRuntime {
...
+            .as_secs();
+        let start_time =
+            system_info.get_process_list()[&process::id().try_into().unwrap()].start_time();
```
arsing: Same here. Either of the two errors here are unlikely (process ID `.try_into()` fails, or process list does not contain this process ID), but please handle it instead of `.unwrap()`ping by making the `start_time` 0.

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -590,6 +596,93 @@ impl ModuleRuntime for DockerModuleRuntime {
...
+            .try_into()
+            .unwrap_or_default();
+
+        let current_time = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs();
+        let start_time = process::id()
+            .try_into()
+            .map(|id| system_info.get_process_list()[&id].start_time())
```
arsing: This will still panic if the process list doesn't contain this process ID, as I said before. You need to use `.get()` instead of `[]`

PR #2052, opened 2019-11-28
EdgeHub: Add support for limiting the SSL protocols enabled in EdgeHub

varunpuranik:
Currently EdgeHub supports the default SSL Protocols - Tls, Tls1.1 and Tls1.2. There is no way to limit the SSL protocols that are supported by EdgeHub. This change allows setting an environment variable SslProtocols to the set of SSL Protocols you want EdgeHub to support. The default remains the same. 
The value for the SslProtocols environment variable is a comma separated string of the supported protocols (case insensitive and with or without the period). Examples - 
"tls, tls1.1, tls1.2"
"Tls12"
"Tls1.1, Tls1.2"


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Constants.cs
```diff
@@ -28,6 +28,7 @@ public static class ConfigKey
             public const string EdgeHubDevServerPrivateKeyFile = "EdgeHubDevServerPrivateKeyFile";
             public const string EdgeHubDevTrustBundleFile = "EdgeHubDevTrustBundleFile";
             public const string EdgeHubClientCertAuthEnabled = "ClientCertAuthEnabled";
+            public const string SslProtocols = "SslProtocols";
```
ancaantochi: >SslProtocols [](start = 48, length = 12)

We should use uppercase as this is the naming convention. We do have old ones which are lowercase and can't be changed.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -24,6 +25,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Service
     public class Program
     {
         const int DefaultShutdownWaitPeriod = 60;
+        const SslProtocols DefaultSslProtocols = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
```
ancaantochi: >DefaultSslProtocols [](start = 27, length = 19)

AMQP library has SslProtocols.Default by default
HTTP protocol head the current default is Tls11 and Tls12, that means we open to support Tls1.0 by default which is less secure.
MQTT has SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12 by default

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Constants.cs
varunpuranik: Are you suggesting any change here?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
varunpuranik: Confirmed with Anca - AMQP library we use also only supports SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12. So we should be good from a backward compat perspective.

varunpuranik:
> Noticed that we support different conversion while parsing TLS version setting. Should we make it identical? Or it is good as is because it is different parts of the system?
> https://github.com/Azure/iotedge/pull/2101/files#diff-14dbbbc5dca73b659a5aafce9e0fa7a1R917-R936
> vs
> https://github.com/Azure/iotedge/pull/2052/files#diff-3f8bd6d0b127b688c6da5c23dba7f5fbR13-R30

yes, I noticed that too. I followed the SslProtocols enum values more or less. Did you get the iotedged ones from some rust source? If not, then we should try to unify.

PR #2057, opened 2019-12-02
Reported properties test - added one more device

ancaantochi:
Syncing reported properties to the cloud had a bug when multiple devices were updating in the same time and one of the update was not synced to the cloud. This PR add a new device to the unit test to make sure the reported property is synced to the cloud for each device.


varunpuranik:
Can you pls add some details about why you made this change?

PR #2058, opened 2019-12-02
[k8s] Runtime now only gets node info if it has cluster RBAC permissions.

darobs:
Added a "enableGetNodesRBAC" to Helm chart.

This gets translated to adding the ClusterRole and ClusterRoleBinding to allow iotedged to get/list nodes.  

iotedged will check if it has permission to list nodes, and if it does, it will get node information, otherwise it will set a default.


edgelet/edgelet-kube/src/settings.rs
```diff
@@ -24,6 +24,7 @@ pub struct Settings {
     device_id: Option<String>,
     device_hub_selector: String,
     proxy: ProxySettings,
+    has_cluster_rbac: Option<bool>,
```
dmolokanov: We can make default `true` via serde features. This way we don't need the field to be Option. https://serde.rs/attr-default.html

edgelet/edgelet-kube/src/runtime.rs
```diff
@@ -183,46 +183,56 @@ where
...
-                        architectures
-                    });
-
-                // Convert a map to a list of architectures
-                let architectures = architectures
-                    .into_iter()
-                    .map(|(name, count)| Architecture {
-                        name,
-                        nodes_count: count,
+        let has_rbac = *self.settings.has_cluster_rbac().unwrap_or(&true);
```
dmolokanov: Interesting. I think bool implements copy trait so `*` is redundant. 
The way we have it right now is ok, just `*` makes me nervous when I see it in the code.

darobs: I think the compiler complained on the "if" statement if I didn't dereference the bool, but that was an earlier iteration of the code, so I can try it again.

edgelet/edgelet-kube/src/settings.rs
darobs: I was looking for that, didn't find it in Rust online docs, thank you!

veyalla:
If we are just looking for Node Observer permissions, can we be a bit more specific with naming? 

“IOTEDGE_HAS_CLUSTER_RBAC” seems very generic.

kubernetes/charts/edge-kubernetes/templates/iotedged-deployment.yaml
```diff
@@ -29,6 +29,12 @@ spec:
           env:
           - name: "IOTEDGE_LOG"
             value: "debug"
+          - name: "IOTEDGE_HAS_CLUSTER_RBAC"
```
massand: We don't require this env var in the daemon, if we just fail gracefully always.

edgelet/edgelet-kube/src/runtime.rs
```diff
@@ -183,46 +183,55 @@ where
...
-                        architectures
-                    });
-
-                // Convert a map to a list of architectures
-                let architectures = architectures
-                    .into_iter()
-                    .map(|(name, count)| Architecture {
-                        name,
-                        nodes_count: count,
+        let fut = if self.settings.has_cluster_rbac() {
```
massand: As mentioned in the helm chart changes, we should aim to query what we can without the added complexity of settings. The existence of the role is sufficient to communicate the operator is interested in grabbing system information.

edgelet/edgelet-kube/src/settings.rs
```diff
@@ -75,6 +77,14 @@ impl Settings {
     pub fn device_hub_selector(&self) -> &str {
         &self.device_hub_selector
     }
+
+    pub fn has_cluster_rbac(&self) -> bool {
+        self.has_cluster_rbac
```
massand: This setting can be translated into a single arch get call that sets this bool.

kubernetes/charts/edge-kubernetes/templates/iotedged-deployment.yaml
darobs: I think I found a nice, clean way to do this.  It might be a little more code, but then I won't have to pass the env var.

PR #2059, opened 2019-12-02
Fail fast sending messages if unerlying channel is not active.

vipeller:
At high load it could happen, that a module got disconnected, but because of batch send, messages still were being sent to the inactive channel, waiting for timeout for every send, causing up to ~5 minutes of hanging before acquiring a new active proxy.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
@@ -404,10 +404,16 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Device
             // The messages need to be processed in order.
             using (await this.serializeMessagesLock.LockAsync())
             {
+                var taskCompletionSource = new TaskCompletionSource<bool>();
+                if (!this.IsActive)
+                {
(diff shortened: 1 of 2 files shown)


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
```diff
@@ -404,10 +404,16 @@ public async Task SendMessageAsync(IMessage message, string input)
             // The messages need to be processed in order.
             using (await this.serializeMessagesLock.LockAsync())
             {
+                var taskCompletionSource = new TaskCompletionSource<bool>();
+                if (!this.IsActive)
+                {
+                    taskCompletionSource.SetException(new EdgeHubIOException("Cannot send message. The communication channel is not active anymore."));
+                    await taskCompletionSource.Task;
```
varunpuranik: We should probably return from here, right?

varunpuranik: Also, do you need the TCS here?

vipeller: it is a convoluted way to throw an exception (and so return), mechanically copied from few lines below - where it makes more sense because those lines not necessarily throw - and you are right, I can throw it directly without the completion source.

PR #2069, opened 2019-12-04
LoadGen changes for connectivity test

philipktlin:
1. Add test run duration and tracking id to Settings for connectivity test.
2. use simple loop instead of PeriodicTask to send messages.

Tested by long haul: https://msazure.visualstudio.com/One/_build/results?buildId=27219029


edge-modules/load-gen/Program.cs
```diff
@@ -77,8 +77,14 @@ static async Task GenerateMessageAsync(ModuleClient client, Guid batchId)
                     messageIdCounter++;
                     message.Properties.Add("sequenceNumber", messageIdCounter.ToString());
                     message.Properties.Add("batchId", batchId.ToString());
+                    message.Properties.Add("trackingId", trackingId);
 
                     await client.SendEventAsync(Settings.Current.OutputName, message);
+
+                    if (messageIdCounter % 1000 == 0)
+                    {
+                        Logger.LogInformation($"Sending {messageIdCounter} messages.");
```
and-rewsmith:
```suggestion
                        Logger.LogInformation($"Sent {messageIdCounter} messages.");
```

edge-modules/load-gen/Program.cs
```diff
@@ -77,8 +77,14 @@ static async Task GenerateMessageAsync(ModuleClient client, Guid batchId)
                     messageIdCounter++;
                     message.Properties.Add("sequenceNumber", messageIdCounter.ToString());
                     message.Properties.Add("batchId", batchId.ToString());
+                    message.Properties.Add("trackingId", trackingId);
```
and-rewsmith: What will this tracking Id be used for?

edge-modules/load-gen/Settings.cs
```diff
@@ -29,20 +29,31 @@ public class Settings
                     configuration.GetValue<ulong>("messageSizeInBytes", 1024),
                     configuration.GetValue("transportType", TransportType.Amqp_Tcp_Only),
                     configuration.GetValue("outputName", "output1"),
-                    configuration.GetValue("startDelay", TimeSpan.FromSeconds(2)));
+                    configuration.GetValue("testStartDelay", TimeSpan.FromMinutes(5)),
+                    configuration.GetValue("testRunDuration", TimeSpan.MaxValue),
+                    configuration.GetValue("trackingId", string.Empty));
```
and-rewsmith: We should also update config.settings.json. Also, since the defaults are specified there, we don't really need these default values here. They do help with readability but at the cost of having to maintain in two places.

edge-modules/load-gen/Settings.cs
```diff
@@ -29,20 +29,31 @@ public class Settings
...
             });
 
         Settings(
```
and-rewsmith: Lee made a comment about DefaultSettings not needing lazy. Do you agree?

and-rewsmith: Lee made a comment about DefaultSettings not needing Lazy. Do you agree?

edge-modules/load-gen/Settings.cs
```diff
@@ -29,20 +29,31 @@ public class Settings
...
+
             this.MessageSizeInBytes = Preconditions.CheckRange<ulong>(messageSizeInBytes, 1);
             this.OutputName = Preconditions.CheckNonWhiteSpace(outputName, nameof(outputName));
-            this.StartDelay = startDelay;
+
+            this.MessageFrequency = messageFrequency;
+            this.TestRunDuration = testRunDuration;
+            this.TestStartDelay = testStartDelay;
+            this.TrackingId = trackingId ?? string.Empty;
             this.TransportType = transportType;
```
and-rewsmith: Recommend adding Preconditions for TransportType and other fields

edge-modules/load-gen/Program.cs
philipktlin: it is used for connectivity test to uniquely identify all test results belong to this test run.  This will help to easily find all results upload to log analytics using this tracking id.

edge-modules/load-gen/Settings.cs
philipktlin: I prefer to put defaults in the code rather than in config file.  Therefore I didn't make change in config/settings.json.

and-rewsmith: If we are keeping the config file it might be a good idea to add the key for the new fields.

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -113,7 +120,9 @@ function process_args() {
                 '-iotHubConnectionString' ) saveNextArg=7;;
                 '-eventHubConnectionString' ) saveNextArg=8;;
                 '-eventHubConsumerGroupId' ) saveNextArg=9;;
-                '-loadGenMessageFrequency' ) saveNextArg=10;;
+                '-testDuration' ) saveNextArg=10;;
+                '-loadGenMessageFrequency' ) saveNextArg=11;;
+
```
and-rewsmith: Is this whitespace intentional?

edge-modules/load-gen/Settings.cs
philipktlin: updated.

edge-modules/load-gen/Settings.cs
philipktlin: TransportType is enum and all values are valid here; messageFrequency, testStartDelay, and testDuration are already validated above.  TrackingId is allowed to be null or whitespace to compatible with loadGen in stress/long haul scenario.

test/connectivity/scripts/connectivityTest.sh
philipktlin: I followed what we did in runE2ETests.sh; and I used ShellCheck online to verify, and it doesn't complain this space.  I will keep it for now.

edge-modules/load-gen/Settings.cs
and-rewsmith: nit: Setting tracking id default empty string in constructor seems awkward to me as we are defining defaults in multiple places. I feel a cleaner mental model is configuration.getValue for defaults and constructor for input validation.

edge-modules/load-gen/Program.cs
```diff
@@ -34,32 +34,32 @@ static async Task Main()
...
+                Logger.LogInformation($"Load gen delay start for {Settings.Current.TestStartDelay}.");
+                await Task.Delay(Settings.Current.TestStartDelay).ConfigureAwait(false);
+
+                DateTime testStartAt = DateTime.UtcNow;
+                while (!cts.IsCancellationRequested && DateTime.UtcNow - testStartAt < Settings.Current.TestDuration)
                 {
-                    Logger.LogInformation("Load gen running.");
-                    await cts.Token.WhenCanceled();
+                    await GenerateMessageAsync(moduleClient, batchId, Settings.Current.TrackingId);
+                    await Task.Delay(Settings.Current.MessageFrequency).ConfigureAwait(false);
```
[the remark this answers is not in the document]
philipktlin: Good question.  In general configureAwait(false) is recommended to use if context switch is not needed.  However it is not used in our edge solution as one of our dependencies, dotnetty requires context synchronization; if configureAwait(false) is used, it will cause unexpected error.  That's why you can't find it used in EdgeAgent/EdgeHub.

I used here as it is applied to Task.Delay; not anything related to SDK.  If you have any concerns, I am ok to remove it.

edge-modules/load-gen/Settings.cs
philipktlin: To answer your question, I would say both are ok; and it is not so critical.  For long answer, please read this post, https://csharpindepth.com/articles/singleton.

philipktlin: If you want to simplify it without using Lazy, please help to create another PR to update all Settings class in Edge solution.  thanks.

edge-modules/load-gen/Program.cs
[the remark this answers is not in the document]
and-rewsmith: My understanding is that .NET core does not support SynchronizationContext and that is the reason configure await false is not needed. I think I am misunderstanding something because if dotnetty requires SynchronizationContext, then I would think that would mean:
- we would not be able to run on .NET core, since .NET core does not support it
- configure await false might be needed to avoid context switches to synchronous methods, but this error you mention would not allow us to do so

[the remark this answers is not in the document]
varunpuranik: We don't use ConfigureAwait anywhere in our code, please remove it from here. It should not be needed.

edge-modules/load-gen/Program.cs
```diff
@@ -77,8 +77,14 @@ static async Task GenerateMessageAsync(ModuleClient client, Guid batchId)
                     messageIdCounter++;
                     message.Properties.Add("sequenceNumber", messageIdCounter.ToString());
                     message.Properties.Add("batchId", batchId.ToString());
+                    message.Properties.Add("trackingId", trackingId);
 
                     await client.SendEventAsync(Settings.Current.OutputName, message);
+
+                    if (messageIdCounter % 1000 == 0)
```
varunpuranik: Nit: Move this to the loop instead..

edge-modules/load-gen/Program.cs
```diff
@@ -34,32 +34,32 @@ static async Task Main()
...
             catch (Exception ex)
             {
                 Logger.LogError($"Error occurred during load gen.\r\n{ex}");
             }
+
+            Logger.LogInformation("Load Gen complete. Exiting.");
         }
 
-        static async Task GenerateMessageAsync(ModuleClient client, Guid batchId)
+        static async Task GenerateMessageAsync(ModuleClient client, Guid batchId, string trackingId)
```
varunpuranik: Nit: Can we rename this to SendMessage or something like that?

edge-modules/load-gen/Program.cs
```diff
@@ -34,32 +34,32 @@ static async Task Main()
...
-                    () => GenerateMessageAsync(moduleClient, batchId),
-                    Settings.Current.MessageFrequency,
-                    Settings.Current.StartDelay,
-                    Logger,
-                    "Generate message"))
+                Logger.LogInformation("Load gen running.");
+                Logger.LogInformation($"Load gen delay start for {Settings.Current.TestStartDelay}.");
+                await Task.Delay(Settings.Current.TestStartDelay).ConfigureAwait(false);
+
+                DateTime testStartAt = DateTime.UtcNow;
```
varunpuranik: Nit: Can you also pls make messageId a local variable in this method?

edge-modules/load-gen/Settings.cs
```diff
@@ -29,20 +29,31 @@ public class Settings
                     configuration.GetValue<ulong>("messageSizeInBytes", 1024),
                     configuration.GetValue("transportType", TransportType.Amqp_Tcp_Only),
                     configuration.GetValue("outputName", "output1"),
-                    configuration.GetValue("startDelay", TimeSpan.FromSeconds(2)));
+                    configuration.GetValue("testStartDelay", TimeSpan.FromMinutes(5)),
```
varunpuranik: Do we really need 5 mins delay? (how about say 2 mins?)

philipktlin: it depends on platform; RPi generally takes longer time to download all required modules.

edge-modules/load-gen/Program.cs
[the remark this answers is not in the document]
varunpuranik: remove COnfigureAwait?

edge-modules/load-gen/Program.cs
```diff
@@ -34,32 +32,39 @@ static async Task Main()
...
+                Logger.LogInformation($"Load gen delay start for {Settings.Current.TestStartDelay}.");
+                await Task.Delay(Settings.Current.TestStartDelay);
+
+                DateTime testStartAt = DateTime.UtcNow;
+                long messageIdCounter = 0;
+                while (!cts.IsCancellationRequested && DateTime.UtcNow - testStartAt < Settings.Current.TestDuration)
                 {
-                    Logger.LogInformation("Load gen running.");
-                    await cts.Token.WhenCanceled();
+                    messageIdCounter++;
```
varunpuranik: Super nit: Start from 1 or 0? :)

varunpuranik: Also, we should not update the messageIdCounter if the sending fails

philipktlin: I want to start id from 1; and counter will increment before calling send method.  It will report to TestAnalyzer (implement soon), for both success and failed calls of SendMessage method.  I expected messageCounter will increase even fail case.

philipktlin: let me know if you prefer starting from 0.

varunpuranik: If we increase the messageCounter in failure case, you will run into the same issue that we have been facing. 
Also, will reporting for each message sent scale? What if we want to use this to stress edgeHub? We will be stressing the Analyzer then as well.

and-rewsmith: In the analyzer there is a trade off here wrt storage. Either we:
1 - Store only a count of status types in-memory which uses a constant amount of memory. We currently use this approach for twins and direct methods.
2 - Store each status in memory so as to make errors more traceable (as we can store individual times). We currently use this approach for messages.

For http request overload, Anca and I have discussed sharding the analyzer in the past which I think could solve the issue if it arises.

edge-modules/load-gen/Program.cs
```diff
@@ -15,75 +15,83 @@ class Program
...
+                while (!cts.IsCancellationRequested &&
+                    (Settings.Current.TestDuration == TimeSpan.Zero || DateTime.UtcNow - testStartAt < Settings.Current.TestDuration))
                 {
-                    Logger.LogInformation("Load gen running.");
-                    await cts.Token.WhenCanceled();
+                    try
+                    {
+                        await SendEventAsync(moduleClient, batchId, Settings.Current.TrackingId, messageIdCounter);
+                        messageIdCounter++;
+                        await Task.Delay(Settings.Current.MessageFrequency).ConfigureAwait(false);
```
varunpuranik:
```suggestion
                        await Task.Delay(Settings.Current.MessageFrequency);
```

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -235,6 +244,7 @@ process_args "$@"
 
 CONTAINER_REGISTRY="${CONTAINER_REGISTRY:-edgebuilds.azurecr.io}"
 E2E_TEST_DIR="${E2E_TEST_DIR:-$(pwd)}"
+TEST_DURATION="${TEST_DURATION:-05:00:00}"
```
varunpuranik: Nit: Let's make the default duration something like 1hr.

PR #2070, opened 2019-12-05
Clean up relay module

philipktlin:
Clean up relay module; and add InputName to Settings.


edge-modules/Relayer/Settings.cs
```diff
@@ -24,24 +26,40 @@ public class Settings
...
 
         Settings(
             TransportType transportType,
+            string inputName,
             string outputName)
         {
-            this.TransportType = Preconditions.CheckNotNull(transportType);
+            this.InputName = Preconditions.CheckNonWhiteSpace(inputName, nameof(inputName));
             this.OutputName = Preconditions.CheckNonWhiteSpace(outputName, nameof(outputName));
+            this.TransportType = transportType;
```
yophilav: Is there a Check if the enum is valid within a set of `TransportType`?

edge-modules/Relayer/Program.cs
```diff
@@ -38,7 +40,7 @@ static async Task Main(string[] args)
             }
             catch (Exception ex)
             {
-                Logger.LogError($"Error occurred during Relayer.\r\n{ex}");
+                Logger.LogError(ex, $"Error occurred during Relayer.");
```
and-rewsmith: We don't need the dollar sign anymore

edge-modules/Relayer/Settings.cs
and-rewsmith: configuration.GetValue will throw an error if invalid

edge-modules/Relayer/Settings.cs
```diff
@@ -24,24 +26,40 @@ public class Settings
...
+        {
+            // serializing in this pattern so that secrets don't accidentally get added anywhere in the future
+            var fields = new Dictionary<string, string>
+            {
+                { nameof(this.InputName), this.InputName },
+                { nameof(this.OutputName), this.OutputName },
+                { nameof(this.TransportType), Enum.GetName(typeof(TransportType), this.TransportType) },
+            };
+
+            return $"Settings:{Environment.NewLine}{string.Join(Environment.NewLine, fields.Select(f => $"{f.Key}={f.Value}"))}";
```
and-rewsmith: I prefer JsonConvert.SerializeObject as it is more readable, but your call here.

edge-modules/Relayer/Settings.cs
philipktlin: Currently all values in TransportType enum are valid here.  Also afaik i don't know there is any method in Precondition class to verify enum.

edge-modules/Relayer/Program.cs
philipktlin:
```suggestion
                Logger.LogError(ex, "Error occurred during Relayer.");
```

edge-modules/Relayer/Settings.cs
philipktlin: I just made it consistently with loadGen with same reason.

and-rewsmith:
LGTM

PR #2072, opened 2019-12-05
[k8s] Runtime now only gets node info if it has cluster RBAC permissi…

darobs:
…ons. (#2058)

* Runtime now only gets node info if it has cluster RBAC permissions.

* Use a serde default, so `has_nodes_rbac` does not need to be optional.

* At startup, have module runtime check "list nodes"
access before attempting to list nodes.

* Changed "enable Cluster RBAC" to "enable get nodes RBAC"

### DIFF
--- a/edgelet/edgelet-kube/src/module/trust_bundle.rs
+++ b/edgelet/edgelet-kube/src/module/trust_bundle.rs
@@ -12,9 +12,10 @@ use kube_client::TokenSource;
 use crate::convert::trust_bundle_to_config_map;
 use crate::{Error, ErrorKind, KubeModuleRuntime};
 
+#[allow(clippy::needless_pass_by_value)]
 pub fn init_trust_bundle<T, S>(
     runtime: &KubeModuleRuntime<T, S>,
-    crypto: &impl GetTrustBundle,
+    crypto: impl GetTrustBundle,
 ) -> impl Future<Item = (), Error = Error>
 where
     T: TokenSource,
@@ -116,7 +117,7 @@ mod tests {
         let crypto = TestHsm::default().with_fail_call(true);
 
         let runtime = create_runtime(settings, service);
-        let task = init_trust_bundle(&runtime, &crypto);
+        let task = init_trust_bundle(&runtime, crypto);
 
         let mut runtime = Runtime::new().unwrap();
         let err = runtime.block_on(task).unwrap_err();
@@ -141,7 +142,7 @@ mod tests {
         let crypto = TestHsm::default().with_cert(cert);
 
         let runtime = create_runtime(settings, service);
-        let task = init_trust_bundle(&runtime, &crypto);
+        let task = init_trust_bundle(&runtime, crypto);
 
         let mut runtime = Runtime::new().unwrap();
         let err = runtime.block_on(task).unwrap_err();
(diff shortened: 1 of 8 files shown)


edgelet/edgelet-kube/src/module/trust_bundle.rs
```diff
@@ -12,9 +12,10 @@ use kube_client::TokenSource;
 use crate::convert::trust_bundle_to_config_map;
 use crate::{Error, ErrorKind, KubeModuleRuntime};
 
+#[allow(clippy::needless_pass_by_value)]
```
dmolokanov: Yes, that's why I changed to borrow initilally.

edgelet/edgelet-kube/src/settings.rs
```diff
@@ -75,6 +82,14 @@ impl Settings {
     pub fn device_hub_selector(&self) -> &str {
         &self.device_hub_selector
     }
+
+    pub fn has_nodes_rbac(&self) -> bool {
+        self.has_nodes_rbac
+    }
+
+    pub fn default_nodes_rbac() -> bool {
```
dmolokanov: It doesn't have to be `pub` right?

edgelet/kube-client/src/client.rs
```diff
@@ -77,6 +78,58 @@ where
     S::Error: Fail,
     Body: From<<S as Service>::ResBody>,
 {
+    pub fn is_subject_allowed(
+        &mut self,
+        resource: Option<String>,
```
dmolokanov: Why options for function args? Doest it make sense to call this method if either of args is `None`?

edgelet/kube-client/src/client.rs
```diff
@@ -77,6 +78,58 @@ where
...
+        verb: Option<String>,
+    ) -> impl Future<Item = bool, Error = Error> {
+        let subject_access_review = api_authorize::SelfSubjectAccessReview {
+            spec: api_authorize::SelfSubjectAccessReviewSpec {
```
dmolokanov: k8s_openapi contracts usually have Default trait implemented for them.
```rust
        api_authorize::SelfSubjectAccessReview::create_self_subject_access_review(
            &api_authorize::SelfSubjectAccessReview {
                spec: api_authorize::SelfSubjectAccessReviewSpec {
                    resource_attributes: Some(api_authorize::ResourceAttributes {
                        resource,
                        verb,
                        ..api_authorize::ResourceAttributes::default()
                    }),
                    ..api_authorize::SelfSubjectAccessReviewSpec::default()
                },
                ..api_authorize::SelfSubjectAccessReview::default()
            },
            api_authorize::CreateSelfSubjectAccessReviewOptional::default(),
        )
```

edgelet/kube-client/src/client.rs
```diff
@@ -77,6 +78,58 @@ where
     S::Error: Fail,
     Body: From<<S as Service>::ResBody>,
 {
+    pub fn is_subject_allowed(
+        &mut self,
+        resource: Option<String>,
+        verb: Option<String>,
+    ) -> impl Future<Item = bool, Error = Error> {
```
dmolokanov: I noticed that all the rest methods return k8s objects, but this one returns `bool`. Does k8s payload provide additional info that can be helpfull later on?

darobs: Doesn't hurt to return the SubjectAccessReviewStatus struct.

darobs:
cherry-picked changes from #2080

PR #2074, opened 2019-12-05
DirectMethodCloudSender / DirectMethodSender Consistency

and-rewsmith:
This PR aims to make the both direct method modules consistent.


edge-modules/DirectMethodCloudSender/Settings.cs
```diff
@@ -0,0 +1,70 @@
...
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Configuration;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Converters;
+    using Newtonsoft.Json.Serialization;
+
+    [JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
+    public class Settings
+    {
+        static readonly Lazy<Settings> DefaultSettings = new Lazy<Settings>(
```
and-rewsmith: I am leaving this Lazy

edge-modules/DirectMethodSender/src/Settings.cs
```diff
@@ -0,0 +1,65 @@
...
+        public string TargetModuleId { get; }
+
+        [JsonConverter(typeof(StringEnumConverter))]
+        public TransportType TransportType { get; }
+
+        public TimeSpan DirectMethodDelay { get; }
+
+        public Uri AnalyzerUrl { get; }
+
+        public override string ToString()
```
and-rewsmith: If we decide to move forward with this PR I will follow the pattern from loadgen

edge-modules/DirectMethodCloudSender/Settings.cs
```diff
@@ -0,0 +1,70 @@
...
+        [JsonConverter(typeof(StringEnumConverter))]
+        public TransportType TransportType { get; }
+
+        public TimeSpan DirectMethodDelay { get; }
+
+        public Uri AnalyzerUrl { get; }
+
+        public string ServiceClientConnectionString { get; }
+
+        public override string ToString()
```
and-rewsmith: If we decide to move forward with this PR I will follow the pattern from loadgen

edge-modules/DirectMethodCloudSender/Settings.cs
```diff
@@ -0,0 +1,70 @@
...
+            string targetModuleId,
+            TransportType transportType,
+            TimeSpan directMethodDelay,
+            Uri analyzerUrl,
+            string serviceClientConnectionString)
+        {
+            this.DeviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
+            this.TargetModuleId = Preconditions.CheckNonWhiteSpace(targetModuleId, nameof(targetModuleId));
+            this.TransportType = Preconditions.CheckNotNull(transportType, nameof(transportType));
+            this.DirectMethodDelay = Preconditions.CheckNotNull(directMethodDelay);
```
ancaantochi: >CheckNotNull [](start = 51, length = 12)

it can't be null because TimeSpan is struct

edge-modules/DirectMethodCloudSender/src/Program.cs
```diff
@@ -2,119 +2,90 @@
...
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Azure.Devices;
     using Microsoft.Azure.Devices.Edge.ModuleUtil;
     using Microsoft.Azure.Devices.Edge.Util;
-    using Microsoft.Extensions.Configuration;
     using Microsoft.Extensions.Logging;
-    using IotHubConnectionStringBuilder = Microsoft.Azure.Devices.IotHubConnectionStringBuilder;
-    using TransportType = Microsoft.Azure.Devices.Client.TransportType;
+    using TransportType2 = Microsoft.Azure.Devices.TransportType;
```
ancaantochi: >TransportType2 [](start = 10, length = 14)

why is this TransportType2?

edge-modules/DirectMethodCloudSender/Settings.cs
ancaantochi: yes, please make sure you use the same patttern

---
In reply to: [354614192](https://github.com/Azure/iotedge/pull/2074#discussion_r354614192) [](ancestors = 354614192)

edge-modules/DirectMethodCloudSender/src/Program.cs
```diff
@@ -2,119 +2,90 @@
...
-            completed.Set();
-            handler.ForEach(h => GC.KeepAlive(h));
-            Logger.LogInformation("DirectMethodCloudSender Main() finished.");
-            return 0;
+                completed.Set();
+                handler.ForEach(h => GC.KeepAlive(h));
+                Logger.LogInformation("DirectMethodCloudSender Main() finished.");
+            }
+            catch (Exception ex)
+            {
```
ancaantochi: I think we should try the exception because otherwise the initialization fails but the module keeps running and not doing anything. If it throws it will exit and then edgeAgent will retry to start

edge-modules/DirectMethodSender/src/Program.cs
```diff
@@ -2,87 +2,101 @@
...
 
-            completed.Set();
-            handler.ForEach(h => GC.KeepAlive(h));
-            Logger.LogInformation("DirectMethodSender Main() finished.");
-            return 0;
+                completed.Set();
+                handler.ForEach(h => GC.KeepAlive(h));
+                Logger.LogInformation("DirectMethodSender Main() finished.");
+            }
+            catch (Exception e)
```
ancaantochi: Same as other comment in CloudSender, throw

edge-modules/DirectMethodSender/src/Program.cs
```diff
@@ -2,87 +2,101 @@
...
-                        await moduleClient.SendEventAsync("AnyOutput", new Message(Encoding.UTF8.GetBytes("Direct Method Call succeeded.")));
-                        break;
+                        Logger.LogDebug(statusMessage);
+                    }
+                    else
+                    {
+                        Logger.LogError(statusMessage);
                     }
+
+                    CallAnalyzerToReportStatus(targetModuleId, response, analyzerClient);
```
ancaantochi: >CallAnalyzerToReportStatus [](start = 20, length = 26)

nit: rename to ReportStatus

edge-modules/DirectMethodCloudSender/src/Program.cs
and-rewsmith: I think I'm missing something. If initialization fails then won't the module enter this catch block then exit?

varunpuranik:
There seems significant overlap between the 2. Any reason we cannot merge them? Seems like all we will need is a wrapper around the ServiceClient and ModuleClient that supports invoking methods and returning the result.

and-rewsmith:
@varunpuranik We discussed in standup and decided that it would be best to merge this change and handle the module combination in a separate PR. That way the work doesn't interfere with Bear's modifications and he can handle this in a way that makes sense for both longhaul and connectivity tests.

and-rewsmith:
I smoke tested this and it failed due to the analyzer client 404-ing on the report. The direct method calls themselves were fine.

PR #2076, opened 2019-12-06
Refactor for test result entity for TestAnalyzer

philipktlin:
Currently CloudOperationStatus is used for Twin and Direct Method test to report result to TestAnalyzer.  For connectivity test, we would like to create a new 'Analyzer" named TestResultCoordinator, and introduce a new result entity instead of using the existing one.  We want to reuse DM Cloud Sender and Twin Tester in connectivity test, but we don't want to make these module to send different result data to TestAanalyzer and TestResultCoordinator.  As a result, I refactored existing code to handle new test result type and convert back to the old type, therefore underlying implementation in TestAnalyzer don't need to change.

Note: we expect TestAnalyzer will be retired in the future.  Also this PR doesn't relocate some of the file location.  This will be done in another PR.


edge-modules/TestAnalyzer/Controllers/TestResultType.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestAnalyzer.Controllers
+{
+    public enum TestResultType
+    {
+        Undefined,
```
and-rewsmith: Why define this? Does it act as a default type since it is the first?

edge-modules/TestAnalyzer/swagger.yaml
```diff
@@ -127,19 +106,21 @@ definitions:
...
+    - "resultType"
+    - "createdOn"
     properties:
-      moduleId:
+      source:
         type: "string"
-      statusCode:
+      result:
+        type: "string"
+      resultType:
```
and-rewsmith: There is a way to make this an enum: https://swagger.io/docs/specification/data-models/enums/

I'm not sure if it is worth doing since the type would remain as string.

edge-modules/TestAnalyzer/Controllers/TestOperationResultController.cs
```diff
@@ -0,0 +1,33 @@
...
+    using System.IO;
+    using System.Net;
+    using System.Threading.Tasks;
+    using Microsoft.AspNetCore.Mvc;
+
+    [Route("api/[controller]")]
+    [ApiController]
+    public class TestOperationResultController : Controller
+    {
+        [HttpPost]
```
and-rewsmith: In the other controllers we usually add a comment with the uri endpoint

edge-modules/TestAnalyzer/Controllers/TestOperationResultController.cs
```diff
@@ -0,0 +1,33 @@
...
+        public async Task<StatusCodeResult> PostAsync(TestOperationResult result)
+        {
+            LegacyTestOperationResult legacyResult = LegacyTestOperationResult.Convert(result);
+
+            switch (result.ResultType)
+            {
+                case TestResultType.LegacyDirectMethod:
+                    await ReportingCache.Instance.AddDirectMethodStatusAsync(legacyResult);
```
varunpuranik: Why can't these methods take TestOperationResult instead? It should not be too much work?

edge-modules/TestAnalyzer/LegacyTestOperationResult.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestAnalyzer
+{
+    using System;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using TestAnalyzer.Controllers;
+
+    public class LegacyTestOperationResult
```
varunpuranik: Can we try to avoid this? How much work would it be?

edge-modules/TestAnalyzer/TestOperationResult.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestAnalyzer
+{
+    using System;
+    using TestAnalyzer.Controllers;
+
+    public class TestOperationResult
```
varunpuranik: Please make it immutable

edge-modules/TestAnalyzer/LegacyTestOperationResult.cs
```diff
@@ -0,0 +1,33 @@
...
+            this.ResponseDateTime = responseDateTime;
+        }
+
+        public string ModuleId { get; }
+
+        public string StatusCode { get; }
+
+        public DateTime ResponseDateTime { get; }
+
+        public static LegacyTestOperationResult Convert(TestOperationResult result)
```
varunpuranik: Nit: rename to Create

edge-modules/TestAnalyzer/Controllers/TestOperationResultController.cs
```diff
@@ -0,0 +1,33 @@
...
+    using Microsoft.AspNetCore.Mvc;
+
+    [Route("api/[controller]")]
+    [ApiController]
+    public class TestOperationResultController : Controller
+    {
+        [HttpPost]
+        public async Task<StatusCodeResult> PostAsync(TestOperationResult result)
+        {
+            LegacyTestOperationResult legacyResult = LegacyTestOperationResult.Convert(result);
```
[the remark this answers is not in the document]
philipktlin: No.  All modules are updated to send TestOperationResult; the reason of keeping LegacyTestOperationResult is to minimize changes to existing logic in TestAnalyzer.

edge-modules/TestAnalyzer/Controllers/TestResultType.cs
philipktlin: Undefined will be the default value of TestResultType if someone define it without assigning a value.  Therefore this will ensure values other than "undefined" is explicitly assigned.

edge-modules/TestAnalyzer/Controllers/TestOperationResultController.cs
philipktlin: Initially I can do it and see.

edge-modules/TestAnalyzer/TestOperationResult.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestAnalyzer
```
dylanbronson: Where do we foresee TestOperationResult living for good? If other tests will use it, it should eventually be under test/ subfolder. Maybe in the Microsoft.Azure.Devices.Edge.Test.Common project?

edge-modules/TestAnalyzer/TestOperationResult.cs
```diff
@@ -0,0 +1,31 @@
...
+    using System;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public class TestOperationResult
+    {
+        public TestOperationResult(
+            string source,
+            TestResultType resultType,
+            string result,
+            DateTime createdOn)
```
dylanbronson: nit: "createdAt" might fit better than "createdOn"

edge-modules/TestAnalyzer/swagger.yaml
philipktlin: I changed back to use string for test result type in TestOperationResult class; which don't need to re-generate AnalyzerClient when trying to support a new result type.

philipktlin: @and-rewsmith, I don't need to re-generate AnalyzerClient since swagger.yaml is not changed. right?

edge-modules/TestAnalyzer/TestOperationResult.cs
philipktlin: we will move it when start working on TestResultCoordinator.

edge-modules/TestAnalyzer/TestOperationResult.cs
```diff
@@ -0,0 +1,29 @@
...
+            this.ResultType = Preconditions.CheckNonWhiteSpace(result, nameof(resultType));
+            this.Result = Preconditions.CheckNonWhiteSpace(result, nameof(result));
+            this.CreatedAt = createdAt;
+        }
+
+        public string Source { get; }
+
+        public string ResultType { get; }
+
+        public string Result { get; }
```
ancaantochi: >Result [](start = 22, length = 6)

maybe call it ResultDetailsJson or ResultContentJson?

edge-modules/TestAnalyzer/ReportingCache.cs
```diff
@@ -175,5 +183,19 @@ public int Compare(MessageDetails msg1, MessageDetails msg2)
...
+
+        interface IResultHandler
+        {
+            Task AddResultAsync(TestOperationResult result);
+        }
+    }
+
+    static class TestResultConstants
+    {
+        internal static class ResultType
```
ancaantochi: >ResultType [](start = 30, length = 10)

remove TestResultConstants class and have ResultType directly

edge-modules/TestAnalyzer/ReportingCache.cs
```diff
@@ -175,5 +183,19 @@ public int Compare(MessageDetails msg1, MessageDetails msg2)
                 return msg1.SequenceNumber.CompareTo(msg2.SequenceNumber);
             }
         }
+
+        interface IResultHandler
```
ancaantochi: >IResultHandler [](start = 18, length = 14)

Is this used somewhere?

philipktlin: cleaned up.  thanks.

edge-modules/TestAnalyzer/TestOperationResult.cs
philipktlin: in order to backward compatible, Result property can be just the value in plain text.

edge-modules/TestAnalyzer/swagger.yaml
philipktlin: re-generated since I rename ResultType to Type in TestOperationResult.

PR #2081, opened 2019-12-06
Reduce tempSensor message count in end-to-end tests

damonbarry:
The 'TempSensor' end-to-end test frequently fails in our test runs. This is because the previous test, 'TempFilter', uses a different instance of the tempSensor module (tempSensor1) to send a couple messages before the test completes and deregisters the module. When the 'TempSensor' test begins, edgeHub is still trying to send at least one of tempSensor1's messages but keeps getting a 404 from IoT Hub because that moduleId no longer exists. The error handling and retries in edgeHub make it difficult for the 'TempSensor' test to successfully send a message to the cloud before the test times out.

This change limits the number of messages the tempSensor module sends in the end-to-end tests. The default behavior of the module is to send every 5 seconds, up to 500 messages. In these tests we only need to send and verify 1 message. So that's what we'll do.


test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -86,7 +90,8 @@ public async Task TempFilter()
             EdgeDeployment deployment = await this.runtime.DeployConfigurationAsync(
                 builder =>
                 {
-                    builder.AddModule(tempSensorModule.Name, tempSensorModule.Image);
+                    builder.AddModule(tempSensorModule.Name, tempSensorModule.Image)
+                        .WithEnvironment(new[] { ("MessageCount", "1") });
```
yophilav: Woops, I forgot about `TempFilterFunc()`. It needs to be updated as well.

damonbarry: Done. Good catch, thanks!

PR #2088, opened 2019-12-07
Adding TestResultCoordinator to build images

dylanbronson:
Adding dockerfiles, and adding TestResultCoordinator to build images.


builds/misc/images.yaml
```diff
@@ -119,6 +119,13 @@ jobs:
           filePath: scripts/linux/buildImage.sh
           arguments: -r $(registry.address) -u $(registry.user) -p $(registry.password) -i azureiotedge-relayer -n microsoft -P Relayer --target-arch aarch64
 
+      # TestResultCoordinator - Not Using Template for ARM64 because we have 2 different .NET Core.
+      - task: Bash@3
+        displayName: Build Image - Test Result Coordinator - aarch64
+        inputs:
+          filePath: scripts/linux/buildImage.sh
+          arguments: -r $(registry.address) -u $(registry.user) -p $(registry.password) -i azureiotedge-test-result-coordinator -n microsoft -P TestResultCoordinator --target-arch aarch64
```
yophilav: Just making sure: we currently don't want to publish this module to our official Container Registry, right?
Because if you want to publish it to the CR, you will need to add to https://github.com/Azure/iotedge/blob/56b85393f51f8807476923af20710f1bda5dfa41/builds/misc/images.yaml#L440-L452

yophilav:
Is `TestResultCoordinator` going to be used in Stress & LongHaul tests? If so, I suggest move the `iotedge/test/modules/TestResultCoordinator` to `iotedge/edge-modules/TestResultCoordinator`.

builds/misc/images.yaml
dylanbronson: We don't want to do this because we generally don't give out our test modules (except Temp sensor) to support them continuously

dylanbronson:
We are eventually going to move all of our test modules to iotedge/test/modules, so we want this one there

PR #2092, opened 2019-12-09
Fix analyzer to work with the new test result operation

ancaantochi:
Change result type to legacytwin and legacydirectmethod because this is what the modules report
fix type value that was set to result


edge-modules/TestAnalyzer/ReportingCache.cs
```diff
@@ -187,8 +187,8 @@ public int Compare(MessageDetails msg1, MessageDetails msg2)
     
     static class TestOperationResultType
     {
-        internal const string AnalyzerDirectMethod = "analyzerdirectmethod";
-        internal const string AnalyzerTwin = "analyzertwin";
+        internal const string AnalyzerDirectMethod = "LegacyDirectMethod";
+        internal const string AnalyzerTwin = "LegacyTwin";
```
and-rewsmith: nit: I agree with module name prefaces when they add context, but here I feel like it might not be needed. The downside to this is if we change the module name we need to update all the fields sharing the preface.

philipktlin: if you want to change values for these constans, please make sure updating the usage of analyzerClient.ReportResultAsync in DirectMethodCloudSender and TwinTester.

PR #2096, opened 2019-12-10
Metrics Upload to IoTHub

lfitchett:
Make metrics IoTHub publisher match Damon's changes. It also extends the module client wrapper to include sending a message. There is a follow-up pr that adds retry logic: https://github.com/Azure/iotedge/pull/2100 

It also adds a verbose flag to PeriodicTask


edge-util/src/Microsoft.Azure.Devices.Edge.Util/PeriodicTask.cs
```diff
@@ -85,9 +89,22 @@ async Task DoWork()
...
+                        void Log(string message)
+                        {
+                            if (this.verbose)
+                            {
+                                this.logger.LogInformation(message);
+                            }
+                            else
+                            {
+                                this.logger.LogDebug(message);
+                            }
```
and-rewsmith: What is the difference between these methods?

edge-util/src/Microsoft.Azure.Devices.Edge.Util/PeriodicTask.cs
```diff
@@ -85,9 +89,22 @@ async Task DoWork()
                 {
                     try
                     {
-                        this.logger.LogInformation($"Starting periodic operation {this.operationName}...");
+                        // void Log(string message) => this.verbose ? this.logger.LogInformation(message) : this.logger.LogDebug(message);
```
and-rewsmith: You didn't like the one-liner?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -331,7 +353,10 @@ enum EventIds
                 UpdatedReportedProperties,
                 ErrorUpdatingReportedProperties,
                 GotModuleClient,
-                GettingModuleClient
+                GettingModuleClient,
+                SendEvent,
+                SendEventClientEmpty,
+                ErrorSendingEvent,
```
and-rewsmith: I'm assuming no retries?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -119,6 +119,28 @@ public async Task UpdateReportedPropertiesAsync(TwinCollection patch)
...
+        {
+            Events.UpdatingReportedProperties();
+            try
+            {
+                Option<IModuleClient> moduleClient = this.moduleConnection.GetModuleClient();
+                if (!moduleClient.HasValue)
+                {
+                    Events.SendEventClientEmpty();
+                    return;
+                }
```
and-rewsmith: I see you followed the pattern here. Is this protection against the moduleClient being disposed?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/IoTHubMetricsUpload.cs
```diff
@@ -2,34 +2,44 @@
...
-                await this.moduleClient.SendEventAsync(message);
+                Message message = this.BuildMessage(data);
+                await this.edgeAgentConnection.SendEventAsync(message);
             }
         }
+
+        Message BuildMessage(byte[] data)
+        {
+            Message message = new Message(data);
+            message.ContentType = "application/x-azureiot-edgeruntimediagnostics";
```
and-rewsmith: Splitting on "iot" and "edge" seems awkward to me. Is this a standard for variations of ```application/x-azureiot-XXX```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/IoTHubMetricsUpload.cs
```diff
@@ -2,34 +2,44 @@
...
 
         public async Task PublishAsync(IEnumerable<Metric> metrics, CancellationToken cancellationToken)
         {
             Preconditions.CheckNotNull(metrics, nameof(metrics));
             byte[] data = MetricsSerializer.MetricsToBytes(metrics).ToArray();
-            byte[] compressedData = Compression.CompressToGzip(data);
 
             // TODO: add check for too big of a message
-            if (compressedData.Length > 0)
+            if (data.Length > 0)
```
and-rewsmith: Should these checks go in the module client instead?

edge-util/src/Microsoft.Azure.Devices.Edge.Util/PeriodicTask.cs
lfitchett: It wasn't valid c#. I should remove it. I thought it would work, but it didn't for some reason.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
lfitchett: I have logic that adds retries, that will be a pr after this one. I wanted to keep my prs small.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
lfitchett: I don't think this should ever be empty, thats why it errors. I'm not sure, but I just followed the pattern.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/IoTHubMetricsUpload.cs
lfitchett: Yes. This is what damon decided fit best in the iot hub codebase. We're just adding to an existing path.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/IoTHubMetricsUpload.cs
lfitchett: No, not sending an empty message is specific to this code. Sometimes an empty body is valid.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/PeriodicTask.cs
lfitchett: debug is more verbose than information. So if log output is set to information and below, debug won't be output.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/PeriodicTask.cs
varunpuranik: this syntax requires you to return a value :(

PR #2097, opened 2019-12-10
Add Web controller and storage for TestResultCoordinator

dylanbronson:
Adding API Controller for TestResultCoordinator. 

Also adding Settings and Storage

The WebController is simply a POST request that stores a TestOperationResult object. The storage is split up into rocksDB's for each TestOperationResult source.

The sources are added in the config by adding each source in a string separated by semicolons for each ResultSource and by commas within each ResultSource for the Source string and the Type string


test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,84 @@
...
+            catch (Exception ex) when (!ExceptionEx.IsFatal(ex))
+            {
+                Logger.LogError(ex, "Error creating RocksDB store. Falling back to in-memory store.");
+                storeProvider = new StoreProvider(new InMemoryDbStoreProvider());
+            }
+
+            this.resultStores = await this.InitializeStoresAsync(storeProvider, resultSources);
+        }
+
+        private async Task<Dictionary<string, ISequentialStore<TestOperationResult>>> InitializeStoresAsync(StoreProvider storeProvider, List<ResultSource> resultSources)
```
varunpuranik: Avoid private on methods

test/modules/TestResultCoordinator/config/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "OptimizeForPerformance": "true",
+  "StoragePath": "",
+  "ResultSources": "loadGen1.send,messages;loadGen2.send,messages;relayer1.receive,messages;relayer2.receive,messages;relayer;relayer1.send,messages;relayer2.send,messages"
```
varunpuranik: Can we make this an array? Or an object?

test/modules/TestResultCoordinator/Startup.cs
```diff
@@ -0,0 +1,38 @@
...
+        public IConfiguration Configuration { get; }
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddMvc();
+        }
+
+        // TODO: Remove warning disable for Obsolete when project is moved to dotnetcore 3.0
+#pragma warning disable 612, 618
```
varunpuranik: Why do you need this?

test/modules/TestResultCoordinator/config/settings.json
and-rewsmith: If you decide to go with a stringified json array and deserialize in the module, this might be hard to integrate into the pipeline considering the backslash escape chain you have to work around (which behaves differently for linux and windows). I've ran into this issue before which caused problems. If you want to represent objects that aren't stringified, it might be a good idea to set this value in the module twin's desired properties rather than in the environment vars.

test/modules/TestResultCoordinator/Startup.cs
and-rewsmith: This is also in the TestAnalyzer [here](https://github.com/ancaantochi/iotedge/blob/ancan/network-controller/edge-modules/TestAnalyzer/Startup.cs#L25). Looked at the blame and it was taken from this [PR to update dotnet core to 3.0 for arm64](https://github.com/ancaantochi/iotedge/commit/ad345efae692bbf3e28dc3d763f32ab25d667265).

test/modules/TestResultCoordinator/Startup.cs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator
+{
+    using Microsoft.AspNetCore.Builder;
+    using Microsoft.AspNetCore.Hosting;
+    using Microsoft.Extensions.Configuration;
+    using Microsoft.Extensions.DependencyInjection;
+
+    public class Startup
```
and-rewsmith: We might want to consolidate the Startup classes in TestAnalyzer and here into one class living somewhere shared like ModuleLib

test/modules/TestResultCoordinator/Controllers/TestOperationResultController.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    [Route("api/[controller]")]
+    [ApiController]
+    public class TestOperationResultController : Controller
+    {
+        // POST api/TestOperationResult
+        [HttpPost]
+        public async Task<StatusCodeResult> PostAsync(TestOperationResult result)
+        {
+            await TestOperationResultStorage.Instance.AddTestOperationResultAsync(result);
```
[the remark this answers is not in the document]
philipktlin: In Swagger.yaml and TestOperationResult class, it already validates for required fields.  Other validation should be done in verification stage; in order to minimize the processing time for reporting a test result.

test/modules/TestResultCoordinator/Program.cs
```diff
@@ -12,15 +15,27 @@ class Program
...
         {
+            Logger.LogInformation($"Starting load gen with the following settings:\r\n{Settings.Current}");
+
             // TODO: Add Coordinator logic
             (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromSeconds(5), Logger);
 
+            await TestOperationResultStorage.Instance.InitAsync(Settings.Current.StoragePath, new SystemEnvironment(), Settings.Current.OptimizeForPerformance, Settings.Current.ResultSources);
+            Console.WriteLine("TestOperationResultStorage fired up");
+            await CreateWebHostBuilder(args).Build().RunAsync(cts.Token);
+            Console.WriteLine("WebHostBuilder fired up");
```
philipktlin: Move this line up; otherwise this line will be logged until the web service stopped.

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -0,0 +1,87 @@
...
+                   .SetBasePath(Directory.GetCurrentDirectory())
+                   .AddJsonFile("config/settings.json")
+                   .AddEnvironmentVariables()
+                   .Build();
+
+                return new Settings(
+                    configuration.GetValue("WebhostPort", DefaultWebhostPort),
+                    configuration.GetValue<string>("StoragePath", DefaultStoragePath),
+                    configuration.GetValue<bool>("StorageOptimizeForPerformance", true),
+                    configuration.GetValue<string>("ResultSources", DefaultResultSources));
```
philipktlin: result sources can't be empty; so remove default value here.

test/modules/TestResultCoordinator/config/settings.json
varunpuranik: You should be able to use objects and arrays.. take a look at this example - 
https://github.com/Azure/iotedge/blob/master/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs#L111

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -0,0 +1,87 @@
...
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Configuration;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Serialization;
+
+    [JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
+    class Settings
+    {
+        const string DefaultStoragePath = "";
+        const string DefaultWebhostPort = "5001";
```
philipktlin: use ushort

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -0,0 +1,87 @@
...
+                    configuration.GetValue<string>("StoragePath", DefaultStoragePath),
+                    configuration.GetValue<bool>("StorageOptimizeForPerformance", true),
+                    configuration.GetValue<string>("ResultSources", DefaultResultSources));
+            });
+        Settings(string webhostPort, string storagePath, bool optimizeForPerformance, string resultSources)
+        {
+            this.WebhostPort = Preconditions.CheckNonWhiteSpace(webhostPort, nameof(webhostPort));
+            this.StoragePath = storagePath;
+            this.OptimizeForPerformance = Preconditions.CheckNotNull(optimizeForPerformance);
+            this.ResultSources = this.ParseResultSources(Preconditions.CheckNonWhiteSpace(resultSources, nameof(resultSources)));
```
philipktlin: nit: move "Preconditions.CheckNonWhiteSpace(resultSources, nameof(resultSources))" inside ParseResultSources method.

test/modules/TestResultCoordinator/Startup.cs
philipktlin: Not for now; and TestAnalyzer should be removed in the near future.

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,84 @@
...
+    using Microsoft.Azure.Devices.Edge.Storage.RocksDb;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    class TestOperationResultStorage
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger("TestResultCoordinator");
+        Dictionary<string, ISequentialStore<TestOperationResult>> resultStores;
+
+        public static TestOperationResultStorage Instance { get; } = new TestOperationResultStorage();
```
philipktlin: You can initialize this Instance using Settings instead of calling InitAsync later.

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            ISequentialStore<TestOperationResult> resultStore;
+            if (this.resultStores.TryGetValue(testOperationResult.Source, out resultStore))
+            {
+                await resultStore.Append(testOperationResult);
+                return true;
+            }
+            else
+            {
+                throw new InvalidDataException($"result type should be 'messages', 'directMethod', or 'twin'. Current is '{testOperationResult.Type}'.");
```
philipktlin: This is about no match source; you need to add validation for result type.

test/modules/TestResultCoordinator/TestResultCoordinator.csproj
```diff
@@ -21,16 +21,30 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.1.1" />
```
philipktlin: make sure use existing version in the solution for all dependencies; e.g. AspNetCore uses 2.2.0.

test/modules/TestResultCoordinator/config/settings.json
philipktlin: Varun shared that you can use colon (:) in envrionment variable to represent a json value.  For example, key1: { key1.1:value1.1, key1.2: value1.2}; environment variables should be "key1:key1.1"="value1.1" and "key1:key1.2"="value1.2".

philipktlin: After discussion with Varun, we will hard-code for now and enhance later.

test/modules/TestResultCoordinator/Startup.cs
dylanbronson: Since TestAnalyzer is going away, we won't consolidate right now

test/modules/TestResultCoordinator/config/settings.json
dylanbronson: So we've decided to go with environment variables as objects. But we are going to combine it with the configuration for the Report types. So we are going to do this at a later date.

I've hard coded a sample list that encapsulates the first version of tests we will make, and I've added a TODO to change it when we decide exactly how we will have those env vars.

test/modules/TestResultCoordinator/Program.cs
```diff
@@ -12,15 +15,27 @@ class Program
...
 
-        static async Task Main()
+        static async Task Main(string[] args)
         {
+            Logger.LogInformation($"Starting load gen with the following settings:\r\n{Settings.Current}");
+
             // TODO: Add Coordinator logic
             (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromSeconds(5), Logger);
 
+            await TestOperationResultStorage.Instance.InitAsync(Settings.Current.StoragePath, new SystemEnvironment(), Settings.Current.OptimizeForPerformance, Settings.Current.ResultSources);
```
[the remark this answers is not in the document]
dylanbronson: Changed TestOperationResultStorage to be static

test/modules/TestResultCoordinator/Program.cs
```diff
@@ -12,15 +15,27 @@ class Program
...
+            Console.WriteLine("TestOperationResultStorage fired up");
```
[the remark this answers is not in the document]
dylanbronson: Testing remnant that I forgot to clean up. Changed to Logger

test/modules/TestResultCoordinator/Controllers/TestOperationResultController.cs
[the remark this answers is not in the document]
dylanbronson: Added a BadResponse response if validation fails (validation is done in the TestOperationResultStorage class)

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
dylanbronson: Changed this to be a static class, but still using InitAsync

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
dylanbronson: Added type validation.

test/modules/TestResultCoordinator/TestResultCoordinator.csproj
dylanbronson: Updated

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                baseStoragePath = Path.GetTempPath();
+            }
+
+            string storagePath = Path.Combine(baseStoragePath, "TestResultCoordinator");
+            Directory.CreateDirectory(storagePath);
+            return storagePath;
+        }
+
+        public static async Task<bool> AddTestOperationResultAsync(TestOperationResult testOperationResult)
```
philipktlin: nit: AddResultAsync
```suggestion
        public static async Task<bool> AddResultAsync(TestOperationResult testOperationResult)
```

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            (string, ISequentialStore<TestOperationResult>) typeAndResultStore;
+            if (resultStores.TryGetValue(testOperationResult.Source, out typeAndResultStore))
+            {
+                if (testOperationResult.Type.Equals(typeAndResultStore.Item1))
```
philipktlin:
```suggestion
                if (string.Equals(testOperationResult.Type, typeAndResultStore.Item1, StringComparison.OrdinalIgnoreCase))
```

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,89 @@
...
+    using Microsoft.Azure.Devices.Edge.Storage;
+    using Microsoft.Azure.Devices.Edge.Storage.RocksDb;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    class TestOperationResultStorage
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger("TestResultCoordinator");
+        static Dictionary<string, (string, ISequentialStore<TestOperationResult>)> resultStores;
+
```
ancaantochi: I think the storage should just save the TestOperationResult type based on the source, it should not care about result type

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,89 @@
...
+        public static async Task InitAsync(string storagePath, ISystemEnvironment systemEnvironment, bool optimizeForPerformance, List<ResultSource> resultSources)
+        {
+            StoreProvider storeProvider;
```
yophilav:
``` suggestion
            StoreProvider storeProvider;
            Preconditions.CheckNotNull<ISystemEnvironment >(systemEnvironment);
            Preconditions.CheckNotNull<bool>(optimizeForPerformance);
            Preconditions.CheckNotNull<List<ResultSource>>(resultSources);
```

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                Logger.LogError(ex, "Error creating RocksDB store. Falling back to in-memory store.");
+                storeProvider = new StoreProvider(new InMemoryDbStoreProvider());
+            }
+
+            resultStores = await InitializeStoresAsync(storeProvider, resultSources);
+        }
+
+        static async Task<Dictionary<string, (string, ISequentialStore<TestOperationResult>)>> InitializeStoresAsync(StoreProvider storeProvider, List<ResultSource> resultSources)
+        {
```
yophilav:
``` suggestion
        {
            Preconditions.CheckNotNull<StoreProvider>(storeProvider);
            Preconditions.CheckNotNull<List<ResultSource>>(resultSources);
```

test/modules/TestResultCoordinator/Program.cs
```diff
@@ -12,15 +15,26 @@ class Program
     {
         static readonly ILogger Logger = ModuleUtil.CreateLogger("TestResultCoordinator");
 
-        static async Task Main()
+        static async Task Main(string[] args)
         {
-            // TODO: Add Coordinator logic
+            Logger.LogInformation($"Starting load gen with the following settings:\r\n{Settings.Current}");
```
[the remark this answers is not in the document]
philipktlin: good catch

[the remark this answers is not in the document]
dylanbronson: Good catch

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -0,0 +1,83 @@
...
+    using System.Linq;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Configuration;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Serialization;
+
+    [JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
+    class Settings
+    {
+        const string DefaultStoragePath = "";
```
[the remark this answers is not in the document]
dylanbronson: No. Should it?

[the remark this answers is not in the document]
yophilav: Welp, I just got yelled by it 🤣 
I'd say go ahead and do that anyway to be consistence with other modules.

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (resultStores.TryGetValue(testOperationResult.Source, out typeAndResultStore))
+            {
+                if (string.Equals(testOperationResult.Type, typeAndResultStore.Item1, StringComparison.OrdinalIgnoreCase))
+                {
+                    await typeAndResultStore.Item2.Append(testOperationResult);
+                    return true;
+                }
+                else
+                {
+                    string message = $"Result type should be 'messages', 'directMethod', or 'twin'. Current is '{testOperationResult.Type}'.";
```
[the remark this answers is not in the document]
dylanbronson: Yup. Added Enum. 

Relayer and LoadGen will give TestOperationResult

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        public static async Task InitAsync(string storagePath, ISystemEnvironment systemEnvironment, bool optimizeForPerformance, List<ResultSource> resultSources)
+        {
+            StoreProvider storeProvider;
+            try
+            {
+                IDbStoreProvider dbStoreprovider = DbStoreProvider.Create(
+                    new RocksDbOptionsProvider(systemEnvironment, optimizeForPerformance),
+                    GetStoragePath(storagePath),
+                    resultSources.Select(r => r.Source));
```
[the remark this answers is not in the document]
dylanbronson: If the container restarts, we expect the storage to persist (append). If the entire image is remade, then we expect the storage to be recreated. This is what currently happens- the same as it is in current TestAnalyzer. 

We currently have no way of not overflowing memory on disk. That feature will be added probably in the next month.

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -0,0 +1,83 @@
...
+
+                return new Settings(
+                    configuration.GetValue("WebhostPort", DefaultWebhostPort),
+                    configuration.GetValue<string>("StoragePath", DefaultStoragePath),
+                    configuration.GetValue<bool>("StorageOptimizeForPerformance", true));
+            });
+        Settings(ushort webhostPort, string storagePath, bool optimizeForPerformance)
+        {
+            this.WebhostPort = Preconditions.CheckNotNull(webhostPort);
+            this.StoragePath = storagePath;
```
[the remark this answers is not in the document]
philipktlin:
```suggestion
            this.StoragePath = storagePath ?? string.Empty;
```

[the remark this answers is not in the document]
dylanbronson: storagePath by default is empty, so this check shouldn't be there

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
dylanbronson: Changed back

edge-modules/ModuleLib/ResultType.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.ModuleUtil
```
yophilav: Danke! 👍

edge-modules/ModuleLib/ResultType.cs
```diff
@@ -0,0 +1,10 @@
...
+{
+    public enum ResultType
```
philipktlin: Please rename it as "TestResultType" or "TestOperationResultType"

dylanbronson: done

yophilav:
🚢  :shipit:

PR #2101, opened 2019-12-10
Add support of min tls version in config.yaml

dmolokanov:
To provide a user capability to specify min TLS version that `iotedged` exposes for both `workload` and `management` apis:
- create a new parameter under `listen` section in `config.yaml` file (empty by default)
- make sure that both `workload` and `management` apis uses given minimal version for TLS server


edgelet/edgelet-core/src/settings.rs
```diff
@@ -459,6 +464,62 @@ impl Listen {
...
     }
+
+    pub fn min_tls_version(&self) -> Option<&Protocol> {
+        self.min_tls_version.as_ref()
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Protocol {
+    Ssl3,
```
varunpuranik: Ssl3 is deprecated. I don't think we need it.

edgelet/edgelet-core/src/settings.rs
```diff
@@ -847,4 +910,19 @@ mod tests {
...
+    #[test_case("ssl3", Protocol::Ssl3; "when ssl3 provided")]
+    #[test_case("sslv3", Protocol::Ssl3; "when sslv3 provided")]
+    #[test_case("tls" , Protocol::Tls10; "when tls provided")]
+    #[test_case("tls1" , Protocol::Tls10; "when tls1 provided")]
+    #[test_case("tls10", Protocol::Tls10; "when tls10 provided")]
+    #[test_case("tlsv10" , Protocol::Tls10; "when tlsv10 provided")]
+    #[test_case("tls11", Protocol::Tls11; "when tls11 provided")]
+    #[test_case("tlsv11" , Protocol::Tls11; "when tlsv11 provided")]
+    #[test_case("tls12", Protocol::Tls12; "when tls12 provided")]
+    #[test_case("tlsv12" , Protocol::Tls12; "when tlsv12 provided")]
```
varunpuranik: Can you also add upper case tests?

edgelet/edgelet-core/src/settings.rs
dmolokanov: IMO we have to support only the latest TLS version :)

varunpuranik: Yes.. backward compat is the only concern.

edgelet/Cargo.lock
```diff
@@ -2107,6 +2134,18 @@ dependencies = [
  "redox_termios 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
+name = "test-case"
```
arsing: Needs to be added to THIRDPARTYNOTICES

edgelet/edgelet-core/src/settings.rs
```diff
@@ -459,6 +464,59 @@ impl Listen {
     pub fn management_uri(&self) -> &Url {
         &self.management_uri
     }
+
+    pub fn min_tls_version(&self) -> Option<&Protocol> {
+        self.min_tls_version.as_ref()
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
```
arsing: Also `derive(Copy)`

edgelet/edgelet-core/src/settings.rs
```diff
@@ -459,6 +464,59 @@ impl Listen {
     pub fn management_uri(&self) -> &Url {
         &self.management_uri
     }
+
+    pub fn min_tls_version(&self) -> Option<&Protocol> {
```
arsing: `-> Option<Protocol>`

edgelet/edgelet-core/src/settings.rs
```diff
@@ -459,6 +464,59 @@ impl Listen {
...
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for Protocol {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
```
arsing: `let s = deserializer.deserialize_str()?;`

edgelet/edgelet-core/src/settings.rs
```diff
@@ -459,6 +464,59 @@ impl Listen {
...
+        FromStr::from_str(&s).map_err(de::Error::custom)
```
arsing: `FromStr::from_str(s)` -> `s.parse()`

edgelet/edgelet-core/src/settings.rs
dmolokanov: Why do we need Copy? Is it some kind of guideline?

arsing: It's good to have types that can be Copy be Copy. Then clippy can enforce you pass them to functions by value instead of by pointer.

edgelet/edgelet-core/src/settings.rs
dmolokanov: deserialize_str expects a visitor. Why it is better than this?
```rust
impl<'de> Deserialize<'de> for String {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        deserializer.deserialize_string(StringVisitor)
    }

    fn deserialize_in_place<D>(deserializer: D, place: &mut Self) -> Result<(), D::Error>
    where
        D: Deserializer<'de>,
    {
        deserializer.deserialize_string(StringInPlaceVisitor(place))
    }
}
```

edgelet/edgelet-core/src/settings.rs
dmolokanov: Found this one https://stackoverflow.com/questions/44695216/should-simple-enums-derive-copy-is-there-any-difference-in-performance

edgelet/edgelet-core/src/settings.rs
```diff
@@ -847,4 +907,25 @@ mod tests {
...
+    #[test_case("tls11", Protocol::Tls11; "when tls11 provided")]
+    #[test_case("TLS11", Protocol::Tls11; "when uppercase TLS11 Provided")]
+    #[test_case("tlsv11" , Protocol::Tls11; "when tlsv11 Provided")]
+    #[test_case("TLSv11" , Protocol::Tls11; "when uppercase TLSv11 provided")]
+    #[test_case("tls12", Protocol::Tls12; "when tls12 provided")]
+    #[test_case("TLS12", Protocol::Tls12; "when uppercase TLS12 provided")]
+    #[test_case("tlsv12" , Protocol::Tls12; "when tlsv12 provided")]
+    #[test_case("TLSv12" , Protocol::Tls12; "when uppercase TLSv12 provided")]
+    fn it_parses_protocol(value: &str, expected: Protocol) {
+        let actual = Protocol::from_str(value);
```
arsing: `value.parse()` here too.

edgelet/edgelet-core/src/settings.rs
arsing: Oh right, you don't have a visitor here so you can't directly do `deserialize_str`.

The advantage would've been that deserializing to an owned String requires copying and allocation. But it's fine.

THIRDPARTYNOTICES
```diff
@@ -6007,4 +6007,26 @@ OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
 IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 
-***
\ No newline at end of file
+***
+
+MIT License
```
arsing: Missing the name of the crate.

dmolokanov: Fixed

edgelet/edgelet-http/src/lib.rs
```diff
@@ -339,7 +338,7 @@ impl HyperExt for Http {
                         )
                     })?;
 
-                let cert = match cert_manager {
+                let cert = match tls_params.as_ref().map(|params| params.cert_manager) {
```
arsing: If you're going to use a combinator, you might as well use one for the whole thing instead of also using `match`.

```
tls_params.as_ref().and_then(|params| params.cert_manager.get_pkcs12_certificate).ok_or(ErrorKind::CertificateCreationEror)?;
```

edgelet/edgelet-http/src/lib.rs
```diff
@@ -424,3 +435,27 @@ impl HyperExt for Http {
         })
     }
 }
+
+#[allow(dead_code)]
```
arsing: Why not `cfg(unix)` it instead of `allow(dead_code)` ?

arsing: Ah, it *is* used as the parameter regardless. In that case I suggest `#[cfg_attr(not(unix), allow(dead_code))]`

edgelet/edgelet-http/src/lib.rs
```diff
@@ -303,7 +302,7 @@ impl HyperExt for Http {
         &self,
```
arsing: Github won't let me add a comment on the actual line, but also change the attribute on line 299 to `cfg_attr(not(unix), allow(unused_variables))`

Also move it to be on the trait method instead of on the impl.

edgelet/edgelet-http/src/lib.rs
dmolokanov: Looks better. Thanks!

edgelet/edgelet-http/src/lib.rs
```diff
@@ -349,7 +348,19 @@ impl HyperExt for Http {
...
+                        .as_ref()
+                        .map(|protocol| match protocol {
+                            Protocol::Tls10 => native_tls::Protocol::Tlsv10,
+                            Protocol::Tls11 => native_tls::Protocol::Tlsv11,
+                            Protocol::Tls12 => native_tls::Protocol::Tlsv12,
+                        })
+                });
+
                 let tls_acceptor = TlsAcceptor::builder(cert_identity)
+                    .min_protocol_version(min_protocol_version)
```
arsing: This is worse than before, I believe. Before this change, the default was TLS 1.0. With this change, the default is the lowest one that openssl supports, which could be SSL 2.0.

arsing: Do you really need the min version to be `Option` ? You could just make it a required property and do `serde(default = "Tlsv10")` when deserializing it. (Not exactly that because it needs a path of a function, not an expr.)

dmolokanov: Why? if the user didn't put anything `min_protocol_version = None`; then flow as it was before.
If the user put something and we parsed it successfully, it will be either of TLS1.0, TLS1.1, TLS1.2.
Sorry, I didn't quite understand your concern.

edgelet/edgelet-core/src/settings.rs
philipktlin: do we need to test failure case?

dmolokanov: Good idea. Thanks

PR #2103, opened 2019-12-11
Report generator for messaging scenario in Connectivity test

philipktlin:
Continue to work on messaging scenario in connectivity test.  Add ITestResultReportGenerator inteface with CountingReportGenerator implementation; plus add tests for related changes.

### DIFF
--- a/test/modules/TestResultCoordinator/CountingReport.cs
+++ b/test/modules/TestResultCoordinator/CountingReport.cs
@@ -2,24 +2,35 @@
 namespace TestResultCoordinator
 {
     using System.Collections.Generic;
-    
+
+    /// <summary>
+    /// This is a counting report to show test result counts, e.g. expect and match counts; and contains a list of unmatched test results.
+    /// </summary>
+    /// <typeparam name="T">Test result type</typeparam>
     class CountingReport<T> : TestResultReportBase
     {
-        readonly IList<T> missingResults = new List<T>();
-
-        public CountingReport(string expectSource, string actualSource, string resultType)
+        public CountingReport(
+            string expectSource,
+            string actualSource,
+            string resultType,
+            long totalExpectCount,
+            long totalMatchCount,
+            long totalDuplicateResultCount,
+            IReadOnlyList<T> unmatchedResults)
             : base(expectSource, actualSource, resultType)
         {
+            this.TotalExpectCount = totalExpectCount;
+            this.TotalMatchCount = totalMatchCount;
+            this.TotalDuplicateResultCount = totalDuplicateResultCount;
+            this.UnmatchedResults = unmatchedResults ?? new List<T>();
         }
 
-        public int TotalExpectCount { get; set; }
+        public long TotalExpectCount { get; }
 
-        public int TotalMatchCount { get; set; }
+        public long TotalMatchCount { get; }
 
-        public int TotalDuplicateResultCount { get; set; }
+        public long TotalDuplicateResultCount { get; }
 
-        public IList<T> MissingResults => this.missingResults;
-
-        public void AddMissingResult(T result) => this.missingResults.Add(result);
+        public IReadOnlyList<T> UnmatchedResults { get; }
     }
 }
--- a/test/modules/TestResultCoordinator/CountingReport.cs
+++ b/test/modules/TestResultCoordinator/CountingReport.cs
@@ -1,0 +1,25 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator
+{
+    using System.Collections.Generic;
+    
+    class CountingReport<T> : TestResultReportBase
+    {
+        readonly IList<T> missingResults = new List<T>();
+
+        public CountingReport(string expectSource, string actualSource, string resultType)
+            : base(expectSource, actualSource, resultType)
+        {
+        }
+
+        public int TotalExpectCount { get; set; }
+
+        public int TotalMatchCount { get; set; }
+
+        public int TotalDuplicateResultCount { get; set; }
+
+        public IList<T> MissingResults => this.missingResults;
+
+        public void AddMissingResult(T result) => this.missingResults.Add(result);
+    }
+}
--- a/test/modules/TestResultCoordinator/CountingReport.cs
+++ b/test/modules/TestResultCoordinator/CountingReport.cs
@@ -10,14 +10,15 @@ namespace TestResultCoordinator
     class CountingReport<T> : TestResultReportBase
     {
         public CountingReport(
-            string expectSource,
+            string trackingId,
+            string expectedSource,
             string actualSource,
             string resultType,
-            long totalExpectCount,
-            long totalMatchCount,
-            long totalDuplicateResultCount,
+            ulong totalExpectCount,
+            ulong totalMatchCount,
+            ulong totalDuplicateResultCount,
             IReadOnlyList<T> unmatchedResults)
-            : base(expectSource, actualSource, resultType)
+            : base(trackingId, expectedSource, actualSource, resultType)
         {
             this.TotalExpectCount = totalExpectCount;
             this.TotalMatchCount = totalMatchCount;
@@ -25,11 +26,11 @@ namespace TestResultCoordinator
             this.UnmatchedResults = unmatchedResults ?? new List<T>();
         }
 
-        public long TotalExpectCount { get; }
+        public ulong TotalExpectCount { get; }
 
-        public long TotalMatchCount { get; }
+        public ulong TotalMatchCount { get; }
 
-        public long TotalDuplicateResultCount { get; }
+        public ulong TotalDuplicateResultCount { get; }
 
         public IReadOnlyList<T> UnmatchedResults { get; }
     }
--- a/test/modules/TestResultCoordinator/CountingReportGenerator.cs
+++ b/test/modules/TestResultCoordinator/CountingReportGenerator.cs
@@ -19,6 +19,7 @@ namespace TestResultCoordinator
     {
         static readonly ILogger Logger = ModuleUtil.CreateLogger(nameof(CountingReportGenerator));
 
+        readonly string trackingId;
         readonly string expectedSource;
         readonly string actualSource;
         readonly ISequentialStore<TestOperationResult> expectedStore;
@@ -27,12 +28,13 @@ namespace TestResultCoordinator
         readonly ITestResultComparer<TestOperationResult> testResultComparer;
         readonly int batchSize;
 
-        long totalExpectCount = 0;
-        long totalMatchCount = 0;
-        long totalDuplicateResultCount = 0;
+        ulong totalExpectCount = 0;
+        ulong totalMatchCount = 0;
+        ulong totalDuplicateResultCount = 0;
         List<TestOperationResult> unmatchedResults;
 
         public CountingReportGenerator(
+            string trackingId,
             string expectedSource,
             ISequentialStore<TestOperationResult> expectedStore,
             string actualSource,
@@ -41,6 +43,7 @@ namespace TestResultCoordinator
             ITestResultComparer<TestOperationResult> testResultComparer,
             int batchSize = 500)
         {
+            this.trackingId = Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
             this.expectedSource = Preconditions.CheckNonWhiteSpace(expectedSource, nameof(expectedSource));
             this.expectedStore = Preconditions.CheckNotNull(expectedStore, nameof(expectedStore));
             this.actualSource = Preconditions.CheckNonWhiteSpace(actualSource, nameof(actualSource));
@@ -142,6 +145,7 @@ namespace TestResultCoordinator
             }
 
             return new CountingReport<TestOperationResult>(
+                this.trackingId,
                 this.expectedSource,
                 this.actualSource,
                 this.resultType,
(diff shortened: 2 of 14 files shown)


test/modules/TestResultCoordinator/CountingReport.cs
```diff
@@ -0,0 +1,25 @@
...
+    class CountingReport<T> : TestResultReportBase
+    {
+        readonly IList<T> missingResults = new List<T>();
+
+        public CountingReport(string expectSource, string actualSource, string resultType)
+            : base(expectSource, actualSource, resultType)
+        {
+        }
+
+        public int TotalExpectCount { get; set; }
```
varunpuranik: Please make this immutable

test/modules/TestResultCoordinator/CountingReport.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator
+{
+    using System.Collections.Generic;
+    
+    class CountingReport<T> : TestResultReportBase
```
varunpuranik: Can you please add comments on these types to explain what they are supposed to do?

test/modules/TestResultCoordinator/CountingReportGenerator.cs
```diff
@@ -0,0 +1,179 @@
...
+            this.actualSource = Preconditions.CheckNonWhiteSpace(actualSource, nameof(actualSource));
+            this.actualStore = Preconditions.CheckNotNull(actualStore, nameof(actualStore));
+            this.resultType = Preconditions.CheckNonWhiteSpace(resultType, nameof(resultType));
+            this.testResultComparer = Preconditions.CheckNotNull(testResultComparer, nameof(testResultComparer));
+            this.batchSize = batchSize;
+        }
+
+        public string ActualSource => this.actualSource;
+
+        public ISequentialStore<TestOperationResult> AcutalStore => this.actualStore;
```
varunpuranik: typo

test/modules/TestResultCoordinator/CountingReportGenerator.cs
```diff
@@ -0,0 +1,179 @@
...
+            while (expectQueue.Count > 0 && actualQueue.Count > 0)
+            {
+                TestOperationResult expectedResult = expectQueue.Dequeue();
+                TestOperationResult actualResult = actualQueue.Peek();
+
+                this.report.TotalExpectCount++;
+
+                if (this.testResultComparer.Matches(expectedResult, actualResult))
+                {
+                    actualQueue.Dequeue();
```
varunpuranik: Where do you dequeue expectedResult?

test/modules/TestResultCoordinator/CountingReportGenerator.cs
```diff
@@ -0,0 +1,179 @@
...
+                    this.report.TotalMatchCount++;
+                }
+                else
+                {
+                    this.report.AddMissingResult(expectedResult);
+                }
+
+                if (expectQueue.Count == 0 || actualQueue.Count == 0)
```
varunpuranik: Should this be moved out of the while loop?

test/modules/TestResultCoordinator/CountingReportGenerator.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        public ISequentialStore<TestOperationResult> ExpectedStore => this.expectedStore;
+
+        public string ResultType => this.resultType;
+
+        public ITestResultComparer<TestOperationResult> TestResultComparer => this.testResultComparer;
+
+        public int BatchSize => this.batchSize;
+
+        public async Task<ITestResultReport> CreateReportAsync()
```
varunpuranik: This needs some comments as well.

test/modules/TestResultCoordinator/CountingReportGenerator.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            Logger.LogInformation($"Start to generate report by {nameof(CountingReportGenerator)} for Sources [{this.expectedSource}] and [{this.actualSource}]");
+
+            long lastLoadedKeyFromExpectStore = -1;
+            long lastLoadedKeyFromActualStore = -1;
+            TestOperationResult lastLoadedResult = default(TestOperationResult);
+
+            this.report = new CountingReport<TestOperationResult>(this.expectedSource, this.actualSource, this.resultType);
```
varunpuranik: Why is this stored as a field?

test/modules/TestResultCoordinator/CountingReportGenerator.cs
```diff
@@ -0,0 +1,179 @@
...
+            this.resultType = Preconditions.CheckNonWhiteSpace(resultType, nameof(resultType));
+            this.testResultComparer = Preconditions.CheckNotNull(testResultComparer, nameof(testResultComparer));
+            this.batchSize = batchSize;
+        }
+
+        public string ActualSource => this.actualSource;
+
+        public ISequentialStore<TestOperationResult> AcutalStore => this.actualStore;
+
+        public string ExpectedSource => this.expectedSource;
```
varunpuranik: Why are all these being exposed as public properties? Where are they being used?

philipktlin: changed to internal; used in tests.

test/modules/TestResultCoordinator/CountingReportGenerator.cs
philipktlin: removed.

test/modules/TestResultCoordinator/CountingReportGenerator.cs
philipktlin: line 77.

test/modules/TestResultCoordinator/CountingReportGenerator.cs
philipktlin: This is needed to continue to load another batch for verification.

test/modules/TestResultCoordinator/ITestResultComparer.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator
+{
+    /// <summary>
+    /// It is used to compare 2 test results whether they matches or not.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    interface ITestResultComparer<T>
```
ancaantochi: >ITestResultComparer [](start = 14, length = 19)

Why not use IComparer<T> ?

philipktlin: IComparer.Compare method return int to identify whether v1 is greater than or equal to, or less than.  For test result, we just need to know whether it matches or not.  I also don't use IEqualityComparer as it has 2 methods, Equals and GetHashCode; which GetHashCode is not needed for our case.

test/modules/TestResultCoordinator/ITestResultReport.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator
+{
+    /// <summary>
+    /// This defines the basic properties of a test result report.
+    /// </summary>
+    interface ITestResultReport
```
ancaantochi: >ITestResultReport [](start = 14, length = 17)

Can this interface be generic instead of having ResultType? Then ExpectedSource and ActualSource have T as type

test/modules/TestResultCoordinator/ITestResultReport.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        string TrackingId { get; }
+
+        string Title { get; }
+
+        string ResultType { get; }
+
+        string ExpectSource { get; }
```
ancaantochi: >ExpectSource [](start = 15, length = 12)

nit: just Expected and Actual

test/modules/TestResultCoordinator/CountingReportGenerator.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            long lastLoadedKeyFromExpectStore = -1;
+            long lastLoadedKeyFromActualStore = -1;
+            TestOperationResult lastLoadedResult = default(TestOperationResult);
+
+            var expectQueue = new Queue<TestOperationResult>();
+            var actualQueue = new Queue<TestOperationResult>();
+            totalExpectCount = 0;
+            totalMatchCount = 0;
+            totalDuplicateResultCount = 0;
```
ancaantochi: >totalDuplicateResultCount [](start = 12, length = 25)

Where  is this incremented?

test/modules/TestResultCoordinator/CountingReportGenerator.cs
```diff
@@ -0,0 +1,207 @@
...
+                this.totalExpectCount++;
+
+                if (this.testResultComparer.Matches(expectedResult, actualResult))
+                {
+                    actualQueue.Dequeue();
+                    this.totalMatchCount++;
+                }
+                else
+                {
+                    this.unmatchedResults.Add(expectedResult);
```
ancaantochi: >unmatchedResults [](start = 25, length = 16)

If actual is: 1 1 2 3 4 and expected is 1 2 3 4, the second 1 gets reported as unmatched, but is actually a duplicate

ancaantochi: >Add [](start = 42, length = 3)

Add expected and actual

test/modules/TestResultCoordinator/ITestResultReport.cs
philipktlin: This is the expected source name e.g. "loadGen1.Send" or "Relayer1.receive".  Prefer to keep "Source".

test/modules/TestResultCoordinator/ITestResultReport.cs
philipktlin: Every report should only support 1 result type; which is used to derive what result comparer to use.  ExpectedSource and ActualSource is just source name, e.g. "loadGen1.send" or "Relyer1.receive".

test/modules/TestResultCoordinator/CountingReportGenerator.cs
philipktlin: Already have test case covered; results loaded to actual queue already remove duplicates.

philipktlin: what do you mean "Add expected and actual"?

test/modules/TestResultCoordinator/CountingReportGenerator.cs
ancaantochi: it looks like totalExpectCount, totalMatchCount and totalDuplicateResultCount should be local variables. In that case LoadBatchIntoQueueAsync could also return the number of duplicates for the current load

---
In reply to: [356809881](https://github.com/Azure/iotedge/pull/2103#discussion_r356809881) [](ancestors = 356809881)

test/modules/TestResultCoordinator/ITestResultComparer.cs
ancaantochi: yes, but the int 0 means they are equal. IEqualityComparer can also be helpful becuase first it can check that the hashcodes is equal and then that they are equal. If hashcodes are different there is no need to check equals

---
In reply to: [356759050](https://github.com/Azure/iotedge/pull/2103#discussion_r356759050) [](ancestors = 356759050)

test/modules/TestResultCoordinator/CountingReportGenerator.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            (lastLoadedKeyFromExpectStore, _) = await this.LoadBatchIntoQueueAsync(this.expectedSource, this.expectedStore, lastLoadedKeyFromExpectStore, expectQueue);
+            (lastLoadedKeyFromActualStore, lastLoadedResult) = await this.LoadBatchIntoQueueAsync(this.actualSource, this.actualStore, lastLoadedKeyFromActualStore, actualQueue, lastLoadedResult);
+
+            while (expectQueue.Count > 0 && actualQueue.Count > 0)
+            {
+                TestOperationResult expectedResult = expectQueue.Dequeue();
+                TestOperationResult actualResult = actualQueue.Peek();
+
+                this.totalExpectCount++;
```
varunpuranik: You are caching the results in the class.. so if CreateReportAsync is called twice, then it will return incorrect values.
You have to make the values all local, or remember that you have already calculated the result and return the cached values.

test/modules/TestResultCoordinator/ITestResultComparer.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        bool Matches(T value1, T value2);
```
varunpuranik: Super nit: rename to Equals?

test/modules/TestResultCoordinator/CountingReportGenerator.cs
philipktlin: All result is reset between line 86 and 89; therefore it will always report correctly.

test/modules/TestResultCoordinator/ITestResultComparer.cs
philipktlin: I don't want to confuse with object.Equals method; this is just to tell that these 2 results matches.

test/modules/TestResultCoordinator/CountingReportGenerator.cs
philipktlin: Updated to use local variables instead.

PR #2107, opened 2019-12-11
PartitionReceiveHandler for connectivity test

philipktlin:
Create a new PartitionReceiveHandler for connectivity test; which will receive the event from event hub and save event as test result in persisted test result storage.

### DIFF
--- a/test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
+++ b/test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
@@ -1,0 +1,105 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Azure.EventHubs;
+    using Microsoft.Extensions.Logging;
+
+    class PartitionReceiveHandler : IPartitionReceiveHandler
+    {
+        const string DeviceIdPropertyName = "iothub-connection-device-id";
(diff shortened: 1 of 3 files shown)


test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+                        else
+                        {
+                            Logger.LogDebug($"Message for module [{moduleIdFromEvent}] and device [{this.deviceId}] doesn't contain batch id and/or sequence number.");
+                        }
+                    }
+                }
+            }
+        }
+
+        public Task ProcessErrorAsync(Exception error)
```
ancaantochi: >ProcessErrorAsync [](start = 20, length = 17)

Is it used?

test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        public int MaxBatchSize { get; set; }
+
+        public async Task ProcessEventsAsync(IEnumerable<EventData> events)
+        {
+            if (events != null)
+            {
+                foreach (EventData eventData in events)
+                {
+                    eventData.SystemProperties.TryGetValue(TrackingIdPropertyName, out object trackingIdFromEvent);
```
varunpuranik: this cannot be a string?

test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
philipktlin: it is part of IPartitionReceiveHandler.

test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
philipktlin: it doesn't do type conversion; need to get object first and cast to string.

PR #2109, opened 2019-12-12
Clean up loadGen and Relayer modules for connectivity tests

philipktlin:
Changes to clean up and integrate all pieces for connectivity test.


edge-modules/Relayer/Program.cs
```diff
@@ -43,6 +44,11 @@ static async Task Main(string[] args)
             {
                 Logger.LogError(ex, "Error occurred during Relayer.");
             }
+            finally
+            {
+                moduleClient?.CloseAsync();
+                moduleClient?.Dispose();
```
varunpuranik: Why do we need to dispose?

philipktlin: Just to clean up moduleClient as it implements IDisposable interface.

PR #2111, opened 2019-12-12
[k8s] Update Helm charts with new image names, new version.

darobs:


darobs:
> Do we need to update the template in CRD apiVersion: apiextensions.k8s.io/v1beta1? to beta3?

That is a good idea. We don’t need to, but it might make sense- would require code changes.

PR #2114, opened 2019-12-13
Fix flaky CI tests - Overloaded consumer group partitions

and-rewsmith:
Three CI tests are currently failing due to overloaded partitions per consumer group. These tests started failing right when we integrated the C# SDK's private drop bits, however from my investigation I couldn't figure out what exactly changed.   

Regardless, the current implementation makes this problem worse by reading from all event hub partitions. 

Since we know how the C# SDK resolves the event hub partition when sending messages, we can choose only the partitions that are relevant when reading. This PR uses their [implementation of partition resolution](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/service/src/Common/EventHubPartitionKeyResolver.cs#L13), which will continue to work even if they later change resolution logic.

I made this change to see if it would fix the issue without adding extra consumer groups, but it did not fix it and I had to create another. I also attempted to cut down on the time cost of these three tests, as they take worst case 250 seconds, but have discovered that these tests are extremely brittle so I will address this in another PR. 

If anyone has insight into what specifically is causing these tests to fail now, but not before, I'd love to hear your thoughts.


edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
```diff
@@ -24,22 +24,31 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test
...
         [TestPriority(401)]
         public async Task SendMessageTest()
         {
+            string device2ConnectionString = await SecretsHelper.GetSecretFromConfigKey(Device2ConnStrKey);
+            string deviceId2 = IotHubConnectionStringBuilder.Create(device2ConnectionString).DeviceId;
+
             IMessage message = MessageHelper.GenerateMessages(1)[0];
             DateTime startTime = DateTime.UtcNow.Subtract(ClockSkew);
-            ICloudProxy cloudProxy = await this.GetCloudProxyWithConnectionStringKey("device2ConnStrKey");
+            ICloudProxy cloudProxy = await this.GetCloudProxyFromConnectionStringKey("device2ConnStrKey");
```
philipktlin: you introduced constant, i guess you want to use here.
```suggestion
            ICloudProxy cloudProxy = await this.GetCloudProxyFromConnectionStringKey(Device2ConnStrKey);
```

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
```diff
@@ -49,43 +58,64 @@ public async Task SendMessageMultipleDevicesTest()
...
         [TestPriority(403)]
         public async Task SendMessageBatchTest()
         {
+            string device2ConnectionString = await SecretsHelper.GetSecretFromConfigKey(Device2ConnStrKey);
+            string deviceId2 = IotHubConnectionStringBuilder.Create(device2ConnectionString).DeviceId;
+
             IList<IMessage> messages = MessageHelper.GenerateMessages(4);
             DateTime startTime = DateTime.UtcNow.Subtract(ClockSkew);
-            ICloudProxy cloudProxy = await this.GetCloudProxyWithConnectionStringKey("device2ConnStrKey");
+            ICloudProxy cloudProxy = await this.GetCloudProxyFromConnectionStringKey("device2ConnStrKey");
```
philipktlin: Do you want to use constant here? more usages below.
```suggestion
            ICloudProxy cloudProxy = await this.GetCloudProxyFromConnectionStringKey(Device2ConnStrKey);
```

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/MessageHelper.cs
```diff
@@ -51,17 +51,20 @@ public static IList<IMessage> GenerateMessages(int count)
...
             {
-                EventData eventData = events.FirstOrDefault(
-                    m =>
-                        m.Properties.ContainsKey("id") &&
-                        m.Properties["id"] as string == message.Properties["id"]);
-                if (eventData == null || !message.Body.SequenceEqual(eventData.Body.Array))
+                foreach (IMessage message in sentMessagesByDevice[deviceId])
                 {
-                    return false;
+                    EventData eventData = receivedMessagesByDevice[deviceId].FirstOrDefault(
```
philipktlin: you can't assume deviceId must exist in receivedMessagesByDevice.Keys.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/EventHubReceiver.cs
```diff
@@ -4,52 +4,39 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test
...
     using System.Collections.Generic;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Common;
     using Microsoft.Azure.EventHubs;
 
     public class EventHubReceiver
     {
-        readonly EventHubClient eventHubClient;
+        const string EventHubConsumerGroup = "ci-tests";
+        readonly string eventHubConnectionString;
```
philipktlin: why remove EventHubClient and store eventHubConnectionString instead?

and-rewsmith: If we are going to retry the GetMessagesForDevice method, then we should start the process from the beginning, which means creating the EventHubClient again. Do you agree?

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/MessageHelper.cs
and-rewsmith: Before this method is called, I make sure that the receivedMessagesByDevice has all the deviceIds initialized. But I agree as we should decouple the logic.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/MessageHelper.cs
```diff
@@ -51,18 +51,26 @@ public static IList<IMessage> GenerateMessages(int count)
         /// Checks if the events have items corresponding to all messages, and
         /// returns false if it doesn't.
         /// </summary>
-        public static bool CompareMessagesAndEventData(IList<IMessage> messages, IList<EventData> events)
+        public static bool CompareMessagesAndEventData(IDictionary<string, IList<IMessage>> sentMessagesByDevice, IDictionary<string, List<EventData>> receivedMessagesByDevice)
```
and-rewsmith: I will rename

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
```diff
@@ -125,12 +155,12 @@ public async Task CanListenForDesiredPropertyUpdates()
         {
             var update = new TaskCompletionSource<IMessage>();
             var edgeHub = new Mock<IEdgeHub>();
-            string deviceConnectionStringKey = "device2ConnStrKey";
+            string deviceConnectionStringKey = Device2ConnStrKey;
```
philipktlin: nit: remove deviceConnectionStringKey

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
```diff
@@ -49,43 +58,64 @@ public async Task SendMessageMultipleDevicesTest()
...
             await cloudProxy.SendMessageBatchAsync(messages);
+
+            Dictionary<string, IList<IMessage>> sentMessagesByDevice = new Dictionary<string, IList<IMessage>>();
+            sentMessagesByDevice.Add(deviceId2, messages);
+
             bool disconnectResult = await cloudProxy.CloseAsync();
             Assert.True(disconnectResult);
 
-            await CheckMessageInEventHub(messages, startTime);
+            await CheckMessageInEventHub(sentMessagesByDevice, startTime);
```
philipktlin:
```suggestion
            await CheckMessageInEventHub(messagesSentByDevice, startTime);
```

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
and-rewsmith: I will remove here and where appropriate

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
```diff
@@ -216,28 +246,39 @@ public async Task TestHandleNonRecoverableExceptions(Type exceptionType)
...
+                    receivedMessagesByDevice[deviceId].AddRange(await eventHubReceiver.GetMessagesForDevice(deviceId, startTime));
+                }
+
+                messagesFound = MessageHelper.ValidateSentMessagesWereReceived(sentMessagesByDevice, receivedMessagesByDevice);
+
                 if (messagesFound)
                 {
                     break;
                 }
+            }
```
philipktlin: add delay

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
and-rewsmith: I tried to emphasize the difference by putting it first

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/EventHubReceiver.cs
```diff
@@ -4,52 +4,39 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test
...
-                    }
-                }
+            EventHubClient eventHubClient = EventHubClient.CreateFromConnectionString(this.eventHubConnectionString);
+            PartitionReceiver partitionReceiver = eventHubClient.CreateReceiver(
+                EventHubConsumerGroup,
+                EventHubPartitionKeyResolver.ResolveToPartition(deviceId, (await eventHubClient.GetRuntimeInformationAsync()).PartitionCount),
+                EventPosition.FromEnqueuedTime(startTime));
 
-                await partitionReceiver.CloseAsync();
+            IEnumerable<EventData> events = await partitionReceiver.ReceiveAsync(maxPerPartition, TimeSpan.FromSeconds(waitTimeSecs));
```
and-rewsmith: Add a filter to just get device id messages

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/CloudProxyTest.cs
```diff
@@ -216,29 +243,30 @@ public async Task TestHandleNonRecoverableExceptions(Type exceptionType)
...
         {
             string eventHubConnectionString = await SecretsHelper.GetSecretFromConfigKey("eventHubConnStrKey");
             var eventHubReceiver = new EventHubReceiver(eventHubConnectionString);
-            var cloudMessages = new List<EventData>();
-            bool messagesFound = false;
-            // Add retry mechanism to make sure all the messages sent reached Event Hub.
-            for (int i = 0; i < EventHubMessageReceivedRetry; i++)
+            var receivedMessagesByDevice = new Dictionary<string, List<EventData>>();
+
+            await Task.Delay(TimeSpan.FromSeconds(30));
```
philipktlin: since you changed to wait a fixed time once, please move this line at the top and add comment to explain why you do this.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/MessageHelper.cs
```diff
@@ -51,18 +51,26 @@ public static IList<IMessage> GenerateMessages(int count)
         /// Checks if the events have items corresponding to all messages, and
         /// returns false if it doesn't.
         /// </summary>
-        public static bool CompareMessagesAndEventData(IList<IMessage> messages, IList<EventData> events)
+        public static bool ValidateSentMessagesWereReceived(IDictionary<string, IList<IMessage>> sentMessagesByDevice, IDictionary<string, List<EventData>> receivedMessagesByDevice)
         {
-            foreach (IMessage message in messages)
+            foreach (string deviceId in sentMessagesByDevice.Keys)
```
philipktlin: I think you need to check number of keys in receivedMessagesByDevice should be greater than or equal to number of keys in sentMessagesByDevice. Similar check you can do before verifying each message, check receivedMessagesByDevice[deviceId].Count should be greater than or equal to sentMessagesByDevice[deviceId].Count.

btw receivedMessagesByDevice doesn't mean it stores received message only for that device.  It stores received messages only for that partition id derived by that device id.  do i understand correctly? if so, please rename it.

and-rewsmith: Isn't the check I'm doing already sufficient? The current logic makes sure sent messages is a subset of the received messages for each device.

You understand correctly. I will rename.

philipktlin:
Please try out the approach we discussed today and submit another PR if it works.

PR #2120, opened 2019-12-13
Install Trust Bundle in EdgeAgent and EdgeHub

varunpuranik:
Currently EdgeAgent and EdgeHub are able to communicate only with an IoTHub with server cert that is rooted in a well known cert. To allow them to communicate with an IoTHub with a private rooted cert, we need to install the Trust bundle that iotedge is configured with. 
This change does that. It has been tested to work with Linux. Need to test on Windows.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
```diff
@@ -161,6 +163,10 @@ public static async Task<int> MainAsync(IConfiguration configuration)
                         apiVersion = configuration.GetValue<string>(Constants.EdgeletApiVersionVariableName);
                         builder.RegisterModule(new AgentModule(maxRestartCount, intensiveCareTime, coolOffTimeUnitInSeconds, usePersistentStorage, storagePath, Option.Some(new Uri(workloadUri)), Option.Some(apiVersion), moduleId, Option.Some(moduleGenerationId), enableNonPersistentStorageBackup, storageBackupPath));
                         builder.RegisterModule(new EdgeletModule(iothubHostname, edgeDeviceHostName, deviceId, new Uri(managementUri), new Uri(workloadUri), apiVersion, dockerAuthConfig, upstreamProtocol, proxy, productInfo, closeOnIdleTimeout, idleTimeout));
+
+                        IList<X509Certificate2> trustBundle = (await CertificateHelper.GetTrustBundleFromEdgelet(new Uri(workloadUri), apiVersion, Constants.WorkloadApiVersion, moduleId, moduleGenerationId)).ToList();
```
dmolokanov: `InstallCertificates` already takes `IEnumerable<X509Certificate2>`  as arg. Do we still need to call `ToList()` here?

edge-util/src/Microsoft.Azure.Devices.Edge.Util/CertificateHelper.cs
```diff
@@ -207,13 +208,17 @@ public static bool ValidateCertExpiry(X509Certificate2 certificate, ILogger logg
...
-        public static void InstallCerts(StoreName name, StoreLocation location, IEnumerable<X509Certificate2> certs)
+        public static void InstallCertificates(IEnumerable<X509Certificate2> certificateChain, ILogger logger)
         {
-            List<X509Certificate2> certsList = Preconditions.CheckNotNull(certs, nameof(certs)).ToList();
-            using (var store = new X509Store(name, location))
+            X509Certificate2[] certs = Preconditions.CheckNotNull(certificateChain, nameof(certificateChain)).ToArray();
+
+            StoreName storeName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? StoreName.CertificateAuthority : StoreName.Root;
+
+            logger?.LogInformation($"Installing certificates {string.Join(",", certs.Select(c => $"[{c.Subject}:{c.GetExpirationDateString()}]"))} to {storeName}");
```
dmolokanov: Shouldn't it be a precondition check for logger in the beginnging of method instead of null check here?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
dmolokanov: Inside `InstallCertificates` method there is additional array allocation for the same list.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/CertificateHelper.cs
varunpuranik: The idea is that dependencies _may_ not have the logger, and we need not fail if logger is null. So this is typically the pattern we follow in helpers.

dmolokanov: Should we change it to Option to make our intention explicit?

varunpuranik: Okay, I will add the null check.. option will make using it a bit awkward

PR #2123, opened 2019-12-13
Fix Direct Method Sender for E2E tests

and-rewsmith:
When standardizing the logic for the direct method modules, I took out the old reporting mechanism which E2E used. I added it back in this PR.

The direct method sender now has two modes. If no analyzer url is passed in, it will assume the old E2E test logic.


edge-modules/DirectMethodSender/Settings.cs
```diff
@@ -25,15 +25,15 @@ public class Settings
                     configuration.GetValue<string>("TargetModuleId", "DirectMethodReceiver"),
                     configuration.GetValue<TransportType>("TransportType", TransportType.Amqp_Tcp_Only),
                     configuration.GetValue<TimeSpan>("DirectMethodDelay", TimeSpan.FromSeconds(5)),
-                    configuration.GetValue<Uri>("AnalyzerUrl", new Uri("http://analyzer:15000")));
+                    configuration.GetValue<string>("AnalyzerUrl", string.Empty));
```
and-rewsmith: I didn't want to change this away from Uri type, but couldn't think of a way around this. Uri with empty string throws an exception.

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -29,12 +29,22 @@ public static async Task<int> MainAsync()
...
                     Logger);
 
-                Uri analyzerUrl = Settings.Current.AnalyzerUrl;
-                AnalyzerClient analyzerClient = new AnalyzerClient { BaseUrl = analyzerUrl.AbsoluteUri };
-
                 (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromSeconds(5), Logger);
 
-                await CallDirectMethod(moduleClient, analyzerClient, Settings.Current.DirectMethodDelay, cts);
+                string analyzerUrl = Settings.Current.AnalyzerUrl;
+                if (analyzerUrl != string.Empty)
```
philipktlin:
```suggestion
                if (!string.IsNullOrWhiteSpace(analyzerUrl))
```

philipktlin: Do you need try-catch for this if block

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -50,9 +60,9 @@ public static async Task<int> MainAsync()
             return 0;
```
philipktlin: Add finally to dispose module client.

and-rewsmith:
I smoked this with modules for long haul using latest build from master

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -29,12 +29,22 @@ public static async Task<int> MainAsync()
...
+                if (analyzerUrl != string.Empty)
+                {
+                    Uri analyzerUri = new Uri(analyzerUrl);
+                    AnalyzerClient analyzerClient = new AnalyzerClient { BaseUrl = analyzerUri.AbsoluteUri };
+                    Action<MethodResponse> reportResult = async (response) => await ReportStatus(Settings.Current.TargetModuleId, response, analyzerClient);
+                    await StartDirectMethdTests(moduleClient, reportResult, Settings.Current.DirectMethodDelay, cts);
+                }
+                else
+                {
+                    Action<MethodResponse> reportResult = async (response) => await moduleClient.SendEventAsync("AnyOutput", new Message(Encoding.UTF8.GetBytes("Direct Method Call succeeded.")));
```
philipktlin: initialize moduleClient here

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -29,12 +29,22 @@ public static async Task<int> MainAsync()
...
+                    await StartDirectMethdTests(moduleClient, reportResult, Settings.Current.DirectMethodDelay, cts);
+                }
+
                 await moduleClient.CloseAsync();
```
philipktlin: move this line inside else block

and-rewsmith: module client is needed in both cases

and-rewsmith: It is passed into the StartDirectMethodTests()

edge-modules/DirectMethodSender/Settings.cs
philipktlin: (Option.Maybe(configuration.GetValue<string>("AnalyzerUrl"))), then use Map method to convert it to Option<Uri>.

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -29,12 +29,22 @@ public static async Task<int> MainAsync()
                     ModuleUtil.DefaultTransientRetryStrategy,
                     Logger);
 
-                Uri analyzerUrl = Settings.Current.AnalyzerUrl;
-                AnalyzerClient analyzerClient = new AnalyzerClient { BaseUrl = analyzerUrl.AbsoluteUri };
-
                 (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromSeconds(5), Logger);
```
and-rewsmith: I should move this higher

edge-modules/DirectMethodSender/Program.cs
philipktlin: my bad. module client used in both case.

edge-modules/DirectMethodSender/Program.cs
varunpuranik: but both cases are inside the try, right?

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -21,38 +21,55 @@ public static async Task<int> MainAsync()
...
+                    analyzerUrl.ForEach(async (uri) =>
+                    {
+                        AnalyzerClient analyzerClient = new AnalyzerClient { BaseUrl = uri.AbsoluteUri };
+                        Action<MethodResponse> reportResult = async (response) => await ReportStatus(Settings.Current.TargetModuleId, response, analyzerClient);
+                        await StartDirectMethdTests(moduleClient, reportResult, Settings.Current.DirectMethodDelay, cts);
+                    });
+                }
+                else
+                {
+                    Action<MethodResponse> reportResult = async (response) => await moduleClient.SendEventAsync("AnyOutput", new Message(Encoding.UTF8.GetBytes("Direct Method call succeeded.")));
```
varunpuranik: How will this message get processed? Is this existing logic?

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -21,38 +21,55 @@ public static async Task<int> MainAsync()
...
                     ModuleUtil.DefaultTimeoutErrorDetectionStrategy,
                     ModuleUtil.DefaultTransientRetryStrategy,
                     Logger);
 
-                Uri analyzerUrl = Settings.Current.AnalyzerUrl;
-                AnalyzerClient analyzerClient = new AnalyzerClient { BaseUrl = analyzerUrl.AbsoluteUri };
-
-                (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromSeconds(5), Logger);
+                Option<Uri> analyzerUrl = Settings.Current.AnalyzerUrl;
+                if (analyzerUrl.HasValue)
```
varunpuranik: Use the new Option.ForEach overload that takes in 2 actions.

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -21,38 +21,55 @@ public static async Task<int> MainAsync()
...
+                {
+                    analyzerUrl.ForEach(async (uri) =>
+                    {
+                        AnalyzerClient analyzerClient = new AnalyzerClient { BaseUrl = uri.AbsoluteUri };
+                        Action<MethodResponse> reportResult = async (response) => await ReportStatus(Settings.Current.TargetModuleId, response, analyzerClient);
```
varunpuranik: Make this a Func that returns a task

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -79,7 +96,7 @@ public static async Task<int> MainAsync()
                         Logger.LogError(statusMessage);
                     }
 
-                    await ReportStatus(targetModuleId, response, analyzerClient);
+                    reportResult.Invoke(response);
```
varunpuranik: await reportResult(response);

edge-modules/DirectMethodSender/Program.cs
and-rewsmith: Yes, existing logic in e2e

edge-modules/DirectMethodSender/Program.cs
and-rewsmith: I tried using this, but it is not compatible with Func.

edge-modules/DirectMethodSender/Settings.cs
```diff
@@ -25,22 +25,22 @@ public class Settings
                     configuration.GetValue<string>("TargetModuleId", "DirectMethodReceiver"),
                     configuration.GetValue<TransportType>("TransportType", TransportType.Amqp_Tcp_Only),
                     configuration.GetValue<TimeSpan>("DirectMethodDelay", TimeSpan.FromSeconds(5)),
-                    configuration.GetValue<Uri>("AnalyzerUrl", new Uri("http://analyzer:15000")));
+                    Option.Maybe(configuration.GetValue<Uri>("AnalyzerUrl")));
```
philipktlin: is it safer to check "AnalyzerUrl" is not empty or white spaces?

edge-modules/load-gen/Settings.cs
```diff
@@ -100,10 +100,14 @@ public override string ToString()
                 { nameof(this.TestStartDelay), this.TestStartDelay.ToString() },
                 { nameof(this.TestDuration), this.TestDuration.ToString() },
                 { nameof(this.TrackingId), this.TrackingId },
-                { nameof(this.TransportType), Enum.GetName(typeof(TransportType), this.TransportType) },
-                { nameof(this.TestResultCoordinatorUrl), this.TestResultCoordinatorUrl.ToString() },
+                { nameof(this.TransportType), Enum.GetName(typeof(TransportType), this.TransportType) }
             };
 
+            this.TestResultCoordinatorUrl.ForEach((url) =>
```
philipktlin: Option class overrided tostring method and it is good to use.  But I would like to make it consistent to use Uri as return type for TestResultCoordinatorUrl.

edge-modules/DirectMethodSender/Settings.cs
and-rewsmith: It will throw an exception if it tries to make a Uri from an empty string or null. This seems fine to me as it is not a valid value for Uri.

edge-modules/load-gen/Settings.cs
and-rewsmith: I can make consistent in another PR

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -21,38 +21,53 @@ public static async Task<int> MainAsync()
...
                     ModuleUtil.DefaultTimeoutErrorDetectionStrategy,
                     ModuleUtil.DefaultTransientRetryStrategy,
                     Logger);
 
-                Uri analyzerUrl = Settings.Current.AnalyzerUrl;
-                AnalyzerClient analyzerClient = new AnalyzerClient { BaseUrl = analyzerUrl.AbsoluteUri };
-
-                (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromSeconds(5), Logger);
+                Option<Uri> analyzerUrl = Settings.Current.AnalyzerUrl;
+                analyzerUrl.ForEach(
```
philipktlin: you need to use ForEachAsync

PR #2126, opened 2019-12-13
Refactor Storage for TestResultCoordinator

dylanbronson:
Moving away from static and introducing DI to TRC.

### DIFF
--- a/test/modules/TestResultCoordinator/Controllers/TestOperationResultController.cs
+++ b/test/modules/TestResultCoordinator/Controllers/TestOperationResultController.cs
@@ -6,11 +6,16 @@ namespace TestResultCoordinator.Controllers
     using System.Net;
     using System.Threading.Tasks;
     using Microsoft.AspNetCore.Mvc;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Extensions.Logging;
+    using TestOperationResult = TestResultCoordinator.TestOperationResult;
 
     [Route("api/[controller]")]
     [ApiController]
     public class TestOperationResultController : Controller
     {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger(nameof(TestOperationResultController));
+
         // POST api/TestOperationResult
         [HttpPost]
         public async Task<StatusCodeResult> PostAsync(TestOperationResult result)
@@ -18,6 +23,7 @@ namespace TestResultCoordinator.Controllers
             try
             {
                 bool success = await TestOperationResultStorage.AddResultAsync(result);
+                Logger.LogDebug($"Received test result: {result.Source}, {result.Type}, {success}");
                 return success ? this.StatusCode((int)HttpStatusCode.NoContent) : this.StatusCode((int)HttpStatusCode.BadRequest);
             }
             catch (Exception)
--- a/e2e_deployment_files/connectivity_deployment.template.json
+++ b/e2e_deployment_files/connectivity_deployment.template.json
@@ -45,7 +45,7 @@
               "messageFrequency": {
                 "value": "<LoadGen.MessageFrequency>"
               },
-              "testRunDuration": {
+              "testDuration": {
                 "value": "<TestDuration>"
               },
               "testStartDelay": {
@@ -78,7 +78,7 @@
               "messageFrequency": {
                 "value": "<LoadGen.MessageFrequency>"
               },
-              "testRunDuration": {
+              "testDuration": {
                 "value": "<TestDuration>"
               },
               "testStartDelay": {
@@ -114,6 +114,9 @@
               "inputName": {
                 "value": "input1"
               },
+              "outputName": {
+                "value": "output1"
+              },
               "testResultCoordinatorUrl": {
                 "value": "http://testResultCoordinator:5001"
               }
@@ -135,6 +138,9 @@
               "inputName": {
                 "value": "input2"
               },
+              "outputName": {
+                "value": "output2"
+              },
               "testResultCoordinatorUrl": {
                 "value": "http://testResultCoordinator:5001"
               }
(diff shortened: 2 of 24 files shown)


test/modules/TestResultCoordinator/Controllers/TestOperationResultController.cs
```diff
@@ -11,13 +10,20 @@ namespace TestResultCoordinator.Controllers
     [ApiController]
     public class TestOperationResultController : Controller
     {
+        private readonly ITestOperationResultStorage testOperationResultStorage;
```
varunpuranik: Nit: avoid private keyword for fields, it is the default

test/modules/TestResultCoordinator/Controllers/TestOperationResultController.cs
```diff
@@ -11,13 +10,20 @@ namespace TestResultCoordinator.Controllers
...
+
+        public TestOperationResultController(ITestOperationResultStorage testOperationResultStorage)
+        {
+            this.testOperationResultStorage = testOperationResultStorage;
```
varunpuranik: check null

test/modules/TestResultCoordinator/ITestOperationResultStorage.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator
+{
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Storage;
+    public interface ITestOperationResultStorage
+    {
+        ISequentialStore<TestOperationResult> GetStoreFromSource(string source);
+
+        Task<bool> AddResultAsync(TestOperationResult testOperationResult);
```
varunpuranik: How do you retrieve the results?

test/modules/TestResultCoordinator/ITestOperationResultStorage.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator
+{
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Storage;
+    public interface ITestOperationResultStorage
+    {
+        ISequentialStore<TestOperationResult> GetStoreFromSource(string source);
```
varunpuranik: What is this used for?

test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
```diff
@@ -19,11 +19,14 @@ class PartitionReceiveHandler : IPartitionReceiveHandler
...
         readonly string trackingId;
+        readonly ITestOperationResultStorage storage;
 
-        public PartitionReceiveHandler(string trackingId, string deviceId)
+        public PartitionReceiveHandler(string trackingId, string deviceId, ITestOperationResultStorage storage)
         {
             this.trackingId = Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
             this.deviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
+            this.storage = Preconditions.CheckNotNull(storage);
+            Logger.LogInformation($"registration worked: {storage.GetStoreFromSource("relayer1.send")}");
```
varunpuranik: ? You are getting the store only to log?

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -12,12 +12,23 @@ namespace TestResultCoordinator
...
 
-    class TestOperationResultStorage
+    public class TestOperationResultStorage : ITestOperationResultStorage
     {
         static readonly ILogger Logger = ModuleUtil.CreateLogger("TestResultCoordinator");
-        static Dictionary<string, ISequentialStore<TestOperationResult>> resultStores;
+        private Dictionary<string, ISequentialStore<TestOperationResult>> ResultStores { get; set; }
 
-        public static async Task InitAsync(string storagePath, ISystemEnvironment systemEnvironment, bool optimizeForPerformance, List<string> resultSources)
+        public static async Task<TestOperationResultStorage> CreateStorage(string storagePath, ISystemEnvironment systemEnvironment, bool optimizeForPerformance, List<string> resultSources)
```
varunpuranik: Nit: suggest renaming to Create

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
```diff
@@ -12,12 +12,23 @@ namespace TestResultCoordinator
...
+        {
+            Dictionary<string, ISequentialStore<TestOperationResult>> resultStores = await TestOperationResultStorage.InitAsync(storagePath, systemEnvironment, optimizeForPerformance, resultSources);
```
varunpuranik: Remove InitAsync. This is essentially your Init now.

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
varunpuranik: We should pass in the DbStoreProvider... that way we can use RocksDb / InMemoryStore depending on config.

test/modules/TestResultCoordinator/ITestOperationResultStorage.cs
dylanbronson: We use GetStoreFromSource with ReportMetadata, which contains the following info about the report: ExpectedSource, ActualSource, TestReportType, and TestOperationResultType. 

For each ReportType we have, we will have a new testResultReportGenerator that will generate a report for that Type.

GetStoreFromSource is used in TestReportGeneratorFactory to create TestResultReportGenerator's (currently only CountingReportGenerator)

test/modules/TestResultCoordinator/TestOperationResultStorage.cs
dylanbronson: Why would that be something we want configured in our tests? I can't think of why we wouldn't want to use persistent storage?

PR #2127, opened 2019-12-14
Fix test result coordinator

philipktlin:
Use task.Run to wrap task delay to ensure running on different thread; therefore web host can start processing requests and save result to store.

### DIFF
--- a/test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
+++ b/test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
@@ -30,6 +30,8 @@ namespace TestResultCoordinator
 
         public async Task ProcessEventsAsync(IEnumerable<EventData> events)
         {
+            Logger.LogInformation("Processing events from event hub.");
+
             if (events != null)
             {
                 foreach (EventData eventData in events)
@@ -52,17 +54,18 @@ namespace TestResultCoordinator
                         {
                             if (long.TryParse(sequenceNumberFromEvent.ToString(), out long sequenceNumber))
                             {
+                                Logger.LogInformation($"Recieved event hub event: tracking Id={(string)trackingIdFromEvent}, deviceId={(string)deviceIdFromEvent}, moduleId={(string)moduleIdFromEvent}");
                                 DateTime enqueuedtime = GetEnqueuedTime(deviceIdFromEvent.ToString(), moduleIdFromEvent.ToString(), eventData);
 
                                 // TODO: remove hardcoded eventHub string in next line
-                                TestOperationResult result = new TestOperationResult(
-                                    moduleIdFromEvent.ToString() + ".eventHub",
+                                var result = new TestOperationResult(
+                                    (string)moduleIdFromEvent + ".eventHub",
                                     "Messages",
                                     ModuleUtil.FormatTestResultValue(
                                         (string)trackingIdFromEvent,
                                         (string)batchIdFromEvent,
                                         (string)sequenceNumberFromEvent),
-                                    DateTime.UtcNow);
+                                    enqueuedtime);
                                 await TestOperationResultStorage.AddResultAsync(result);
                             }
                             else
--- a/test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
+++ b/test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
@@ -40,6 +40,8 @@ namespace TestResultCoordinator
                     eventData.SystemProperties.TryGetValue(DeviceIdPropertyName, out object deviceIdFromEvent);
                     eventData.SystemProperties.TryGetValue(ModuleIdPropertyName, out object moduleIdFromEvent);
 
+                    Logger.LogInformation($"Received event from Event Hub: trackingId={(string)trackingIdFromEvent}, deviceId={(string)deviceIdFromEvent}, moduleId={(string)moduleIdFromEvent}");
+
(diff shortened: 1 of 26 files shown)


test/modules/TestResultCoordinator/service/TestResultReportingService.cs
```diff
@@ -0,0 +1,90 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util.AzureLogAnalytics;
+    using Microsoft.Extensions.Hosting;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+    using TestResultCoordinator.Report;
+
+    // Reference from https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-3.1&tabs=visual-studio
```
ancaantochi: please add comments that the background task with Timer is from this reference

test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
```diff
@@ -47,6 +51,8 @@ public async Task ProcessEventsAsync(IEnumerable<EventData> events)
                         eventData.Properties.TryGetValue(TestConstants.Message.SequenceNumberPropertyName, out object sequenceNumberFromEvent);
                         eventData.Properties.TryGetValue(TestConstants.Message.BatchIdPropertyName, out object batchIdFromEvent);
 
+                        Logger.LogInformation($"Recieved event from Event Hub: sequenceNumber={(string)sequenceNumberFromEvent}, batchId={(string)batchIdFromEvent}");
```
dylanbronson: nit: minor typo - recieved -> received

test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
```diff
@@ -55,14 +61,14 @@ public async Task ProcessEventsAsync(IEnumerable<EventData> events)
                                 DateTime enqueuedtime = GetEnqueuedTime(deviceIdFromEvent.ToString(), moduleIdFromEvent.ToString(), eventData);
 
                                 // TODO: remove hardcoded eventHub string in next line
-                                TestOperationResult result = new TestOperationResult(
-                                    moduleIdFromEvent.ToString() + ".eventHub",
+                                var result = new TestOperationResult(
```
dylanbronson: What's the benefit of changing this to var over TestOperationResult if we know it is going to be a TestOperationResult and won't change?

test/modules/TestResultCoordinator/service/TestResultEventReceivingService.cs
```diff
@@ -0,0 +1,45 @@
...
+    using Microsoft.Azure.Devices.Common;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Azure.EventHubs;
+    using Microsoft.Extensions.Hosting;
+    using Microsoft.Extensions.Logging;
+
+    class TestResultEventReceivingService : BackgroundService
+    {
+        readonly ILogger logger = ModuleUtil.CreateLogger(nameof(TestResultEventReceivingService));
```
dylanbronson: very nit: all our other loggers are capitalized - we should stay consistent

test/modules/TestResultCoordinator/service/TestResultEventReceivingService.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            await Task.CompletedTask;
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        {
+            this.logger.LogInformation("Test Result Event Receiving Service running.");
+
+            DateTime eventEnqueuedFrom = DateTime.UtcNow;
+            var builder = new EventHubsConnectionStringBuilder(Settings.Current.EventHubConnectionString);
```
dylanbronson: Why make this a var instead of EventHubsConnectionStringBuilder type?

test/modules/TestResultCoordinator/service/TestResultEventReceivingService.cs
```diff
@@ -0,0 +1,45 @@
...
+            this.logger.LogInformation("Test Result Event Receiving Service is stopping.");
+
+            await Task.CompletedTask;
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        {
+            this.logger.LogInformation("Test Result Event Receiving Service running.");
+
+            DateTime eventEnqueuedFrom = DateTime.UtcNow;
```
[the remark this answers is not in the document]
philipktlin: let's do this later; try to minimize the changes in this PR.

test/modules/TestResultCoordinator/service/TestResultEventReceivingService.cs
philipktlin: it is small case here due to it is class variable, not static.

test/modules/TestResultCoordinator/service/TestResultEventReceivingService.cs
philipktlin: my general rule is to use var if you can determine the type from right side.

test/modules/TestResultCoordinator/PartitionReceiverHandler.cs
philipktlin: just for readability to shorten the line. :P

PR #2140, opened 2019-12-16
Metrics Upload Retries

lfitchett:
This pr adds retry logic to the metrics upload. 

It also moves the metrics de-duplication logic to it's own class and makes a util folder to store it and the prometheous metrics parser.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsScraper.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics
     using System.Text.RegularExpressions;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Util;
```
rafilho: >using Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Util; [](start = 4, length = 58)

still needed?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/storage/MetricsFileStorage.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            return Directory.GetFiles(this.directory)
+                .SelectMany(filename =>
+                {
+                    Metric[] fileMetrics;
+                    try
+                    {
+                        string rawMetrics = File.ReadAllText(filename);
+                        fileMetrics = JsonConvert.DeserializeObject<Metric[]>(rawMetrics) ?? new Metric[0];
+                        File.Delete(filename);
```
CindyXing: Why we are deleting the file here?

lfitchett: That's an old commit, not the pr changes.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsScraper.cs
lfitchett: yes, for PrometheusMetricsParser on line 48

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/IoTHubMetricsUpload.cs
```diff
@@ -30,8 +30,18 @@ public async Task PublishAsync(IEnumerable<Metric> metrics, CancellationToken ca
             if (data.Length > 0)
             {
                 Message message = this.BuildMessage(data);
-                await this.edgeAgentConnection.SendEventAsync(message);
+
+                try
+                {
+                    await this.edgeAgentConnection.SendEventAsync(message);
+                }
+                catch (Exception ex) when (ex.HasTimeoutException())
```
rafilho: >catch (Exception ex) when (ex.HasTimeoutException()) [](start = 16, length = 52)

This exception will be pretty much swallowed. Should we at least log it?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/MetricsFileWriter.cs
```diff
@@ -28,6 +28,8 @@ public async Task PublishAsync(IEnumerable<Metric> metrics, CancellationToken ca
                     await file.WriteAsync(buffer, 0, buffer.Length, cancellationToken);
```
rafilho: >WriteAsync [](start = 31, length = 10)

Do we need to handle/log potential exceptions when writing to the file, in a similar way to what you did on the IoTHubMetricsUpload implementation?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/IoTHubMetricsUpload.cs
lfitchett: The timeout exception is swallowed, b/c a timeout exception means that there was no internet connection and we should retry. In this case, the worker will log that it is starting to retry and the backoff timeout. It's an unfortunate case of control flow exceptions stemming from our sdk dependency.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/MetricsFileWriter.cs
lfitchett: Any exceptions will propagate back to the worker, which will log them. The true/false is just for retry, not unexpected exceptions. The exceptions writing to a file will throw (probably permission denied) should not be retried.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/IoTHubMetricsUpload.cs
lfitchett: All other exceptions will propagate back to the worker and be logged without retry.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/MetricsFileWriter.cs
lfitchett: Although, I just realized the metrics are preserved in the case of an unexpected exception. I will change MetricsWorker so they are deleted instead, so if it continually fails to upload and can't retry it doesn't fill the disk.

rafilho: Got it. In this case, I would consider implementing a retry strategy for the MetricsPublisher, encapsulating the decision about which exceptions to retry or not. It would be a bit cleaner and also easier to extend it later (e.g. for additional exceptions). 

---
In reply to: [358507791](https://github.com/Azure/iotedge/pull/2140#discussion_r358507791) [](ancestors = 358507791)

lfitchett: That's what the bool does. The publisher decides whether it should retry or not. True for success, false for retry, any exception for un-retryable failure.

rafilho: Sure. I understand. I was just thinking about encapsulating that on the detectionStrategy. Something similar to https://github.com/Azure/iotedge/blob/3eafe054159320ef61547c8fbadf812b767d7da1/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/statemachine/EndpointExecutorFsm.cs#L535. But to be honest, take it with a grain of salt, since I am not fully familiar with the coding guideline yet.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/IoTHubMetricsUpload.cs
varunpuranik: emm, I agree, we should log it

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/MetricsDeDuplication.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Util
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+
+    public static class MetricsDeDuplication
+    {
+        public static IEnumerable<Metric> RemoveDuplicateMetrics(IEnumerable<Metric> metrics)
```
varunpuranik: Where is Metric defined? Is it generic enough to put it in Util?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Test/UtilTest.cs
```diff
@@ -0,0 +1,139 @@
...
+    using Moq;
+    using Newtonsoft.Json;
+    using Xunit;
+
+    public class UtilTest
+    {
+        Random rand = new Random();
+
+        [Fact]
+        public void TestRemoveDuplicateMetrics()
```
varunpuranik: UT should be in the same place as the code.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/MetricsDeDuplication.cs
lfitchett: edge-agent\src\Microsoft.Azure.Devices.Edge.Agent.Diagnostics\Metric.cs

I think it makes sense to leave it in Diagnostics for now. It's only used internally

kkdawkins: I don't like the name here - You are esentially filtering one enumerable into another enumerable. You could make this an extension method called GetDistinct(). This would allow you to write code like: `var foo = metrics.GetDistinct();`

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/MetricsDeDuplication.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            Dictionary<int, (Metric metric, bool)> previousValues = new Dictionary<int, (Metric, bool)>();
+
+            foreach (Metric newMetric in metrics)
```
kkdawkins: I think that there is a much simpler way to write this? Why not use a set as a filter? Lookat how c# implements distinct https://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,4ab583c7d8e84d6d

kkdawkins: Your set would be similar but you could use a "key" instead of the raw metric (since the metric could be the same but with a different value)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/MetricsDeDuplication.cs
lfitchett: renamed to RemoveDuplicateMetrics and made extension function

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/MetricsDeDuplication.cs
lfitchett: Added comments to clarify behavior

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/IoTHubMetricsUpload.cs
lfitchett: added debug log

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsWorker.cs
```diff
@@ -53,38 +57,67 @@ internal async Task Scrape(CancellationToken cancellationToken)
...
+                }
+                catch (Exception ex)
+                {
+                    // If unexpected error publishing metrics, delete stored ones to prevent storage buildup.
+                    Log.LogError($"Unexpected error publishing metrics. Deleting stored metrics.");
+                    await this.storage.RemoveAllReturnedMetricsAsync();
+                    throw ex;
+                }
+
+                if (uploaded)
```
kkdawkins: Why not just move all of this into the try? Then you wouldn't need the bool uploaded declared out on line 74? Doesn't seem like it buys you anything having this code out here.

kkdawkins: This is for lines 87-97

PR #2141, opened 2019-12-16
Give docker stats default values

lfitchett:
Docker sometimes returns null for some stats values, especially on startup. there are now default values to prevent null ref exceptions.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
```diff
@@ -8,6 +8,19 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker.Models
...
+        public DockerStats(string name, Dictionary<string, DiskIO[]> block_io_stats, DockerCpuStats cpu_stats, MemoryStats memory_stats, Dictionary<string, NetworkInfo> networks, int? num_processes, PidsStats pids_stats, DateTime? read)
+        {
+            this.Name = name ?? "error_name";
+            this.BlockIoStats = block_io_stats ?? new Dictionary<string, DiskIO[]>();
+            this.CpuStats = cpu_stats ?? new DockerCpuStats(null, null, null);
+            this.MemoryStats = memory_stats ?? new MemoryStats(null, null, null);
+            this.Networks = networks ?? new Dictionary<string, NetworkInfo>();
+            this.NumProcesses = num_processes ?? 0;
+            this.PidsStats = pids_stats ?? new PidsStats(null);
+            this.Read = read ?? DateTime.MinValue;
```
and-rewsmith: What are your thoughts on finding a way to use defaults that don't convey meaning? I can think of solutions to this, but they seem hacky. We should discuss.

lfitchett: Agree. Will wrap in options

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -161,33 +161,41 @@ void SetModuleStats(SystemResources systemResources)
...
-                    this.diskWrite.Set(module.BlockIoStats.Sum(io => io.Value.Where(d => d.Op == "Write").Sum(d => d.Value)), tags);
-                }
+                    this.diskRead.Set(disk.Sum(io => io.Value.Where(d => d.Op.Exists(op => op == "Read")).Sum(d => d.Value.OrDefault())), tags);
+                    this.diskWrite.Set(disk.Sum(io => io.Value.Where(d => d.Op.Exists(op => op == "Write")).Sum(d => d.Value.OrDefault())), tags);
+                });
             }
         }
 
-        Option<double> GetCpuUsage(DockerStats module)
+        Option<double> GetCpuUsage(DockerStats module, string name)
```
CindyXing: What's the usage pattern for this API? Not clear what's the "name" parameter is for.

lfitchett: The module name has a bit of processing done on it. Right now it just removes '/' from the start, but depending on GDPR requirements it may be hashed as well. This prevents duplication of this logic.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
```diff
@@ -4,98 +4,170 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker.Models
...
+            this.BlockIoStats = Option.Maybe(block_io_stats);
+            this.CpuStats = Option.Maybe(cpu_stats);
+            this.MemoryStats = Option.Maybe(memory_stats);
+            this.Networks = Option.Maybe(networks);
+            this.NumProcesses = Option.Maybe(num_processes);
+            this.PidsStats = Option.Maybe(pids_stats);
+            this.Read = Option.Maybe(read);
+
+            // Check for null entries in dictionarys
+            if (this.BlockIoStats.Exists(n => n.Any(o => o.Key == null || o.Value == null)))
```
philipktlin: o.Key cannot be null.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
```diff
@@ -4,98 +4,170 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker.Models
...
+            {
+                this.BlockIoStats = Option.None<Dictionary<string, DiskIO[]>>();
```
philipktlin: why not just filter out any value is null, but instead set to Option.None?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
```diff
@@ -4,98 +4,170 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker.Models
...
+    using System.Linq;
+    using Microsoft.Azure.Devices.Edge.Util;
     using Newtonsoft.Json;
 
     public class DockerStats
     {
+        [JsonConstructor]
+        public DockerStats(string name, Dictionary<string, DiskIO[]> block_io_stats, DockerCpuStats cpu_stats, MemoryStats memory_stats, Dictionary<string, NetworkInfo> networks, int? num_processes, PidsStats pids_stats, DateTime? read)
+        {
+            this.Name = Option.Maybe(name);
```
philipktlin: what is this name used for? should it be required?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
```diff
@@ -4,98 +4,170 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker.Models
...
-        public DateTime Read { get; set; }
+        public Option<DateTime> Read { get; }
     }
 
     public class DiskIO
     {
+        [JsonConstructor]
+        public DiskIO(long? major, long? minor, string op, long? value)
+        {
+            this.Major = Option.Maybe(major);
```
philipktlin: If all fields are optional, but just create Option.None<DiskIO>().  Some fields should be required, right?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
philipktlin: DockerStats also have a name property; are both module name?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
lfitchett: switched to filter out instead

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
lfitchett: It is required by the stats that use this, but this class doesn't and shouldn't know that. All this does is parse the docker api result. The class that uses this handles the empty name case

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
philipktlin: Please put a comment to the class, mentioned this class is just used for serializing docker stats result.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
lfitchett: this is the name from docker stats, but with a bit of processing done to it. I don;t want to repeat the processing

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
lfitchett: added

PR #2143, opened 2019-12-17
Disabling flakey test - RefreshCacheWithRefreshRequestTest

dylanbronson:
Disabling RefreshCacheWithRefreshRequestTest. It breaks builds pretty often, but is not important enough to be worked on right at this moment. When we get the chance we will fix it. 

Bug is filed: 5928143


edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/DeviceScopeIdentitiesCacheTest.cs
```diff
@@ -135,7 +135,7 @@ public async Task RefreshCacheTest()
             Assert.Contains("d2/m4", removedIdentities);
         }
 
-        [Fact]
+        [Fact(Skip ="Consistently flakey test - bug logged: 5928143")]
```
varunpuranik: Don't log bug Id in here.

and-rewsmith:
How often does this fail?

dylanbronson:
@and-rewsmith this fails pretty often - so dotnet checkins fail often as well, but re-run usually succeeds

PR #2145, opened 2019-12-17
CloudProxyTest - Don't read duplicate messages

and-rewsmith:
We can pass the time of the last received message to a PartitionReceiver in order to stop reading duplicate messages and simplify the retry logic around event hub.


philipktlin:
Please test to verify if it makes test stable.

and-rewsmith:
This is unstable. This implies very weird behavior with ReceiveAsync().

PR #2147, opened 2019-12-17
Fix flaky BackupAndRestoreTest

and-rewsmith:
There is an Assert before a needed wait that will cause an amqp error if Assert fails (execution immediately jumps into finally without waiting).

### DIFF
--- a/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/BackupAndRestoreTest.cs
+++ b/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/BackupAndRestoreTest.cs
@@ -177,10 +177,10 @@ namespace Microsoft.Azure.Devices.Edge.Hub.E2E.Test
                 // Send more messages after the receiver is registered.
                 Task<int> task2 = sender.SendMessagesByCountAsync("output1", beforeBackupMessageCount, afterBackupMessageCount, TimeSpan.FromMinutes(2));
                 sentMessagesCount = await task2;
-                Assert.Equal(afterBackupMessageCount, sentMessagesCount);
 
                 waitTime = waitTimeComputer(expectedMessageCountAfterRestore);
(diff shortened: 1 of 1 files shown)


edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/BackupAndRestoreTest.cs
```diff
@@ -177,7 +177,6 @@ async Task BackupAndRestoreCorruptBackupDataTest(ITransportSettings[] transportS
                 // Send more messages after the receiver is registered.
                 Task<int> task2 = sender.SendMessagesByCountAsync("output1", beforeBackupMessageCount, afterBackupMessageCount, TimeSpan.FromMinutes(2));
                 sentMessagesCount = await task2;
-                Assert.Equal(afterBackupMessageCount, sentMessagesCount);
```
shantanu1singh: Did this assertion fail because of a timeout while sending the messages? If so, I would rather increase the timeout instead of moving this assertion to after the wait before asserting received messages.

This is so that we can ensure that all the messages were sent before we even attempt to check the receipt of messages later.

and-rewsmith: I thought this was an error caused by messages in flight in the EdgeHubFixture. This is not the case and I am investigating more.

and-rewsmith:
Totally misunderstood the stack trace when investigating last night, so closing and will investigate more.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/BackupAndRestoreTest.cs
and-rewsmith: I've updated the description with a link to the failing CI run and tested version of BackupAndRestore.cs.

PR #2148, opened 2019-12-17
Update edgehub-proxy

dmolokanov:
Update edgehub-proxy to rust edition 2018.
Make edgelet-hsm as dev-dependency for edgelet-http.


edgelet/edgelet-http/Cargo.toml
```diff
@@ -50,6 +49,7 @@ winapi = { version = "0.3.5", features = ["winsock2"] }
 lazy_static = "1.0"
 tempfile = "3"
 tempdir = "0.3.7"
+edgelet-hsm = { path = "../edgelet-hsm" }
 
 edgelet-test-utils = { path = "../edgelet-test-utils" }
```
myagley: This is a huge nit, but can you please group the `edgelet-hsm` dependency with `edgelet-test-utils`? Something like:

```toml
lazy_static = "1.0"
tempfile = "3"
tempdir = "0.3.7"

edgelet-hsm = { path = "../edgelet-hsm" }
edgelet-test-utils = { path = "../edgelet-test-utils" }
```

dmolokanov: Done

PR #2158, opened 2019-12-18
Add and filter by ms_telemetry tag in edgeAgent

lfitchett:
This adds ms_telemetry=true to the edgeAgent metrics. This tag serves two purposes: 
* It tells edgeAgent which metrics to upload
* It acts as documentation for customers.
    * WIll need to contact legal team to determine if this is sufficient for GDPR.

This also removes tags from metrics to reduce bandwidth. Device Id and Hub Name are already known in iothub, so damon's code adds them back in. 

It also changes tags from a string to a dictionary


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// After filtering, it will add or remove tags.
+    /// </summary>
+    public class MetricFilter
+    {
+        readonly Dictionary<string, string> tagsWhitelist;
+        readonly Dictionary<string, string> tagsBlacklist;
+        readonly Dictionary<string, string> tagsToAdd;
+        readonly List<string> tagsToRemove;
+
+        public MetricFilter(Dictionary<string, string> tagsWhitelist = null, Dictionary<string, string> tagsBlacklist = null, Dictionary<string, string> tagsToAdd = null, List<string> tagsToRemove = null)
```
varunpuranik: typically if all of these are optional, then a builder pattern works pretty well.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            this.tagsWhitelist = tagsWhitelist;
```
varunpuranik: null args should be initialized to actual values in ctor

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+            this.tagsBlacklist = tagsBlacklist;
+            this.tagsToAdd = tagsToAdd;
+            this.tagsToRemove = tagsToRemove;
+        }
+
+        public IEnumerable<Metric> FilterMetrics(IEnumerable<Metric> metrics)
+        {
+            foreach (Metric metric in metrics)
+            {
+                Dictionary<string, string> tags = JsonConvert.DeserializeObject<Dictionary<string, string>>(metric.Tags);
```
varunpuranik: var?

varunpuranik: If the metrics are going to go through the filter every time... maybe we should consider storing tags as a dictionary... and then serializing it at the end?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+
+                // Skip if the blacklist has any or the whitelist does not contain it.
+                if ((this.tagsBlacklist != null && this.tagsBlacklist.Any(tags.Contains)) || (this.tagsWhitelist != null && !this.tagsWhitelist.Any(tags.Contains)))
+                {
+                    continue;
+                }
+
+                // Add or remove tags if needed.
+                if ((this.tagsToAdd != null && this.tagsToAdd.Any()) || (this.tagsToRemove != null && this.tagsToRemove.Any()))
```
varunpuranik: No nulls pls..

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    this.tagsToAdd?.ToList().ForEach(toAdd => tags.Add(toAdd.Key, toAdd.Value));
```
varunpuranik: Store as list instead of allocating a list every time

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// This class acts as a pass through for a group of metrics.
+    /// It will exclude metrics based on a tag whitelist and blacklist.
```
varunpuranik: I get the idea, but having both a whitelist and blacklist seems odd somehow. Can't we do just with a whitelist?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// After filtering, it will add or remove tags.
+    /// </summary>
+    public class MetricFilter
```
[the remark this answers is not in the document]
lfitchett: Metrics worker uses it

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
lfitchett: I'm only using the whitelist, but I wanted to keep it generic. Hopefully this will be re-used when we expose it to customers, so giving them a robust filter is good.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
lfitchett: I prefer being explicit. I don't think it detracts from the readability.

lfitchett: made dictionary

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
```diff
@@ -3,16 +3,20 @@
...
     public sealed class Metric : IEquatable<Metric>
     {
         public DateTime TimeGeneratedUtc { get; }
         public string Name { get; }
         public double Value { get; }
-        public string Tags { get; }
+        public IReadOnlyDictionary<string, string> Tags { get; }
 
-        public Metric(DateTime timeGeneratedUtc, string name, double value, string tags)
+        int? metricKey;
```
lt72: Why not init this in the ctor and not make it a nullable value at all?

lfitchett: It's not used always, so I want it to be lazy. In this case, scraping never needs this value.

lt72: may be fine, consider simplicity and readability too.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
```diff
@@ -3,45 +3,68 @@
...
-            // TODO: replace with "return HashCode.Combine(Name.GetHashCode(), Tags.GetHashCode());"
-            // when upgraded to .net standard 2.1: https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=netstandard-2.1
-            int hash = 17;
-            hash = hash * 31 + this.Name.GetHashCode();
-            hash = hash * 31 + this.Tags.GetHashCode();
+            return CombineHash(this.Name.GetHashCode(), GetOrderIndependentHash(this.Tags));
+        }
 
-            return hash;
+        public static int GetOrderIndependentHash<T1, T2>(IEnumerable<KeyValuePair<T1, T2>> dictionary)
```
varunpuranik: private

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,92 @@
...
+                    yield return new Metric(metric.TimeGeneratedUtc, metric.Name, metric.Value, newTags);
+                }
+                else
+                {
+                    yield return metric;
+                }
+            }
+        }
+
+        public MetricFilter AddToWhitelist(params KeyValuePair<string, string>[] pairs)
```
kkdawkins: I don't see any place where you make use of the params keyword... why not just take a list of KVPairs?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,92 @@
...
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// This class acts as a pass through for a group of metrics.
+    /// It will exclude metrics based on a tag whitelist and blacklist.
+    /// After filtering, it will add or remove tags.
+    /// </summary>
+    public class MetricFilter
+    {
+        Option<List<KeyValuePair<string, string>>> tagsWhitelist = Option.None<List<KeyValuePair<string, string>>>();
```
kkdawkins: IIRC, Whitelist and blacklist are restricted keywords and will flag SDL scans. Can you change those to "allowed" and "blocked" or something simliar to that?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
kkdawkins: This applies to all functions here.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
```diff
@@ -39,5 +41,21 @@ public async Task<bool> PublishAsync(IEnumerable<Metric> metrics, CancellationTo
                 return false;
             }
         }
+
+        class MetricWithStringTags
```
kkdawkins: I believe that our coding standards specify 1 class per file. Split this out?

kkdawkins: Also, why not just call this MetricWithTags? String is leaking implementation details.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
lfitchett: Made it a Lazy<T>

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
lfitchett: Removed blacklist to make it simple. It is easy to add in later if needed

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
lfitchett: Currently where this is used is 
```
this.metricFilter = new MetricFilter()
    .AddToWhitelist(new KeyValuePair<string, string>(MetricsConstants.MsTelemetry, true.ToString()))
    .AddTagsToRemove(MetricsConstants.MsTelemetry, MetricsConstants.IotHubLabel, MetricsConstants.DeviceIdLabel);
```

So both are using the params functionality. Currently, whitelist is only 1 value, but it makes sense to use params in case that grows in the future. It also keeps the signatures consistent.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
lfitchett: This class is a result of changing the metric class Tags from a string to a Dictionary. I think it should remain a private class nested in LogAnalyticsUpload, since it's a very slight transformation of the data that only makes sense just prior to upload. I will rename it to UploadMetric.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
lfitchett: Do we run SDL scans in check-in? It's been passing the pr build. If possible, I'd like to leave it whitelist because I think it's more clear.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/LogAnalyticsUpload.cs
lfitchett: If that's enforced by our coding pattern I can switch it. I'd like to keep it's scope small though. It should really only exist for LogAnalyticsUpload

lfitchett: I could also change it to 
```
var metricsToUpload = metrics.Select(m => new { m.TimeGeneratedUtc, m.Name, m.Value, Tags = JsonConvert.SerializeObject(m.Tags) });
```
and not have a declared class

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
damonbarry: Despite the inconvenience, I think it's better to find a different name. `allowedTags`, `filteredTags`, `tagsList`, `tagsFilter`, ...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+            else
+            {
+                this.tagsWhitelist = Option.Some(new List<KeyValuePair<string, string>>(pairs));
+            }
+
+            return this;
+        }
+
+        public MetricFilter AddTagsToAdd(params KeyValuePair<string, string>[] pairs)
```
damonbarry: Nit: Rename methods to just `AddTags` and `RemoveTags`?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
damonbarry: If you're replacing "whitelist", this method could be called `metricFilter.AllowTags(...)`...

edge-util/test/Microsoft.Azure.Devices.Edge.Util.Test.Common/TestUtilities.cs
```diff
@@ -11,5 +15,30 @@ public static void ApproxEqual(double expected, double actual, double tolerance)
...
+                {
+                    if (actualItem.Equals(expectedNode.Value))
+                    {
+                        break;
+                    }
+
+                    expectedNode = expectedNode.Next;
+                }
+
+                Assert.False(expectedNode == null, $"Didn't expected actual to contain {JsonConvert.SerializeObject(actualItem)}");
```
damonbarry: Nit: typo "Didn't expected..." should be "Didn't expect..."

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
```diff
@@ -3,45 +3,68 @@
...
-            return hash;
+        static int GetOrderIndependentHash<T1, T2>(IEnumerable<KeyValuePair<T1, T2>> dictionary)
+        {
+            return CombineHash(dictionary.Select(o => CombineHash(o.Key.GetHashCode(), o.Value.GetHashCode())).OrderBy(h => h).ToArray());
         }
 
-        public bool Equals(Metric other)
+        // TODO: replace with "return HashCode.Combine();"
+        // when upgraded to .net standard 2.1: https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=netstandard-2.1
+        static int CombineHash(int hash1, int hash2)
```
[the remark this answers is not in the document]
lfitchett:
```
        static int CombineHash(int hash1, int hash2)
        {
            return (17 * 31 + hash1) * 31 + hash2;
        }
```
Could be removed entirely since the second one uses params. I just made it b/c its slightly more efficient. 

Should I keep it for the efficiency or remove it for simplicity?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricFilter.cs
lfitchett: I like Add prefix since it shows intent

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Metric.cs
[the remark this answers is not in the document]
damonbarry: Looks like you removed it :)

PR #2162, opened 2019-12-18
needinfo should not be subjected to stale state

yophilav:
- `needinfo` is no longer consider stale

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -13,4 +13,6 @@ jobs:
         stale-issue-label: 'no-issue-activity'
         days-before-stale: 30
         days-before-close: 99999
-        exempt-issue-label: 'enhancement'
+        exempt-issue-label:
+          - 'enhancement'
+          - 'needinfo'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -16,3 +16,4 @@ jobs:
         exempt-issue-label:
           - 'enhancement'
           - 'needinfo'
+          - 'bug'


.github/workflows/stale.yml
```diff
@@ -13,4 +13,6 @@ jobs:
         stale-issue-label: 'no-issue-activity'
         days-before-stale: 30
         days-before-close: 99999
-        exempt-issue-label: 'enhancement'
+        exempt-issue-label:
```
varunpuranik: bug as well

yophilav: 🐛 🐞

.github/workflows/stale.yml
```diff
@@ -13,4 +13,7 @@ jobs:
...
+          - 'enhancement'
+          - 'needinfo'
```
philipktlin: It is correct for enhancement and bug label; but incorrect for needinfo label.

If a github issue is created and our dev mark it with "needinfo" label; then after stale period, customers doesn't provide required info, this issue should be marked as stale.

If a github issue is created, no one from our team is reviewed => no action taken; therefore even it is longer than stale period, this issue should not be marked as stale.  This case doesn't support in current github stale action.

It is good to have "bug" and "enhancement" for now.  I can see that there is more customized integration with github, e.g. enforce PR commit description.  I think we should think about an extensible solution to address them.

.github/workflows/stale.yml
```diff
@@ -13,4 +13,7 @@ jobs:
         stale-issue-label: 'no-issue-activity'
         days-before-stale: 30
```
philipktlin: @lt72, please confirm if you want to change days before stale from 30 to 10.  thanks.

lt72: yes, please

yophilav:
After some searching, it looks like the Git Stale bot does not support a conditional labeling e.g. if both label `bug` and `needinfo` are labeled, then the task can be exempted from label as a stale.

PR #2163, opened 2019-12-18
Fix E2E flake - Increase timeouts

and-rewsmith:
TempFilterFunctions is a large image which takes a long time to download, so we need to increase the timeout.

Our direct method tests rely on a message sent through event hub which serves as a signal that the direct method was successful. We should change this logic to have the direct method sender module report status directly rather than relying on eventHub

### DIFF
--- a/smoke/IotEdgeQuickstart/details/Details.cs
+++ b/smoke/IotEdgeQuickstart/details/Details.cs
@@ -364,7 +364,7 @@ namespace IotEdgeQuickstart.Details
                 EventPosition.FromEnd());
 
             var result = new TaskCompletionSource<bool>();
-            using (var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10)))
+            using (var cts = new CancellationTokenSource(TimeSpan.FromMinutes(20)))
             {
                 using (cts.Token.Register(() => result.TrySetCanceled()))
                 {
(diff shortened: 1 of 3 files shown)


smoke/IotEdgeQuickstart/details/Details.cs
```diff
@@ -364,7 +364,7 @@ protected async Task VerifyDataOnIoTHub(string moduleId)
                 EventPosition.FromEnd());
 
             var result = new TaskCompletionSource<bool>();
-            using (var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10)))
+            using (var cts = new CancellationTokenSource(TimeSpan.FromMinutes(20))) // This long timeout is needed in case event hub is slow to process messages
```
philipktlin: Please add TODO here, we should use the similar approach as what you did in long haul to verify test result instead of using events from event hub.

smoke/IotEdgeQuickstart/details/Details.cs
```diff
@@ -363,8 +363,9 @@ protected async Task VerifyDataOnIoTHub(string moduleId)
                     (await eventHubClient.GetRuntimeInformationAsync()).PartitionCount),
                 EventPosition.FromEnd());
 
+            // TODO: Don't rely on eventhub to verify module to module direct methods. The module sending the direct method should report success / failure directly.
```
philipktlin: Suggest to put more positive statement
```suggestion
            // TODO: [Improvement] should verify test results without using event hub, which introduce latency.
```

PR #2166, opened 2019-12-19
Make FormatTestResult more specific

dylanbronson:
FormatTestResult in ModuleUtil is specific to Messages. Renaming it to be more specific to Messages. 

Also adding FormatDirectMethodTestResult for DirectMethod test results. 

The resultMessage in DirectMethod tests will be the result of calling the Direct Method - which will be a success message or an exception.


test/modules/TestResultCoordinator/service/PartitionReceiverHandler.cs
```diff
@@ -68,7 +68,7 @@ public async Task ProcessEventsAsync(IEnumerable<EventData> events)
                                 var result = new TestOperationResult(
                                     (string)moduleIdFromEvent + ".eventHub",
                                     "Messages",
```
yophilav:
``` suggestion
TestOperationResultType.Messages.ToString(),
```

PR #2167, opened 2019-12-19
Ensure Metrics Ordering

lfitchett:
This switches the rocksdb metrics store to use a sequential store to ensure ordering is preserved. This is because the uploader relies on chronological order to reduce the size.

This also changes the scraper to query the ip each time instead of caching, so it doesn't break if the ip changes.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsScraper.cs
```diff
@@ -35,16 +35,16 @@ public MetricsScraper(IList<string> endpoints, ISystemTime systemTime = null)
             Preconditions.CheckNotNull(endpoints, nameof(endpoints));
 
             this.httpClient = new HttpClient();
-            this.endpoints = new Lazy<IDictionary<string, string>>(() => endpoints.ToDictionary(e => e, this.GetUriWithIpAddress));
+            this.endpoints = endpoints;
```
lt72: It seems we init the MetricsScraper instance with a list of endpoints as strings, but then we parse those strings *every time* we need to scrape.
Parsing should be done at construction time, thus validating input early and avoiding repeated cpu/memory usage.

lfitchett: To put this method in context, it is a temporary workaround for an issue in how edgehub exposes its metrics. The metrics library provides 2 ways to expose metrics: a webserver implementation and a way to hook into asp.net. We want to use the asp.net implementation, because the metrics webserver doesn't advertise correctly on the docker network (the reason the GetUriWithIpAddress function exists, we're manually resolving the ip). However, asp.net currently forces all traffic through https, so we need to add an exception to edgehub's auto-redirect. Prometheous is supposed to be exposed over http.

Initially, the result of resolving the ip and rebuilding the uri was stored. However, this breaks if the ip of the container changes. So I removed the cache. While this does add a little overhead, it is only a single precompiled regex check every scrape (default 1 per hour). The performance impact should be very negligible, especially since this method should be gone before GA.

lt72: I see, since this will go before GA, we can proceed.

PR #2170, opened 2019-12-20
Fix docker cpu stats

lfitchett:
Docker CPU stats were not multiplied by 100. 

Additionally, there was a bug where sometimes docker returns null network or block stats. those nulls are now ignored


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -158,54 +158,67 @@ void SetModuleStats(SystemResources systemResources)
...
                 if (this.previousModuleCpu.TryGetValue(module.Name, out ulong prevModule) && this.previousSystemCpu.TryGetValue(module.Name, out ulong prevSystem))
                 {
                     double moduleDiff = module.CpuStats.CpuUsage.TotalUsage - prevModule;
                     double systemDiff = module.CpuStats.SystemCpuUsage - prevSystem;
-                    result = moduleDiff / systemDiff;
+                    if (systemDiff > 0)
+                    {
+                        double result = 100 * moduleDiff / systemDiff;
+
+                        // Occasionally on startup results in a very large number (billions of percent). Ignore this point.
```
and-rewsmith: Good catch

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -111,7 +111,7 @@ public void Start(ILogger logger)
         {
             if (ApiVersion.ParseVersion(this.apiVersion).Value >= ApiVersion.Version20191105.Value)
             {
-                this.updateResources = new PeriodicTask(this.Update, TimeSpan.FromMinutes(5), TimeSpan.FromMinutes(1), logger, "Get system resources", false);
+                this.updateResources = new PeriodicTask(this.Update, TimeSpan.FromSeconds(30), TimeSpan.FromMinutes(1), logger, "Get system resources", false);
```
and-rewsmith: Why increase the frequency?

lfitchett: It's much more accurate to cpu usage. every 5 minutes doesn't catch any spikes. 30 seconds still has very little performance impact and is better about spikes. This should probably be configurable, I'll add that.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -40,11 +41,13 @@ public class SystemResourcesMetrics : IDisposable
         Dictionary<string, ulong> previousSystemCpu = new Dictionary<string, ulong>();
         Dictionary<string, DateTime> previousReadTime = new Dictionary<string, DateTime>();
 
-        public SystemResourcesMetrics(IMetricsProvider metricsProvider, Func<Task<SystemResources>> getSystemResources, string apiVersion)
+        public SystemResourcesMetrics(IMetricsProvider metricsProvider, Func<Task<SystemResources>> getSystemResources, string apiVersion, TimeSpan updateFrequency)
         {
             Preconditions.CheckNotNull(metricsProvider, nameof(metricsProvider));
             this.getSystemResources = Preconditions.CheckNotNull(getSystemResources, nameof(getSystemResources));
             this.apiVersion = Preconditions.CheckNotNull(apiVersion, nameof(apiVersion));
+            Preconditions.CheckArgument(updateFrequency > TimeSpan.FromSeconds(5), "Performance metrics cannot update faster than 5 seconds.");
```
lt72: I would suggest proving a 'public static readonly TimeSpan MaxUpdateFrequency' (or 'public const') member for user to easily discover what the ceiling is, and also remove the magic number.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/EdgeletModule.cs
```diff
@@ -67,6 +69,7 @@ public class EdgeletModule : Module
             this.productInfo = Preconditions.CheckNotNull(productInfo, nameof(productInfo));
             this.closeOnIdleTimeout = closeOnIdleTimeout;
             this.idleTimeout = idleTimeout;
+            this.performanceMetricsUpdateFrequency = performanceMetricsUpdateFrequency;
```
and-rewsmith: nit: Philip has told me in the past to validate inputs at every level, however not having maxUpdateFrequency in scope means you can't do full validation. Is there a formal convention for input validation?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -18,9 +18,12 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker
```
and-rewsmith: nit: System.Text not needed

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/EdgeletModule.cs
lfitchett: I don't want to validate it at this level, b/c if metrics are disabled this value is ignored. I only want to validate it if it is used.

PR #2172, opened 2019-12-20
Update the Direct Method Sender with Connectivity Test Parameters

yophilav:
- Update the Direct Method Sender with Connectivity Test Parameters
- Update InvokeDirectMethodAsync() to now return the DirectMethod number as well.


e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -107,6 +107,106 @@
...
+                "value": "Cloud"
+              },
+              "ServiceClientConnectionString": {
+                "value": "<ServiceClientConnectionString>"
+              },
+              "TargetModuleId": {
+                "value": "DirectMethodReceiver"
+              },
+              "TransportType" : {
+                "value": "Mqtt"
```
philipktlin: 1 pair of DM sender and receiver uses amqp and the other one uses mqtt.  Please set transport type for DirectMethodSender and DirectMethodReceiver to Amqp.

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -240,7 +340,8 @@
           "loadGen1ToRelayer1": "FROM /messages/modules/loadGen1/outputs/output1 INTO BrokeredEndpoint(\"/modules/relayer1/inputs/input1\")",
           "relayer1ToCloud": "FROM /messages/modules/relayer1/outputs/output1 INTO $upstream",
           "loadGen2ToRelayer2": "FROM /messages/modules/loadGen2/outputs/output2 INTO BrokeredEndpoint(\"/modules/relayer2/inputs/input2\")",
-          "relayer2ToCloud": "FROM /messages/modules/relayer2/outputs/output2 INTO $upstream"
+          "relayer2ToCloud": "FROM /messages/modules/relayer2/outputs/output2 INTO $upstream",
+          "directMethodSender1ToUpstream": "FROM /messages/modules/DirectMethodSender/outputs/output1 INTO $upstream"
```
philipktlin: Remove this line; it is not used in connectivity test.

edge-modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -27,7 +27,7 @@ public abstract class DirectMethodSenderBase : IDisposable
 
         public abstract void Dispose();
 
-        public async Task<HttpStatusCode> InvokeDirectMethodAsync(CancellationTokenSource cts)
+        public async Task<Tuple<HttpStatusCode, long>> InvokeDirectMethodAsync(CancellationTokenSource cts)
```
philipktlin: I think you need to pass tracking Id and directMethodCount when calling this.InvokeDirectMethodAsync; therefore when DM receiver received your call, it can extract trackingId and directMethodCount; and construct TestOperationResult to report to TRC.

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -24,21 +26,47 @@ public static async Task<int> MainAsync()
             (CancellationTokenSource cts, ManualResetEventSlim completed, Option<object> handler) = ShutdownHandler.Init(TimeSpan.FromSeconds(5), Logger);
             DirectMethodSenderBase directMethodClient = null;
             ModuleClient reportClient = null;
+            Option<Uri> analyzerUrl = Settings.Current.AnalyzerUrl;
+            Option<Uri> testReportCoordinatorUrl = Settings.Current.TestResultCoordinatorUrl;
+
             try
             {
+                Guid batchId = Guid.NewGuid();
+                Logger.LogInformation($"Batch Id={batchId}");
```
philipktlin: If you introduce batch id, please have this value sent to receiver, so it should have enough data to report to TRC.

yophilav: It is done below. Line:58-68

edge-modules/DirectMethodSender/DirectMethodSenderBase.cs
yophilav: It is done below. Line:58-68

philipktlin: The InvokeDeviceMethodAsync(string deviceId, string targetModuleId, CancellationToken none) signature only has device Id, target module id; but no tracking id and direct method count.  How DM receiver can get these 2 numbers?

edge-modules/DirectMethodSender/Program.cs
philipktlin: I am talking about when receiving on DM receiver; not in DM sender.

yophilav: Oh I see. I created a task for it.

edge-modules/DirectMethodSender/DirectMethodSenderBase.cs
yophilav: I see, I misread. I thought you meant the TRC, not the DMReceiver .
I think that should be a separate PR to have DMReceiver takes the Connectivity params.

philipktlin: Please add todo here if you plan to do it in another PR.

edge-modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -49,12 +49,12 @@ public async Task<HttpStatusCode> InvokeDirectMethodAsync(CancellationTokenSourc
 
                 logger.LogInformation($"Invoke DirectMethod with count {this.directMethodCount}: finished.");
                 this.directMethodCount++;
-                return (HttpStatusCode)resultStatus;
+                return new Tuple<HttpStatusCode, long>((HttpStatusCode)resultStatus, this.directMethodCount);
```
philipktlin: should it be better to make this.directMethodCount as protected, so derived class can access it instead of return from method?

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -24,21 +26,47 @@ public static async Task<int> MainAsync()
...
+                    await testReportCoordinatorUrl.ForEachAsync(
+                        async (Uri uri) =>
+                        {
+                            TestResultCoordinatorClient trcClient = new TestResultCoordinatorClient { BaseUrl = uri.AbsoluteUri };
+                            await ModuleUtil.ReportStatus(
+                                    trcClient,
+                                    Logger,
+                                    Settings.Current.ModuleId + ".send",
+                                    ModuleUtil.FormatDirectMethodTestResultValue(
+                                        Settings.Current.TrackingId.Expect(() => new ArgumentException("TrackingId is empty")),
```
philipktlin: I think for long haul and e2e test, tracking id is not required.  Why throws argument exception unconditionally if tracking id is missing?

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -24,21 +26,47 @@ public static async Task<int> MainAsync()
...
+                                        batchId.ToString(),
+                                        dmCounter.ToString(),
+                                        result.ToString()),
+                                    TestOperationResultType.DirectMethod.ToString());
+                        });
+
```
philipktlin: For connectivity test, analyzerUrl should not have value.  Therefore below code block will call "SendEventAsync" method.    You may need to work on the todo in this PR to create an abstract class to handle reporting.

edge-modules/DirectMethodSender/Settings.cs
```diff
@@ -38,16 +43,29 @@ public class Settings
...
             this.DirectMethodDelay = directMethodDelay;
+            this.InvocationSource = invocationSource;
             this.AnalyzerUrl = analyzerUrl;
             this.ServiceClientConnectionString = serviceClientConnectionString;
-            this.InvocationSource = invocationSource;
+            this.ModuleId = Preconditions.CheckNonWhiteSpace(moduleId, nameof(moduleId));
+            this.TestDuration = testDuration;
+            this.TestStartDelay = testStartDelay;
+            this.TestResultCoordinatorUrl = testResultCoordinatorUrl;
+            this.TrackingId = trackingId;
```
philipktlin: I think you should validate tracking id if testresultcoordinatorurl is provided

edge-modules/DirectMethodSender/DirectMethodSenderBase.cs
yophilav: Correct me if I misunderstanding but currently the derived classes `DirectMethodCloudSender` and `DirectMethodLocalSender` is `this.directMethodCount` agnostic. The reason this function has to return the value as a tuple because the program (caller) of the method needs a way to access the parameter for reporting.

edge-modules/DirectMethodSender/Program.cs
yophilav: This module is shared across E2E, LH, and Stress test. By the nature of the LH, the reporting is done using the default routing from the config file to send the report to EventHub

edge-modules/DirectMethodSender/DirectMethodSenderBase.cs
yophilav: The task: https://msazure.visualstudio.com/One/_workitems/edit/5954482 was created to address this issue.

edge-modules/DirectMethodSender/Settings.cs
yophilav: I wouldn't necessary want to do that in case if a developer wants to deploy the code locally and too lazy to provide a trackingId since it doesn't really provide an additional value for a local smoke test.

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -107,14 +107,111 @@
...
           },
+          "directMethodSender1": {
+            "version": "1.0",
+            "type": "docker",
+            "status": "running",
+            "restartPolicy": "always",
+            "env": {
+              "InvocationSource": {
+                "value": "Cloud"
+              },
```
ancaantochi: Module to module DM should be also included in connectivity tests.

edge-modules/DirectMethodSender/Settings.cs
```diff
@@ -28,7 +28,12 @@ public class Settings
                     configuration.GetValue<TimeSpan>("DirectMethodDelay", TimeSpan.FromSeconds(5)),
                     Option.Maybe(configuration.GetValue<Uri>("AnalyzerUrl")),
                     configuration.GetValue<InvocationSource>("InvocationSource", InvocationSource.Local),
```
ancaantochi: >InvocationSource [](start = 81, length = 16)

Is InvocationSource used in other modules or why is it it the module util?

edge-modules/DirectMethodSender/Settings.cs
ancaantochi: I agree with Philip that it should be validated, sorry for lazy developers 

---
In reply to: [361519203](https://github.com/Azure/iotedge/pull/2172#discussion_r361519203) [](ancestors = 361519203)

e2e_deployment_files/connectivity_deployment.template.json
yophilav: I created a task to add this and test: https://msazure.visualstudio.com/One/_workitems/edit/5972370

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -24,21 +26,47 @@ public static async Task<int> MainAsync()
...
+                await Task.Delay(Settings.Current.TestStartDelay);
+
+                DateTime testStartAt = DateTime.UtcNow;
+                while (!cts.Token.IsCancellationRequested && IsTestTimeUp(testStartAt))
                 {
-                    HttpStatusCode result = await directMethodClient.InvokeDirectMethodAsync(cts);
+                    (HttpStatusCode result, long dmCounter) = await directMethodClient.InvokeDirectMethodAsync(cts);
 
                     // TODO: Create an abstract class to handle the reporting client generation
+                    await testReportCoordinatorUrl.ForEachAsync(
```
ancaantochi: Can you please add an interface with SendReportAsync (like in twinTester changes) and have 2 implementations, one for trc, one for analyzer. Based on analyzerUrl or trc url you can create an instance of the analyzerReportHandler or trcReportHandler

edge-modules/DirectMethodSender/Settings.cs
yophilav: I was thinking that other module may use it.

edge-modules/DirectMethodSender/Program.cs
yophilav: Yeah, I had some idea on how to do this with a nice way by using abstract class: https://msazure.visualstudio.com/One/_workitems/edit/5949787. 
Since it can be used across multiple module, I'm thinking I will do it in Module Utils and then back port it to DM and Twin testers.  Would it be alright with you? I'm thinking of just merging the baseline (function block) of the DM tester first. Then slowly introduce more feature to it.

edge-modules/DirectMethodSender/DirectMethodSenderBase.cs
philipktlin: Your understanding is correct, my question is why don't make direct method count accessible by derived classes if all derived classes need to use it?

edge-modules/DirectMethodSender/Program.cs
philipktlin: For any changes added to existing code base, you need to make sure that all existing use cases work as before; and new use case should work as designed.  This part is not what it should behave in connectivity test.  You should introduce a setting parameter to distinguish what use case this module is running.  That's why @ancaantochi ask you to have an interface for reporting in this PR.

philipktlin: If you think this reporting code can be shared (e.g. interface/abstract), please implement in DM projects first.  Then have another PR to refactor and reuse in other projects.

edge-modules/DirectMethodSender/Program.cs
```diff
@@ -24,21 +26,47 @@ public static async Task<int> MainAsync()
...
                 reportClient = await ModuleUtil.CreateModuleClientAsync(
                     Settings.Current.TransportType,
                     ModuleUtil.DefaultTimeoutErrorDetectionStrategy,
                     ModuleUtil.DefaultTransientRetryStrategy,
                     Logger);
-                while (!cts.Token.IsCancellationRequested)
+
+                Logger.LogInformation($"Load gen delay start for {Settings.Current.TestStartDelay}.");
+                await Task.Delay(Settings.Current.TestStartDelay);
+
```
ancaantochi: add cts.Token

PR #2175, opened 2019-12-20
Refactoring NetworkController

dylanbronson:
Refactoring NetworkController for the upcoming DirectMethod Report changes. 

NetworkControllerResult will be in shared space ModuleUtil so that both Network Controller and TestResultCoordinator can use it.

Description now becomes NetworkStatus and NetworkStatus now becomes enabled.
This makes more sense with the future design of NetworkController. Deleted NetworkInterfaceController because it hasn't been working. We can add it back once we get it working.


edge-modules/ModuleLib/networkcontrollerreuslt/NetworkControllerResult.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace ModuleUtil.NetworkControllerResult
+{
+    using Newtonsoft.Json;
```
ancaantochi: >using Newtonsoft.Json; [](start = 4, length = 22)

add a newline after usings

edge-modules/ModuleLib/networkcontrollerreuslt/NetworkControllerResult.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public string TrackingId { get; set; }
+
+        public string Operation { get; set; }
+
+        public string OperationStatus { get; set; }
+
+        public NetworkStatus NetworkStatus { get; set; }
+
+        public bool Enabled { get; set; }
```
ancaantochi: It looks like the naming is a bit confusing, I would say NetworkStatus should be more like NetworkControllerType and Enabled is should be NetworkStatus enum with Enabled/Disabled.

test/connectivity/modules/NetworkController/LinuxFirewallOfflineController.cs
```diff
@@ -34,11 +35,11 @@ public async Task<NetworkStatus> GetStatusAsync(CancellationToken cs)
                 // parse output to see if online or offline
                 if (output.Contains("qdisc noqueue"))
                 {
-                    return NetworkStatus.Default;
+                    return NetworkStatus.Disabled;
```
ancaantochi: >NetworkStatus.Disabled [](start = 27, length = 22)

NetworkStatus.Disabled it can be understood as the network is disabled, is not clear that actually the rule is disabled. I think NetworkStatus should be renamed to ControllerStatus or similar

test/connectivity/modules/NetworkController/Program.cs
```diff
@@ -97,32 +94,34 @@ static async Task RemoveAllControllingRules(IList<INetworkController> controller
...
 
-        static async Task<bool> CheckSetStatusAsyncResult(bool success, NetworkStatus status, INetworkController controller, CancellationToken cs)
+        static async Task<bool> CheckSetNetworkStatusAsyncResult(bool success, NetworkStatus networkStatus, INetworkController controller, CancellationToken cs)
         {
-            NetworkStatus reportedStatus = await controller.GetStatusAsync(cs);
+            NetworkStatus reportedStatus = await controller.GetNetworkStatusAsync(cs);
 
             string resultMessage = success ? "succeded" : "failed";
-            Log.LogInformation($"Command SetStatusAsync {status} execution {resultMessage}, network status {reportedStatus}");
+            Log.LogInformation($"Command SetEnabledAsync to {networkStatus} execution {resultMessage}, network status {networkStatus}");
```
ancaantochi: >SetEnabledAsync [](start = 41, length = 15)

Should be SetNetworkStatus

test/connectivity/modules/NetworkController/Program.cs
```diff
@@ -97,32 +94,34 @@ static async Task RemoveAllControllingRules(IList<INetworkController> controller
...
-                    online = await CheckSetStatusAsyncResult(online, NetworkStatus.Default, controller, cancellationToken);
+                    Log.LogInformation($"Network restriction is enabled for {controller.NetworkControllerType}. Setting default");
+                    bool online = await controller.SetNetworkStatusAsync(NetworkStatus.Disabled, cancellationToken);
+                    online = await CheckSetNetworkStatusAsyncResult(online, NetworkStatus.Disabled, controller, cancellationToken);
                     if (!online)
                     {
                         Log.LogError($"Failed to ensure it starts with default values.");
                         throw new TestInitializationException();
                     }
+                    else
```
ancaantochi: >else [](start = 20, length = 4)

this should be after the foreach as it was before because it has to check all rules before it reports the status

PR #2177, opened 2019-12-21
Twin test results validation

ancaantochi:
Add twin report generation to the TestResultCoordinator

### DIFF
--- a/edge-modules/TwinTester/Program.cs
+++ b/edge-modules/TwinTester/Program.cs
@@ -22,17 +22,20 @@ namespace TwinTester
 
             try
             {
+                await Task.Delay(Settings.Current.TestStartDelay);
+
                 RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Settings.Current.ServiceClientConnectionString);
 
-                using (ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager, Settings.Current.AnalyzerUrl))
+                using (ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager))
                 {
                     await twinOperator.Start();
-
-                    await cts.Token.WhenCanceled();
-                    completed.Set();
-                    handler.ForEach(h => GC.KeepAlive(h));
-                    Logger.LogInformation("TwinTester exiting.");
+                    await Task.Delay(Settings.Current.TestDuration, cts.Token);
                 }
+
+                await cts.Token.WhenCanceled();
+                completed.Set();
+                handler.ForEach(h => GC.KeepAlive(h));
+                Logger.LogInformation("TwinTester exiting.");
             }
             catch (Exception ex)
             {
@@ -40,21 +43,21 @@ namespace TwinTester
             }
         }
 
-        static async Task<ITwinTestInitializer> GetTwinOperatorAsync(RegistryManager registryManager, Uri analyzerClientUri)
+        static async Task<ITwinTestInitializer> GetTwinOperatorAsync(RegistryManager registryManager)
         {
-            switch (Settings.Current.TestMode)
+            switch (Settings.Current.TwinTestMode)
             {
-                case TestMode.TwinCloudOperations:
-                    return await TwinCloudOperationsInitializer.CreateAsync(registryManager, new TwinEdgeOperationsResultHandler(analyzerClientUri, Settings.Current.ModuleId));
-                case TestMode.TwinEdgeOperations:
+                case TwinTestMode.TwinCloudOperations:
+                    return await TwinCloudOperationsInitializer.CreateAsync(registryManager, new TwinEdgeOperationsResultHandler(Settings.Current.ReporterUrl, Settings.Current.ModuleId, Settings.Current.TrackingId));
+                case TwinTestMode.TwinEdgeOperations:
                     ModuleClient moduleClient = await ModuleUtil.CreateModuleClientAsync(
                         Settings.Current.TransportType,
                         ModuleUtil.DefaultTimeoutErrorDetectionStrategy,
                         ModuleUtil.DefaultTransientRetryStrategy,
                         Logger);
-                    return await TwinEdgeOperationsInitializer.CreateAsync(registryManager, moduleClient, new TwinEdgeOperationsResultHandler(analyzerClientUri, Settings.Current.ModuleId));
+                    return await TwinEdgeOperationsInitializer.CreateAsync(registryManager, moduleClient, new TwinEdgeOperationsResultHandler(Settings.Current.ReporterUrl, Settings.Current.ModuleId, Settings.Current.TrackingId));
                 default:
-                    return await GetTwinAllOperationsInitializer(registryManager, analyzerClientUri);
+                    return await GetTwinAllOperationsInitializer(registryManager, Settings.Current.ReporterUrl);
             }
         }
 
--- a/edge-modules/TwinTester/Settings.cs
+++ b/edge-modules/TwinTester/Settings.cs
@@ -23,7 +23,7 @@ namespace TwinTester
                 string moduleId = configuration.GetValue<string>("IOTEDGE_MODULEID");
 
                 return new Settings(
-                    configuration.GetValue<string>("IOTEDGE_DEVICEID", string.Empty),
+                    configuration.GetValue<string>("IOTEDGE_DEVICEID"),
                     moduleId,
                     configuration.GetValue<string>("TargetModuleId", moduleId),
                     configuration.GetValue<int>("TwinUpdateSize", 1),
--- a/edge-modules/TwinTester/Settings.cs
+++ b/edge-modules/TwinTester/Settings.cs
@@ -30,7 +30,7 @@ namespace TwinTester
                     configuration.GetValue<TimeSpan>("TwinUpdateFrequency", TimeSpan.FromSeconds(10)),
                     configuration.GetValue<TimeSpan>("TwinUpdateFailureThreshold", TimeSpan.FromMinutes(1)),
                     configuration.GetValue<TransportType>("TransportType", TransportType.Amqp_Tcp_Only),
-                    configuration.GetValue<string>("AnalyzerUrl", http://analyzer:15000),
+                    configuration.GetValue<string>("AnalyzerUrl", "http://analyzer:15000"),
                     configuration.GetValue<string>("TestResultCoordinatorUrl"),
                     configuration.GetValue<string>("ServiceClientConnectionString"),
                     configuration.GetValue<string>("StoragePath"),
--- a/edge-modules/TwinTester/Settings.cs
+++ b/edge-modules/TwinTester/Settings.cs
@@ -30,7 +30,7 @@ namespace TwinTester
                     configuration.GetValue<TimeSpan>("TwinUpdateFrequency", TimeSpan.FromSeconds(10)),
                     configuration.GetValue<TimeSpan>("TwinUpdateFailureThreshold", TimeSpan.FromMinutes(1)),
                     configuration.GetValue<TransportType>("TransportType", TransportType.Amqp_Tcp_Only),
-                    configuration.GetValue<string>("AnalyzerUrl"),
+                    configuration.GetValue<string>("AnalyzerUrl", http://analyzer:15000),
                     configuration.GetValue<string>("TestResultCoordinatorUrl"),
                     configuration.GetValue<string>("ServiceClientConnectionString"),
                     configuration.GetValue<string>("StoragePath"),
@@ -74,14 +74,14 @@ namespace TwinTester
             this.TestStartDelay = testStartDelay;
             this.TestDuration = testDuration;
 
-            if (!string.IsNullOrWhiteSpace(analyzerUrl))
+            if (!string.IsNullOrWhiteSpace(testResultCoordinatorUrl))
             {
-                this.ReporterUrl = new Uri(analyzerUrl);
+                this.ReporterUrl = new Uri(testResultCoordinatorUrl);
+                trackingId.Expect(() => new ArgumentNullException(nameof(trackingId)));
             }
             else
             {
-                this.ReporterUrl = new Uri(Preconditions.CheckNonWhiteSpace(testResultCoordinatorUrl, nameof(testResultCoordinatorUrl)));
-                trackingId.Expect(() => new ArgumentNullException(nameof(trackingId)));
+                this.ReporterUrl = new Uri(Preconditions.CheckNonWhiteSpace(analyzerUrl, nameof(analyzerUrl)));
             }
         }
 
--- a/edge-modules/TwinTester/Settings.cs
+++ b/edge-modules/TwinTester/Settings.cs
@@ -20,47 +20,69 @@ namespace TwinTester
                     .AddEnvironmentVariables()
                     .Build();
 
-                string moduleId = configuration.GetValue<string>("TargetModuleId");
-                moduleId = string.IsNullOrWhiteSpace(moduleId) ? configuration.GetValue<string>("IOTEDGE_MODULEID", string.Empty) : moduleId;
+                string moduleId = configuration.GetValue<string>("IOTEDGE_MODULEID");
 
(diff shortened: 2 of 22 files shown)


edge-modules/TwinTester/Settings.cs
```diff
@@ -20,26 +20,32 @@ public class Settings
                     .AddEnvironmentVariables()
                     .Build();
 
-                string moduleId = configuration.GetValue<string>("TargetModuleId");
-                moduleId = string.IsNullOrWhiteSpace(moduleId) ? configuration.GetValue<string>("IOTEDGE_MODULEID", string.Empty) : moduleId;
+                string moduleId = configuration.GetValue<string>("IOTEDGE_MODULEID", string.Empty);
+                string targetModuleId = configuration.GetValue<string>("TargetModuleId", moduleId);
+
+                string analyzerUrl = configuration.GetValue<string>("AnalyzerUrl", "http://analyzer:15000");
+                string reporterUrl = configuration.GetValue<string>("testResultCoordinatorUrl", analyzerUrl);
```
dylanbronson: Why do we need both of these endpoints? Your later changes only include reporterUrl, which I thought would be a general reporter (configurable for wherever we report to)

edge-modules/TwinTester/Settings.cs
```diff
@@ -20,26 +20,32 @@ public class Settings
...
                     configuration.GetValue<TimeSpan>("TwinUpdateFailureThreshold", TimeSpan.FromMinutes(1)),
                     configuration.GetValue<TransportType>("TransportType", TransportType.Amqp_Tcp_Only),
-                    configuration.GetValue<Uri>("AnalyzerUrl", new Uri("http://analyzer:15000")),
+                    new Uri(reporterUrl),
                     configuration.GetValue<string>("ServiceClientConnectionString"),
                     configuration.GetValue<string>("StoragePath"),
                     configuration.GetValue<bool>("StorageOptimizeForPerformance", true),
-                    configuration.GetValue<TestMode>("TestMode", TestMode.TwinAllOperations));
+                    configuration.GetValue<TestMode>("TestMode", TestMode.TwinAllOperations),
+                    Option.Maybe(configuration.GetValue<string>("trackingId", null)));
```
dylanbronson: This is Optional because it is only used in Connectivity Tests - is that right?

edge-modules/TwinTester/Settings.cs
```diff
@@ -48,19 +54,22 @@ public class Settings
...
         {
             this.DeviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
             this.ModuleId = Preconditions.CheckNonWhiteSpace(moduleId, nameof(moduleId));
+            this.TargetModuleId = Preconditions.CheckNonWhiteSpace(moduleId, nameof(targetModuleId));
             this.TwinUpdateSize = Preconditions.CheckRange(twinUpdateSize, 1);
             this.TwinUpdateFrequency = Preconditions.CheckNotNull(twinUpdateFrequency);
             this.TwinUpdateFailureThreshold = Preconditions.CheckNotNull(twinUpdateFailureThreshold);
             this.TransportType = Preconditions.CheckNotNull(transportType);
-            this.AnalyzerUrl = Preconditions.CheckNotNull(analyzerUrl, nameof(analyzerUrl));
+            this.ReporterUrl = Preconditions.CheckNotNull(analyzerUrl, nameof(analyzerUrl));
```
dylanbronson: Also confused why we are using the analyzerUrl for the reporterUrl when we have a reporterUrl variable above

edge-modules/TwinTester/Settings.cs
ancaantochi: Because TwinTester reports to analyzer in long haul and testResultCoordinator in connectvity tests. So reporterUrl will try to use testResultCoordinatorUrl, if it doesn't exist use analyzerUrl

edge-modules/TwinTester/Settings.cs
ancaantochi: yes

edge-modules/TwinTester/Settings.cs
ancaantochi: I renamed to reporterUrl. The idea is that reporterUrl is the value that is passed by analyzerUrl or testResultCoordinatorUrl env variable.

edge-modules/TwinTester/Program.cs
```diff
@@ -24,7 +24,7 @@ static async Task Main()
             {
                 RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Settings.Current.ServiceClientConnectionString);
 
-                using (ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager, Settings.Current.AnalyzerUrl))
+                using (ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager, Settings.Current.ReporterUrl))
```
philipktlin: update argument name "analyzerClientUri" to "reportUrl" in GetTwinOperatorAsync method.

philipktlin: should it be better to pass Settings to GetTwinOperatorAsync method?

edge-modules/TwinTester/Settings.cs
```diff
@@ -20,47 +20,56 @@ public class Settings
...
                     configuration.GetValue<TimeSpan>("TwinUpdateFailureThreshold", TimeSpan.FromMinutes(1)),
                     configuration.GetValue<TransportType>("TransportType", TransportType.Amqp_Tcp_Only),
-                    configuration.GetValue<Uri>("AnalyzerUrl", new Uri("http://analyzer:15000")),
+                    new Uri(reporterUrl),
                     configuration.GetValue<string>("ServiceClientConnectionString"),
                     configuration.GetValue<string>("StoragePath"),
                     configuration.GetValue<bool>("StorageOptimizeForPerformance", true),
-                    configuration.GetValue<TestMode>("TestMode", TestMode.TwinAllOperations));
+                    configuration.GetValue<TestMode>("TestMode", TestMode.TwinAllOperations),
+                    Option.Maybe(configuration.GetValue<string>("trackingId", null)));
```
philipktlin: I think you don't need to provide default value "null'.

edge-modules/TwinTester/Settings.cs
philipktlin: since either testResultCoordinatorUrl or analyzerUrl will be used and assigned to reportUrl; should we just define a report url in config file?

edge-modules/TwinTester/Settings.cs
```diff
@@ -20,47 +20,56 @@ public class Settings
...
         {
             this.DeviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
             this.ModuleId = Preconditions.CheckNonWhiteSpace(moduleId, nameof(moduleId));
+            this.TargetModuleId = Preconditions.CheckNonWhiteSpace(moduleId, nameof(targetModuleId));
             this.TwinUpdateSize = Preconditions.CheckRange(twinUpdateSize, 1);
             this.TwinUpdateFrequency = Preconditions.CheckNotNull(twinUpdateFrequency);
             this.TwinUpdateFailureThreshold = Preconditions.CheckNotNull(twinUpdateFailureThreshold);
             this.TransportType = Preconditions.CheckNotNull(transportType);
-            this.AnalyzerUrl = Preconditions.CheckNotNull(analyzerUrl, nameof(analyzerUrl));
+            this.ReporterUrl = Preconditions.CheckNotNull(reporterUrl, nameof(reporterUrl));
```
philipktlin:
```suggestion
            this.ReporterUrl = Preconditions.CheckNonWhiteSpace(reporterUrl, nameof(reporterUrl));
```

edge-modules/TwinTester/Settings.cs
philipktlin: Another point is, we should handle this in a consistent way in both TwinTester and DirectMethodSender projects.  Please sync up with @yophilav.

edge-modules/TwinTester/TwinEdgeOperationsResultHandler.cs
```diff
@@ -4,36 +4,40 @@ namespace TwinTester
...
 
     class TwinEdgeOperationsResultHandler : ITwinTestResultHandler
     {
         static readonly ILogger Logger = ModuleUtil.CreateLogger(nameof(TwinEdgeOperationsResultHandler));
         readonly TestResultCoordinatorClient trcClient;
         readonly string moduleId;
+        readonly Option<string> trackingId;
 
-        public TwinEdgeOperationsResultHandler(Uri analyzerClientUri, string moduleId)
+        public TwinEdgeOperationsResultHandler(Uri analyzerClientUri, string moduleId, Option<string> trackingId)
```
philipktlin:
```suggestion
        public TwinEdgeOperationsResultHandler(Uri reportUri, string moduleId, Option<string> trackingId)
```

edge-modules/TwinTester/Settings.cs
```diff
@@ -20,47 +20,56 @@ public class Settings
...
             TimeSpan twinUpdateFrequency,
             TimeSpan twinUpdateFailureThreshold,
             TransportType transportType,
-            Uri analyzerUrl,
+            Uri reporterUrl,
             string serviceClientConnectionString,
             string storagePath,
             bool storageOptimizeForPerformance,
-            TestMode testMode)
+            TestMode testMode,
```
philipktlin: Please rename enum TestMode to TwinTestMode.

edge-modules/TwinTester/TwinEdgeOperationsResultHandler.cs
philipktlin: is this class only used in Connectivity test?

edge-modules/TwinTester/TwinEdgeOperationsResultHandler.cs
```diff
@@ -43,27 +47,23 @@ public Task HandleTwinValidationStatusAsync(string status)
 
         public Task HandleReportedPropertyUpdateExceptionAsync(string propertyKey, string failureStatus)
         {
-            return this.SendReportAsync(StatusCode.ReportedPropertyUpdateCallFailure, propertyKey, failureStatus);
+            TwinCollection properties = this.GetTwinCollection(propertyKey);
```
philipktlin: Why do we need to report fail twin update? is this used for long haul test?

edge-modules/TwinTester/Settings.cs
ancaantochi: reporterUrl has Uri type, for sure is not empty

edge-modules/TwinTester/TwinEdgeOperationsResultHandler.cs
ancaantochi: It was for long haul and I also reported for connectivity, but since it is not used in connectivity I removed it. This class handles connectivity reporting so it's fine to remove it.

edge-modules/TwinTester/TwinEdgeOperationsResultHandler.cs
ancaantochi: yes, it's used for connectivity only

edge-modules/TwinTester/Program.cs
ancaantochi: GetTwinOperatorAsync not uses the Settings.Current.ReporterUrl directly

test/modules/Modules.Test/TestResultCoordinator/CountingReportGeneratorTest.cs
```diff
@@ -245,7 +245,7 @@ public async Task TestCreateReportAsyncWithEmptyResults()
         {
             string expectedSource = "expectedSource";
             string actualSource = "actualSource";
-            string resultType = "resultType1";
+            string resultType = "Twin";
```
yophilav: Would it make sense to use the `TestOperationResultType` instead?

edge-modules/TwinTester/Settings.cs
```diff
@@ -20,47 +20,70 @@ public class Settings
                     .AddEnvironmentVariables()
                     .Build();
 
-                string moduleId = configuration.GetValue<string>("TargetModuleId");
-                moduleId = string.IsNullOrWhiteSpace(moduleId) ? configuration.GetValue<string>("IOTEDGE_MODULEID", string.Empty) : moduleId;
+                string moduleId = configuration.GetValue<string>("IOTEDGE_MODULEID", string.Empty);
+                string targetModuleId = configuration.GetValue<string>("TargetModuleId", moduleId);
+
+                string analyzerUrl = configuration.GetValue<string>("AnalyzerUrl");
+                string testResultCoordinatorUrl = configuration.GetValue<string>("testResultCoordinatorUrl");
```
yophilav: Is the TwinTester going to be part of the Long Haul and Stress tests? I'm thinking we probably can have a default endpoint here.

edge-modules/TwinTester/TwinEdgeOperationsResultHandler.cs
```diff
@@ -43,27 +47,22 @@ public Task HandleTwinValidationStatusAsync(string status)
 
         public Task HandleReportedPropertyUpdateExceptionAsync(string propertyKey, string failureStatus)
```
yophilav: Since this function no longer needs `string propertyKey ` and ` string failureStatus` please get rid of the input arguments.

edge-modules/TwinTester/Settings.cs
ancaantochi: yes, added the default and changed code to first check if trcUrl is set, otherwise use analyzer

---
In reply to: [361727022](https://github.com/Azure/iotedge/pull/2177#discussion_r361727022) [](ancestors = 361727022)

edge-modules/TwinTester/TwinEdgeOperationsResultHandler.cs
ancaantochi: I removed propertyKey, failureStatus is needed in TwinAllOperationsResultHandler for long haul 

---
In reply to: [361752363](https://github.com/Azure/iotedge/pull/2177#discussion_r361752363) [](ancestors = 361752363)

edge-modules/TwinTester/TwinEdgeOperationsResultHandler.cs
```diff
@@ -43,27 +47,22 @@ public Task HandleTwinValidationStatusAsync(string status)
...
-
-            public string TwinProperty { get; set; }
-
-            public string ErrorMessage { get; set; }
-
-            public override string ToString()
-            {
-                return JsonConvert.SerializeObject(this, Formatting.Indented);
-            }
+            var result = new TwinTestResult() { Operation = statusCode.ToString(), Properties = details, ErrorMessage = exception, TrackingId = this.trackingId.GetOrElse(string.Empty) };
```
yophilav: ` TrackingId = this.trackingId.GetOrElse(string.Empty) ` 
Are you sure you want to use an `string.Empty` instead of throwing an assert when `trackingId` is not set? 
I'd suggest using `TrackingId = this.trackingId.Expect(() => new ArgumentException("trackingId is null or empty"));`

edge-modules/TwinTester/Settings.cs
```diff
@@ -20,47 +20,70 @@ public class Settings
...
 
                 return new Settings(
                     configuration.GetValue<string>("IOTEDGE_DEVICEID", string.Empty),
                     moduleId,
```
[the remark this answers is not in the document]
ancaantochi: I moved trackingId inline, but moduleId is used as the default for targetId that's why is not inline

---
In reply to: [361703396](https://github.com/Azure/iotedge/pull/2177#discussion_r361703396) [](ancestors = 361703396)

edge-modules/TwinTester/Settings.cs
ancaantochi: As discussed, refactoring PR for direct method should take this into consideration

---
In reply to: [361517522](https://github.com/Azure/iotedge/pull/2177#discussion_r361517522) [](ancestors = 361517522)

PR #2178, opened 2019-12-21
Prepare edgehub-proxy builder for arm32v7 images

dmolokanov:
- Prepare Dockerfile for edgehub-proxy arm32v7 images
- Create Dockerfile for edgehub-proxy cross-compile builder image
- Add support for build.sh to build both amd64 and arm32v7 images


arsing:
This isn't new with this PR, but there's a problem with using `ekidd/rust-musl-builder`. It updates its rust version out of sync with us, which means its build breaks will not happen at the same time as the rest of our code, and despite https://github.com/Azure/iotedge/pull/2192 it means even release branches can have breakages.

One option is that we freeze the particular tag of `ekidd/rust-musl-builder` in this script to a particular Rust version tag when we create the release branch, instead of always using `latest`.

Another option is that we use a regular alpine container and install Rust inside it, which is determined by the `rust-toolchain` file, just like the packages build does with that PR.

Is it possible for you to use the second approach? Or is there some benefit to using `ekidd/rust-musl-builder` ?

dmolokanov:
@arsing We don't need to release edgehub-proxy along with the rest of iotedge platform for now. We just need to provide a way for customers to build their own custom image. Later on, we may want to improve this though.

PR #2186, opened 2019-12-23
Support module priority in IoT Edge

shantanu1singh:
With this change, a new property ‘Priority’ will be added to the deployment configuration of each module. This value will be used by Edge Agent to determine the order in which all other modules are to be created.

The value of ‘Priority’ will be positive and zero-based with 0 being the highest priority value. A higher numeric ‘Priority’ value would indicate a lower priority assigned to that module. The maximum value of this property will be UInt32.MaxValue (4294967295).

Modules that possess the same priority will be created at the same time and will have no deterministic startup order imposed amongst themselves. Modules with higher priority values will be created and started first and once Edge Agent has made an attempt to create and start them (only the modules with a desired state of ‘Running’ will be started), it will proceed with the creation (and optionally startup) of other lower priority modules in order.

Edge Agent will not support the ‘Priority’ property and will by default have the highest priority of zero (0).

Modules that do not have a ‘Priority’ value specified will be started in a non-deterministic order (as is the case today) and will be assigned a priority of UInt32.MaxValue which indicates the lowest possible priority. Edge Agent will trigger the creation of these modules after all modules with a higher priority have been created.

Module boot ordering will initially not be supported for the Kubernetes execution mode of IoT Edge.


sandervandevelde:
At this moment, the edgeHub module deployed together with the first module(s) a user adds to the running Edge.

Which priority does the edgeHub module get?

shantanu1singh:
@sandervandevelde If no priority value is explicitly specified for a module, it'll take the default priority value of UInt32.MaxValue. That applies to all modules other than Edge Agent including EdgeHub.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerEnvironment.cs
```diff
@@ -61,7 +61,7 @@ public async Task<ModuleSet> GetModulesAsync(CancellationToken token)
 
                 if (!moduleSet.Modules.TryGetValue(dockerRuntimeInfo.Name, out IModule configModule) || !(configModule is DockerModule dockerModule))
                 {
-                    dockerModule = new DockerModule(dockerRuntimeInfo.Name, string.Empty, ModuleStatus.Unknown, Core.RestartPolicy.Unknown, new DockerConfig(Constants.UnknownImage, new CreateContainerParameters()), ImagePullPolicy.OnCreate, new ConfigurationInfo(), null);
+                    dockerModule = new DockerModule(dockerRuntimeInfo.Name, string.Empty, ModuleStatus.Unknown, Core.RestartPolicy.Unknown, new DockerConfig(Constants.UnknownImage, new CreateContainerParameters()), ImagePullPolicy.OnCreate, Core.Constants.HighestPriority, new ConfigurationInfo(), null);
```
varunpuranik: This is HighestPriority because you want to remove it first, right?
Can you add a comment saying so..

shantanu1singh: Sure thing

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/planners/HealthRestartPlanner.cs
```diff
@@ -70,67 +70,95 @@ public async Task<Plan> CreateShutdownPlanAsync(ModuleSet current)
...
-                .Concat(removeState)
-                .Concat(addedCommands)
-                .Concat(updatedCommands)
-                .Concat(stateChangedCommands)
-                .Concat(desiredStatedChangedCommands.Select(d => d.command))
-                .Concat(resetHealthStatus)
-                .ToList();
+            foreach (IGrouping<uint, KeyValuePair<string, IModule>> priorityGroup in priorityGroups)
+            {
+                // The desired set is all the desired modules that have the priority of the current priority group being evaluated.
```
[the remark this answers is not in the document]
shantanu1singh: Sure, I'll change it to reduce the extra computation

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/planners/HealthRestartPlanner.cs
```diff
@@ -70,67 +70,95 @@ public async Task<Plan> CreateShutdownPlanAsync(ModuleSet current)
...
+                // - All the current modules that have the priority of the current priority group being evaluated and were not specified in the desired deployment config.
+                //   These are included so that they can be stopped and removed in the right priority order.
+                ModuleSet priorityBasedCurrentSet = ModuleSet.Create(
+                    current.Modules.Where(x => priorityBasedDesiredSet.Modules.ContainsKey(x.Key) ||
+                    priorityGroup.Any(y => y.Key.Equals(x.Key, StringComparison.OrdinalIgnoreCase)))
+                    .Select(y => y.Value)
+                    .ToArray());
+
+                // extract list of modules that need attention
+                (IList<IModule> added, IList<IModule> updateDeployed, IList<IModule> desiredStatusChanged, IList<IRuntimeModule> updateStateChanged, IList<IRuntimeModule> removed, IList<IRuntimeModule> runningGreat) = this.ProcessDiff(priorityBasedDesiredSet, priorityBasedCurrentSet);
```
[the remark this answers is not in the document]
shantanu1singh: Yep yep, will do

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/planners/HealthRestartPlanner.cs
```diff
@@ -70,6 +73,49 @@ public async Task<Plan> CreateShutdownPlanAsync(ModuleSet current)
...
+            {
+                // The desired set is all the desired modules that have the priority of the current priority group being evaluated.
+                ModuleSet priorityBasedDesiredSet = ModuleSet.Create(desiredPriorityGroups[priority].Select(x => x.Value).ToArray());
+
+                // The current set is:
+                // - All the current modules that correspond to the desired modules present in the current priority group.
+                // - All the current modules that have the priority of the current priority group being evaluated which were not specified in the desired deployment config
+                //   -and- have not already been processed yet.
+                //   These are included so that they can be stopped and removed in the right priority order.
+                IEnumerable<KeyValuePair<string, IModule>> desiredMatchingCurrentModules = current.Modules.Where(x => priorityBasedDesiredSet.Modules.ContainsKey(x.Key));
```
varunpuranik: So, we need this in case the priority of a deployed module has changed?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/planners/HealthRestartPlanner.cs
```diff
@@ -70,6 +73,49 @@ public async Task<Plan> CreateShutdownPlanAsync(ModuleSet current)
...
+            List<ICommand> commands = new List<ICommand>();
+
+            // Create a grouping of desired and current modules based on their priority.
+            // We want to process all the modules in the deployment (desired modules) and also include the modules
+            // that are not specified in the deployment but are currently running on the device. This is so that
+            // their processing is done in the right priority order.
+            ILookup<uint, KeyValuePair<string, IModule>> desiredPriorityGroups = desired.Modules.ToLookup(x => x.Value.Priority);
+            ILookup<uint, KeyValuePair<string, IModule>> currentPriorityGroups = current.Modules.ToLookup(x => x.Value.Priority);
+            ImmutableSortedSet<uint> orderedPriorities = desiredPriorityGroups.Select(x => x.Key).Union(currentPriorityGroups.Select(x => x.Key)).ToImmutableSortedSet();
+            HashSet<string> processedDesiredMatchingCurrentModules = new HashSet<string>();
```
varunpuranik: Nit: var

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/planners/HealthRestartPlanner.cs
varunpuranik: This is a bit confusing. Can we simplify it?

shantanu1singh: Umm, that's the best I could think of.

The problem is that we need to create priority based desired and current sets but also include the current modules in the current set that have the same names as modules in the desired set otherwise the commands generated will be incorrect and modules that should keep running might get removed or something

shantanu1singh: That's correct.

PR #2187, opened 2019-12-24
Revert "Ensure Metrics Ordering (#2167)"

ancaantochi:
This reverts commit f9ad608cfd047f4426beddaa48b8e0cd9d742530 because causes edgeAgent startup crash.


yophilav:
Did you try to deploy the artifact on any of the test pipeline? 
If those pass, I'd gladly approve this PR.   The commit actually compile successfully but it fails upon a deployment.

PR #2190, opened 2019-12-25
Enable metrics collector to report test type

and-rewsmith:
The metrics collector currently uses the test agent to determine test type. For connectivity tests multiple different test types will run on the same agent, so this strategy will stop working. Now we need to pass test type upon initialization.


and-rewsmith:
I am not sure how useful this will be going forward. I will close and reopen if we need to.

PR #2191, opened 2019-12-26
Update lang version to fix windows build

ancaantochi:


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/Microsoft.Azure.Devices.Edge.Hub.Core.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
```
varunpuranik: This is needed only for Windows?

ancaantochi: The build fails only on Windows. Not sure how it works on Linux, I assume the msbuild version is different

PR #2203, opened 2019-12-30
Fix E2E flake - Increase timeouts (#2163)

and-rewsmith:
Cherry pick of #2163
* increase timeout for tempFilterFunctions
* increase timeout for direct methods

### DIFF
--- a/smoke/IotEdgeQuickstart/details/Details.cs
+++ b/smoke/IotEdgeQuickstart/details/Details.cs
@@ -363,8 +363,9 @@ namespace IotEdgeQuickstart.Details
                     (await eventHubClient.GetRuntimeInformationAsync()).PartitionCount),
                 EventPosition.FromEnd());
 
+            // TODO: [Improvement] should verify test results without using event hub, which introduce latency.
             var result = new TaskCompletionSource<bool>();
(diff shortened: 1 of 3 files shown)


smoke/IotEdgeQuickstart/details/Details.cs
```diff
@@ -363,8 +363,9 @@ protected async Task VerifyDataOnIoTHub(string moduleId)
                     (await eventHubClient.GetRuntimeInformationAsync()).PartitionCount),
                 EventPosition.FromEnd());
 
+            // TODO: [Improvement] should verify test results without using event hub, which introduce latency.
             var result = new TaskCompletionSource<bool>();
-            using (var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10)))
+            using (var cts = new CancellationTokenSource(TimeSpan.FromMinutes(20))) // This long timeout is needed in case event hub is slow to process messages
```
philipktlin: I don't think it is slow in event hub; it is slow using event hub partition handler to pull events from event hub endpoints.  Please update the comment.

and-rewsmith: This is getting into semantics, but isn't the PartitionReceiver part of event hub? We can still be more specific though, so I will update in master then port to 109.

PR #2173, opened 2019-12-20
E2E flake - Increase timeouts for slow registry manager device creation

and-rewsmith:
Failing test:
https://msazure.visualstudio.com/One/_releaseProgress?_a=release-environment-logs&releaseId=1555799&environmentId=4042218

PR #2204, opened 2019-12-30
E2E flake - Increase timeouts for slow registry manager device creation

and-rewsmith:
Cherry pick of #2173 
* increase timeout for slow registry manager device creation

### DIFF
--- a/smoke/LeafDevice/details/Details.cs
+++ b/smoke/LeafDevice/details/Details.cs
@@ -143,29 +143,56 @@ namespace LeafDeviceTest
 
         protected async Task ConnectToEdgeAndSendDataAsync()
         {
-            var builder = IotHubConnectionStringBuilder.Create(this.iothubConnectionString);
-            DeviceClient deviceClient;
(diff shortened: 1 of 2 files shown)


smoke/LeafDevice/details/Details.cs
```diff
@@ -143,29 +143,56 @@ protected Task InitializeTrustedCertsAsync()
...
+                    {
+                        try
+                        {
+                            await deviceClient.SendEventAsync(message);
+                            Console.WriteLine("Message Sent.");
+                            await deviceClient.SetMethodHandlerAsync("DirectMethod", DirectMethod, null);
+                            Console.WriteLine("Direct method callback is set.");
+                            break;
+                        }
+                        catch (Exception e)
```
philipktlin: does it throw a specific exception when the device is registered, but not ready to use? if so, please catch that specific exception instead.

and-rewsmith: I will address this in a PR to master then port back to 109

PR #2205, opened 2019-12-30
Twin tester use target module id to update twin

ancaantochi:
- Fix twinTester to use targetModuleId. If env variable for targetModuleId is not set it is set to the moduleId value by default.
- Fix twinTester initialization for edge operations mode
- Replaced Dispose with Stop for test initializer to continue receiving desired properties updates after test duration

### DIFF
--- a/test/modules/TwinTester/Program.cs
+++ b/test/modules/TwinTester/Program.cs
@@ -20,11 +20,12 @@ namespace TwinTester
 
             Logger.LogInformation($"Starting twin tester with the following settings:\r\n{Settings.Current}");
 
+            ITwinTestInitializer twinOperator = null;
             try
             {
                 using (RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Settings.Current.ServiceClientConnectionString))
                 {
-                    ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager);
+                    twinOperator = await GetTwinOperatorAsync(registryManager);
                     await twinOperator.StartAsync(cts.Token);
                     await Task.Delay(Settings.Current.TestDuration, cts.Token);
 
@@ -40,6 +41,7 @@ namespace TwinTester
             catch (Exception ex)
             {
                 Logger.LogError(ex, $"Error occurred during twin test setup.");
+                twinOperator?.Stop();
             }
         }
 
--- a/test/modules/TwinTester/Program.cs
+++ b/test/modules/TwinTester/Program.cs
@@ -22,19 +22,20 @@ namespace TwinTester
 
             try
             {
-                RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Settings.Current.ServiceClientConnectionString);
+                using (RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Settings.Current.ServiceClientConnectionString))
+                {
+                    ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager);
+                    await twinOperator.StartAsync(cts.Token);
+                    await Task.Delay(Settings.Current.TestDuration, cts.Token);
 
-                ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager);
-                await twinOperator.StartAsync(cts.Token);
-                await Task.Delay(Settings.Current.TestDuration, cts.Token);
+                    Logger.LogInformation($"Test run completed after {Settings.Current.TestDuration}");
+                    twinOperator.Stop();
 
-                Logger.LogInformation($"Test run completed after {Settings.Current.TestDuration}");
-                twinOperator.Stop();
-
-                await cts.Token.WhenCanceled();
-                completed.Set();
-                handler.ForEach(h => GC.KeepAlive(h));
-                Logger.LogInformation("TwinTester exiting.");
+                    await cts.Token.WhenCanceled();
+                    completed.Set();
+                    handler.ForEach(h => GC.KeepAlive(h));
+                    Logger.LogInformation("TwinTester exiting.");
+                }
             }
             catch (Exception ex)
             {
--- a/test/modules/TwinTester/Program.cs
+++ b/test/modules/TwinTester/Program.cs
@@ -22,13 +22,11 @@ namespace TwinTester
 
             try
             {
-                await Task.Delay(Settings.Current.TestStartDelay);
-
                 RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Settings.Current.ServiceClientConnectionString);
 
                 using (ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager))
                 {
-                    await twinOperator.Start();
+                    await twinOperator.StartAsync(cts.Token);
                     await Task.Delay(Settings.Current.TestDuration, cts.Token);
                 }
 
--- a/test/modules/TwinTester/Program.cs
+++ b/test/modules/TwinTester/Program.cs
@@ -24,11 +24,12 @@ namespace TwinTester
             {
                 RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Settings.Current.ServiceClientConnectionString);
 
-                using (ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager))
-                {
-                    await twinOperator.StartAsync(cts.Token);
-                    await Task.Delay(Settings.Current.TestDuration, cts.Token);
-                }
+                ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager);
+                await twinOperator.StartAsync(cts.Token);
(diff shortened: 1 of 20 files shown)


test/modules/TestResultCoordinator/report/TwinCountingReport.cs
```diff
@@ -5,20 +5,23 @@ namespace TestResultCoordinator.Report
...
         }
 
         public ulong TotalExpectCount { get; }
 
         public ulong TotalMatchCount { get; }
 
-        public ulong TotalPatches { get; }
+        public ulong TotalPatchesCount { get; }
+
+        public ulong TotalDuplicateResultCount { get; }
```
philipktlin: why Twin will be duplicated?

test/modules/TwinTester/Program.cs
```diff
@@ -22,15 +22,14 @@ static async Task Main()
...
-                await Task.Delay(Settings.Current.TestStartDelay);
-
                 RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Settings.Current.ServiceClientConnectionString);
 
-                using (ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager))
-                {
-                    await twinOperator.Start();
-                    await Task.Delay(Settings.Current.TestDuration, cts.Token);
-                }
+                ITwinTestInitializer twinOperator = await GetTwinOperatorAsync(registryManager);
```
philipktlin: does registryManager implement IDisposable, right? if so, please call Dispose in Try...Finally.

test/modules/TwinTester/Program.cs
```diff
@@ -22,15 +22,14 @@ static async Task Main()
 
             try
             {
-                await Task.Delay(Settings.Current.TestStartDelay);
```
philipktlin: why remove test start delay?

test/modules/TwinTester/TwinEdgeOperationsInitializer.cs
```diff
@@ -17,41 +18,56 @@ class TwinEdgeOperationsInitializer : ITwinTestInitializer
...
-        public async Task Start()
+        public async Task StartAsync(CancellationToken cancellationToken)
         {
+            await Task.Delay(Settings.Current.TestStartDelay, cancellationToken);
+            await this.DisplayInitialTwin();
             this.periodicUpdate = new PeriodicTask(this.UpdateAsync, Settings.Current.TwinUpdateFrequency, Settings.Current.TwinUpdateFrequency, Logger, "TwinReportedPropertiesUpdate");
             await this.desiredPropertiesReceiver.UpdateAsync();
         }
 
-        public void Dispose()
```
philipktlin: why remove this method? should this class implement IDisposable since it owns periodic task which implements IDisposable. right?

test/modules/TestResultCoordinator/report/TwinCountingReport.cs
ancaantochi: I had a test run that failed to generate the report because of duplicates. It is not an expected behavior, it needs investigation, but I wasn't able to reproduce it.

test/modules/TwinTester/TwinEdgeOperationsInitializer.cs
ancaantochi: I changed the method to stop because it should not dispose the initializer, but just stop the periodictask. After the test completes it should continue to receive desired properties updates.

test/modules/TestResultCoordinator/report/TwinCountingReportGenerator.cs
```diff
@@ -46,13 +47,14 @@ public async Task<ITestResultReport> CreateReportAsync()
...
                 Option<TwinTestResult> twinTestResult = this.GetTwinTestResult(this.expectedTestResults.Current);
+                Logger.LogDebug($"Expected test results {twinTestResult}");
 
                 twinTestResult.ForEach(
                     r =>
                     {
                         foreach (var prop in r.Properties)
                         {
-                            propertiesUpdated.Add(prop.ToString(), this.expectedTestResults.Current.CreatedAt);
+                            propertiesUpdated.TryAdd(prop.ToString(), this.expectedTestResults.Current.CreatedAt);
```
philipktlin: The reason to use TryAdd here because avoiding crash if duplicates really happen, do I understand correctly?

test/modules/TwinTester/TwinEdgeOperationsInitializer.cs
```diff
@@ -17,41 +18,56 @@ class TwinEdgeOperationsInitializer : ITwinTestInitializer
...
+        async Task DisplayInitialTwin()
```
philipktlin: nit: rename
```suggestion
        async Task LogEdgeDeviceTwin()
```

test/modules/TwinTester/TwinTesterUtil.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TwinTester
+{
+    using System.Collections.Generic;
+    using Microsoft.Azure.Devices.Shared;
+
+    static class TwinTesterUtil
+    {
+        public static TwinCollection GetResetedReportedPropertiesTwin(Twin originalTwin)
```
philipktlin: currently have 2 usages; and both will use this method output to call moduleClient to reset report properties.  I think it be better to make this method to take moduleClient and do cleanup here.  what do you think?

test/modules/TwinTester/TwinEdgeOperationsResultHandler.cs
```diff
@@ -49,10 +49,10 @@ public Task HandleReportedPropertyUpdateExceptionAsync(string failureStatus)
             return Task.CompletedTask;
         }
 
-        TwinCollection GetTwinCollection(string propertyKey)
+        TwinCollection GetTwinCollection(string propertyKey, string value)
```
philipktlin:
```suggestion
        TwinCollection CreateTwinCollection(string propertyKey, string value)
```

test/modules/TwinTester/Program.cs
```diff
@@ -22,20 +22,20 @@ static async Task Main()
...
+                    await twinOperator.StartAsync(cts.Token);
                     await Task.Delay(Settings.Current.TestDuration, cts.Token);
-                }
 
-                await cts.Token.WhenCanceled();
-                completed.Set();
-                handler.ForEach(h => GC.KeepAlive(h));
-                Logger.LogInformation("TwinTester exiting.");
+                    Logger.LogInformation($"Test run completed after {Settings.Current.TestDuration}");
+                    twinOperator.Stop();
```
philipktlin: do you need to call stop explicitly when exception caught?

test/modules/TestResultCoordinator/report/TwinCountingReportGenerator.cs
ancaantochi: yes, that's correct

test/modules/TwinTester/Program.cs
ancaantochi: when there is an exception here the program ends.

test/modules/TwinTester/TwinAllOperationsInitializer.cs
```diff
@@ -111,7 +110,7 @@ public async Task PerformValidationAsync(CancellationToken cancellationToken)
             }
         }
 
-        public void Dispose()
+        public void Stop()
         {
             this.periodicValidation.Dispose();
```
philipktlin: it is safe to use null-conditional operator here as it will throw if Stop gets called without StartAsync.
```suggestion
            this.periodicValidation?.Dispose();
            this.periodicUpdate?.Dispose();
```

test/modules/TwinTester/TwinEdgeOperationsInitializer.cs
```diff
@@ -17,41 +18,56 @@ class TwinEdgeOperationsInitializer : ITwinTestInitializer
...
+            }
+            catch (Exception e)
+            {
+                Logger.LogError(e, $"Failure to get twin");
+            }
+        }
+
+        public void Stop()
         {
             this.periodicUpdate.Dispose();
```
philipktlin:
```suggestion
            this.periodicUpdate?.Dispose();
```

test/modules/TwinTester/TwinAllOperationsInitializer.cs
```diff
@@ -111,9 +110,10 @@ public async Task PerformValidationAsync(CancellationToken cancellationToken)
             }
         }
 
-        public void Dispose()
+        public void Stop()
         {
-            this.periodicValidation.Dispose();
+            this.periodicValidation?.Dispose();
+            this.periodicUpdate?.Dispose();
             this.periodicUpdate.Dispose();
```
philipktlin: remove this
```suggestion
```

PR #2206, opened 2019-12-30
Update k8s client, Watch extension changed slightly in update

darobs:


darobs:
Closing. Client version 1.6.8 does not work because we set a working directory in agent, and client cannot find the service token.  See issue: https://github.com/kubernetes-client/csharp/issues/341

PR #2209, opened 2019-12-31
Clean up Test Operation Result types and generated codes

philipktlin:
1. Clean up TestOperationResult and move to ModuleUtil for sharing.
2. Remove report test result method from AnalyzerClient swagger file and rename TRC Client swagger file to TestResultReportingClient.


test/modules/TwinTester/TwinAllOperationsResultHandler.cs
```diff
@@ -11,13 +11,13 @@ namespace TwinTester
...
         static readonly ILogger Logger = ModuleUtil.CreateLogger(nameof(TwinAllOperationsResultHandler));
-        readonly AnalyzerClient analyzerClient;
+        readonly TestResultReportingClient testResultReportingClient;
         readonly string moduleId;
         readonly TwinEventStorage storage;
 
         public TwinAllOperationsResultHandler(Uri analyzerClientUri, TwinEventStorage storage, string moduleId)
         {
-            this.analyzerClient = new AnalyzerClient() { BaseUrl = analyzerClientUri.AbsoluteUri };
+            this.testResultReportingClient = new TestResultReportingClient { BaseUrl = analyzerClientUri.AbsoluteUri };
```
yophilav: Um... I'm confused. Does the TwinTester no longer reporting to Analyzer instead it reports to TRC?  Or why is it using TRC Client/API to analyzer endpoint?

yophilav:
Thank you for the duplication of `TestOperationResult` between swaggers.

test/modules/TwinTester/TwinAllOperationsResultHandler.cs
philipktlin: You can think of both TestAnalyzer and TestResultCoordinator implemeted the same contract [PostAsync(TestOperationResult)].  Therefore I combine the client generated code for this contract and named it as TestResultReportingClient.  When using analyzerUrl, it means it send test result to TestAnalyzer.  The next step is to combine AnalyzerUrl and TestResultCoordinatorUrl into 1 property "reportUrl" in Settings as only one of these 2 variables should be used.

yophilav: I see where was my misunderstanding. Previously the TestAnalyzerClient does both Write (EndpointW) and Read(EndpointR) to the Analyzer module endpoints. Now your change made the TestAnalyzerClient a read-only module (only talks to EndpointR).  Since the TRC module and Analyzer module shares the same write endpoing (EndpointW), you can share the same API but only change the URI.  It was the TestAnalyzerClient changes from read-write to read-only that confused me because I saw two complete different endpoints using the same client on the diff version.

test/modules/ModuleLib/TestResultReportingClient.swagger.yaml
```diff
@@ -23,14 +23,14 @@ paths:
         description: "Test result details"
         required: true
         schema:
-          $ref: "#/definitions/TestOperationResult"
+          $ref: "#/definitions/ClientTestOperationResult"
```
yophilav: Can you make the changes to `description` on the top of the file to make it explicit that this is a writer API which is shared between Analyzer Module and TRC Module for a reporting client please?

test/modules/ModuleLib/TestAnalyzer.swagger.yaml
```diff
@@ -6,27 +6,6 @@ info:
 schemes:
 - "https"
 paths:
```
yophilav: Can you please update the `description` on top of the file that this is the reader endpoint for TestAnalyzer module please?

PR #2215, opened 2020-01-01
Disable flaky RefreshCacheWithRefreshRequestTest for dotnet checkin

and-rewsmith:


and-rewsmith:
I must have been looking at 109 when I saw this was still failing. It appears stable now.

PR #2219, opened 2020-01-02
Merge prometheus metrics additions to 109

lfitchett:
Merges in the additional metrics added to edgeHub and edgeAgent


lfitchett:
Going to make separate pr's for agent metrics and for host/docker metrics

PR #2222, opened 2020-01-03
Add different transport types for direct methods

and-rewsmith:
We should test both transport types for direct methods.


and-rewsmith:
Discussed with Philip and we don't need to test this as ServiceClient always should use amqp. The reason there are two DirectMethodSender modules is for sending to the two receivers (for which we do need different transport types).

PR #2231, opened 2020-01-03
Change k8s-openapi to use V1.15 (latest for current k8s client)

darobs:
Update to v1.15 and make minor changes to get it to compile and test.


edgelet/edgelet-kube/Cargo.toml
```diff
@@ -11,7 +11,7 @@ failure = "0.1"
 futures = "0.1"
 hyper = "0.12"
 hyper-tls = "0.3"
-k8s-openapi = { version = "0.5.1", features = ["v1_12"] }
+k8s-openapi = { version = "0.5.1", features = ["v1_15"] }
```
CindyXing: What's the reason to use 1.15 instead of 1.17? 

In our release note, we need to call out the supported K8s and Docker version(s).

darobs: That's the latest this version of k8s-openapi supports. v0.6.0 of k8s-openapi introduces some breaking changes which we don't want to include in the project yet.

darobs: Based on our current API usage (everything is V1 except the CRD in the Helm charts), we can operate on v1.10 through v.1.17.

PR #2236, opened 2020-01-04
Enable deployment test

philipktlin:
Add deployment test module in connectivity test deployment.

### DIFF
--- a/e2e_deployment_files/connectivity_deployment.template.json
+++ b/e2e_deployment_files/connectivity_deployment.template.json
@@ -117,7 +117,7 @@
                 "value": "Cloud"
               },
               "ServiceClientConnectionString": {
-                "value": "<ServiceClientConnectionString>"
+                "value": "<IoTHubConnectionString>"
               },
               "TargetModuleId": {
                 "value": "directMethodReceiver1"
@@ -171,7 +171,7 @@
                 "value": "Cloud"
               },
               "ServiceClientConnectionString": {
-                "value": "<ServiceClientConnectionString>"
+                "value": "<IoTHubConnectionString>"
               },
               "TargetModuleId": {
                 "value": "directMethodReceiver2"
@@ -304,7 +304,7 @@
                 "value": "<TestResultCoordinator.LogAnalyticsLogType>"
               },
               "ServiceClientConnectionString": {
-                "value": "<ServiceClientConnectionString>"
+                "value": "<IoTHubConnectionString>"
               }
             },
             "settings": {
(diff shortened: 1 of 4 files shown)


e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -444,6 +444,63 @@
               "createOptions": ""
             }
           },
+          "deploymentTest1": {
+            "version": "1.0",
+            "type": "docker",
+            "status": "running",
+            "restartPolicy": "always",
+            "env": {
+              "DEPLOYMENT_TESTER_MODE": {
```
ancaantochi: Prefix with IOTEDGE_

test/modules/DeploymentTester/Settings.cs
```diff
@@ -25,33 +25,48 @@ public class Settings
...
             this.DeviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
             this.ModuleId = Preconditions.CheckNonWhiteSpace(moduleId, nameof(moduleId));
-            this.IoTHubConnectionString = Preconditions.CheckNonWhiteSpace(iotHubConnectionString, nameof(iotHubConnectionString));
             this.TestResultCoordinatorUrl = Preconditions.CheckNotNull(testResultCoordinatorUrl, nameof(testResultCoordinatorUrl));
             this.TrackingId = Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
-            this.TargetModuleId = Preconditions.CheckNonWhiteSpace(targetModuleId, nameof(targetModuleId));
+
+            if (testMode == DeploymentTesterMode.Sender)
+            {
+                if (!iotHubConnectionString.HasValue)
```
ancaantochi: nit: You can use Preconditions.CheckArgument(!iotHubConnectionString.HasValue, nameof(iotHubConnectionString)). It will throw ArgumentException instead of ArgumentNullException

e2e_deployment_files/connectivity_deployment.template.json
philipktlin: Why do we want to add IOTEDGE_ Prefix; should it be used only when it is added by edgeAgent?  This "DEPLOYMENT_TESTER_MODE" is a custom env variable for DeploymentTester module only.

ancaantochi: I think you are right, the idea to add IOTEDGE_ was not to conflict with a var set by the custom module.

PR #2248, opened 2020-01-06
Fix tempfilter with IEnumerable

and-rewsmith:
TempFilter tests sometimes fail due to the modules starting before the necessary edgeHub routes are in place. I made a fix for this where the new e2e framework deploys in configurations. First it deploys edgeAgent / edgeHub, then the rest of the modules.


test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
```diff
@@ -84,9 +88,15 @@ public EdgeConfiguration Build()
                         ["properties.desired"] = module.DesiredProperties
                     };
                 }
+
+                if (module.Name == "$edgeHub")
```
philipktlin: use string.Equals with comparison option

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
```diff
@@ -72,7 +76,7 @@ public EdgeConfiguration Build()
             var moduleNames = new List<string>();
             var moduleImages = new List<string>();
 
-            foreach (ModuleConfiguration module in modules)
+            foreach (ModuleConfiguration module in modules.OrderBy(m => m.Name != "$edgeHub"))
```
philipktlin: should it be better to handle edgeHub first and use where clause for filtering to handle other modules

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeRuntime.cs
```diff
@@ -48,13 +48,18 @@ public async Task<EdgeDeployment> DeployConfigurationAsync(Action<EdgeConfigBuil
             addConfig(builder);
 
             DateTime deployTime = DateTime.Now;
-            EdgeConfiguration config = builder.Build();
+            IEnumerable<EdgeConfiguration> configs = builder.BuildConfigurationStages();
```
and-rewsmith: Let me know what you think of this name change

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
and-rewsmith: I made this change

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
and-rewsmith: I made this change

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
```diff
@@ -72,21 +75,25 @@ public EdgeConfiguration Build()
             var moduleNames = new List<string>();
             var moduleImages = new List<string>();
 
-            foreach (ModuleConfiguration module in modules)
+            string edgeHubModuleId = "$edgeHub";
```
philipktlin: uses Constants.EdgeHubModuleIdentityName in EdgeAgent.Core project

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
```diff
@@ -167,6 +171,25 @@ ModuleConfiguration BuildEdgeAgent(IEnumerable<ModuleConfiguration> configs)
             return agentBuilder.Build();
         }
 
+        bool IsSystemModule(ModuleConfiguration module)
+        {
+            return module.Name.Equals("$edgeHub") || module.Name.Equals("$edgeAgent");
```
philipktlin: Check with Damon if he have any concerns to reference EdgeAgent.Core project for these constants.

PR #2260, opened 2020-01-07
update connectivity test script to wait for test complete

philipktlin:
update connectivity test script to wait for test complete  in order to show test result in VSTS pipeline.

Verified with Connectivity Test: https://msazure.visualstudio.com/One/_build/results?buildId=27867503&view=logs&j=7745b47e-e2a2-549a-b47a-b50c2a0d5e6c&t=ddd08a52-8c2a-57b6-3587-e2d73b1904a3

### DIFF
--- a/test/connectivity/scripts/connectivityTest.sh
+++ b/test/connectivity/scripts/connectivityTest.sh
@@ -8,6 +8,11 @@ set -e
 # Import test-related functions
 . $(dirname "$0")/testHelper.sh
 
+function examine_test_result() {
+    #TODO: check TRC log for test report result
+    return 0
+}
+
 function prepare_test_from_artifacts() {
     print_highlighted_message 'Prepare test from artifacts'
 
@@ -58,19 +63,25 @@ function prepare_test_from_artifacts() {
     
 }
 
-function print_logs() {
-    local ret=$1
-    local test_end_time=$2
-    local elapsed_seconds=$3
+function print_deployment_logs() {
+    print_highlighted_message 'LOGS FROM IOTEDGED'
+    journalctl -u iotedge -u docker --since "$test_start_time" --no-pager || true
 
-    elapsed_time="$(TZ=UTC0 printf '%(%H:%M:%S)T\n' "$elapsed_seconds")"
-    print_highlighted_message "Test completed at $test_end_time, took $elapsed_time."
+    print_highlighted_message 'EDGE AGENT LOGS'
+    docker logs edgeAgent || true
+}
+
+function print_test_run_logs() {
+    local ret=$1
+
+    print_highlighted_message 'Print logs'
+    print_highlighted_message 'testResultCoordinator LOGS'
+    docker logs testResultCoordinator || true
 
     if (( ret < 1 )); then
         return;
     fi
 
-    print_highlighted_message 'Print logs'
     print_highlighted_message 'LOGS FROM IOTEDGED'
     journalctl -u iotedge -u docker --since "$test_start_time" --no-pager || true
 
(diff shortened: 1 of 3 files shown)


test/connectivity/scripts/connectivityTest.sh
```diff
@@ -58,33 +68,73 @@ function prepare_test_from_artifacts() {
...
         return;
     fi
 
-    print_highlighted_message 'Print logs'
     print_highlighted_message 'LOGS FROM IOTEDGED'
     journalctl -u iotedge -u docker --since "$test_start_time" --no-pager || true
 
-    print_highlighted_message 'EDGE AGENT LOGS'
+    print_highlighted_message 'edgeAgent LOGS'
     docker logs edgeAgent || true
```
and-rewsmith: Why ```|| true```?

Is this to prevent a failing exit code where module doesn't exist?

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -58,33 +68,73 @@ function prepare_test_from_artifacts() {
     
 }
 
-function print_logs() {
+function print_deployment_logs() {
+    print_highlighted_message 'LOGS FROM IOTEDGED'
+    journalctl -u iotedge -u docker --since "$test_start_time" --no-pager || true
+
+    print_highlighted_message 'EDGE AGENT LOGS'
```
and-rewsmith: ~We are potentially printing edge agent logs in two places.~ Also, the naming is inconsistent with ```edgeAgent LOGS```

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -175,6 +228,8 @@ function process_args() {
                 '-verificationDelay' ) saveNextArg=18;;
                 '-upstreamProtocol' ) saveNextArg=19;;
                 '-deploymentTestUpdatePeriod' ) saveNextArg=20;;
+                '-timeForReportingGeneration' ) saveNextArg=21;;
+                '-waitForTestComplete' ) WAIT_FOR_TEST_COMPLETE=1;;
```
and-rewsmith: nit: setting a precedent for bypassing the arg parsing pattern might decrease readability in the future

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -58,33 +68,73 @@ function prepare_test_from_artifacts() {
...
+    docker logs twinTester3 || true
+
+    print_highlighted_message 'twinTester4 LOGS'
+    docker logs twinTester4 || true
+
+    print_highlighted_message 'networkController LOGS'
+    docker logs networkController || true
 }
 
 function process_args() {
```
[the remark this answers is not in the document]
and-rewsmith: Having deployment scripts implemented in python would mean we wouldn't have to have separate scripts for windows / linux.

test/connectivity/scripts/connectivityTest.sh
philipktlin: yes.

test/connectivity/scripts/connectivityTest.sh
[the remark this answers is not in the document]
philipktlin: I think shell script/bash can probably be run on Windows 10 as well.  For 1st comment, I already extracted some functions in testHelper.sh in the same folder.

test/connectivity/scripts/connectivityTest.sh
philipktlin: user should just pass -waitForTestComplete argument only when run the script; therefore we need to assign the value here.

PR #2266, opened 2020-01-07
Add IsPassed Property in Test Reports

philipktlin:
Add IsPassed in TestReports and Create a TestSummary to wrap all test result and provide AreAllTestsPassed property.

### DIFF
--- a/test/modules/TestResultCoordinator/report/TestSummary.cs
+++ b/test/modules/TestResultCoordinator/report/TestSummary.cs
@@ -8,11 +8,11 @@ namespace TestResultCoordinator.Report
         internal TestSummary(ITestResultReport[] testResultReports)
         {
             this.TestResultReports = testResultReports;
-            this.AreAllTestsPassed = testResultReports.All(r => r.IsPassed);
+            this.IsPassed = testResultReports.All(r => r.IsPassed);
         }
 
-        public bool AreAllTestsPassed { get; private set; }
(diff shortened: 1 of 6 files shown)


test/modules/TestResultCoordinator/report/TestSummary.cs
```diff
@@ -0,0 +1,18 @@
...
+
+    class TestSummary
+    {
+        internal TestSummary(ITestResultReport[] testResultReports)
+        {
+            this.TestResultReports = testResultReports;
+            this.AreAllTestsPassed = testResultReports.All(r => r.IsPassed);
+        }
+
+        public bool AreAllTestsPassed { get; private set; }
```
ancaantochi: you can remove private set

test/modules/TestResultCoordinator/report/TestSummary.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace TestResultCoordinator.Report
+{
+    using System.Linq;
+
+    class TestSummary
+    {
+        internal TestSummary(ITestResultReport[] testResultReports)
+        {
+            this.TestResultReports = testResultReports;
+            this.AreAllTestsPassed = testResultReports.All(r => r.IsPassed);
```
ancaantochi: suggestion: name this IsPassed, I think it should be clear since it is in TestSummary that it means all test passed

PR #2271, opened 2020-01-08
Prevent new dependencies that pull in winapi 0.2 from being added

arsing:
This change adds a sanity check script for the checkin gates.
It performs the same patching that the Windows ARM32 build does to remove
all known uses of winapi 0.2, then asserts that winapi 0.2 is really gone
from the lockfile.


philipktlin:
Just to confirm; currently in master, no code references winapi 0.2, right?

arsing:
>Just to confirm; currently in master, no code references winapi 0.2, right?

iovec, mio, miow do, which is why we patch them. Apart from that nothing else does, otherwise ARM32 won't build at all.

PR #2286, opened 2020-01-10
Ping for EdgeAgent

yophilav:
Bugfix: 
- Update DirectMethod Receiver to correctly send the test result to TRC. 

Updates: 
- Deploy DirectMethod Sender to send `ping` method to EdgeAgent


test/modules/ModuleLib/ModuleUtil.cs
```diff
@@ -56,7 +56,7 @@ public static ILogger CreateLogger(string categoryName, LogEventLevel logEventLe
             return new LoggerFactory().AddSerilog().CreateLogger(categoryName);
         }
 
-        public static async Task ReportTestResultAsync(TestResultReportingClient apiClient, ILogger logger, TestResultBase testResult)
+        public static async Task ReportTestResultAsync(TestResultReportingClient apiClient, ILogger logger, TestResultBase testResult, bool isRethrow = false)
```
philipktlin: Can we always call throw? and lets' caller to handle.

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -166,6 +167,7 @@ List<string> GetResultSources()
                 "relayer2.eventHub",
                 "directMethodSender1.send",
                 "directMethodSender2.send",
+                "directMethodSender3.send",
```
philipktlin: what are you going to expect if both actual and expected are always the same?

test/modules/ModuleLib/ModuleUtil.cs
yophilav: It looks like there are multiple caller to this function that I don't want to make change right now. 
So, I'll make the default behavior a swallow exception. And if a new test module want the manage the exception, the caller can set the flag to be true.

test/modules/TestResultCoordinator/Settings.cs
yophilav: I actually have no idea. Dylan is working creating&verify process for DM tests.

test/modules/ModuleLib/ModuleUtil.cs
philipktlin: revert this fix.  I will take a look and better to handle in a separate PR.

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -143,6 +143,7 @@ List<IReportMetadata> InitializeReportMetadataList()
                 new CountingReportMetadata("relayer2.send", "relayer2.eventHub", TestOperationResultType.Messages, TestReportType.CountingReport),
                 new CountingReportMetadata("directMethodSender1.send", "directMethodReceiver1.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
                 new CountingReportMetadata("directMethodSender2.send", "directMethodReceiver2.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
+                new CountingReportMetadata("directMethodSender3.send", "directMethodSender3.send", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
```
philipktlin: Remove this line for now.

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -222,6 +222,46 @@
...
+              "DirectMethodResultType": {
+                "value": "DirectMethodTestResult"
+              },
+              "DirectMethodName": {
+                "value": "ping"
+              },
+              "InvocationSource": {
+                "value": "Cloud"
+              },
+              "ServiceClientConnectionString": {
```
philipktlin: Rename it to IOT_HUB_CONNECTION_STRING.

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -143,6 +143,8 @@ List<IReportMetadata> InitializeReportMetadataList()
                 new CountingReportMetadata("relayer2.send", "relayer2.eventHub", TestOperationResultType.Messages, TestReportType.CountingReport),
                 new CountingReportMetadata("directMethodSender1.send", "directMethodReceiver1.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
```
philipktlin: please comment out the other 2 direct method reports as well.

e2e_deployment_files/connectivity_deployment.template.json
yophilav: The `deploymentTest1` is the only module that uses `IOT_HUB_CONNECTION_STRING` in Connectivity Test. Would you like me to change that one instead?

test/modules/TestResultCoordinator/Settings.cs
yophilav: yessir

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -141,8 +141,10 @@ List<IReportMetadata> InitializeReportMetadataList()
                 new CountingReportMetadata("relayer1.send", "relayer1.eventHub", TestOperationResultType.Messages, TestReportType.CountingReport),
                 new CountingReportMetadata("loadGen2.send", "relayer2.receive", TestOperationResultType.Messages, TestReportType.CountingReport),
                 new CountingReportMetadata("relayer2.send", "relayer2.eventHub", TestOperationResultType.Messages, TestReportType.CountingReport),
-                new CountingReportMetadata("directMethodSender1.send", "directMethodReceiver1.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
-                new CountingReportMetadata("directMethodSender2.send", "directMethodReceiver2.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
+                // TODO: Enable Direct Method Cloud-to-Module and Cloud-to-EdgeAgent once the verification scheme is finallized.
```
philipktlin: typo
```suggestion
                // TODO: Enable Direct Method Cloud-to-Module and Cloud-to-EdgeAgent once the verification scheme is finalized.
```

test/modules/TestResultCoordinator/Settings.cs
dylanbronson: Ping for EdgeAgent test won't need two stores - it can have only one. I think we should either make a new DirectMethodReport type for this case, or make one of the stores Optional and generate a different report depending on if the option is filled in or not.

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -119,7 +119,7 @@
               "InvocationSource": {
                 "value": "Cloud"
               },
-              "ServiceClientConnectionString": {
+              "IOT_HUB_CONNECTION_STRING": {
```
varunpuranik: Why this change?

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -222,6 +222,46 @@
               "createOptions": ""
             }
           },
+          "directMethodSender3": {
```
varunpuranik: Maybe we should call this edgeAgentDirectMethodSender or something like that?

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -141,8 +141,10 @@ List<IReportMetadata> InitializeReportMetadataList()
                 new CountingReportMetadata("relayer1.send", "relayer1.eventHub", TestOperationResultType.Messages, TestReportType.CountingReport),
                 new CountingReportMetadata("loadGen2.send", "relayer2.receive", TestOperationResultType.Messages, TestReportType.CountingReport),
                 new CountingReportMetadata("relayer2.send", "relayer2.eventHub", TestOperationResultType.Messages, TestReportType.CountingReport),
-                new CountingReportMetadata("directMethodSender1.send", "directMethodReceiver1.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
-                new CountingReportMetadata("directMethodSender2.send", "directMethodReceiver2.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
+                // TODO: Enable Direct Method Cloud-to-Module and Cloud-to-EdgeAgent once the verification scheme is finalized.
```
varunpuranik: Can you clarify why we need to comment this out?

yophilav: The verification & report generating part is not there yet. 
It has a dependency on this PR: https://github.com/Azure/iotedge/pull/2300/files

e2e_deployment_files/connectivity_deployment.template.json
yophilav: `ServiceClientConnectionString` is a variable used in Long Haul test. According to Philip's comment (https://github.com/Azure/iotedge/pull/2286#discussion_r365924786), he would like to use the variable `IOT_HUB_CONNECTION_STRING` in connectivity test instead.

e2e_deployment_files/connectivity_deployment.template.json
yophilav: we could do `edgeAgentDirectMethodSender` but we decided to do the enumeration for the test module instead (at least for now).

test/modules/TestResultCoordinator/Settings.cs
philipktlin: It is better to comment it out since these counting report should not be used for DM scenario which created noise in test report result.

e2e_deployment_files/connectivity_deployment.template.json
philipktlin: I think Varun's suggestion is doable, can you update in this PR?

PR #2287, opened 2020-01-10
Tests clean up

philipktlin:
1. make environment variable readonly in DeploymentTestResult
2. use internal access modifier for class if possible

### DIFF
--- a/test/modules/DeploymentTester/Settings.cs
+++ b/test/modules/DeploymentTester/Settings.cs
@@ -8,7 +8,7 @@ namespace DeploymentTester
     using Microsoft.Azure.Devices.Edge.Util;
     using Microsoft.Extensions.Configuration;
 
-    public class Settings
+    class Settings
     {
         public const string EnvironmentVariablePrefix = "IOTEDGE_DT";
 
@@ -66,7 +66,7 @@ namespace DeploymentTester
             this.DeploymentUpdatePeriod = deploymentUpdatePeriod;
         }
 
-        public static Settings Current => DefaultSettings.Value;
+        internal static Settings Current => DefaultSettings.Value;
 
         public string DeviceId { get; }
 
--- a/test/modules/ModuleLib/TestResults/DeploymentTestResult.cs
+++ b/test/modules/ModuleLib/TestResults/DeploymentTestResult.cs
@@ -11,6 +11,7 @@ namespace Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults
         public DeploymentTestResult(string trackingId, string source, Dictionary<string, string> environmentVariables, DateTime createdAt)
             : base(source, TestOperationResultType.Deployment, createdAt)
         {
+            this.TrackingId = trackingId;
             this.EnvironmentVariables = Preconditions.CheckNotNull(environmentVariables, nameof(environmentVariables));
         }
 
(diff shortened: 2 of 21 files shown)


test/modules/DeploymentTester/Settings.cs
```diff
@@ -66,7 +66,7 @@ public class Settings
             this.DeploymentUpdatePeriod = deploymentUpdatePeriod;
         }
 
-        public static Settings Current => DefaultSettings.Value;
+        internal static Settings Current => DefaultSettings.Value;
 
         public string DeviceId { get; }
```
yophilav: I think `DeviceId` can also be `internal` right? Since it's in the same namespace

test/modules/ModuleLib/TestResults/DeploymentTestResult.cs
```diff
@@ -3,19 +3,20 @@ namespace Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults
 {
     using System;
     using System.Collections.Generic;
+    using Microsoft.Azure.Devices.Edge.Util;
     using Microsoft.Azure.Devices.Edge.Util.Json;
 
     public class DeploymentTestResult : TestResultBase
     {
-        public DeploymentTestResult(string source, DateTime createdAt)
+        public DeploymentTestResult(string trackingId, string source, Dictionary<string, string> environmentVariables, DateTime createdAt)
```
yophilav: `trackingId` is passed in but not used?

test/modules/ModuleRestarter/src/Settings.cs
```diff
@@ -49,9 +47,8 @@ public class Settings
             }
         }
 
-        public static Settings Current => DefaultSettings.Value;
+        internal static Settings Current => DefaultSettings.Value;
 
-        [JsonIgnore]
         public string ServiceClientConnectionString { get; }
```
yophilav: Should these properties be internal as well?

test/modules/TemperatureFilter/MessageBody.cs
```diff
@@ -25,7 +25,7 @@ namespace TemperatureFilter
     ///Humidity: int, %
     ///Pressure: double, psi
     /// </summary>
-    public class MessageBody
+    internal class MessageBody
```
yophilav: `internal` should not be necessary for these classes.

test/modules/DeploymentTester/Settings.cs
philipktlin: all properties below can be internal; since class is internal already; it is not necessary to change these readonly properties to internal.

test/modules/ModuleLib/TestResults/DeploymentTestResult.cs
philipktlin: it is used when calling GetFormattedResult() which will cast this class to json.

test/modules/ModuleRestarter/src/Settings.cs
philipktlin: same explanation as above

test/modules/ModuleLib/TestResults/DeploymentTestResult.cs
yophilav: `trackingId ` is not set in the constructor

philipktlin: good catch

PR #2291, opened 2020-01-10
Dispose all clients when network disconnection

ancaantochi:
When the network disconnects all clients need to be disposed so a new connection is created. The dispose is in a lock so no cloud connections are created before all existing clients are disposed because with AMQP multiplexing the connection is recreated only when there are no other connections.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
```diff
@@ -65,17 +65,8 @@ public async Task<bool> CloseAsync()
...
-                try
-                {
-                    await this.client.CloseAsync();
-                }
-                catch (Exception ex)
-                {
-                    Events.ErrorClosingClient(this.clientId, ex);
-                }
+                // SDK calls CloseAsync when Dispose is called, but it doesn't await on the CloseAsync and no exception is thrown
+                this.client.Dispose();
```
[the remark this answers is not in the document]
ancaantochi: I changed the code so it still calls CloseAsync from CloudProxy and call Dispose in the CloseAsync when SDK is called. So the wrapper layer will have the Open/Close.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceClientWrapper.cs
```diff
@@ -24,9 +24,12 @@ public DeviceClientWrapper(DeviceClient deviceClient)
 
         public Task AbandonAsync(string messageId) => this.underlyingDeviceClient.AbandonAsync(messageId);
 
-        public Task CloseAsync() => this.isActive.GetAndSet(false)
-            ? this.underlyingDeviceClient.CloseAsync()
-            : Task.CompletedTask;
+        public Task CloseAsync()
+        {
+            this.isActive.Set(false);
+            this.underlyingDeviceClient?.Dispose();
```
varunpuranik: Calling this twice might throw right?
Why not do the isAcive.GetAndSet(false) like the code was?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
```diff
@@ -265,12 +266,8 @@ await device.CloudConnection.Filter(cp => cp.IsActive)
...
                     Events.InvokingCloudConnectionLostEvent(device.Identity);
                     this.CloudConnectionLost?.Invoke(this, device.Identity);
-                    await device.CloudConnection.Filter(cp => cp.IsActive).ForEachAsync(
-                        cp =>
-                        {
-                            Events.CloudConnectionLostClosingClient(device.Identity);
-                            return cp.CloseAsync();
-                        });
+                    // Closing all connections so new connections are created a network disconnection
+                    await this.CloseAllCloudConnections();
```
varunpuranik: Is this correct? This is the callback for each individual connection, right?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
```diff
@@ -280,6 +277,23 @@ await device.CloudConnection.Filter(cp => cp.IsActive)
             }
         }
 
+        async Task CloseAllCloudConnections()
+        {
+            using (await this.connectToCloudLock.LockAsync())
```
varunpuranik: We should log that we are closing all connections because device is disconnected

varunpuranik: You need changed to disconnected state callback

philipktlin: this lock is used to protect which resources/properties? this.devices?  If so, there are some other usages of this.devices in this class, should locking be required for other usages?

philipktlin: Also, what deviceConnLock is used for? it looks like it also protect access to this.devices.

ancaantochi: The lock protects that there is no new connection open to the cloud while a disconnect from cloud is handled (closing all cloud connection). This is because of how the sdk works with AMQP multiplexing, in order to drop the tcp connection and create a new one there should be no existing open connection. 
deviceConnLock protects that there are no more ConnectedDevice created at the same time so it would exceed the limit, but it doesn't connect to cloud.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceClientWrapper.cs
ancaantochi: Calling Dispose twice doesn't throw, but I changed back to GetAndSet just to be more safe

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
ancaantochi: Yes, this is the callback for each connection. I added IDeviceConnectivityManager to Connectionmanager and subscribe to DeviceDisconnected

PR #2292, opened 2020-01-10
Update C# KubernetesClient

darobs:
Update the KubernetesClient nuget package for C# to 1.6.10.
Make changes to reflect updated Watch.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentOperator.cs
```diff
@@ -60,7 +60,7 @@ async Task OnListEdgeDeploymentsCompleted(Task<HttpOperationResponse<object>> ta
             HttpOperationResponse<object> response = await task;
 
             this.operatorWatch = Option.Some(
-                response.Watch<EdgeDeploymentDefinition>(
+                response.Watch<EdgeDeploymentDefinition, object>(
```
dmolokanov: Weird. Could you look it up and put real type by any chance?

darobs: I did attempt to create an EdgeDeploymentDefinitionList, but using that in the watch failed.

The Watch signature is:
```
Watcher<T> Watch<T, L>(this Task<HttpOperationResponse<L>> responseTask,
            Action<WatchEventType, T> onEvent,
            Action<Exception> onError = null,
            Action onClosed = null)
```

"L" needs to match the response type. In this case, it's `Task<HttpOperationResponse<object>> ListNamespacedCustomObjectWithHttpMessagesAsync()` so `object` is what it needs to be.

PR #2294, opened 2020-01-10
Add metrics for direct methods

and-rewsmith:
This PR adds metrics for direct methods, a timer and a counter.

I also address the wrong id ("callee id" vs correct "caller id") being passed to the RoutingEdgeHub InvokeMethodAsync() function in direct method scenarios coming from the cloud.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/cloud/CloudListener.cs
```diff
@@ -15,7 +15,7 @@ public CloudListener(IEdgeHub edgeHub, string clientId)
             this.clientId = Preconditions.CheckNotNull(clientId, nameof(clientId));
         }
 
-        public Task<DirectMethodResponse> CallMethodAsync(DirectMethodRequest request) => this.edgeHub.InvokeMethodAsync(this.clientId, request);
+        public Task<DirectMethodResponse> CallMethodAsync(DirectMethodRequest request) => this.edgeHub.InvokeMethodAsync("$upstream", request);
```
and-rewsmith: The first argument to InvokeMethodAsync should be the caller id. Cloud Listener would previously report this incorrectly, as this.clientId is the name of the callee device or module.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
```diff
@@ -368,9 +368,13 @@ internal async Task<MethodResponse> MethodCallHandler(MethodRequest methodreques
...
-                DirectMethodResponse directMethodResponse = await this.cloudListener.CallMethodAsync(direceMethodRequest);
-                MethodResponse methodResponse = directMethodResponse.Data == null ? new MethodResponse(directMethodResponse.Status) : new MethodResponse(directMethodResponse.Data, directMethodResponse.Status);
-                return methodResponse;
+
+                using (Metrics.TimeDirectMethod("$upstream", this.cloudProxy.clientId))
+                {
+                    DirectMethodResponse directMethodResponse = await this.cloudListener.CallMethodAsync(direceMethodRequest);
+                    MethodResponse methodResponse = directMethodResponse.Data == null ? new MethodResponse(directMethodResponse.Status) : new MethodResponse(directMethodResponse.Data, directMethodResponse.Status);
+                    return methodResponse;
+                }
```
and-rewsmith: This timer needs to be duplicated in the TwinController because the common call to RoutingEdgeHub.InvokeMethodAsync() does not await the relevant Task.

and-rewsmith:
I need to fix the failing checkin

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
```diff
@@ -368,9 +368,13 @@ internal async Task<MethodResponse> MethodCallHandler(MethodRequest methodreques
 
                 Events.MethodCallReceived(this.cloudProxy.clientId);
                 var direceMethodRequest = new DirectMethodRequest(this.cloudProxy.clientId, methodrequest.Name, methodrequest.Data, DeviceMethodMaxResponseTimeout);
-                DirectMethodResponse directMethodResponse = await this.cloudListener.CallMethodAsync(direceMethodRequest);
-                MethodResponse methodResponse = directMethodResponse.Data == null ? new MethodResponse(directMethodResponse.Status) : new MethodResponse(directMethodResponse.Data, directMethodResponse.Status);
-                return methodResponse;
+
+                using (Metrics.TimeDirectMethod("$upstream", this.cloudProxy.clientId))
```
varunpuranik: Nit: Move the const to the method, and change it to upstream pls. 
We should make sure to use the same string for upstream (upstream/cloud/IoTHub) for all metrics..

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/cloud/CloudListener.cs
varunpuranik: Nit: upstream

PR #2296, opened 2020-01-11
Remove device descriptors from cloud proxy logging

and-rewsmith:
Currently, cloud proxy prefixes its logs with "for device [CloudProxy.ClientId]". In some cases this is not true (i.e. might apply to modules as well). 

Since there is already precedent in the logs for removing device/module descriptors I will remove.


and-rewsmith:
If we want to have descriptors I can add them back specifying "device or module" where appropriate.

PR #2300, opened 2020-01-13
DirectMethodReport generation for Connectivity Tests

dylanbronson:
Adding the ability to generate DirectMethodReports for connectivity tests. 

The report will contain the results of the direct method test. 
Currently we have eight buckets for results based on what occurred during the test.
For example, if a DirectMethod succeeds while the network was online, networkOnSuccess++.
If a DirectMethod fails while the network was online, networkOnFailure++.
If a DirectMethod fails when the network was offline, however, that is okay and networkOffSuccess++.


test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
```diff
@@ -16,7 +18,7 @@ public TestReportGeneratorFactory(ITestOperationResultStorage storage)
             this.storage = Preconditions.CheckNotNull(storage, nameof(storage));
         }
 
-        public ITestResultReportGenerator Create(
+        public async Task<ITestResultReportGenerator> Create(
             string trackingId,
             IReportMetadata reportMetadata)
         {
```
yophilav:
```
Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
Preconditions.CheckNotNull(reportMetadata);
```

test/modules/TestResultCoordinator/Reports/ITestResultCollection.cs
```diff
@@ -5,7 +5,7 @@ namespace TestResultCoordinator.Reports
     using System.Threading.Tasks;
 
     // TODO: Replace ITestResultCollection<T> with IAsyncEnumerable<T> when project is moved to .Net Core 3.1
-    interface ITestResultCollection<T> : IDisposable
+    public interface ITestResultCollection<T> : IDisposable
```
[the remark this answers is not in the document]
dylanbronson: I put this in for testing... It is actually already changed in master to be public

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
```diff
@@ -0,0 +1,507 @@
...
+    using Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults;
+    using Microsoft.Azure.Devices.Edge.Storage;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Moq;
+    using Newtonsoft.Json;
+    using Xunit;
+
+    public class DirectMethodReportGeneratorTest
+    {
+        public static IEnumerable<object[]> GetCreateReportData =>
```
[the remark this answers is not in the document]
dylanbronson: 1. I will put a comment there to make it more clear. (For now, you can see what they represent by going to the method that uses GetCreateReportData and seeing the parameters)

2. The DMReportGenerator assumes that the SequentialStores (and, in turn, the TestResultCollections) are in order. This is very similar to how CountingReportGenerator currently is. We match up on the sequence number, but expect the Expected sequence number to always increment by 1. 

I could add a check to make sure that the Expected sequence number always increments by 1, though.

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
```diff
@@ -0,0 +1,507 @@
...
+                        new DateTime(2020, 1, 1, 9, 10, 24, 10),
+                        new DateTime(2020, 1, 1, 9, 10, 24, 15)
+                    },
+                    10, 7, 0, 0, 0, 0, 0, 0, 0, true
+                },
+                new object[]
+                {
+                    Enumerable.Range(1, 7).Select(v => v.ToString()),
+                    new[] { "1", "2", "4", "5", "6", "7" },
+                    new int[] { 200, 200, 500, 200, 200, 200, 200 },
```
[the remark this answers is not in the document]
dylanbronson: 1. Can definitely add those
2. If I'm not mistaken, the status code of the DirectMethod Result can only be 200 or 500, since we define it in the DirectMethodSender. So we only have to worry about these two codes.

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
```diff
@@ -0,0 +1,507 @@
...
+            };
+
+        static Option<NetworkStatusTimeline> NetworkStatusTimeline => MockNetworkStatusTimeline.GetMock(new TimeSpan(0, 0, 0, 0, 5)).Result;
+
+        [Fact]
+        public void TestConstructorSuccess()
+        {
+            string expectedSource = "expectedSource";
+            string actualSource = "actualSource";
+            int batchSize = 10;
```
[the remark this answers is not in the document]
dylanbronson: In my opinion, unit tests should be exactly reproducible. If we wanted to add what you are talking about, that is more like fuzz testing. It could definitely have its place, but I don't think in this PR. But I am also interested to hear what others think

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
```diff
@@ -0,0 +1,507 @@
...
+            Assert.Equal(expectedSource, reportGenerator.ExpectedSource);
+            Assert.Equal(expectedResults, reportGenerator.ExpectedTestResults);
+            Assert.Equal(resultType, reportGenerator.ResultType);
+            Assert.Equal(typeof(DirectMethodTestOperationResultComparer), reportGenerator.TestResultComparer.GetType());
+        }
+
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        public void TestConstructorThrowsWhenTrackingIdIsNotProvided(string trackingId)
```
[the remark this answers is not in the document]
dylanbronson: I agree. I'll remove unless anyone thinks otherwise. I was following CountingReportGeneratorTest, so maybe there is a good reason

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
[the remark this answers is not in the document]
philipktlin: Unit tests should be exactly reproduced.  Sometimes, you may see something like Args.Any<int> in some unit tests for some arguments which are not important for that unit test.  We can discuss fuzz testing in team meeting or retrospective meeting.

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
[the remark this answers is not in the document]
philipktlin: We would like to have test coverage for every single line of code.  This test is verify constructor will throw exception if tracking id is not provided.  There are 2 ways to verify, 1) mock preconditions and verify precondition method get called. 2) Verify that argument exception throws.  We took 2nd option because it is easier to implement.

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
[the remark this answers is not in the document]
philipktlin: For #1, I think adding comment to refer to TestCreateReportAsync method should be enough.  I expected devs should have xunit knowledge to find out what arguments are from TestCreateReportAsync method.

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/MockNetworkStatusTimeline.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        static string[] NetworkControllerOperationArray => new[] { "SettingRule", "RuleSet", "SettingRule", "RuleSet", "SettingRule", "RuleSet", "SettingRule", "RuleSet" };
+
+        static async Task<Option<NetworkStatusTimeline>> GetNetworkStatusTimeline(StoreTestResultCollection<TestOperationResult> results, TimeSpan tolerancePeriod)
+        {
+            try
+            {
+                return Option.Some(await NetworkStatusTimeline.Create(results, tolerancePeriod));
+            }
+            catch (Exception)
```
philipktlin: GetNetworkStatusTimeline is private, you manage all usages.  why do you need to catch exception?

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReport.cs
```diff
@@ -0,0 +1,55 @@
...
+            this.NetworkOffSuccess = networkOffSuccess;
+            this.NetworkOnToleratedSuccess = networkOnToleratedSuccess;
+            this.NetworkOffToleratedSuccess = networkOffToleratedSuccess;
+            this.NetworkOnFailure = networkOnFailure;
+            this.NetworkOffFailure = networkOffFailure;
+            this.MismatchSuccess = mismatchSuccess;
+            this.MismatchFailure = mismatchFailure;
+        }
+
+        public ulong NetworkOnSuccess { get; }
```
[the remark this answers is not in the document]
philipktlin: I think so.  I forgot what does mismatchSuccess mean? :P

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+            Option<NetworkStatusTimeline> networkStatusTimeline)
+        {
+            this.trackingId = Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
+            this.ExpectedTestResults = Preconditions.CheckNotNull(expectedTestResults, nameof(expectedTestResults));
+            this.ExpectedSource = Preconditions.CheckNonWhiteSpace(expectedSource, nameof(expectedSource));
+            this.ActualSource = Preconditions.CheckNonWhiteSpace(actualSource, nameof(actualSource));
+            this.ActualTestResults = Preconditions.CheckNotNull(actualTestResults, nameof(actualTestResults));
+            this.TestResultComparer = Preconditions.CheckNotNull(testResultComparer, nameof(testResultComparer));
+            this.ResultType = Preconditions.CheckNonWhiteSpace(resultType, nameof(resultType));
+            this.NetworkStatusTimeline = Preconditions.CheckNotNull(networkStatusTimeline, nameof(networkStatusTimeline));
```
philipktlin: Option is struct, no need to call PreConditions

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+            string expectedSource,
+            ITestResultCollection<TestOperationResult> expectedTestResults,
+            string actualSource,
+            ITestResultCollection<TestOperationResult> actualTestResults,
+            string resultType,
+            ITestResultComparer<TestOperationResult> testResultComparer,
+            Option<NetworkStatusTimeline> networkStatusTimeline)
+        {
+            this.trackingId = Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
+            this.ExpectedTestResults = Preconditions.CheckNotNull(expectedTestResults, nameof(expectedTestResults));
```
philipktlin: nit:have same order as arguments.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+    using Newtonsoft.Json;
+    using TestResultCoordinator.Reports;
+
+    class DirectMethodReportGenerator : ITestResultReportGenerator
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger(nameof(DirectMethodReportGenerator));
+
+        readonly string trackingId;
+
+        public DirectMethodReportGenerator(
```
philipktlin: please use internal here; I will have another PR to change all other report generator's constructor to use internal.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil.NetworkController;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+    using TestResultCoordinator.Reports;
+
+    class DirectMethodReportGenerator : ITestResultReportGenerator
```
philipktlin: use sealed

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+            bool hasExpectedResult = await this.ExpectedTestResults.MoveNextAsync();
+            bool hasActualResult = await this.ActualTestResults.MoveNextAsync();
+
+            while (hasExpectedResult && hasActualResult)
+            {
+                this.ValidateDataSource(this.ExpectedTestResults.Current, this.ExpectedSource);
+                this.ValidateDataSource(this.ActualTestResults.Current, this.ActualSource);
+
+                (NetworkControllerStatus networkControllerStatus, bool isWithinTolerancePeriod) =
+                        this.NetworkStatusTimeline.Expect<InvalidOperationException>(
```
philipktlin: do you expect NetworkStatusTimeline always exists? if so, why not enforce in constructor?

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+                        if (isWithinTolerancePeriod)
+                        {
+                            networkOffToleratedSuccess++;
+                        }
+                        else
+                        {
+                            networkOffFailure++;
+                        }
+                    }
+                    else
```
philipktlin: throw exception right after GetNetworkControllerStatusAndWithinToleranceAt call if status is not enabled or disabled.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+                    else if (NetworkControllerStatus.Enabled.Equals(networkControllerStatus))
+                    {
+                        // If network off at time, fail unless in tolerance period
+                        if (isWithinTolerancePeriod)
+                        {
+                            networkOffToleratedSuccess++;
+                        }
+                        else
+                        {
+                            networkOffFailure++;
```
philipktlin: network off failure means network is offline and DM call failed; which is positive result.  However this line cannot be reached as for failure scenario, DM receiver will not report that sequence number to TRC, therefore if block on line 82 is never true for failure scenario.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        internal ITestResultComparer<TestOperationResult> TestResultComparer { get; }
+
+        internal Option<NetworkStatusTimeline> NetworkStatusTimeline { get; }
+
+        public async Task<ITestResultReport> CreateReportAsync()
+        {
+            Logger.LogInformation($"Start to generate report by {nameof(DirectMethodReportGenerator)} for Sources [{this.ExpectedSource}] and [{this.ActualSource}]");
+
+            ulong networkOnSuccess = 0;
```
philipktlin: wrap all these counts into a private class/struct; and using it will help to make code more readable.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+                    {
+                        // If network on at time, succeed
+                        networkOnSuccess++;
+                    }
+                    else if (NetworkControllerStatus.Enabled.Equals(networkControllerStatus))
+                    {
+                        // If network off at time, fail unless in tolerance period
+                        if (isWithinTolerancePeriod)
+                        {
+                            networkOffToleratedSuccess++;
```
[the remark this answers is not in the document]
dylanbronson: I don't think so, since, in a way, the whole report is a log of what happened. But this is just report generation based on events that have already happened, so I don't think logging is as useful here

test/modules/TestResultCoordinator/Reports/ITestReportGeneratorFactory.cs
```diff
@@ -1,12 +1,13 @@
...
 namespace TestResultCoordinator.Reports
 {
+    using System.Threading.Tasks;
     /// <summary>
     /// This is used to create an instance of test report generator based on given report-related parameters.
     /// </summary>
     interface ITestReportGeneratorFactory
     {
-        ITestResultReportGenerator Create(
+        Task<ITestResultReportGenerator> Create(
```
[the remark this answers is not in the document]
dylanbronson: I'll change the name to be CreateAsync

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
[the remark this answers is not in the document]
yophilav: @philipktlin I see where you're coming from. But doesn't that implies that we should also verify the constructor correctly set the correct input argument to the right variable? 

@dylanbronson  Per Phililps' comment, I think it would make sense to do `[InlineData(null)]` and `[InlineData("cash me outside how bout dat")]`? What do you think? I mean we are not really want to test if `Precontidion.*` correctly catch both `null` and `""` - that is done. We want to test more in lue of the constructor if it can handle input argument correctly.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
dylanbronson: I defined networkOffFailure as "if the network is off and the DM call succeeds but is not in the tolerance period" which is a weird (and negative) result case

networkOffSuccess is a positive result  that means the network is off and the DM call failed

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,214 @@
...
+            TestOperationResult testOperationResult,
+            NetworkControllerStatus networkControllerStatus,
+            bool isWithinTolerancePeriod,
+            ulong networkOffSuccess,
+            ulong networkOnToleratedSuccess,
+            ulong networkOnFailure,
+            ulong mismatchSuccess)
+        {
+            DirectMethodTestResult dmTestResult = JsonConvert.DeserializeObject<DirectMethodTestResult>(testOperationResult.Result);
+            HttpStatusCode statusCode = JsonConvert.DeserializeObject<HttpStatusCode>(dmTestResult.Result);
```
philipktlin: convert it to int and use "YourEnum foo = (YourEnum)yourInt;"

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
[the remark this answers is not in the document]
yophilav: 2. For your reference:  https://github.com/Azure/iotedge/blob/master/test/modules/DirectMethodSender/DirectMethodSenderBase.cs#L37-L58
The logic is basically when there is an exception, the HTTP500 will be thrown; otherwise, report whatever HTTP status the SDK team returns (either moduleClient or serviceClient).

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportMetadata.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public TimeSpan TolerancePeriod { get; }
+
+        public string ExpectedSource { get; }
+
+        public string ActualSource { get; }
+
+        public TestReportType TestReportType { get; }
+
+        public TestOperationResultType TestOperationResultType { get; }
```
philipktlin: With below suggestion, you can remove line 14.
```suggestion
        public TestOperationResultType TestOperationResultType => TestOperationResultType.DirectMethod;
```

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportMetadata.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator.Reports.DirectMethod
+{
+    using System;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using TestResultCoordinator.Reports;
+
+    class DirectMethodReportMetadata : IReportMetadata
```
philipktlin: extends from TestResultReportBase

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodTestOperationResultComparer.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            if (!TestOperationResultType.DirectMethod.ToString().Equals(value1.Type) || !TestOperationResultType.DirectMethod.ToString().Equals(value2.Type))
+            {
+                throw new InvalidDataException($"Incorrect TestOperationResult Type for comparer {nameof(DirectMethodTestOperationResultComparer)}. Types are: {value1.Type} and {value2.Type}");
+            }
+
+            DirectMethodTestResult dmtr1 = JsonConvert.DeserializeObject<DirectMethodTestResult>(value1.Result);
+            DirectMethodTestResult dmtr2 = JsonConvert.DeserializeObject<DirectMethodTestResult>(value2.Result);
+
+            return string.Equals(value1.Type, value2.Type, StringComparison.OrdinalIgnoreCase) &&
```
philipktlin: don't need this line which is already covered in line 28.

philipktlin: you need to compare trackingId, batchId, sequenceNumber and result.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodTestOperationResultComparer.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                return true;
+            }
+
+            if (!TestOperationResultType.DirectMethod.ToString().Equals(value1.Type) || !TestOperationResultType.DirectMethod.ToString().Equals(value2.Type))
+            {
+                throw new InvalidDataException($"Incorrect TestOperationResult Type for comparer {nameof(DirectMethodTestOperationResultComparer)}. Types are: {value1.Type} and {value2.Type}");
+            }
+
+            DirectMethodTestResult dmtr1 = JsonConvert.DeserializeObject<DirectMethodTestResult>(value1.Result);
```
philipktlin: I think you don't need to deserialize it, just use string.Equals; check out the implementation of GetFormattedResult in DirectMethodTestResult class.

test/modules/TestResultCoordinator/Reports/ITestResultCollection.cs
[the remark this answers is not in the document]
philipktlin: public is needed for interface in order to be able to use in Modules.Test project.

test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
```diff
@@ -65,13 +67,46 @@ public TestReportGeneratorFactory(ITestOperationResultStorage storage)
...
                 }
             }
         }
 
+        async Task<Option<NetworkStatusTimeline>> GetNetworkStatusTimeline(TimeSpan tolerancePeriod)
+        {
+            try
+            {
+                return Option.Some(
+                    await NetworkStatusTimeline.Create(
```
philipktlin: rename Create method to CreateAsync method.

test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
```diff
@@ -65,13 +67,46 @@ public TestReportGeneratorFactory(ITestOperationResultStorage storage)
...
+                }
+
                 default:
                 {
                     throw new NotSupportedException($"Report type {reportMetadata.TestReportType} is not supported.");
                 }
             }
         }
 
+        async Task<Option<NetworkStatusTimeline>> GetNetworkStatusTimeline(TimeSpan tolerancePeriod)
```
philipktlin: Rename GetNetworkStatusTimeline method to GetNetworkStatusTimelineAsync method

test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
```diff
@@ -16,7 +18,7 @@ public TestReportGeneratorFactory(ITestOperationResultStorage storage)
             this.storage = Preconditions.CheckNotNull(storage, nameof(storage));
         }
 
-        public ITestResultReportGenerator Create(
+        public async Task<ITestResultReportGenerator> Create(
```
philipktlin: rename to CreateAsync

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/MockNetworkStatusTimeline.cs
dylanbronson: Remnant from moving it around. Removed the try catch

philipktlin:
update settings class in TRC project for this new report.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
dylanbronson: My thinking is that we should be able to generate an empty report even if there is no timeline. Therefore, no timeline is allowed for that one case.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
dylanbronson: I'm going to keep it like this for now. Currently, the test result won't be deserialized properly, because it's not being serialized from the DMSender, but I think it should be serialized from the DMSender. (currently it's just a string with semicolon delimiters, but that is less desirable than a serialized json)

The DM connectivity test won't work until that is in place, but I think we should keep the Generator and its tests as is and fix the DMSender

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodTestOperationResultComparer.cs
dylanbronson: Same as above comment for this

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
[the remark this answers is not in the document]
dylanbronson: @yophilav Do you mean we should test:
1. Null case
2. Case where argument is given correctly?
(and not 3. "")

Currently, we test null, empty string, and we test all the arguments get correctly set in the test named "TestConstructorSuccess." Does that cover what you are asking?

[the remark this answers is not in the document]
yophilav: yeah.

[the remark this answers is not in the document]
yophilav: Oh I see. There are 2 cases doing somewhat similar testing. It's fine then.

test/modules/TestResultCoordinator/Reports/ITestResultCollection.cs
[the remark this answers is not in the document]
yophilav: Understood.

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGeneratorTest.cs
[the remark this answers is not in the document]
yophilav: Gotcha.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReport.cs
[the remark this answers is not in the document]
dylanbronson: Added

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
dylanbronson: Talked offline - making it non-optional. The default is set to restriction disabled (aka online), so if there are no network controller results passed to TRC, the network will always be assumed to be online.

test/modules/TestResultCoordinator/Reports/TestReportHelper.cs
```diff
@@ -19,7 +19,7 @@ internal static async Task<ITestResultReport[]> GenerateTestResultReports(ITestO
                 var testResultReportList = new List<Task<ITestResultReport>>();
                 foreach (IReportMetadata reportMetadata in Settings.Current.ReportMetadataList)
                 {
-                    ITestResultReportGenerator testResultReportGenerator = testReportGeneratorFactory.Create(Settings.Current.TrackingId, reportMetadata);
+                    ITestResultReportGenerator testResultReportGenerator = await testReportGeneratorFactory.CreateAsync(Settings.Current.TrackingId, reportMetadata);
```
yophilav: `GenerateTestResultReports` -> `GenerateTestResultReportsAsync`

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -147,7 +148,7 @@ List<IReportMetadata> InitializeReportMetadataList()
                 new TwinCountingReportMetadata("twinTester2.reportedReceived", "twinTester2.reportedUpdated", TestReportType.TwinCountingReport, TwinTestPropertyType.Reported),
                 new TwinCountingReportMetadata("twinTester3.desiredUpdated", "twinTester4.desiredReceived", TestReportType.TwinCountingReport, TwinTestPropertyType.Desired),
                 new TwinCountingReportMetadata("twinTester4.reportedReceived", "twinTester4.reportedUpdated", TestReportType.TwinCountingReport, TwinTestPropertyType.Reported),
-                new DeploymentTestReportMetadata("deploymentTester1.send",  "deploymentTester2.receive")
+                new DeploymentTestReportMetadata("deploymentTester1.send",  "deploymentTester2.receive"),
```
yophilav: typo `,` at the end of the line.

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -135,18 +135,26 @@ internal List<ITestReportMetadata> GetReportMetadataList()
...
+                return new List<ITestReportMetadata>
+                {
+                    new CountingReportMetadata("loadGen1.send", "relayer1.receive", TestOperationResultType.Messages, TestReportType.CountingReport),
+                    new CountingReportMetadata("relayer1.send", "relayer1.eventHub", TestOperationResultType.Messages, TestReportType.CountingReport),
+                    new CountingReportMetadata("loadGen2.send", "relayer2.receive", TestOperationResultType.Messages, TestReportType.CountingReport),
+                    new CountingReportMetadata("relayer2.send", "relayer2.eventHub", TestOperationResultType.Messages, TestReportType.CountingReport),
+                // TODO: Enable Direct Method Cloud-to-Module and Cloud-to-EdgeAgent once the verification scheme is finalized.
+                // new CountingReportMetadata("directMethodSender1.send", "directMethodReceiver1.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
+                // new CountingReportMetadata("directMethodSender2.send", "directMethodReceiver2.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
```
yophilav: With changes within this PR, I think 
```
new CountingReportMetadata("directMethodSender1.send", "directMethodReceiver1.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
new CountingReportMetadata("directMethodSender2.send", "directMethodReceiver2.receive", TestOperationResultType.DirectMethod, TestReportType.CountingReport),
```
might be ready to be enabled?  They are cloud-to-module connectivity test.

Issue #2279, opened 2020-01-09
Ubuntu iotedge debian (deb) package mishandles configuration file installation


gri6507:
I believe there is an issue in the current debian package for iotedge. I tested this on Ubuntu 18.04, but I am almost certain that other debian based distros are equally affected.

When installing `iotedge` package, dpkg installs the `/etc/iotedge/config.yaml` configuration file. Then, the `iotedge.postinst` debian maintainer script (which can be viewed in `/var/lib/dpkg/info/iotedge.postinst` on a system that already had the `iotedge` package installed) modifies the  `/etc/iotedge/config.yaml` configuration file **in place**. Such in place modification is not ideal, but not really a problem at this point.

However, imagine that at some point in the future, a new version of `iotedge` package is released **and that new version has some differences to the default  `/etc/iotedge/config.yaml` file**. When upgrading from the old version of `iotedge` package to the new version, dpkg will notice that the default config file is different and the `/etc/iotedge/config.yaml` is not the old or new default version. This causes dpkg to go into interactive mode (even if export DEBIAN_FRONTEND=noninteractive), printing out something like

```
Configuration file '/etc/configtest/configtest.conf'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** configtest.conf (Y/I/N/O/D/Z) [default=N] ?
```

I believe there are actually two problems here:

1. Instead of modifying `/etc/iotedge/config.yaml` configuration file **in place** in `iotedge.postinst`, the debian package should instead be leveraging `debconf`. See http://manpages.ubuntu.com/manpages/precise/man7/debconf-devel.7.html
and http://manpages.ubuntu.com/manpages/bionic/man8/dpkg-reconfigure.8.html for more information
[...]

PR #2301, opened 2020-01-13
Issue #2279 - Allow for clean updates of iotedge deb packages

gri6507:
PROBLEM STATEMENT:
When installing iotedge package, dpkg installs the /etc/iotedge/config.yaml configuration file. Then, the iotedge.postinst debian maintainer script modifies the /etc/iotedge/config.yaml configuration file in place. Then, when iotedge package is updated and that update has a new config.yaml, the upgrade causes dpkg to go into interactive mode (even if export DEBIAN_FRONTEND=noninteractive), asking for what action to take on the conflicting configration file.

SOLUTION:
Per @arsing: "we will never want a new iotedge package to remove the changes that the user made to their config.yaml, even if the new config.yaml has some new defaults. We'll also not want to make targeted replacements to check if the new defaults can be inserted into the config.yaml safely, because this is too much effort to get right. Basically any changes to the defaults will need to be explicitly added by the user." Therefore, when deploying the deb, deliver a **template** of config.yaml which is then copied to config.yaml in postinst IFF that file does not already exist.

DETAILS:
 - Update Makefile to install config.yaml as template
 - Updated debain*/install to deploy config.yaml as template
 - Updated debain*/postinst to create a copy of config.yaml template as config.yaml if that file does not exist. This allows for subsequent in place modification of config.yaml without causing any upgrade problems.

TESTING:
1. Installed 1.0.8-2 from Microsoft maintained Ubuntu package repo/
2. Installed locally built debs, which have a different config.yaml (template) and observed no problems


edgelet/Makefile
```diff
@@ -51,6 +51,8 @@ MAKE=make
 MKDIR_P=mkdir -p
 SED=sed
 HOST_URI_OPTION=
+CONFIG_YAML="config.yaml"
```
arsing: Why do these two need to be variables? They can't have any other value anyway, so they don't need to be configurable.

edgelet/contrib/debian8/postinst
```diff
@@ -10,9 +10,13 @@ case "$1" in
             chown ${USER}:${GROUP} /var/lib/iotedge
             chown ${USER}:${GROUP} /var/log/iotedge
             chown ${USER}:${GROUP} /etc/iotedge/config.yaml
+            if [ ! -f /etc/iotedge/config.yaml ]; then
+              cp -a /etc/iotedge/config.yaml.template /etc/iotedge/config.yaml
```
arsing: Indentation is wrong. The rest of the file uses 4 spaces.

edgelet/contrib/debian/postinst
```diff
@@ -9,10 +9,14 @@ case "$1" in
         if id -u $USER > /dev/null 2>&1; then
             chown ${USER}:${GROUP} /var/lib/iotedge
             chown ${USER}:${GROUP} /var/log/iotedge
-            chown ${USER}:${GROUP} /etc/iotedge/config.yaml
+            chown ${USER}:${GROUP} /etc/iotedge/config.yaml.template
+            if [ ! -f /etc/iotedge/config.yaml ]; then
+              cp -a /etc/iotedge/config.yaml.template /etc/iotedge/config.yaml
```
arsing: Indentation is wrong. The rest of the file uses 4 spaces.

edgelet/Makefile
```diff
@@ -123,7 +125,7 @@ rpm:
 install: release
 	$(INSTALL_PROGRAM) -D $(TARGET)/iotedged $(DESTDIR)$(bindir)/iotedged
 	$(INSTALL_PROGRAM) -D $(TARGET)/iotedge $(DESTDIR)$(bindir)/iotedge
-	$(INSTALL_DATA) -D $(srcdir)/contrib/config/linux/config.yaml $(DESTDIR)$(sysconfdir)/iotedge/config.yaml
+	$(INSTALL_DATA) -D $(srcdir)/contrib/config/linux/$(CONFIG_YAML) $(DESTDIR)$(sysconfdir)/iotedge/$(CONFIG_YAML)
```
arsing: We should do the same thing for deb and rpm. So the rpm should also have the `.yaml.tempate` and  copy it to `.yaml` if that doesn't already exist.

gri6507: Could you help me a bit with RPM changes? I haven't done RPM packaging

edgelet/contrib/debian/postinst
gri6507: fixed

edgelet/contrib/debian8/postinst
gri6507: fixed

edgelet/Makefile
gri6507: If your preference is not to have variables, I can certainly understand that. Reverted to filenames.

edgelet/Makefile
arsing: The equivalent place where you need to insert the "if `config.yaml` doesn't exist then copy `config.yaml.template` to it" line is around https://github.com/Azure/iotedge/blob/a85430e98c8fa436d8578fdcfa6d921f15ef1423/edgelet/contrib/centos/iotedge.spec#L74

To test it, you can build an rpm using `package.sh` with `PACKAGE_OS=centos7`, then install it in a `centos:7` Docker container to verify.

arsing:
Given that the deb no longer knows about the existence of `/etc/iotedge/config.yaml`, does that mean that `apt purge iotedge` will leave it behind?

gri6507:
> Given that the deb no longer knows about the existence of `/etc/iotedge/config.yaml`, does that mean that `apt purge iotedge` will leave it behind?

Currently, yes. My original assumption was that you only want to modify `/etc/iotedge/config.yaml` once at installation time. Reinstallation and Upgrade would certainly honor that. However, I intentionally did not add the removal of `/etc/iotedge/config.yaml` to prerm (remove and deconfigure) with the assumption that you wanted to keep the same "configure once" paradigm even on purging.

After thinking about it some more, I don't that was was the correct choice. I will make a new push which cleans up on deb purge (and RPM uninstall).

edgelet/Makefile
gri6507: > The equivalent place where you need to insert the "if `config.yaml` doesn't exist then copy `config.yaml.template` to it" line is around
> 
> https://github.com/Azure/iotedge/blob/a85430e98c8fa436d8578fdcfa6d921f15ef1423/edgelet/contrib/centos/iotedge.spec#L74
> 
> To test it, you can build an rpm using `package.sh` with `PACKAGE_OS=centos7`, then install it in a `centos:7` Docker container to verify.

While trying to test the RPM changes, during `PACKAGE_OS=centos7 PACKAGE_ARCH=amd64 ./edgelet/build/linux/package.sh`, the build process gave me an error that is unrelated to my changes

```
<snip>
[ 68%] Built target aziotsharedutil
Scanning dependencies of target utpm
[ 69%] [ 71%] [ 72%] [ 73%] [ 75%] Building C object deps/utpm/CMakeFiles/utpm.dir/src/Marshal.c.o
Building C object deps/utpm/CMakeFiles/utpm.dir/src/tpm_codec.c.o
Building C object deps/utpm/CMakeFiles/utpm.dir/src/Memory.c.o
Building C object deps/utpm/CMakeFiles/utpm.dir/src/gbfiledescript.c.o
[ 76%] Building C object deps/utpm/CMakeFiles/utpm.dir/src/tpm_socket_comm.c.o
Building C object deps/utpm/CMakeFiles/utpm.dir/src/tpm_comm_linux.c.o
Linking C static library libutpm.a
[ 76%] Built target utpm
Scanning dependencies of target iothsm
[ 77%] [ 78%] [ 80%] [ 81%] [ 82%] Building C object CMakeFiles/iothsm.dir/src/certificate_info.c.o
Building C object CMakeFiles/iothsm.dir/src/constants.c.o
Building C object CMakeFiles/iothsm.dir/src/edge_hsm_client_crypto.c.o
Building C object CMakeFiles/iothsm.dir/src/edge_enc_openssl_key.c.o
Building C object CMakeFiles/iothsm.dir/src/edge_hsm_client_store.c.o
[ 84%] Building C object CMakeFiles/iothsm.dir/src/edge_hsm_client_x509.c.o
[ 85%] [ 86%] Building C object CMakeFiles/iothsm.dir/src/edge_hsm_key_interface.c.o
Building C object CMakeFiles/iothsm.dir/src/edge_openssl_common.c.o
[ 88%] [ 89%] Building C object CMakeFiles/iothsm.dir/src/edge_sas_perform_sign_with_key.c.o
Building C object CMakeFiles/iothsm.dir/src/edge_pki_openssl.c.o
[ 90%] [ 92%] Building C object CMakeFiles/iothsm.dir/src/edge_sas_key.c.o
[ 93%] [ 94%] [ 96%] Building C object CMakeFiles/iothsm.dir/src/hsm_certificate_props.c.o
Building C object CMakeFiles/iothsm.dir/src/hsm_client_tpm_in_mem.c.o
Building C object CMakeFiles/iothsm.dir/src/hsm_client_data.c.o
Building C object CMakeFiles/iothsm.dir/src/hsm_client_tpm_device.c.o
[ 97%] Building C object CMakeFiles/iothsm.dir/src/hsm_client_tpm_select.c.o
make[2]: *** No rule to make target `/usr/lib/x86_64-linux-gnu/libssl.so', needed by `libiothsm.so.1.0.8'.  Stop.
make[2]: *** Waiting for unfinished jobs....
[ 98%] [100%] Building C object CMakeFiles/iothsm.dir/src/hsm_log.c.o
Building C object CMakeFiles/iothsm.dir/src/hsm_utils.c.o
make[1]: *** [CMakeFiles/iothsm.dir/all] Error 2
make: *** [all] Error 2
```

Is there some other step I need before running package.sh?

arsing: >`/usr/lib/x86_64-linux-gnu/libssl.so`

It's reusing the makefile from an ubuntu build. Make sure to `git clean` to remove the output of the ubuntu build first.

gri6507: > It's reusing the makefile from an ubuntu build. Make sure to `git clean` to remove the output of the ubuntu build first.

Thanks for the clarification. This is something else to add to Rust devguide along with the other information we discussed, which was
- git clone --recurse-submodules -j4 https://github.com/Azure/iotedge.git
- cd iotedge
- PACKAGE_OS=ubuntu18.04 PACKAGE_ARCH=amd64 ./edgelet/build/linux/package.sh

arsing:
>I will make a new push which cleans up on deb purge (and RPM uninstall).

No, uninstalling the RPM should not remove `/etc/iotedge/config.yaml`. The way RPM systems work is that the `.spec` annotates it as a config file with the `%config` directive, so uninstalling the package renames it to `config.yaml.rpmsave` instead of deleting it (so that it can be recovered after the package is installed, without conflicting with a reinstalled package's file). Also, if the `config.yaml` already exists before the package is installed, then the `config.yaml` is left as-is and the new one is saved as `config.yaml.rpmnew` (so that the package can be pre-configured before the package is installed)

That said, because the `config.yaml` needs to be annotated in the `.spec` file, I think you'll be unable to build the package without including that file in the package. Try it. If that is indeed the case, then leave out any changes to the `.spec`, though it'll be unfortunate to have different sets of instructions to bootstrap the daemon post-package-install.

gri6507:
> > I will make a new push which cleans up on deb purge (and RPM uninstall).
> 
> No, uninstalling the RPM should not remove `/etc/iotedge/config.yaml`. The way RPM systems work is that the `.spec` annotates it as a config file with the `%config` directive, so uninstalling the package renames it to `config.yaml.rpmsave` instead of deleting it (so that it can be recovered after the package is installed, without conflicting with a reinstalled package's file). Also, if the `config.yaml` already exists before the package is installed, then the `config.yaml` is left as-is and the new one is saved as `config.yaml.rpmnew` (so that the package can be pre-configured before the package is installed)
> 
> That said, because the `config.yaml` needs to be annotated in the `.spec` file, I think you'll be unable to build the package without including that file in the package. Try it. If that is indeed the case, then leave out any changes to the `.spec`, though it'll be unfortunate to have different sets of instructions to bootstrap the daemon post-package-install.

If this is the case, then I do not see a point in using the same `config.yaml.template` -->  `config.yaml` scheme in RPMs. The YAML file should simply be deployed as is in the RPM. In other words, no changes to the RPM are needed. Did I understand that correctly?

arsing:
There is a point - it simplifies the instructions if they're the same for all distros. Which is why I said:

>>That said, because the config.yaml needs to be annotated in the .spec file, I think you'll be unable to build the package without including that file in the package. **Try it.** If that is indeed the case, then leave out any changes to the .spec, though it'll be unfortunate to have different sets of instructions to bootstrap the daemon post-package-install.

gri6507:
> > > That said, because the config.yaml needs to be annotated in the .spec file, I think you'll be unable to build the package without including that file in the package. **Try it.** If that is indeed the case, then leave out any changes to the .spec, though it'll be unfortunate to have different sets of instructions to bootstrap the daemon post-package-install.
> There is a point - it simplifies the instructions if they're the same for all distros. Which is why I said:
> 

I'm sorry, I think I may not have conveyed my point well. Let me rephrase. If RPM already takes care of config files on install/upgrade/reinstall/uninstall such that there is no conflict as in DEB, then there is no reason to use the "template" scheme in RPMs. The RPM spec file would not change and deliver 
 `/etc/iotedge/config.yaml` as is, having declared it as a %config file. The customer facing instructions for configuring their RPM based system would not change because the config file is still calls config.yaml.

To your point, I did try modifying `edgelet/contrib/centos/iotedge.spec` so that the %config file is listed as `config.yaml` (not `config.yaml.template`) and 

```rpm
%post
if [ ! -f /etc/iotedge/config.yaml ]; then
    cp -a /etc/iotedge/config.yaml.template /etc/iotedge/config.yaml
fi
```

The RPM built successfully and I was able to install it successfully. The installation had `/etc/iotedge/config.yaml` and upon removal, the file was changed to `config.yaml.rpmsave`. So, there doesn't appear to be a need for the "template" file and the overall customer facing configuration instructions do not need to change.

arsing:
Sorry, you're right. The bootstrap instructions will be the same in either case. I was thinking about future documentation that we would write that says "We ship a `config.yaml.template` file that contains the latest defaults. Look at it to discover any new defaults that you want to incorporate into your existing config.yaml". This documentation would end up being different for .deb users and .rpm users.

It's not a big deal to say it's `config.yaml.template` for .deb users and `config.yaml.rpmnew` for .rpm users, so they don't *have* to be the same, though it would be easier for us as well as users if they were.

>To your point, I did try modifying `edgelet/contrib/centos/iotedge.spec` so that the %config file is listed as `config.yaml` (not `config.yaml.template`) and

Actually, I think that'll just end up including both the `config.yaml.template` and `config.yaml` in the package, so...

>```
>%post
>if [ ! -f /etc/iotedge/config.yaml ]; then
>    cp -a /etc/iotedge/config.yaml.template /etc/iotedge/config.yaml
>fi
>```

... this `if` block will never actually run. I think what we would want is to remove `config.yaml` from the .spec file and only have `config.yaml.template` (not marked as `%config` so that it gets replaced like a regular file).

But it seems to be more complicated to do this than I was thinking. Let's just scope the change to the .deb packaging for now. Sorry for the run-around.

edgelet/Makefile
arsing: (Decided against doing this for the rpm package in the main discussion thread.)

gri6507:
@arsing - in that case, since RPM change is out of scope, and with my push a  minute about that cleans up the DEB on purging, I believe this should complete the PullRequest

gri6507:
I wanted to make sure that I'm properly following the PullRequest process. I see that this request still shows "1 change requested" @arsing, but I believe I have addressed all comments. Is that not correct?

Also, how long does a typical PullRequest last before it is reviewed by others and merged?

arsing:
I've approved, but I would like someone else to sign off too. Preferably Mike, but he's out-of-office for a month or so, so either Damon or David (already requested).

edgelet/contrib/debian8/postinst
```diff
@@ -10,9 +10,13 @@ case "$1" in
             chown ${USER}:${GROUP} /var/lib/iotedge
             chown ${USER}:${GROUP} /var/log/iotedge
             chown ${USER}:${GROUP} /etc/iotedge/config.yaml
```
damonbarry: Your `debian/postinst` change replaces this line with a `chown` statement for `config.yaml.template`. Why did you not do the same thing here (for `debian8`)?

edgelet/contrib/debian/prerm
```diff
@@ -2,7 +2,13 @@
 set -e
 
 case "$1" in
-    remove|upgrade|deconfigure)
+    remove|deconfigure)
+        if [ -f /etc/iotedge/config.yaml ]; then
+            rm /etc/iotedge/config.yaml
+        fi
+    ;;
```
damonbarry: It looks like we're changing the behavior here to delete the user's `config.yaml` on uninstall...@arsing is that what we want?

arsing: It ought to be only on `purge`, not on `remove` (ref https://github.com/Azure/iotedge/pull/2301#issuecomment-573898598 )

arsing: @gri6507 Can you confirm that the `rm` only runs on `purge`, not on `remove` ?

gri6507: @arsing  `prerm` does not get called for `purge`. Therefore, it has to be on `remove`. (see https://www.debian.org/doc/debian-policy/ap-flowcharts.html for more info). The problem with moving the deletion to `postrm purge` is that you loose the ability to reconfigure the package because `reconfigure` is not an option for `postrm`. 

Since earlier in this PullRequest we were considering using `DebConf`, the `reconfigure` is critical. However, now that we are no longer using DebConf, I will move back to deleting in `postrm purge` .

edgelet/contrib/debian8/postinst
gri6507: Good catch. That was an error. I missed that test case for Deb8.

gri6507:
Sorry for the delay in pushing the update. Canonical servers seem to be having issues last night and this morning. I finally got the DEBs built so I could test it locally. Please take a look. Thanks!

gri6507:
Just wanted to check in on this PullRequest status.

arsing:
(Waiting for packages build, then manual testing.)

arsing:
Looks fine to me with local testing of the Debian 9 package.

I did notice that `apt remove iotedge` leaves both `/etc/iotedge/config.yaml.template` and `/etc/iotedge/config.yaml` behind, where ideally it would just leave the latter. But I don't think there's a way to avoid this - I guess apt considers all files under `/etc` to be config files since there's no equivalent to RPM's `%config` directive. It's not a big deal either way since `purge` does remove it.

veyalla:
@gri6507 Thank for your contribution and working through the feedback. We appreciate your contribution very much! 🎉

PR #2302, opened 2020-01-13
Update settings in Test modules

philipktlin:
1. Refactor Settings class in test module projects to use static variable for Singleton instead of Lazy.
2. Refactor Settings class in TRC for calling GetTwin to get report metadatas later.

### DIFF
--- a/test/modules/DeploymentTester/Settings.cs
+++ b/test/modules/DeploymentTester/Settings.cs
@@ -10,29 +10,9 @@ namespace DeploymentTester
 
     class Settings
     {
-        public const string EnvironmentVariablePrefix = "IOTEDGE_DT";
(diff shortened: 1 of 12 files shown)


test/modules/DeploymentTester/Settings.cs
```diff
@@ -66,7 +46,26 @@ class Settings
             this.DeploymentUpdatePeriod = deploymentUpdatePeriod;
         }
 
-        internal static Settings Current => DefaultSettings.Value;
+        static Settings Create()
+        {
+            IConfiguration configuration = new ConfigurationBuilder()
+                .SetBasePath(Directory.GetCurrentDirectory())
+                .AddJsonFile("config/settings.json", optional: true)
```
and-rewsmith: Should we use optional: true?  In some other places we don't use it.

PR #2305, opened 2020-01-14
Add agent to custom network 

damonbarry:
Currently, when edge agent is started by iotedged for the first time (container doesn't already exist on the host) it is added to the default docker network, e.g., "bridge" on Linux.

When the agent's container is upgraded by the deployment (i.e., the deployment specifies a docker image that doesn't match what was given in config.yaml), it is added to the network specified by config.yaml's `moby_runtime.network[.name]` field, e.g., "azure-iot-edge" by default on Linux.

This behavior was inadvertently introduced a long time ago (see e2006e0a4). The common case seems to be to use the default `mcr.microsoft.com/azureiotedge-agent:1.0` image to bootstrap then change to a specific image in the deployment, so for most people the agent lives in the right network most of the time. And even if it's added to the default network instead, that's usually not a problem, which probably explains why this hasn't been an issue. I ran into it while working on the new diagnostics feature (the agent needs to connect to edge hub's Prometheus endpoint, but it's in a different network so it can't resolve the name).

This change makes the behavior more consistent. Now, when the daemon creates the agent container, it will add it to the custom network--the same network it will live on if it ever gets upgraded by a deployment.

### DIFF
--- a/edgelet/edgelet-docker/src/runtime.rs
+++ b/edgelet/edgelet-docker/src/runtime.rs
@@ -17,7 +17,10 @@ use url::Url;
 
 use docker::apis::client::APIClient;
 use docker::apis::configuration::Configuration;
-use docker::models::{ContainerCreateBody, ContainerCreateBodyNetworkingConfig, EndpointSettings, InlineResponse200, Ipam, NetworkConfig};
+use docker::models::{
+    ContainerCreateBody, ContainerCreateBodyNetworkingConfig, EndpointSettings, InlineResponse200,
+    Ipam, NetworkConfig,
+};
 use edgelet_core::{
     AuthId, Authenticator, GetTrustBundle, Ipam as CoreIpam, LogOptions, MakeModuleRuntime,
     MobyNetwork, Module, ModuleId, ModuleRegistry, ModuleRuntime, ModuleRuntimeState, ModuleSpec,
--- a/edgelet/edgelet-docker/src/runtime.rs
+++ b/edgelet/edgelet-docker/src/runtime.rs
@@ -17,14 +17,15 @@ use url::Url;
 
 use docker::apis::client::APIClient;
 use docker::apis::configuration::Configuration;
-use docker::models::{ContainerCreateBody, InlineResponse200, Ipam, NetworkConfig};
+use docker::models::{ContainerCreateBody, ContainerCreateBodyNetworkingConfig, EndpointSettings, InlineResponse200, Ipam, NetworkConfig};
 use edgelet_core::{
     AuthId, Authenticator, GetTrustBundle, Ipam as CoreIpam, LogOptions, MakeModuleRuntime,
     MobyNetwork, Module, ModuleId, ModuleRegistry, ModuleRuntime, ModuleRuntimeState, ModuleSpec,
     RegistryOperation, RuntimeOperation, SystemInfo as CoreSystemInfo, SystemResources, UrlExt,
 };
 use edgelet_http::{Pid, UrlConnector};
-use edgelet_utils::{ensure_not_empty_with_context, log_failure};
+use edgelet_utils::{ensure_not_empty_with_context, log_failure, serde_clone};
+
 use provisioning::ProvisioningResult;
 
 use crate::client::DockerClient;
@@ -350,13 +351,32 @@ impl ModuleRuntime for DockerModuleRuntime {
                     module.config().image()
                 );
 
-                let create_options = create_options
+                let mut create_options = create_options
                     .with_image(module.config().image().to_string())
                     .with_env(merged_env)
                     .with_labels(labels);
 
-                // Here we don't add the container to the iot edge docker network as the edge-agent is expected to do that.
-                // It contains the logic to add a container to the iot edge network only if a network is not already specified.
+                // In the bootstrap scenario (edge runtime module starting up for the first time),
+                // add the container to the custom network. We identify this scenario by looking
+                // for 'NetworkId' in the module's environment. The edge runtime module is then
+                // responsible for adding all subsequent modules to the custom network.
+                if let Some(network_id) = module.env().get("NetworkId") {
+                    let mut network_config = create_options
+                        .networking_config()
+                        .and_then(|network_config| serde_clone(network_config).ok())
+                        .unwrap_or_else(ContainerCreateBodyNetworkingConfig::new);
+
+                    let mut endpoints_config = network_config
+                        .endpoints_config()
+                        .and_then(|endpoints_config| serde_clone(endpoints_config).ok())
+                        .unwrap_or_else(HashMap::new);
+                    if !endpoints_config.contains_key(network_id.as_str()) {
+                        endpoints_config.insert(network_id.to_string(), EndpointSettings::new());
+                    }
+
+                    network_config = network_config.with_endpoints_config(endpoints_config);
+                    create_options = create_options.with_networking_config(network_config);
+                }
 
                 Ok(self
                     .client
--- a/edgelet/edgelet-docker/tests/runtime.rs
+++ b/edgelet/edgelet-docker/tests/runtime.rs
@@ -775,6 +775,13 @@ fn container_create_handler(req: Request<Body>) -> ResponseFuture {
                         .unwrap()
                 );
 
+                assert!(create_options
+                    .networking_config()
+                    .unwrap()
+                    .endpoints_config()
+                    .unwrap()
+                    .contains_key(&"test-network-id".to_string()));
+
                 let volumes = create_options.volumes().unwrap();
                 let mut expected = ::std::collections::HashMap::new();
                 expected.insert("test1".to_string(), json!({}));
@@ -821,6 +828,7 @@ fn container_create_succeeds() {
             env.insert("k1".to_string(), "v1".to_string());
             env.insert("k2".to_string(), "v2".to_string());
             env.insert("k3".to_string(), "v3".to_string());
+            env.insert("NetworkId".to_string(), "test-network-id".to_string());
 
             // add some create options
             let mut port_bindings = HashMap::new();
--- a/doc/IPv6Configuration.md
+++ b/doc/IPv6Configuration.md
@@ -1,17 +1,12 @@
 # IPv6 support in IoT Edge (Linux only)
 
-IoT Edge can be configured to work on Linux devices that are on IPv6 networks. On Linux devices, IoT Edge creates modules in two different docker networks:
-
-1. Default 'bridge' network: This network is created by docker. The Edge Agent module is deployed to this network.
-
-2. Custom 'azure-iot-edge' network: This network is created by `iotedged`. All user defined modules along with Edge Hub are deployed to this network.
+IoT Edge can be configured to work on Linux devices that are on IPv6 networks. On Linux devices, a user-defined network named 'azure-iot-edge' is created by `iotedged`. All modules, including Edge Agent and Edge Hub, are deployed to this network.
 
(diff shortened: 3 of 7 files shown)


edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -350,13 +354,32 @@ impl ModuleRuntime for DockerModuleRuntime {
...
 
-                let create_options = create_options
+                let mut create_options = create_options
                     .with_image(module.config().image().to_string())
                     .with_env(merged_env)
                     .with_labels(labels);
 
-                // Here we don't add the container to the iot edge docker network as the edge-agent is expected to do that.
-                // It contains the logic to add a container to the iot edge network only if a network is not already specified.
+                // In the bootstrap scenario (edge runtime module starting up for the first time),
```
arsing: "edge runtime" -> "edge agent" ? Or is calling it "runtime" intentional?

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -350,13 +354,32 @@ impl ModuleRuntime for DockerModuleRuntime {
...
+                // add the container to the custom network. We identify this scenario by looking
+                // for 'NetworkId' in the module's environment. The edge runtime module is then
+                // responsible for adding all subsequent modules to the custom network.
+                if let Some(network_id) = module.env().get("NetworkId") {
```
arsing: Just to clarify, `NetworkId` will actually always be defined, right? All user modules created by Edge Agent will have it set by Edge Agent already, and now with this change Edge Agent will have it set by the watchdog?

edgelet/edgelet-docker/tests/runtime.rs
```diff
@@ -775,6 +775,13 @@ fn container_create_handler(req: Request<Body>) -> ResponseFuture {
                         .unwrap()
                 );
 
+                assert!(create_options
+                    .networking_config()
+                    .unwrap()
+                    .endpoints_config()
+                    .unwrap()
+                    .contains_key(&"test-network-id".to_string()));
```
arsing: `&"test-network-id".to_string()` -> `"test-network-id"`, because `HashMap::contains_key` allows you to use a `Borrow`, ie `str`

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -350,13 +354,32 @@ impl ModuleRuntime for DockerModuleRuntime {
...
+                    let mut network_config = create_options
+                        .networking_config()
+                        .and_then(|network_config| serde_clone(network_config).ok())
```
arsing: What does `serde_clone` do that a normal `.clone()` wouldn't?

edgelet/edgelet-docker/src/runtime.rs
varunpuranik: Why can't we update the spec for the EdgeAgent that the watchdog uses, instead? 
https://github.com/Azure/iotedge/blob/master/edgelet/iotedged/src/lib.rs#L1978
(In short, instead of fixing it in the lower layer, why don't we fix it in the higher layer)?

damonbarry: I suspect it was intentional, although I tend to think of the runtime as the collection of all system modules (agent + hub). Anyway, elsewhere in the code it was referring to the edge runtime module, so I kept that here. I can change it to edge agent.

damonbarry: Just noticed the comment I replaced referred to edge agent not runtime, so yeah, I'll change it.

edgelet/edgelet-docker/src/runtime.rs
damonbarry: I _think_ that because of the merge of environments (env from create options + env from spec) earlier in the function, most if not all modules will have `NetworkId` defined, but only edge agent in the bootstrap case will have it defined in the spec. I'm specifically looking in the spec (`module.env()`) here, not `merged_env`. But I may not be following the logic correctly...what do you think?

edgelet/edgelet-docker/tests/runtime.rs
damonbarry: Excellent, will do.

edgelet/edgelet-docker/src/runtime.rs
damonbarry: Frankly not sure. This was a copy/paste of the original logic that was removed by e2006e0 (and replaced with the comment "Here we don't add the container to the iot edge docker network..."). Maybe @varunpuranik remembers?

edgelet/edgelet-docker/src/runtime.rs
arsing: The watchdog uses an opaque generic for the module spec because it's generic on the module runtime, so it can't see it as a `DockerConfig` and mutate it.

edgelet/edgelet-docker/src/runtime.rs
arsing: Yeah, I'd spelunked yesterday and found it was originally added in https://github.com/Azure/iotedge/commit/aa2740644eba812bb96c0ee0299f97296db719e6 , but that doesn't explain either.

edgelet/edgelet-docker/src/runtime.rs
damonbarry: That's what I originally tried to do. But `DockerModuleRuntime::create` is the first function in that call chain that knows anything about Docker. The network needs to be added to the create options, and I can't do that at the higher level. But the problem with `create` is that it gets called for all modules, not just agent and not just the first time it starts. So I had to come up with a way to discriminate.

There might be a way to do it but I couldn't find one...

arsing: We could add a new trait for the config, with a `fn set_network(&mut self, network_id: String);` and make it a bound for https://github.com/Azure/iotedge/blob/b1e6ba543f5be7b4c6cb9b6d7d7f2d97b401e203/edgelet/edgelet-core/src/module.rs#L146 . But it would touch a lot more files than what this PR does now, though it *would* definitely be cleaner.

Alternatively we could add the trait but only add the bound for the watchdog function rather than `ModuleSpec` itself. I'm leaning towards this now that I think about it.

arsing: @damonbarry I suggest:

- Add
    ```rust
    trait Config {
        fn set_network(&mut self, network_id: String);
    }
    ```

    somewhere under `edgelet-core`

- impl it for `edgelet_docker::DockerConfig` and friends

- change https://github.com/Azure/iotedge/blob/b1e6ba543f5be7b4c6cb9b6d7d7f2d97b401e203/edgelet/edgelet-core/src/watchdog.rs#L42 to

    ```rust
    <M::Module as Module>::Config: Clone + Config,
    ```

  (Adding the bound here rather than on `struct ModuleSpec` itself means you don't have to change the rest of the codebase that uses `ModuleSpec`.

- and add a `spec.set_network()` call in https://github.com/Azure/iotedge/blob/b1e6ba543f5be7b4c6cb9b6d7d7f2d97b401e203/edgelet/edgelet-core/src/watchdog.rs#L56 You'll also have to thread `settings.moby_runtime().network()` to the watchdog from main.

damonbarry: I'll give it a try. I definitely wasn't loving taking advantage of a side-effect as a solution. Thanks!

edgelet/edgelet-docker/src/runtime.rs
arsing: Ah yes, I forgot there are two different env maps.

edgelet/edgelet-docker/src/runtime.rs
arsing: I checked with Raj. He says it may have been because he didn't want to add `Clone` to the `docker::models::*` types by hand. But I saw Varun added them by hand in https://github.com/Azure/iotedge/commit/2cbe1bd74109fad7a1fc69398de2218eae423d44 anyway, so that doesn't matter any more.

I think we should `Clone` instead of `serde_clone`, at least for the "new" code in this PR. Though if you remove `serde_clone` entirely in this PR I'll support it. The only other use is for `ContainerCreateBody` which is already `Clone` from Varun's change.

edgelet/edgelet-docker/src/runtime.rs
damonbarry: > You'll also have to thread settings.moby_runtime().network() to the watchdog from main.

@arsing Does that mean I need to add a new trait for Settings too (with, e.g., `fn get_network()`)? `M::Settings` constraints in Main are super generic...not docker-specific.

arsing: Oh, drat. I forgot about that.

There's already [`RuntimeSettings`](https://github.com/Azure/iotedge/blob/53757e2be8bc9aedab41d46b9472ac7b8cca3cd0/edgelet/edgelet-core/src/settings.rs#L677) for that, but looking at the impls I see that `moby_runtime` is [`edgelet-docker`-specific.](https://github.com/Azure/iotedge/blob/53757e2be8bc9aedab41d46b9472ac7b8cca3cd0/edgelet/edgelet-docker/src/settings.rs#L52) So it would run counter to the purpose of that trait to expose `fn network(&self) -> &str` on it, and not be possible to do for [`edgelet-kube`'s struct](https://github.com/Azure/iotedge/blob/53757e2be8bc9aedab41d46b9472ac7b8cca3cd0/edgelet/edgelet-kube/src/settings.rs#L19) anyway.

arsing:
(Assuming we keep the env-var implementation and not the alternatives in the subthread)

Aren't you missing a change to https://github.com/Azure/iotedge/blob/53757e2be8bc9aedab41d46b9472ac7b8cca3cd0/edgelet/iotedged/src/lib.rs#L1996 to actually set the `NetworkId` env var?

damonbarry:
> (Assuming we keep the env-var implementation and not the alternatives in the subthread)
> 
> Aren't you missing a change to
> 
> https://github.com/Azure/iotedge/blob/53757e2be8bc9aedab41d46b9472ac7b8cca3cd0/edgelet/iotedged/src/lib.rs#L1996
> 
> to actually set the `NetworkId` env var?

@arsing No, NetworkId is already inserted into the agent section of the settings by `Settings::new()`, which is called by `unix::run()` to create the settings struct that gets passed into `main.run_until()`. Further on, in `start_runtime()`, the agent section of `settings` is cloned into the new `spec`, whose `env()` is passed into `build_env()`, which copies `spec_env` wholesale.

arsing:
Okay, all the way up there. So since that is Docker-specific, and thus has access to the actual types of settings as well as spec as well as container create options, can you change that to set the `NetworkingConfig` directly with the code you added instead of / in addition to the env var?

damonbarry:
> So since that is Docker-specific...

It's not Docker-specific. Those are just environment variables getting copied around, without any real context as to what they mean, where they're from, etc. That's how my original PR changes worked; they leveraged the fact that some Docker-specific info happened to be getting into the spec through the environment variables (rather than through docker-specific methods on a struct).

arsing:
>>So since that is Docker-specific...
>
>It's not Docker-specific.

The "that" that I'm talking about is https://github.com/Azure/iotedge/blob/master/edgelet/edgelet-docker/src/settings.rs#L174-L180

damonbarry:
Ah, I get it, sorry. 😐 Investigating...

damonbarry:
Alright, making the changes inside `edgelet_docker::settings::Settings` was easy and clean. I created a new PR for it; check out #2318. Thanks @arsing for the hot tip!

PR #2308, opened 2020-01-15
Refactor and fix test reports

philipktlin:
1. fix missing module id assignment in settings class.
2. update to use IOT_HUB_CONNECTION_STRING as key in setting class.
3. Remove actual and expected source from ITestReportMetadata since not all reports have 2 result sources
4. Remove actual and expected source from ITestResultReport  since not all reports have 2 result sources

### DIFF
--- a/test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
+++ b/test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
@@ -18,90 +18,88 @@ namespace TestResultCoordinator.Reports
 
         public ITestResultReportGenerator Create(
             string trackingId,
-            IReportMetadata reportMetadata)
+            ITestReportMetadata testReportMetadata)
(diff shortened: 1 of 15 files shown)


test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
```diff
@@ -18,90 +18,85 @@ internal TestReportGeneratorFactory(ITestOperationResultStorage storage)
 
         public ITestResultReportGenerator Create(
             string trackingId,
-            IReportMetadata reportMetadata)
+            ITestReportMetadata testReportMetadata)
         {
```
yophilav:
```
Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
Preconditions.CheckNotNull(testReportMetadata, nameof(testReportMetadata));
```

PR #2310, opened 2020-01-15
Add metrics collector to connectivity test pipeline

and-rewsmith:
Currently, we are not collecting metrics for our connectivity tests. This PR adds the metrics collector to the connectivity pipeline, and also adds some additional logic to the module for reporting test specific data.

Here is a breakdown of this additional logic in the metrics collector:
- Report test type so that the metrics collector can tag metrics with that type. This will be helpful in the dashboard.
- Get all additional, test-specific tags from the **module twin's desired properties**. This is desirable over environment variables because the metrics collector module can simply pass all the key-value pairs into the scraped metrics without coupling the implementation to the current set of test-specific tags. This also allows us to keep the metrics collector module as a "non-test" module that is targeted at customers.


e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -596,6 +626,15 @@
...
         }
       }
+    },
+    "metricsCollector": {
+      "properties.desired": {
+        "upstreamProtocol": "<UpstreamProtocol>",
+        "offlineFrequency": "<NetworkController.OfflineFrequency0>",
+        "onlineFrequency": "<NetworkController.OnlineFrequency0>",
+        "runsCount": "<NetworkController.RunsCount0>",
+        "testDuration": "<TestDuration>"
```
philipktlin: You may need to consider testStartDelay and plus some buffer for metrics collector module.

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -59,9 +57,16 @@ IEnumerable<Metric> GetGuidTaggedMetrics(IEnumerable<Metric> metrics)
         {
             foreach (Metric metric in metrics)
             {
-                Dictionary<string, string> tagsWithGuid = new Dictionary<string, string>(metric.Tags);
-                tagsWithGuid.Add("guid", this.metricsCollectorRuntimeId.ToString());
-                yield return new Metric(metric.TimeGeneratedUtc, metric.Name, metric.Value, tagsWithGuid);
+                Dictionary<string, string> metricTags = new Dictionary<string, string>(metric.Tags);
+
+                metricTags.Add("testId", this.metricsCollectorRuntimeId.ToString());
```
philipktlin: I think you should pass testId and testType using testSpecificTags instead.

e2e_deployment_files/connectivity_deployment.template.json
and-rewsmith: Do we want to delay the metrics collector? It is not part of the test, so I was thinking I could start immediately.

edge-modules/MetricsCollector/Program.cs
```diff
@@ -29,6 +32,8 @@ private static async Task<int> MainAsync()
             try
             {
                 moduleClient = await ModuleClient.CreateFromEnvironmentAsync(transportSettings);
+                Dictionary<string, string> testSpecificTags = await GetTestSpecificTagsFromTwin(moduleClient);
```
philipktlin: why do you want to get additional metric tags from Twin? requirements from customer?

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -59,9 +57,16 @@ IEnumerable<Metric> GetGuidTaggedMetrics(IEnumerable<Metric> metrics)
         {
             foreach (Metric metric in metrics)
             {
-                Dictionary<string, string> tagsWithGuid = new Dictionary<string, string>(metric.Tags);
-                tagsWithGuid.Add("guid", this.metricsCollectorRuntimeId.ToString());
```
philipktlin: why remove this tag for this.metricsCollectorRuntimeId?

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -11,30 +11,15 @@ namespace MetricsCollector
...
-                });
+        internal static Settings Current = Create();
 
         Settings(
             string azMonWorkspaceId,
             string azMonWorkspaceKey,
             string azMonLogType,
             string endpoints,
             int scrapeFrequencySecs,
+            string testType,
```
philipktlin: should it be generic to get metric additional tags?

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -55,11 +40,28 @@ public class Settings
...
+            IConfiguration configuration = new ConfigurationBuilder()
+                .SetBasePath(Directory.GetCurrentDirectory())
+                .AddJsonFile("config/settings.json")
+                .AddEnvironmentVariables()
+                .Build();
+
+            return new Settings(
+                configuration.GetValue<string>("AzMonWorkspaceId"),
+                configuration.GetValue<string>("AzMonWorkspaceKey"),
+                configuration.GetValue<string>("AzMonLogType", "testMetrics"),
```
philipktlin: what are the possible values for AzMonLogType? what is AzMonLogType used for?

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -11,30 +11,15 @@ namespace MetricsCollector
 
     public class Settings
```
philipktlin:
```suggestion
    class Settings
```

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -71,6 +73,8 @@ public class Settings
 
         public int ScrapeFrequencySecs { get; }
 
+        public string TestType { get; }
```
philipktlin: Add TestType in ToString method

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
and-rewsmith: testId is a Guid that is generated at runtime rather than injected from the pipeline, so I don't think it fits well in testSpecificTags.

My logic for testSpecificTags was to include tags that only exist for a given test type. Since the testType field is used for every test I didn't put it in testSpecificTags.

What do you think?

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
and-rewsmith: I renamed the tag name to testId

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -596,6 +626,15 @@
           "timeToLiveSecs": 86400
         }
       }
+    },
+    "metricsCollector": {
+      "properties.desired": {
+        "upstreamProtocol": "<UpstreamProtocol>",
+        "offlineFrequency": "<NetworkController.OfflineFrequency0>",
+        "onlineFrequency": "<NetworkController.OnlineFrequency0>",
+        "runsCount": "<NetworkController.RunsCount0>",
```
philipktlin: please add prefix "networkController" for offlineFrequency,  onlineFrequency and runsCount.

e2e_deployment_files/connectivity_deployment.template.json
philipktlin: since TestDuration is used as additional metrics tag, that's fine.  I will update connectivity test to shut down iotedge service after test.

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -23,7 +23,14 @@
             "env": {
               "UpstreamProtocol": {
                 "value": "<UpstreamProtocol>"
+              },
+              "experimentalfeatures__enabled": {
```
philipktlin: do you know what happen if edgeAgent is enabled for metrics?  edgeHub still will send out metrics, right?

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -59,13 +43,28 @@ public class Settings
             this.UploadTarget = uploadTarget;
```
philipktlin: There are 2 values of upload target enum - eventhub and log analytics; what if upload target is eventhub? if eventhub is given, why user need to provide log analytics parameters?

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -579,6 +597,36 @@
               "image": "<Container_Registry>/microsoft/azureiotedge-network-controller:<Build.BuildNumber>-linux-<Architecture>",
               "createOptions": "{\"HostConfig\":{\"Privileged\":\"true\",\"NetworkMode\":\"host\",\"Binds\":[\"/var/run/docker.sock:/var/run/docker.sock\"]},\"NetworkingConfig\":{\"EndpointsConfig\":{\"host\":{}}}}"
             }
+          },
+          "metricsCollector": {
```
philipktlin: does metrics collector support offline?

e2e_deployment_files/connectivity_deployment.template.json
and-rewsmith: These flags are independent. You need both if you want both agent / hub to collect metrics.

e2e_deployment_files/long_haul_deployment.template.json
```diff
@@ -435,6 +438,12 @@
           "timeToLiveSecs": 86400
         }
       }
+    },
+    "metricsCollector": {
+      "properties.desired": {
+        "testType": "longhaul",
+        "testId": "<TrackingId>"
```
philipktlin: add upstream protocol?

e2e_deployment_files/connectivity_deployment.template.json
and-rewsmith: Metrics collector supports an offline flow through event hub so that it can use edge hub for retries and storage. The flow to log analytics does not support offline.

e2e_deployment_files/long_haul_deployment.template.windows.json
```diff
@@ -438,6 +441,12 @@
           "timeToLiveSecs": 86400
         }
       }
+    },
+    "metricsCollector": {
+      "properties.desired": {
+        "testType": "longhaul",
+        "testId": "<TrackingId>"
+      }
```
philipktlin: add upstream protocol

e2e_deployment_files/stress_deployment.template.windows.json
```diff
@@ -358,6 +361,12 @@
           "timeToLiveSecs": 86400
         }
       }
+    },
+    "metricsCollector": {
+      "properties.desired": {
+        "testType": "stress",
+        "testId": "<TrackingId>"
+      }
```
philipktlin: add upstream protocol

edge-modules/MetricsCollector/Settings.cs
and-rewsmith: You are correct. I will only validate if upload target is appropriate.

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -9,26 +9,21 @@ namespace MetricsCollector
...
     {
         static readonly ILogger Logger = MetricsUtil.CreateLogger("MetricsCollector");
         readonly IMetricsScraper scraper;
         readonly IMetricsPublisher publisher;
+        readonly Dictionary<string, string> testSpecificTags;
         PeriodicTask periodicScrapeAndUpload;
-        Guid metricsCollectorRuntimeId;
 
-        public MetricsScrapeAndUpload(IMetricsScraper scraper, IMetricsPublisher publisher, Guid metricsCollectorRuntimeId)
+        public MetricsScrapeAndUpload(IMetricsScraper scraper, IMetricsPublisher publisher, Dictionary<string, string> testSpecificTags)
```
philipktlin:
```suggestion
        public MetricsScrapeAndUpload(IMetricsScraper scraper, IMetricsPublisher publisher, Dictionary<string, string> additionalTags)
```

edge-modules/MetricsCollector/Program.cs
```diff
@@ -58,5 +63,11 @@ private static async Task<int> MainAsync()
             Logger.LogInformation("MetricsCollector Main() finished.");
             return 0;
         }
+
+        static async Task<Dictionary<string, string>> GetTestSpecificTagsFromTwin(ModuleClient moduleClient)
```
and-rewsmith: Need to rename

edge-modules/MetricsCollector/Program.cs
```diff
@@ -29,18 +32,20 @@ private static async Task<int> MainAsync()
             try
             {
```
philipktlin: Please add catch block and log error.

edge-modules/MetricsCollector/Program.cs
```diff
@@ -29,18 +32,20 @@ private static async Task<int> MainAsync()
...
-                    publisher = new LogAnalyticsUpload(Settings.Current.AzMonWorkspaceId, Settings.Current.AzMonWorkspaceKey, Settings.Current.AzMonLogType);
+                    publisher = new LogAnalyticsUpload(Settings.Current.LogAnalyticsWorkspaceId, Settings.Current.LogAnalyticsWorkspaceKey, Settings.Current.LogAnalyticsLogType);
                 }
                 else
                 {
                     publisher = new EventHubMetricsUpload(moduleClient);
                 }
 
-                using (MetricsScrapeAndUpload metricsScrapeAndUpload = new MetricsScrapeAndUpload(scraper, publisher, Guid.NewGuid()))
+                using (MetricsScrapeAndUpload metricsScrapeAndUpload = new MetricsScrapeAndUpload(scraper, publisher, testSpecificTags))
```
philipktlin: nit:use var
```suggestion
                using (var metricsScrapeAndUpload = new MetricsScrapeAndUpload(scraper, publisher, testSpecificTags))
```

edge-modules/MetricsCollector/Program.cs
```diff
@@ -29,18 +32,20 @@ private static async Task<int> MainAsync()
             try
             {
                 moduleClient = await ModuleClient.CreateFromEnvironmentAsync(transportSettings);
+                Dictionary<string, string> testSpecificTags = await GetTestSpecificTagsFromTwin(moduleClient);
```
philipktlin: tags may not be test related.  Please rename it.
```suggestion
                Dictionary<string, string> additionalMetricsTags = await GetAdditionalMetricsTagsFromTwin(moduleClient);
```

e2e_deployment_files/long_haul_deployment.template.json
and-rewsmith: Good idea

edge-modules/MetricsCollector/Program.cs
philipktlin: Maybe call log.Info for number of additional tags here

edge-modules/MetricsCollector/Program.cs
```diff
@@ -29,18 +32,20 @@ private static async Task<int> MainAsync()
...
                 else
                 {
                     publisher = new EventHubMetricsUpload(moduleClient);
                 }
 
-                using (MetricsScrapeAndUpload metricsScrapeAndUpload = new MetricsScrapeAndUpload(scraper, publisher, Guid.NewGuid()))
+                using (MetricsScrapeAndUpload metricsScrapeAndUpload = new MetricsScrapeAndUpload(scraper, publisher, testSpecificTags))
                 {
                     TimeSpan scrapeAndUploadInterval = TimeSpan.FromSeconds(Settings.Current.ScrapeFrequencySecs);
                     metricsScrapeAndUpload.Start(scrapeAndUploadInterval);
```
philipktlin: add log message for "start running metrics ..."

and-rewsmith: The periodic task does a good bit of logging. Still want me to add?

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -228,6 +240,9 @@ function process_args() {
                 '-upstreamProtocol' ) saveNextArg=19;;
                 '-deploymentTestUpdatePeriod' ) saveNextArg=20;;
                 '-timeForReportingGeneration' ) saveNextArg=21;;
+                '-metricsEndpointsCSV' ) saveNextArg=22;;
+                '-metricsScrapeFrequencyInSecs' ) saveNextArg=23;;
+                '-metricsUploadTarget' ) saveNextArg=24;;
                 '-waitForTestComplete' ) WAIT_FOR_TEST_COMPLETE=1;;
 
                 '-cleanAll' ) CLEAN_ALL=1;;
```
philipktlin: Please help to add "eventHubConnectionString" field in below required parameters checking.

Also default value for log type is set to "connectivity"; please change it to "testMetrics".

e2e_deployment_files/connectivity_deployment.template.json
and-rewsmith: Although I assumed CollectMetrics was required for edgeHub, but it might not be (might only be required for MetricsV0)

test/connectivity/scripts/connectivityTest.sh
and-rewsmith: The log type of "connectivity" is for test _reporting_, not metrics, so it should not be using the "testMetrics" log type. Because all the metrics are of the same type, the log type for "testMetrics" is hardcoded in the deployment file.

e2e_deployment_files/long_haul_deployment.template.json
and-rewsmith: Longhaul only uses amqp as upstream so I will hardcode this

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -579,6 +597,36 @@
...
+                "ScrapeFrequencyInSecs": {
+                    "value": "<MetricsCollector.ScrapeFrequencyInSecs>"
+                },
+                "UploadTarget": {
+                  "value": "<MetricsCollector.UploadTarget>"
+                }
+            },
+            "settings": {
+                "image": "<Container_Registry>/microsoft/azureiotedge-metrics-collector:<Build.BuildNumber>-linux-<Architecture>",
+                "createOptions": ""
```
philipktlin: please use host network; and make sure to open port in edgeAgent and edgeHub of metric endpoints to the host.

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -36,6 +43,17 @@
               "image": "<Container_Registry>/microsoft/azureiotedge-hub:<Build.BuildNumber>-linux-<Architecture>",
               "createOptions": "{\"HostConfig\": {\"PortBindings\": {\"8883/tcp\": [{\"HostPort\": \"8883\"}],\"443/tcp\": [{\"HostPort\": \"443\"}],\"5671/tcp\": [{\"HostPort\": \"5671\"}]}}}"
```
philipktlin: open metric endpoint port

e2e_deployment_files/connectivity_deployment.template.json
and-rewsmith: It is running in the host network so this will work for connectivity tests

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -23,7 +23,14 @@
...
+              "experimentalfeatures__enabled": {
+                  "value": "true"
+              },
+              "experimentalfeatures__enableMetrics": {
+                  "value": "true"
               }
+            }
              },
             "settings": {
               "image": "<Container_Registry>/microsoft/azureiotedge-agent:<Build.BuildNumber>-linux-<Architecture>",
```
philipktlin: open metric endpoint port

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -23,18 +23,36 @@
...
-              "createOptions": ""
+              "createOptions": "{\"HostConfig\": {\"PortBindings\": {\"9600/tcp\": [{\"HostPort\": \"9600\"}]}}}",
             }
           },
           "edgeHub": {
             "type": "docker",
             "settings": {
               "image": "<Container_Registry>/microsoft/azureiotedge-hub:<Build.BuildNumber>-linux-<Architecture>",
-              "createOptions": "{\"HostConfig\": {\"PortBindings\": {\"8883/tcp\": [{\"HostPort\": \"8883\"}],\"443/tcp\": [{\"HostPort\": \"443\"}],\"5671/tcp\": [{\"HostPort\": \"5671\"}]}}}"
+              "createOptions": "{\"HostConfig\": {\"PortBindings\": {\"8883/tcp\": [{\"HostPort\": \"8883\"}],\"443/tcp\": [{\"HostPort\": \"443\"}],\"5671/tcp\": [{\"HostPort\": \"5671\"}],\"9600/tcp\": [{\"HostPort\": \"9600\"}]}}}"
```
and-rewsmith: Here is the relevant change:
```,\"9600/tcp\": [{\"HostPort\": \"9600\"}]```

File: edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs [hunk 27]
```diff
@@ -9,26 +9,21 @@ namespace MetricsCollector
...
-    using Serilog.Core;
-    using Serilog.Events;
     using ILogger = Microsoft.Extensions.Logging.ILogger;
 
     class MetricsScrapeAndUpload : IDisposable
     {
         static readonly ILogger Logger = MetricsUtil.CreateLogger("MetricsCollector");
         readonly IMetricsScraper scraper;
         readonly IMetricsPublisher publisher;
+        readonly Dictionary<string, string> testSpecificTags;
```
and-rewsmith: Will rename

edge-modules/MetricsCollector/Settings.cs
```diff
@@ -77,8 +79,8 @@ public override string ToString()
         {
             var fields = new Dictionary<string, string>()
             {
-                { nameof(this.AzMonWorkspaceId), this.AzMonWorkspaceId },
-                { nameof(this.AzMonLogType), this.AzMonLogType },
+                { nameof(this.LogAnalyticsWorkspaceId), this.LogAnalyticsWorkspaceId },
```
philipktlin: same for LogAnalyticsLogType 
```suggestion
                { nameof(this.LogAnalyticsWorkspaceId), this.LogAnalyticsWorkspaceId ?? string.Empty },
```

scripts/linux/runE2ETest.sh
```diff
@@ -169,6 +169,7 @@ function prepare_test_from_artifacts() {
 
                 local escapedSnitchAlertUrl
                 local escapedBuildId
+                local tracking_id=$(cat /proc/sys/kernel/random/uuid)
```
philipktlin: nit: line break

test/connectivity/scripts/connectivityTest.sh
philipktlin: yes, you are right.  Please help to add 1 required field checking for "eventHubConnectionString".  thanks.

edge-modules/MetricsCollector/Program.cs
and-rewsmith: I prefer not to use var. Is this a strict rule?

e2e_deployment_files/long_haul_deployment.template.json
```diff
@@ -412,8 +413,10 @@
                   "value": "<MetricsCollector.UploadTarget>"
                 }
             },
-            "status": "running",
-            "restartPolicy": "always"
```
and-rewsmith: I reordered this for consistency

edge-hub/docker/linux/amd64/Dockerfile
```diff
@@ -24,10 +24,11 @@ WORKDIR /app
 
 COPY $EXE_DIR/ ./	
 
-# Expose MQTT, AMQP and HTTPS ports	
+# Expose MQTT, AMQP, HTTPS, and metrics ports
 EXPOSE 8883/tcp	
 EXPOSE 5671/tcp	
 EXPOSE 443/tcp	
+EXPOSE 9600/tcp
```
philipktlin: if customer runs metric module in iotedge, this port is not required to expose.  Therefore you should not add this line.

and-rewsmith: I brought this point up with Anca and we decided this was fine because expose does not actually expose the port, but allows the port to be exposed with CreateOptions.

philipktlin:
Next time, you should create a new PR for changes in iotedge quickstart for deploy edge agent 1.0.

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -23,18 +23,35 @@
...
+              },
+              "experimentalfeatures__enableMetrics": {
+                  "value": "true"
               }
-             },
+            },
             "settings": {
               "image": "<Container_Registry>/microsoft/azureiotedge-agent:<Build.BuildNumber>-linux-<Architecture>",
-              "createOptions": ""
+              "createOptions": "{\"HostConfig\": {\"PortBindings\": {\"9600/tcp\": [{\"HostPort\": \"9600\"}]}}}"
```
philipktlin: see if ExposedPorts option works for edgeAgent and edgeHub.  https://docs.docker.com/engine/api/v1.40/#operation/ContainerCreate

and-rewsmith: I was going to try this, but from the docker inspects I did earlier it is possible there is more configuration on top of PortBindings and ExposedPorts if passing by createOptions. Once I verify that this works I will check if ExposedPorts works.

edge-hub/docker/windows/arm32v7/Dockerfile
```diff
@@ -8,7 +8,7 @@ WORKDIR /app
 
 COPY $EXE_DIR/ ./
 
-# Expose MQTT, AMQP and HTTPS ports
```
and-rewsmith: I can revert this, but I thought this was better so when we add exposed ports we don't need to worry about the "and".

PR #2314, opened 2020-01-15
Add new layered deployment integration tests

massand:
Add a new integration test that composes a base and add-on deployment, deploys it to a fresh Edge device and validates reporting via ADM deployment metrics functionality


darobs:
>                 IEnumerable<IRequestHandler> requestHandlers = new List<IRequestHandler> { new PingRequestHandler() };

Seems like this could be a function, as this is identical code to 359-387

---
Refers to: edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/EdgeAgentConnectionTest.cs:266 in 6e0443a. [](commit_id = 6e0443a4b0ac2184185ba8aec10466e87a223db4, deletion_comment = False)

darobs:
>                 Assert.Equal(1, config.SystemMetrics.Results["appliedCount"]);

We might be able to remove this check since we are checking both base and addon deployment in the next test.

---
Refers to: edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/EdgeAgentConnectionTest.cs:418 in 6e0443a. [](commit_id = 6e0443a4b0ac2184185ba8aec10466e87a223db4, deletion_comment = False)

massand:
@darobs - The validation, while somewhat redundant, is in place to ensure backwards compatibility of the service API. I will retain the current tests validation. I have addressed your feedback about reducing redundant code. Thanks for the review!

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/EdgeAgentConnectionTest.cs
```diff
@@ -31,23 +31,51 @@ public class EdgeAgentConnectionTest
         const string DockerType = "docker";
         static readonly TimeSpan DefaultRequestTimeout = TimeSpan.FromSeconds(60);
 
-        public static async Task<Configuration> CreateConfigurationAsync(RegistryManager registryMananger, string configurationId, string targetCondition, int priority)
+        private static async Task CreateConfigurationAsync(RegistryManager registryManager, string configurationId, string targetCondition, int priority)
```
philipktlin: nit: remove private which is the default access modifier for method.
```suggestion
        static async Task CreateConfigurationAsync(RegistryManager registryManager, string configurationId, string targetCondition, int priority)
```

PR #2315, opened 2020-01-15
Dispose all clients when network disconnection (#2291)

ancaantochi:
Cherrypick #2291 from master

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceClientWrapper.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceClientWrapper.cs
@@ -24,9 +24,15 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
 
         public Task AbandonAsync(string messageId) => this.underlyingDeviceClient.AbandonAsync(messageId);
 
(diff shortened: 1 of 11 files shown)


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceClientWrapper.cs
```diff
@@ -24,9 +24,15 @@ public DeviceClientWrapper(DeviceClient deviceClient)
...
         public Task AbandonAsync(string messageId) => this.underlyingDeviceClient.AbandonAsync(messageId);
 
-        public Task CloseAsync() => this.isActive.GetAndSet(false)
-            ? this.underlyingDeviceClient.CloseAsync()
-            : Task.CompletedTask;
+        public Task CloseAsync()
+        {
+            if (this.isActive.GetAndSet(false))
+            {
+                this.underlyingDeviceClient?.Dispose();
```
varunpuranik: Do you need the null check on underlyingClient?

ancaantochi: not really needed, I did same as Dispose. I can remove it, but I would like to do it in a PR that has to go to master and than backport to 1.0.9

PR #2316, opened 2020-01-15
Set timer for Connectivity checker to run every minute while offline

ancaantochi:
Checking every minute helps detecting faster when device comes back online. 
When online the checker runs every 5 minutes, no change to that in this PR.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/RoutingModule.cs
```diff
@@ -178,7 +178,7 @@ protected override void Load(ContainerBuilder builder)
                         var edgeHubCredentials = c.ResolveNamed<IClientCredentials>("EdgeHubCredentials");
                         IDeviceConnectivityManager deviceConnectivityManager = this.experimentalFeatures.DisableConnectivityCheck
                             ? new NullDeviceConnectivityManager()
-                            : new DeviceConnectivityManager(this.connectivityCheckFrequency, TimeSpan.FromMinutes(2), edgeHubCredentials.Identity) as IDeviceConnectivityManager;
+                            : new DeviceConnectivityManager(this.connectivityCheckFrequency, TimeSpan.FromMinutes(1), edgeHubCredentials.Identity) as IDeviceConnectivityManager;
```
varunpuranik: Looks good. Only concern would be - will this be too quick? As in will the earlier call complete before the new one starts? For that the timeout of 20s on the OpenAsync will have to work as expected.

PR #2317, opened 2020-01-15
Get report metadata from Twin in TRC

philipktlin:
update to get report metadata from Twin for TRC module; add unit tests.


test/modules/Modules.Test/TestResultCoordinator/TestUtilTest.cs
```diff
@@ -0,0 +1,253 @@
...
+
+        [Fact]
+        public async Task TestGenerateTestResultReportsAsync_NullReportMetadata()
+        {
+            var mockLogger = new Mock<ILogger>();
+            var mockTestReportGeneratorFactory = new Mock<ITestReportGeneratorFactory>();
+
+            ArgumentNullException ex = await Assert.ThrowsAsync<ArgumentNullException>(() =>
+                 TestUtil.GenerateTestResultReportsAsync(
+                     "fakeTrackngId",
```
dylanbronson: nit: typo

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -131,51 +134,34 @@ public override string ToString()
             return $"Settings:{Environment.NewLine}{string.Join(Environment.NewLine, fields.Select(f => $"{f.Key}={f.Value}"))}";
         }
 
-        internal List<ITestReportMetadata> GetReportMetadataList()
+        internal async Task<List<ITestReportMetadata>> GetReportMetadataListAsync(ILogger logger)
```
ancaantochi: It looks like this method should be in the TestUtil. Same for GetResultSources

test/modules/TestResultCoordinator/TestUtil.cs
```diff
@@ -0,0 +1,136 @@
...
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json.Linq;
+    using TestResultCoordinator.Reports;
+    using TestResultCoordinator.Reports.DirectMethod;
+
+    static class TestUtil
```
ancaantochi: Suggestion: rename to Util or ReportGeneratorUtil

test/modules/TestResultCoordinator/TestUtil.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+
+        internal static List<ITestReportMetadata> ParseReportMetadataJson(string reportMetadataListJson, ILogger logger)
+        {
+            var reportMetadataList = new List<ITestReportMetadata>();
+
+            try
+            {
+                JObject reportMetadatas = JObject.Parse(reportMetadataListJson);
```
ancaantochi: Would deserializing the json to its corresponding report metadata be easier?

test/modules/TestResultCoordinator/Settings.cs
philipktlin: This method used a class field to persist the result from GetTwin and ReportMetadata is part of settings.  Therefore I think it is better to be in the class compared to TestUtil static class.

test/modules/TestResultCoordinator/TestUtil.cs
philipktlin: It is not only used for report generation; how about TestReportUtil.

test/modules/TestResultCoordinator/TestUtil.cs
philipktlin: updated to use json deserializaiton

test/modules/TestResultCoordinator/Settings.cs
varunpuranik: With the "Current" pattern, it feels like the Twin should be obtained and cached in Create() itself.

test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -0,0 +1,109 @@
...
+            catch (Exception ex)
+            {
+                logger.LogError(ex, $"Incorrect parsing for report metadata list: {reportMetadataListJson}{Environment.NewLine}");
+                throw;
+            }
+
+            return reportMetadataList;
+        }
+
+        private static TEnum GetEnumValueFromReportMetadata<TEnum>(JToken metadata, string key)
```
varunpuranik: Nit: avoid private access modifier on methods and fields

test/modules/TestResultCoordinator/Settings.cs
philipktlin: Current pattern is the singleton pattern to ensure only 1 Settings in the process.  I think it is reasonable for this class to provide methods to manage more setting-related data.

If you prefer to change Create method become async, then I need to:
1. Rename Create method to GetOrCreateAsync.
2. Update all usage of Settings.Current to become Settings.GetOrCreateAsync; which will use a simple lock and check whether to return cached object or create a new instance.

Please confirm which implementation you recommended; or suggest another way.  thanks.

PR #2330, opened 2020-01-17
Add Network Test Report

philipktlin:
Add network controller report and generator; will enable in another PR.

### DIFF
--- a/test/modules/TestResultCoordinator/Reports/CountingReportMetadata.cs
+++ b/test/modules/TestResultCoordinator/Reports/CountingReportMetadata.cs
@@ -2,13 +2,14 @@
 namespace TestResultCoordinator.Reports
 {
     using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util;
 
     class CountingReportMetadata : ITestReportMetadata
     {
         public CountingReportMetadata(string expectedSource, string actualSource, TestOperationResultType testOperationResultType, TestReportType testReportType)
         {
-            this.ExpectedSource = expectedSource;
-            this.ActualSource = actualSource;
+            this.ExpectedSource = Preconditions.CheckNonWhiteSpace(expectedSource, nameof(expectedSource));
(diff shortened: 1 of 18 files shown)


test/modules/TestResultCoordinator/Reports/CountingReportMetadata.cs
```diff
@@ -2,13 +2,14 @@
...
     using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util;
 
     class CountingReportMetadata : ITestReportMetadata
     {
         public CountingReportMetadata(string expectedSource, string actualSource, TestOperationResultType testOperationResultType, TestReportType testReportType)
         {
-            this.ExpectedSource = expectedSource;
-            this.ActualSource = actualSource;
+            this.ExpectedSource = Preconditions.CheckNonWhiteSpace(expectedSource, nameof(expectedSource));
```
ancaantochi: nit: I think if it is an internal class we don't need to check the preconditions here, it should be checked when they are read from the metadata

test/modules/TestResultCoordinator/Reports/NetworkControllerReportGenerator.cs
```diff
@@ -0,0 +1,71 @@
...
+        internal string ResultType { get; }
+
+        public async Task<ITestResultReport> CreateReportAsync()
+        {
+            Logger.LogInformation($"Start to generate report by {nameof(NetworkControllerReportGenerator)} for Source [{this.Source}].");
+
+            var results = new List<TestOperationResult>();
+            bool hasResult = await this.TestResults.MoveNextAsync();
+
+            while (hasResult)
```
ancaantochi: because it always move to next you can use: while (await this.TestResults.MoveNextAsync())

test/modules/TestResultCoordinator/Reports/CountingReportMetadata.cs
philipktlin: leave it as-is; it is not harm to check.

PR #2339, opened 2020-01-21
contentTrust Add repo root key to AuthConfig

sozeng0:
This is based on the design which we think the root key will be put into registryCredntial. The code enable the ability to get root key into CombinedDockerConfig in PlanAsync

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/RegistryCredentials.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/RegistryCredentials.cs
@@ -15,6 +15,14 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker
             this.Password = Preconditions.CheckNonWhiteSpace(password, nameof(password));
         }
 
+        public RegistryCredentials(string address, string username, string password, string rootKey)
+        {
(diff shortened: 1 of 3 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/RegistryCredentials.cs
```diff
@@ -24,6 +32,9 @@ public RegistryCredentials(string address, string username, string password)
         [JsonProperty(Required = Required.Always, PropertyName = "password")]
         public string Password { get; }
 
+        [JsonProperty(PropertyName = "rootKey")]
```
kkdawkins: Is it required like address and password are?

sozeng0: No. it is not required.

arsing:
To ensure I remember, this still requires verification that there actually is such a thing as a registry-level root key. Sofie and Rajeev are doing that, so this shouldn't be merged for now.

(I don't believe there is, given that signing using `docker` generates repository-level keys, so just from `docker`'s implementation POV there shouldn't be any way for it to reuse another arbitrary repository's keys for this one.)

massand:
@sozeng0 - thanks for sharing the branch. We can go over details in your private repo.

PR #2344, opened 2020-01-21
Fix Windows build to compile against openssl 1.1.1

lfitchett:
Our build always uses latest vcpkg to install openssl.
vcpkg updated openssl from 1.0.2 to 1.1.1, which broke the Windows build:

- We've hard-coded the names of libssl and libcrypto import libs and DLLs in
  a few cmake files and the CAB package manifest file. These names are
  different in 1.1.1. This change updates the files to use the new names.

  This in turn means that we can no longer compile with 1.0.2,
  but we don't need to. If we do need that, we can always revert
  this change.

- Some of our libiothsm tests failed with 1.1.1. The tests expected a PEM blob
  to contain Unix newlines, but it actually contained Windows newlines.

  `read_file_into_buffer_impl` on Windows opens files for reading
  using win32 `CreateFile`, as opposed to POSIX `fopen`. This means the files
  would be read without newline translation, ie the equivalent of `rb` mode
  instead of `r`.

  However the code that used openssl BIOs, such as `write_certificate_file`,
  used `BIO_new_file` with modes like `r` and `w` that get passed down to
  `fopen` unchanged. This means these files would get opened
  with newline translation.

  The end result was that the code ended up writing a file with
  Windows newlines but believed it had not, which broke some tests
  that asserted one against the other.

  This change makes all calls to `BIO_new_file` specify `b` in the mode.

This change also removes the strawberry perl download from the Windows script
that installs openssl. We'd seen in the past that our build would break
because the original host that vcpkg pulled this file from from went down,
so we'd mirrored the file to our infra. However vcpkg has since started
using a new version of strawberry perl, so the one we downloaded ourselves
was being ignored and just wasting build time.


edgelet/build/windows/iotedge.wm.xml
```diff
@@ -14,8 +14,8 @@
     <file source="$(_REPO_ROOT)\edgelet\target\$(_Arch)\release\iotedged.pdb" destinationDir="$(runtime.programFiles)\iotedge" />
     <file source="$(_REPO_ROOT)\edgelet\target\$(_Arch)\release\iotedged_eventlog_messages.dll" destinationDir="$(runtime.programFiles)\iotedge" />
     <file source="$(_REPO_ROOT)\edgelet\hsm-sys\azure-iot-hsm-c\build\Release\iothsm.dll" destinationDir="$(runtime.programFiles)\iotedge" />
-    <file source="$(_OPENSSL_ROOT_DIR)\bin\libeay32.dll" destinationDir="$(runtime.programFiles)\iotedge" />
-    <file source="$(_OPENSSL_ROOT_DIR)\bin\ssleay32.dll" destinationDir="$(runtime.programFiles)\iotedge" />
+    <file source="$(_LIBSSL)" destinationDir="$(runtime.programFiles)\iotedge" />
```
arsing: Since you're switching to the new names, keep the original `$(_OPENSSL_ROOT_DIR)\...` form and just change the filenames.

arsing: Done.

edgelet/build/windows/iotedge.wm.xml
```diff
@@ -14,8 +14,8 @@
     <file source="$(_REPO_ROOT)\edgelet\target\$(_Arch)\release\iotedged.pdb" destinationDir="$(runtime.programFiles)\iotedge" />
     <file source="$(_REPO_ROOT)\edgelet\target\$(_Arch)\release\iotedged_eventlog_messages.dll" destinationDir="$(runtime.programFiles)\iotedge" />
     <file source="$(_REPO_ROOT)\edgelet\hsm-sys\azure-iot-hsm-c\build\Release\iothsm.dll" destinationDir="$(runtime.programFiles)\iotedge" />
-    <file source="$(_OPENSSL_ROOT_DIR)\bin\libeay32.dll" destinationDir="$(runtime.programFiles)\iotedge" />
-    <file source="$(_OPENSSL_ROOT_DIR)\bin\ssleay32.dll" destinationDir="$(runtime.programFiles)\iotedge" />
+    <file source="$(_OPENSSL_ROOT_DIR)\bin\libcrypto.dll" destinationDir="$(runtime.programFiles)\iotedge" />
```
arsing: These are wrong; the DLLs have an arch suffix like `-x64` in the filename.

arsing:
Successful packages build: https://dev.azure.com/msazure/One/_build/results?buildId=28140702&view=logs&j=db8f0d42-1e4a-5959-0d5f-4976b4d9df00

edgelet/build/windows/iotedge.wm.xml
```diff
@@ -14,8 +14,8 @@
     <file source="$(_REPO_ROOT)\edgelet\target\$(_Arch)\release\iotedged.pdb" destinationDir="$(runtime.programFiles)\iotedge" />
     <file source="$(_REPO_ROOT)\edgelet\target\$(_Arch)\release\iotedged_eventlog_messages.dll" destinationDir="$(runtime.programFiles)\iotedge" />
     <file source="$(_REPO_ROOT)\edgelet\hsm-sys\azure-iot-hsm-c\build\Release\iothsm.dll" destinationDir="$(runtime.programFiles)\iotedge" />
-    <file source="$(_OPENSSL_ROOT_DIR)\bin\libeay32.dll" destinationDir="$(runtime.programFiles)\iotedge" />
-    <file source="$(_OPENSSL_ROOT_DIR)\bin\ssleay32.dll" destinationDir="$(runtime.programFiles)\iotedge" />
+    <file source="$(_OPENSSL_ROOT_DIR)\bin\libcrypto-1_1-$(_OPENSSL_DLL_SUFFIX).dll" destinationDir="$(runtime.programFiles)\iotedge" />
```
lfitchett: why does the new build require the suffix when the old one didn't?

arsing: openssl sets the filenames, so we have to use what it gives us. If you remember even in your local amd64 build that there was an -x64 suffix.

PR #2347, opened 2020-01-22
Fix Windows build by downgrading openssl to 1.0.2

arsing:
Our build always uses latest vcpkg to install openssl.
vcpkg updated openssl from 1.0.2 to 1.1.1, which breaks our build:

- We've hard-coded the names of libssl and libcrypto import libs and DLLs in
  a few cmake files and the CAB package manifest file. These names are
  different in 1.1.1

- Some of our libiothsm tests fail with 1.1.1. This appears to be benign -
  the tests expect a PEM blob to contain Unix newlines, but they actually
  contain Windows newlines. (This is also why the tests are fine on Linux.)

Fixing the latter will take some time to investigate, so for now, this commit
unbreaks the Windows build by forcing vcpkg to a commit right before it updated
openssl to 1.1.1

### DIFF
--- a/edgelet/build/windows/openssl.ps1
+++ b/edgelet/build/windows/openssl.ps1
@@ -8,10 +8,14 @@ function Get-OpenSSL
 
     $ErrorActionPreference = 'Continue'
 
-    if (!((Test-Path -Path $env:HOMEDRIVE\vcpkg) -and ((Test-Path -Path $env:HOMEDRIVE\vcpkg\vcpkg.exe))))
+    # Override vcpkg directory from the default $env:HOMEDRIVE\vcpkg that VSTS CI machines have.
+    # See TODO below for the reason.
+    $vcpkgDirectory = "$env:HOMEDRIVE\vcpkg-2"
+
+    if (!((Test-Path -Path $vcpkgDirectory) -and ((Test-Path -Path $vcpkgDirectory\vcpkg.exe))))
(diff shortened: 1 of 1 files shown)


edgelet/build/windows/openssl.ps1
```diff
@@ -8,40 +8,52 @@ function Get-OpenSSL
...
         Write-Host "Bootstrapping vcpkg..."
-        & "$env:HOMEDRIVE\vcpkg\bootstrap-vcpkg.bat"
+        & "$vcpkgDirectory\bootstrap-vcpkg.bat"
         if ($LastExitCode)
         {
             Throw "Failed to bootstrap vcpkg with exit code $LastExitCode"
         }
         Write-Host "Installing vcpkg..."
-        & $env:HOMEDRIVE\\vcpkg\\vcpkg.exe integrate install
+        & $vcpkgDirectory\vcpkg.exe integrate install
```
philipktlin: do you need double quote here?

arsing: No, otherwise the original code won't have worked either.

arsing:
I fixed the tests in https://github.com/Azure/iotedge/pull/2344/commits/29ed6d7660cbcdf0aa772354e89deec4997e7890 , so we can go with https://github.com/Azure/iotedge/pull/2344 after all.

PR #2349, opened 2020-01-22
[k8s] Put IDeploymentBackupSource back in k8s module.

darobs:
We lost registration of IDeploymentBackupSource  in a recent merge, this puts it back.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/KubernetesModule.cs
```diff
@@ -227,6 +227,17 @@ protected override void Load(ContainerBuilder builder)
                 .As<IRuntimeInfoSource>()
                 .SingleInstance();
 
+            // Task<IBackupSource>
```
dmolokanov:
```suggestion
            // Task<IDeploymentBackupSource>
```

Issue #2306, opened 2020-01-14
iotedge debian purge leaves data in '/var/lib/iotedge'


gri6507:
After installing iotedge debian, then manually configuring `/etc/iotedge/config.yaml`, followed by `systemctl restart iotedge`, and then waiting for the IoT Edge device to become fully functional, it looks like `iotedge` creates some content in '/var/lib/iotedge' directory. However, during `apt purge iotedge`, that content is not cleaned up, causing 

```bash
dpkg: warning: while removing iotedge, directory '/var/lib/iotedge' not empty so not removed 

gri6507@abyss:~>ls -al /var/lib/iotedge
total 16
drwxr-xr-x  4 iotedge iotedge 4096 Jan 13 15:51 .
drwxr-xr-x 38 root    root    4096 Jan 13 14:33 ..
drwxr-xr-x  2 iotedge iotedge 4096 Jan 13 15:51 cache
drwxr-xr-x  5 iotedge iotedge 4096 Jan 13 14:34 hsm
```

arsing:
Yes, that is the home directory of the `iotedge` user, and contains working files required for the `iotedge` service at runtime.

The files must be kept across updates, but can be removed when the package is uninstalled or purged.

I see it's currently listed in [`debian/dirs`](https://github.com/Azure/iotedge/blob/b1e6ba543f5be7b4c6cb9b6d7d7f2d97b401e203/edgelet/contrib/debian/dirs) along with the logs directory. Perhaps they should be created in `postinst` and removed in `postrm` instead?

gri6507:
> Perhaps they should be created in `postinst` and removed in `postrm` instead?

That would make sense.

gri6507:
I created https://github.com/Azure/iotedge/pull/2350 for this

arsing:
Fixed by #2350

PR #2350, opened 2020-01-22
Azure/iotedge issue #2306 - cleanup /var/lib/iotedge

gri6507:
PROBLEM STATEMENT:
`apt purge iotedge` causes dpkg: warning: while removing iotedge, directory '/var/lib/iotedge' not empty so not removed

SOLUTION:
Create /var/lib/iotedge in postinst and cleanup in postrm

DETAILS:
For debian, removed creation of /var/lib/iotedge from dirs and moved it
into postinst.

For debian and debian8, remove /var/lib/iotedge in postrm

TESTING:
Built using package.sh for both ubuntu18.04 and debian8. Tested in local ubuntu 18.04 based docker container.


gri6507:
I'm not sure how to add the automerge_label

arsing:
Do we want to delete `/var/log/iotedge` too? I don't know what Debian convention is.

Actually, with this change, does `apt uninstall` not complain about not deleting `/var/log/iotedge` like it used to complain about `/var/lib/iotedge` before? Or is it that it special-cases directories under `/var/log` ?


>I'm not sure how to add the automerge_label

You can't add the label. Only committers can do that.

gri6507:
> Do we want to delete `/var/log/iotedge` too? I don't know what Debian convention is.
> 
> Actually, with this change, does `apt uninstall` not complain about not deleting `/var/log/iotedge` like it used to complain about `/var/lib/iotedge` before? Or is it that it special-cases directories under `/var/log` ?

It is actually `apt purge` which now does the cleanup of  `/var/lib/iotedge`. With regards to `/var/log/iotedge` cleanup, apparently my testing did not have any logs in that directory because `apt purge` was not complaining about that directory being not empty. However, when I tested the debian behavior in a simple test deb package which puts content in `/var/log/foo`, the act of purging does, in fact, complain that it cannot delete `/var/log/foo` because it is non-empty. So, I am going to update this PullRequest to treat the `/var/log/iotedge` directory in a similar way.

arsing:
Yeah sorry, I meant `apt purge`, not `apt uninstall`.

I couldn't find any search results to confirm where `purge`ing should also delete log files or not. Mike and I talked offline and we think it's fine to do it anyway.

I'll queue up a packages build and test manually, then we can merge this.

PR #2355, opened 2020-01-22
Hash device and module id before upload

lfitchett:
This is required to comply with GDPR. It replaces any device Ids with "device". Device Id's are replaced instead of hashed because compliance determined that deviceId hashes must use a rotating salt. 

Module id is simply hashed.

It also changes host module tag to module_name to match agent metrics


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricTransformer.cs
```diff
@@ -61,7 +70,7 @@ public MetricFilter AddAllowedTags(params KeyValuePair<string, string>[] pairs)
             return this;
         }
 
-        public MetricFilter AddTagsToAdd(params KeyValuePair<string, string>[] pairs)
+        public MetricTransformer AddTagsToAdd(params KeyValuePair<string, string>[] pairs)
         {
             if (this.tagsToAdd.HasValue)
```
and-rewsmith: nit: w1 could be more descriptive name

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricTransformer.cs
```diff
@@ -30,13 +31,21 @@ public IEnumerable<Metric> FilterMetrics(IEnumerable<Metric> metrics)
                     continue;
```
and-rewsmith: So allowed tags will bypass transformation? What is the reason for naming this "allowed"? What about "tagsIndicatingNoTransform"

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsWorker.cs
```diff
@@ -125,6 +126,32 @@ async Task BeginUploadRetry(CancellationToken cancellationToken)
...
+        string ReplaceDeviceId(string id)
+        {
+            const string deviceIdReplacement = "device";
+
+            // Id is in the form of 'deviceId/moduleId'
+            string[] parts = id.Split('/');
+            if (parts.Length == 2)
+            {
+                parts[0] = deviceIdReplacement;
+                parts[1] = parts[1].CreateSha256(); // Hash moduleId
```
damonbarry: Any benefit to **NOT** hashing the module ID if it's a system module (i.e., begins with `$`)?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricTransformer.cs
lfitchett: nope, this acts as a whitelist. Only metrics that contain the given tags will be allowed through, all others are dropped. It used to be called whitelist, but was renamed.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsWorker.cs
lfitchett: easier to read, I'll change it.

PR #2357, opened 2020-01-23
Make auto_generated_cert_lifetime not required (#2356)

kkdawkins:
Fix bug: Auto generated certificate lifetime configuration was required which would break updating of Edge deployments.


edgelet/edgelet-core/src/settings.rs
```diff
@@ -331,6 +332,10 @@ pub struct DeviceCertificate {
     trusted_ca_certs: String,
 }
 
+fn default_auto_generated_ca_lifetime_days() -> u16 {
+    90
```
darobs: Would it make sense to use `DEFAULT_AUTO_GENERATED_CA_LIFETIME_DAYS` here?

PR #2359, opened 2020-01-23
Metrics Collector: Add host platform to scraped metrics

and-rewsmith:
We should tag metrics with the host platform of the test agent. I have passed this host platform information from the pipeline yaml definitions into the metrics collector where I tag the metrics.


test/connectivity/scripts/connectivityTest.sh
```diff
@@ -69,6 +69,7 @@ function prepare_test_from_artifacts() {
     sed -i -e "s@<MetricsCollector.MetricsEndpointsCSV>@$METRICS_ENDPOINTS_CSV@g" "$deployment_working_file"
```
philipktlin: btw some of these metric parameters are required, right? if so, please check if it is provided.

scripts/linux/runE2ETest.sh
```diff
@@ -1040,6 +1045,7 @@ function usage() {
     echo ' -metricsEndpointsCSV              Optional csv of exposed endpoints for which to scrape metrics.'
     echo ' -metricsScrapeFrequencyInSecs     Optional frequency at which the MetricsCollector module will scrape metrics from the exposed metrics endpoints. Default is 300 seconds.'
     echo ' -metricsUploadTarget              Optional upload target for metrics. Valid values are AzureLogAnalytics or IoTHub. Default is AzureLogAnalytics.'
+    echo ' -hostPlatform                     Describes the host OS and cpu architecture.'
```
philipktlin: In general, it is possible to get host platform for linux and RPi by script, instead of passing in.  It is required to pass in for Windows since it has many different skus.

scripts/windows/test/Run-E2ETest.ps1
```diff
@@ -316,6 +319,8 @@ Param (
 
     [string] $TwinUpdateFailureThreshold = $null,
 
+    [switch] $HostPlatform = $null,
```
philipktlin: Please validate host platform is provided.

scripts/linux/runE2ETest.sh
and-rewsmith: I would rather keep it consistent. There is also the benefit of matching the names directly to the pipeline.

scripts/windows/test/Run-E2ETest.ps1
and-rewsmith: I also did this in the runE2E.sh

scripts/windows/test/Run-E2ETest.ps1
```diff
@@ -1495,6 +1501,7 @@ Function ValidateTestParameters
         If ([string]::IsNullOrEmpty($SnitchAlertUrl)) {Throw "Required snith alert URL."}
         If ([string]::IsNullOrEmpty($SnitchStorageAccount)) {Throw "Required snitch storage account."}
         If ([string]::IsNullOrEmpty($SnitchStorageMasterKey)) {Throw "Required snitch storage master key."}
+        If ([string]::IsNullOrEmpty($HostPlatform)) {Throw "Required host platform."}
```
philipktlin: Are you going to remove this line here and check only if test is stress or long haul, right?

and-rewsmith: Wait, isn't that what it is currently doing?

philipktlin: Sorry; the code is collapsed and confused me.

PR #2360, opened 2020-01-23
upload logs in TRC

philipktlin:
After test completes, send upload log request from TRC to edgeAgent; and add log blob container uri in test summary.

Tested and it can successfully upload the log to azure blob container.

### DIFF
--- a/test/modules/TestResultCoordinator/TestReportUtil.cs
+++ b/test/modules/TestResultCoordinator/TestReportUtil.cs
@@ -3,9 +3,16 @@ namespace TestResultCoordinator
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using System.Threading;
     using System.Threading.Tasks;
+    using Azure.Storage.Blobs;
+    using Microsoft.Azure.Devices.Client;
     using Microsoft.Azure.Devices.Edge.Util;
     using Microsoft.Extensions.Logging;
+    using Microsoft.WindowsAzure.Storage;
+    using Microsoft.WindowsAzure.Storage.Blob;
     using Newtonsoft.Json;
     using Newtonsoft.Json.Linq;
     using TestResultCoordinator.Reports;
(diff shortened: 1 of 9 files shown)


test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -103,9 +110,100 @@ internal static List<ITestReportMetadata> ParseReportMetadataJson(string reportM
             return reportMetadataList;
         }
 
+        internal static async Task<Uri> GetOrCreateBlobContainerSasUriForLogAsync(string storageAccountConnectionString)
```
dylanbronson: Can we move all upload logic to a new class? This one will get bloated, and TestReportUtil is already a generic name

test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -103,9 +110,100 @@ internal static List<ITestReportMetadata> ParseReportMetadataJson(string reportM
...
+        }
+
+        internal static async Task UploadLogsAsync(string iotHubConnectionString, Uri blobContainerWriteUri, ILogger logger)
+        {
+            Preconditions.CheckNonWhiteSpace(iotHubConnectionString, nameof(iotHubConnectionString));
+            Preconditions.CheckNotNull(blobContainerWriteUri, nameof(blobContainerWriteUri));
+            Preconditions.CheckNotNull(logger, nameof(logger));
+
+            DateTime uploadLogStartAt = DateTime.UtcNow;
+            logger.LogInformation("Send upload logs request to edgeAgent.");
```
dylanbronson:
```suggestion
            logger.LogInformation("Sending upload logs request to edgeAgent.");
```

test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -103,9 +110,100 @@ internal static List<ITestReportMetadata> ParseReportMetadataJson(string reportM
...
+
+            ModuleClient moduleClient = await ModuleClient.CreateFromEnvironmentAsync(TransportType.Amqp_Tcp_Only);
```
dylanbronson: This has to be AMQP? Shouldn't it be whatever protocol we're using to send upstream?

test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -103,9 +110,100 @@ internal static List<ITestReportMetadata> ParseReportMetadataJson(string reportM
...
+            var uploadLogRequest = new MethodRequest(
+                "UploadLogs",
+                Encoding.UTF8.GetBytes($"{{ \"schemaVersion\": \"1.0\", \"sasUrl\": \"{blobContainerWriteUri.AbsoluteUri}\", \"items\": [{{ \"id\": \".*\", \"filter\": {{}} }}], \"encoding\": \"gzip\", \"contentType\": \"json\" }}"));
+            MethodResponse uploadLogResponse = await moduleClient.InvokeMethodAsync(Settings.Current.DeviceId, "$edgeAgent", uploadLogRequest);
+
+            (string status, string correlationId) = GetUploadLogResponseResult(uploadLogResponse.ResultAsJson);
+            logger.LogInformation($"Upload logs response: status={status}, correlationId={correlationId}");
+
+            int checkUploadStatusPeriod = 60 * 1000;    // 1 min
```
dylanbronson: Do we expect the blob storage upload to take a minute?

test/modules/TestResultCoordinator/TestReportUtil.cs
philipktlin: This is used for module client to connect to edgeHub and routing to edgeAgent on IoT Edge device.  And this is not related to the test; therefore I chose to use AMQP for this request.

test/modules/TestResultCoordinator/TestReportUtil.cs
philipktlin: it depends log file size and number of modules; since we have up to 20 modules and debug log level is used.  I can change it to 30s.

PR #2364, opened 2020-01-23
Twin tester fix

dylanbronson:
Adding a startup delay to TwinTester module's settings. 

Currently TwinTester could start sending DesiredProperty updates before EdgeHub or IoTHub is fully aware of the desired property update subscription. This leads to a problem in our longhaul tests, where it seems like we get a desired property update failure near startup time. 

This change will take care of that problem, as TwinTester will now wait 2 min before sending desired property updates. This should be ample time for EdgeHub to fully start up and subscribe to the desired property updates with IoTHub.


test/modules/TwinTester/config/settings.json
```diff
@@ -8,5 +8,6 @@
     "serviceClientConnectionString": "",
     "StoragePath": "",
     "StorageOptimizeForPerformance": true,
-    "TwinTestMode": "TwinAllOperations"
+    "TwinTestMode": "TwinAllOperations",
+    "TestStartDelay":  "00:02:00"
```
varunpuranik: I suppose we already have the code to use the parameter if it exists?

dylanbronson: Yes. This will just set it by default

PR #2365, opened 2020-01-23
Fix E2E: Pass edge agent base image name to IoTEdgeSecurityDaemon.ps1

and-rewsmith:
I accidentally broke E2E because I didn't catch the fact that IoTEdgeSecurityDaemon.ps1 needs an agent image passed in. I fix this by passing edge agent 1.0.

### DIFF
--- a/smoke/IotEdgeQuickstart/details/Details.cs
+++ b/smoke/IotEdgeQuickstart/details/Details.cs
@@ -259,7 +259,7 @@ namespace IotEdgeQuickstart.Details
 
                     return new DeviceProvisioningMethod(connectionString);
                 });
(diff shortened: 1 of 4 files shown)


smoke/IotEdgeQuickstart/details/Details.cs
```diff
@@ -511,25 +511,30 @@ async Task CreateEdgeDeviceIdentity(RegistryManager rm)
...
         }
 
         string TempSensorImage()
         {
-            return this.BuildImageName("azureiotedge-simulated-temperature-sensor");
+            return this.BuildImageName("azureiotedge-simulated-temperature-sensor", this.imageTag);
         }
 
-        string BuildImageName(string name)
+        string BuildImageName(string name, string customImageTag)
```
and-rewsmith: nit: will change back to imageTag because in the case where

PR #2368, opened 2020-01-23
Update dotnet task in image build

philipktlin:
Update dotnet 3.0 to use latest .net core task in image build pipeline.


darobs:
Do we want to update CI build, too?
builds/ci/dotnet.yaml

philipktlin:
Sure. Will do.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #2376, opened 2020-01-24
IoTEdgeQuickstart: use mcr agent image before deployment

and-rewsmith:
There is a known bug upon re-deployment where edge agent fails to update configuration settings (env vars, createOptions, etc.) without the image field changing.

Our test framework is affected by this bug because it first replaces the config yaml with the to-be-tested edge agent image. Edgelet downloads this edge agent image and spins it up, disregarding any additional configuration. When edge agent comes up, it should update its env vars and createOptions, but does not because the image has not changed.

This fix will deploy "mcr.microsoft.com/azureiotedge-agent:1.0" first instead of the to-be-tested image, then have the desired state applied through a deployment.

I also removed Iotedgectl.cs because iotedgectl (python tool) has been deprecated in favor of the iotedged security daemon.


smoke/IotEdgeQuickstart/details/IBootstrapper.cs
```diff
@@ -13,7 +13,7 @@ public interface IBootstrapper
 
         Task Install();
 
-        Task Configure(DeviceProvisioningMethod method, string image, string hostname, string deviceCaCert, string deviceCaPk, string deviceCaCerts, LogLevel runtimeLogLevel);
+        Task Configure(DeviceProvisioningMethod method, string hostname, string deviceCaCert, string deviceCaPk, string deviceCaCerts, LogLevel runtimeLogLevel);
```
philipktlin: Please add comments here to mention the root cause for not using provided image tag in this step.  The root cause is because edgeAgent will not be triggered a new deployment if same image tag is used; even there are environment variable changes in the edgeAgent config.  And mentioned this is a bug in edgeAgent, it will be better if you can find the corresponding github issue.

and-rewsmith: Couldn't find a relevant issue. Do we even have this logged anywhere?

smoke/IotEdgeQuickstart/details/IotedgedWindows.cs
```diff
@@ -175,11 +175,6 @@ public async Task Configure(DeviceProvisioningMethod method, string image, strin
                 args += method.ManualConnectionString.Map(
                     cs => { return $" -Manual -DeviceConnectionString '{cs}'"; }).GetOrElse(string.Empty);
 
-                foreach (RegistryCredentials c in this.credentials)
-                {
-                    args += $" -Username '{c.User}' -Password (ConvertTo-SecureString '{c.Password}' -AsPlainText -Force)";
-                }
-
```
and-rewsmith: We don't need to define the image manually if we remove this information.These credentials are only needed for edge agent image pull

and-rewsmith:
This can now be merged. I have verified linux is fixed for VM proxy. Windows is having an unrelated issue also in master.

PR #2384, opened 2020-01-25
[k8s] Improve proxy logging

dmolokanov:
- Include proxy request URL even in case of an error in logs
- Fix the way how proxy builds destination URL for the service

### DIFF
--- a/edgelet/iotedge-proxy/src/proxy/client.rs
+++ b/edgelet/iotedge-proxy/src/proxy/client.rs
@@ -74,7 +74,7 @@ where
                 // add authorization header with bearer token to authenticate request
                 if let Some(token) = self.config.token().get() {
                     let token = HeaderValue::from_str(format!("Bearer {}", token).as_str())
-                        .context(ErrorKind::HeaderValue("Authorization".to_owned()))?;
+                        .with_context(|_| ErrorKind::HeaderValue("Authorization".to_owned()))?;
 
                     req.headers_mut().insert(header::AUTHORIZATION, token);
                 }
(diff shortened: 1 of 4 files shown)


edgelet/iotedge-proxy/src/proxy/client.rs
```diff
@@ -103,21 +103,28 @@ where
...
                     .get(header::CONTENT_LENGTH)
                     .and_then(|length| length.to_str().ok().map(ToString::to_string))
                     .unwrap_or_else(|| "-".to_string());
 
-                info!("\"{}\" {} {}", request, res.status(), body_length);
-
-                res
-            });
+                info!(
+                    "Success: \"{}\" {} {}",
```
darobs: >Success [](start = 21, length = 7)

"Response completed"?

dmolokanov: Make changes to this part

edgelet/iotedge-proxy/src/proxy/client.rs
```diff
@@ -94,6 +87,31 @@ where
     }
 }
 
+fn build_uri(destination_url: Url, requested_uri: &Uri) -> Result<Uri, Error> {
+    let path = percent_decode(requested_uri.path().as_bytes())
+        .decode_utf8()
+        .context(ErrorKind::Uri(requested_uri.to_string()))?;
```
arsing: `.with_context`, because you don't want to run `.to_string()` eagerly.

Same for the next two calls.

edgelet/iotedge-proxy/src/proxy/client.rs
```diff
@@ -94,6 +87,31 @@ where
     }
 }
 
+fn build_uri(destination_url: Url, requested_uri: &Uri) -> Result<Uri, Error> {
```
arsing: Maybe rename `destination_url` to `base_url` (since from the POV of the caller it's just the base URL), and rename the `url` on line 104 to `destination_url` (since inside the function it becomes the destination URL) ?

edgelet/iotedge-proxy/src/proxy/client.rs
dmolokanov: Good idea!

PR #2385, opened 2020-01-25
Prefetch edgehub config from twin store

ancaantochi:
To speed up edgeHub starting while offline it won't wait for the config if it already has it in the twin store. 
Fixed update callback call to also update the currentConfig and check if it has Updates


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
```diff
@@ -82,6 +82,26 @@ public async Task<IMessage> GetTwinAsync(string id)
                 });
         }
 
+        public async Task<Option<IMessage>> GetLocalTwinAsync(string id)
```
varunpuranik: Nit: GetCachedTwinAsync?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
```diff
@@ -82,6 +82,26 @@ public async Task<IMessage> GetTwinAsync(string id)
...
+        {
+            return await this.TwinStore.Match(
```
varunpuranik: If you instead use Map, it will automatically return an Option, no?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
```diff
@@ -82,6 +82,26 @@ public async Task<IMessage> GetTwinAsync(string id)
...
+                async (store) =>
+                {
+                    Option<TwinInfo> cached = await this.TwinStore.Match(s => s.Get(id), () => Task.FromResult(Option.None<TwinInfo>()));
```
varunpuranik: Store

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -24,24 +24,44 @@ public class ConfigUpdater
         Option<EdgeHubConfig> currentConfig;
         Option<IConfigSource> configProvider;
 
-        public ConfigUpdater(Router router, IMessageStore messageStore, TimeSpan configUpdateFrequency, IStorageSpaceChecker storageSpaceChecker)
+        private ConfigUpdater(Router router, IMessageStore messageStore, TimeSpan configUpdateFrequency, IStorageSpaceChecker storageSpaceChecker)
```
varunpuranik: Nit: avoid private modifier

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -24,24 +24,44 @@ public class ConfigUpdater
...
         {
             this.router = Preconditions.CheckNotNull(router, nameof(router));
             this.messageStore = messageStore;
             this.configUpdateFrequency = configUpdateFrequency;
             this.storageSpaceChecker = Preconditions.CheckNotNull(storageSpaceChecker, nameof(storageSpaceChecker));
         }
 
+        public static async Task<ConfigUpdater> Create(Router router, IMessageStore messageStore, TimeSpan configUpdateFrequency, IStorageSpaceChecker storageSpaceChecker, Option<EdgeHubConfig> initialConfig)
```
varunpuranik: Nit: I think it will be cleaner to add a ConfigUpdaterProvider that first gets the local config and then inits ConfigUpdater with it.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -24,24 +24,44 @@ public class ConfigUpdater
...
+        {
+            var instance = new ConfigUpdater(router, messageStore, configUpdateFrequency, storageSpaceChecker);
+            await instance.UpdateConfig(initialConfig);
```
varunpuranik: Should we do this in Init? Else can we get rid of Init?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core.Config
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class EdgeHubDesiredProperties
+    {
+        [JsonConstructor]
+        public EdgeHubDesiredProperties(string schemaVersion, IDictionary<string, string> routes, StoreAndForwardConfiguration storeAndForwardConfiguration)
```
varunpuranik: Hmm, EdgeHub config need not come from the twin. 
So this should be EdgeHubConfig and the other one can be EdgeHubParsedConfig or something like that..

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/LocalTwinConfigSource.cs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core.Config
+{
+    using System;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    public class LocalTwinConfigSource : IConfigSource
```
varunpuranik: I wonder if it will be cleaner to make this a TwinConfigSource that always gets the twin from the TwinManager (either cached or the regular call based on a flag), and bypass the EdgeHubConnection altogether.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/RoutingModule.cs
```diff
@@ -512,8 +512,24 @@ protected override void Load(ContainerBuilder builder)
...
+                        RouteFactory routeFactory = await c.Resolve<Task<RouteFactory>>();
                         Router router = await c.Resolve<Task<Router>>();
-                        var configUpdater = new ConfigUpdater(router, messageStore, this.configUpdateFrequency, storageSpaceChecker);
+                        var twinManagerTask = c.Resolve<Task<ITwinManager>>();
+                        var twinMessageConverter = c.Resolve<Core.IMessageConverter<Twin>>();
+                        var twinManager = await twinManagerTask;
+                        Option<EdgeHubConfig> prefetchConfig;
+                        if (this.useTwinConfig)
+                        {
+                            var localConfigSource = new LocalTwinConfigSource(edgeHubCredentials.Identity.Id, twinManager, twinMessageConverter, routeFactory);
```
varunpuranik: We should move this out of registration. Suggest something like ConfigUpdaterProvider

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
ancaantochi: the issue I had with Map is when twin is null I could not return Option.None

ancaantochi: I would return a Option<Task<Option<IMessage>>>

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
ancaantochi: This is used when it comes from twin to deserialize the twin properties.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/config/ConfigUpdaterTest.cs
```diff
@@ -25,6 +25,124 @@ public class ConfigUpdaterTest
...
+
+            // Act
+            var configUpdater = new ConfigUpdater(router, messageStore.Object, updateFrequency, storageSpaceChecker.Object);
+            await configUpdater.Init(configProvider.Object);
+
+            // Assert
+            // First only has updated from prefeched config
+            Assert.Equal(2, router.Routes.Count);
+
+            // After 10 seconds updates from init received
```
vipeller: 10 or 6? :)

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/config/ConfigUpdaterTest.cs
```diff
@@ -25,6 +25,124 @@ public class ConfigUpdaterTest
...
+            // Act
+            var configUpdater = new ConfigUpdater(router, messageStore.Object, updateFrequency, storageSpaceChecker.Object);
+            await configUpdater.Init(configProvider.Object);
+
+            // Assert
+            Assert.Equal(2, router.Routes.Count);
+
+            // After 5 seconds no updates
+            await Task.Delay(TimeSpan.FromSeconds(5));
+            Assert.Equal(2, router.Routes.Count);
```
vipeller: The delay at the GetConfig() mock is also 5 seconds which is the same time we wait here before check. Maybe the result is 2 becuse the GetConfig() hasn't finished yet?

ancaantochi: increased to 6 secs

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/config/ConfigUpdaterTest.cs
ancaantochi: 6, updated comment

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/TwinManager.cs
```diff
@@ -82,6 +82,24 @@ public async Task<IMessage> GetTwinAsync(string id)
                 });
         }
 
+        public async Task<Option<IMessage>> GetCachedTwinAsync(string id)
+        {
+            return await this.TwinStore.Match(
```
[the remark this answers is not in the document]
ancaantochi: not needed

PR #2387, opened 2020-01-27
Metrics Collector: Retry sending failed upload messages

and-rewsmith:
When using log analytics, the Metrics Collector module does not attempt to retry failed uploaded messages. Here, I make a change so that it attempts to upload the previously failed metrics before each new scrape.


edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -38,18 +41,44 @@ public void Dispose()
...
         {
+            IEnumerable<Metric> scrapedMetrics = Enumerable.Empty<Metric>();
             try
             {
-                IEnumerable<Metric> metrics = await this.scraper.ScrapeEndpointsAsync(cancellationToken);
-                await this.publisher.PublishAsync(this.GetGuidTaggedMetrics(metrics), cancellationToken);
+                await this.PublishPreviouslyFailedMetricsAsync(cancellationToken);
+
+                scrapedMetrics = await this.scraper.ScrapeEndpointsAsync(cancellationToken);
+                scrapedMetrics = this.GetGuidTaggedMetrics(scrapedMetrics);
```
and-rewsmith: nit: might want to do renaming here instead of assigning twice

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -38,18 +41,44 @@ public void Dispose()
...
+                {
+                    Logger.LogError(e, "Error uploading metrics from prior scrape");
+                    break;
+                }
+            }
+
+            this.failedUploadMetrics.RemoveRange(0, Math.Min(processedCount, this.failedUploadMetrics.Count));
+        }
+
         IEnumerable<Metric> GetGuidTaggedMetrics(IEnumerable<Metric> metrics)
```
philipktlin: Please rename "ApplyAdditionalMetricTags"; or suggest to change it as extension method and name it as "WithAdditionalMetricTags".

edge-modules/MetricsCollector/MetricsScrapeAndUpload.cs
```diff
@@ -38,18 +41,44 @@ public void Dispose()
...
+
+                scrapedMetrics = await this.scraper.ScrapeEndpointsAsync(cancellationToken);
+                scrapedMetrics = this.GetGuidTaggedMetrics(scrapedMetrics);
+                await this.publisher.PublishAsync(scrapedMetrics, cancellationToken);
                 Logger.LogInformation("Successfully scraped and uploaded metrics");
             }
             catch (Exception e)
             {
                 Logger.LogError(e, "Error scraping and uploading metrics");
+                this.failedUploadMetrics.AddRange(scrapedMetrics);
```
philipktlin: publish logic of publisher is done by batch, right? For example. when exception happens at batch x; then batch 1 to (x-1) is sent and should not resend again.  Metrics in batch x or later should be added to failed metrics and will resend later on.  If my understanding is correct, you should not add all scraped metrics to failed upload metrics.

and-rewsmith:
Closing as I am re-using the retry logic from the diagnostics module.

PR #2393, opened 2020-01-28
Add .NET code coverage to the checkin build

damonbarry:
The change enables code coverage analysis for our .NET Core code in the checkin build, and (similar to our iotedged checkin build) fails the build if % line coverage is below a certain threshold. Current coverage is measuring at 61% so I set the threshold to 60%.

Note this change relies on the built-in code coverage functionality that has recently been added in .NET Core. Downsides are: it's currently only available on Windows, and it produces a .coverage file which has to be opened in Visual Studio. Upsides are: this is the "officially supported" option for Azure DevOps, so while support is minimal now, future improvements to analytics in DevOps would likely come for free. Also, because this option works out-of-the-box in the dotnet tooling, I was able to remove a **bunch** of gunk from our project files.

The other option would be to use the 3rd-party [Coverlet](https://github.com/tonerdo/coverlet) tool, which would allow us to run on Linux (might shave a couple minutes off job time) and outputs in Cobertura format just like our Rust code coverage. More details [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/dotnet-core?view=azure-devops#collect-code-coverage-metrics-with-coverlet). So we'd have a more consistent experience between languages. But we'd have to add back a bunch of project references in our test projects for arguably minimal benefit since this change seems to give us everything we need, at least for now.

This change also adds the "Unit" category to 132 unit tests which were not categorized.


edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Core.Test/DeploymentConfigInfoTest.cs
```diff
@@ -3,8 +3,10 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Core.Test
 {
     using System.Collections.Generic;
     using Microsoft.Azure.Devices.Edge.Agent.Core.Test.ConfigSources;
+    using Microsoft.Azure.Devices.Edge.Util.Test.Common;
     using Xunit;
 
+    [Unit]
```
arsing: So all these tests weren't running at all? Did you discover this because the code coverage tool flagged these files as untested?

damonbarry: They weren't running in the gated builds because we're always careful to specify a filter. I guess they'd run if you just do a vanilla `dotnet test` at your desk.

I noticed one assembly was reporting 0% coverage even though it had tests. That led me to open up VS Test Explorer, group tests by category, and expand the "uncategorized" tree. :)

damonbarry: I don't know of a way to flag uncategorized tests, but that would be nice...

builds/checkin/dotnet.yaml
```diff
@@ -67,3 +67,44 @@ jobs:
...
+    variables:
+      coverage.goal: 64
+    pool:
+      vmImage: windows-2019
+    steps:
+      - task: DotNetCoreCLI@2
+        displayName: Build test projects
+        inputs:
+          command: build
+          projects: edge-*\test\Microsoft.Azure.Devices.*Test\Microsoft.Azure.Devices.*Test.csproj
```
philipktlin: we also have unit tests in test\modules\Modules.Test\Modules.Test.csproj

builds/checkin/dotnet.yaml
```diff
@@ -67,3 +67,44 @@ jobs:
           testRunner: VSTest
           testResultsFiles: '**/TestResults/*.trx'
         condition: succeededOrFailed()
+
+################################################################################
+  - job: code_coverage
+################################################################################
+    displayName: Code coverage
+    variables:
+      coverage.goal: 64
```
philipktlin: suggest to define it in pipeline variable instead.

builds/checkin/dotnet.yaml
```diff
@@ -67,3 +67,44 @@ jobs:
...
+          command: build
+          projects: edge-*\test\Microsoft.Azure.Devices.*Test\Microsoft.Azure.Devices.*Test.csproj
+          arguments: '-o target'
+      - task: VSTest@2
+        displayName: Run unit tests with code coverage
+        inputs:
+          testSelector: testAssemblies
+          testAssemblyVer2: |
+            target\Microsoft.Azure*Test.dll
+          testFiltercriteria: Category=Unit
```
philipktlin: Since code coverage only runs unit test, what if a feature is added, but it cannot be easily tested by unit test; but integration test is added.  Therefore the code coverage percentage may go below the goal.  Or we agreed that we should always refactor our code to have some unit tests.  Also do we plan to adding more tests to improve test coverage?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Microsoft.Azure.Devices.Edge.Agent.Core.csproj
```diff
@@ -3,23 +3,10 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
-    <Configurations>Debug;Release;CodeCoverage;CheckInBuild</Configurations>
+    <Configurations>Debug;Release;CheckInBuild</Configurations>
```
philipktlin: Thanks for cleaning up this configuration.

builds/checkin/dotnet.yaml
damonbarry: Oh right, that throws this glob spec off...let me rework it.

builds/checkin/dotnet.yaml
damonbarry: The iotedged code coverage goal is a job variable; I'm following the existing pattern. Why make it a pipeline variable if only one job needs it?

builds/checkin/dotnet.yaml
damonbarry: A couple thoughts about this:
- Integration tests often require secrets, which is why they don't run in the checkin build. So I can't measure integration tests here.
- I'm only interested in understanding our coverage for unit tests because unit tests are the primary means of "covering" all our features with tests. In most cases, if something can't easily be tested with unit tests it should be refactored until it can. Integration tests are for testing our integration with isolated external systems. I know it's common to write an integration test when unit testing is hard, but I don't want to encourage that behavior.

builds/checkin/dotnet.yaml
philipktlin: so you can change it by editing the pipeline (not query time) instead of yaml change.

builds/checkin/dotnet.yaml
philipktlin: I am on your side.

builds/checkin/dotnet.yaml
damonbarry: Ah, I see what you're saying. But I think there's value in storing coverage goals in source control, alongside the code. That way if you send an older commit through our build system, the goal value will match the state of our code at the time of that commit.

If you publish a PR that increases coverage, it's easy enough to bump the goal in that PR as well.

builds/checkin/dotnet.yaml
damonbarry: Fixed. Note that I went from building just test project to building everything because it's actually faster. 😃

PR #2403, opened 2020-01-28
C2D Connectivity test modules

dylanbronson:
Adding the modules for C2D connectivity tests.

These are the CloudMessageSender and the CloudMessageReceiver. They are very simple. CloudMessageSender sends a message to a device using the ServiceClient, and CloudMessageReceiver receives the message using the DeviceClient.

Both will report the message to the TestResultCoordinator (after sending and receiving respectively).

Still need to add them to the deployment. That will be in another PR


test/modules/CloudToDeviceMessageTester/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    class Program
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger("CloudMessageSender");
```
philipktlin:
```suggestion
        static readonly ILogger Logger = ModuleUtil.CreateLogger("CloudToDeviceMessageTester");
```

test/modules/CloudToDeviceMessageTester/Receiver.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public override async Task InitAsync(CancellationTokenSource cts, DateTime testStartAt)
+        {
+            Microsoft.Azure.Devices.RegistryManager registryManager = null;
+            try
+            {
+                registryManager = Microsoft.Azure.Devices.RegistryManager.CreateFromConnectionString(this.iotHubConnectionString);
+                await registryManager.AddDeviceAsync(new Microsoft.Azure.Devices.Device(this.deviceId + "_leaf"), cts.Token);
+                this.deviceClient = DeviceClient.CreateFromConnectionString(this.iotHubConnectionString, this.deviceId, Microsoft.Azure.Devices.Client.TransportType.Amqp_Tcp_Only);
```
dylanbronson: Think I need to append ";GatewayHostname=127.0.0.1" to connectionString here. Will confirm soon

test/modules/CloudToDeviceMessageTester/Receiver.cs
```diff
@@ -0,0 +1,71 @@
...
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Client;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults;
+    using Microsoft.Extensions.Logging;
+    using TransportType = Microsoft.Azure.Devices.TransportType;
+
+    sealed class Receiver : CloudToDeviceMessageTesterBase
```
philipktlin:
```suggestion
    sealed class CloudToDeviceMessageReceiver : CloudToDeviceMessageTesterBase
```

test/modules/CloudToDeviceMessageTester/Receiver.cs
```diff
@@ -0,0 +1,71 @@
...
+            var testResultReceived = new MessageTestResult(this.moduleId + ".receive", DateTime.UtcNow)
+            {
+                TrackingId = trackingId,
+                BatchId = batchId,
+                SequenceNumber = sequenceNumber
+            };
+            await ModuleUtil.ReportTestResultAsync(this.testResultReportingClient, this.logger, testResultReceived);
+        }
+
+        public override async Task InitAsync(CancellationTokenSource cts, DateTime testStartAt)
```
philipktlin:
```suggestion
        public override async Task StartAsync(CancellationTokenSource cts, DateTime testStartAt)
```

test/modules/CloudToDeviceMessageTester/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+            catch (Exception ex)
+            {
+                Logger.LogError(ex, $"Error occurred during CloudToDeviceMessageTester while in {Settings.Current.TestMode} mode.");
+            }
+            finally
+            {
+                // Implicit CloseAsync()
+                cloudToDeviceMessageTester?.Dispose();
+            }
+
```
philipktlin:
```suggestion

await cts.Token.WhenCanceled();
```

test/modules/CloudToDeviceMessageTester/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+                        Settings.Current.ModuleId,
+                        Settings.Current.TransportType,
+                        Settings.Current.TestDuration,
+                        reportClient,
+                        Settings.Current.TrackingId,
+                        Settings.Current.MessageDelay,
+                        Settings.Current.TestStartDelay);
+                }
+
+                await cloudToDeviceMessageTester.InitAsync(cts, testStartAt);
```
philipktlin: don't need testStartAt parameter; just define inside InitAsync method

test/modules/CloudToDeviceMessageTester/Receiver.cs
```diff
@@ -0,0 +1,71 @@
...
+            TestResultReportingClient testResultReportingClient)
+            : base(logger, iotHubConnectionString, deviceId, moduleId, transportType, testDuration, testResultReportingClient)
+        {
+        }
+
+        public override void Dispose() => this.deviceClient?.Dispose();
+
+        public async Task ReportTestResult(Message message)
+        {
+            string trackingId = message.Properties[TestConstants.Message.TrackingIdPropertyName];
```
philipktlin: think about if this can be shared with existing message scenario; do it in another PR.

test/modules/CloudToDeviceMessageTester/Receiver.cs
```diff
@@ -0,0 +1,71 @@
...
+        public override async Task InitAsync(CancellationTokenSource cts, DateTime testStartAt)
+        {
+            Microsoft.Azure.Devices.RegistryManager registryManager = null;
+            try
+            {
+                registryManager = Microsoft.Azure.Devices.RegistryManager.CreateFromConnectionString(this.iotHubConnectionString);
+                await registryManager.AddDeviceAsync(new Microsoft.Azure.Devices.Device(this.deviceId + "_leaf"), cts.Token);
+                this.deviceClient = DeviceClient.CreateFromConnectionString(this.iotHubConnectionString, this.deviceId, Microsoft.Azure.Devices.Client.TransportType.Amqp_Tcp_Only);
+                await this.deviceClient.OpenAsync();
+                while (!cts.IsCancellationRequested && this.IsTestTimeUp(testStartAt))
```
philipktlin: for receiver, you don't even need to check test ends

test/modules/CloudToDeviceMessageTester/Receiver.cs
```diff
@@ -0,0 +1,71 @@
...
+                {
+                    this.logger.LogInformation("Ready to receive message");
+                    Message message = await this.deviceClient.ReceiveAsync();
```
philipktlin: what happens if there is not message sending to this module; will this line throw exception? e.g. timeout exception.

test/modules/DirectMethodSender/Program.cs
```diff
@@ -28,7 +28,7 @@ public static async Task<int> MainAsync()
             {
                 Guid batchId = Guid.NewGuid();
                 Logger.LogInformation($"Batch Id={batchId}");
-                Logger.LogInformation($"Load gen delay start for {Settings.Current.TestStartDelay}.");
+                Logger.LogInformation($"Direct Method Sender delay start for {Settings.Current.TestStartDelay}.");
```
philipktlin: please send out another PR to fix this line.

test/modules/CloudToDeviceMessageTester/Settings.cs
```diff
@@ -0,0 +1,105 @@
...
+            Preconditions.CheckArgument(Enum.IsDefined(typeof(CloudToDeviceMessageTesterMode), testMode));
+
+            this.DeviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
+            this.IoTHubConnectionString = Preconditions.CheckNonWhiteSpace(iotHubConnectionString, nameof(iotHubConnectionString));
+            this.TestMode = testMode;
+            this.ModuleId = Preconditions.CheckNonWhiteSpace(moduleId, nameof(moduleId));
+            this.TrackingId = Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
+            this.TransportType = transportType;
+            this.MessageDelay = messageDelay;
+            this.ReportingEndpointUrl = reportingEndpointUrl;
```
philipktlin: Preconditions.CheckNotNull

test/modules/CloudToDeviceMessageTester/Receiver.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace CloudToDeviceMessageTester
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Client;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults;
+    using Microsoft.Extensions.Logging;
+    using TransportType = Microsoft.Azure.Devices.TransportType;
```
yophilav: Why are there duplication of `TransportType`?

test/modules/CloudToDeviceMessageTester/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+                        Settings.Current.IoTHubConnectionString,
+                        Settings.Current.DeviceId,
+                        Settings.Current.ModuleId,
+                        Settings.Current.TransportType,
+                        Settings.Current.TestDuration,
+                        reportClient);
+                }
+                else
+                {
+                    cloudToDeviceMessageTester = new Sender(
```
yophilav: Nit: `Receiver` and `Sender` are too generic and can be confusing if the constructors is used with other module client. Please rename to be more indicative of the `CloudToDeviceMessage` properties of these modules.

test/modules/CloudToDeviceMessageTester/Program.cs
yophilav: suggest passing in `cts.Token` (a cancellation) instead of the `cts` which is the source of cancellation.

test/modules/CloudToDeviceMessageTester/Receiver.cs
```diff
@@ -0,0 +1,71 @@
...
+            await ModuleUtil.ReportTestResultAsync(this.testResultReportingClient, this.logger, testResultReceived);
+        }
+
+        public override async Task InitAsync(CancellationTokenSource cts, DateTime testStartAt)
+        {
+            Microsoft.Azure.Devices.RegistryManager registryManager = null;
+            try
+            {
+                registryManager = Microsoft.Azure.Devices.RegistryManager.CreateFromConnectionString(this.iotHubConnectionString);
+                await registryManager.AddDeviceAsync(new Microsoft.Azure.Devices.Device(this.deviceId + "_leaf"), cts.Token);
```
yophilav: Isn't there an upper limit of how long a deviceName can be? Please double check me and make sure that by adding "_leaf" would not exceed the current deviceName length.

test/modules/CloudToDeviceMessageTester/Receiver.cs
yophilav: Did you miss a `Token` here? cts.`Token`.IsCancellationRequested

Though, this may work.

test/modules/CloudToDeviceMessageTester/Settings.cs
yophilav: I thought `URI` is a non-nullable type?

test/modules/DirectMethodSender/Program.cs
yophilav: Type.

test/modules/CloudToDeviceMessageTester/Receiver.cs
dylanbronson: (as we discussed in person): It will. And then it will bubble up to program, where it is caught, logged, and test is ended. That will signify a problem with our setup, so failing fast is a good thing here.

test/modules/CloudToDeviceMessageTester/Receiver.cs
dylanbronson: Changed to use token

test/modules/CloudToDeviceMessageTester/Receiver.cs
dylanbronson: Apparently it should be fine with this length, as it is quite short. If there is a problem we will catch it at the very start of the test.

test/modules/CloudToDeviceMessageTester/Receiver.cs
dylanbronson: Need the full name since it conflicts with another TransportType somewhere else. This one is for deviceClient specifically which can only take AMQP

test/modules/CloudToDeviceMessageTester/CloudToDeviceMessageReceiver.cs
```diff
@@ -0,0 +1,83 @@
...
+                await this.deviceClient.OpenAsync();
+
+                while (!ct.IsCancellationRequested)
+                {
+                    this.logger.LogInformation("Ready to receive message");
+                    try
+                    {
+                        Message message = await this.deviceClient.ReceiveAsync();
+                        this.logger.LogInformation($"Message received. " +
+                        $"Sequence Number: {message.Properties[TestConstants.Message.SequenceNumberPropertyName]}, " +
```
philipktlin: nit: alignment; should have a tab space for new line of a method call.  Please try to build in release configuration to make sure no style error.

dylanbronson: dotnet build -c release didn't care about this, but I agree it is easier to read when tabbed in, so I updated it.

PR #2409, opened 2020-01-28
run connectivity test with different configs

philipktlin:
use strategy in connectivity yaml to run different configs


builds/e2e/templates/connectivity-deploy.yaml
```diff
@@ -73,6 +77,14 @@ steps:
           -edgeletBranchName "${{ parameters['edgelet.branchName'] }}" \
           -hostPlatform "${{ parameters['metricsCollector.hostPlatform'] }}" \
           -storageAccountConnectionString "${{ parameters['testResultCoordinator.storageAccountConnectionString'] }}" \
+          -devOpsAccessToken $(System.AccessToken) \
```
philipktlin: Need these 2 parameter to query rest api for build cancel

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -455,6 +471,12 @@ function usage() {
     exit 1;
 }
 
+is_build_canceled=$(is_cancel_build_requested $DEVOPS_ACCESS_TOKEN $DEVOPS_BUILDID)         
+if [ $is_build_canceled -eq 1 ]; then
+    print_highlighted_message "build is canceled."
+    exit 3
```
philipktlin: use exit code 3 as it is the first custom exit code to use; for details, http://tldp.org/LDP/abs/html/exitcodes.html.

builds/e2e/connectivity.yaml
```diff
@@ -11,7 +11,33 @@ jobs:
 ################################################################################
     displayName: Linux AMD64 Moby
     condition: and(eq(variables['run.linux.amd64.moby'], 'true'), ne(variables['agent.group'], ''))
-    timeoutInMinutes: 120
+    timeoutInMinutes: 180
+    strategy:
+      matrix:
```
yophilav: Is there a way to decide if we want to use matrix strategy or not?
Say... I want to only test the `Offline(5/20/3)` case without going through the rest of the cases.

builds/e2e/templates/connectivity-deploy.yaml
yophilav: whose accessToken are you using? Can we have a generic one for the team use?

philipktlin: This is a security token used by the running build. https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#systemaccesstoken

builds/e2e/connectivity.yaml
philipktlin: I added a pipeline variable “config.to.run”; you can specify a number between 1-6 to run a specific config.

test/connectivity/scripts/testHelper.sh
```diff
@@ -117,3 +131,10 @@ function print_highlighted_message() {
     local color_reset='\033[0m'
     echo -e "${cyan}$message${color_reset}"
 }
+
+function stop_iotedge_service() {
+    echo 'Stop IoT Edge services'
+    systemctl stop iotedge.socket iotedge.mgmt.socket || true
+    systemctl kill iotedge || true
+    systemctl stop iotedge || true
+}
```
and-rewsmith: Why can't you only do ```systemctl stop iotedge```?

builds/e2e/connectivity.yaml
```diff
@@ -22,14 +48,33 @@ jobs:
...
+            output1=$(curl -s -u :$accessToken --request GET "https://dev.azure.com/msazure/one/_apis/build/builds/$buildId?api-version=5.1" | grep -oe '"status":"cancel')
+            output2=$(curl -s -u :$(System.AccessToken) --request GET "https://dev.azure.com/msazure/one/_apis/build/builds/$(Build.BuildId)/Timeline?api-version=5.1" | grep -oe '"result":"canceled"')
+            
+            if [[ ( "$(testrun.config)" == "$(config.to.run)" || "$(config.to.run)" == "0" ) && ( -z "$output1" ) && ( -z "$output2" ) ]]; then
+              echo "##vso[task.setvariable variable=run.steps]1"
+              echo "run.steps=1"
+            else
+              echo "##vso[task.setvariable variable=run.steps]0"
+              echo "run.steps=0"
+            fi
```
and-rewsmith: Just confirming my understanding: This step allows the test to be cancelled before the next matrix entry gets run?

builds/e2e/templates/connectivity-deploy.yaml
```diff
@@ -73,6 +77,14 @@ steps:
...
+          -devOpsBuildId $(Build.BuildId) \
```
and-rewsmith: This field already exists, right? See testBuildNumber on line 75.

test/connectivity/scripts/testHelper.sh
philipktlin: I just refactor and extract the original implementation.  We can discuss with Arnav and handle it in different PR

builds/e2e/connectivity.yaml
philipktlin: All matrix entry will be run and it will skip all steps if build is canceled

builds/e2e/templates/connectivity-deploy.yaml
philipktlin: Build number e.g. 20200201.3 is different from build Id, e.g. 11263738.

PR #2411, opened 2020-01-29
Vipeller/agent reconnect

vipeller:
Added connectivity test to Edge Agent in order to detect network problems earlier and mitigate reconnection times


vipeller:
not needed anymore

PR #2426, opened 2020-01-29
Fix windows e2e metric tests

lfitchett:
Metrics tests break on windows. This hopefully fixes that.


test/Microsoft.Azure.Devices.Edge.Test/Metrics.cs
```diff
@@ -48,9 +48,15 @@ async Task Deploy(CancellationToken token)
...
                     {
                         builder.AddModule(ModuleName, metricsValidatorImage);
-                        builder.GetModule("$edgeHub")
+
+                        var edgeHub = builder.GetModule("$edgeHub")
                             .WithEnvironment(("experimentalfeatures__enabled", "true"), ("experimentalfeatures__enableMetrics", "true"))
                             .WithDesiredProperties(new Dictionary<string, object> { { "routes", new { All = "FROM /messages/* INTO $upstream" } } });
+                        if (OsPlatform.IsWindows())
+                        {
+                            edgeHub.WithSettings(("User", "ContainerAdministrator"));
```
damonbarry: Needs to be under "createOptions"

test/Microsoft.Azure.Devices.Edge.Test/Metrics.cs
```diff
@@ -26,7 +26,10 @@ public async Task ValidateMetrics()
             CancellationToken token = this.TestToken;
             await this.Deploy(token);
 
-            var result = await this.iotHub.InvokeMethodAsync(Context.Current.DeviceId, ModuleName, new CloudToDeviceMethod("ValidateMetrics"), token);
+            // System resource metrics take 1 minute to start. Wait before testing
+            await Task.Delay(TimeSpan.FromMinutes(1.1));
+
+            var result = await this.iotHub.InvokeMethodAsync(Context.Current.DeviceId, ModuleName, new CloudToDeviceMethod("ValidateMetrics"), CancellationToken.None);
```
damonbarry: You went back to using `CancellationToken.None`. Should we continue to use `token` instead?

lfitchett: yes, don't know how that went back

Issue #1562, opened 2019-08-13
1.0.9 RC 1 Default sock configuration doesn't work properly


mma-mm:
Hi,

i just tried to make Verison 1.0.9 RC1 of the IoT Edge Security work.
In the default config.yaml, the sockets are configured like this:

    connect:
      management_uri: "unix:///var/lib/iotedge/mgmt.sock"
      workload_uri: "unix:///var/lib/iotedge/workload.sock"
    
    listen:
      management_uri: "unix:///var/lib/iotedge/mgmt.sock"
      workload_uri: "unix:///var/lib/iotedge/workload.sock"

whilst according to iotedge --help the default connect socket is:
    
    unix:///var/run/iotedge/mgmt.sock
    
so, commands like iotedge list won't work if  `-H unix:///var/lib/iotedge/mgmt.sock`is not provided.

The default configuration should work out of the box.

## Expected Behavior
After setting the connection string / dps information and iotedge started, `iotedge list` should give a result like this:
    
    NAME             STATUS           DESCRIPTION      CONFIG
    edgeAgent        running          Up 29 minutes    mcr.microsoft.com/azureiotedge-agent:1.0

    

## Current Behavior
Call to `iotedge list` never returns.

## Steps to Reproduce

1. pull, build and install iotedge security daemon Version 1.0.9 RC1
2. change config.yaml such that it uses dps with x509-certificates.
3. restart iotedge service
4. attempt to list modules with `iotedge list` -> won't return, needs to be canceled using ctrl+c
5. call `iotedge list -H unix:///var/lib/iotedge/mgmt.sock` -> the desired output is given.

## Context (Environment)

Ubuntu 18.04 (AMD64)
iotedge 1.0.9-rc1

### Output of `iotedge check`

<details>
<summary>Click here</summary>

```
Configuration checks
--------------------
√ config.yaml is well-formed - OK
× config.yaml has well-formed connection string - Error
    Device is not using manual provisioning, so Azure IoT Hub hostname needs to be specified with --iothub-hostname
√ container engine is installed and functional - OK
√ config.yaml has correct hostname - OK
× config.yaml has correct URIs for daemon mgmt endpoint - Error
    Unable to find image 'mcr.microsoft.com/azureiotedge-diagnostics:1.0.9-rc1' locally
    docker: Error response from daemon: manifest for mcr.microsoft.com/azureiotedge-diagnostics:1.0.9-rc1 not found: manifest unknown: manifest unknown.
[...]

PR #2428, opened 2020-01-30
Update edgelet devguide.

arsing:
- Talk about building the packages as an alternative to building the binaries.

  Discussed in https://github.com/Azure/iotedge/pull/2301

- Building binaries on Windows requires VS / VSBT and cmake.

- Both `OPENSSL_DIR` and `OPENSSL_ROOT_DIR` need to be set for
  macOS and Windows.

- Take into account that builds don't necessarily use "stable", but whatever
  happens to be in the rust-toolchain file.

- `cargo build --all` spends time building Kubernetes dependencies.
  `cargo build -p iotedged -p iotedge` is sufficient.

- Talk about how to actually run the binaries.

- Update some URLs, such as from the merger of the rust-lang-nursery and
  rust-lang GitHub orgs.


gri6507:
Can you please add a quick sentence to the `Building OS packages` section that when re-building for different `PACKAGE_ARCH`, you should do a `make clean`

Otherwise, this looks good. Thank you!

arsing:
Ah yes, forgot about that. Done now.

arsing:
@darobs @damonbarry @dmolokanov Ping.

edgelet/doc/devguide.md
```diff
@@ -1,163 +1,264 @@
...
+
+After installing `rustup`, or if you already have it installed, install the toolchain that will be used to build the daemon binaries.
+
+```sh
+rustup self update   # Update rustup itself to the latest version
+
+git clone --recurse-submodules 'https://github.com/Azure/iotedge'
+cd iotedge/edgelet/
+
+rustup update   # Install / update the toolchain used to build the daemon binaries. This is controlled by the rust-toolchain file in this directory.
```
dmolokanov: This line is too wide and user will need to scroll right to read the full comment. Should we re-arrange comment that the user doesn't need to do additional scrolling?

edgelet/doc/devguide.md
```diff
@@ -1,163 +1,264 @@
...
+1. Set the `OPENSSL_DIR` and `OPENSSL_ROOT_DIR` environment variables to point to the local openssl installation.
 
-	```powershell
-	.\vcpkg install openssl:x64-windows
-	```
+    ```powershell
+    # $PWD is the root of the vcpkg repository
 
-1. Set `OPENSSL_ROOT_DIR`
+    $env:OPENSSL_DIR = "$PWD\installed\x64-windows"
```
dmolokanov: Shouldn't both of them to lead to openssl lib folder inside "$PWD\installed\x64-windows"? I remember having problems with doing as the old doc says.

edgelet/doc/devguide.md
```diff
@@ -1,163 +1,264 @@
...
+This will create `iotedged` and `iotedge` binaries under `edgelet/target/debug`
+
+
+### Run
+
+To run `iotedged` locally:
+
+1. Create a directory that it will use as its home directory, such as `~/iotedge`
+
+    - Linux
```
dmolokanov:
```suggestion
    - Linux / macOS
```

edgelet/doc/devguide.md
```diff
@@ -1,163 +1,264 @@
...
+
+    ```sh
+    java -jar swagger-codegen-cli.jar generate -i api/workload.yaml -l rust -o {root}/edgelet/workload
+    ```
+
+    Note that we've manually fixed up the generated code so that it satisfies rustfmt and clippy. As such, if you ever need to run `swagger-codegen-cli` against new definitions, or need to regenerate existing ones, you will want to perform the same fixups manually. Make sure to run clippy and rustfmt against the new code yourself, and inspect the diffs of modified files before checking in.
+
+- IDE
+
+    VS Code has good support for Rust. Consider installing the following extensions:
```
dmolokanov: Should we mentioned IntelliJ IDEA with Rust plugin?

edgelet/doc/devguide.md
```diff
@@ -1,163 +1,264 @@
...
+
+1. `PACKAGE_ARCH`: This is the architecture of the OS on which the resulting packages will be installed. It should be one of `amd64`, `arm32v7` or `aarch64`.
+
+For example:
+
+```sh
+git clone --recurse-submodules 'https://github.com/Azure/iotedge'
+cd iotedge/
+
+PACKAGE_OS='debian10' PACKAGE_ARCH='arm32v7' ./edgelet/build/linux/package.sh
```
dmolokanov: We also have a [Makefile](https://github.com/Azure/iotedge/blob/master/edgelet/Makefile) to do things with source code. Should we mention the `make release` command rather than exact script call?

edgelet/doc/devguide.md
arsing: The path for `libcrypto.dll` must be `$env:OPENSSL_DIR\bin\libcrypto.dll`

See https://github.com/Azure/iotedge/blob/70f9fbe7ea18de4360351b99dfd2288c1111ee7b/edgelet/build/windows/openssl.ps1#L60-L61

edgelet/doc/devguide.md
arsing: The point of this change is the makefile is the wrong thing to use when you want to make packages.

Eg with `make release` on Debian 8 you'll end up using the wrong value for `HOST_URI_OPTION` (the default empty one).

Eg if you go further ahead and use `make install` on any Debian or Ubuntu, you'll end up having a misconfigured iotedged, because despite its generic-sounding name `make install` is only used for CentOS. ( https://github.com/Azure/iotedge/issues/1562 )

The script is there precisely to build the packages properly. That's why the preamble lists two ways of working on iotedged - use the script if you just want packages, or use `cargo` if you want to hack the code.

edgelet/doc/devguide.md
arsing: I don't know anything about it. What do you want it to say?

edgelet/doc/devguide.md
arsing: Yeah, good point. Done.

edgelet/doc/devguide.md
arsing: Done.

edgelet/doc/devguide.md
dmolokanov: I see. It is confusing. Thanks for explanation.
Wonder if we need to spend some time and fix makefile in a way it does what it is intended to do.

arsing: I don't think I'd want to bake OS detection in a makefile. The script can be more deterministic because it knows which OS it's spawning a Docker container for.

The script also helps because you don't have to install and configure the build dependencies, esp cross-compiler toolchains. Also, for some of the targets, the build needs to run in a Docker image anyway (the ones that use Linaro images).

edgelet/doc/devguide.md
dmolokanov: > Alternatively, [Intellij IDEA Community Edition](https://www.jetbrains.com/idea) with [Rust plugin](https://intellij-rust.github.io) provides full IDE experience for programming in Rust.

Something like this maybe?

edgelet/doc/devguide.md
dmolokanov: Makes sense. Agree

edgelet/doc/devguide.md
dmolokanov: Ok, maybe it was something else.

edgelet/doc/devguide.md
arsing: Done.

PR #2435, opened 2020-01-30
Create [EndToEnd] test category

damonbarry:
While not strictly necessary, people have asked that end-to-end tests be organized in their own category. This change adds the NUnit `EndToEnd` attribute, and adds it to all the existing end-to-end tests.

### DIFF
--- a/edge-util/test/Microsoft.Azure.Devices.Edge.Util.Test.Common/nunit/EndToEndAttribute.cs
+++ b/edge-util/test/Microsoft.Azure.Devices.Edge.Util.Test.Common/nunit/EndToEndAttribute.cs
@@ -1,0 +1,10 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Util.Test.Common.NUnit
+{
+    using System;
+    using global::NUnit.Framework;
+
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true)]
+    public class EndToEndAttribute : CategoryAttribute
+    { }
+}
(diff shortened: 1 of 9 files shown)


edge-util/test/Microsoft.Azure.Devices.Edge.Util.Test.Common/nunit/EndToEndAttribute.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Util.Test.Common.NUnit
+{
+    using System;
+    using global::NUnit.Framework;
+
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true)]
```
philipktlin: should we force that all tests in 1 test file should belong to the same category? if so, we can make it only allow to use on Class level, but not method.  And AllowMultiple should be false, right?

damonbarry: > should we force that all tests in 1 test file should belong to the same category?
I don't think we need to enforce that.

> AllowMultiple should be false, right?
It's true in all the attributes we currently define. I don't have an opinion one way or the other, so I just followed the existing pattern.

philipktlin: Sure. let's unblock this PR.  And we can decide if we need another PR to set "AllowMultiple to false".  I still think it is better to only set in class level to enforce all tests in a test class should belong to the same category.

PR #2436, opened 2020-01-30
Add callback to set TLS recovation flag for AQMP test on Linux

richma-ms:


test/Microsoft.Azure.Devices.Edge.Test.Common/linux/OsPlatform.cs
```diff
@@ -43,8 +45,18 @@ public async Task<CaCertificates> GenerateCaCertificatesAsync(string deviceId, s
...
             this.GetEdgeQuickstartCertificates("/var/lib/iotedge/hsm");
 
-        public void InstallCaCertificates(IEnumerable<X509Certificate2> certs, ITransportSettings _) =>
-            this.InstallTrustedCertificates(certs, StoreName.Root);
+        public void InstallCaCertificates(IEnumerable<X509Certificate2> certs, ITransportSettings transportSettings)
+        {
+            if (transportSettings.GetTransportType() == TransportType.Amqp_WebSocket_Only ||
+                transportSettings.GetTransportType() == TransportType.Amqp_Tcp_Only)
+            {
+                transportSettings.SetupCertificateValidation(certs.First());
```
varunpuranik: Why do we need diff behavior for AMQP?

richma-ms: this is only an issue with AMQP.  For MQTT, I think it's better to preserve the behavior of installing the CA cert, so that our test reflects the customer scenarios.

---
In reply to: [373151134](https://github.com/Azure/iotedge/pull/2436#discussion_r373151134) [](ancestors = 373151134)

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/OsPlatform.cs
```diff
@@ -4,12 +4,14 @@ namespace Microsoft.Azure.Devices.Edge.Test.Common.Linux
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
```
dmolokanov: some imports maybe unnecessary

PR #2438, opened 2020-01-30
Report error to TRC

philipktlin:
Allow to report unexpected error message to TRC for follow up action; will generate a new error report to list out reported errors in another PR.


yophilav:
Why do we need a errorReport? Shouldn't the error report already be parted of each test result to TRC?

philipktlin:
> Why do we need a errorReport? Shouldn't the error report already be parted of each test result to TRC?

Sometimes there are unexpected behavior may happen but it is not the focus of the test. We just add error message to module log.  For now, everyone will just look at the report and no one will aware of those unexpected behavior.

For example, messaging scenario. We expect the message sender should be always able to send messages to edgeHub and to simplify report generation; sequence number will only be increased if message is successfully sent.

PR #2442, opened 2020-01-30
Eliminate redundant deployments from end-to-end tests

damonbarry:
We recently split up end-to-end test deployments into two configs: one for just the system modules and another for all modules. The problem is that our tests will double-deploy even when the config _only_ contains system modules.

This change only produces/deploys the second config if it differs from the first. Tests that only deploy $edgeAgent and $edgeHub won't do it twice.


test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
```diff
@@ -52,9 +52,11 @@ public IModuleConfigBuilder AddModule(string name, string image)
...
         }
 
-        /* Will output edgeHub and edgeAgent configurations, then a full configuration at the end.
-           This is done to assure routes are set up for the most recent $edgeHub deployment before the test modules start sending messages (i.e. assure messages won't get dropped).
-           Another way to handle this is to define all possible routes at the very beginning of the test, but there is added complexity as module names are assigned dynamically. */
+        // Returns two configurations: one with just $edgeAgent and $edgeHub; the other with
+        // everything. This is done to ensure edgeHub's routes are ready before the test modules
+        // start sending messages, to avoid dropped messages. Another way to handle this is to
+        // define all possible routes at the beginning of the test run, but there is added
+        // complexity as module names are assigned dynamically.
```
and-rewsmith: nit: stylecop might complain about no /* syntax

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
```diff
@@ -75,19 +76,29 @@ public IEnumerable<EdgeConfiguration> BuildConfigurationStages()
             var moduleNames = new List<string>();
             var moduleImages = new List<string>();
 
+            var modulesLookup = modules.ToLookup(m => m.IsSystemModule() ? "system" : "other");
```
and-rewsmith: I'm assuming this was done to avoid enumerating over the modules twice?

test/Microsoft.Azure.Devices.Edge.Test.Common/config/ModuleConfiguration.cs
```diff
@@ -20,5 +20,10 @@ public ModuleConfiguration(string name, string image, IReadOnlyDictionary<string
             this.Image = image;
             this.Name = name;
         }
+
+        public bool IsSystemModule()
+        {
+            return this.Name.StartsWith('$');
```
and-rewsmith: smart!

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
damonbarry: StyleCop doesn't complain. Actually I thought that our team coding standard was the opposite (prefer '//' to C-style '/* */'), although I can't find that written anywhere right now...I guess I was just confused. Anyway, if you look for instances of C-style comments in our codebase, the vast majority of them were added by newer devs in the last 6 months. 😃

PR #2446, opened 2020-02-01
Updating SDK 

dylanbronson:
Updating the service and device sdk's to the latest versions.
Latest versions found here: https://github.com/Azure/azure-iot-sdk-csharp/releases
Microsoft.Azure.Devices.Client 1.18.2 -> 1.18.4
Microsoft.Azure.Devices.Client 1.21.3 -> 1.22.0

Also updated Newtonsoft.Json 12.0.1 -> 12.0.3 where applicable.

Also had to resolve a version conflict regarding azure-functions-vs-build-sdk. It uses Newtonsoft.Json 11.0.2. I manually added 12.0.3 as it says to do in the FAQ on its github:
https://github.com/Azure/azure-functions-vs-build-sdk

This resulted in more warnings for us.


edge-modules/MetricsCollector/MetricsCollector.csproj
```diff
@@ -1,64 +1,64 @@
...
-  <ItemGroup>
-    <AdditionalFiles Include="..\..\stylecop.json" Link="stylecop.json" />
-  </ItemGroup>
-  <PropertyGroup>
-    <CodeAnalysisRuleSet>..\..\stylecop.ruleset</CodeAnalysisRuleSet>
-  </PropertyGroup>  
-  <Import Project="..\..\stylecop.props" />
-
-</Project>
+<Project Sdk="Microsoft.NET.Sdk">
```
varunpuranik: Is this some formatting change? Can we revert it?

dylanbronson: IDE did it automatically. I tried to, but there is still some difference that Git detects that we can't see. I don't think it is an issue

PR #2448, opened 2020-02-02
E2E: Support windows arm32

and-rewsmith:
We are currently initializing the iotedge installation with agent image 1.0 on startup. This won't work for windows arm32 (or any future new platform under development) so I am adding support for a quickstart arg "initialize-with-agent-artifact" that can be used to initialize the installation with the desired agent artifact, skipping 1.0.


and-rewsmith:
Successful Win arm32 test: 
https://msazure.visualstudio.com/One/_releaseProgress?_a=release-environment-logs&releaseId=1741313&environmentId=4613930

smoke/IotEdgeQuickstart/Quickstart.cs
```diff
@@ -34,16 +35,18 @@ public class Quickstart : Details.Details
...
             Option<DPSAttestation> dpsAttestation)
-            : base(bootstrapper, credentials, iothubConnectionString, eventhubCompatibleEndpointWithEntityPath, upstreamProtocol, proxy, imageTag, deviceId, hostname, deploymentFileName, twinTestFileName, deviceCaCert, deviceCaPk, deviceCaCerts, optimizedForPerformance, runtimeLogLevel, cleanUpExistingDeviceOnSuccess, dpsAttestation)
+            : base(bootstrapper, credentials, iothubConnectionString, eventhubCompatibleEndpointWithEntityPath, upstreamProtocol, proxy, imageTag, deviceId, hostname, deploymentFileName, twinTestFileName, deviceCaCert, deviceCaPk, deviceCaCerts, optimizedForPerformance, initializeWithAgentArtifact, runtimeLogLevel, cleanUpExistingDeviceOnSuccess, dpsAttestation)
         {
             this.leaveRunning = leaveRunning;
             this.noVerify = noVerify;
             this.bypassEdgeInstallation = bypassEdgeInstallation;
             this.verifyDataFromModule = verifyDataFromModule;
             this.dpsProvisionTest = dpsAttestation.HasValue;
+            this.initializeWithAgentArtifact = initializeWithAgentArtifact;
```
philipktlin: why do you need this? it is already passed to base class.

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
```diff
@@ -174,9 +174,17 @@ public Task Install()
                 300); // 5 min timeout because install can be slow on raspberry pi
         }
 
-        public async Task Configure(DeviceProvisioningMethod method, string hostname, string deviceCaCert, string deviceCaPk, string deviceCaCerts, LogLevel runtimeLogLevel)
+        public async Task Configure(DeviceProvisioningMethod method, Option<string> agentImage, string hostname, string deviceCaCert, string deviceCaPk, string deviceCaCerts, LogLevel runtimeLogLevel)
         {
-            Console.WriteLine($"Setting up iotedged with agent image 1.0");
+            agentImage.ForEach(
```
philipktlin: is it better to combine this ForEach with ForEach on line 232?

and-rewsmith: I thought about this too, but liked this executing at the very beginning.

smoke/IotEdgeQuickstart/Quickstart.cs
and-rewsmith: Good catch, I don't

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
philipktlin: nit: should put together; will make sure any updates will not miss the other?  your call.

and-rewsmith: My only concern is the unhappy path where something fails. If we move the log statement down then we will see less information in this error case.

PR #2452, opened 2020-02-03
Test edge hub metrics

lfitchett:
This adds tests for edgeHub metrics. It validates number of messages sent to hub, number sent upstream and queue length. 

It also consolidates the module tag (module => module_name)


test/modules/MetricsValidator/src/Program.cs
```diff
@@ -37,10 +37,19 @@ static async Task<int> MainAsync()
                     .AddEnvironmentVariables()
                     .Build();
 
+                if (Environment.OSVersion.Platform != PlatformID.Unix)
```
damonbarry: Elsewhere in the tests, the pattern is
```csharp
    if (OsPlatform.IsWindows()) { ... }
```

test/modules/MetricsValidator/src/Program.cs
```diff
@@ -37,10 +37,19 @@ static async Task<int> MainAsync()
...
+                {
+                    Logger.LogInformation("Windows detected, delaying");
```
damonbarry: We want to be _very_ tidy and concise in the information-level logs. So either remove these `LogInformation` statements or at least move them to `LogDebug`.

test/modules/MetricsValidator/src/Program.cs
```diff
@@ -37,10 +37,19 @@ static async Task<int> MainAsync()
...
+                    await Task.Delay(TimeSpan.FromMinutes(2));
```
damonbarry: Are you going to figure out what is necessitating this delay before merging this PR?

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
```diff
@@ -97,6 +98,9 @@ async Task SeedMetrics(CancellationToken cancellationToken)
             await this.moduleClient.SetMethodHandlerAsync(methodName, (_, __) => Task.FromResult(new MethodResponse(200)), null);
             await this.moduleClient.InvokeMethodAsync(Environment.GetEnvironmentVariable("IOTEDGE_DEVICEID"), Environment.GetEnvironmentVariable("IOTEDGE_MODULEID"), new MethodRequest(methodName), cancellationToken);
 
+            await this.moduleClient.UpdateReportedPropertiesAsync(new TwinCollection(), cancellationToken);
+            await this.moduleClient.GetTwinAsync(cancellationToken);
```
damonbarry: Why are you making these calls?

test/Microsoft.Azure.Devices.Edge.Test.Common/IotHub.cs
```diff
@@ -146,23 +146,31 @@ public Task<Device> CreateEdgeDeviceIdentityAsync(string deviceId, Authenticatio
             string deviceId,
             string moduleId,
             CloudToDeviceMethod method,
-            CancellationToken token)
+            CancellationToken token,
+            bool retry = true)
```
damonbarry: This is a workaround for the problem your module has connecting to edge hub, right?

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
lfitchett: There's a couple metrics like twin update duration and message latency that only show up if the relevant action has happened. This just does everything needed to populate all the metrics.

test/Microsoft.Azure.Devices.Edge.Test.Common/IotHub.cs
lfitchett: It isn't a solution, it just makes it fail fast and actually show the error. For the test, if there's any kind of connection error, I don't want it to retry until timeout, I want it to fail.

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
damonbarry: Maybe a comment to that effect...

PR #2463, opened 2020-02-04
Fixing c2d amqp receiver

dylanbronson:
Using the same object in certificate installation and creating device client in C2D connectivity tests. With the new change to AMQP certificate installation, we set a callback for certificate validation. 

We need to use that same TransportSettings object in DeviceClient.CreateFromConnectionString, so that it is validated.


varunpuranik:
The change looks good. But I am not sure what the issue is here..

dylanbronson:
We need to use the same TransportSettings object in CreateFromConnectionString as we do when we install the certs. This is because in the cert installation code, we set a callback on the transportSettings (for AMQP)

You can see that change in this PR: https://github.com/Azure/iotedge/commit/9fe8eeea7c92fc6f801ab1cc19ea917b2564c0e8

PR #2467, opened 2020-02-04
[k8s]handling module status based on pod condition object

ggjjj:


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
```diff
@@ -83,19 +87,23 @@ static ReportedModuleStatus ConvertPodStatusToModuleStatus(Option<V1PodStatus> p
...
+                                        {
                                             return new ReportedModuleStatus(ModuleStatus.Stopped, $"Module Stopped reason: {c.State.Terminated.Reason}");
+                                        }            
                                     }
                                     else
                                     {
                                         return new ReportedModuleStatus(ModuleStatus.Backoff, $"Started at {c.State.Running.StartedAt}");
                                     }
-                                }).GetOrElse(() => new ReportedModuleStatus(ModuleStatus.Failed, $"Module Failed with container status Unknown More Info: K8s reason: {status.Reason} with message: {status.Message}"));
+                                }).GetOrElse(() => new ReportedModuleStatus(ModuleStatus.Failed, $"Module Failed with container status Unknown More Info: K8s reason: {status.Conditions.Last().Reason} with message: {status.Conditions.Last().Message}"));
```
kkdawkins: Are you somehow guaranteeing order in your Enumerable here such that you can depend on .Last? Why not .First or if this Enumerable has only one item, Single? Also, IIRC, Last/Single/First will throw if the Enumerable is empty... is there a chance that it is empty? If so, I sugest you use the OrDefault versions of these functions.

kkdawkins: Comment applies to all uses of Last.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
```diff
@@ -61,15 +61,19 @@ static ReportedModuleStatus ConvertPodStatusToModuleStatus(Option<V1PodStatus> p
                                     else if (c.State.Terminated != null)
                                     {
                                         if (c.State.Terminated.ExitCode != 0)
+                                        {
```
kkdawkins: Why the formatting change?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesRuntimeInfoProviderTest.cs
```diff
@@ -305,25 +319,25 @@ public static IEnumerable<object[]> GetListOfPodsInAbnormalPhase()
             {
                 new object[]
                 {
-                    CreatePodWithPodParametersOnly("Running", "Unschedulable", "persistentvolumeclaim module-a-pvc not found"),
+                    CreatePodWithPodParametersOnly("Pending", null, null, "Unschedulable", "persistentvolumeclaim module-a-pvc not found"),
```
kkdawkins: Null or string.Empty? The latter is generally safer.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/PodExtensions.cs
ggjjj: Rider suggestion of inconsistent brackets

kkdawkins:
LGTM but I would get someone like @darobs who has Kubernetes knowledge to also look at it :)

PR #2469, opened 2020-02-05
Parse priority and TTL for routes and plumb down to messaging layer

richma-ms:
This change adds JSON deserialization for the new priority and TTL properties on routes, as per the following schema:
    "routes": {
      "route1": {
        "route": <string>,
        "priority": <uint>,
        "timeToLiveSecs": <uint>
        },
      "route2": <string>
    }

The meat of the change is in the following files:
- EdgeHubConfigParser.cs
- Evaluator.cs
- RouteConfiguration.cs
- RouteResult.cs

Everything else is plumbing and updating calling params due to having changed Route instantiation signature.

Also took the opportunity to clean up legacy code where a route can have multiple endpoints.  In our scenarios, each route must have a single endpoint only.


edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Evaluator.cs
```diff
@@ -53,9 +51,9 @@ public ISet<Route> Routes
             }
         }
 
-        public ISet<Endpoint> Evaluate(IMessage message)
+        public ISet<RouteResult> Evaluate(IMessage message)
         {
-            var endpoints = new HashSet<Endpoint>();
+            var results = new HashSet<RouteResult>();
```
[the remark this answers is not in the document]
dmolokanov: Actually this whole block can be done via single linq chain. I prefer the declarative approach to work with collections in c#
```csharp
// Because we are only reading here, it doesn't matter that it is under a lock
// ReSharper disable once InconsistentlySynchronizedField - compiledRoutes is immutable
ImmutableDictionary<string, CompiledRoute> snapshot = this.compiledRoutes;

// Multiple routes for the same endpoint can exist, in which case
// the message should be routed with the highest available priority
var routes = snapshot.Values
    .Where(cr => cr.Route.Source.Match(message.MessageSource) && EvaluateInternal(cr, message))
    .Select(cr => new RouteResult(cr.Route.Endpoint, cr.Route.Priority, cr.Route.TimeToLiveSecs))
    .GroupBy(r => r.Endpoint)
    .Select(dups => dups.OrderBy(r => r.Priority).First());

var results = new HashSet<RouteResult>(routes);
```

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Routing.Core
+{
+    using System;
+
+    public class RouteResult : IEquatable<RouteResult>
```
dmolokanov: I noticed that a lot of classes implement IEquatable interface, but I think it is a pretty strong constraint for classes. https://rules.sonarsource.com/csharp/RSPEC-4035 

@varunpuranik Could you please elaborate it little bit about reasons to have them?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Router.cs
```diff
@@ -130,14 +130,13 @@ public async Task RemoveRoute(string id)
...
-                ImmutableHashSet<Endpoint> currentEndpoints = this.Routes.SelectMany(r => r.Endpoints).ToImmutableHashSet();
-                this.routes.Value = snapshot.Remove(id);
-                ISet<string> removedEndpointIds = currentEndpoints.Except(this.Routes.SelectMany(r => r.Endpoints))
-                    .Select(e => e.Id)
-                    .ToImmutableHashSet();
-
-                this.evaluator.RemoveRoute(id);
-                await this.dispatcher.RemoveEndpoints(removedEndpointIds);
+
+                if (snapshot.TryGetValue(id, out Route removedRoute))
```
dmolokanov: we have `TryRemove` method to make 1 method call intead of 2.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
```diff
@@ -0,0 +1,57 @@
...
+                return false;
+            }
+
+            // Check endpoint, priority, and TTL
+            bool areEqual =
+                this.Endpoint.Equals(other.Endpoint) &&
+                this.Priority == other.Priority &&
+                this.TimeToLiveSecs == other.TimeToLiveSecs;
+
+            return ReferenceEquals(this, other) || areEqual;
```
dylanbronson: When would references be equal but not the inner values?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft. All rights reserved.
```
dylanbronson: Can you add unit tests for this class?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
```diff
@@ -6,18 +6,19 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Config
...
-            this.Routes = routes;
-            this.StoreAndForwardConfiguration = storeAndForwardConfiguration;
-        }
-
-        public string SchemaVersion { get; }
-
-        public IDictionary<string, string> Routes { get; }
-
+        [JsonConstructor]
+        public EdgeHubDesiredProperties(string schemaVersion, IDictionary<string, RouteConfiguration> routes, StoreAndForwardConfiguration storeAndForwardConfiguration)
```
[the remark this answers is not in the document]
philipktlin: you can call Preconditions methods to validate inputs.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            var routes = new Dictionary<string, RouteConfiguration>();
+            JToken token = JToken.Load(reader);
+
+            foreach (JToken child in token.Children())
+            {
+                if (child.First.Type == JTokenType.String)
+                {
+                    // Legacy routes are just string properties
+                    routes.Add(child.First.Path, new RouteConfiguration(child.First.ToObject<string>()));
```
philipktlin: make sure child.First.ToObject<string>() not return null or white spaces.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
```diff
@@ -0,0 +1,57 @@
...
+        public bool Equals(RouteResult other)
+        {
+            if (ReferenceEquals(null, other))
+            {
+                return false;
+            }
+
+            // Check endpoint, priority, and TTL
+            bool areEqual =
+                this.Endpoint.Equals(other.Endpoint) &&
```
philipktlin: please use string.Comparison (https://docs.microsoft.com/en-us/dotnet/api/system.stringcomparison?view=netframework-4.8); is it case sensitive or insensitive?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
philipktlin: Suggest to move ReferenceEquals check to its own if block and before line 27.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Evaluator.cs
```diff
@@ -64,30 +62,43 @@ public ISet<Endpoint> Evaluate(IMessage message)
             {
                 if (EvaluateInternal(compiledRoute, message))
                 {
-                    endpoints.UnionWith(compiledRoute.Route.Endpoints);
+                    RouteResult toAdd = new RouteResult(compiledRoute.Route.Endpoint, compiledRoute.Route.Priority, compiledRoute.Route.TimeToLiveSecs);
+
+                    // Multiple routes for the same endpoint can exist, in which case
+                    // the message should be routed with the highest available priority
+                    IEnumerable<RouteResult> dupeResults = results.Where(r => r.Endpoint.Equals(compiledRoute.Route.Endpoint));
```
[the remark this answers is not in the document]
varunpuranik: +1

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
varunpuranik: The reason is mainly to have a strongly typed way to check equality to another object of the same type. I think most of these classes can be made sealed..

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Routing.Core
+{
+    using System;
+
+    public class RouteResult : IEquatable<RouteResult>
+    {
+        public RouteResult(Endpoint endpoint, uint priority, uint timeToLiveSecs)
+        {
+            this.Endpoint = endpoint;
```
varunpuranik: Precondition checks pls

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
```diff
@@ -0,0 +1,57 @@
...
+    public class RouteResult : IEquatable<RouteResult>
+    {
+        public RouteResult(Endpoint endpoint, uint priority, uint timeToLiveSecs)
+        {
+            this.Endpoint = endpoint;
+            this.Priority = priority;
+            this.TimeToLiveSecs = timeToLiveSecs;
+        }
+
+        public readonly Endpoint Endpoint;
```
varunpuranik: Please make these properties

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
varunpuranik: yes, the ref check should be before property check

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
```diff
@@ -0,0 +1,57 @@
...
+            // Check endpoint, priority, and TTL
+            bool areEqual =
+                this.Endpoint.Equals(other.Endpoint) &&
+                this.Priority == other.Priority &&
+                this.TimeToLiveSecs == other.TimeToLiveSecs;
+
+            return ReferenceEquals(this, other) || areEqual;
+        }
+
+        public override bool Equals(object obj)
```
varunpuranik: You shoudl be able to call the other Equals

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                return true;
+            }
+
+            return obj.GetType() == this.GetType() && this.Equals((RouteResult)obj);
+        }
+
+        public override int GetHashCode()
+        {
+            // Not an accurate cast from uint to int, but for hashing purposes
```
varunpuranik: Something like this instead? https://github.com/Azure/iotedge/blob/3eafe054159320ef61547c8fbadf812b767d7da1/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Route.cs#L66

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -48,7 +48,7 @@ public class StoringAsyncEndpointExecutor : IEndpointExecutor
 
         public EndpointExecutorStatus Status => this.machine.Status;
 
-        public async Task Invoke(IMessage message)
+        public async Task Invoke(IMessage message, uint priority, uint timeToLiveSecs)
```
varunpuranik: To be implemented in next PR?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
varunpuranik: Do you have tests for this?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
```diff
@@ -6,18 +6,19 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Config
...
+        {
+            this.SchemaVersion = schemaVersion;
+            this.Routes = routes;
+            this.StoreAndForwardConfiguration = storeAndForwardConfiguration;
+        }
+
+        public string SchemaVersion { get; }
+
+        [JsonConverter(typeof(RouteConfigurationDictionaryConverter))]
+        public IDictionary<string, RouteConfiguration> Routes;
```
philipktlin: Just want to bring this existing issue: IDctionary doesn't guarantee the order; should we use OrderedDictionary https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary?redirectedfrom=MSDN&view=netframework-4.8.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Evaluator.cs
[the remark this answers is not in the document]
philipktlin: my 2 cents: use ToHashSet method if above impl is used.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
richma-ms: list is actually a better approach now that we have a customer converter.  I don't see any place where we're actually using this as a dictionary.

---
In reply to: [376650794](https://github.com/Azure/iotedge/pull/2469#discussion_r376650794) [](ancestors = 376650794)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
richma-ms: was going to add testing as part of the integration tests WI, right now I just verified that existing manifests aren't broken via E2E tests

---
In reply to: [376244854](https://github.com/Azure/iotedge/pull/2469#discussion_r376244854) [](ancestors = 376244854)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public class RouteConfigurationDictionaryConverter : JsonConverter
+    {
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer) => throw new NotSupportedException();
+
+        public override object ReadJson(JsonReader reader, Type type, object value, JsonSerializer serializer)
+        {
+            var routes = new Dictionary<string, RouteConfiguration>();
+            JToken token = JToken.Load(reader);
```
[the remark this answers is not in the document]
richma-ms: yes, the foreach just noops because token.Children() is empty

---
In reply to: [376142815](https://github.com/Azure/iotedge/pull/2469#discussion_r376142815) [](ancestors = 376142815)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
richma-ms: I think this should be checked in the constructor of the concrete object, not here.

---
In reply to: [376169406](https://github.com/Azure/iotedge/pull/2469#discussion_r376169406) [](ancestors = 376169406)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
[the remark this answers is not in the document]
richma-ms: Assuming you meant "private".

Making the ctor private might be future testing harder, as we'll likely need to construct custom instances of this object.  Although I think your concerns are addressed if I add checks for null params.

---
In reply to: [376095386](https://github.com/Azure/iotedge/pull/2469#discussion_r376095386) [](ancestors = 376095386)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
richma-ms: see other comments, I think it makes more sense to do validation check in the object-ctor.  Null checks shouldn't be necessary as we're expected to throw an exception in that case anyways.

---
In reply to: [376237140](https://github.com/Azure/iotedge/pull/2469#discussion_r376237140) [](ancestors = 376237140)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                if (child.First.Type == JTokenType.String)
+                {
+                    // Legacy routes are just string properties
+                    routes.Add(child.First.Path, new RouteConfiguration(child.First.ToObject<string>()));
+                }
+                else
+                {
+                    // Newer routes are objects with their own properties
+                    routes.Add(child.First.Path, child.First.ToObject<RouteConfiguration>());
```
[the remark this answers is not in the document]
richma-ms: The existing behavior is that we're expecting an exception from JSON deserialize if the payload is malformed, which then gets translated into an error that goes upstream.  So I don't think do an explicit null-check here adds anything.

---
In reply to: [376143105](https://github.com/Azure/iotedge/pull/2469#discussion_r376143105) [](ancestors = 376143105)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
richma-ms: I came to the realization that just verifying that existing behavior isn't broken already takes me half way there anyways, so I'll just add the other half and be done with integration tests.

---
In reply to: [377288342](https://github.com/Azure/iotedge/pull/2469#discussion_r377288342) [](ancestors = 377288342,376244854)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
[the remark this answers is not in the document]
dmolokanov: Will JSON converter handle exception from the constructor properly with precondition checks? All I wanted to suggest to keep public API the way that client code that uses this class cannot create an object in invalid state (with null routes collection for instance)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Evaluator.cs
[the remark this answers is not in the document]
dmolokanov: Cool. I didn't know we have this method!

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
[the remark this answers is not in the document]
richma-ms: I thought it'd throw a null deref exception (which is desired here), but turns out if child.First is never null, it just becomes empty.

Added logic to handle this case, and also check if the route name is empty.

---
In reply to: [376143105](https://github.com/Azure/iotedge/pull/2469#discussion_r376143105) [](ancestors = 376143105)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
richma-ms: I don't think such an UT would have anything meaningful. There's no real logic in this class, it's effectively a struct for all intents and purposes.  The IEquatable implementation was only added to facilitate updating the existing test collateral, and is not used anywhere else.

---
In reply to: [376164043](https://github.com/Azure/iotedge/pull/2469#discussion_r376164043) [](ancestors = 376164043)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
richma-ms: I only added it to facilitate updating existing test collateral, where Endpoint comparison now becomes RouteResult comparison.  It's not actually used anywhere in product code, but updating all our tests without IEquatable here would be too expensive.

---
In reply to: [376152050](https://github.com/Azure/iotedge/pull/2469#discussion_r376152050) [](ancestors = 376152050)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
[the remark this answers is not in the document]
richma-ms: JSON converter would just bubble the exception up, which is what we're expecting.  The existing behavior is already such that if deserialization throws an exception, we catch up and translate it into a deployment error upstream.  Adding preconditions here would just reinforce the same pattern.

---
In reply to: [377366379](https://github.com/Azure/iotedge/pull/2469#discussion_r377366379) [](ancestors = 377366379)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
richma-ms: I don't think this class should make any assumptions about how the Endpoint.Equals is implemented, even if it's using a string compare underneath right now.

---
In reply to: [376178099](https://github.com/Azure/iotedge/pull/2469#discussion_r376178099) [](ancestors = 376178099)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: yes, this PR just plumbs everything down to this layer.

---
In reply to: [376244577](https://github.com/Azure/iotedge/pull/2469#discussion_r376244577) [](ancestors = 376244577)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Evaluator.cs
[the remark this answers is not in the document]
richma-ms: neat! =)

---
In reply to: [376150004](https://github.com/Azure/iotedge/pull/2469#discussion_r376150004) [](ancestors = 376150004)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Router.cs
richma-ms: I need a ref to removedRoute to get the endpoint id, which is different then the route id being passed in.

---
In reply to: [376152598](https://github.com/Azure/iotedge/pull/2469#discussion_r376152598) [](ancestors = 376152598)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core.Config
+{
+    using System.ComponentModel;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Azure.Devices.Routing.Core;
+    using Newtonsoft.Json;
+
+    public class RouteConfiguration
```
philipktlin: Would it be possible to have only 1 implementation, either RouteConfig and RouteConfiguration? If really not possible, please think about to rename RouteConfiguration class in order to easy distinguish from existing RouteConfig class.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
philipktlin: Sure.  Agreed this existing issue, please create a bug for tracking.  thanks.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfigurationConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public override IDictionary<string, RouteConfiguration> ReadJson(JsonReader reader, Type type, IDictionary<string, RouteConfiguration> value, bool hasValue, JsonSerializer serializer)
+        {
+            var routes = new Dictionary<string, RouteConfiguration>();
+            JToken token = JToken.Load(reader);
+
+            foreach (JToken child in token.Children())
+            {
+                // The route name must not be empty
+                Preconditions.CheckNonWhiteSpace(child.First.Path, nameof(child.First.Path));
```
philipktlin: Suggest to provide more detailed message, e.g. print out which children node.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Evaluator.cs
```diff
@@ -53,41 +51,36 @@ public ISet<Route> Routes
...
+
+            // Multiple routes for the same endpoint can exist, in which case
+            // the message should be routed with the highest available priority
+            var routes = snapshot.Values
+                .Where(cr => cr.Route.Source.Match(message.MessageSource) && EvaluateInternal(cr, message))
+                .Select(cr => new RouteResult(cr.Route.Endpoint, cr.Route.Priority, cr.Route.TimeToLiveSecs))
+                .GroupBy(r => r.Endpoint)
+                .Select(dupes => dupes.OrderBy(r => r.Priority).First());
+
+            var results = new HashSet<RouteResult>(routes);
```
philipktlin: nit: you can combine line 62 and 68 by calling ToHashSet().

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Route.cs
```diff
@@ -11,13 +11,15 @@ namespace Microsoft.Azure.Devices.Routing.Core
 
     public class Route : IEquatable<Route>
     {
-        public Route(string id, string condition, string iotHubName, IMessageSource source, ISet<Endpoint> endpoints)
```
philipktlin: Just curious, I found the usages of this constructor in RouteFactory; and it always pass only 1 element in endpoints argument.  Do you know why this argument type is ISet<Endpoint>? what use cases?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Route.cs
```diff
@@ -45,7 +51,7 @@ public bool Equals(Route other)
             return string.Equals(this.Id, other.Id) &&
                    string.Equals(this.Condition, other.Condition) &&
                    this.Source.Equals(other.Source) &&
-                   this.Endpoints.SetEquals(other.Endpoints);
+                   this.Endpoint.Equals(other.Endpoint);
```
philipktlin: 1. Do you know why IoTHubName is not considered in Equals method?
2. Also IMessageSource doesn't implement IEquatable or similar interface, are we doing reference comparison here?
3. For string comparison, should we pass string comparison option?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Route.cs
```diff
@@ -71,12 +77,12 @@ public override int GetHashCode()
...
                 hash = hash * 31 + this.Source.GetHashCode();
-                hash = this.Endpoints.Aggregate(hash, (acc, b) => acc * 31 + b.GetHashCode());
+                hash = hash * 31 + this.Endpoint.GetHashCode();
                 return hash;
             }
         }
 
         public override string ToString() =>
-            string.Format(CultureInfo.InvariantCulture, "Route(\"{0}\", {1}, \"{2}\" => ({3})", this.Id, this.Source, this.Condition, string.Join(", ", this.Endpoints));
+            string.Format(CultureInfo.InvariantCulture, "Route(\"{0}\", {1}, \"{2}\" => ({3})", this.Id, this.Source, this.Condition, this.Endpoint);
```
philipktlin: btw only ConsoleEndpoint and NullEndpoint implemented ToString().  Please create a bug for tracking if we want to log Endpoint details.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteFactory.cs
```diff
@@ -12,6 +12,12 @@ namespace Microsoft.Azure.Devices.Routing.Core
 
     public abstract class RouteFactory
     {
+        // Default route priority at half of uint.MaxValue,
+        // so that routes with no custom priority fall below
+        // everything else, and reserve room for routes that
+        // are below default priority.
+        public const uint DefaultPriority = uint.MaxValue / 2;
```
philipktlin: UInt.MaxValue is 4294967295 and therefore default priority is 2147483647.  Should we round it and make it easier to remember, e.g. 2M?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
philipktlin: I think you need to implement unit test because:
1. even very simple logic can be incorrect; e.g. copy-and-paste and assign to wrong variable; string comparison whether is case sensitive or insensitive.
2. since code coverage is turned on, we should try to have code coverage for all code added/changed in order to increase our code coverage.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public override bool Equals(object obj) => obj.GetType() == this.GetType() && this.Equals((RouteResult)obj);
+
+        public override int GetHashCode()
+        {
+            // Not accurate to cast from uint to int, but for hashing purposes
+            // that's okay, and much better than getting an int overflow
+            unchecked
+            {
+                return this.Endpoint.Id.GetHashCode() + (int)this.Priority + (int)this.TimeToLiveSecs;
```
philipktlin: Should we find a better way to increase the uniqueness of hash code value?  Normally a prime number will involve in this calculation.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/AsyncEndpointExecutor.cs
```diff
@@ -78,7 +78,7 @@ public AsyncEndpointExecutor(Endpoint endpoint, ICheckpointer checkpointer, Endp
 
         public EndpointExecutorStatus Status => this.machine.Status;
 
-        public virtual Task Invoke(IMessage message) => this.SendToTplHead(message);
+        public virtual Task Invoke(IMessage message, uint _, uint __) => this.SendToTplHead(message);
```
philipktlin: why not provide name to these 2 arguments? will these be used in next PRs?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/SyncEndpointExecutor.cs
```diff
@@ -45,7 +45,7 @@ public SyncEndpointExecutor(Endpoint endpoint, ICheckpointer checkpointer, Endpo
 
         public EndpointExecutorStatus Status => this.machine.Status;
 
-        public async Task Invoke(IMessage message)
+        public async Task Invoke(IMessage message, uint _, uint __)
```
philipktlin: please name these arguments

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteFactory.cs
```diff
@@ -25,11 +31,14 @@ protected RouteFactory(IEndpointFactory endpointFactory)
 
         public abstract string GetNextRouteId();
 
-        public Route Create(string routeString)
+        public Route Create(string routeString) =>
+            this.Create(routeString, DefaultPriority, 0);
```
philipktlin: why default TTL is zero?  Does this mention in the spec?

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/config/EdgeHubDesiredPropertiesTest.cs
```diff
@@ -0,0 +1,198 @@
...
+            Assert.Equal(RouteFactory.DefaultPriority, desiredProperties.Routes["route1"].Priority);
+            Assert.Equal(0u, desiredProperties.Routes["route1"].TimeToLiveSecs);
+
+            Assert.Equal("from /* INTO $upstream", desiredProperties.Routes["route2"].Route);
+            Assert.Equal(1u, desiredProperties.Routes["route2"].Priority);
+            Assert.Equal(7200u, desiredProperties.Routes["route2"].TimeToLiveSecs);
+
+            Assert.Equal(20, desiredProperties.StoreAndForwardConfiguration.TimeToLiveSecs);
+
+            string emptyRoutesSection =
```
philipktlin: Please split into different test methods.

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouteTest.cs
```diff
@@ -19,29 +20,29 @@ public class RouteTest : RoutingUnitTestBase
...
-        static readonly Route Route4 = new Route("id2", "rule2", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint> { Endpoint1 });
-        static readonly Route Route5 = new Route("id3", "rule3", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint>());
-        static readonly Route Route6 = new Route("id3", "rule3", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint>());
-        static readonly Route Route7 = new Route("id2", "rule1", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint> { Endpoint1 });
+        static readonly Route Route1 = new Route("id1", "rule1", "hub", TelemetryMessageSource.Instance, Endpoint1, 0, 0);
+        static readonly Route Route2 = new Route("id1", "rule1", "hub", TelemetryMessageSource.Instance, Endpoint2, 0, 0);
+        static readonly Route Route3 = new Route("id1", "rule1", "hub", TelemetryMessageSource.Instance, Endpoint3, 0, 0);
+        static readonly Route Route4 = new Route("id2", "rule2", "hub", TelemetryMessageSource.Instance, Endpoint1, 0, 0);
+        static readonly Route Route5 = new Route("id3", "rule3", "hub", TelemetryMessageSource.Instance, new Mock<Endpoint>("id1").Object, 0, 0);
```
philipktlin: why test data is changed? previously it is empty endpoint, and now a mock endpoint is provided.

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouteTest.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.Devices.Routing.Core.Test
     using Microsoft.Azure.Devices.Routing.Core.MessageSources;
     using Microsoft.Azure.Devices.Routing.Core.Query;
     using Microsoft.Azure.Devices.Routing.Core.Test.Endpoints;
+    using Moq;
     using Xunit;
 
     [ExcludeFromCodeCoverage]
```
philipktlin: Since this file is changed, I think [ExcludeFromCodeCoverage] should be removed.

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouteTest.cs
```diff
@@ -19,29 +20,29 @@ public class RouteTest : RoutingUnitTestBase
...
 
         [Fact]
         [Unit]
         public void TestConstructor()
         {
-            Assert.Throws<ArgumentNullException>(() => new Route(null, "condition", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint>()));
-            Assert.Throws<ArgumentNullException>(() => new Route("id", null, "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint>()));
-            Assert.Throws<ArgumentNullException>(() => new Route("id", "condition", null, TelemetryMessageSource.Instance, new HashSet<Endpoint>()));
-            Assert.Throws<ArgumentNullException>(() => new Route("id", "condition", "hub", TelemetryMessageSource.Instance, null));
+            Assert.Throws<ArgumentNullException>(() => new Route(null, "condition", "hub", TelemetryMessageSource.Instance, new Mock<Endpoint>("id1").Object, 0, 0));
```
philipktlin: nit: try to use [Theory] and [InlineData]

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouteTest.cs
```diff
@@ -19,29 +20,29 @@ public class RouteTest : RoutingUnitTestBase
...
+            Assert.Throws<ArgumentNullException>(() => new Route("id", "condition", null, TelemetryMessageSource.Instance, new Mock<Endpoint>("id1").Object, 0, 0));
+            Assert.Throws<ArgumentNullException>(() => new Route("id", "condition", "hub", TelemetryMessageSource.Instance, null, 0, 0));
         }
 
         [Fact]
         [Unit]
         public void SmokeTest()
         {
-            var route = new Route("id", "true", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint> { new TestEndpoint("id1") });
+            var route = new Route("id", "true", "hub", TelemetryMessageSource.Instance, new Mock<Endpoint>("id1").Object, 0, 0);
```
philipktlin: nit:It seems all test cases in this class, use zero for both TTL and priority.  Please just use different values.

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouterTest.cs
```diff
@@ -152,16 +148,15 @@ public async Task TestRemoveRoute()
...
 
-            var allEndpoints = new HashSet<Endpoint> { endpoint1, endpoint2, endpoint3, endpoint4 };
+            var allEndpoints = new HashSet<Endpoint> { endpoint1, endpoint2, endpoint3 };
 
-            var route1 = new Route("id1", "key1 = \"value1\"", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint> { endpoint1, endpoint2 });
-            var route2 = new Route("id2", "key1 = \"value2\"", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint> { endpoint3, endpoint4 });
-            var route3 = new Route("id3", "key1 = \"value1\"", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint> { endpoint3, endpoint4 });
+            var route1 = new Route("id1", "key1 = \"value1\"", "hub", TelemetryMessageSource.Instance, endpoint1, 0, 0);
+            var route2 = new Route("id2", "key1 = \"value2\"", "hub", TelemetryMessageSource.Instance, endpoint2, 0, 0);
+            var route3 = new Route("id3", "key1 = \"value1\"", "hub", TelemetryMessageSource.Instance, endpoint3, 0, 0);
```
philipktlin: should have route with same endpoint

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/config/EdgeHubDesiredPropertiesTest.cs
```diff
@@ -0,0 +1,198 @@
...
+        {
+            string noRoutesSection =
+                @"{
+                  'schemaVersion': '1.0',
+                  'storeAndForwardConfiguration': {
+                    'timeToLiveSecs': 20
+                  },
+                  '$version': 2
+                }";
+            Assert.Throws<ArgumentNullException>(() => JsonConvert.DeserializeObject<EdgeHubDesiredProperties>(noRoutesSection));
```
philipktlin: either split into different tests or use [Theory] and [MemberData] attributes

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/EvaluatorTest.cs
```diff
@@ -43,35 +43,38 @@ public void TestSetRoute()
...
+            Assert.Equal(2, results.Count);
+            Assert.Contains(result1, results);
+            Assert.Contains(result2, results);
             Assert.True(evaluator.Routes.SetEquals(new HashSet<Route> { route3, route2, route1 }));
 
             // Replace route2
             var endpoint3 = new TestEndpoint("id3");
-            var route4 = new Route("id2", "true", "hub", TelemetryMessageSource.Instance, new HashSet<Endpoint> { endpoint3 });
+            var result3 = new RouteResult(endpoint3, 0, 0);
+            var route4 = new Route("id2", "true", "hub", TelemetryMessageSource.Instance, endpoint3, 0, 0);
```
philipktlin: For replace, please use different priority and TTL.

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/query/ConditionVisitorTest.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.Azure.Devices.Routing.Core.Test.Query
     using Microsoft.Azure.Devices.Edge.Util.Test.Common;
     using Microsoft.Azure.Devices.Routing.Core.MessageSources;
     using Microsoft.Azure.Devices.Routing.Core.Query;
+    using Moq;
     using Xunit;
 
     [ExcludeFromCodeCoverage]
```
philipktlin: Please remove [ExcludeFromCodeCoverage]

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/endpoints/StoringAsyncEndpointExecutorTest.cs
```diff
@@ -36,7 +36,7 @@ public async Task InvokeTest()
             // Act - Send messages to invoke
             foreach (IMessage message in messages)
             {
-                await storingAsyncEndpointExecutor.Invoke(message);
+                await storingAsyncEndpointExecutor.Invoke(message, 0, 3600);
```
philipktlin: In general, should we use random number for TTL and priority for existing tests.  Rather than hard-coded a fixed value.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
richma-ms: yea, I noticed this as well.  I plan on renaming this to RouteSpec after I get a sign-off.  I figured if I rename this during the PR, then the commenting will get really really weird.

---
In reply to: [377442628](https://github.com/Azure/iotedge/pull/2469#discussion_r377442628) [](ancestors = 377442628)

richma-ms: or rename in a separate PR to reduce confusion.

---
In reply to: [377814817](https://github.com/Azure/iotedge/pull/2469#discussion_r377814817) [](ancestors = 377814817,377442628)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Evaluator.cs
richma-ms: not sure if you saw my other comment, but there is no ToHashSet() that I could find, only ToImmutableHashSet(), which doesn't work here because we need to modify the collection right below.

---
In reply to: [377443584](https://github.com/Azure/iotedge/pull/2469#discussion_r377443584) [](ancestors = 377443584)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Route.cs
richma-ms: I talked to Varun about this, it's legacy code from when we adapted the routing stack from IoT Hub.  We don't actually have a usecase that takes more than 1 endpoint, that's why I cleaned it up.  It would have actually taken more time if I needed to update every caller with the older signature.

---
In reply to: [377445441](https://github.com/Azure/iotedge/pull/2469#discussion_r377445441) [](ancestors = 377445441)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteFactory.cs
richma-ms: hmm...lemme think about it.

---
In reply to: [377447237](https://github.com/Azure/iotedge/pull/2469#discussion_r377447237) [](ancestors = 377447237)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteFactory.cs
richma-ms: 0 is treated as a special value for TTL, there's no valid scenario where any message should have 0sec TTL for real.  We use zero to signify that the operator did not specify a TTL on the route, at which point the global TTL value should take over.

I'll add a comment to explain this.

---
In reply to: [377458687](https://github.com/Azure/iotedge/pull/2469#discussion_r377458687) [](ancestors = 377458687)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteResult.cs
richma-ms: I actually think this is desirable, for a given endpoint, if the priority and TTL are the same, then the two results are effectively the same thing.  The hash from the underlying endpoint provides enough uniqueness for all other comparisons.

---
In reply to: [377450086](https://github.com/Azure/iotedge/pull/2469#discussion_r377450086) [](ancestors = 377450086)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/AsyncEndpointExecutor.cs
richma-ms: these won't be used ever, this is in the AsyncEndpointExecutor implementation, not the StoringAsyncEndpointExecutor.

---
In reply to: [377458214](https://github.com/Azure/iotedge/pull/2469#discussion_r377458214) [](ancestors = 377458214)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/SyncEndpointExecutor.cs
richma-ms: these won't be used ever

---
In reply to: [377458446](https://github.com/Azure/iotedge/pull/2469#discussion_r377458446) [](ancestors = 377458446)

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouteTest.cs
richma-ms: the class paradigm changed, previous it took a collection, which could be empty.  That's a bug and not valid for Edge scenarios.  For our routes, there must be one endpoint and one endpoint only, now the code is updated to reflect that.

---
In reply to: [377461524](https://github.com/Azure/iotedge/pull/2469#discussion_r377461524) [](ancestors = 377461524)

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouteTest.cs
richma-ms: out of scope here, trying to keep changes to existing test collateral to a minimum (or I won't ever finish).

---
In reply to: [377462258](https://github.com/Azure/iotedge/pull/2469#discussion_r377462258) [](ancestors = 377462258)

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouteTest.cs
richma-ms: that's because the tests don't verify TTL and priority, so they're just fill in values much like mock endpoints.  Tests that actually cover TTL and priority will have more interesting values, and that'll come in a separate WI.

---
In reply to: [377462568](https://github.com/Azure/iotedge/pull/2469#discussion_r377462568) [](ancestors = 377462568)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Route.cs
richma-ms: I'm not clear on the history here, but if we want to make improvements to the Endpoint implementation we can do it in a separate PR.

---
In reply to: [377446071](https://github.com/Azure/iotedge/pull/2469#discussion_r377446071) [](ancestors = 377446071)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Route.cs
richma-ms: I think it's fine for now, it's been this way forever and I don't think we had any trouble diagnosing logs around endpoints.  If we come back and update the Endpoint class, then we can prob opportunistically improve it then.

---
In reply to: [377446707](https://github.com/Azure/iotedge/pull/2469#discussion_r377446707) [](ancestors = 377446707)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
richma-ms: 6257467

---
In reply to: [377442952](https://github.com/Azure/iotedge/pull/2469#discussion_r377442952) [](ancestors = 377442952)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfigurationConverter.cs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core.Config
+{
+    using System;
+    using System.Collections.Generic;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+    using Org.BouncyCastle.Security;
```
richma-ms: >using Org.BouncyCastle.Security; [](start = 4, length = 32)

for some reason VS keeps on randomly adding these to files I'm editing.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfigurationConverter.cs
richma-ms: CheckNonWhiteSpace doesn't take a custom string message.

---
In reply to: [377443190](https://github.com/Azure/iotedge/pull/2469#discussion_r377443190) [](ancestors = 377443190)

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouteTest.cs
richma-ms: doesn't this mean the testcode shouldn't be included in CodeCoverage?  If so, seems like that's still what we want.

---
In reply to: [377461886](https://github.com/Azure/iotedge/pull/2469#discussion_r377461886) [](ancestors = 377461886)

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/query/ConditionVisitorTest.cs
richma-ms: See other reply.  Seems like this still applies, we don't want test code to affect code coverage numbers?

---
In reply to: [377464887](https://github.com/Azure/iotedge/pull/2469#discussion_r377464887) [](ancestors = 377464887)

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/RouteTest.cs
richma-ms: talked and resolved offline

---
In reply to: [377894483](https://github.com/Azure/iotedge/pull/2469#discussion_r377894483) [](ancestors = 377894483,377461886)

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/query/ConditionVisitorTest.cs
richma-ms: talked and resolved offline

---
In reply to: [377895001](https://github.com/Azure/iotedge/pull/2469#discussion_r377895001) [](ancestors = 377895001,377464887)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/RouteFactory.cs
```diff
@@ -12,6 +12,13 @@ namespace Microsoft.Azure.Devices.Routing.Core
 
     public abstract class RouteFactory
     {
+        // Default route priority at half of uint.MaxValue,
+        // so that routes with no custom priority fall below
+        // everything else, and reserve room for routes that
+        // are below default priority.  We round down to an
+        // even 2bn for easy remembering.
+        public const uint DefaultPriority = 2000000000;
```
philipktlin: nit: you can use Digit Separator, e.g. 2_000_000_000.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/AsyncEndpointExecutor.cs
philipktlin: Would you tell me in what scenario, these 2 field will be used? If no use case, why do we need these 2 arguments?

PR #2477, opened 2020-02-05
Update IotEdgeSecurityDaemon.ps1

SlavyMihov:
Add descriptive message to every generic exception thrown.
Reason for this is that *Write-Host* and *Write-HostRed* printouts do not get into the DevOps logs (which makes the logs uninformative).


scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -964,7 +975,7 @@ function Install-Packages(
                 Write-HostRed ('IoT Edge is already installed. To update, run "Update-IoTEdge". ' +
                     'Alternatively, if you want to finalize the installation, run "Initialize-IoTEdge".')
             }
-            throw
+            throw 'IoT Edge is already installed.'
```
philipktlin: why extract the message on line 975 and use here?

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -309,46 +309,52 @@ function Initialize-IoTEdge {
     Set-StrictMode -Version 5
 
     if (-not (Test-EdgeAlreadyInstalled)) {
+        $error = ('IoT Edge is not yet installed. ' + $InstallMessage)
         Write-HostRed
-        Write-HostRed ('IoT Edge is not yet installed. ' + $InstallMessage)
-        throw
+        Write-HostRed $error
+        throw $error
```
philipktlin: but why do we need to throw again with same error message? error message should already print out by Write-HostRed.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
SlavyMihov: This message intends to summarize the if-else branches into one concise message:
 * "IoT Edge or the IoT Edge Moby Engine is installed in an invalid location."
 * "IoT Edge is already installed."

I am open to suggestions or leave the throw command with no message.

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
SlavyMihov: *Write-Host* and *Write-HostRed* do not print anything in the DevOps logs, not sure why. That's the reason to have the same error message thrown as an exception (that makes it into the DevOps logs).

philipktlin: is it because iotedgequickstart swallow all logs from IotEdgeSecurityDaemon.ps1; we should try to fix it if that's the case.

SlavyMihov: Good point.

PR #2492, opened 2020-02-07
EdgeHub Restart Tester (Test Verification)

yophilav:
- Enable Verification in the Test Result Coordinator for EdgeHub Restart Tester
- Calculate statistical analysis in the TRC along with report


test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator.Reports.EdgeHubRestartTest
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Net;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    /// <summary>
+    /// This is a counting report to show test result counts, e.g. expect and match counts; and contains a list of unmatched test results.
```
philipktlin: Please update summary text

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    class EdgeHubRestartDirectMethodReport : TestResultReportBase
+    {
+        public EdgeHubRestartDirectMethodReport(
+            string trackingId,
+            string resultType,
+            bool isPassing,
+            long passedMessageCount,
+            Dictionary<string, ulong> messageCount,
```
philipktlin: please explain what is stored in this "messageCount" dictionary.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,66 @@
...
+            Dictionary<HttpStatusCode, List<TimeSpan>> completedStatusHistogram,
```
philipktlin: if i have 3 results - 2 results has status code "x" and both completed test in 5s and 1 result also has status code "x" and completed in 3s.  How completedStatusHistogram to support?  Again, what question do we want to answer by using histogram?  I don't want the report to grow too big.  If you want to share this information, please just write to TRC log.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,66 @@
...
+            TimeSpan minPeriod,
+            TimeSpan maxPeriod,
+            TimeSpan medianPeriod,
+            TimeSpan meanPeriod,
+            TimeSpan variancePeriod)
+            : base(trackingId, resultType)
+        {
+            this.isPassing = isPassing;
+            this.PassedMessageCount = passedMessageCount;
```
philipktlin: why only have passedMessageCount? but no failed message count?  Please define TotalCallsInvoked and TotalCallsReceived; and remove isPassing; which isPassed should be determined by TotalCallsInvoked and TotalCallsReceived.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,66 @@
...
+            this.MinPeriod = minPeriod;
+            this.MaxPeriod = maxPeriod;
+            this.MedianPeriod = medianPeriod;
+            this.MeanPeriod = meanPeriod;
+            this.VariancePeriod = variancePeriod;
+        }
+
+        bool isPassing;
+
+        internal long PassedMessageCount { get; }
```
philipktlin: do we use long type in sending code?

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        internal Dictionary<string, ulong> MessageCount { get; }
```
philipktlin: is this a problem when json serialization? use public.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        internal TimeSpan MedianPeriod { get; }
+
+        internal TimeSpan MeanPeriod { get; }
+
+        internal TimeSpan VariancePeriod { get; }
+
+        public override bool IsPassed => this.isPassing;
+
+        public override string Title => $"{this.ResultType} Report between {this.SourceList}";
```
philipktlin: this.SourceList here will just print the type reference instead of values inside the list.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        internal long PassedMessageCount { get; }
+
+        internal Dictionary<string, ulong> MessageCount { get; }
+
+        internal Dictionary<HttpStatusCode, ulong> RestartStatusHistogram { get; }
+
+        internal Dictionary<HttpStatusCode, List<TimeSpan>> CompletedStatusHistogram { get; }
+
+        internal List<string> SourceList { get; }
```
philipktlin: Split into 2 properties

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <summary>
+    /// This is a counting report to show test result counts, e.g. expect and match counts; and contains a list of unmatched test results.
+    /// </summary>
+    class EdgeHubRestartDirectMethodReport : TestResultReportBase
+    {
+        public EdgeHubRestartDirectMethodReport(
+            string trackingId,
+            string resultType,
+            bool isPassing,
+            long passedMessageCount,
```
philipktlin:
```suggestion
            ulong passedMessageCount,
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,304 @@
...
+                meanPeriod = totalSpan / orderedCompletedPeriods.Count();
+
+                // Compute Variance: var = sum((x - mean)^2) / N
+                //                       = sum(x^2) / N - mean^2
+                double variancePeriodInMilisec = (totalSpanSquareInMilisec / orderedCompletedPeriods.Count()) - Math.Pow(meanPeriod.TotalMilliseconds, 2);
+                variancePeriod = TimeSpan.FromMilliseconds(variancePeriodInMilisec);
+            }
+
+            // Make sure the maximum restart period is within a passable threshold
+            isPassing &= maxPeriod < this.Metadata.PassableEdgeHubRestartPeriod;
```
philipktlin: why do we need this?  this is doing similar thing as what runExpirationTime during each edgeHub restart test run.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,304 @@
...
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    sealed class EdgeHubRestartDirectMethodReportGenerator : ITestResultReportGenerator
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger(nameof(EdgeHubRestartDirectMethodReportGenerator));
+
+        internal EdgeHubRestartDirectMethodReportGenerator(
+            string trackingId,
+            EdgeHubRestartDirectMethodReportMetadata metadata,
```
philipktlin: Please pass SenderSource, and ReceiverSource instead of EdgeHubRestartDirectMethodReportMetadata.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,304 @@
...
+        public async Task<ITestResultReport> CreateReportAsync()
+        {
+            Logger.LogInformation($"Generating report: {nameof(EdgeHubRestartDirectMethodReport)} for [{this.Metadata.SenderSource}] and [{this.Metadata.ReceiverSource}]");
+
+            bool isPassing = true;
+            long previousSeqNum = 0;
+            long passedDirectMethodCount = 0;
+
+            // Value: (source, numOfResults)
+            Dictionary<string, ulong> resultCount = new Dictionary<string, ulong>()
```
philipktlin: don't use dictionary, just use 2 ulong count

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,304 @@
...
+                    // Increment sender result to have the same seq as the receiver
+                    await this.IncrementSequenceNumberAsync(
+                        this.SenderTestResults,
+                        nameof(this.SenderTestResults),
+                        receiverSeqNum,
+                        resultCount,
+                        completedStatusHistogram);
+
+                    // Fail the test
+                    isPassing = false;
```
philipktlin: store failed count, remove isPassing

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,304 @@
...
+
+                // Both sender & receiver have their dm results
+                resultCount[nameof(this.SenderTestResults)]++;
+                resultCount[nameof(this.ReceiverTestResults)]++;
+
+                // Adjust seqeunce number from both source to be equal before doing any comparison
+                EdgeHubRestartDirectMethodResult senderResult = JsonConvert.DeserializeObject<EdgeHubRestartDirectMethodResult>(this.SenderTestResults.Current.Result);
+                DirectMethodTestResult receiverResult = JsonConvert.DeserializeObject<DirectMethodTestResult>(this.ReceiverTestResults.Current.Result);
+
+                long senderSeqNum = this.ConvertStringToLong(senderResult.SequenceNumber);
```
philipktlin: Please have a PR to fix SequenceNubmer in DirectMethodTestResult to use uint instead of string.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
philipktlin:
```suggestion
        internal ulong PassedMessageCount { get; }
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,304 @@
...
+                        nameof(this.SenderTestResults),
+                        receiverSeqNum,
+                        resultCount,
+                        completedStatusHistogram);
+
+                    // Fail the test
+                    isPassing = false;
+                }
+
+                if (receiverSeqNum < senderSeqNum)
```
philipktlin: can you tell me any possibilities for this condition become true?

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,304 @@
...
+                        nameof(this.ReceiverTestResults),
+                        senderSeqNum,
+                        resultCount,
+                        completedStatusHistogram);
+
+                    // Fail the test
+                    isPassing = false;
+                }
+
+                // Note: In the current verification, if the receiver obtained more result dm than
```
philipktlin: I don't think you will have duplicate sequence number in direct method scenario, right?

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReport.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    class EdgeHubRestartMessageReport : TestResultReportBase
+    {
+        public EdgeHubRestartMessageReport(
+            string trackingId,
+            string resultType,
+            bool isPassing,
+            long passedMessageCount,
+            Dictionary<string, ulong> messageCount,
+            Dictionary<HttpStatusCode, List<TimeSpan>> completedStatusHistogram,
```
philipktlin: Remove all statistics in both report; it should be covered by metric.

test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -85,6 +86,12 @@ internal static List<ITestReportMetadata> ParseReportMetadataJson(string reportM
                         case TestReportType.CountingReport:
                             reportMetadataList.Add(JsonConvert.DeserializeObject<CountingReportMetadata>(((JProperty)metadata).Value.ToString()));
                             break;
+                        case TestReportType.EdgeHubRestartDirectMethodReport:
```
philipktlin: Add tests in TestReportUtilTest.cs

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
yophilav: So the `messageCount` is the received result from each of the source, namely, the sender and a reciever. this number should be always be equal if the test runs properly.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
yophilav: The `completedStatusHistrogram` is used to record of all the sender's result categorized by the status code as a key. In a passing case, we want to only look at the bucket of HTTP.OK and find the max of the timespan within that bucket. If the timespan is less than a minute the test passed.

The idea behind the histogram is that you can have a statistic result of how long the EdgeHub takes to restart.  For an example, imagine a case scenario, case 1: mean=50s, variance=0.1; I won't not have a confidence that the edgeHub "restarts" under 1 minutes since the variance is so low with mean pushing the boundary of our limit ( ideally this case should fail the test ).

With that said, I can have the report not print out all the timespan that it records, but instead a statistic of each bucket, namely {min, max, mean, mid, variance}

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
yophilav: The `TotalCallsInvoked` and `TotocallsReceived` are tracked by the dictionary `messageCount`. The passedMessageCount is another metric that counts how many of the messages are "good" independent of the previous messages.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
yophilav: Good catch.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
yophilav: In normal scenario, it shouldn't happen. I can think if somehow edgeHub retries a directMethod (which shouldn't happen).

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
yophilav: From the sender logic, this should never happen. But if it does, let's make sure we fail it.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReport.cs
yophilav: How does one introduce metric here? Metric does not know about the condition of the test. It only cares with at the scope of the module, not the test conditions.  Plus, this value will be available on LogAnalytics, right? you can just treat them like stat in the report.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
yophilav: I see what you mean. 
I'll keep this param for Direct Method, but remove it for Message. 
For Message, I'll just check the complete HTTP status if it's passing.

test/modules/Modules.Test/TestResultCoordinator/TestReportUtilTest.cs
```diff
@@ -101,12 +115,14 @@ public async Task TestGenerateTestResultReportsAsync_NullReportMetadata()
...
-                    directMethodReportMetadataWithoutReceiverSource
+                    directMethodReportMetadataWithoutReceiverSource,
+                    edgeHubRestartMessageReportMetadata,
+                    edgeHubRestartDirectMethodReportMetadata
                 },
                 mockTestReportGeneratorFactory.Object,
                 mockLogger.Object);
 
-            Assert.Equal(expectedReportCount, reports.Length);
+            Assert.Equal(expectedExceptionCount, reports.Length);
```
philipktlin: should not change this, right?

test/modules/Modules.Test/TestResultCoordinator/TestReportUtilTest.cs
```diff
@@ -312,6 +341,56 @@ public void ParseReportMetadataList_ParseErrorReportMetadata()
...
+            List<ITestReportMetadata> results = TestReportUtil.ParseReportMetadataJson(testDataJson, new Mock<ILogger>().Object);
+
+            Assert.Single(results);
+            var reportMetadata = results[0] as EdgeHubRestartDirectMethodReportMetadata;
+            Assert.NotNull(reportMetadata);
+            Assert.Equal(TestOperationResultType.EdgeHubRestartDirectMethod, reportMetadata.TestOperationResultType);
+            Assert.Equal(TestReportType.EdgeHubRestartDirectMethodReport, reportMetadata.TestReportType);
+            Assert.Equal("edgeHubRestartTester1.EdgeHubRestartDirectMethod", reportMetadata.SenderSource);
+            Assert.Equal("directMethodReceiver1.receive", reportMetadata.ReceiverSource);
+            Assert.Equal("00:01:00", reportMetadata.PassableEdgeHubRestartPeriod.ToString());
```
philipktlin: please specify expected value in TimeSpan

test/modules/Modules.Test/TestResultCoordinator/TestReportUtilTest.cs
yophilav: I rename it since the `expectedReportCount` implies the number of report being generated, but that's not what the test case is counting. It is counting the total number of exceptions being thrown in this routine.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
philipktlin: Do we get this information from metric?

we use hard-limit used in test to timeout invoking DM call or send message call and decide if that test run is failed; not based on mean and variation.  Please simplify this report if there is no strong reason to have it.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,309 @@
...
+        static readonly ILogger Logger = ModuleUtil.CreateLogger(nameof(EdgeHubRestartDirectMethodReportGenerator));
+
+        internal EdgeHubRestartDirectMethodReportGenerator(
+            string trackingId,
+            string senderSource,
+            string receiverSource,
+            TestReportType testReportType,
+            ITestResultCollection<TestOperationResult> senderTestResults,
+            ITestResultCollection<TestOperationResult> receiverTestResults,
+            TimeSpan passableEdgeHubRestartPeriod)
```
philipktlin: Please explain why do we need this.  Current timeout should be good enough to decide whether test run is passed or not.

yophilav: This is for direct method which include the SDK connection time for the receiver module reconnection to edgeHub.

test/modules/EdgeHubRestartTester/src/DirectMethodEdgeHubConnectorTest.cs
```diff
@@ -79,7 +79,7 @@ class DirectMethodEdgeHubConnectorTest : IEdgeHubConnectorTest
                     MethodRequest request = new MethodRequest(
                         directMethodName,
                         Encoding.UTF8.GetBytes($"{{ \"Message\": \"Hello\", \"DirectMethodCount\": \"{this.directMethodCount}\" }}"),
-                        Settings.Current.SdkOperationTimeout,
+                        TimeSpan.FromSeconds(5),   // Minimum value accepted by SDK
```
philipktlin: how did you know this? https://github.com/Azure/azure-iot-sdk-csharp/blob/58d2379f8a556e67098f7283223c328822819d9e/iothub/device/src/MethodInvokeRequest.cs#L64; it should accept any number of seconds.

test/modules/EdgeHubRestartTester/src/Program.cs
```diff
@@ -61,17 +61,23 @@ static async Task<int> MainAsync()
 
                     // Setup Message Task
                     Task sendMessageTask = Task.CompletedTask;
-                    sendMessageTask = edgeHubMessageConnector?.StartAsync(
-                        eachTestExpirationTime,
-                        restartTime,
-                        cts.Token);
+                    if (edgeHubMessageConnector != null)
```
philipktlin: Just FYI; you can also do this.

sendMessageTask = edgeHubMessageConnector?.StartAsync(eachTestExpirationTime, restartTime, cts.Token) ?? Task.Completed;

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
philipktlin: So report generator will calculate time duration between edgehub restart timestamp and receiver timestamp; then if  time duration is longer then "passableEdgeHubRestartPeriod", it is failed; otherwise passed. do I understand correctly?  I think this is not worth to do; @varunpuranik what do you think?

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
philipktlin: General comment is to simplify report generation logic; for example, the flow should try focusing on happy path and fail all other cases generally.  If code get complex due to distinguish some edge cases failed results, I would prefer not distinguish them; fail it and report all failed data and investigate later.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReport.cs
philipktlin: Would you please tell me what is not covered by metric?  And this will increase the size of report; which currently histogram is not our test focus now.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReport.cs
```diff
@@ -0,0 +1,75 @@
...
+            ulong receiverMessageCount,
+            Dictionary<HttpStatusCode, List<TimeSpan>> completedStatusHistogram,
+            TimeSpan minPeriod,
+            TimeSpan maxPeriod,
+            TimeSpan medianPeriod,
+            TimeSpan meanPeriod,
+            TimeSpan variancePeriod)
+            : base(trackingId, resultType)
+        {
+            this.isPassing = isPassing;
```
philipktlin: isPassing should not be passing in

test/modules/EdgeHubRestartTester/src/DirectMethodEdgeHubConnectorTest.cs
yophilav: I gives me an error saying the number must be between 5 to 300 seconds.

test/modules/EdgeHubRestartTester/src/Program.cs
yophilav: Yeah. I was thinking about that but it's not as quite readable as explicit `if`. Thanks for the info though.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReport.cs
yophilav: Why not? I'm thinking the report itself should not know anything about the logic to determine if the test is passnig or not. I'm thinking a report as a vessel just to get the report out.

test/modules/Modules.Test/TestResultCoordinator/TestReportUtilTest.cs
yophilav: TODO: I can rename it back.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,73 @@
...
+            this.SenderSource = Preconditions.CheckNonWhiteSpace(senderSource, nameof(senderSource));
+            this.ReceiverSource = Preconditions.CheckNonWhiteSpace(receiverSource, nameof(receiverSource));
+            this.MinPeriod = minPeriod;
+            this.MaxPeriod = maxPeriod;
+            this.MedianPeriod = medianPeriod;
+            this.MeanPeriod = meanPeriod;
+            this.VariancePeriodInMilisec = variancePeriodInMilisec;
+        }
+
+        bool isPassing;
```
varunpuranik: Why do you need this field?

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
varunpuranik: +1 on that. Please keep tests as simple as possible.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
yophilav: I don't need this field. It was there for only the testing purpose.

test/modules/EdgeHubRestartTester/src/DirectMethodEdgeHubConnectorTest.cs
```diff
@@ -25,10 +25,12 @@ class DirectMethodEdgeHubConnectorTest : IEdgeHubConnectorTest
 
         public DirectMethodEdgeHubConnectorTest(
             Guid batchId,
-            ILogger logger)
+            ILogger logger,
+            ModuleClient dmModuleClient = null)
```
philipktlin: why dmModuleClient allows null and optional?  module client should be created in main program and clean in there as well.  Connector test do not need to implement IDisposable.

test/modules/EdgeHubRestartTester/src/MessageEdgeHubConnectorTest.cs
```diff
@@ -24,10 +24,12 @@ class MessageEdgeHubConnectorTest : IEdgeHubConnectorTest, IDisposable
 
         public MessageEdgeHubConnectorTest(
             Guid batchId,
-            ILogger logger)
+            ILogger logger,
+            ModuleClient msgModuleClient = null)
```
philipktlin: Same comment in DirectMethodEdgeHubConnectorTest.cs

test/modules/EdgeHubRestartTester/src/Program.cs
```diff
@@ -37,16 +38,32 @@ static async Task<int> MainAsync()
...
                 }
 
                 if (Settings.Current.DirectMethodEnabled)
                 {
+                    ModuleClient dmModuleClient = await ModuleUtil.CreateModuleClientAsync(
+                        Settings.Current.TransportType,
+                        ModuleUtil.DefaultTimeoutErrorDetectionStrategy,
+                        ModuleUtil.DefaultTransientRetryStrategy,
+                        Logger);
+
```
philipktlin: I think you need to set OperationTimeoutInMilliseconds.  Please extract module client creation code as a method and reuse to create module client for messaging and DM scenario.

test/modules/EdgeHubRestartTester/src/Program.cs
```diff
@@ -37,16 +38,32 @@ static async Task<int> MainAsync()
...
                     edgeHubDirectMethodConnector = new DirectMethodEdgeHubConnectorTest(
                         batchId,
-                        Logger);
+                        Logger,
+                        dmModuleClient);
                 }
 
                 DateTime testStart = DateTime.UtcNow;
```
philipktlin: rename testExpirationTime to testCompletionTime

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
philipktlin: Please remove "completedStatusHistogram" here; you can write this data to log in report generator.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,69 @@
...
+            string senderSource,
+            string receiverSource,
+            ulong senderResultCount,
+            ulong receiverResultCount,
+            Dictionary<HttpStatusCode, List<TimeSpan>> completedStatusHistogram,
+            TimeSpan minPeriod,
+            TimeSpan maxPeriod,
+            TimeSpan medianPeriod,
+            TimeSpan meanPeriod,
+            double variancePeriodInMilisec)
```
philipktlin: why variancePeriodInMillisec use double?

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+                receiverResultCount,
+                this.completedStatusHistogram,
+                minPeriod,
+                maxPeriod,
+                medianPeriod,
+                meanPeriod,
+                variancePeriodInMilisec);
+        }
+
+        async Task<(ulong resultCount, bool isNotEmpty, long sequenceNum)> IncrementSequenceNumberAsync(
```
philipktlin: Please rename this since this method is not just increment sequence number at all.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+            bool isNotEmpty,
```
philipktlin: Rename isNotEmpty to HasResult

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+            MoveNextResultAsync MoveNextResultAsync,
+            long targetSequenceNumber,
+            ulong resultCount)
+        {
+            long seqNum = 0;
+
+            while (isNotEmpty && (seqNum < targetSequenceNumber))
```
philipktlin: use do-while loop instead.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+                throw new InvalidDataException($"Result source is '{result.Source}' but expected should be '{expectedSource}'.");
+            }
+
+            if (!result.Type.Equals(testOperationResultType, StringComparison.Ordinal))
+            {
+                throw new InvalidDataException($"Result type is '{result.Type}' but expected should be '{testOperationResultType}'.");
+            }
+        }
+
+        //////////////////////////////////////////////////////////////// HELPER LAND
```
philipktlin: Clean up

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReport.cs
philipktlin: Please remove "completedStatusHistogram" and below parameters; you can write these data to log in report generator.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportMetadata.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public string SenderSource { get; }
+
+        public string ReceiverSource { get; }
+
+        public string[] ResultSources => new string[] { this.SenderSource, this.ReceiverSource };
+
+        public TestReportType TestReportType => TestReportType.EdgeHubRestartDirectMethodReport;
+
+        public TestOperationResultType TestOperationResultType => (TestOperationResultType)Enum.Parse(typeof(TestOperationResultType), this.SenderSource.Split('.').LastOrDefault());
```
philipktlin: Please hardcore the value here.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+    using Newtonsoft.Json;
+
+    sealed class EdgeHubRestartDirectMethodReportGenerator : ITestResultReportGenerator
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger(nameof(EdgeHubRestartDirectMethodReportGenerator));
+
+        // Value: (completedStatusCode, DirectMethodCompletedTime - EdgeHubRestartedTime)
+        private Dictionary<HttpStatusCode, List<TimeSpan>> completedStatusHistogram;
+
+        delegate Task<(ulong count, bool isNotEmpty, long sequenceNumber)> MoveNextResultAsync(ulong count);
```
philipktlin: create a task to review/refactor to align the type of count and sequence number.  They should use the same type.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+                {
+                    // Increment receiver result to have the same seq as the sender
+                    (receiverResultCount, hasReceiverResult, receiverSeqNum) = await this.IncrementSequenceNumberAsync(
+                        hasReceiverResult,
+                        this.MoveNextReceiverResultAsync,
+                        senderSeqNum,
+                        receiverResultCount);
+                }
+
+                if (hasSenderResult ^ hasReceiverResult)
```
philipktlin: nit
```suggestion
                if (!hasSenderResult || !hasReceiverResult)
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+                // Note: In the current verification, if the receiver obtained more result dm than
+                //   sender, the test is consider a fail. We are disregarding duplicated/unique seqeunce number
+                //   of receiver's result.
+
+                // Check if the current dm is passing
+                bool isCurrentDirectMethodPassing = true;
+                EdgeHubRestartDirectMethodResult senderResult = JsonConvert.DeserializeObject<EdgeHubRestartDirectMethodResult>(this.SenderTestResults.Current.Result);
+
+                // Verified the sequence numbers are the same
+                isCurrentDirectMethodPassing &= senderSeqNum == receiverSeqNum;
```
philipktlin: combine these 3 conditions in a single statement.
```suggestion
                isCurrentDirectMethodPassing = (senderSeqNum == receiverSeqNum) && (senderSeqNum > previousSeqNum) && (senderResult.DirectMethodCompletedStatusCode == HttpStatusCode.OK);
```

test/modules/EdgeHubRestartTester/src/Program.cs
yophilav: DM is SDK timeout is set upon request. 
Message SDK timeout is set upon client creation.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+                long.MaxValue,
+                senderResultCount);
+
+            (receiverResultCount, _, _) = await this.IncrementSequenceNumberAsync(
+                hasReceiverResult,
+                this.MoveNextReceiverResultAsync,
+                long.MaxValue,
+                receiverResultCount);
+
+            return this.CalculateStatistic(
```
philipktlin: Rename this method name; which it turns a test result report, not statistic.  Just write statistic to log file, and put into the report.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
yophilav: I don't want to iterate even once if the `isNotEmpty` is set; hence, the do-while-loop would not quite work here.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportMetadata.cs
yophilav: Oh right, I forgot about this.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
yophilav: I think XOR is more readable to me.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+            TimeSpan completedPeriod = senderResult.DirectMethodCompletedTime - senderResult.EdgeHubRestartedTime;
+            // Try to allocate the list if it is the first time HttpStatusCode shows up
+            this.completedStatusHistogram.TryAdd(completedStatus, new List<TimeSpan>());
+            this.completedStatusHistogram[completedStatus].Add(completedPeriod);
+        }
+
+        long ConvertStringToLong(string result)
+        {
+            long seqNum;
+            long.TryParse(result.Split(';').LastOrDefault(), out seqNum);
```
philipktlin: seq number should be the third element after split; please check if splited result has 3 element exactly and return third element.  Don't use LastOrDefault().

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
yophilav: I'd rather have them spread across 3 lines and be very very explicit of what each passing conditions are instead of combining into one massive if condition. I'm not going to address this comment. Sorry.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+            return seqNum;
+        }
+
+        private async Task<(ulong resultCount, bool hasValue, long sequenceNumber)> MoveNextSenderResultAsync(ulong senderResultCount)
+        {
+            bool hasValue = await this.SenderTestResults.MoveNextAsync();
+            long seqNum = 0;
+            if (hasValue)
```
philipktlin: invert the if statement; immediately return if hasValue is false here.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,303 @@
...
+            return (resultCount: senderResultCount,
+                hasValue: hasValue,
+                sequenceNumber: seqNum);
+        }
+
+        private async Task<(ulong resultCount, bool hasValue, long sequenceNumber)> MoveNextReceiverResultAsync(ulong receiverResultCount)
+        {
+            bool hasValue = await this.ReceiverTestResults.MoveNextAsync();
+            long seqNum = 0;
+            if (hasValue)
```
philipktlin: invert the if statement; immediately return if hasValue is false here.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
philipktlin: The reason to combine is to avoid set isCurrentDirectMethodPassing to true initially.  You are welcome to use other syntax/approach in order to prevent isCurrentDirectMethodPassing initialize to true at the beginning.  Even if-statement is used, these 3 conditions can still be written in different lines, I think it is still very clear.  Please kindly address.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,319 @@
...
+                {
+                    // Increment sender result to have the same seq as the receiver
+                    (senderMessageCount, hasSenderResult, senderSeqNum) = await this.IncrementSequenceNumberAsync(
+                        hasSenderResult,
+                        this.MoveNextSenderResultAsync,
+                        receiverSeqNum,
+                        senderMessageCount);
+                }
+
+                if (receiverSeqNum < senderSeqNum)
```
philipktlin: This case should not be happened in message scenario.  The sender report seq number should be same as number of restart count.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,319 @@
...
+                //   of receiver's result.
+
+                // Check if the current message is passing
+                bool isCurrentMessagePassing = true;
+
+                EdgeHubRestartMessageResult senderResult = JsonConvert.DeserializeObject<EdgeHubRestartMessageResult>(this.SenderTestResults.Current.Result);
+                string receiverResult = this.ReceiverTestResults.Current.Result;
+
+                // Verified "TrackingId;BatchId;SequenceNumber" altogether.
+                isCurrentMessagePassing &= senderResult.GetMessageTestResult() == receiverResult;
```
philipktlin: use string.Eqauls with string comparison option

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportMetadata.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public string SenderSource { get; }
+
+        public string ReceiverSource { get; }
+
+        public string[] ResultSources => new string[] { this.SenderSource, this.ReceiverSource };
+
+        public TestReportType TestReportType => TestReportType.EdgeHubRestartMessageReport;
+
+        public TestOperationResultType TestOperationResultType => (TestOperationResultType)Enum.Parse(typeof(TestOperationResultType), this.SenderSource.Split('.').LastOrDefault());
```
philipktlin: Please hardcore the value here.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
yophilav: https://msazure.visualstudio.com/One/_workitems/edit/6274658

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
yophilav: I agree but let's keep this here just in case.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,319 @@
...
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    sealed class EdgeHubRestartMessageReportGenerator : ITestResultReportGenerator
+    {
+        static readonly ILogger Logger = ModuleUtil.CreateLogger(nameof(EdgeHubRestartMessageReportGenerator));
+
+        // Value: (completedStatusCode, DirectMethodCompletedTime - EdgeHubRestartedTime)
+        private Dictionary<HttpStatusCode, List<TimeSpan>> completedStatusHistogram;
```
varunpuranik: How is the histogram being used?

varunpuranik: Nit: Don't mark fields as private

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,319 @@
...
+                isCurrentMessagePassing &= senderResult.MessageCompletedStatusCode == HttpStatusCode.OK;
+
+                // If this message passed, increment the count for a good message
+                passedMessageCount += isCurrentMessagePassing ? 1UL : 0UL;
+
+                // Make sure the sequence number is incremental
+                isIncrementalSeqeunce &= (previousSeqNum + 1) == senderSeqNum;
+                previousSeqNum++;
+
+                (senderMessageCount, hasSenderResult, senderSeqNum) = await this.MoveNextSenderResultAsync(senderMessageCount);
```
varunpuranik: Nit: Use inline declarations = (ulong resultCount, bool isNotEmpty, long sequenceNum)

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,319 @@
...
+                foreach (TimeSpan eachTimeSpan in orderedCompletedPeriods)
+                {
+                    totalSpan += eachTimeSpan;
+                    totalSpanSquareInMilisec += Math.Pow(eachTimeSpan.TotalMilliseconds, 2);
+                }
+
+                // Compute Mean : mean = sum(x) / N
+                meanPeriod = totalSpan / Math.Max(orderedCompletedPeriods.Count(), 1);
+
+                // Compute Sample Variance: var = sum((x - mean)^2) / (N - 1)
```
varunpuranik: Why do we need this? How are we using this?

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,319 @@
...
+                throw new InvalidDataException($"Result source is '{result.Source}' but expected should be '{expectedSource}'.");
+            }
+
+            if (!result.Type.Equals(testOperationResultType, StringComparison.Ordinal))
+            {
+                throw new InvalidDataException($"Result type is '{result.Type}' but expected should be '{testOperationResultType}'.");
+            }
+        }
+
+        //////////////////////////////////////////////////////////////// HELPER LAND
```
varunpuranik: Avoid comments like this, pls. The method name should indicate the helper nature of it. Maybe you can make it static.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
varunpuranik: +1 on Philip's comment. DOn't use bit operators like XOR on boolean variables.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
varunpuranik: I am confused here. The aim of this test is to check EdgeHub startup time. How does duplicate methods have anything to do with that?

varunpuranik:
I would like to understand the logic of this test, and test verification. The aim of the test is to check EdgeHub startup time. The test and the verification seems to be doing way more than that.

test/modules/EdgeHubRestartTester/src/Program.cs
philipktlin: Thanks for explanation. @davilu would you please confirm that OperationTimeoutInMilliseconds is not used for Direct Method call?

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public ulong SenderResultCount { get; }
+
+        public ulong ReceiverResultCount { get; }
+
+        public string SenderSource { get; }
+
+        public string ReceiverSource { get; }
+
+        public Dictionary<HttpStatusCode, List<TimeSpan>> CompletedStatusHistogram { get; }
```
philipktlin: Please remove.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,51 @@
...
+            this.SenderSource = Preconditions.CheckNonWhiteSpace(senderSource, nameof(senderSource));
+            this.ReceiverSource = Preconditions.CheckNonWhiteSpace(receiverSource, nameof(receiverSource));
+            this.MedianPeriod = medianPeriod;
+        }
+
+        public override string Title => $"{this.ResultType} Report between {this.SenderSource} and {this.ReceiverSource}";
+
+        public override bool IsPassed => (this.PassedDirectMethodCount == this.SenderResultCount) && (this.SenderResultCount > 0);
+
+        public ulong PassedDirectMethodCount { get; }
```
philipktlin:
```suggestion
        public ulong PassedCount { get; }
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,258 @@
...
+                        receiverResultCount);
+                }
+
+                if (hasSenderResult ^ hasReceiverResult)
+                {
+                    break;
+                }
+
+                // Note: In the current verification, if the receiver obtained more result dm than
+                //   sender, the test is consider a fail. We are disregarding duplicated/unique seqeunce number
```
philipktlin:
```suggestion
                //   sender, the test is considered as test failure . We are disregarding duplicated/unique sequence number
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        long ConvertStringToLong(string result)
+        {
+            long seqNum;
+            long.TryParse(result, out seqNum);
+            return seqNum;
+        }
+
+        private async Task<(ulong resultCount, bool hasValue, long sequenceNumber)> MoveNextSenderResultAsync(ulong senderResultCount)
```
philipktlin: remove private access modifier

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,258 @@
...
+            seqNum = this.ConvertStringToLong(senderResult.SequenceNumber);
+
+            this.AddEntryToCompletedStatusHistogram(senderResult);
+
+            return (resultCount: senderResultCount,
+                hasValue: hasValue,
+                sequenceNumber: seqNum);
+        }
+
+        private async Task<(ulong resultCount, bool hasValue, long sequenceNumber)> MoveNextReceiverResultAsync(ulong receiverResultCount)
```
philipktlin: remove private access modifier

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReport.cs
```diff
@@ -0,0 +1,53 @@
...
+            this.MedianPeriod = medianPeriod;
+        }
+
+        public override string Title => $"{this.ResultType} Report between {this.SenderSource} and {this.ReceiverSource}";
+
+        public override bool IsPassed => this.IsIncrementalSeqeunce && (this.PassedMessageCount == this.SenderMessageCount) && (this.SenderMessageCount > 0);
+
+        public bool IsIncrementalSeqeunce { get; }
+
+        public ulong PassedMessageCount { get; }
```
philipktlin:
```suggestion
        public ulong PassedCount { get; }
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public ulong SenderResultCount { get; }
```
philipktlin:
```suggestion
        public ulong SenderCount { get; }
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public ulong ReceiverResultCount { get; }
```
philipktlin:
```suggestion
        public ulong ReceiverCount { get; }
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReport.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public ulong SenderMessageCount { get; }
```
philipktlin:
```suggestion
        public ulong SenderCount { get; }
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReport.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public ulong ReceiverMessageCount { get; }
```
philipktlin:
```suggestion
        public ulong ReceiverCount { get; }
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
philipktlin: Without this logic, if this case really happens, I think the report will show all result not matched. right?  If that's the case, please remove this logic.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,272 @@
...
+                {
+                    // Increment receiver result to have the same seq as the sender
+                    (receiverMessageCount, hasReceiverResult, receiverSeqNum) = await this.IterateResultToSequenceNumberAsync(
+                        hasReceiverResult,
+                        this.MoveNextReceiverResultAsync,
+                        senderSeqNum,
+                        receiverMessageCount);
+                }
+
+                if (hasSenderResult ^ hasReceiverResult)
```
philipktlin: Please don't use bitwise operation here.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,258 @@
...
+        {
+            long seqNum;
+            long.TryParse(result, out seqNum);
+            return seqNum;
+        }
+
+        private async Task<(ulong resultCount, bool hasValue, long sequenceNumber)> MoveNextSenderResultAsync(ulong senderResultCount)
+        {
+            bool hasValue = await this.SenderTestResults.MoveNextAsync();
+            long seqNum = 0;
```
philipktlin: Combine assignment in line 227

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            if (!hasValue)
+            {
+                return (resultCount: senderResultCount,
+                    hasValue: hasValue,
+                    sequenceNumber: seqNum);
```
philipktlin: should return original sequence number rather than zero.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,258 @@
...
+            ulong resultCount)
+        {
+            long seqNum = 0;
+
+            while (hasValue && (seqNum < targetSequenceNumber))
+            {
+                (resultCount, hasValue, seqNum) = await MoveNextResultAsync(resultCount);
+            }
+
+            return (resultCount: resultCount,
```
philipktlin: should return original sequence number rather or return null or option type, but not zero.

yophilav: Given the current use cases for the sequence number, making sequence number nullable/Option<> does not add anything to the code since the `hasValue` is required to be checked before accessing the seqNum to get a meaningful value.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
philipktlin: Quote from Bear's reply:

If I want to return a previous value, I'd need to deserialize the result which can be expensive operation given the routine
	if I want to return null, then I need to make sure the caller can handle Null
	if I do Option, then the caller needs to handle option type properly

Why is 0 not a sufficient given the flag hasValue is also a return param?

I mean the seqNum is essentially an Option<> with respect to hasValue flag.

philipktlin: That's fair.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+                {
+                    // Increment sender result to have the same seq as the receiver
+                    (senderMessageCount, hasSenderResult, senderSeqNum) = await this.IterateResultToSequenceNumberAsync(
+                        hasSenderResult,
+                        this.MoveNextSenderResultAsync,
+                        receiverSeqNum,
+                        senderMessageCount);
+                }
+
+                if ((receiverSeqNum < senderSeqNum) && hasSenderResult)
```
philipktlin: This cannot be happened.  If this if-block is removed, and this condition happens, the report should report all unmatched results. right? if so, please remove it.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+            Logger.LogInformation($"Generating report: {nameof(EdgeHubRestartMessageReport)} for [{this.SenderSource}] and [{this.ReceiverSource}]");
+
+            bool isIncrementalSeqeunce = true;
+            long previousSeqNum = 0;
+            ulong passedMessageCount = 0;
+            ulong senderMessageCount = 0;
+            ulong receiverMessageCount = 0;
+            bool hasSenderResult = true;
+            long senderSeqNum = 0;
+            bool hasReceiverResult = true;
```
philipktlin: default to false.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        internal ITestResultCollection<TestOperationResult> SenderTestResults { get; }
+
+        internal ITestResultCollection<TestOperationResult> ReceiverTestResults { get; }
+
+        public async Task<ITestResultReport> CreateReportAsync()
+        {
+            Logger.LogInformation($"Generating report: {nameof(EdgeHubRestartMessageReport)} for [{this.SenderSource}] and [{this.ReceiverSource}]");
+
+            bool isIncrementalSeqeunce = true;
```
philipktlin: default to false.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,273 @@
...
+            long previousSeqNum = 0;
+            ulong passedMessageCount = 0;
+            ulong senderMessageCount = 0;
+            ulong receiverMessageCount = 0;
+            bool hasSenderResult = true;
```
philipktlin: default to false.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
philipktlin: change it to sth, e.g. "HasGap", then default to false.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
philipktlin: Please address if instead of bitwise operator in both report generator.  thanks.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
yophilav: Ok.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
philipktlin: please put in multiple lines

builds/e2e/connectivity.yaml
```diff
@@ -19,38 +19,43 @@ jobs:
           testrun.network.runProfile: "Online"
           testrun.network.frequencies: "00:00:00 00:00:00 0"
           testrun.duration: "01:00:00"
-        Offline(5/5/5):
+        Offline:
           testrun.config: 2
+          testrun.network.runProfile: "Offline"
```
yophilav: 1. Remove this case OFFLINE
2. Introduce the dup + backlog items to deploy w/ separate connectivity deployment file (use foreach log).

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
yophilav: TODO: Centralize this conditions for verification.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,286 @@
...
+                    break;
+                }
+
+                bool isCurrentMessagePassing = this.VerifyCurrentResult(
+                    senderSeqNum,
+                    receiverSeqNum);
+
+                // Make sure the sequence number is incremental
+                isDiscontinuousSequenceNumber |= (previousSeqNum + 1) != senderSeqNum;
+                previousSeqNum++;
```
philipktlin:
```suggestion
                previousSeqNum = senderSeqNum;
```

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartMessageReportGenerator.cs
```diff
@@ -0,0 +1,286 @@
...
+
+                // If this message passed, increment the count for a good message
+                passedMessageCount += isCurrentMessagePassing ? 1UL : 0UL;
```
philipktlin: if isCurrentMessagePassing is used only here, please remove it and directly call this.VerifyCurrentResult method.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReportGenerator.cs
```diff
@@ -0,0 +1,275 @@
...
+                    break;
+                }
+
+                bool isCurrentDirectMethodPassing = this.VerifyCurrentResult(
+                    senderSeqNum,
+                    receiverSeqNum,
+                    previousSeqNum);
+
+                // If the current DM result is passed, increment the count for a good dm
+                passedDirectMethodCount += isCurrentDirectMethodPassing ? 1UL : 0UL;
```
philipktlin: please replace isCurrentDirectMethodPassing by calling VerifyCurrentResutl directly.

PR #2498, opened 2020-02-10
[k8s] Make Docker volume declarations work as anticipated

darobs:
Recent testing revealed how Edge on k8s was handling Docker volumes wrong. 

We also noticed that there is only a 1:1 mapping between PVCs and PVs - you can only bind one PVC to a PV, and we had expected you could do more than one.  I also discovered that if the user mounts the same volume twice in a container, this will also cause trouble in the Edge on k8s  deployment.

In docker (single node) mode, if a user mounts the same volume in 2 modules, they will expect that the modules will access the same directory. This PR fixes this problem.

This PR does the following: 
- If user starts runtime with PersistentVolume name, all volume mounts get the same PVC mount. 
- If user starts runtime with a StorageClass name only, each volume is a unique PVC mount.
- If the user starts runtime with neither, each volume is a unique PVC mount - users are expected to provide PVC for each volume.

With specifying a PV name, there is a caveat that all modules will be given that same file share to access - because the user gave only one file share to use.

This will break existing public preview deployments, since it will tear down existing PVCs and replace them with new PVCs.


edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/edgedeployment/pvc/KubernetesPvcByValueComparerTest.cs
```diff
@@ -139,79 +140,94 @@ public void FieldComparisonTests()
...
         {
-            var x = new V1PersistentVolumeClaim
+            (var x, var y) = MakeIdenticalPVCs();
+            Assert.True(Comparer.Equals(x, y));
+        }
+
+        [Fact]
+        public void DistictionTest()
+        {
+            (var x, var y) = MakeIdenticalPVCs();
```
dmolokanov:
```suggestion
            var (x, y) = MakeIdenticalPVCs();
```

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/edgedeployment/pvc/KubernetesPvcByValueComparerTest.cs
```diff
@@ -139,79 +140,94 @@ public void FieldComparisonTests()
...
+            var distinct = new List<V1PersistentVolumeClaim> { x, y }.Distinct(Comparer);
```
dmolokanov:
```suggestion
            var claims = new List<V1PersistentVolumeClaim> { x, y };
            
            var distinct = claims.Distinct(Comparer);
```

darobs: Interesting. VS suggested `new List<V1PersistentVolumeClaim> { x, y }.Distinct(Comparer);` 
I write it as 2 lines.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -82,6 +82,13 @@ public async Task<EdgeDeploymentStatus> DeployModulesAsync(ModuleSet desiredModu
                     [KubernetesConstants.K8sEdgeHubNameLabel] = KubeUtils.SanitizeLabelValue(this.resourceName.Hostname)
                 };
 
+                var desiredServiceAccounts = desiredModules.Modules
```
CindyXing: So the serviceaccount is needed before creating PVC. Thanks for finding this.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -359,25 +361,28 @@ static string ToProxyLogLevel(LogEventLevel level)
 
         V1Volume GetVolume(KubernetesModule module, Mount mount)
         {
-            string volumeName = KubeUtils.SanitizeK8sValue(mount.Source);
+            // Volume name will be pvName if persistent volume is set, else will be volumeName
+            string volumeName = this.persistentVolumeName
+                .Map(KubeUtils.SanitizeK8sValue)
+                .GetOrElse(() => KubeUtils.SanitizeDNSValue(mount.Source));
+            // PVC name will be volume name
+            string pvcName = volumeName;
```
CindyXing: In the past, we hit issue that two modules share the same mount name; and resulted in a clash. How are we addressing this scenario with this change?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
darobs: It's not required.  We should move this up because the deployment references it.  (We have another PBI for this, I chose to change it now since I was testing agent E2E.)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
darobs: That will be a problem if the user chooses a storage class that only allows ReadWriteOnce.  If the user chooses to attempt the same volume twice in a deployment, and it's RWOnce, there's a conflict regardless of our implementation. 

I also discovered that part of the problem wasn't that K8s had a problem with 2 pods mounting the same PVC, the problem was that our code was trying to create the same PVC twice.  It was our fault all along. :(

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/edgedeployment/pvc/KubernetesPvcByValueComparerTest.cs
dmolokanov: Yeah, it's kinda Arrage, Act, Assert thing

PR #2499, opened 2020-02-10
Make logs upload method since take relative times

lfitchett:
This changes the since parameter for built in log upload to accept relative times: https://github.com/Azure/iotedge/blob/master/doc/built-in-logs-pull.md

### Edgelet
It moves the parse_since function in support bundle to a shared func in core and uses it for logs.

### EdgeAgent
Chances since from an int to a string. This is directly passed to edgelet.
Also changes management yaml


edgelet/api/managementVersion_2019_11_05.yaml
```diff
@@ -298,9 +298,9 @@ paths:
           default: "all"
         - in: query
           name: since
-          description: Only return logs since this time, as a UNIX timestamp.
-          type: integer
-          default: 0
+          description: Only return logs since this time, as a duration (1 day, 1d, 90m, 2 days 3 hours 2 minutes), rfc3339 timestamp, or UNIX timestamp.
```
arsing: Did you confirm that you don't need to bump the API version for this?

edgelet/iotedge/src/main.rs
```diff
@@ -362,7 +360,8 @@ fn run() -> Result<(), Error> {
             let since = args
                 .value_of("since")
                 .map(|s| parse_since(s))
-                .transpose()?
+                .transpose()
+                .map_err(|err| Error::from(err.context(ErrorKind::BadSinceParameter)))?
```
arsing:
```diff,rust
-                .map_err(|err| Error::from(err.context(ErrorKind::BadSinceParameter)))?
+                .context(ErrorKind::BadSinceParameter)?
```

edgelet/api/managementVersion_2019_11_05.yaml
lfitchett: Yes, confirmed w/ varun and mike that it's ok.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesRuntimeInfoProviderTest.cs
```diff
@@ -185,21 +185,6 @@ public void ConstructorChecksNull()
             Assert.Throws<ArgumentNullException>(() => new KubernetesRuntimeInfoProvider("namespace", client.Object, null));
         }
 
-        [Fact]
-        public async void GetModuleLogsTest()
```
arsing: Why are you removing it?

lfitchett: I removed the functionality from edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesRuntimeInfoProvider.cs. It was never fully implemented for k8s, and after talking to david we decided to remove it. The feature isn't very useful on k8s, which has its own ways to get logs.

Issue #2313, opened 2020-01-15
iotedge check incorrectly shows an error when using DPS


jlorich:
Running `iotedge check` when configured with symmetric key DPS displays the following problem:

```
× config.yaml has well-formed connection string - Error
    Device is not using manual provisioning, so Azure IoT Hub hostname needs to be specified with --iothub-hostname
```

This occurs even though Edge is running perfectly fine against DPS. An IoT Hub hostname should explicitly _not_ be specified with DPS.  I believe this to be an error with the `iotedge check` command.

veyalla:
There currently isn't a way for the tool to discovery which hub the device is assigned to when using DPS. There are scenarios where customers know which hub their device will be assigned to so they can specify the address using `--iothub-name` to verify connectivity to it.

Does making this a warning instead of an error help?

arsing:
>An IoT Hub hostname should explicitly *not* be specified with DPS.

That reasoning applies to the actual provisioning done by `iotedged`, not to `iotedge check`. `iotedge check` does not do provisioning itself, so it has no way to know what the IoT Hub hostname is for the "Can the device connect to the IoT Hub?" tests.

>Does making this a warning instead of an error help?

Not unless we want to also disable the connectivity tests.

JeeWeetje:
@veyalla I think a warning is a better option. But with the explanation you give about - - iothub-name option and url to dps provioning iot edge docs (as - - verbose output or just normal output? ).

arsing:
>>Does making this a warning instead of an error help?
>
>Not unless we want to also disable the connectivity tests.

@veyalla and I discussed offline and he asked for the connectivity tests to also downgrade to warnings if `--iothub-hostname` is not specified..

Note that this means we'll not run some tests just because the user ran the tool incorrectly. It also means users whose devices genuinely can't connect to IoT Hub for whatever reason (aggressive firewall rules, etc - the kind of things the connectivity tests are intended to catch in the first place) will see no errors from `iotedge check`.

JeeWeetje:
Would it be an option to show that if the user does not provide the ```--iothub-name``` parameter when testing a DPS registered edge:

### normal warning
```
Configuration checks
--------------------
√ config.yaml is well-formed - OK
!!! config.yaml has well-formed connection string - Warning
    Device is not using manual provisioning, so Azure IoT Hub hostname needs to be specified with --iothub-hostname. The connectivity tests can not be run without the --iothub-name parameter.
...
Connectivity checks
-------------------
[...]

PR #2501, opened 2020-02-10
Change DPS provisioning check error to warning

lfitchett:
In the context of https://github.com/Azure/iotedge/issues/2313 

Users often find that particular error and message a bit jarring/confusing. I propose downgrading this to a warning and updating the warning message to be more descriptive:
```Device is configured with automatic provisioning, in this configuration 'iotedge check' is not able to discover the device's backing IoT Hub. To run connectivity checks in this configuration please specify the backing IoT Hub name using --iothub-hostname switch if you have that information.​```

It also changes the documentation for certs in the config yaml. While file paths still work, we want users to only use uris for consistency with the new x509.

### DIFF
--- a/edgelet/contrib/config/linux/config.yaml
+++ b/edgelet/contrib/config/linux/config.yaml
@@ -162,9 +162,8 @@ provisioning:
 #                                       Defaults to 90 days.
 #
 # Note:
-# The values of all of these fields can be specified either as a
-# "file" scheme URI such as "file:///path/cert_key.pem" or a
-#  file path such as "/path/cert_key.pem"
+# The values of all of these fields must be specified as a
+# "file" scheme URI such as "file:///path/cert_key.pem"
 ###############################################################################
 
 # certificates:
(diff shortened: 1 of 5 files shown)


arsing:
>It also changes the documentation for certs in the config yaml, since non-file path urls aren't accepted.

I don't understand. This sentence implies "URLs other than the file scheme aren't accepted", which is true, but the existing comment in the config.yaml doesn't say that other schemes are accepted in the first place, so there doesn't seem to be anything that needed to be changed.

Then the actual change is removing the mention of file paths, as if only URIs are accepted, which is also not true. Both paths and URIs are accepted.

arsing:
>I propose downgrading this to a warning and updating the warning message to be more descriptive:

I'm down for making the message more descriptive, but I still disagree with making it a warning, for the reason I gave in https://github.com/Azure/iotedge/issues/2313#issuecomment-574817425 . If people don't agree to specify `--iothub-hostname` (as in that issue) after reading the error, I don't think making it a warning is going to convince them. So they'll get a successful result out from `iotedge check` because it didn't run any of the connectivity tests, then wonder why they have connectivity problems.

I believe if you use a tool incorrectly, the tool should complain. If we absolutely want to support users running `iotedge check` without `--iothub-hostname`, it should be opt-in, not the default.

PR #2506, opened 2020-02-11
enforce SDK device client to use tls 1.2

davilu:


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -58,6 +59,8 @@ static async Task<int> MainAsync(IConfigurationRoot configuration)
             SslProtocols sslProtocols = SslProtocolsHelper.Parse(sslProtocolsConfig, DefaultSslProtocols, logger);
             logger.LogInformation($"Enabling SSL protocols: {sslProtocols.Print()}");
 
+            TlsVersions.SetMinimumTlsVersions(SslProtocols.Tls12);
```
varunpuranik: We should also do it for EdgeAgent.

PR #2509, opened 2020-02-12
Fix old e2e VM proxy test: Adapt to new service sdk exceptions 

and-rewsmith:
The new service SDK 1.18.4 throws a new exception only when configured with a proxy. I am editing our quickstart application to adjust to this, however will also open a bug noting that this is a breaking change.


and-rewsmith:
Closing as this new behavior is an SDK bug

PR #2510, opened 2020-02-12
Add metrics for direct methods (#2294)

and-rewsmith:
Cherry-pick #2294
* add direct method metrics


and-rewsmith:
Holding off on these backports. Not going into 109.

PR #2511, opened 2020-02-12
Remove device descriptors from cloud proxy logging (#2296)

and-rewsmith:
Cherry pick: #2296
* remove cloud proxy misleading descriptors


and-rewsmith:
Holding off on these backports. Not going into 109.

PR #2516, opened 2020-02-13
Removing workaround for SDK cert authentication bug

dylanbronson:
We will be receiving a new SDK version that contains a fix for a problem that we made a workaround for. The workaround was to attach a callback to transport settings in order to validate certificates, when normal certificate validation was not previously working.

This removes that workaround (https://github.com/Azure/iotedge/commit/9fe8eeea7c92fc6f801ab1cc19ea917b2564c0e8). It also removes the workaround from our LeafDevice tests.


smoke/LeafDevice/details/Details.cs
```diff
@@ -120,18 +120,7 @@ protected Task InitializeTrustedCertsAsync()
         {
             if (!string.IsNullOrEmpty(this.trustedCACertificateFileName))
             {
-                // Since Windows will pop up security warning when add certificate to current user store location;
-                // Therefore we will use CustomCertificateValidator instead.
-                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
philipktlin, line 125: did you test on Windows? I think the reason we need this logic here is because it is not possible to install CA certificate on Windows silently (without human interaction).  Therefore we use the callback approach.

philipktlin: It will be great if we can automate to install cert on Windows. If so, you can remove these code.

dylanbronson: Quickstart Certs still fails for Windows. Which is odd because TransparentGateway test passes with workaround removed.

PR #2517, opened 2020-02-13
[k8s] Better PVC story for iotedged.

darobs:
For HA, name is required.  If only name is given, we will mount the PVC
name in iotedged Deployment, but will not create a PVC.
If name and storageclass and size are given, we will create a PVC in the
chart. VolumeName is optional.


darobs:
> Looks good, presuming you've tested that it works :)

Yes, yes I did. :)

PR #2518, opened 2020-02-13
Fix CentOS arm32v7 and aarch64 packages to depend on correct openssl sonames

### LINKED ISSUE #2398: How to install IoTEdge for centos7 ARM on raspbian Pi
I tried installing IoT Edge for Centos ARM32 on Raspberry PI. During installation of the libiothsm-std-1.0.9-0.1.rc3.el7.armv7hl.rpm
rpm -ivh libiothsm-std-1.0.9-0.1.rc3.el7.armv7hl.rpm
error: Failed dependencies:
libcrypto.so.1.0.0 is needed by libiothsm-std-1.0.9-0.1.rc3.el7.armv7hl
libssl.so.1.0.0 is needed by libiothsm-std-1.0.9-0.1.rc3.el7.armv7hl

arsing:
Before this change, these two targets were built using Linaro compilers
and against custom-compiled openssl. However the packages built this way
ended up depending on `libcrypto.so.1.0.0` and `libssl.so.1.0.0`,
whereas CentOS's openssl-libs package provides `libcrypto.so.10` and
`libssl.so.10`

Since CentOS 7 does not ship cross-compiler packages and openssl,
this commit changes the build steps to run the native arm32v7 and arm64v8
CentOS Docker containers in qemu instead of cross-compiling.

Since qemu builds are very slow (2h30m instead of 15m), this commit moves them
to a new pipeline job so that it does not block existing downstream jobs like
E2E tests that don't need the CentOS packages.

Lastly, the new job runs on Ubuntu 18.04 hosts instead of Ubuntu 16.04 hosts.
Ubuntu 16.04 supports getrandom, but its qemu is just old enough to
not implement it, which breaks processes inside the container that require it.

Fixes #2398

### DIFF
--- a/builds/misc/packages.slow.yaml
+++ b/builds/misc/packages.slow.yaml
@@ -1,0 +1,71 @@
+# This is like packages.yaml, but for packages that are much slower to build.
+# This runs independently of the packages.yaml job so that it does not impact other jobs that use the packages.yaml artifacts, like E2E test runs.
+
+trigger:
+  batch: true
+  branches:
+    include:
+      - master
+pr: none
+variables:
+  REVISION: '1'
+jobs:
+
+################################################################################
+  - job: linux
+################################################################################
+    displayName: Linux
+    timeoutInMinutes: 180 # Observed to take just under 2h30m, so round up
+    pool:
+      # qemu on ubuntu-16.04 is too old to support some syscalls used by CentOS 7 containers, so ensure this is ubuntu-18.04 or newer.
+      vmImage: 'ubuntu-18.04'
+    strategy:
+      matrix:
+        Centos75-arm32v7:
+          arch: arm32v7
+          os: centos7
+          target.iotedged: edgelet/target/rpmbuild/RPMS/armv7hl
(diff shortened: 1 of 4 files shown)


arsing:
Package build: https://dev.azure.com/msazure/One/_build/results?buildId=28703443&view=results

The arm32 build took 2h30m. I need to fix the arm64 build and some other stuff, but it'll take about the same time. In comparison all the other Linuces took 15m, and even Windows only takes 1h.

Are we okay with making these the packages build take 2h30m now?

Venkat suggested we could disable these two in the packages build and only run them in the release build. The downside is that if the release build breaks, we won't know until release time.

Alternatively we can make a separate daily build for these two.

@darobs, @myagley: Thoughts?

arsing:
Talked with myagley offline and he agreed with making a separate build job for CentOS packages.

arsing:
I still need to set up the new pipeline in VSTS, but [here's](https://dev.azure.com/msazure/One/_build/results?buildId=28763946&view=results) a successful run from when it was still part of the regular packages job.

PR #2519, opened 2020-02-13
Removing unused property ConnectionIdleTimeout

dylanbronson:
Removing unused property ConnectionIdleTimeout that was causing a build error with the new SDK


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -16,8 +16,6 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
 
     public class CloudConnectionProvider : ICloudConnectionProvider
     {
-        // Minimum value allowed by the SDK for Connection Idle timeout for AMQP Multiplexed connections.
-        static readonly TimeSpan MinAmqpConnectionMuxIdleTimeout = TimeSpan.FromSeconds(5);
```
varunpuranik: Interesting.. I could have sworn this was being used in the earlier versions of the SDK at least.

PR #2524, opened 2020-02-13
Import VSTS pipeline status

philipktlin:
Add new project for import vsts pipeline data to database for test dashboard.

### DIFF
--- a/tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
+++ b/tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 namespace DevOpsLib
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
@@ -38,25 +39,62 @@ namespace DevOpsLib
 
             // TODO: need to think about how to handle unexpected exception during REST API call
             string requestPath = string.Format(LatestBuildPathSegmentFormat, this.accessSetting.Organization, this.accessSetting.Project);
-            IFlurlRequest latestBuildRequest = DevOpsAccessSetting.BaseUrl
-                .AppendPathSegment(requestPath)
-                .SetQueryParam("definitions", string.Join(",", buildDefinitionIds.Select(b => b.IdString())))
-                .SetQueryParam("queryOrder", "finishTimeDescending")
-                .SetQueryParam("maxBuildsPerDefinition", "1")
-                .SetQueryParam("api-version", "5.1")
-                .SetQueryParam("branchName", branchName)
+            IFlurlRequest latestBuildRequest = GetBuildsRequestUri(buildDefinitionIds, branchName, requestPath, DateTime.MinValue, 1)
                 .WithBasicAuth(string.Empty, this.accessSetting.PersonalAccessToken);
 
             string resultJson = await latestBuildRequest.GetStringAsync().ConfigureAwait(false);
             JObject result = JObject.Parse(resultJson);
 
-            if (!result.ContainsKey("count") || (int) result["count"] <= 0)
+            if (!result.ContainsKey("count") || (int)result["count"] <= 0)
             {
                 return buildDefinitionIds.Select(i => VstsBuild.CreateBuildWithNoResult(i, branchName)).ToList();
             }
 
             Dictionary<BuildDefinitionId, VstsBuild> latestBuilds = JsonConvert.DeserializeObject<VstsBuild[]>(result["value"].ToString()).ToDictionary(b => b.DefinitionId, b => b);
             return buildDefinitionIds.Select(i => latestBuilds.ContainsKey(i) ? latestBuilds[i] : VstsBuild.CreateBuildWithNoResult(i, branchName)).ToList();
+        }
+
+        public async Task<IList<VstsBuild>> GetBuildsAsync(HashSet<BuildDefinitionId> buildDefinitionIds, string branchName, DateTime minTime, int? maxBuildsPerDefinition = null)
+        {
+            ValidationUtil.ThrowIfNulOrEmptySet(buildDefinitionIds, nameof(buildDefinitionIds));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(branchName, nameof(branchName));
+
+            // TODO: need to think about how to handle unexpected exception during REST API call
+            string requestPath = string.Format(LatestBuildPathSegmentFormat, this.accessSetting.Organization, this.accessSetting.Project);
+            IFlurlRequest latestBuildRequest = GetBuildsRequestUri(buildDefinitionIds, branchName, requestPath, minTime, maxBuildsPerDefinition)
+                .WithBasicAuth(string.Empty, this.accessSetting.PersonalAccessToken);
+
+            string resultJson = await latestBuildRequest.GetStringAsync().ConfigureAwait(false);
+            JObject result = JObject.Parse(resultJson);
+
+            if (!result.ContainsKey("count") || (int)result["count"] <= 0)
+            {
+                return buildDefinitionIds.Select(i => VstsBuild.CreateBuildWithNoResult(i, branchName)).ToList();
+            }
+
+            return JsonConvert.DeserializeObject<VstsBuild[]>(result["value"].ToString()).ToList();
+        }
+
+        private static Url GetBuildsRequestUri(HashSet<BuildDefinitionId> buildDefinitionIds, string branchName, string requestPath, DateTime minTime, int? maxBuildsPerDefinition)
+        {
+            Url requestUri = DevOpsAccessSetting.BaseUrl
+                .AppendPathSegment(requestPath)
+                .SetQueryParam("definitions", string.Join(",", buildDefinitionIds.Select(b => b.IdString())))
+                .SetQueryParam("queryOrder", "finishTimeDescending")                
+                .SetQueryParam("api-version", "5.1")
+                .SetQueryParam("branchName", branchName);
+
+            if (maxBuildsPerDefinition.HasValue)
+            {
+                requestUri = requestUri.SetQueryParam("maxBuildsPerDefinition", "1");
+            }
+
+            if (minTime > DateTime.MinValue)
+            {
+                requestUri = requestUri.SetQueryParam("minTime", minTime.ToString("o"));
+            }
+
+            return requestUri;
         }
     }
 }
--- a/tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
+++ b/tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
@@ -39,7 +39,7 @@ namespace DevOpsLib
 
             // TODO: need to think about how to handle unexpected exception during REST API call
             string requestPath = string.Format(LatestBuildPathSegmentFormat, this.accessSetting.Organization, this.accessSetting.Project);
-            IFlurlRequest latestBuildRequest = GetBuildsRequestUri(buildDefinitionIds, branchName, requestPath, DateTime.MinValue, 1)
+            IFlurlRequest latestBuildRequest = GetBuildsRequestUri(buildDefinitionIds, branchName, requestPath, null, 1)
                 .WithBasicAuth(string.Empty, this.accessSetting.PersonalAccessToken);
 
             string resultJson = await latestBuildRequest.GetStringAsync().ConfigureAwait(false);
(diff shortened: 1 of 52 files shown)


tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
```diff
@@ -0,0 +1,122 @@
...
+            var lastUpdatePerBranchPerDefinition = new Dictionary<string, Dictionary<BuildDefinitionId, DateTime>>();
+
+            while (!ct.IsCancellationRequested)
+            {
+                try
+                {
+                    Console.WriteLine($"Import Vsts Builds data started at {DateTime.UtcNow}");
+
+                    lastUpdatePerBranchPerDefinition.Upsert("refs/heads/master", await ImportVstsBuildsDataAsync(buildManagement, "refs/heads/master", BuildExtension.MasterBranchBuildDefinitions,  lastUpdatePerBranchPerDefinition));
+                }
```
and-rewsmith: nit: I found this confusing when I first read it. Might make it easier to split into two lines.

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
```diff
@@ -0,0 +1,122 @@
...
+            ValidationUtil.ThrowIfNull(devOpsAccessSetting, nameof(devOpsAccessSetting));
+
+            this.devOpsAccessSetting = devOpsAccessSetting;
+            this.dbConnectionString = dbConnectionString;
+        }
+
+        public async Task RunAsync(TimeSpan waitPeriodAfterEachUpdate, CancellationToken ct)
+        {
+            var buildManagement = new BuildManagement(devOpsAccessSetting);
+            var lastUpdatePerBranchPerDefinition = new Dictionary<string, Dictionary<BuildDefinitionId, DateTime>>();
```
and-rewsmith: nit: this could be a member variable and then it we wouldn't need to define it inside the func and pass it as an arg to another func.

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            Dictionary<BuildDefinitionId, DateTime> lastUpdatePerDefinition = lastUpdatePerBranchPerDefinition.GetIfExists(branch);
+            if (lastUpdatePerDefinition == null)
+            {
+                lastUpdatePerDefinition = new Dictionary<BuildDefinitionId, DateTime>();
+            }
+
+            foreach (BuildDefinitionId buildDefinitionId in buildDefinitionIds)
+            {
+                DateTime maxLastChange = await ImportVstsBuildsDataForSpecificDefinitionAsync(buildManagement, branch, buildDefinitionId, lastUpdatePerDefinition.GetIfExists(buildDefinitionId));
```
and-rewsmith: I'm assuming this will default to DateTime.MinValue according to link?
https://stackoverflow.com/questions/13957701/new-datetime-vs-defaultdatetime

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
```diff
@@ -0,0 +1,122 @@
...
+                {
+                    var cmd = new SqlCommand
+                    {
+                        Connection = sqlConnection,
+                        CommandType = System.Data.CommandType.StoredProcedure,
+                        CommandText = "UpsertVstsBuildRun"
+                    };
+
+                    cmd.Parameters.Add(new SqlParameter("@BuildNumber", build.BuildNumber));
+                    cmd.Parameters.Add(new SqlParameter("@DefinitionId", build.DefinitionId));
```
and-rewsmith: We are going to need the readable name here

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
philipktlin: yes, correct.

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
philipktlin: done

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
```diff
@@ -38,25 +39,62 @@ public async Task<IList<VstsBuild>> GetLatestBuildsAsync(HashSet<BuildDefinition
...
+
+            string resultJson = await latestBuildRequest.GetStringAsync().ConfigureAwait(false);
+            JObject result = JObject.Parse(resultJson);
+
+            if (!result.ContainsKey("count") || (int)result["count"] <= 0)
+            {
+                return buildDefinitionIds.Select(i => VstsBuild.CreateBuildWithNoResult(i, branchName)).ToList();
+            }
+
+            return JsonConvert.DeserializeObject<VstsBuild[]>(result["value"].ToString()).ToList();
```
yophilav: Do we need to do `result.ContainsKey("value")` check before accessing it?

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
```diff
@@ -38,25 +39,62 @@ public async Task<IList<VstsBuild>> GetLatestBuildsAsync(HashSet<BuildDefinition
...
+        private static Url GetBuildsRequestUri(HashSet<BuildDefinitionId> buildDefinitionIds, string branchName, string requestPath, DateTime minTime, int? maxBuildsPerDefinition)
+        {
+            Url requestUri = DevOpsAccessSetting.BaseUrl
+                .AppendPathSegment(requestPath)
+                .SetQueryParam("definitions", string.Join(",", buildDefinitionIds.Select(b => b.IdString())))
+                .SetQueryParam("queryOrder", "finishTimeDescending")                
+                .SetQueryParam("api-version", "5.1")
+                .SetQueryParam("branchName", branchName);
+
+            if (maxBuildsPerDefinition.HasValue)
```
yophilav:
``` suggestion
            if (!maxBuildsPerDefinition.HasValue)
```

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
```diff
@@ -38,25 +39,62 @@ public async Task<IList<VstsBuild>> GetLatestBuildsAsync(HashSet<BuildDefinition
...
+            {
+                requestUri = requestUri.SetQueryParam("maxBuildsPerDefinition", "1");
+            }
+
+            if (minTime > DateTime.MinValue)
```
yophilav:
``` suggestion
            if (minTime >= DateTime.MinValue)
```

tools/IoTEdgeDevOps/DevOpsLibTest/BuildManagementTest.cs
```diff
@@ -35,7 +35,7 @@ public void TearDown()
         }
 
         [Test]
-        public async Task TestGetLatestBuildsAsyncWithNoDefinitionId()
+        public void TestGetLatestBuildsAsyncWithNoDefinitionId()
         {
             ArgumentException ex = Assert.ThrowsAsync<ArgumentException>(
```
yophilav: Does `Assert.ThrowsAsync<ArgumentException>()` has an implicit await for the await-task you are passing in?

tools/IoTEdgeDevOps/VstsPipelineSync/DictionaryExtension.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace VstsPipelineSync
+{
+    using System.Collections.Generic;
+
+    public static class DictionaryExtension
+    {
+        public static TValue GetIfExists<TKey, TValue>(this Dictionary<TKey, TValue> dict, TKey key)
```
yophilav: Did you mean to do this?
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netframework-4.8

tools/IoTEdgeDevOps/VstsPipelineSync/DictionaryExtension.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            if (dict.ContainsKey(key))
+            {
+                return dict[key];
+            }
+
+            return default(TValue);
+        }
+
+        public static bool Upsert<TKey, TValue>(this Dictionary<TKey, TValue> dict, TKey key, TValue value)
```
yophilav: Did you mean to do this? 
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.tryadd?view=netcore-3.1

Note: please read the "Remark" section

tools/IoTEdgeDevOps/VstsPipelineSync/Program.cs
```diff
@@ -0,0 +1,28 @@
...
+        static async Task Main(string[] args)
+        {
+            (string pat, string dbConnectionString, TimeSpan waitPeriodBeforeNextUpdate) = GetInputsFromArgs(args);
+            Console.WriteLine($"Wait period before next update=[{waitPeriodBeforeNextUpdate}]");
+            await new VstsBuildBatchUpdate(new DevOpsAccessSetting(pat), dbConnectionString).RunAsync(waitPeriodBeforeNextUpdate, CancellationToken.None);
+        }
+
+        private static (string pat, string dbConnectionString, TimeSpan waitPeriodBeforeNextUpdate) GetInputsFromArgs(string[] args)
+        {
+            if (args.Length != 3)
```
yophilav: What's the format of this `args` gonna look like? What's in it? Maybe add a comment there on what format of data is expecting?

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
```diff
@@ -38,25 +39,62 @@ public async Task<IList<VstsBuild>> GetLatestBuildsAsync(HashSet<BuildDefinition
...
+            if (!result.ContainsKey("count") || (int)result["count"] <= 0)
             {
                 return buildDefinitionIds.Select(i => VstsBuild.CreateBuildWithNoResult(i, branchName)).ToList();
             }
 
             Dictionary<BuildDefinitionId, VstsBuild> latestBuilds = JsonConvert.DeserializeObject<VstsBuild[]>(result["value"].ToString()).ToDictionary(b => b.DefinitionId, b => b);
             return buildDefinitionIds.Select(i => latestBuilds.ContainsKey(i) ? latestBuilds[i] : VstsBuild.CreateBuildWithNoResult(i, branchName)).ToList();
         }
+
+        public async Task<IList<VstsBuild>> GetBuildsAsync(HashSet<BuildDefinitionId> buildDefinitionIds, string branchName, DateTime minTime, int? maxBuildsPerDefinition = null)
```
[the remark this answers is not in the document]
philipktlin: For test dashboard scenario, we want to show history as well; therefore we need to get as many as API returns.

tools/IoTEdgeDevOps/VstsPipelineSync/DictionaryExtension.cs
and-rewsmith: From the remarks:
"TryAdd doesn't override the element if the element with the given key exists in the dictionary"

So this is doing something different.

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
philipktlin: let it fail if it doesn't exist; means something wrong with VSTS API result.

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
philipktlin: Good catch.

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
```diff
@@ -38,25 +39,62 @@ public async Task<IList<VstsBuild>> GetLatestBuildsAsync(HashSet<BuildDefinition
...
+
+            if (!result.ContainsKey("count") || (int)result["count"] <= 0)
+            {
+                return buildDefinitionIds.Select(i => VstsBuild.CreateBuildWithNoResult(i, branchName)).ToList();
+            }
+
+            return JsonConvert.DeserializeObject<VstsBuild[]>(result["value"].ToString()).ToList();
+        }
+
+        private static Url GetBuildsRequestUri(HashSet<BuildDefinitionId> buildDefinitionIds, string branchName, string requestPath, DateTime minTime, int? maxBuildsPerDefinition)
```
and-rewsmith: I'm assuming you opted for nullable vs Option because it was simpler?

tools/IoTEdgeDevOps/VstsPipelineSync/DictionaryExtension.cs
philipktlin: Any obvious benefit? I prefer not to use out param.

and-rewsmith: Why is out not desirable?

tools/IoTEdgeDevOps/DevOpsLibTest/BuildManagementTest.cs
philipktlin: I guess so.  I changed this method to void as VS2019 gives warning/suggestion.  And I ran all tests and all passed.

tools/IoTEdgeDevOps/VstsPipelineSync/DictionaryExtension.cs
philipktlin: out keyword is to show intent explicitly this parameter will be updated inside the method.  I try my best to avoid changing parameter value of a method.  That's why I prefer not to use out.  Even you can't find this syntax in Java.  It is just my general practice.

tools/IoTEdgeDevOps/DevOpsLib/BuildManagement.cs
philipktlin: yes

PR #2525, opened 2020-02-13
Accumulate and maintain route priorities per endpoint

richma-ms:
This takes the routes that were parsed in the previous change, and creates a mapping of each endpoint to their respective list of priorities.  It also handles the case where priorities are changed on routes (or routes added/removed) after initial deployment by re-calculating the priorities for any affected endpoint.

The following files have the relevant logic, everything else is updating calling signatures and test collateral:
- Router.cs
- Dispatcher.cs
- StoringAsyncEndpointExecutor.cs


edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Router.cs
```diff
@@ -181,11 +187,24 @@ protected virtual void Dispose(bool disposing)
...
-            config.Fallback.ForEach(f => endpoints.Add(f.Endpoint));
-            return endpoints;
+            var endpoints = new HashSet<(Endpoint, IList<uint>)>(config.Endpoints.Select(e => (e, GetPrioritiesForEndpoint(e, config.Routes))));
+            config.Fallback.ForEach(f => endpoints.Add((f.Endpoint, new List<uint>() { RouteFactory.DefaultPriority })));
+
+            // De-dupe for the case where the fallback route uses the same
+            // endpoint as a normal route
+            return endpoints
+                .GroupBy(e => e.Item1)
+                .Select(dupe => dupe.First())
```
dylanbronson: should this be a merge of all the dupes, instead of just taking the first one and discarding the rest?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Router.cs
```diff
@@ -181,11 +187,24 @@ protected virtual void Dispose(bool disposing)
...
+                .Where(r => r.Endpoint == endpoint)
+                .Select(r => r.Priority).ToList();
+        }
+
+        static ISet<(Endpoint, IList<uint>)> GetEndpointsWithPriority(RouterConfig config)
         {
-            var endpoints = new HashSet<Endpoint>(config.Routes.Select(r => r.Endpoint));
-            config.Fallback.ForEach(f => endpoints.Add(f.Endpoint));
-            return endpoints;
+            var endpoints = new HashSet<(Endpoint, IList<uint>)>(config.Endpoints.Select(e => (e, GetPrioritiesForEndpoint(e, config.Routes))));
```
dylanbronson: maybe break this into 2 statements for readability

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Router.cs
richma-ms: just realized that this is prob better served as a dictionary, then we don't have the deduping concerns

---
In reply to: [381575308](https://github.com/Azure/iotedge/pull/2525#discussion_r381575308) [](ancestors = 381575308)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Router.cs
```diff
@@ -181,10 +191,31 @@ protected virtual void Dispose(bool disposing)
             }
         }
 
-        static ISet<Endpoint> GetEndpoints(RouterConfig config)
+        static void MergeEndpointPriorities(ref Dictionary<Endpoint, IList<uint>> existing, Endpoint endpoint, uint newPriority)
```
varunpuranik: Why the ref?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -122,8 +132,23 @@ public async Task SetEndpoint(Endpoint newEndpoint)
...
 
+        void UpdatePriorities(IList<uint> priorities)
+        {
+            // Update priorities by merging the new ones with the existing.
+            // We don't ever remove stale priorities, otherwise stored messages
+            // pending for a removed priority will never get sent.
+            IList<uint> snapshot = this.priorities.Value;
+            this.priorities.Value = priorities
+                .Union(snapshot)
+                .GroupBy(p => p)
```
dylanbronson: Do we need GroupBy if we have Union?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Dispatcher.cs
```diff
@@ -48,29 +48,28 @@ public class Dispatcher : IDisposable
 
         ImmutableDictionary<string, IEndpointExecutor> Executors => this.executors;
 
-        public static async Task<Dispatcher> CreateAsync(string id, string iotHubName, ISet<Endpoint> endpoints, IEndpointExecutorFactory factory)
+        public static async Task<Dispatcher> CreateAsync(string id, string iotHubName, IDictionary<Endpoint, IList<uint>> endpointsWithPriorities, IEndpointExecutorFactory factory)
```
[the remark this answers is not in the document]
richma-ms: talked offline, this is a list of priorities accumulated from multiple routes targeting the same endpoint, so that the executor knows how many underlying queues/checkpointers to maintain.

---
In reply to: [381807857](https://github.com/Azure/iotedge/pull/2525#discussion_r381807857) [](ancestors = 381807857)

PR #2530, opened 2020-02-14
Incorrect metrics timespan config doesn't crash

lfitchett:
Fixes bug 6022799


edge-util/src/Microsoft.Azure.Devices.Edge.Util/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        static readonly ILogger Log = Logger.Factory.CreateLogger(typeof(ConfigurationExtensions));
+
+        public static T GetValueIgnoreException<T>(this IConfiguration configuration, string key, T defaultValue)
+        {
+            try
+            {
+                return configuration.GetValue(key, defaultValue);
+            }
+            catch (Exception ex)
```
damonbarry: We don't want to ignore _all_ exceptions, just the ones that specifically come from malformed input. `TimeSpan.Parse` docs say it can throw `FormatException` or `OverflowException`; let's catch those (and add tests to validate). Since we're catching specific exceptions from `TimeSpan.Parse`, let's also scope this extension method down to just `TimeSpan` instead of `T`. Finally after all that, a more appropriate name might be `GetTimeSpanValue`?

lfitchett: Changed to be timespan specific. It will be hard to unit test, since GetValue is an extension function and Moq can't mock it. It's a simple enough function I think its safe leaving it untested.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+            when (IsParseException(ex))
+            {
+                Log.LogWarning(ex, $"Could not parse timespan for {key}");
+                return defaultValue;
+            }
+        }
+
+        static bool IsParseException(Exception ex)
+        {
+            return ex != null && (ex is OverflowException || ex is FormatException || IsParseException(ex.InnerException));
```
damonbarry: Did you observe that the parse exception is sometimes (or always) wrapped in another exception, or is this theoretical?

edge-util/src/Microsoft.Azure.Devices.Edge.Util/ConfigurationExtensions.cs
damonbarry: Have a look at all the `edge-util/test/Microsoft.Azure.Devices.Edge.Util.Test/*ExTest.cs` files...it's easy to get some basic unit testing in place for the extension class, no mocks needed.

damonbarry: Started to worry that it might be hard to test (due to Logger?), so tried it really quick. It worked an now I have some tests on my machine, so I pushed them. ;)

edge-util/src/Microsoft.Azure.Devices.Edge.Util/ConfigurationExtensions.cs
lfitchett: Observed

damonbarry: Ah I see, Configuration wraps it in `InvalidOperationException`. We could simplify things by catching InvalidOperationException (no filtering), but this is fine too.

PR #2540, opened 2020-02-18
Update the Direct Method Seqeunce number for connectivity test to be Ulong

yophilav:
- Update the Direct Method Seqeunce number for connectivity test to be Ulong


philipktlin:
Would you please change DirectMethodTestResult.SequenceNumber from string to ulong?

test/modules/DirectMethodReceiver/src/DirectMethodReceiver.cs
```diff
@@ -69,7 +69,7 @@ public async Task ReportTestResult(string directMethodCount)
                             DateTime.UtcNow,
                             this.trackingId.GetOrElse(string.Empty),
                             this.batchId,
-                            directMethodCount,
+                            ulong.Parse(directMethodCount),
```
kkdawkins: Is this wise? If the .Parse fails, it will throw an exception. Should you do ulong.TryParse and handle the error case yourself?

yophilav: I was thinking about that as well.

Background: 
In our test scenarios, there are multiple test modules that send Direct Method and use DirectMethodReceiver as a target in a test.

In my opinion, I intentionally use `.Parse()` and not the `.TryParse()` as a way to strictly enforce that the sender must a number range within `ulong`; otherwise, the DirectMethodReceiver would not work at all.

PR #2546, opened 2020-02-19
[1.0.9] Limit the RocksDb LOG file size

yophilav:
- Limit Rocks DB Log file size
- Set the default RocksDB's default log file to be 1 file, 0 MB
- Make the maximum number of log files and the size of the log file configurable from the module (EdgeAgent & EdgeHub).

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Constants.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Constants.cs
@@ -105,6 +105,10 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Core
 
         public const string StorageMaxTotalWalSize = "RocksDB_MaxTotalWalSize";
 
+        public const string StorageMaxLogFileNum = "Storage_MaxLogFileNum";
+
+        public const string StorageMaxLogFileSize = "Storage_MaxLogFileSize";
+
         public const string WorkloadApiVersion = "2019-01-30";
 
         public static class Labels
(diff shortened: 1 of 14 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Constants.cs
```diff
@@ -105,6 +105,10 @@ public static class Constants
 
         public const string StorageMaxTotalWalSize = "RocksDB_MaxTotalWalSize";
 
+        public const string StorageMaxLogFileNum = "Storage_MaxLogFileNum";
```
varunpuranik: RocksDB_ to be consistent

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbOptionsProvider.cs
```diff
@@ -44,16 +44,29 @@ public class RocksDbOptionsProvider : IRocksDbOptionsProvider
         // Once write-ahead logs exceed this size, we will start forcing the flush of
         // column families whose memtables are backed by the oldest live WAL file
         const ulong DefaultMaxTotalWalSize = 512 * 1024 * 1024;
+        // Logging to ONE file as a default
+        const ulong DefaultMaxLogFileNum = 1;
```
varunpuranik: Can we set this to 0?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbOptionsProvider.cs
```diff
@@ -62,7 +75,9 @@ public DbOptions GetDbOptions()
                 .SetCreateIfMissing()
                 .SetCreateMissingColumnFamilies();
 
-            options.SetMaxTotalWalSize(this.maxTotalWalSize);
+            options.SetMaxTotalWalSize(this.maxTotalWalSize)
+                .SetKeepLogFileNum(this.maxLogFileNum)
```
varunpuranik: Is there a TotalLog file size option?

yophilav:
Per isolated RocksDB option tests, these parameters being set do not result in LOG file size limitation as expected.  This PR is closed; a new PR will be issued in its place.

edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbOptionsProvider.cs
yophilav: 0 would be a default value to RocksDB which is interpreted as no-limits.

PR #2552, opened 2020-02-19
[1.0.9] Use LogLevel to limit RocksDB LOG file size

yophilav:
- Use Allow LogLevel to be configurable using an environment variable from a module
- Set default RocksDB logging to be None ( Create the minimal 15KB LOG file and leave it alone without logging any activity to it.)


edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbInfoLogLevel.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Storage.RocksDb
+{
+    // Ref: https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/InfoLogLevel.java
+    public enum RocksDbInfoLogLevel
```
ancaantochi: Please name this class StorageLogLevel and move it the Storage project

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Constants.cs
```diff
@@ -105,6 +105,8 @@ public static class Constants
 
         public const string StorageMaxTotalWalSize = "RocksDB_MaxTotalWalSize";
 
+        public const string StorageLogLevel = "Storage_LogLevel";
```
ancaantochi: Would StoreAndForward_LogLevel naming be more clear?

yophilav: 1. I think the structure (or tuble) `StoreAndForward` should really be broken into 2 tuples.
2. Since the LogLevel is only use for the RocksDb (and not for RoutingModule or the other module), I think the Storage_LogLevel is reasonable. 

Now that I think about it, I think I will break the `StoreAndForward` into `StoreConfig` and `???Config`

yophilav: I think `Storage_LogLevel` is justified since the config does not use for the `Forwarding` part. The parameter is only used for `CommonModule` but not for `RoutingModule` and `MqttModule`

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/StorageLogLevel.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Storage
+{
+    // Ref: https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/InfoLogLevel.java
+    public enum StorageLogLevel
```
varunpuranik: RocksDbLogLevel

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
```diff
@@ -320,19 +322,19 @@ static string GetLocalConfigFilePath(IConfiguration configuration, ILogger logge
...
-                storageMaxTotalWalSize = configuration.GetValue<ulong>(Constants.StorageMaxTotalWalSize);
+                storageParamValue = configuration.GetValue<T>(fieldName);
             }
             catch
             {
                 // ignored
             }
 
-            return storageMaxTotalWalSize <= 0 ? Option.None<ulong>() : Option.Some(storageMaxTotalWalSize);
+            return EqualityComparer<T>.Default.Equals(storageParamValue, default(T)) ? Option.None<T>() : Option.Some(storageParamValue);
```
ancaantochi: please test if you provide a negative value the conguration.GetValue throws so this method will return Option.None<T>

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/StorageLogLevel.cs
yophilav: https://github.com/Azure/iotedge/pull/2552#discussion_r381604100

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
```diff
@@ -320,19 +322,19 @@ static string GetLocalConfigFilePath(IConfiguration configuration, ILogger logge
             return localConfigPath;
         }
 
-        static Option<ulong> GetStorageMaxTotalWalSizeIfExists(IConfiguration configuration)
+        static Option<T> GetStorageConfigIfExists<T>(string fieldName, IConfiguration configuration)
```
philipktlin: why do you need to make it with type T?  Just return Option<ulong>

yophilav: It's used for multiple types. One is an enum, the other is ulong

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
yophilav: I tested with value (-1) and the LOG is using `NONE` config.

PR #2554, opened 2020-02-20
Update `make dist` to vendor rust dependencies

myagley:
`make dist` uses `cargo vendor` to grab all of the dependency source and include it in the tarball. This coupled with the `rust-toolchain` file will allow reproducible and offline builds (assuming the rust toolchain is installed).

The old `make dist` target (using git) was renamed to `rpm-dist` as this was used by the rpm build. This was done to limit any impact to the rpm build

To create a dist tarball in `target/release` run `make dist`


arsing:
I kicked off a packages build. Waiting for that.

arsing:
If you run `make dist` with an enlistment that already has build outputs under `target/`, it ends up trying to `tar` those too. Is it possible to skip that directory?

Now that we have this, should we think about creating proper source deb / rpm packages too?

edgelet/Makefile
```diff
@@ -61,22 +63,22 @@ release:
...
+dist:
+	$(MKDIR_P) $(TARGET)
+	$(MKDIR_P) .cargo
+	$(MKDIR_P) $(vendordir)
+	$(CARGO) vendor $(vendordir) > .cargo/config
+	$(TAR) --transform "flags=r;s/^\./$(PACKAGE_NAME)-$(VERSION)/" -c -f $(TARGET)/$(PACKAGE_NAME)-$(VERSION).tar -v --exclude-vcs --exclude-vcs-ignores .
+
+	# Update the version.txt in the archive with $VERSION
+	echo "${VERSION}" > $(TARGET)/version.txt
+	$(TAR) --update -f $(TARGET)/$(PACKAGE_NAME)-$(VERSION).tar --transform 'flags=r;s|$(TARGET)/version.txt|$(PACKAGE_NAME)-$(VERSION)/version.txt|' $(TARGET)/version.txt
```
arsing: Should we include `LICENSE` and `THIRDPARTYNOTICES` too?

myagley:
> If you run `make dist` with an enlistment that already has build outputs under `target/`, it ends up trying to `tar` those too. Is it possible to skip that directory?
> 
> Now that we have this, should we think about creating proper source deb / rpm packages too?

It should skip the `target` dir if you run `make dist` from the context of the git repo. If you run `make dist` after untar the tarball, it'll include the `target` because it uses `tar --exclude-vcs-ignore`. I'll update to exclude `target` explicitly.

+1 on doing proper deb/rpm source packages. This will make things much easier to package.

myagley:
@arsing I believe I addressed your comments. It now includes the `LICENSE` and `THIRDPARTYNOTICES` and explicitly excludes `target`.

arsing:
>It should skip the `target` dir if you run `make dist` from the context of the git repo. If you run `make dist` after untar the tarball, it'll include the `target` because it uses `tar --exclude-vcs-ignore`.

It *was* in the context of the git repo, and I saw it packaging up a bunch of `.rlib`s before I killed it. I'll double-check when I come in.

arsing:
Ah, never mind. I forgot it's because I set `CARGO_TARGET_DIR` to use different directories depending on target arch and for clippy, so the default `.gitignore` doesn't cover them.

PR #2558, opened 2020-02-20
Update network test info reporting

philipktlin:
Update network test info reporting.


philipktlin:
Tested https://msazure.visualstudio.com/One/_build/results?buildId=28948248&view=logs&j=6ecc137b-3a85-553a-1aab-7cdc03381ddc&t=b063a85b-e6b4-50ca-fe62-f176da76457a.  It works as expected, testinfo get reported to TRC.

PR #2562, opened 2020-02-21
Introduce MQTT EdgeHub Restart Tester to connectivity deployment file

yophilav:
Introduce MQTT EdgeHub Restart Tester to connectivity deployment file


test/modules/EdgeHubRestartTester/src/DirectMethodEdgeHubConnectorTest.cs
```diff
@@ -69,7 +72,7 @@ class DirectMethodEdgeHubConnectorTest : IEdgeHubConnectorTest
             DateTime runExpirationTime,
             CancellationToken cancellationToken)
         {
-            ModuleClient moduleClient = await this.GetModuleClientAsync();
+            ModuleClient moduleClient = this.dmModuleClient;
```
philipktlin: remove this and just use it directly in line 87.

test/modules/EdgeHubRestartTester/src/EdgeHubConnectorConfig.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace EdgeHubRestartTester
+{
+    using System;
+    using Microsoft.Azure.Devices.Client;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    class EdgeHubConnectorConfig
+    {
+        public EdgeHubConnectorConfig(
```
philipktlin: I suggest to rename since this class contains config for both direct method and message connector.
```suggestion
        public EdgeHubConnectorsConfig(
```

test/modules/EdgeHubRestartTester/src/MessageEdgeHubConnectorTest.cs
```diff
@@ -96,7 +82,7 @@ TestResultReportingClient GetReportClient()
             DateTime runExpirationTime,
             CancellationToken cancellationToken)
         {
-            ModuleClient moduleClient = await this.GetModuleClientAsync();
+            ModuleClient moduleClient = this.msgModuleClient;
```
philipktlin: remove this line and directly use it in line 98

test/modules/EdgeHubRestartTester/src/Settings.cs
```diff
@@ -51,9 +46,19 @@ class Settings
             this.TestStartDelay = testStartDelay;
             this.TrackingId = Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
 
-            if (!(this.DirectMethodEnabled || this.MessageEnabled))
+            this.GetConnectorConfigAsync().Wait();
```
philipktlin: Try not to use Wait() method.  Please change ConnectorConfig property as private field; and make GetConnectorConfigAsync as public method; you can cache the result in ConnectorCOnfig private field inside GetConnectorConfigAsync method.

PR #2563, opened 2020-02-21
Support bundle can output to console

lfitchett:
This gives support bundle the ability to output to stdout instead of a file. This will be used to collect the bundle from a direct method. Allowing console output removes the need to write the bundle to disk


arsing:
It's a common convention to have `--output -` to mean "write to stdout". You don't need to add a new parameter.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -75,45 +87,80 @@ where
...
             zip_writer,
+            output_location: self.output_location,
+        })
+    }
+
+    fn make_vector_state(self) -> Result<BundleState<M, Cursor<Vec<u8>>>, Error> {
+        let file_options =
+            zip::write::FileOptions::default().compression_method(zip::CompressionMethod::Deflated);
+
+        let zip_writer = zip::ZipWriter::new(Cursor::new(Vec::new()));
```
kkdawkins: Should we worry that this project is no longer actively maintained? https://github.com/mvdnes/zip-rs

kkdawkins: This means that we must track security issues/bugs/etc, right?

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -47,25 +51,33 @@ where
...
+
+        match self.output_location {
+            OutputLocation::File(_) => Box::new(
+                Self::bundle_all(future::result(self.make_file_state())).map(|state| {
+                    let path = PathBuf::from(state.output_location.get_file_location());
+                    println!(
+                        "Created support bundle at {}",
+                        path.canonicalize()
+                            .unwrap_or_else(|_| path)
+                            .to_string_lossy()
```
dmolokanov: Path implements Display trait. Is this conversion required?

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -22,21 +23,24 @@ use crate::Command;
...
+{
     runtime: M,
     log_options: LogOptions,
-    location: PathBuf,
     include_ms_only: bool,
     verbose: bool,
     iothub_hostname: Option<String>,
     file_options: zip::write::FileOptions,
-    zip_writer: zip::ZipWriter<File>,
+    zip_writer: zip::ZipWriter<W>,
```
dmolokanov: import zip module maybe?

edgelet/iotedge/src/support_bundle.rs
arsing: `Path` doesn't impl `Display`, but it does have a `display()` method that would be better to use here than `to_string_lossy()` (the result of `Display` never allocates a `String`).

edgelet/iotedge/src/support_bundle.rs
lfitchett: This could be a possible alternative, it has a very similar interface: https://crates.io/crates/mini-fs. However, it has 900 lifetime downloads vs 1.1 million: https://crates.io/crates/zip. I think it makes sense to stick with zip for now.

lfitchett:
> It's a common convention to have `--output -` to mean "write to stdout". You don't need to add a new parameter.

I wanted to match the pattern of check (--output switches between text and json). I think it makes sense to keep the flags consistent within the tool. Thoughts?

arsing:
`check`'s switch is for an output format. There's no way to get that information without a switch. For `support-bundle`, the only difference is where it writes to, which can be expressed with the existing switch. Like I said, `-` for stdout (or stdin if it's an input) is a common convention, so if anything special-casing it is the more intuitive approach.

edgelet/iotedge/src/support_bundle.rs
kkdawkins: Perhaps, but I am still concerned with the support burdeon

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -58,9 +58,7 @@ where
                     let path = PathBuf::from(state.output_location.get_file_location());
                     println!(
```
dmolokanov: I thought something about this (no direct call to display())
```rust
println!("Created support bundle at {:?}", path.canonicalize().unwrap_or_else(|_| path));
```

lfitchett: image

lfitchett: :? works, but is escapes the string ```C:\\Users\\Lee\\source``` instead of ```C:\Users\Lee\source```

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -200,20 +257,25 @@ where
...
+                        output_location,
                     };
                     state.print_verbose(&format!("Wrote {} logs to file", module_name));
                     state
                 })
             })
     }
 
-    fn write_edgelet_log_to_file(mut state: BundleState<M>) -> Result<BundleState<M>, Error> {
+    fn write_edgelet_log_to_file<W>(
```
arsing: `s/to_file/to_writer/g`

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -22,21 +23,24 @@ use crate::Command;
...
     log_options: LogOptions,
-    location: PathBuf,
     include_ms_only: bool,
     verbose: bool,
     iothub_hostname: Option<String>,
-    file_options: zip::write::FileOptions,
-    zip_writer: zip::ZipWriter<File>,
+    file_options: FileOptions,
+    zip_writer: ZipWriter<W>,
+    output_location: OutputLocation,
```
arsing: Doesn't look like this field is ever used.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -75,45 +85,78 @@ where
...
             include_ms_only: self.include_ms_only,
             verbose: self.verbose,
             iothub_hostname: self.iothub_hostname,
             file_options,
             zip_writer,
+            output_location: self.output_location,
+        })
+    }
+
+    fn make_vector_state(self) -> Result<BundleState<M, Cursor<Vec<u8>>>, Error> {
```
arsing: Might be cleaner to make this `fn make_state<W>(self, writer: W) -> Result<BundleState<M, W>, Error>` and have the caller pass in the appropriate writer. Then you don't have two methods that are otherwise nearly identical.

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -75,45 +85,78 @@ where
...
+    {
+        state
+            .and_then(Self::write_all_logs)
+            .and_then(Self::write_edgelet_log_to_file)
+            .and_then(Self::write_docker_log_to_file)
+            .and_then(Self::write_check_to_file)
+            .and_then(Self::write_all_inspects)
+            .and_then(Self::write_all_network_inspects)
+    }
+    fn write_all_logs<W>(
```
arsing: Missing empty line between these two functions.

edgelet/iotedge/src/support_bundle.rs
lfitchett: Talked to arnav, he said it's ok for now

PR #2567, opened 2020-02-21
Fix staged deployment in end-to-end tests

damonbarry:
The end-to-end tests were recently updated (see #2248 and #2442) to do deployments in two stages: the first just adds the system modules (including any edge hub routes the test will need), and the second adds the remaining modules. This was done to give edge hub a chance to get routes set up before test devices/modules begin using them.

Following the changes, output showed that the staged deployments were happening as intended, but they weren't. A deployment config really breaks down into two major parts: edge agent's desired properties (which describe all the modules in the deployment), and all the other modules' desired properties (e.g. edge hub routes, test module config, etc.). The first deployment contained the first part and the second deployment contained the first and second parts, which means that all modules were being sent down to the device in the first deployment.

This change fixes the problem by calling the internal function `EdgeConfigBuilder.BuildEdgeAgent` twice--once for each deployment. Each call receives only the modules that are expected to start with that deployment (just system modules first, all modules second).

I also added constants for `$edgeAgent` and `$edgeHub`.


test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
```diff
@@ -59,32 +59,48 @@ public IModuleConfigBuilder AddModule(string name, string image)
...
+                    {
+                        ["properties.desired"] = module.DesiredProperties
+                    };
+                }
+            }
+
+            // Edge agent is not optional; add if necessary
+            if (!this.moduleBuilders.ContainsKey(ModuleName.EdgeAgent))
+            {
+                this.AddEdgeAgent();
```
lfitchett: Should this throw an error instead? Is it even possible to create a deployment w/o agent with this framework?

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
```diff
@@ -59,32 +59,48 @@ public IModuleConfigBuilder AddModule(string name, string image)
...
+            foreach (ModuleConfiguration module in modules)
             {
-                AddModuleToConfiguration(module, moduleNames, moduleImages, config);
+                UpdateConfiguration(module);
             }
 
-            if (modulesLookup.Contains("other"))
+            if (moduleConfigs.Contains("other"))
             {
                 Dictionary<string, IDictionary<string, object>> modulesContent =
```
lfitchett: Recommend using copy constructor here

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
damonbarry: This logic was already present; I just moved it into this function from `BuildEdgeAgent`. Yes, you can create a test that doesn't add edge agent, in which case it would get added here as a convenience. I don't think you could write such a test by using the higher-level `EdgeRuntime` class, but you could absolutely do it using the lower level classes directly.

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfigBuilder.cs
damonbarry: I can do that

damonbarry: @lfitchett Done

PR #2569, opened 2020-02-22
Fix agent matrix

philipktlin:
allow agent capability with empty value and add connectiviyt test column in agent matrix

### DIFF
--- a/tools/IoTEdgeDevOps/IoTEdgeDashboard/.config/dotnet-tools.json
+++ b/tools/IoTEdgeDevOps/IoTEdgeDashboard/.config/dotnet-tools.json
@@ -1,0 +1,5 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {}
+}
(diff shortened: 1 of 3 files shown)


tools/IoTEdgeDevOps/IoTEdgeDashboard/.config/dotnet-tools.json
```diff
@@ -0,0 +1,5 @@
+{
```
philipktlin: this file is automatically added by VS, which is introduced in .net core 3.1

varunpuranik: Do we need it? If not, we can remove it.

philipktlin: Leave it as-is; otherwise I believe it will add to the project again when open in VS2019.

PR #2570, opened 2020-02-22
Check the wait in EdgeHub restarter to not run into a negative number

yophilav:


test/modules/EdgeHubRestartTester/src/Program.cs
```diff
@@ -100,7 +100,8 @@ static async Task<int> MainAsync()
                     await Task.WhenAll(taskList);
 
                     // Wait until the specified restart period to do another restart
-                    await Task.Delay((int)(eachTestExpirationTime - DateTime.UtcNow).TotalMilliseconds, cts.Token);
+                    var waitTime = (eachTestExpirationTime < DateTime.UtcNow) ? 0 : (int)(eachTestExpirationTime - DateTime.UtcNow).TotalMilliseconds;
```
varunpuranik: Use Actual type where the type is not obvious (not var)

varunpuranik: Also, do you need to convert to TotalMilliseconds? Why not use the TimeSpan?

philipktlin: Current implementation is safe since comparison and calculation are in 2 statements. 
```suggestion
                   TimeSpan waitTime = eachTestExpirationTime - DateTime.UtcNow;
                   if (waitTime.TotalMilliseconds > 0)
                   {
                       await Task.Delay(waitTime, cts.Token);
                   }
```

PR #2571, opened 2020-02-24
Add test info in LH and Stress reports

philipktlin:
Add test info into test reports for LH and stress test, which will send to Log analytics and used in test dashboard.


test/modules/TestAnalyzer/ModuleMessagesReport.cs
```diff
@@ -41,6 +42,10 @@ public ModuleMessagesReport(string moduleId, StatusCode statusCode, long receive
 
         public IList<MissedMessagesDetails> MissedMessages { get; }
 
+        public bool IsPassed => this.StatusCode == StatusCode.AllMessages;
```
and-rewsmith: Won't this always report as passing?

test/modules/TestAnalyzer/Settings.cs
```diff
@@ -41,7 +40,12 @@ class Settings
             this.WebhostPort = Preconditions.CheckNonWhiteSpace(webhostPort, nameof(webhostPort));
             this.ToleranceInMilliseconds = Preconditions.CheckRange(tolerance, 0);
             this.StoragePath = storagePath;
-            this.TestInfo = testInfo.Split(",", StringSplitOptions.RemoveEmptyEntries);
+            this.TestInfo = testInfo.Split(",", StringSplitOptions.RemoveEmptyEntries)
```
and-rewsmith: Beware that StringSplitOptions might not work as you expect.
https://stackoverflow.com/questions/10682301/stringsplitoptions-removeemptyentries-doesnt-work-as-advertised/10682341

```"      " by definition is not empty (it is actually whitespace), so it is not removed from resulting array.```

and-rewsmith:
I'm going to make the changes to upload to different log type.

test/modules/TestAnalyzer/Settings.cs
philipktlin: it should be fine since I ensure '=' should be there.

PR #2575, opened 2020-02-24
DevOpsLib: Ingest connectivity, longhaul, stress data into database for test dashboard

and-rewsmith:
We do not currently ingest vsts build api data for our connectivity, longhaul, or stress tests.


tools/IoTEdgeDevOps/DevOpsLib/BuildDefinitionId.cs
```diff
@@ -10,6 +10,10 @@ public enum BuildDefinitionId
         EdgeletRelease = 31845,
         EndToEndTest = 87020,
         ImageRelease = 31987,
-        LibiohsmCI = 39853
+        LibiohsmCI = 39853,
+        ConnectivityTest = 108949,
+        LonghaulTest = 66429,
```
philipktlin: how about long haul release and RC pipeline?

PR #2579, opened 2020-02-24
Use LogLevel to limit RocksDB LOG file size

yophilav:
Porting fixes in 
https://github.com/Azure/iotedge/pull/2552/files
to master branch. 

- Use Allow LogLevel to be configurable using an environment variable from a module
- Set default RocksDB logging to be None ( Create the minimal 15KB LOG file and leave it alone without logging any activity to it.)


philipktlin:
is this exactly the same as https://github.com/Azure/iotedge/pull/2552/files?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
```diff
@@ -452,19 +454,19 @@ static async Task CloseDbStoreProviderAsync(IContainer container)
...
-            ulong storageMaxTotalWalSize = 0;
+            T storageParamValue = default(T);
             try
             {
-                storageMaxTotalWalSize = configuration.GetValue<ulong>(Constants.StorageMaxTotalWalSize);
+                storageParamValue = configuration.GetValue<T>(fieldName);
             }
             catch
             {
                 // ignored
```
philipktlin: do we print out what storage configed value in log file?  and why you need to make it generic?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -262,22 +265,23 @@ static string GetProductInfo()
...
+            T storageParamValue = default(T);
             try
             {
-                storageMaxTotalWalSize = this.configuration.GetValue<ulong>(Constants.ConfigKey.StorageMaxTotalWalSize);
+                storageParamValue = configuration.GetValue<T>(fieldName);
             }
             catch
             {
-                // ignored
+                logger?.LogError($"Invalid parameter for {fieldName}");
```
philipktlin: If this field is not define, it is fine.  Therefore please update the message.
```suggestion
                logger?.LogError($"Can't get parameter '{fieldName}' from config.");
```

PR #2580, opened 2020-02-24
Add priority-aware methods to MessageStore interface

richma-ms:
This changes readies the MessageStore implementation for prioritized message queues.  There are some pieces of code that are in a transitional state as a result, but I left notes for them and they'll be resolved in the next PR where StoringAsyncEndpointExecutor actually starts to tie everything together.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -142,6 +160,25 @@ public void Dispose()
...
+                // have the "_Pri<x>" suffix. We use the default
+                // priority (2,000,000,000) for this, which means
+                // the store ID is just the endpoint ID.
+                return endpointId;
+            }
+            else
+            {
+                // The actual ID for the underlying store is of string format:
+                //      <endpointId>_Pri<priority>
+                return endpointId + "_Pri" + priority.ToString();
```
varunpuranik: $"{endpointId}_Pri{priority}"

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -55,31 +55,44 @@ public void SetTimeToLive(TimeSpan timeSpan)
             Events.TtlUpdated(timeSpan);
         }
 
-        public async Task AddEndpoint(string endpointId)
+        public Task AddEndpoint(string endpointId) => this.AddEndpointQueue(endpointId, RouteFactory.DefaultPriority);
+
+        public async Task AddEndpointQueue(string endpointId, uint priority)
```
varunpuranik: So, thinking about this a bit more - can we move this logic one level up? That way, the MessageStore just gets an endpoint, but now instead of it being "endpointId", it is "endpointId_Pri10". That way the MessageStore API doesn't need to change at all.

richma-ms: talked offline, the suggested approach will likely be more flexible for when we implement metered connection management, where the messagestore would need to expose even more parameters to set.

I'm abandoning this PR and will just have the executor calculate the queue name in the next PR.

richma-ms:
Abandoning this PR in favor of having the executor calculate the sequentialstore ID, so that the messagestore implementation doesn't need to change at all.

PR #2581, opened 2020-02-24
Improve staged deployments in end-to-end tests

damonbarry:
It turns out #2567 didn't go far enough. Deployments were still duplicated instead of staged because we weren't making a deep copy of the configuration for the 1st stage.

This change makes the following corrections:
- Makes a deep copy of the configuration for each stage, so that the 2nd one won't update the 1st
- Moved some of the configuration creation code into a factory method on `EdgeConfiguration` for readability
- Instead of separating deployments via a hard-coded 10-second delay, wait for the 1st deployment to complete before sending the next

### DIFF
--- a/test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
+++ b/test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
@@ -29,6 +29,45 @@ namespace Microsoft.Azure.Devices.Edge.Test.Common.Config
                 .ToArray();
         }
 
+        public static EdgeConfiguration Create(string deviceId, List<ModuleConfiguration> modules)
+        {
+            var config = new ConfigurationContent
+            {
+                ModulesContent = new Dictionary<string, IDictionary<string, object>>()
+            };
+
+            var moduleNames = new HashSet<string>();
+            var moduleImages = new HashSet<string>();
+
+            void UpdateConfiguration(ModuleConfiguration module)
+            {
+                moduleNames.Add(module.Name);
+                moduleImages.Add(module.Image);
+
+                if (module.DesiredProperties.Count != 0)
+                {
+                    config.ModulesContent[module.Name] = new Dictionary<string, object>
+                    {
+                        ["properties.desired"] = module.DesiredProperties
+                    };
+                }
+            }
+
+            foreach (ModuleConfiguration module in modules)
+            {
+                UpdateConfiguration(module);
+            }
+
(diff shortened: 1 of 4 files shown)


test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
```diff
@@ -29,6 +30,30 @@ public class EdgeConfiguration
...
+                ModulesContent = modules
+                    .Where(m => m.DesiredProperties.Count != 0)
+                    .Select(m => new KeyValuePair<string, IDictionary<string, object>>(
+                        m.Name,
+                        new Dictionary<string, object>
+                        {
+                            ["properties.desired"] = m.DesiredProperties
+                        }))
+                    .ToDictionary(x => x.Key, x => x.Value)
+            }); // serialize/deserialize to make a copy
```
and-rewsmith: nit: no need for key value pair

damonbarry: Yep, I'll simplify this as well

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
```diff
@@ -29,6 +30,30 @@ public class EdgeConfiguration
                 .ToArray();
         }
 
+        public static EdgeConfiguration Create(string deviceId, List<ModuleConfiguration> modules)
+        {
+            var names = modules.Select(m => new StringBuilder(m.Name).ToString()).ToArray();
+            var images = modules.Select(m => new StringBuilder(m.Image).ToString()).ToArray();
```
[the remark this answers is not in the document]
damonbarry: No. Since strings are immutable, this isn't really a concern...updated references will point to a new string, other references will still point to the old. I'll simplify this.

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
damonbarry: Done.

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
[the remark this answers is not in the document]
damonbarry: Fixed.

PR #2586, opened 2020-02-25
Import more data for test dashboard

philipktlin:
Import deployments and tasks from vsts release envrionments for test dashboard.


tools/IoTEdgeDevOps/DevOpsLibTest/AgentMatrixTest.cs
```diff
@@ -182,14 +182,20 @@ public void TestProperties()
                 {
                     new AgentCapability("run-e2e-tests", "true"),
                 });
+            var connectivityTestsds = new AgentDemandSet(
```
and-rewsmith: nit:typo

philipktlin: it is not typo; it means connectivity tests demand set.  :P

PR #2593, opened 2020-02-26
Update vsts batch import

philipktlin:
Update to import pipeline task for a new perfix; and fix date time type parameter when calling to stored procedures; and all more logs.


tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
```diff
@@ -253,11 +264,11 @@ void UpsertVstsReleaseDeploymentToDb(SqlConnection sqlConnection, int releaseEnv
             cmd.Parameters.Add(new SqlParameter("@ReleaseEnvironmentId", releaseEnvironmentId));
             cmd.Parameters.Add(new SqlParameter("@Attempt", deployment.Attempt));
             cmd.Parameters.Add(new SqlParameter("@Status", deployment.Status.ToString()));
-            cmd.Parameters.Add(new SqlParameter("@LastModifiedOn", deployment.LastModifiedOn));
+            cmd.Parameters.Add(new SqlParameter("@LastModifiedOn", SqlDbType.DateTime2) { Value = deployment.LastModifiedOn });
```
and-rewsmith: Any reason why you needed to change the type?

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
```diff
@@ -180,13 +182,22 @@ async Task ImportVstsReleasesDataAsync(ReleaseManagement releaseManagement, stri
                             {
                                 UpsertVstsReleaseDeploymentToDb(sqlConnection, releaseEnvironment.Id, deployment);
 
-                                foreach(IoTEdgePipelineTask pipelineTask in deployment.Tasks.Where(x => IsTestTask(x)))
+                                const string testTaskPrefix = "Test:";
```
and-rewsmith: My understanding is that you are adding a prefix to all test tasks in the test yaml then stripping the substring out when loading the tasks into the db. 

Do we have the changes in master yet to change the task names, or is it a hardcoded yaml in the pipeline?

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
philipktlin: Previously the type is wrong, it is the right type to use; check what is done for vsts build.

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
philipktlin: Old e2e tests is not in yaml; just in vsts pipeline. :(

PR #2595, opened 2020-02-26
Longhaul/Stress: Add test name when reporting to log analytics

and-rewsmith:
Add test name when reporting to log analytics. Also fix the naming of build definition Ids.


and-rewsmith:
This has been tested in longhaul

builds/e2e/longhaul.yaml
```diff
@@ -83,6 +83,7 @@ jobs:
       # Deploy long haul
       - template: templates/longhaul-deploy.yaml
         parameters:
+          testName: 'LongHaul'
```
yophilav: How is the `testName` different than a built-in-predefined variable `Build.DefinitionName` in this particular use case?
https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml

and-rewsmith: We want this name as LongHaul no matter what pipeline it is run in

PR #2597, opened 2020-02-26
Update the DirectMethod test in the Old E2E test to have a retry logic

yophilav:
Bug: 
- Currently the old E2E test framework is flaky because it only tries to send a direct method once. If the device is not ready to receive a direct method the test would just terminate instead of retry -- this is not a desired trait of the test. 

Fix: 
- This introduce a retry logic for a direct method so that it can gracefully handle the case when the device is not ready for a direct method.


smoke/LeafDevice/details/Details.cs
```diff
@@ -299,14 +299,33 @@ protected async Task VerifyDirectMethodAsync()
...
+                        Console.WriteLine($"Failed to send direct method from device '{this.context.Device.Id}' with payload '{cloudToDeviceMethod}'");
+
+                        if (DateTime.UtcNow > endTime)
+                        {
+                            isRetrying = false;
+                        }
+                    }
+                }
 
                 if (result.Status != 200)
```
philipktlin:
```suggestion
                if (result?.Status != 200)
```

smoke/LeafDevice/details/Details.cs
```diff
@@ -299,14 +299,33 @@ protected async Task VerifyDirectMethodAsync()
...
+                        result = await serviceClient.InvokeDeviceMethodAsync(
+                            this.context.Device.Id,
+                            cloudToDeviceMethod,
+                            cts.Token);
+                    }
+                    catch (Exception)
+                    {
+                        Console.WriteLine($"Failed to send direct method from device '{this.context.Device.Id}' with payload '{cloudToDeviceMethod}'");
+
+                        if (DateTime.UtcNow > endTime)
```
philipktlin:
```suggestion
                        if (DateTime.UtcNow >= endTime)
```

smoke/LeafDevice/details/Details.cs
```diff
@@ -299,14 +299,33 @@ protected async Task VerifyDirectMethodAsync()
...
+                    try
+                    {
+                        result = await serviceClient.InvokeDeviceMethodAsync(
+                            this.context.Device.Id,
+                            cloudToDeviceMethod,
+                            cts.Token);
+                    }
+                    catch (Exception)
+                    {
+                        Console.WriteLine($"Failed to send direct method from device '{this.context.Device.Id}' with payload '{cloudToDeviceMethod}'");
```
philipktlin:
```suggestion
                        Console.WriteLine($"Failed to send direct method from device '{this.context.Device.Id}' with payload '{cloudToDeviceMethod}: {ex}'");
```

smoke/LeafDevice/details/Details.cs
```diff
@@ -299,14 +299,33 @@ protected async Task VerifyDirectMethodAsync()
...
+
+                        if (DateTime.UtcNow > endTime)
+                        {
+                            isRetrying = false;
```
philipktlin: log full exception message

PR #2604, opened 2020-02-27
Update connectivity yamls

philipktlin:
stop running tasks if any previous task is failed in connectivity test.

### DIFF
--- a/builds/e2e/connectivity.yaml
+++ b/builds/e2e/connectivity.yaml
@@ -88,7 +88,7 @@ jobs:
               echo "run.steps=0"
             fi
       - checkout: none
-        condition: eq(variables['run.steps'], 1)
+        condition: and(succeeded(), eq(variables['run.steps'], 1))
       - task: AzureKeyVault@1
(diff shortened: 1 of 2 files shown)


builds/e2e/connectivity.yaml
```diff
@@ -88,7 +88,7 @@ jobs:
               echo "run.steps=0"
             fi
       - checkout: none
-        condition: eq(variables['run.steps'], 1)
+        condition: and(succeeded(), eq(variables['run.steps'], 1))
```
yophilav: Per https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops,
you may want to be explicated that you want a strictly `Succeeded` status? 

By default, `succeeded()` is a combination of `Succeeded` and `SucceededWithIssues` (continued w/ error)

philipktlin: As talked, succeeded() will ensure all previous steps are passed.  This is exactly what we want.

PR #2613, opened 2020-02-28
fix connectivity yaml for arm32

philipktlin:
fix connectivity yaml for arm32


builds/e2e/connectivity.yaml
```diff
@@ -177,14 +209,30 @@ jobs:
...
+        condition: and(succeeded(), eq(variables['run.steps'], 1))
       - task: AzureKeyVault@1
+        condition: and(succeeded(), eq(variables['run.steps'], 1))
         displayName: 'Azure Key Vault'
         inputs:
           azureSubscription: $(azure.subscription)
           KeyVaultName: $(azure.keyVault)
           SecretsFilter: 'edgebuilds-azurecr-io-username,edgebuilds-azurecr-io-pwd,EdgeConnectivityTestHubConnString,EdgeConnectivityEventHubConnString,kvLogAnalyticWorkspaceId,kvLogAnalyticSharedKey,EdgeConnectivityStorageAccountConnString'
       - task: DownloadBuildArtifacts@0
+        condition: and(succeeded(), eq(variables['run.steps'], 1))
```
yophilav: By default, `succeeded()` is a combination of `Succeeded` and `SucceededWithIssues` (continued w/ error)

Are you sure that is what you want there?

builds/e2e/connectivity.yaml
```diff
@@ -195,6 +243,7 @@ jobs:
           downloadPath: '$(Build.StagingDirectory)'
           artifactName: $(edgelet.artifact.name)
       - task: DownloadBuildArtifacts@0
+        condition: and(succeeded(), eq(variables['run.steps'], 1))
```
yophilav: Where did `run.steps` get set? I didn't see it in the built-in DevOps document.

yophilav: Please rename the var.

builds/e2e/templates/set-run-flag.yaml
```diff
@@ -0,0 +1,17 @@
+steps:
+  - task: Bash@3
+    name: Set_run_flag
```
yophilav: Please add more description that this block of code is for job cancellation.

builds/e2e/connectivity.yaml
philipktlin: this is exactly what we want as succeeded() will ensure all previous tasks succeeded.

PR #2623, opened 2020-03-02
[mqtt] Import mqtt-broker project

dmolokanov:
Import mqtt-broker project from its own repo
1. location: ./mqtt
2. disable dotnet checks for mqtt-broker
3. import mqtt client project
4. make it compile on windows


mqtt-broker/mqtt-broker/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+futures-util = { version = "0.3", features = ["sink"] }
+lazy_static = "1"
+tokio = { version = "0.2", features = ["macros", "rt-threaded", "signal", "stream", "sync", "tcp"] }
+tokio-io-timeout = "0.4"
+tokio-util = { version = "0.2", features = ["codec"] }
+tracing = "0.1"
+tracing-futures = "0.2"
+uuid = { version = "0.8", features = ["v4"] }
+
+mqtt = { git = "https://github.com/myagley/mqtt", branch = "v0.2.x" }
```
arsing: You should import this too.

dmolokanov: Like, put it next to the broker code locally?

varunpuranik:
Should we instead create a ./mqtt top level folder, which can then contain the broker, client, bridge, etc?

mqtt-broker/mqtt-broker/Cargo.toml
dmolokanov: done

mqtt/mqtt3-fuzz/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
```
arsing: Ah, if you're importing this too, then I have a couple of scripts I use locally to run `afl` that we should add here as a README or something. Remind me when I come in tomorrow.

mqtt/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# -*- mode: dockerfile -*-
+#
+# An example Dockerfile showing how to build a Rust executable using this
+# image, and deploy it with a tiny Alpine Linux container.
+
+# You can override this `--build-arg BASE_IMAGE=...` to use different
+# version of Rust or OpenSSL.
+ARG BASE_IMAGE=ekidd/rust-musl-builder:beta
```
arsing: It needs `beta` ? `stable` doesn't work?

.gitignore
```diff
@@ -322,3 +322,7 @@ stylecop.json
 *.swp
 *.swo
 *.vi
+
+# fuzz
+in
+out
```
arsing: Missing newline at end-of-file.

mqtt/mqtt3/src/proto/packet.rs
```diff
@@ -159,16 +163,8 @@ impl PacketMeta for Connect {
         let protocol_name = super::Utf8StringDecoder::default()
             .decode(&mut src)?
             .ok_or(super::DecodeError::IncompletePacket)?;
-        if protocol_name != "MQTT" {
```
arsing: As discussed, you should add this and the protocol level check back.

mqtt/mqtt3/examples/publisher.rs
```diff
@@ -103,11 +103,11 @@ fn main() {
         .shutdown_handle()
         .expect("couldn't get shutdown handle");
     runtime.spawn(async move {
-        let () = tokio::signal::ctrl_c()
```
arsing: :(

mqtt/mqtt3/tests/common/mod.rs
```diff
@@ -100,14 +101,15 @@ impl mqtt3::IoSource for IoSource {
...
-                let () = tokio::time::delay_for(std::time::Duration::from_secs(5)).await;
+                tokio::time::delay_for(std::time::Duration::from_secs(5)).await;
                 unreachable!();
             })
         }
     }
 }
 
 /// A single connection between a client and a server
+#[allow(clippy::type_complexity)]
```
arsing: As discussed, you should be able to move this to the specific field that it complained about instead of applying it to the whole struct.

mqtt/mqtt3/README.md
```diff
@@ -1,28 +0,0 @@
-An MQTT v3.1.1 Client implementation in Rust
```
arsing: As discussed, this README is specific to the crate, so you shouldn't delete it.

mqtt/mqtt3/src/proto/packet.rs
dmolokanov: True, but Mike build logic around these fields. Broker decides to disconnect a client, but not the connection itself. We can change this is the future. I decided to leave it like this for now.

mqtt/mqtt3/examples/publisher.rs
dmolokanov: Clippy was super opinionated about this. But... I agree with you :)

mqtt/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# RUN sudo chown -R rust:rust /home/rust
+
+# Build our application.
+RUN cargo build -p mqttd --release && strip /home/rust/src/target/x86_64-unknown-linux-musl/release/mqttd
+
+FROM scratch
+ENV RUST_LOG=info
+EXPOSE 1883/tcp
+EXPOSE 8883/tcp
+
```
lfitchett: Do we want to run as non-root like edgeHub? This is eventually going to move to hubs container, so its probably best to make sure it can run as non-root now.

mqtt/Dockerfile
myagley: ah good catch. It was using beta because I started on this before async/await was stabilized.

mqtt/Dockerfile
dmolokanov: It's just a stub to be able to compile it without installing the whole toolchain. Should not be used for building a final container.

mqtt/mqtt3/src/proto/packet.rs
myagley: We can add these checks back in. We will also need to add the protocol_name and protocol_level to the connect packet.

mqtt/mqtt3-fuzz/Cargo.toml
dmolokanov: I'm going to add it as a separate PR

PR #2625, opened 2020-03-02
Fix libiothsm-std centos7-amd64 and centos7-aarch64 packages to install to /usr/lib64.

arsing:
c1e2d5e93bf08d5b2ea7e52ab2bd8957652a7e88 defined CMAKE_SYSTEM_NAME and
CMAKE_SYSTEM_VERSION for all targets, however this breaks
GNUInstallDirs.cmake's detection of the system architecture, so it doesn't
realize CentOS amd64's libdir is /usr/lib64. Instead it defaults to /usr/lib,
which causes the libiothsm-std package to install libiothsm.so to /usr/lib and
causes iotedged to fail to find it.

Before the package.sh refactoring, `CMAKE_SYSTEM_NAME` and
`CMAKE_SYSTEM_VERSION` were only defined for arm32 and aarch64 targets,
via 0dac12c17f03261c2a2979d09457327012e44f91. So this commit restores that
behavior.

For CentOS specifically, we also don't want to define `CMAKE_SYSTEM_NAME` and
`CMAKE_SYSTEM_VERSION` for the arm32v7 and aarch64 builds anyway,
because they're only supposed to be defined when cross-compiling, whereas
the CentOS arm32v7 and aarch64 builds run natively (via qemu).

### DIFF
--- a/edgelet/build/linux/package.sh
+++ b/edgelet/build/linux/package.sh
@@ -13,7 +13,7 @@ REVISION="${REVISION:-1}"
 DEFAULT_VERSION="$(cat "$PROJECT_ROOT/version.txt")"
 VERSION="${VERSION:-$DEFAULT_VERSION}"
 
-CMAKE_ARGS='-DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_VERSION=1 -DCMAKE_BUILD_TYPE=Release'
+CMAKE_ARGS='-DCMAKE_BUILD_TYPE=Release'
 CMAKE_ARGS="$CMAKE_ARGS -DBUILD_SHARED=On -Drun_unittests=Off -Duse_default_uuid=On -Duse_emulator=Off -Duse_http=Off"
 
 DOCKER_VOLUME_MOUNTS=''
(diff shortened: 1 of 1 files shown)


arsing:
CentOS ARM32 build failed because it looks like it doesn't have epel. I think I'll revert the `cmake3` part of the change.

edgelet/build/linux/package.sh
```diff
@@ -183,13 +183,15 @@ case "$PACKAGE_ARCH" in
         ;;
 
     'arm32v7')
+        CMAKE_ARGS="$CMAKE_ARGS -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_VERSION=1"
```
darobs: Should we set `CMAKE_SYSTEM_VERSION` to something more accurate. i.e the target system version?

arsing: I'll go through the `.cmake`s again and see what exactly cmake was looking for in these too to determine the libdir. Technically even before the `package.sh` refactoring we were defining `Linux/1` (see Mahesh's commit linked in the commit message), so it shouldn't be worse than what it was back then. Then again, the arm32 and aarch64 packages before the refactoring were broken because of the malformed openssl dependency anyway, so they clearly had never been tested and `Linux/1` may never have been right for CentOS at all.

I also need to check what the libdir is on centos aarch64; if it's supposed to be `/usr/lib64` then `Linux/1` will probably break it too.

arsing: Yeah, CentOS aarch64 does use `/usr/lib64`, and `linux/1` does do the wrong thing for it. So we do need to figure out what exactly is needed to make it work.

arsing: I have been enlightened.

`CMakeDetermineSystem.cmake`

>```
># if CMAKE_SYSTEM_NAME is here already set, either it comes from a toolchain file
># or it was set via -DCMAKE_SYSTEM_NAME=...
># if that's the case, assume we are crosscompiling
>if(CMAKE_SYSTEM_NAME)
>  if(NOT DEFINED CMAKE_CROSSCOMPILING)
>    set(CMAKE_CROSSCOMPILING TRUE)
>  endif()
>else()
>  set(CMAKE_SYSTEM_NAME      "${CMAKE_HOST_SYSTEM_NAME}")
>  set(CMAKE_SYSTEM_VERSION   "${CMAKE_HOST_SYSTEM_VERSION}")
>  set(CMAKE_SYSTEM_PROCESSOR "${CMAKE_HOST_SYSTEM_PROCESSOR}")
>  set(CMAKE_CROSSCOMPILING FALSE)
>endif()
>```

So it should also not be defined for the CentOS arm32 and aarch64 targets since those are being compiled natively. I confirmed that an aarch64 build on my machine with the two not set ends up using `/usr/lib64` correctly.

---

Also, re: the original question of what value to use for `CMAKE_SYSTEM_VERSION`:

>```
># CMAKE_SYSTEM_VERSION - on unix this is uname -r, for windows it is empty
>```

So this is actually the kernel version. I suppose we could set it to the output of `uname -r` ourselves, but for the stuff we're building the kernel version doesn't really make a difference. I'm leaning towards just leaving it as `1`

PR #2626, opened 2020-03-03
Using unique id for stress and longhaul

dylanbronson:
We currently reuse the same deviceId for each run of stress and longHaul. 

We recently ran into a problem in LH, where that old device was in a state that did not allow LongHaul TwinTester to start up (it had too many reported properties associated with it, and it's updates were too large).

We should be using unique Ids for each run. This PR achieves that. 

We've logged a bug for the problem that got the old device into that unfixable state, as well.


scripts/linux/runE2ETest.sh
```diff
@@ -687,7 +687,8 @@ function run_longhaul_test() {
     print_highlighted_message "Run Long Haul test for $image_architecture_label"
     test_setup
 
-    local device_id="$RELEASE_LABEL-Linux-$image_architecture_label-longhaul"
+    local random_id=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 8)
```
philipktlin: wrap this line in a method; same for powershell script as well.

and-rewsmith:
Dylan and I synced and decided that before we merge this, we need to make sure we change the metrics dashboard to not rely upon this static device id.

scripts/linux/runE2ETest.sh
```diff
@@ -468,6 +468,11 @@ function process_args() {
     echo 'Required parameters are provided'
 }
 
+function get_random_id()
```
philipktlin: Currently you can find that I create another file named testHelper.sh for connectivity case, the intent is to break down tests into its own file.  What does it mean? let's take runE2ETest.sh as an example, we started this script from first tempSensor test.  And ppls add more and more tests to it.  However it is getting mess due to different tests need different input parameters; which means runE2ETest.sh contains the union of all the input parameters for all supported tests.  To solve this issue, we should break down our implementation into its own test files and share some common logic/functions in a shared file (which is testHelper.sh).  Think about even input parameter method (ProcessArgs), we should start sharing by breaking all common parameters in one function as well.

why I explain this?  I think it is good to start with the first step (the easiest one), which is to extract those reusable function in a testHelper.sh file like get_random_id().  Please help to move at least this function to its own file and make sure it is specificed in download artifact step in e2e, LH and stress.

What do you think?

philipktlin: Currently you can find that I create another file named testHelper.sh for connectivity case, the intent is to break down tests into its own file.  What does it mean? let's take runE2ETest.sh as an example, we started this script from first tempSensor test.  And ppls add more and more tests to it.  However it is getting mess due to different tests need different input parameters; which means runE2ETest.sh contains the union of all the input parameters for all supported tests.  To solve this issue, we should break down our implementation into its own test files and share some common logic/functions in a shared file (which is testHelper.sh).  Think about even input parameter method (ProcessArgs), we should start sharing by breaking all common parameters in one function as well.

why I explain this?  I think it is good to start with the first step (the easiest one), which is to extract those reusable function in a testHelper.sh file like get_random_id().  Please help to move at least this function to its own file and make sure it is specificed in download artifact step in e2e, LH and stress.

What do you think?

PR #2628, opened 2020-03-03
Fixed the TwinTester to use a proper key param casing

yophilav:
- testStartDelay
- testDuration
- testResultCoordinatorUrl


test/modules/TwinTester/config/settings.json
```diff
@@ -9,5 +9,5 @@
     "StoragePath": "",
     "StorageOptimizeForPerformance": true,
     "TwinTestMode": "TwinAllOperations",
-    "TestStartDelay":  "00:02:00"
+    "testStartDelay":  "00:02:00"
```
philipktlin: We should not define testStartDelay in settings.json as default value; for LH and stress test, I think test start delay should be TimeSpan.Zero.  Please confirm with @and-rewsmith and remove it if my understanding is correct.  thanks.

yophilav: I'm making this default value to 2 min and remove the config from the setting file.

PR #2631, opened 2020-03-03
Fixing twin tests for connectivity

dylanbronson:
We are seeing a bug in Connectivity Twin Tests. 

When we start the tests, we do a cleanup - where we grab the current twin, set its properties to null, and send an UpdateReportedProperties call. Due to how the SDK works, this will automatically subscribe us for receiving DesiredPropertiesUpdates (DPU). But then we wait for 2 minutes due to the testStartDelay. After that, we set the callback for receiving DPUs.

The issue is that if we receive a DPU during the testStartDelay, the callback won't be set, and we will do nothing. 

So this PR moves the callback up before the testStartDelay, so that we are ready to receive even during the waiting period. 

It also removes the cleanup because it is unnecessary now that we are using a fresh device every time.

### DIFF
--- a/test/modules/TwinTester/TwinEdgeOperationsInitializer.cs
+++ b/test/modules/TwinTester/TwinEdgeOperationsInitializer.cs
@@ -35,8 +35,8 @@ namespace TwinTester
 
         public async Task StartAsync(CancellationToken cancellationToken)
         {
-            Logger.LogInformation($"Waiting for {Settings.Current.TestStartDelay} based on TestStartDelay setting before starting.");
             await this.desiredPropertiesReceiver.UpdateAsync();
+            Logger.LogInformation($"Waiting for {Settings.Current.TestStartDelay} based on TestStartDelay setting before starting.");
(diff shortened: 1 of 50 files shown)


test/modules/TwinTester/TwinEdgeOperationsInitializer.cs
```diff
@@ -28,29 +28,18 @@ class TwinEdgeOperationsInitializer : ITwinTestInitializer
...
-            catch (Exception e)
-            {
-                throw new Exception($"Shutting down module. Initialization failure: {e}");
-            }
+            return Task.FromResult(new TwinEdgeOperationsInitializer(registryManager, moduleClient, reporter, 0));
         }
 
         public async Task StartAsync(CancellationToken cancellationToken)
         {
             Logger.LogInformation($"Waiting for {Settings.Current.TestStartDelay} based on TestStartDelay setting before starting.");
```
ancaantochi: Please move the log line just before the delay

PR #2635, opened 2020-03-03
Synchronize deployments in end-to-end tests

damonbarry:
This change improves the way the end-to-end tests wait for a configuration to be deployed to the edge. Previously the tests simply looked for the expected module names to match the set of currently running docker containers. But with this change, the tests also wait until Edge Agent's reported properties match the expect config, meaning that Edge Agent has processed the entire deployment and believes it is complete.

When a single test deploys multiple configurations (esp. with only minor differences between deployments), our wait logic didn't realize that the current set of running containers was from the previous deployment rather than the current deployment, so the test would continue on before the deployment was complete. These changes fix that problem.

Details:
- Add method `EdgeAgent.WaitForReportedConfigurationAsync`, which waits until Edge Agent's reported properties match the given object
- Add method `EdgeConfiguration.VerifyAsync`, which calls `EdgeAgent.WaitForReportedConfigurationAsync`, effectively waiting until edge agent reports to the cloud that the configuration has been deployed
- Add `expectedConfig` argument to `EdgeConfiguration` constructor, and generate the new argument in the factory method `EdgeConfiguration.Create`
- Move logic from `EdgeModule.WaitForReportedPropertyUpdatesAsync` into `WaitForReportedPropertyUpdatesInternalAsync` so `EdgeAgent.WaitForReportedConfigurationAsync` can call it with different logging
- Call `EdgeConfiguration.VerifyAsync` from `EdgeRuntime.DeployConfigurationAsync` to wait until each configuration is actually deployed before continuing
- Save Edge Agent's module name internally as "$edgeAgent" instead of "edgeAgent"


builds/e2e/e2e.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       arch: arm32v7
       artifactDir: iotedged-debian9-arm32v7
 
-    timeoutInMinutes: 90
+    timeoutInMinutes: 120
```
and-rewsmith: Does this verification add ~30 mins?

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
```diff
@@ -130,31 +130,71 @@ public Task UpdateDesiredPropertiesAsync(object patch, CancellationToken token)
                 this.Id);
         }
 
-        public Task WaitForReportedPropertyUpdatesAsync(object expectedPatch, CancellationToken token)
```
and-rewsmith: nit: expected could be more descriptive

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
```diff
@@ -130,31 +130,71 @@ public Task UpdateDesiredPropertiesAsync(object patch, CancellationToken token)
...
+        // Newtonsoft.Json.Linq.JToken.SelectToken(path)
+        static bool JsonEquals(
+            (object obj, string rootPath) reference,
+            (object obj, string rootPath) comparand)
+        {
+            Dictionary<string, JValue> ProcessJson(object obj, string rootPath)
+            {
+                // return all json values under root path, with their relative
+                // paths as keys
+                int length = rootPath.Length + (rootPath == string.Empty ? 0 : 1);
```
and-rewsmith: Assuming there is a root path with length > 1, won't the substring call below work fine without adding an extra 1?

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
```diff
@@ -130,31 +130,71 @@ public Task UpdateDesiredPropertiesAsync(object patch, CancellationToken token)
...
+                    .FromObject(obj)
+                    .SelectToken(rootPath)
+                    .Cast<JContainer>()
+                    .DescendantsAndSelf()
+                    .OfType<JValue>()
+                    .ToDictionary(v => v.Path.Substring(length));
+            }
+
+            var referenceValues = ProcessJson(reference.obj, reference.rootPath);
+            var comparandValues = ProcessJson(comparand.obj, comparand.rootPath);
```
and-rewsmith: Recommend using explicit types instead of var

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
```diff
@@ -130,31 +130,71 @@ public Task UpdateDesiredPropertiesAsync(object patch, CancellationToken token)
...
+
+            // comparand equals reference if, for each json value in reference:
+            // - comparand has a json value with the same path
+            // - the json values match
```
and-rewsmith: Based on this logic, we don't care if comparand has extra json values. Is this fine?

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
```diff
@@ -130,31 +130,71 @@ public Task UpdateDesiredPropertiesAsync(object patch, CancellationToken token)
...
+
+            if (!match)
+            {
+                var missing = referenceValues
+                    .Where(kvp =>
+                        !comparandValues.ContainsKey(kvp.Key) ||
+                        !kvp.Value.Equals(comparandValues[kvp.Key]))
+                    .Select(kvp => kvp.Key);
+                Log.Verbose(
+                    "Expected configuration values missing in agent's reported properties:\n  {MissingValues}",
```
and-rewsmith: Not just missing, but not matching also, right? If so I would recommend changing the var name above as well.

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
```diff
@@ -48,7 +54,60 @@ public static EdgeConfiguration Create(string deviceId, IEnumerable<ModuleConfig
...
+                    reported = reported
+                }
+            };
+
+            return new EdgeConfiguration(deviceId, names, images, config, expected);
+        }
+
+        static object CreateExpectedModuleConfig(JObject source)
+        {
+            var image = source.SelectToken($"settings.image"); // not optional
```
and-rewsmith: If this doesn't exist in the source, will there be an exception?

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
```diff
@@ -34,7 +34,7 @@ public EdgeModule(string id, string deviceId, IotHub iotHub)
 
         public static Task WaitForStatusAsync(IEnumerable<EdgeModule> modules, EdgeModuleStatus desired, CancellationToken token)
         {
-            string[] moduleIds = modules.Select(m => m.Id).Distinct().ToArray();
+            string[] moduleIds = modules.Select(m => m.Id.TrimStart('$')).Distinct().ToArray();
```
and-rewsmith: Isn't it better to store it with the identifier?

builds/e2e/e2e.yaml
damonbarry: This extends the maximum time the tests can take on Pi. These changes don't really slow the tests down that much but in the worst case, where several tests fail, the Pi can easily go over the 90 min threshold. The agent can be left in a rather weird state if the test run times out--our cleanup code won't run for example--so absent some more complicated work on our part it's easiest to just prevent the timeout from happening.

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
damonbarry: Given the name of the method, it seems clear to me that you'd call it when you want to wait until the reported properties on this module match the `expected` object you passed in.

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
damonbarry: There's a dot separator between path segments, so doing:
```
"this.is.my.path".Substring("this.is".Length)
```
yields `".my.path"` and we want to knock off that leading `.`.

I could make it more clear by doing `v.Path.Substring(rootPath.Length).TrimStart('.')`. What do you think?

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
damonbarry: Oh yeah meant to go back and do that. I'll make a pass.

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
damonbarry: Yes it's fine. Comparand needs to have _at least_ everything in reference to match. We don't care about extra stuff.

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
damonbarry: I was thinking about it in the strictest sense: If there's a similar-but-not-exactly-the-same property in the object (e.g., the name is the same but the value is off by one character), then that's no more helpful than if the property wasn't there at all. The expected value is present (exactly as specified) or it's missing.

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
damonbarry: I believe so, either at this line, or later on when we try to dereference null. But that's good; the exception will bubble up and the test will fail.

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
damonbarry: When we're talking about a system module (starts with `$`), most of the time we need to preserve that leading dollar sign. There's only one case where we need to drop it; that's when we're matching up module names with docker container names. But in `EdgeModule` previously we were storing the module name _without_ the dollar sign, then adding it back when we needed it (which was most of the time). I just reversed the behavior, but the net effect is the same.

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
damonbarry: Updated.

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
damonbarry: Done.

PR #2637, opened 2020-03-03
Backport CentOS fixes to 1.0.9

### LINKED ISSUE #2398: How to install IoTEdge for centos7 ARM on raspbian Pi

arsing:
- 20b5f1fbd4aa350f2645d3b717e6ef51908feb8a

    Update edgelet devguide. (#2428)
    
    - Talk about building the packages as an alternative to building the binaries.
    
      Discussed in https://github.com/Azure/iotedge/pull/2301
    
    - Building binaries on Windows requires VS / VSBT and cmake.
    
    - Both `OPENSSL_DIR` and `OPENSSL_ROOT_DIR` need to be set for
      macOS and Windows.
    
    - Take into account that builds don't necessarily use "stable", but whatever
      happens to be in the rust-toolchain file.
    
    - `cargo build --all` spends time building Kubernetes dependencies.
      `cargo build -p iotedged -p iotedge` is sufficient.
    
    - Talk about how to actually run the binaries.
    
    - Update some URLs, such as from the merger of the rust-lang-nursery and
      rust-lang GitHub orgs.

- 958e4b05622a429bed08547f6e7915bc4a25fe0e

    Fix CentOS arm32v7 and aarch64 packages to depend on correct openssl sonames. (#2518)
    
    Before this change, these two targets were built using Linaro compilers
    and against custom-compiled openssl. However the packages built this way
    ended up depending on `libcrypto.so.1.0.0` and `libssl.so.1.0.0`,
    whereas CentOS's openssl-libs package provides `libcrypto.so.10` and
    `libssl.so.10`
    
    Since CentOS 7 does not ship cross-compiler packages and openssl,
    this commit changes the build steps to run the native arm32v7 and arm64v8
    CentOS Docker containers in qemu instead of cross-compiling.
    
    Since qemu builds are very slow (2h30m instead of 15m), this commit moves them
    to a new pipeline job so that it does not block existing downstream jobs like
    E2E tests that don't need the CentOS packages.
    
    Lastly, the new job runs on Ubuntu 18.04 hosts instead of Ubuntu 16.04 hosts.
    Ubuntu 16.04 supports getrandom, but its qemu is just old enough to
    not implement it, which breaks processes inside the container that require it.
    
    Fixes #2398

- fd2dc5a670fc05d2fbc111fd8a3256251869bfe5

    Fix libiothsm-std centos7-amd64 and centos7-aarch64 packages to install to /usr/lib64. (#2625)
    
    c1e2d5e93bf08d5b2ea7e52ab2bd8957652a7e88 defined CMAKE_SYSTEM_NAME and
    CMAKE_SYSTEM_VERSION for all targets, however this breaks
    GNUInstallDirs.cmake's detection of the system architecture, so it doesn't
    realize CentOS amd64's libdir is /usr/lib64. Instead it defaults to /usr/lib,
    which causes the libiothsm-std package to install libiothsm.so to /usr/lib and
    causes iotedged to fail to find it.
    
    Before the package.sh refactoring, `CMAKE_SYSTEM_NAME` and
    `CMAKE_SYSTEM_VERSION` were only defined for arm32 and aarch64 targets,
    via 0dac12c17f03261c2a2979d09457327012e44f91. So this commit restores that
    behavior.
    
    For CentOS specifically, we also don't want to define `CMAKE_SYSTEM_NAME` and
    `CMAKE_SYSTEM_VERSION` for the arm32v7 and aarch64 builds anyway,
    because they're only supposed to be defined when cross-compiling, whereas
    the CentOS arm32v7 and aarch64 builds run natively (via qemu).

### DIFF
--- a/builds/misc/packages.slow.yaml
+++ b/builds/misc/packages.slow.yaml
@@ -1,0 +1,71 @@
+# This is like packages.yaml, but for packages that are much slower to build.
+# This runs independently of the packages.yaml job so that it does not impact other jobs that use the packages.yaml artifacts, like E2E test runs.
+
+trigger:
+  batch: true
+  branches:
+    include:
+      - master
+pr: none
+variables:
+  REVISION: '1'
+jobs:
+
+################################################################################
+  - job: linux
+################################################################################
+    displayName: Linux
+    timeoutInMinutes: 180 # Observed to take just under 2h30m, so round up
+    pool:
+      # qemu on ubuntu-16.04 is too old to support some syscalls used by CentOS 7 containers, so ensure this is ubuntu-18.04 or newer.
+      vmImage: 'ubuntu-18.04'
+    strategy:
+      matrix:
+        Centos75-arm32v7:
+          arch: arm32v7
+          os: centos7
+          target.iotedged: edgelet/target/rpmbuild/RPMS/armv7hl
+        Centos75-aarch64:
+          arch: aarch64
+          os: centos7
+          target.iotedged: edgelet/target/rpmbuild/RPMS/aarch64
+    steps:
+      # Ensure any changes here are replicated in packages.yaml
+
+      - bash: |
+          BASE_VERSION=`cat $BUILD_SOURCESDIRECTORY/edgelet/version.txt`
+          VERSION="$BASE_VERSION$BUILD_BUILDNUMBER"
+          echo "##vso[task.setvariable variable=VERSION;]$VERSION"
+
+          echo "##vso[task.setvariable variable=PACKAGE_ARCH;]$(arch)"
+          echo "##vso[task.setvariable variable=PACKAGE_OS;]$(os)"
+        displayName: Set Version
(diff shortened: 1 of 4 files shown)


arsing:
(The devguide commit is there because the openssl sonames commit makes changes on top of it, and rather than resolve the merge conflict I figured it would be useful to c-p the devguide commit on its own worth.)

PR #2640, opened 2020-03-04
Test Dashboard: Update vsts pull job to read from keyvault using managed service identities

and-rewsmith:
Instead of relying solely on the args for the credentials, it is better if we can use RBAC to pull keys from keyvault from our Azure Container Instance.

I enabled this functionality, while also allowing the old way of passing args. I changed the argument order though, so it's not entirely backwards compatible.


tools/IoTEdgeDevOps/VstsPipelineSync/Program.cs
```diff
@@ -2,37 +2,64 @@ namespace VstsPipelineSync
...
+                Console.WriteLine(" vsts-pat: personal access token to vsts");
+                Console.WriteLine(" db-connection-string: connection string found in the azure portal");
                 Environment.Exit(1);
             }
 
-            var branches = new HashSet<string> { "refs/heads/master" };
+            HashSet<string> branches = new HashSet<string>(args[0].Split(","));
+            TimeSpan waitPeriodBeforeNextUpdate = TimeSpan.Parse(args[1]);
+            string pat;
+            string dbConnectionString;
```
philipktlin: nit: line break

tools/IoTEdgeDevOps/VstsPipelineSync/Program.cs
```diff
@@ -2,37 +2,64 @@ namespace VstsPipelineSync
...
             if (args.Length == 4)
             {
-                branches.UnionWith(args[3].Split(","));
+                pat = args[2];
```
philipktlin: suggestion: return here directly.

and-rewsmith: I feel like it is cleaner to return once at the end and have a more readable return statement.

PR #2645, opened 2020-03-04
Connectivity Tests: Add test description to report

and-rewsmith:
The test dashboard needs easily readable test names to display connectivity test results. I have added these test name identifiers under a field, "test description" for every report. Here are the test descriptions I intend to support:
```
messages | local | amqp
messages | upstream | amqp
messages | local | mqtt
messages | upstream | mqtt
twin | desired property | amqp
twin | reported property | amqp
twin | desired property | mqtt
twin | reported property | mqtt
deployment
direct method | cloud | amqp
direct method | cloud | mqtt
edge agent ping
C2D | amqp
C2D | mqtt
network controller
test configuration
test info
``` 

There are two similar fields already existing, but using / changing these fields is not ideal:
- **Result Type** (currently implemented as enum): Specifies the high level test type (e.g. messaging, twins, etc.). This identifier is too vague to fit my needs. I would like the freedom to differentiate within these enum types.
- **Title**: Describes the test in more detail, describing information potentially useful when debugging (e.g. ```Counting Report between [loadGen1.send] and [relayer1.receive] (Messages)```).  Changing this is not ideal because we need this detailed information when things go wrong.


and-rewsmith:
Still testing. Should have marked as draft in retrospect.

and-rewsmith:
Test run complete:
https://msazure.visualstudio.com/One/_build/results?buildId=29257914&view=results

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -649,9 +649,6 @@
               "testStartDelay": {
                 "value": "<TestStartDelay>"
               },
-              "transportType": {
```
philipktlin: why removed? and btw I found that this key casing is not matched.  In C2DMessageTester settings.cs, it started with capital T.  Please help to create a bug and let Dylan aware of it.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
```diff
@@ -20,7 +21,7 @@ class EdgeHubRestartDirectMethodReport : TestResultReportBase
             ulong senderCount,
             ulong receiverCount,
             TimeSpan medianPeriod)
-            : base(trackingId, resultType)
+            : base(testDescription, trackingId, resultType)
```
philipktlin: why do you prefer to put test description as 1st argument, not 3rd one?

test/modules/TestResultCoordinator/Reports/ErrorReportMetadata.cs
```diff
@@ -2,13 +2,17 @@
...
 {
     using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Util;
 
     class ErrorReportMetadata : ITestReportMetadata
     {
-        public ErrorReportMetadata()
+        public ErrorReportMetadata(string testDescription)
         {
+            this.TestDescription = Preconditions.CheckNonWhiteSpace(testDescription, nameof(testDescription));
```
philipktlin: Since you are adding code commonly used for all reportMetadata, I think it is better to create an abstract class and implements constructor, TestDescription property and ToString() method.

test/modules/TestResultCoordinator/Reports/TestResultReportBase.cs
```diff
@@ -8,12 +8,15 @@ namespace TestResultCoordinator.Reports
     /// </summary>
     abstract class TestResultReportBase : ITestResultReport
     {
-        protected TestResultReportBase(string trackingId, string resultType)
+        protected TestResultReportBase(string trackingId, string resultType, string testDescription)
```
philipktlin: why put testDescription in 3rd argument, but not 1st?  Please think about consistently between other test reports and TestResultReportBase.

e2e_deployment_files/connectivity_deployment.template.json
and-rewsmith: I noticed that transportType is not used by a c2dMessageSender in send mode. I talked with Dylan to confirm and we decided to remove it.

test/modules/TestResultCoordinator/Reports/ErrorReportMetadata.cs
and-rewsmith: Good suggestion. Will fix.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
and-rewsmith: My logic at the time was that it wasn't worth disturbing the ordering for the metadata, but I felt for the Reports themselves that most useful descriptor for the report should come first. I made this distinction because the report is the presentation, however the report doesn't get serialized in order anyway so this is not sound logic. I should change the reportMetadata for consistency.

test/modules/TestResultCoordinator/Reports/TestResultReportBase.cs
and-rewsmith: Oops. Will fix.

test/modules/TestResultCoordinator/Reports/EdgeHubRestartTest/EdgeHubRestartDirectMethodReport.cs
and-rewsmith: Just realized I was also inconsistent within the reportMetadata classes as well. I'll lyk when I fix.

e2e_deployment_files/connectivity_deployment.template.json
philipktlin: But this is not great in current settings class, we should not have a value for transport type in settings if it runs in sender mode.  Maybe create a bug to track.

and-rewsmith: Are you saying the desired behavior is to throw an exception in the case where you use transportType and are a sender?

test/modules/TestResultCoordinator/Reports/TestReportMetadataBase.cs
```diff
@@ -0,0 +1,18 @@
...
+        public TestReportMetadataBase(string testDescription)
+        {
+            this.TestDescription = testDescription;
+        }
+
+        public string TestDescription { get; }
+
+        public override string ToString()
+        {
+            return $"TestDescription: {this.TestDescription}";
```
philipktlin: any usage of this method? I saw you implement ToString() for derived test report metadata, or you should call base.ToString method in child class ToString method?

test/modules/TestResultCoordinator/Reports/TestReportMetadataBase.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator.Reports
+{
+    public abstract class TestReportMetadataBase
+    {
+        public TestReportMetadataBase(string testDescription)
+        {
+            this.TestDescription = testDescription;
```
philipktlin: Please implement TestReportType and TestOperationResultType as well.

test/modules/TestResultCoordinator/Reports/TestInfoReportMetadata.cs
```diff
@@ -19,7 +20,7 @@ public TestInfoReportMetadata()
 
         public override string ToString()
         {
-            return $"Source: {this.Source}, TestOperationResultType: {this.TestOperationResultType.ToString()}, ReportType: {this.TestReportType.ToString()}";
+            return $"TestDescription: {this.TestDescription}, Source: {this.Source}, TestOperationResultType: {this.TestOperationResultType.ToString()}, ReportType: {this.TestReportType.ToString()}";
```
philipktlin: Suggest to call base.ToString() here.

test/modules/TestResultCoordinator/Reports/TestReportMetadataBase.cs
and-rewsmith: I tried doing this initially, but thought it was a bad idea due to misunderstanding with properties. Figured it out though and made the change.

test/modules/TestResultCoordinator/Reports/TestReportMetadataBase.cs
and-rewsmith: Yes, I will call the base class ToString() in the child class

and-rewsmith:
Passing test run:
https://msazure.visualstudio.com/One/_build/results?buildId=29492695&view=logs&j=bce603fb-13ef-563c-8375-f1c939d7117f&t=765e9a87-f4df-5ea8-7946-e2b14c835574

e2e_deployment_files/connectivity_deployment.template.json
philipktlin: No.  If you are running a sender, the code to extract transport type should not even be run.

PR #2646, opened 2020-03-05
Connectivity test: Add testinfo to report

and-rewsmith:
**Overview**
This PR adds contextual information to the connectivity test report that will be consumed by the dashboard. I chose to add this information to the TestSummary (not the test info report). 

**Where this contextual information will be**
This test information will be stored in the TestSummary as a dictionary under the name TestInfo. It will contain these inner key value pairs:
- build id
- build source branch
- edgelet source branch
- images source branch
- test build number
- device id
- host platform

**Implementation**
In the connectivity-deploy yaml I concatenate some pipeline variables in dictionary format, then parse into a dict in the TRC. I also move some pre-existing root-level test summary fields to this TestInfo report (i.e. device id, test build number).


e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -337,9 +337,6 @@
               "Logging:LogLevel:Microsoft": {
                 "value": "Error"
               },
-              "TEST_BUILD_NUMBER": {
-                "value": "<Test.BuildNumber>"
-              },
```
and-rewsmith: removed from the TRC environment variables and passed instead through the yaml testInfo field

test/modules/TestResultCoordinator/Services/TestResultReportingService.cs
```diff
@@ -82,7 +82,7 @@ async void DoWorkAsync(object state)
                 this.logger.LogError(ex, "Exception happened when uploading logs");
             }
 
-            var testSummary = new TestSummary(Settings.Current.TestBuildNumber, Settings.Current.DeviceId, testResultReports, blobContainerUri);
```
and-rewsmith: Test build number and device id now wrapped in test info

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -62,7 +62,15 @@ class Settings
...
             this.StorageAccountConnectionString = Preconditions.CheckNonWhiteSpace(storageAccountConnectionString, nameof(storageAccountConnectionString));
-            this.NetworkControllerType = GetNetworkControllerType(networkControllerRunProfileName);
+            this.NetworkControllerType = this.GetNetworkControllerType(networkControllerRunProfileName);
+
+            this.TestInfo = testInfo.Split(",", StringSplitOptions.RemoveEmptyEntries)
+                                .Select(x => (KeyAndValue: x, SplitIndex: x.IndexOf('=')))
+                                .Where(x => x.SplitIndex >= 1)
+                                .ToDictionary(
+                                    x => x.KeyAndValue.Substring(0, x.SplitIndex),
+                                    x => x.KeyAndValue.Substring(x.SplitIndex + 1, x.KeyAndValue.Length - x.SplitIndex - 1));
```
and-rewsmith: This is the same logic as we implemented in longhaul. Might be a good idea to consolidate it in edge util

and-rewsmith:
Test result here:
https://msazure.visualstudio.com/One/_build/results?buildId=29340366&view=logs&j=bce603fb-13ef-563c-8375-f1c939d7117f&t=765e9a87-f4df-5ea8-7946-e2b14c835574

There was a C2D issue, but it adds the test info as expected.

builds/e2e/connectivity.yaml
```diff
@@ -119,10 +119,12 @@ jobs:
         parameters:
           release.label: 'ct$(agent.group)'
           test.buildNumber: '$(Build.BuildNumber)'
+          test.buildId: '$(Build.BuildId)'
```
philipktlin: why do you need build Id?

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -378,6 +375,9 @@
               },
               "NetworkControllerRunProfile": {
                 "value": "<NetworkController.RunProfile>"
+              },
+              "TEST_INFO": {
```
philipktlin: Please also update edgehub_restart_deployment.template.json

test/modules/TestResultCoordinator/Reports/TestSummary.cs
```diff
@@ -1,24 +1,21 @@
...
     using Microsoft.Azure.Devices.Edge.Util;
 
     class TestSummary
     {
-        internal TestSummary(string testBuildNumber, string deviceId, ITestResultReport[] testResultReports, string blobContainerUri)
+        internal TestSummary(Dictionary<string, string> testInfo, ITestResultReport[] testResultReports, string blobContainerUri)
         {
-            this.TestBuildNumber = Preconditions.CheckNonWhiteSpace(testBuildNumber, nameof(testBuildNumber));
-            this.DeviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
+            this.TestInfo = Preconditions.CheckNotNull(testInfo, nameof(testInfo));
```
philipktlin: nit: do we need to enforce not empty? your call?

test/modules/TestResultCoordinator/Reports/TestSummary.cs
```diff
@@ -1,24 +1,21 @@
...
             this.TestResultReports = Preconditions.CheckNotNull(testResultReports, nameof(testResultReports));
             this.IsPassed = testResultReports.All(r => r.IsPassed);
             this.BlobContainerUri = blobContainerUri ?? string.Empty;
         }
 
-        public string TestBuildNumber { get; }
-
-        public string DeviceId { get; }
```
philipktlin: Since device id is getting from environment value, just keep it.  This will be a bit easy to find it when looking for log in blob storage, which part of the path is the device id.

test/modules/TestResultCoordinator/Reports/TestSummary.cs
```diff
@@ -1,24 +1,21 @@
...
+        public Dictionary<string, string> TestInfo { get; }
```
philipktlin: Can you verify if SortedDictionary will output in alphabetic order when serialization? if so, please use it.

test/modules/TestResultCoordinator/Settings.cs
philipktlin: Agree, but not edge util; should be ModuleUtil project under test folder.

test/modules/TestResultCoordinator/Reports/TestSummary.cs
and-rewsmith: I think so, because this info will be relied upon for proper display in dashboard

builds/e2e/connectivity.yaml
and-rewsmith: For showing a link to Dev Ops

test/modules/TestResultCoordinator/Reports/TestSummary.cs
and-rewsmith: Tested with dotnet fiddle. It does so I made the change.

e2e_deployment_files/edgehub_restart_deployment.template.json
```diff
@@ -267,6 +267,9 @@
               },
               "StartAfter": {
                 "value": "<TestStartDelay>"
+              },
```
philipktlin: added to the wrong module.

test/modules/ModuleLib/ModuleUtil.cs
```diff
@@ -120,5 +122,17 @@ static void WriteLog(ILogger logger, LogLevel logLevel, string message)
                 logger.Log(logLevel, message);
             }
         }
+
+        public static SortedDictionary<string, string> ParseTestInfo(string testInfo)
```
philipktlin: nit: put a line of comment to specify the format of test Info.

e2e_deployment_files/edgehub_restart_deployment.template.json
and-rewsmith: oops good catch, thanks

test/modules/ModuleLib/ModuleUtil.cs
and-rewsmith: I was thinking of putting the specific fields, but decided against it because the comments can easily become outdated. I am going to put a generic description specifying where the test information comes from.

philipktlin: please mention the format: key=value[,key=value]

test/modules/TestResultCoordinator/Reports/TestSummary.cs
philipktlin: are you going to check dictionary is not empty?

and-rewsmith: I'm doing the check in ModuleUtil

builds/e2e/templates/connectivity-deploy.yaml
```diff
@@ -52,6 +54,16 @@ steps:
...
         chmod +x $(Agent.HomeDirectory)/../artifacts/${{ parameters['images.artifact.name'] }}/scripts/connectivityTest.sh
+
+        testInfo="TestBuildNumber=${{ parameters['test.buildNumber'] }}"
+        testInfo="$testInfo,TestBuildId=${{ parameters['test.buildId'] }}"
+        testInfo="$testInfo,BuildSourceBranch=${{ parameters['build.source.branch'] }}"
+        testInfo="$testInfo,EdgeletSourceBranch=${{ parameters['edgelet.source.branch'] }}"
+        testInfo="$testInfo,ImagesSourceBranch=${{ parameters['images.source.branch'] }}"
+        testInfo="$testInfo,HostPlatform=${{ parameters['metricsCollector.hostPlatform'] }}"
+        testInfo="$testInfo,RunProfile=${{ parameters['networkController.runProfile'] }}"
+        testInfo="$testInfo,CustomEdgeAgentImage=${{ parameters['customEdgeAgent.image'] }:-None}"
```
philipktlin: how about edgeRuntime.buildNumber pipeline variable?

test/modules/TestAnalyzer/Settings.cs
```diff
@@ -39,12 +40,7 @@ class Settings
             this.WebhostPort = Preconditions.CheckNonWhiteSpace(webhostPort, nameof(webhostPort));
             this.ToleranceInMilliseconds = Preconditions.CheckRange(tolerance, 0);
             this.StoragePath = storagePath;
-            this.TestInfo = testInfo.Split(",", StringSplitOptions.RemoveEmptyEntries)
-                                .Select(x => (KeyAndValue: x, SplitIndex: x.IndexOf('=')))
-                                .Where(x => x.SplitIndex >= 1)
-                                .ToDictionary(
-                                    x => x.KeyAndValue.Substring(0, x.SplitIndex),
-                                    x => x.KeyAndValue.Substring(x.SplitIndex + 1, x.KeyAndValue.Length - x.SplitIndex - 1));
+            this.TestInfo = ModuleUtil.ParseTestInfo(testInfo);
```
philipktlin: do you need to add Device id in TestInfo?

builds/e2e/templates/connectivity-deploy.yaml
and-rewsmith: I don't think I needed so I didn't add.

test/modules/TestAnalyzer/Settings.cs
and-rewsmith: I don't think I needed so I didn't add. Lmk if you disagree.

builds/e2e/templates/connectivity-deploy.yaml
philipktlin: Then why do you need CustomEdgeAgentImage and CustomEdgeHubImage, but not edgeRuntime.buildNumber? edgeRuntime.buildNumber is used to control connectivity test to run with a specific docker image from other build instead of latest build from given branches.

and-rewsmith: Oh I misunderstood. I'll add.

and-rewsmith: I misunderstood what the custom image was, but the reason I added though was to be able to filter when someone runs connectivity tests for 109. With just the build number, we don't have any information to know that connectivity tests came from the 109 branch. We will need some identifying information about what branch, however if we eventually plan to backport the connectivity tests then we don't have to fix. What do you think?

and-rewsmith: I think we should leave it as is and backport the tests.

philipktlin: 1. I would just to get this PR committed first; and have another PR to address this problem.
2. To address this problem, are you going to add 1 more field and add people to update that new field when edgeRuntime.buildNumber is set?
3. Or we should just add to get edgeRuntime.SourceBranch instead of edgeRuntime.buildNumber; and use source branch to find the corresponding build number?

and-rewsmith: I think we should do 3 only if we think that we will never backport the tests. If we plan on backporting the tests, then we won't need to support this hack anymore.

PR #2647, opened 2020-03-05
[mqtt] Add afl fuzz testing scripts

dmolokanov:
Add scripts and description to run `cargo afl` tool for fuzz testing

### DIFF
--- a/mqtt/build/linux/mqtt-fuzz-rerun.sh
+++ b/mqtt/build/linux/mqtt-fuzz-rerun.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+set -euo pipefail
+
+tmux attach-session -t mqtt-fuzz && exit 0 || :
+
+DIR=$1
+if [ ! -d "$DIR" ]; then
+    echo "$DIR does not exist"
+fi
+
+cd $DIR
+
+rm -rf in/ in.cmin/ in.tmin/
+
+mkdir in.cmin/
+i=0
+for f in out/fuzzer*/queue/*; do
+    cp "$f" "in.cmin/$i"
+    i=$(( i + 1 ))
+done
+
+mkdir in.tmin/
+cargo afl cmin -i in.cmin/ -o in.tmin/ target/release/mqtt3-fuzz
+rm -rf in.cmin/
+
+mkdir in.cmin/
+for f in in.tmin/*; do
+    cargo afl tmin -i "$f" -o "in.cmin/$(basename $f)" target/release/mqtt3-fuzz
+done
+rm -rf in.tmin/
+
+mkdir in/
+cargo afl cmin -i in.cmin/ -o in/ target/release/mqtt3-fuzz
+rm -rf in.cmin/
+
+rm -rf out/
+
+(
+sleep 1
+
+tmux set-option -t 'mqtt-fuzz' pane-border-status top
+tmux set-option -t 'mqtt-fuzz' status-interval 1
+tmux set-option -t 'mqtt-fuzz' status-right '#{pane_title} %Y-%m-%d %H:%M:%S'
+
+tmux split-window -t 0 -h -p 66 'read -rsp "Press enter to continue..." && cargo afl fuzz -i in -o out -S fuzzer02 -t 1000 target/release/mqtt3-fuzz'
+tmux select-pane -t 1 -T 'fuzzer02'
+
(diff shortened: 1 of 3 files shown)


mqtt/mqtt3/README.md
```diff
@@ -20,6 +20,31 @@ The crate is not published to crates.io yet. Please generate docs locally with `
...
+sudo apt install tmux binutils
+cargo install afl
+```
+
+## Run fuzzer
+
+```bash
+echo 'core' | sudo tee /proc/sys/kernel/core_pattern
+
+which gold
```
arsing: I don't think you need `gold`. In any case, you don't need to test for it here.

mqtt/mqtt3/README.md
```diff
@@ -20,6 +20,31 @@ The crate is not published to crates.io yet. Please generate docs locally with `
...
+
+/build/linux/mqtt-fuzz.sh ../mqtt3-fuzz/
```
arsing: This should be `build/linux/mqtt-fuzz.sh ` ? No leading `/`.

Also if you're going to run this from the `mqtt` directory, then the path to `mqtt3-fuzz` would be just `mqtt3-fuzz`, not `../mqtt3-fuzz`

mqtt/mqtt3/README.md
```diff
@@ -20,6 +20,31 @@ The crate is not published to crates.io yet. Please generate docs locally with `
...
+
+```bash
+sudo apt install tmux binutils
+cargo install afl
+```
+
+## Run fuzzer
+
+```bash
+echo 'core' | sudo tee /proc/sys/kernel/core_pattern
```
arsing: Add a comment why this is needed. Something like:

```sh
# afl requires dumps to be taken as quickly as possible, so configure kernel to just write the coredump instead of anything fancy.
```

mqtt/mqtt3/README.md
```diff
@@ -20,6 +20,31 @@ The crate is not published to crates.io yet. Please generate docs locally with `
...
+
+which gold
+
+/build/linux/mqtt-fuzz.sh ../mqtt3-fuzz/
+
+build/linux/mqtt-fuzz-rerun.sh ../mqtt3-fuzz/
+
+echo '|/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h' | sudo tee /proc/sys/kernel/core_pattern
```
arsing: You don't need to mention this. Or if you do want to mention it, at least say that the user should only do this when they're done with fuzzing and want to restore the original value. Even then, this is just what the default was on my openSUSE machine; it wouldn't necessarily be this on everyone else's.

mqtt/mqtt3/README.md
arsing: Add some explanation above each command. Something like:

```sh
# mqtt-fuzz.sh spawns a tmux session with 6 panes. The first pane is the master afl instance, and the remaining five are slave instances.
# Press any key to start the master instance, wait for it to start running, then press any key in the five slave instances to start them too.
```

and

```sh
# mqtt-fuzz-rerun.sh uses the output of a previous run of mqtt-fuzz.sh or mqtt-fuzz-rerun.sh as the starting corpus.
# Use this if you interrupted a previous run of mqtt-fuzz.sh or mqtt-fuzz-rerun.sh and want to resume from where it left off.
```

dmolokanov: Nice additions. Thanks!

mqtt/mqtt3/README.md
```diff
@@ -20,6 +20,37 @@ The crate is not published to crates.io yet. Please generate docs locally with `
...
 
+# Fuzz testing
+
+The crate is coming with fuzz tests. 
+
+## Prerequisites
+
+```bash
+sudo apt install tmux binutils
+cargo install afl
```
arsing: This needs to match the `afl` dep of the crate, so make this `cargo install afl --version '^0.6'`. Then it won't break when afl 0.7 is released.

mqtt/build/linux/mqtt-fuzz-rerun.sh
```diff
@@ -0,0 +1,65 @@
...
+mkdir in/
+cargo afl cmin -i in.cmin/ -o in/ target/release/mqtt3-fuzz
+rm -rf in.cmin/
+
+rm -rf out/
+
+(
+sleep 1
+
+tmux set-option -t 'mqtt-fuzz' pane-border-status top
```
arsing: Sorry, remove these three `set-option` lines too. They're just personal customization (and should belong in `~/.tmux.conf` anyway).

PR #2649, opened 2020-03-05
[mqtt] Fix build errors in mqtt pipeline

dmolokanov:
Make sure MQTT pipeline succeeds for both linux and windows amd64 architectures


mqtt/build/windows/openssl.ps1
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Microsoft. All rights reserved.
```
arsing: Why does it need openssl?

dmolokanov: some script uses it. I just copied all related scripts to build a folder. don't we need it for the same reason as edgelet? or for edgelet it is just to build with libiothsm?

arsing: iotedged needs openssl. mqtt code does not.

Yes, the existing Rust install script also installs openssl, but that doesn't mean we want to install openssl for the mqtt code. It's just going to slow down the build for no reason. You should instead edit the script to make installing openssl optional.

In fact I see you just wholesale copied all the scripts for edgelet to this. Why did you do that? Now we have two copies of identical scripts and have to worry about making fixes to both of them.

dmolokanov: My goal was to make a pipeline for broker up and running, without spending too much time polishing scripts that most likely will change soon. Now I see that scripts are way to different...

PR #2651, opened 2020-03-05
Port x509 fix to 1.0.9

kkdawkins:


scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -280,6 +280,9 @@ function Initialize-IoTEdge {
 
     Set-ProvisioningMode
     Set-Certificates
+    if ($DpsX509 -or $ManualX509) {
```
arsing: These don't exist in this branch's script.

PR #2658, opened 2020-03-06
[mqtt] Add simple file-based persistence

myagley:
Implements very basic file persistence using serde and bincode. The expectation is that this file format is improved in the future to something more efficient and to add the versioning. The focus of this PR is to get the file persistence to be resilient in the face of failure, regardless of the format.

* updates needed types to be `Serialize` and `Deserialize`
* defines a very simple file format based on bincode
* adds a file persistor which atomically persists broker state snapshots
* adds proptest based tests to shake out the bincode file format
* adds proptest and failpoint based tests to "fuzz" the file persistor's atomic guarantees
* updates the snapshot tick to not snapshot immediately


arsing:
I think you need to rebase this on master to get rid of Denis's commits at the beginning of your branch?

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -1,28 +1,385 @@
...
+    async fn load(&mut self) -> Result<Option<BrokerState>, Self::Error> {
+        let dir = self.dir.clone();
+        let format = self.format.clone();
+
+        let res = tokio::task::spawn_blocking(move || {
+            let path = dir.join(format!("{}.{}", STATE_DEFAULT_STEM, STATE_EXTENSION));
+            if path.exists() {
+                info!("loading state from file {}.", path.display());
+
+                fail_point!("filepersistor.load.fileopen", |_| {
```
dmolokanov: It is a seam for us to emulate failure during following operation in tests, right?

myagley:
I tried rebasing it when I started but it was too confusing. The merge was more straightforward

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -1,28 +1,385 @@
...
+        fail_point!("filepersistor.store.spawn_blocking", |_| {
+            Err(Error::from(ErrorKind::TaskJoin))
+        });
+        res.context(ErrorKind::TaskJoin)?
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub enum ErrorReason {
+    FileOpen,
```
dmolokanov: It'd be nice these enum values to contain a path to file

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -1,28 +1,385 @@
...
+    FileUnlink,
+    CreateDir,
+    ReadDir,
+    Symlink,
+    SymlinkUnlink,
+    Serialize,
+    Deserialize,
+}
+
+impl fmt::Display for ErrorReason {
```
dmolokanov: In my "yet another error crate" I tried to get away with [derive_more](https://docs.rs/derive_more/0.99.3/derive_more/). Almost succedded :)

mqtt/mqtt-broker/tests/persist_failpoints.rs
```diff
@@ -0,0 +1,161 @@
...
+            .enable_all()
+            .build()
+            .unwrap()
+            .block_on(test_persistor(count, ops));
+        scenario.teardown();
+    }
+}
+
+#[test]
+fn test_failpoints_regression1() {
```
dmolokanov: right now regression1 suffix makes sense only for the author of a PR. I think more user-friendly name will benefit everyone, if possible.

mqtt/mqtt-broker/tests/persist_failpoints.rs
```diff
@@ -90,7 +90,8 @@ fn test_failpoints_smoketest() {
             let mut persistor = FilePersistor::new(path, BincodeFormat::new());
 
             let result = persistor.load().await;
-            assert!(result.is_err());
+            let err = result.unwrap_err();
+            assert_eq!(&ErrorKind::TaskJoin, err.kind());
```
dmolokanov: `assert_match(result, Err(ErrorKind::TaskJoin.into()));` instead?

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -76,8 +76,8 @@ pub struct FilePersistor<F> {
     format: F,
     previous_count: usize,
 
-    /// epoch is bumped on every store to disambiguate stores in the same timestamp
-    epoch: u16,
+    /// seq is bumped on every store to disambiguate stores in the same timestamp
+    seq: u16,
```
dmolokanov: it's like a `version`, right?

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -1,28 +1,385 @@
...
+        Self
+    }
+}
+
+impl FileFormat for BincodeFormat {
+    type Error = Error;
+
+    fn load<R: Read>(&self, reader: R) -> Result<BrokerState, Self::Error> {
+        let decoder = GzDecoder::new(reader);
+        fail_point!("bincodeformat.load.deserialize_from", |_| {
```
dmolokanov: I wonder if we can make fail_point! to be aware of our error kinds in a way, that we can use it like this
```rust
fail_point_ext(ErrorKind::Persist(ErrorReason::Deserialize));
```
that generates the same output. 
It may require `ErrorKind::Persist` to implement some custom trait to return the correct string though. Probably with proc_macro :)

mqtt/mqtt-broker/tests/persist_failpoints.rs
myagley: yeah, i probably don't even need these. proptest updates a regressions file that it uses to seed future runs.

mqtt/mqtt-broker/tests/persist_failpoints.rs
myagley: I had considered that, but didn't want to bring in another dependency just for this one simple case.

mqtt/mqtt-broker/src/persist.rs
myagley: it's kind of like a version but it resets on every restart.

mqtt/mqtt-broker/src/persist.rs
myagley: i had considered that. perhaps we can optimize this in a future PR?

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -1,28 +1,385 @@
...
+
+        let res = tokio::task::spawn_blocking(move || {
+            let path = dir.join(format!("{}.{}", STATE_DEFAULT_STEM, STATE_EXTENSION));
+            if path.exists() {
+                info!("loading state from file {}.", path.display());
+
+                fail_point!("filepersistor.load.fileopen", |_| {
+                    Err(Error::from(ErrorKind::Persist(ErrorReason::FileOpen)))
+                });
+                let file = OpenOptions::new()
```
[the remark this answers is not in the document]
myagley: I could. However, the following call to `format.load` is going to block anyways. I decided to drop this all into a `spawn_blocking` and not rely on any of tokio's file stuff.

mqtt/mqtt-broker/src/persist.rs
myagley: Yes, this is a test only thing. The macro doesn't generate any code in normal operation. The test code turns on a feature that causes code to be generated to enable fault injection.

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -1,28 +1,385 @@
...
         Ok(())
     }
 }
+
+/// An abstraction over the broker state's file format.
+pub trait FileFormat {
+    type Error: Into<Error>;
+
+    /// Load `BrokerState` from a reader.
+    fn load<R: Read>(&self, reader: R) -> Result<BrokerState, Self::Error>;
```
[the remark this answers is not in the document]
myagley: Having used the async file operations in the past, I think it will be more straight forward to use the standard blocking calls when implementing a new `FileFormat`. This thing gets run on a completely separate thread pool with `spawn_blocking`.

PR #2659, opened 2020-03-06
Fixing C2D sender for connectivity tests

dylanbronson:
When c2d sender throws an exception it crashes. This was the original intent, since the sender shouldn't fail. That assumption turned out to be wrong, though.

it will fail if it exceeds 50 messages that are in the cloud queue. So if edgeHub is offline and 50 messages build up, we should just keep trying to send. Eventually the queue will be drained, and we'll be able to continue sending.


test/modules/CloudToDeviceMessageTester/CloudToDeviceMessageSender.cs
```diff
@@ -58,36 +58,37 @@ public async Task StartAsync(CancellationToken ct)
...
             while (!ct.IsCancellationRequested && this.IsTestTimeUp(testStartAt))
             {
-                MessageTestResult testResult = await this.SendCloudToDeviceMessageAsync(batchId, this.trackingId);
-                await ModuleUtil.ReportTestResultAsync(this.testResultReportingClient, this.logger, testResult);
-                this.messageCount++;
-                await Task.Delay(this.messageDelay, ct);
+                MessageTestResult testResult = null;
+                try
+                {
+                    testResult = await this.SendCloudToDeviceMessageAsync(batchId, this.trackingId);
```
yophilav: Combine line 61 and 64 pls

test/modules/CloudToDeviceMessageTester/CloudToDeviceMessageSender.cs
```diff
@@ -58,36 +58,37 @@ public async Task StartAsync(CancellationToken ct)
...
+                    this.messageCount++;
+                    await ModuleUtil.ReportTestResultAsync(this.testResultReportingClient, this.logger, testResult);
+                    await Task.Delay(this.messageDelay, ct);
+                }
+                catch (Exception ex)
```
yophilav: @philipktlin  Should this be catching a specific exception for full message queue and rethrow the rest of the exception? Or should it just swallow all the exception here?

test/modules/CloudToDeviceMessageTester/CloudToDeviceMessageSender.cs
```diff
@@ -58,36 +58,37 @@ public async Task StartAsync(CancellationToken ct)
...
-                    BatchId = batchId.ToString(),
-                    SequenceNumber = this.messageCount.ToString()
-                };
-            }
-            catch (Exception ex)
+            var message = new Message(Encoding.ASCII.GetBytes("Cloud to device message."));
+            message.Properties.Add(TestConstants.Message.SequenceNumberPropertyName, this.messageCount.ToString());
+            message.Properties.Add(TestConstants.Message.BatchIdPropertyName, batchId.ToString());
+            message.Properties.Add(TestConstants.Message.TrackingIdPropertyName, trackingId);
+            await this.serviceClient.SendAsync(this.deviceId, message);
```
yophilav: Why did you move the try-catch() up the call stack?  (Pure curiosity).. 
I'd expect the exception to be thrown by `await this.serviceClient.SendAsync(this.deviceId, message);` and you can prob just wrap the try-catch() here like in your original code.

test/modules/CloudToDeviceMessageTester/CloudToDeviceMessageSender.cs
dylanbronson: Changed to catch specific exception, log and swallow

Also will catch general exception, but will log and re-throw.

test/modules/CloudToDeviceMessageTester/CloudToDeviceMessageSender.cs
dylanbronson: The way I had it originally, if I catch the DeviceExceededMaximum (legal in this case), then I'll also update the sequence number, which I don't want to do, since I want to retry on that sequence number until it goes through. 

Now, if the Send method throws the DeviceExceededMaximum exception, we will restart the whole call, and we can update the sequence number only once the send goes through

PR #2662, opened 2020-03-06
[k8s PP]Update iotedged image to debian:10-slim to fix vulnerability.

darobs:
[For reference: CVE-2019-5188](https://security-tracker.debian.org/tracker/CVE-2019-5188)


darobs:
> I haven't tested at my end, but presumably you have :)

Yup.  Didn't ask for a review until I did!

PR #2663, opened 2020-03-07
making RetryingSink to set same exceptions in case if the process get…

vipeller:
making RetryingSink to set same exceptions in case if the process gets cancelled (Task.Delay vs ThrowIfCancellationRequested)


edge-hub/src/Microsoft.Azure.Devices.Routing.Core/sinks/RetryingSink.cs
```diff
@@ -74,13 +74,17 @@ public async Task<ISinkResult<T>> ProcessAsync(ICollection<T> ts, CancellationTo
...
             }
-            catch (OperationCanceledException ex)
-            {
-                failed.AddRange(messages);
-                rv = new SinkResult<T>(succeeded, failed, invalid, new SendFailureDetails(FailureKind.InternalError, ex));
-            }
             catch (Exception ex)
             {
+                // this is thrown by ThrowIfCancellationRequested() which is different to that Task.Delay() throws
+                // when it gets cancelled. Warping it to give a consistent output when the process is cancelled.
```
richma-ms: >Warping [](start = 43, length = 7)

Did you mean "wrapping"

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/sinks/RetryingSink.cs
```diff
@@ -74,13 +74,13 @@ public async Task<ISinkResult<T>> ProcessAsync(ICollection<T> ts, CancellationTo
...
                 rv = new SinkResult<T>(succeeded, failed, invalid, failureDetails);
             }
-            catch (OperationCanceledException ex)
-            {
-                failed.AddRange(messages);
-                rv = new SinkResult<T>(succeeded, failed, invalid, new SendFailureDetails(FailureKind.InternalError, ex));
-            }
             catch (Exception ex)
             {
+                if (ex is OperationCanceledException)
```
philipktlin: why do we changing this behavior?  or should we change our test code to catch OperationCanceledException instead? https://stackoverflow.com/questions/13040428/difference-between-operationcanceledexception-and-taskcanceledexception

vipeller: now that it turned out that one is a base-class of the other, then yes, maybe it is best not to transform one to the other, but modify the test to accept not the type, but the inheritance line. Doing the change

PR #2664, opened 2020-03-08
Allow Edge attached Hardware accelerators

mustafakasap:
"device_cgroup_rules" attribute is not allowed to be changed. But this is mandatory setting to be able to access USB based hardware accelerators (like INTEL VPUs). Without this attribute set, only way is to run IoTEdge module with previlidged rights... But using this setting it is not required to give admin previliges and also it allows USB attached dynamic devices.

Here is an example of intel OpenVino docker image (to use intel VPU, FPGA etc. attached to edge devie) https://docs.openvinotoolkit.org/latest/_docs_install_guides_installing_openvino_docker_linux.html
as in the above example to run the iotedge module without admin privilages, you will use following container create parameters:

docker run --device-cgroup-rule='c 189:* rmw' -v /dev/bus/usb:/dev/bus/usb <image_name>

Here there is no way to pass "--device-cgroup-rule". parameter in IoTEdge Hub... Today it allows such manifest:

{
   "HostConfig": {
      "PortBindings": {
         ...
         ]
      },
      "DeviceCgroupRules": [
         "c 189:* rmw"
      ],
....
}

But above manifest always reflected as "DeviceCgroupRules": null on the edge device... 

so above changes hope to solve this issue...

### DIFF
--- a/edgelet/docker-rs/src/models/host_config.rs
+++ b/edgelet/docker-rs/src/models/host_config.rs
@@ -93,9 +93,9 @@ pub struct HostConfig {
     // /// A list of devices to add to the container.
     // #[serde(rename = "Devices", skip_serializing_if = "Option::is_none")]
     // devices: Option<Vec<crate::models::DeviceMapping>>,
-    // /// a list of cgroup rules to apply to the container
-    // #[serde(rename = "DeviceCgroupRules", skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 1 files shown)


arsing:
The property will be preserved if you set it. That's why all the properties are commented out in the first place. See the DEVNOTE comment at the top of the file.

PR #2670, opened 2020-03-09
Message priority in StoringAsyncEndpointExecutor

richma-ms:
This changes adds the last leg of the message priority logic and enables the feature E2E.

The executor now creates a separate sequentialstore, checkpointer, and FSM for each priority during startup.  The code to parse dispatcher names and pass it onto endpoints is removed, because it's never actually used in product.

When a message arrives, it's enqueued into the corresponding sequentialstore, and pumped using the corresponding FSM+checkpointer.  While pumping messages, the loop will visit each queue in priority order, and re-check high priority queues after every batch.


edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -132,63 +132,153 @@ public async Task SetEndpoint(Endpoint newEndpoint, IList<uint> priorities)
...
+        static string GetMessageQueueId(string endpointId, uint priority)
+        {
+            if (priority == RouteFactory.DefaultPriority)
+            {
+                // We need to maintain backwards compatibility
+                // for existing sequential stores that don't
+                // have the "_Pri<x>" suffix. We use the default
+                // priority (2,000,000,000) for this, which means
+                // the store ID is just the endpoint ID.
+                return endpointId;
```
dylanbronson: So this means that even new routes that are set to the default priority won't have the "_Pri" suffix, correct? 
Also, if a customer explicitly set their priority to 2,000,000,000 it won't have pri either, right?

richma-ms: yes for both, but how the underlying stores are handled is opaque to the customer anyways.  Short of cracking open the RockDB files, they'll never be able to tell that priority-2bn results in a different naming format.

---
In reply to: [389952225](https://github.com/Azure/iotedge/pull/2670#discussion_r389952225) [](ancestors = 389952225)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Dispatcher.cs
```diff
@@ -21,19 +20,19 @@ public class Dispatcher : IDisposable
         readonly CancellationTokenSource cts;
         readonly IEndpointExecutorFactory endpointExecutorFactory;
         readonly AtomicReference<ImmutableDictionary<string, IEndpointExecutor>> executors;
-        readonly ICheckpointer checkpointer;
+        readonly MasterCheckpointer masterCheckpointer;
```
dylanbronson: What's the idea of the MasterCheckpointer and why are we using it here?

richma-ms: this has always been an instance of MasterCheckpointer, it just used to be cached in two places:
1. here as the "global" checkpointer to checkpoint orphaned messages
2. in CheckpointerEndpointExecutorFactory as an instance of ICheckpointerFactory

Now that the executor is responsible for creating the checkpointer, CheckpointerEndpointExecutorFactory is not useful anymore so the ICheckpointerFactory reference is gone.  I just changed the type here to the object-class so that we can access both ICheckpointer and ICheckpointerFactory interfaces as needed.

---
In reply to: [390015947](https://github.com/Azure/iotedge/pull/2670#discussion_r390015947) [](ancestors = 390015947)

philipktlin: where do you need to change type from ICheckpointer to MasterCheckpointer? MasterCheckpointer implements both ICheckpointer and ICheckpointerFactory.  Do you need to use ICheckpointerFactory?  if not, ICheckpointer should be used same as before. Do I miss anything?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Dispatcher.cs
```diff
@@ -54,9 +53,11 @@ public static async Task<Dispatcher> CreateAsync(string id, string iotHubName, I
             Preconditions.CheckNotNull(endpointsWithPriorities);
             Preconditions.CheckNotNull(factory);
 
+            var masterCheckpointer = await MasterCheckpointer.CreateAsync(id, NullCheckpointStore.Instance);
+
             IEnumerable<Task<IEndpointExecutor>> tasks = endpointsWithPriorities.Select(e => factory.CreateAsync(e.Key, e.Value));
             IEndpointExecutor[] executors = await Task.WhenAll(tasks);
-            return new Dispatcher(id, iotHubName, executors, factory, new NullCheckpointer());
+            return new Dispatcher(id, iotHubName, executors, factory, masterCheckpointer);
```
philipktlin: why do you change from NullCheckpointer to masterCheckpointer? I traced the usages the reason NullCheckpointer is used as isStoreAndForwardEnabled is passed as false in routing module.  So I think it should be kept as NullCheckpointer here, right?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -23,51 +25,50 @@ public class StoringAsyncEndpointExecutor : IEndpointExecutor
...
-
-            this.UpdatePriorities(priorities);
+            this.prioritiesToFsms = new AtomicReference<OrderedDictionary>(new OrderedDictionary());
         }
 
-        public Endpoint Endpoint => this.machine.Endpoint;
+        public Endpoint Endpoint { get; }
 
-        public EndpointExecutorStatus Status => this.machine.Status;
+        public EndpointExecutorStatus Status => this.lastUsedFsm.Status;
```
dylanbronson: Can you find out exactly what the status is being used for please?

Before, the status would answer "what is the status of the EndpointExecutor's FiniteStateMachine"

Now, status is answering "What is the status of the last used FiniteStateMachine by the EndpointExecutor"

It's very similar, because it will still be true that the last used FSM's status is probably what we care about, and all the other FSM's are Idle by definition, but to me it seems more awkward now to ask the status like this.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -23,51 +25,50 @@ public class StoringAsyncEndpointExecutor : IEndpointExecutor
...
+                {
+                    throw new InvalidOperationException($"Endpoint executor for endpoint {this.Endpoint} is closed.");
                 }
 
                 using (MetricsV0.StoreLatency(this.Endpoint.Id))
                 {
-                    IMessage storedMessage = await this.messageStore.Add(this.Endpoint.Id, message);
-                    this.checkpointer.Propose(storedMessage);
+                    // Get the checkpointer corresponding to the queue for this priority
+                    OrderedDictionary snapshot = this.prioritiesToFsms;
```
dylanbronson: Have you looked at SortedDictionary? It's typed, and it will actually order the dictionary by the sorted key values (priorities)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Dispatcher.cs
richma-ms: the operative change here is to pass in ICheckpointerFactory to the executor, because the executor now needs to create a different checkpointer per priority.  So we need ICheckpointerFactory everytime the Dispatcher creates a new executor, both for startup and for "on-the-fly" route changes.

---
In reply to: [390535732](https://github.com/Azure/iotedge/pull/2670#discussion_r390535732) [](ancestors = 390535732)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Dispatcher.cs
richma-ms: it's a mastercheckpointer created on top of a nullcheckpointer store, which has the same effect of not persisting any checkpointing information.  The difference is that NullCheckpointer does not actually do any checkpointing, which broke some of the existing tests that tried to get away with sending the same message multiple-times without changing the offset (as you can see from the test changes).

As far as production goes, there is no valid scenario for disabling StoreAndForward.  That code path is legacy from the original services code and should be considered dead pending removal.

---
In reply to: [390541357](https://github.com/Azure/iotedge/pull/2670#discussion_r390541357) [](ancestors = 390541357)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: will look into it

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: I followed the call stack, looks like it's only called by tests now.  Maybe something called it before.  We can remove this if we ever get around to cleaning up the SyncEndpointExecutor and AsyncEndpointExecutor implementations, both of which are not valid for Edge scenarios.

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/DispatcherTest.cs
```diff
@@ -20,9 +20,9 @@ namespace Microsoft.Azure.Devices.Routing.Core.Test
     [ExcludeFromCodeCoverage]
```
philipktlin: Please help to remove it, otherwise it doesn't include in test coverage calculation.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Dispatcher.cs
philipktlin: Since that code path is legacy, please create a task/bug to remove it.  thanks.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -23,51 +25,50 @@ public class StoringAsyncEndpointExecutor : IEndpointExecutor
...
         readonly Task sendMessageTask;
         readonly AsyncManualResetEvent hasMessagesInQueue = new AsyncManualResetEvent(true);
-        readonly ICheckpointer checkpointer;
-        readonly AtomicReference<IList<uint>> priorities = new AtomicReference<IList<uint>>(new List<uint>());
         readonly AsyncEndpointExecutorOptions options;
-        readonly EndpointExecutorFsm machine;
         readonly CancellationTokenSource cts = new CancellationTokenSource();
+        readonly ICheckpointerFactory checkpointerFactory;
+        readonly EndpointExecutorConfig config;
+        AtomicReference<OrderedDictionary> prioritiesToFsms;
```
philipktlin: Just curious, can we use ConcurrentDictionary instead of AtomicReference<OrderedDictionary>?  @varunpuranik would you give some context about AtomicReference?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -132,63 +132,153 @@ public async Task SetEndpoint(Endpoint newEndpoint, IList<uint> priorities)
...
+                // have the "_Pri<x>" suffix. We use the default
+                // priority (2,000,000,000) for this, which means
+                // the store ID is just the endpoint ID.
+                return endpointId;
+            }
+            else
+            {
+                // The actual ID for the underlying store is of string format:
+                //      <endpointId>_Pri<priority>
+                return endpointId + "_Pri" + priority.ToString();
```
philipktlin: Do we know that endpointId defined in routes is case-insensitive. right?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -132,63 +132,153 @@ public async Task SetEndpoint(Endpoint newEndpoint, IList<uint> priorities)
...
+
+                        if (!haveMessagesRemaining)
+                        {
+                            // All the message queues have been drained, reset the hasMessagesInQueue flag.
+                            this.hasMessagesInQueue.Reset();
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        Events.SendMessagesError(this, ex);
```
philipktlin: For failed messages, it will pick up and re-try every cycle. right?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -132,63 +132,153 @@ public async Task SetEndpoint(Endpoint newEndpoint, IList<uint> priorities)
...
+
+                            StoreMessagesProvider storeMessagesProvider = pair.Item2;
+                            IMessage[] messages = await storeMessagesProvider.GetMessages();
+                            if (messages.Length > 0)
+                            {
+                                await this.ProcessMessages(messages, fsm);
+                                Events.SendMessagesSuccess(this, messages);
+                                MetricsV0.DrainedCountIncrement(this.Endpoint.Id, messages.Length);
+
+                                // Only move on to the next priority if the queue for the current
```
philipktlin: thumbs up for this comment!

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/endpoints/StoringAsyncEndpointExecutorTest.cs
```diff
@@ -25,23 +26,26 @@ public async Task InvokeTest()
             // Arrange
             const int MessagesCount = 10;
             const string EndpointId = "endpoint1";
+            const uint Priority = 100;
             var endpoint = new NullEndpoint(EndpointId);
-            var priorities = new List<uint>() { 0, 1, 2, 100, 101, 102 };
-            var checkpointer = new NullCheckpointer();
+            var priorities = new List<uint>() { Priority };
```
philipktlin: why this reduce priority test data to 1 only?

philipktlin: you simplify this test here for testing basic scenario and have the new test "MessagePrioritiesTest" for different priority. right?

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/endpoints/StoringAsyncEndpointExecutorTest.cs
```diff
@@ -288,6 +295,128 @@ public async Task StoreMessagesProviderIntermittantMessagesTest()
...
+                        // high-pri and normal-pri should also
+                        // have completed before this
+                        Assert.True(highPriMessagesProcessed == HighPriCount);
+                        Assert.True(normalPriMessagesProcessed == NormalPriCount);
+                    }
+                }
+                else
+                {
+                    // Bad test setup
+                    Assert.True(false);
```
philipktlin: nit: suggest to use Assert.True(false, "Bad test setup or even tell how to resolve it.") with description argument.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
dylanbronson: Since it's unused, can you add those as a task, and add removing this as a part of those tasks as well please, so we don't forget

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
myagley: We've used an `AtomicReference` and an immutable dictionary together before. This is because the immutable dictionary is updated by making a new version. This pairs nicely with AtomicReferences.

I think a concurrent dictionary is an issue because the keys need to be sorted?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -23,51 +25,50 @@ public class StoringAsyncEndpointExecutor : IEndpointExecutor
...
+                    throw new InvalidOperationException($"Endpoint executor for endpoint {this.Endpoint} is closed.");
                 }
 
                 using (MetricsV0.StoreLatency(this.Endpoint.Id))
                 {
-                    IMessage storedMessage = await this.messageStore.Add(this.Endpoint.Id, message);
-                    this.checkpointer.Propose(storedMessage);
+                    // Get the checkpointer corresponding to the queue for this priority
+                    OrderedDictionary snapshot = this.prioritiesToFsms;
+                    ICheckpointer checkpointer = ((EndpointExecutorFsm)snapshot[priority]).Checkpointer;
```
myagley: Do we need this cast? Does the dictionary not have a type for its keys.?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -23,51 +25,50 @@ public class StoringAsyncEndpointExecutor : IEndpointExecutor
...
+
+                    IMessage storedMessage = await this.messageStore.Add(GetMessageQueueId(this.Endpoint.Id, priority), message);
+                    checkpointer.Propose(storedMessage);
```
myagley: where was `checkpointer.propose` called before? I'd be cautious about moving it unless this is a completely separate set of checkpointers.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -132,63 +132,153 @@ public async Task SetEndpoint(Endpoint newEndpoint, IList<uint> priorities)
...
+                    EndpointExecutorFsm fsm = new EndpointExecutorFsm(this.Endpoint, checkpointer, this.config);
+
+                    // Add it to our dictionary
+                    snapshot.Add(priority, fsm);
+                }
+                else
+                {
+                    // Update the existing FSM with the new endpoint
+                    var fsm = (EndpointExecutorFsm)snapshot[priority];
+                    newEndpoint.ForEach(async e => await fsm.RunAsync(Commands.UpdateEndpoint(e)));
```
myagley: does this need to be `ForEachAsync` and awaited:

```suggestion
                    await newEndpoint.ForEachAsync(e => fsm.RunAsync(Commands.UpdateEndpoint(e)));

```

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/DispatcherTest.cs
```diff
@@ -20,9 +20,9 @@ namespace Microsoft.Azure.Devices.Routing.Core.Test
...
     public class DispatcherTest : RoutingUnitTestBase
```
myagley: Are these test changes just updates to get things to compile with priorities? Or are the test cases being changed?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Dispatcher.cs
myagley: Yeah, I think its safe to make the interface use the concrete `MasterCheckpointer`,

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
myagley: Yeah, I would love to be able to get rid of the casts if possible.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
myagley: Just noticed the comment above.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
philipktlin: I am not sure that concurrent dictionary needed to be sorted; but actually its AddOrUpdate and GetOrAdd method doesn't execute under locks and need to be taken care by yourself.

richma-ms: yes, the keys need to be sorted, and ConcurrentDictionary doesn't allow that.

Using AtomicReference here enforces a snapshot mechanism, so that the writer updating the dictionary doesn't mess up the reader.

---
In reply to: [392285442](https://github.com/Azure/iotedge/pull/2670#discussion_r392285442) [](ancestors = 392285442)

richma-ms: the message pump needs the dictionary to be sorted, so that it can iterate over messages in ascending priority order.

---
In reply to: [392302436](https://github.com/Azure/iotedge/pull/2670#discussion_r392302436) [](ancestors = 392302436)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: 6470923 - Remove unused message forwarding modes

---
In reply to: [391714814](https://github.com/Azure/iotedge/pull/2670#discussion_r391714814) [](ancestors = 391714814)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: shouldn't matter either way, right?

Before my change, endpointId was just used directly, all I'm doing now is adding a fixed-format suffix.  So regardless of whether endpointId is case-sensitive or not, it should have no new change on the underlying storage.

---
In reply to: [391350197](https://github.com/Azure/iotedge/pull/2670#discussion_r391350197) [](ancestors = 391350197)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: yes, that's handled by the FSM in ProcessMessages()

---
In reply to: [391350993](https://github.com/Azure/iotedge/pull/2670#discussion_r391350993) [](ancestors = 391350993)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: it was called in the same place previously on the "global" checkpointer, right after the message was added to the store.

Now I'm doing the same thing, except I'm using the dictionary to get the checkpointer that corresponds to the message's priority.

---
In reply to: [392287536](https://github.com/Azure/iotedge/pull/2670#discussion_r392287536) [](ancestors = 392287536)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: is there a difference at runtime? O_o

---
In reply to: [392289666](https://github.com/Azure/iotedge/pull/2670#discussion_r392289666) [](ancestors = 392289666)

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/DispatcherTest.cs
richma-ms: for sure

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
philipktlin: follow up question: what happens if there is some reason causing a specific message throws exception during processing every time.  Will this message block all queues? This may already exist.  we can address this in another PR.

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/endpoints/StoringAsyncEndpointExecutorTest.cs
richma-ms: correct, I previously had multiple priorities here because I needed a way to locally verify my last change.  Now that the feature is done E2E, I was able to write a new, more comprehensive testcase, so this isn't needed here anymore.

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/DispatcherTest.cs
richma-ms: it's to get them to run against the new change.  Previously the test took a shortcut with message offsets, and relied on the fact that the NullCheckpointer implementation ignored the offset values.

Now that there's no option to pass in NullCheckpointer anymore, the message offset is actually being calculated and checked, so the test messages we pass in need to have ascending offsets like real messages.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: talked offline, I should be awaiting on a ForEachAsync

---
In reply to: [392457843](https://github.com/Azure/iotedge/pull/2670#discussion_r392457843) [](ancestors = 392457843,392289666)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: the retry loop is not here, it's in the FSM.  This loop is just to get messages out of storage.

If the FSM encounters a message throws an exception, it's checked against a whitelist of exceptions, then the retry action is resolved against the specified RetryStrategy.  Unrecoverable messages are generally dropped on the floor.

All this is already handled by the lower layers, and the logical layer here doesn't need to deal with it.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: changed to use SortedDictionary, much nicer

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/endpoints/StoringAsyncEndpointExecutorTest.cs
```diff
@@ -288,6 +295,128 @@ public async Task StoreMessagesProviderIntermittantMessagesTest()
             iterator.VerifyAll();
         }
 
+        [Fact]
+        public async Task MessagePrioritiesTest()
```
varunpuranik: Do we need to add more tests to cover the different scenarios?

PR #2671, opened 2020-03-10
IoTEdgeDevOps: Add bug counts to database

and-rewsmith:
We need to start storing bug counts of certain shared queries in Azure Dev Ops. This will allow us to get an up to date picture of the stability of iotedge.

This is implemented by calling the Dev Ops [Wiql API](https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/wiql/get?view=azure-devops-rest-5.1) to execute a shared query, getting the number of bugs returned, then storing this information in our test dashboard database.

### DIFF
--- a/tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
+++ b/tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
@@ -2,13 +2,10 @@
 namespace DevOpsLib
 {
     using System;
-    using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
-    using DevOpsLib.VstsModels;
     using Flurl;
     using Flurl.Http;
-    using Newtonsoft.Json;
     using Newtonsoft.Json.Linq;
 
     public class BugManagement
@@ -26,18 +23,17 @@ namespace DevOpsLib
         /// This method is used to execute a Dev Ops work item query and get a list of bugs. 
         /// If result is not found for a query Id, it will return an entity with no result.
         /// Note: there is no validation of work item query ids.
-        /// Reference: https://docs.microsoft.com/en-us/rest/api/azure/devops/build/builds/list?view=azure-devops-rest-5.1
+        /// Reference: https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/wiql/get?view=azure-devops-rest-5.1
         /// </summary>
-        /// <param name="buildDefinitionIds">build definition Ids</param>
-        /// <param name="branchName">github repository branch name</param>
-        /// <returns>List of vsts build entities</returns>
-        public async Task<IList<VstsBuild>> GetLatestBuildsAsync(WorkItemQueryId workItemQueryId)
+        /// <param name="bugQueryId">bug query id from azure dev ops shared queries</param>
+        /// <returns>Number of bugs output by query</returns>
+        public async Task<int> GetBugsQuery(string bugQueryId)
         {
-            ValidationUtil.ThrowIfNullOrEmptySet(workItemQueryId, nameof(workItemQueryId));
+            ValidationUtil.ThrowIfNullOrEmptySet(bugQueryId, nameof(bugQueryId));
 
             // TODO: need to think about how to handle unexpected exception during REST API call
-            string requestPath = string.Format(WorkItemPathSegmentFormat, this.accessSetting.Organization, this.accessSetting.Project, this.accessSetting.Team, workItemQueryId);
-            IFlurlRequest latestBuildRequest = GetBugsRequestUri(workItemQueryId, requestPath)
+            string requestPath = string.Format(WorkItemPathSegmentFormat, this.accessSetting.Organization, this.accessSetting.Project, this.accessSetting.Team, bugQueryId);
+            IFlurlRequest latestBuildRequest = GetBugsRequestUri(requestPath)
                 .WithBasicAuth(string.Empty, this.accessSetting.PersonalAccessToken);
 
             string resultJson = await latestBuildRequest.GetStringAsync().ConfigureAwait(false);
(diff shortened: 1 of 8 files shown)


tools/IoTEdgeDevOps/DevOpsLib/BugQueryId.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace DevOpsLib
+{
+    public static class BugQueryId
+    {
+        public const string SampleQuery = "d7f4e036-0ccb-46a6-a33d-87a932296543";
+
+        public readonly static Dictionary<string, string> QueryNamestoIds = new Dictionary<string, string>() 
+        { 
+            {nameof(SampleQuery), SampleQuery  } 
+        };
```
and-rewsmith: What is the correct convention for an enum of strings? I looked it up, and this was a proposed solution, but I feel there could be a better way. The dictionary is here so that I can cleanly iterate through the properties, but I feel like there is a better way to do this. Repeating the property names in the dict init seems inefficient.

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
```diff
@@ -32,9 +32,21 @@ public async Task RunAsync(TimeSpan waitPeriodAfterEachUpdate, CancellationToken
...
+                try
+                {
+                    Console.WriteLine($"Import bugs data started at {DateTime.UtcNow}");
+
+                    await ImportVstsBugDataAsync(bugManagement);
+                }
+                catch (Exception ex)
+                {
+                    Console.WriteLine($"Unexcepted Exception: {ex}");
+                }
```
and-rewsmith: There is similar logic below and this loop is getting long. We need to refactor this at some point. If you have any suggestions lmk.

tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Reference: https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/wiql/get?view=azure-devops-rest-5.1
+        /// </summary>
+        /// <param name="bugQueryId">bug query id from azure dev ops shared queries</param>
+        /// <returns>Number of bugs output by query</returns>
+        public async Task<int> GetBugsQuery(string bugQueryId)
+        {
+            ValidationUtil.ThrowIfNullOrEmptySet(bugQueryId, nameof(bugQueryId));
+
+            // TODO: need to think about how to handle unexpected exception during REST API call
+            string requestPath = string.Format(WorkItemPathSegmentFormat, this.accessSetting.Organization, this.accessSetting.Project, this.accessSetting.Team, bugQueryId);
```
philipktlin: nit:use string interpolation

tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
```diff
@@ -0,0 +1,58 @@
...
+            IFlurlRequest latestBuildRequest = GetBugsRequestUri(requestPath)
```
philipktlin: rename "latestBuildRequest"

tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+            else
+            {
+                return result["workItems"].Count();
+            }
+        }
+
+        private static Url GetBugsRequestUri(string requestPath)
+        {
+            Url requestUri = DevOpsAccessSetting.BaseUrl.AppendPathSegment(requestPath);
```
philipktlin: why not just put baseUrl in WorkItemPathSegmentFormat ?

and-rewsmith: I think im going to eliminate this function and do what you suggested where I don't append path

tools/IoTEdgeDevOps/DevOpsLib/DevOpsAccessSetting.cs
```diff
@@ -7,26 +7,31 @@ public class DevOpsAccessSetting
...
         }
 
         public DevOpsAccessSetting(
             string organization,
             string project,
-            string personalAccessToken)
+            string personalAccessToken,
+            string team)
         {
             this.Organization = organization;
```
philipktlin: nit: help to add validationUtil calls.

tools/IoTEdgeDevOps/DevOpsLib/BugQueryId.cs
philipktlin: 1. You don't need to define SampleQuery constant, just define everything in the dictionary should be fine.
2. Do you have different query to find number of bugs of different priority? for example number of P1, P2 bugs?  Bug source from IcM or GitHub or VSTS?

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
philipktlin: Remove try-catch here and move Console.WriteLine inside ImportVstsBugDataAsync and similar to others.

tools/IoTEdgeDevOps/VstsPipelineSync/db/dbo.VstsBug.sql
```diff
@@ -0,0 +1,17 @@
+/****** Object:  Table [dbo].[VstsBuildRun]    Script Date: 2/12/2020 4:56:48 PM ******/
+SET ANSI_NULLS ON
+GO
+
+SET QUOTED_IDENTIFIER ON
+GO
+
+CREATE TABLE [dbo].[VstsBug](
+	[QueryName] [varchar](20) NOT NULL,
+	[BugCount] [int] NOT NULL,
```
philipktlin: you should create 2 more fields:
[InsertedAt] [datetime2](7) NOT NULL,
[UpdatedAt] [datetime2](7) NOT NULL,

tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// This method is used to execute a Dev Ops work item query and get a list of bugs. 
+        /// If result is not found for a query Id, it will return an entity with no result.
+        /// Note: there is no validation of work item query ids.
+        /// Reference: https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/wiql/get?view=azure-devops-rest-5.1
+        /// </summary>
+        /// <param name="bugQueryId">bug query id from azure dev ops shared queries</param>
+        /// <returns>Number of bugs output by query</returns>
+        public async Task<int> GetBugsQuery(string bugQueryId)
```
philipktlin: should it be Task<uint> or Task<ushort>?  if you are going to use unsigned, please make sure to use the right type in sql table, https://social.msdn.microsoft.com/Forums/en-US/ff08c190-a981-4896-9542-3f64b95a84a2/sql-server-data-type-for-signedunsigned-integral-c-types?forum=adodotnetdataproviders.

tools/IoTEdgeDevOps/DevOpsLib/IoTEdgeReleaseDeployment.cs
```diff
@@ -16,7 +16,7 @@ public class IoTEdgeReleaseDeployment : IEquatable<IoTEdgeReleaseDeployment>
 
         public IoTEdgeReleaseDeployment(int id, int attempt, VstsDeploymentStatus status, DateTime lastModifiedOn, HashSet<IoTEdgePipelineTask> tasks)
         {
-            ValidationUtil.ThrowIfNonPositive(id, nameof(id));
+            ValidationUtil.ThrowIfNegative(id, nameof(id));
```
and-rewsmith: Needed to change because we were getting a record with id 0. This seemed weird to me though. I logged the deployment that was trying to be created:
```
[{"id":0,"attempt":1,"deploymentStatus":0,"lastModifiedOn":"2020-03-09T18:37:54.4Z","releaseDeployPhases[0].deploymentJobs[0].tasks":null}]
```

tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
and-rewsmith: Since SQL server doesn't support uint I feel like it is cleanest to use the same data type everywhere. If you disagree I can change.

philipktlin: Not a big concern to me since this is a internal tool.

tools/IoTEdgeDevOps/VstsPipelineSync/db/dbo.UpsertVstsBug.sql
```diff
@@ -0,0 +1,21 @@
+CREATE PROCEDURE [dbo].[UpsertVstsBug]
+	@QueryName varchar(20),
+	@BugCount int
+AS
+    DECLARE @now datetime2;
+    SET @now = SYSDATETIME();
+
+	IF EXISTS (SELECT 1 FROM dbo.VstsBug WHERE QueryName = @QueryName)
```
philipktlin: nit: spacing issue.

PR #2678, opened 2020-03-11
IoTEdgeDevOps: Use dynamic query generation to retrieve bug counts

and-rewsmith:
I added template vsts work item queries (wiql) templates that I now edit at runtime and use to retrieve the desired bugs for different combinations of area, priority, and state.

Note that I am using the same API, but now I am performing POST requests which allows me to use custom queries.
https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/wiql/query%20by%20wiql?view=azure-devops-rest-5.1

I have tested and it correctly loads bug counts into the db.


tools/IoTEdgeDevOps/DevOpsLib/BugPriority.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace DevOpsLib
+{
+    public enum BugPriority
```
philipktlin: In DevOps, bug priority values can be 0-4.

tools/IoTEdgeDevOps/DevOpsLib/BuildExtension.cs
```diff
@@ -22,28 +22,27 @@ public static class BuildExtension
                 BuildDefinitionId.StressTestEnv2,
                 BuildDefinitionId.StressTestEnv3
             };
+        static Dictionary<BuildDefinitionId, string> definitionIdToDisplayNameMapping = new Dictionary<BuildDefinitionId, string>
```
philipktlin: nit: new line

tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
```diff
@@ -21,23 +21,35 @@ public BugManagement(DevOpsAccessSetting accessSetting)
...
-        /// Note: there is no validation of work item query ids.
-        /// Reference: https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/wiql/get?view=azure-devops-rest-5.1
+        /// If result is not found for a query, it will return 0.
+        /// Reference: https://docs.microsoft.com/en-us/rest/api/azure/devops/wit/wiql/query%20by%20wiql?view=azure-devops-rest-5.1
         /// </summary>
-        /// <param name="bugQueryId">bug query id from azure dev ops shared queries</param>
+        /// <param name="bugQuery">Bug query object representing vsts shared queries</param>
         /// <returns>Number of bugs output by query</returns>
-        public async Task<int> GetBugsQuery(string bugQueryId)
+        public async Task<int> GetBugsQuery(BugQuery bugQuery)
```
philipktlin: This method name should be GetBugCount.

tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
```diff
@@ -21,23 +21,35 @@ public BugManagement(DevOpsAccessSetting accessSetting)
...
+                
+                result = await response.GetJsonAsync<JObject>();
+            }
+            catch (FlurlHttpException e)
+            {
+                Console.WriteLine($"Failed making call to vsts work item api: {e.Message}");
+                Console.WriteLine(e.Call.RequestBody);
+                Console.WriteLine(e.Call.Response.StatusCode);
+                Console.WriteLine(e.Call.Response.ResponseMessage);
+                result = JObject.Parse("{}");
```
philipktlin: return 0 here.

tools/IoTEdgeDevOps/DevOpsLib/BugManagement.cs
```diff
@@ -21,23 +21,35 @@ public BugManagement(DevOpsAccessSetting accessSetting)
...
+            }
 
             if (!result.ContainsKey("queryType"))
             {
```
philipktlin: remove else keyword and return directly.

tools/IoTEdgeDevOps/DevOpsLib/BugPriorityExtension.cs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace DevOpsLib
+{
+    using System.Collections.Generic;
+
+    public static class BugPriorityExtension
+    {
+        public static HashSet<BugPriority> BugPriorities =>
```
philipktlin: I think this can be improved by using a single dictionary.

Dictionary<string, string> bugPriorityGroups = new Dictionary<string, string>{
    { "1", "Pri-1" },
    { "2", "Pri-2" },
    { "Other", "Pri-Other" }
}

For any usage of BugPriorities, use bugPriorityGroups.Keys.
For any usage of PriorityValue method, use key of dictionary in the loop.
For any usage of DisplayName method, use bugPriorityGroups[<key>].

Once simplified, remove this class and define this single dictionary in BugQueryGenerator

tools/IoTEdgeDevOps/DevOpsLib/BugPriority.cs
philipktlin: Create a class
1. private constructor with name and value properties.
2. create 3 static variables for each priority (one, two, other).
3. create a static method to return IEnumerable of these 3 static variables.
4. remove BugPriorityExtension

tools/IoTEdgeDevOps/DevOpsLib/BugQuery.cs
```diff
@@ -0,0 +1,55 @@
...
+        public string Title { get; }
+        public string Area { get; }
+        public BugPriority Priority { get; }
+        public bool InProgress { get; }
+
+        public string GetWiqlFromConfiguration()
+        {
+            string query = "";
+
+            if (this.Priority != BugPriority.Other && !this.InProgress)
```
philipktlin: Define a data collection to store the mapping to get BugWiqlQueries string and replace this if-else chain block.
1. Dictionary<BugPriority, Dictionary<bool, string>> or
2. Dictionary<Tuple<BugPriority,bool>, string>

tools/IoTEdgeDevOps/DevOpsLib/BugQuery.cs
```diff
@@ -0,0 +1,55 @@
...
+            ValidationUtil.ThrowIfNullOrWhiteSpace(title, nameof(title));
+            ValidationUtil.ThrowIfNullOrWhiteSpace(area, nameof(area));
+            ValidationUtil.ThrowIfNull(inProgress, nameof(inProgress));
+
+            this.Title = title;
+            this.Area = area;
+            this.Priority = priority;
+            this.InProgress = inProgress;
+        }
+        public string Title { get; }
```
philipktlin: I think the title should be derived inside this class, not passing in.

tools/IoTEdgeDevOps/DevOpsLib/BugQuery.cs
and-rewsmith: I feel like this would work if we weren't grouping multiple different <Priority, InProgress> states to the same wiql template. Doing it the way you described means we have to enumerate every combination. With the if else chain, multiple different configs can evaluate to the same template.

tools/IoTEdgeDevOps/DevOpsLib/BugQuery.cs
and-rewsmith: I used computed properties to achieve this

tools/IoTEdgeDevOps/DevOpsLib/BugPriority.cs
and-rewsmith: I did it this way to avoid comparing non-enum types in the if else chain below.

tools/IoTEdgeDevOps/DevOpsLib/BugPriorityGrouping.cs
```diff
@@ -0,0 +1,29 @@
...
+            ValidationUtil.ThrowIfNullOrWhiteSpace(priority, nameof(priority));
+
+            this.Name = name;
+            this.Priority = priority;
+        }
+
+        public string Name { get; }
+        public string Priority { get; }
+
+        public static readonly BugPriorityGrouping Pri1 = new BugPriorityGrouping("Pri-1", "1");
```
philipktlin: internal?

tools/IoTEdgeDevOps/DevOpsLib/BugPriorityGrouping.cs
```diff
@@ -0,0 +1,29 @@
...
+        public static readonly BugPriorityGrouping Pri2 = new BugPriorityGrouping("Pri-2", "2");
+        public static readonly BugPriorityGrouping PriOther = new BugPriorityGrouping("Pri-Other", "Other");
+
+        public bool Equals(BugPriorityGrouping bugPriorityGrouping)
+        {
+            return this.Name == bugPriorityGrouping.Name && this.Priority == bugPriorityGrouping.Priority;
```
philipktlin: uses Equals method with string comparison option.

tools/IoTEdgeDevOps/DevOpsLib/BugQuery.cs
```diff
@@ -0,0 +1,72 @@
...
+                }
+            }
+        }
+
+        public string GetWiqlFromConfiguration()
+        {
+            string query;
+            string unassignedPriorityPlaceholder = "Other";
+
+            if (this.BugPriorityGrouping.Priority != unassignedPriorityPlaceholder && !this.InProgress)
```
philipktlin: you can define a data collection type to store this mapping.
1. uses Dictionary<string, Dictionary<string, string>>, or
2. uses Dictionary<Tuple<string,string>, string>

Then use TryGetValue if value doesn't exist, throw exception.

tools/IoTEdgeDevOps/DevOpsLib/BugWiqlQueries.cs
```diff
@@ -0,0 +1,79 @@
...
+FROM workitems
+WHERE
+    [System.TeamProject] = @project
+    AND [System.WorkItemType] = 'Bug'
+    AND [System.AreaPath] = 'One\IoT\Platform\{AREA}'
+    AND NOT [System.State] CONTAINS 'Done'
+    AND NOT [System.State] CONTAINS 'Removed'
+    AND [Microsoft.VSTS.Common.Priority] = {PRIORITY}";
+
+        public const string PrioritizedStartedBugTemplate =
```
philipktlin: define PrioritizedStartedBugTemplate by using PrioritizedBugTemplate

tools/IoTEdgeDevOps/DevOpsLib/BugWiqlQueries.cs
```diff
@@ -0,0 +1,79 @@
...
+    [System.State],
+    [System.Tags]
+FROM workitems
+WHERE
+    [System.TeamProject] = @project
+    AND [System.WorkItemType] = 'Bug'
+    AND [System.AreaPath] = 'One\IoT\Platform\{AREA}'
+    AND NOT [System.State] CONTAINS 'Done'
+    AND NOT [System.State] CONTAINS 'Removed'
+    AND NOT [Microsoft.VSTS.Common.Priority] IN (1)
```
philipktlin: This is not modeled in a good way, but I think it is ok for now since this is an internal tool with time constraint

tools/IoTEdgeDevOps/VstsPipelineSync/VstsBuildBatchUpdate.cs
```diff
@@ -152,7 +152,10 @@ void UpsertVstsBugToDb(SqlConnection sqlConnection, string queryName, int bugCou
                 CommandText = "UpsertVstsBug"
             };
 
-            cmd.Parameters.Add(new SqlParameter("@QueryName", queryName));
+            cmd.Parameters.Add(new SqlParameter("@Title", bugQuery.Title));
+            cmd.Parameters.Add(new SqlParameter("@AreaPath", bugQuery.Area));
+            cmd.Parameters.Add(new SqlParameter("@Priority", bugQuery.BugPriorityGrouping.Priority.ToString()));
```
philipktlin: remove ToString()

tools/IoTEdgeDevOps/VstsPipelineSync/db/dbo.VstsBug.sql
```diff
@@ -6,14 +6,16 @@ SET QUOTED_IDENTIFIER ON
...
+	[AreaPath] [varchar](200) NOT NULL,
+	[Priority] [varchar](20) NOT NULL,
+	[InProgress] [bit] NOT NULL,
 	[BugCount] [int] NOT NULL,
 	[InsertedAt] [datetime2](7) NOT NULL,
 	[UpdatedAt] [datetime2](7) NOT NULL,
  CONSTRAINT [PK_QueryName] PRIMARY KEY CLUSTERED 
 (
-	[QueryName] ASC
+	[Title] ASC
```
philipktlin: Suggestion: add a datetime field as part of primary key; therefore this table will keep number of bugs for every day.  This may help to answer more question later.  If this is too big to change, forget it.

tools/IoTEdgeDevOps/DevOpsLib/BugWiqlQueries.cs
and-rewsmith: I intentionally didn't do this because I thought continuing this practice as we change / add wiql queries would be error prone and less readable. I think it is fine either way though so I'll change.

tools/IoTEdgeDevOps/DevOpsLib/BugWiqlQueries.cs
and-rewsmith: Do you think this because of the existence of other possible bug states?

tools/IoTEdgeDevOps/DevOpsLib/BugQuery.cs
and-rewsmith: I don't think this is a good idea because then we have to enumerate all combinations, where if/else chain can capture multiple combinations with the same logic.

tools/IoTEdgeDevOps/DevOpsLib/BugWiqlQueries.cs
and-rewsmith: On second thought, I'm scared that implementing it this way will make it harder to work with. I think it best to not make this change. Do you still disagree?

tools/IoTEdgeDevOps/VstsPipelineSync/db/dbo.VstsBug.sql
and-rewsmith: I feel it's best to wait until we have a plan for what to do with this historical data. Maybe display a graph of bugs over time? I'd have to think of how best to display this because lots of areas/categories.

tools/IoTEdgeDevOps/DevOpsLib/BugQuery.cs
philipktlin: you don't need to enumerate all combinations.  For example Dictionary<tuple<string,bool>, string>

if-eles chain block can be replaced by:

if (!this.Dict.TryGetValue(Tuple.Create(this.BugPriorityGrouping.Priority, this.InProgress), out string query) {
    throw new NotImplementedException();
}

query = query.Replace("{PRIORITY}", this.BugPriorityGrouping.Priority.ToString());
query = query.Replace("{AREA}", this.Area);

is it clear to you?

tools/IoTEdgeDevOps/DevOpsLib/BugPriorityGrouping.cs
```diff
@@ -0,0 +1,29 @@
...
+            ValidationUtil.ThrowIfNullOrWhiteSpace(priority, nameof(priority));
+
+            this.Name = name;
+            this.Priority = priority;
+        }
+
+        public string Name { get; }
+        public string Priority { get; }
+
+        internal static readonly BugPriorityGrouping Pri1 = new BugPriorityGrouping("Pri-1", "1");
```
philipktlin: nit: move before constructor

tools/IoTEdgeDevOps/DevOpsLib/BugWiqlQueries.cs
philipktlin: not a big deal to me.

tools/IoTEdgeDevOps/DevOpsLib/BugWiqlQueries.cs
philipktlin: No as this static class is to provide format of query; the actual value should not be in there; same as you will not create many of this query format for different area; which area should be pass-in.

tools/IoTEdgeDevOps/DevOpsLib/BugQuery.cs
and-rewsmith: This part made sense to me, but when making the dict we will have to enumerate all the mappings, right? Might be missing something obvious, so apologies if I am 😅.

philipktlin: Ok.  Got it because current if-else block is not just use equals, but not equals; therefore need to define more than 4 entry in the dictionary.  my fault I missed that.

Why not nested the if statement? for example you can nested 1st and 2nd if block, right?

PR #2691, opened 2020-03-12
Adding timeout for device and moduleClients

dylanbronson:
Adding timeout on OpenAsync for device and module wrapper clients. 

We are seeing sometimes that OpenAsync hangs forever. This is an SDK issue, but by adding a timeout here, we don't need to rely on the SDK making sure to come back. This way we force a timeout from our side. Even if the SDK hangs forever, we will timeout on this call.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceClientWrapper.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceClientWrapper.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
         {
             try
             {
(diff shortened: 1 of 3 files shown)


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceClientWrapper.cs
```diff
@@ -48,7 +48,7 @@ public async Task OpenAsync()
         {
             try
             {
-                await this.underlyingDeviceClient.OpenAsync();
+                await this.underlyingDeviceClient.OpenAsync().TimeoutAfter(TimeSpan.FromMilliseconds(this.underlyingDeviceClient.OperationTimeoutInMilliseconds));
```
varunpuranik: We should move it one level up and add tests for it

PR #2698, opened 2020-03-16
Longhaul/Stress: Consolidate logic for metrics/test-report upload

and-rewsmith:
This PR consolidates the test-context data upload logic for both metric upload and test reporting. 

For test reporting, we use a custom key value string format to store the test context. This is desirable over the way we handle this for metrics because it decouples the test information from the module logic.

Therefore, let's follow the same convention when uploading test-context with metrics. That being said, the metrics collector module is intended to be public, so at some point we should clean up the implementation of this custom format. Once this PR is merged I will create a work item to track this cleanup task.

**Metrics collector behavior before**
Parses key value pairs from root level desired properties and appends these to each scraped metric

**Metrics collector behavior after**
Reads custom formatted string from field "additionalTags" at root level desired properties, then parses each individual key value pair. Each key value pair will be appended to  each scraped metric.


builds/e2e/templates/deploy-iotuap.yaml
```diff
@@ -109,6 +109,7 @@ steps:
         $testInfo=$testInfo+",ImagesSourceBranch=${{ parameters['images.source.branch'] }}"
         $testInfo=$testInfo+",HostPlatform=${{ parameters['metricsCollector.hostPlatform'] }}"
         $testInfo=$testInfo+",TestName=${{ parameters['testName'] }}"
+        $testInfo=$testInfo+",UpstreamProtocol=amqp"
```
varunpuranik: WHy do we need this?

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -731,7 +731,7 @@
                 "value": "<LogAnalyticsSharedKey>"
               },
               "LogAnalyticsLogType": {
-                "value": "testMetrics"
+                "value": "testMetricsV2"
```
varunpuranik: Again, why v2?

and-rewsmith: Log types are like tables. I am changing the schema of the metric tags so I thought it best to create a new table. This also helped to make the downstream changes to the metrics dashboard / alerts, as they didn't immediately break with differently shaped data going into the existing table.

builds/e2e/templates/deploy-iotuap.yaml
and-rewsmith: TestInfo is now used for both metrics upload and test result reporting. I don't beleive it is currently used for metrics, but since the upstream protocol is configurable then we should know about the setting in the test report.

edge-modules/MetricsCollector/Program.cs
```diff
@@ -32,7 +33,7 @@ static async Task<int> MainAsync()
             try
             {
                 moduleClient = await ModuleClient.CreateFromEnvironmentAsync(transportSettings);
-                Dictionary<string, string> additionalTags = await GetAdditionalTagsFromTwin(moduleClient);
+                Option<SortedDictionary<string, string>> additionalTags = await GetAdditionalTagsFromTwin(moduleClient);
```
philipktlin: I think it is ok to return empty SortedDictionary, right?

scripts/linux/runE2ETest.sh
```diff
@@ -169,12 +169,10 @@ function prepare_test_from_artifacts() {
 
                 local escapedSnitchAlertUrl
                 local escapedBuildId
-                local tracking_id=$(cat /proc/sys/kernel/random/uuid)
```
philipktlin: why move this? tracking id should only be used for long haul and stress test, right? that's why this line is put in this section.  I think adding tracking id to TESTINFO should be put here as well.

philipktlin: Same in Run-E2ETest.ps1

scripts/windows/test/Run-E2ETest.ps1
```diff
@@ -1625,6 +1617,9 @@ Else
     $BypassInstallationFlag = $null
 }
 
+$TrackingId = New-Guid
+$TestInfo=$TestInfo+"TestId,=$TrackingId"
```
philipktlin:
```suggestion
$TestInfo=$TestInfo+"TrackingId=$TrackingId"
```

philipktlin: And btw why the key is "TestId", not "TrackingId"?
```suggestion
$TestInfo=$TestInfo+",TestId=$TrackingId"
```

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -235,43 +230,34 @@ function process_args() {
...
-            saveNextArg=0
-        elif [ $saveNextArg -eq 36 ]; then
-            CUSTOM_EDGE_AGENT_IMAGE="$arg"
-            saveNextArg=0
-        elif [ $saveNextArg -eq 37 ]; then
-            CUSTOM_EDGE_HUB_IMAGE="$arg"
+            TEST_INFO="$arg"
             saveNextArg=0
         elif [ $saveNextArg -eq 38 ]; then
             TEST_INFO="$arg"
```
philipktlin: Duplicated, please remove it.

test/modules/ModuleLib/ModuleUtil.cs
```diff
@@ -127,9 +127,11 @@ static void WriteLog(ILogger logger, LogLevel logLevel, string message)
...
-        public static SortedDictionary<string, string> ParseTestInfo(string testInfo)
+        public static SortedDictionary<string, string> ParseKeyValuePairs(string keyValuePairs, ILogger logger)
         {
-            Dictionary<string, string> unsortedParsedTestInfo = testInfo.Split(",", StringSplitOptions.RemoveEmptyEntries)
+            WriteLog(logger, LogLevel.Information, $"Parsing key value pairs: {keyValuePairs}");
+
+            Dictionary<string, string> unsortedParsedTestInfo = keyValuePairs.Split(",", StringSplitOptions.RemoveEmptyEntries)
                                 .Select(x => (KeyAndValue: x, SplitIndex: x.IndexOf('=')))
                                 .Where(x => x.SplitIndex >= 1)
                                 .ToDictionary(
```
philipktlin: Since this method become more generic, I suggest to remove "CheckArgument" call from this function.  Let's the caller to decide if empty collection is an invalid input or not.

scripts/linux/runE2ETest.sh
and-rewsmith: This is moved because we need to add this field to TestInfo. I think the cleanest place to do that is with all the default value setup near EOF after process arguments. Let me know if you disagree.

edge-modules/MetricsCollector/Program.cs
and-rewsmith: I thought using Option was better because then one would have to check the contents of the collection, which is a little less obvious. Option lets the reader know the intentions all the way through the code. Let me know if you disagree.

scripts/linux/runE2ETest.sh
philipktlin: my point is, we should not run any line of code if it is not needed.  Tracking id is only used for long haul and stress test, we should not run that line if it is running other e2e test.  I can see 2 options; 1) keep tracking id assignment at EOF but add if-statement to run only for stress and long haul; 2) move tracking id and testinfo assigment back to here. Or you have another idea.

and-rewsmith: Oh this is rly good point. I think I prefer 1)

scripts/windows/test/Run-E2ETest.ps1
```diff
@@ -1588,6 +1588,12 @@ If ([string]::IsNullOrWhiteSpace($MetricsUploadTarget))
     $MetricsScrapeFrequencyInSecs="AzureLogAnalytics";
 }
 
+If ($TestName -eq "LongHaul" -Or $TestName -eq "Stress")
```
and-rewsmith: I followed the existing convention, however we use double comma syntax for our linux shell script in order to check the lower cased string, so we aren't being consistent.

scripts/linux/runE2ETest.sh
```diff
@@ -1085,6 +1073,12 @@ TRANSPORT_TYPE_4="${TRANSPORT_TYPE_4:-mqtt}"
 TWIN_UPDATE_FAILURE_THRESHOLD="${TWIN_UPDATE_FAILURE_THRESHOLD:-00:01:00}"
 METRICS_SCRAPE_FREQUENCY_IN_SECS="${METRICS_SCRAPE_FREQUENCY_IN_SECS:-300}"
 METRICS_UPLOAD_TARGET="${METRICS_UPLOAD_TARGET:-AzureLogAnalytics}"
+
+if [[ "${TEST_NAME,,}" == "longhaul" ]] ||
+   [[ "${TEST_NAME,,}" == "stress"   ]]; then
```
philipktlin:
```suggestion
   [[ "${TEST_NAME,,}" == "stress" ]]; then
```

test/modules/ModuleLib/ModuleUtil.cs
```diff
@@ -127,16 +127,21 @@ static void WriteLog(ILogger logger, LogLevel logLevel, string message)
...
                                     x => x.KeyAndValue.Substring(0, x.SplitIndex),
                                     x => x.KeyAndValue.Substring(x.SplitIndex + 1, x.KeyAndValue.Length - x.SplitIndex - 1));
 
-            Preconditions.CheckArgument(unsortedParsedTestInfo.Count > 0);
+            if (shouldBeNonEmpty)
+            {
+                Preconditions.CheckArgument(unsortedParsedTestInfo.Count > 0);
```
philipktlin: add message

edge-modules/MetricsCollector/MetricsCollector.csproj
```diff
@@ -31,6 +31,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\edge-util\src\Microsoft.Azure.Devices.Edge.Util\Microsoft.Azure.Devices.Edge.Util.csproj" />
+    <ProjectReference Include="..\..\test\modules\ModuleLib\Microsoft.Azure.Devices.Edge.ModuleUtil.csproj" />
```
philipktlin: FYI, once this project is officially released like tempSensor; we need to refactor this dependency.

scripts/linux/runE2ETest.sh
```diff
@@ -1085,6 +1073,12 @@ TRANSPORT_TYPE_4="${TRANSPORT_TYPE_4:-mqtt}"
 TWIN_UPDATE_FAILURE_THRESHOLD="${TWIN_UPDATE_FAILURE_THRESHOLD:-00:01:00}"
 METRICS_SCRAPE_FREQUENCY_IN_SECS="${METRICS_SCRAPE_FREQUENCY_IN_SECS:-300}"
 METRICS_UPLOAD_TARGET="${METRICS_UPLOAD_TARGET:-AzureLogAnalytics}"
+
+if [[ "${TEST_NAME,,}" == "longhaul" ]] ||
+   [[ "${TEST_NAME,,}" == "stress"  ]]; then
+    tracking_id=$(cat /proc/sys/kernel/random/uuid)
```
philipktlin: The rule we agreed is that TestInfo should be updated in yaml only except device id.  For tracking id, your call, either leave here or move to yaml.

edge-modules/MetricsCollector/MetricsCollector.csproj
and-rewsmith: I talked with Venkat and we are never going to do this, because we want Azure Monitor to own it.

scripts/linux/runE2ETest.sh
and-rewsmith: I'll open a separate PR

PR #2708, opened 2020-03-18
Iotedged Checkin: Large file download timeout increase / Disk space fix

and-rewsmith:
We saw two errors from the iotedged checkin. 

Firstly, it would timeout when downloading this ~250MB file. This can be solved by adding a timeout of 5 minutes.

Secondly, the agent would crash and run out of disk space caused by a rust update. We can fix this by using compression in our zip command, as well as cleaning up the folder we use for cargo build/test.


builds/checkin/edgelet.yaml
```diff
@@ -181,10 +181,13 @@ jobs:
       - script: $CARGO test --verbose
         displayName: Test
         workingDirectory: edgelet
+      - script: find ./target -type f ! -name '*.gc' -delete
```
darobs: -name '*.gc*'

PR #2710, opened 2020-03-18
Is this a disk space problem?

darobs:


and-rewsmith:
Should we log the disk space at all steps?

and-rewsmith:
Now that this worked, we can delete the logging and check it in.

and-rewsmith:
Yesterday there was also a timeout in the windows_amd64 job when downloading a large file. I added a longer timeout to get around this. Might be a good idea to take this change here?
https://github.com/Azure/iotedge/pull/2708

and-rewsmith:
Merged into master with #2708

PR #2712, opened 2020-03-19
Connectivity for ARM32

yophilav:


test/connectivity/scripts/connectivityTest.sh
```diff
@@ -354,20 +354,44 @@ function run_connectivity_test() {
...
-        -u "$CONTAINER_REGISTRY_USERNAME" \
-        -p "$CONTAINER_REGISTRY_PASSWORD" \
-        -n "$(hostname)" \
-        -t "$ARTIFACT_IMAGE_BUILD_NUMBER-linux-$image_architecture_label" \
-        --leave-running=All \
-        -l "$deployment_working_file" \
-        --runtime-log-level "Debug" \
-        --no-verify && funcRet=$? || funcRet=$?
+
+    case $(hostname) in
```
philipktlin: you can't use hostplatform as Andrew has another PR to remove it from this script.  I think you can use image_architecture_label instead.

PR #2714, opened 2020-03-19
Test Dashboard: Add network controller description to test info

and-rewsmith:
Test info currently contains run profile, however this is not specific enough. We can add a variable to describe the exact network scenario and add to test info.

If unfamiliar, test info is used to describe the context of a test and gets uploaded with the report for connectivity/longhaul/stress.

### DIFF
--- a/builds/e2e/connectivity.yaml
+++ b/builds/e2e/connectivity.yaml
@@ -16,51 +16,51 @@ jobs:
       matrix:
         Online:
           testrun.config: 1
-          testrun.network.runProfile: "Online"
           testrun.network.description: "Online"
           testrun.network.frequencies: "00:00:00 00:00:00 0"
+          testrun.network.runProfile: "Online"
           testrun.duration: "01:00:00"
         Offline(5/5/5):
           testrun.config: 2
+          testrun.network.description: "Offline(5/5/5)"
           testrun.network.frequencies: "00:05:00 00:05:00 5"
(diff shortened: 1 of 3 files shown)


builds/e2e/connectivity.yaml
```diff
@@ -286,6 +303,7 @@ jobs:
           testStartDelay: '$(testStartDelay.arm32)'
           networkController.frequencies: '$(testrun.network.frequencies)'
           networkController.runProfile: '$(testrun.network.runProfile)'
+          networkController.description: '$(testrun.network.description)'
```
philipktlin: Since network controller frequencies and run profile is passed to module for test run and those information already reported in TestInfo in https://github.com/Azure/iotedge/blob/cf39ecd6298892533d6525520dac6a0e07016cb0/test/connectivity/modules/NetworkController/Program.cs#L84

for description of network controller, it is just used in Test dashboard;  therefore this change should be in Test dashboard.  Also for any information already passed to module, it should let the module to report to TRC using TestInfoResult like what did in NetworkController.  After review, testBuildNumber should not be defined in TestInfo; as it is already well-defined in TestSummary which test dashboard logic should extract from there.  Similarly for any information passed to connectivityTest.sh, those field should append to TestInfo inside connectivityTest.sh instead of in connectivity-deploy.yaml, for example, customEdgeAgentImage.

Let me know if you want to chat about this.

PR #2716, opened 2020-03-20
[k8s] Prepare charts for beta7

darobs:
Update helm chart version, update image versions to beta7


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/KubernetesModule.cs
```diff
@@ -33,7 +33,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Service.Modules
 
     public class KubernetesModule : Module
     {
-        static readonly TimeSpan SystemInfoTimeout = TimeSpan.FromSeconds(3);
+        static readonly TimeSpan SystemInfoTimeout = TimeSpan.FromSeconds(120);
```
CindyXing: Would there be any side effect by extending to 2 minutes?

darobs: Originally, we had set this to a very low timeout to fail fast, and quit edgeAgent if we could not connect to iotedged. Then we put in the main Program wait to connect to iotedged, rather than failing fast, but we didn't bother to change this timeout.  This timeout can be longer, because we've already established we can communicate to iotedged when this runs.

CindyXing: Great! Thanks Dave.

PR #2718, opened 2020-03-20
Increase dotnet version to 3.1

and-rewsmith:
What this PR accomplishes:
- Increases build to 3.1
- Alters build / test pipelines to be compatible with dotnet 3.1 increase


and-rewsmith:
This is outdated. There is a new PR out

PR #2719, opened 2020-03-20
Connectivity test on ARM32 

yophilav:
- Introduce a flag to disable log upload to connectivity test


and-rewsmith:
When this is ready for review, ping me and I can look more into this upload problem. Dylan and I do not think that it is because of manual upload.

yophilav:
This is only a dev branch. What do you think is the issue?

e2e_deployment_files/connectivity_deployment.arm32.template.json
```diff
@@ -0,0 +1,823 @@
...
+        "testType": "connectivity",
+        "imagesBranchName": "<MetricsCollector.Images.BranchName>",
+        "edgeletBranchName": "<MetricsCollector.Edgelet.BranchName>",
+        "testId": "<TrackingId>",
+        "testBuildNumber": "<Test.BuildNumber>",
+        "upstreamProtocol": "<UpstreamProtocol>",
+        "networkControllerOfflineFrequency": "<NetworkController.OfflineFrequency0>",
+        "networkControllerOnlineFrequency": "<NetworkController.OnlineFrequency0>",
+        "networkControllerRunsCount": "<NetworkController.RunsCount0>",
+        "testDuration": "<TestDuration>"
```
and-rewsmith: Most of these are no longer needed

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -499,7 +531,7 @@ function usage() {
     echo ' -storageAccountConnectionString          Azure storage account connection string with privilege to create blob container.'
     echo ' -edgeRuntimeBuildNumber                  Build number for specifying edge runtime (edgeHub and edgeAgent)'
     echo ' -testInfo                                Contains comma delimiter test information, e.g. build number and id, source branches of build, edgelet and images.'
-
+    echo ' -testPlatform                            Test agent platform'
```
philipktlin: You don't need test platform parameter as you can use $image_architecture_label variable in connectivity script to decide "optimize_performance_flag".

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -334,20 +342,44 @@ function run_connectivity_test() {
...
+                -c "$IOT_HUB_CONNECTION_STRING" \
+                -e "$EVENTHUB_CONNECTION_STRING" \
+                -r "$CONTAINER_REGISTRY" \
+                -u "$CONTAINER_REGISTRY_USERNAME" \
+                -p "$CONTAINER_REGISTRY_PASSWORD" \
+                -n "$(hostname)" \
+                -t "$ARTIFACT_IMAGE_BUILD_NUMBER-linux-$image_architecture_label" \
+                --leave-running=All \
+                -l "$deployment_working_file" \
+                --runtime-log-level "Debug" \
```
philipktlin: try not to duplicate and use "--optimize_for_performance=true"

e2e_deployment_files/connectivity_deployment.arm32.template.json
```diff
@@ -0,0 +1,771 @@
...
+            }
+          },
+          "edgeHub": {
+            "type": "docker",
+            "settings": {
+              "image": "<Container_Registry>/microsoft/azureiotedge-hub:<EdgeRuntime.BuildNumber>-linux-<Architecture>",
+              "createOptions": "{\"ExposedPorts\": {\"9600/tcp\": {}}, \"HostConfig\": {\"PortBindings\": {\"8883/tcp\": [{\"HostPort\": \"8883\"}],\"443/tcp\": [{\"HostPort\": \"443\"}],\"5671/tcp\": [{\"HostPort\": \"5671\"}],\"9600/tcp\": [{\"HostPort\": \"9601\"}]}}}"
+            },
+            "env": {
+              "CollectMetrics": {
```
philipktlin: Why do you need another deployment file for arm32?  If we are going to have arm32 deployment file, you can define "optimize_performance_flag" here and no need to change connectivity.sh (90% sure).  But use different deployment json will introduce overhead when changes in test modules need to update in 2 places.  Try to use same deployment file first if possible.

yophilav: Oh you're right. I removed a lot of my debugging feature out. Now they are basically the same. Let me fix it.

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -298,7 +299,11 @@ function process_args() {
                 '-testInfo' ) saveNextArg=34;;
                 '-waitForTestComplete' ) WAIT_FOR_TEST_COMPLETE=1;;
                 '-cleanAll' ) CLEAN_ALL=1;;
-                * ) usage;;
+                
+                * ) 
+                    echo "Failed argument: $saveNextArg $arg"
```
philipktlin:
```suggestion
                    echo "unsupported argument: $saveNextArg $arg"
```

test/modules/TestResultCoordinator/Services/TestResultReportingService.cs
```diff
@@ -71,15 +71,18 @@ async void DoWorkAsync(object state)
...
+            if (Settings.Current.LogUploadEnabled)
             {
-                Uri blobContainerWriteUriForLog = await TestReportUtil.GetOrCreateBlobContainerSasUriForLogAsync(Settings.Current.StorageAccountConnectionString);
-                blobContainerUri = $"{blobContainerWriteUriForLog.Scheme}{Uri.SchemeDelimiter}{blobContainerWriteUriForLog.Authority}{blobContainerWriteUriForLog.AbsolutePath}";
-                await TestReportUtil.UploadLogsAsync(Settings.Current.IoTHubConnectionString, blobContainerWriteUriForLog, this.logger);
-            }
-            catch (Exception ex)
-            {
-                this.logger.LogError(ex, "Exception happened when uploading logs");
+                try
```
philipktlin: add log to indicate log upload is run

yophilav: `TestReportUtil.UploadLogsAsync()` already prints out `Send upload logs request to edgeAgent.` I think that should be sufficient log indicator.

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -298,7 +299,11 @@ function process_args() {
                 '-testInfo' ) saveNextArg=34;;
                 '-waitForTestComplete' ) WAIT_FOR_TEST_COMPLETE=1;;
                 '-cleanAll' ) CLEAN_ALL=1;;
-                * ) usage;;
+                
+                * ) 
+                    echo "Unsupported  argument: $saveNextArg $arg"
```
philipktlin:
```suggestion
                    echo "Unsupported argument: $saveNextArg $arg"
```

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -334,6 +339,19 @@ function run_connectivity_test() {
     print_highlighted_message "Run connectivity test with -d '$device_id' started at $test_start_time"
 
     SECONDS=0
+
+    case $image_architecture_label in
```
philipktlin: optimize_for_performance variable already exists in line 554; please move/merge this block to line 552.  Also please use lower case for internal variable (since it is not passing in).

philipktlin:
Please test both amd64 and arm32 for connectivity test before commit.  thanks.

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -346,7 +352,8 @@ function run_connectivity_test() {
         -t "$ARTIFACT_IMAGE_BUILD_NUMBER-linux-$image_architecture_label" \
         --leave-running=All \
         -l "$deployment_working_file" \
-        --runtime-log-level "Debug" \
+        --runtime-log-level "Info" \
+        --optimize_for_performance=$optimize_for_performance \
```
and-rewsmith: Did you ever figure out why the edgehub optimize for performance was being set to false when, according to the quickstart implementation, this flag should not do anything?

(Context: this flag is ignored if there is a deployment file.)

yophilav: In the Dockerfiles for ARM32v7 and ARM64v8, they already set the `ENV OptimizeForPerformance false` by default. I forgot to remove the line.

e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -370,6 +370,9 @@
               "IOT_HUB_CONNECTION_STRING": {
                 "value": "<IoTHubConnectionString>"
               },
+              "logUploadEnabled": {
```
philipktlin: Need to add optimize performance flag here and replace value by script.

yophilav: No I don't. Please see the comment above.

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -68,6 +68,7 @@ function prepare_test_from_artifacts() {
     sed -i -e "s@<TestResultCoordinator.VerificationDelay>@$VERIFICATION_DELAY@g" "$deployment_working_file"
     sed -i -e "s@<TestResultCoordinator.OptimizeForPerformance>@$optimize_for_performance@g" "$deployment_working_file"
```
philipktlin: Please create a bug/pbi to track to remove OptimizeForPerformance flag and we should add it in TRC arm32 docker file; same way as EdgeAgent and EdgeHub.

yophilav: Do we have a PBI that is a pool for connectivity test framework improvement?

and-rewsmith: @philipktlin Shouldn't the optimize for performance flag always be set to false for the TRC? My understanding was that these additional logs are not useful.

e2e_deployment_files/connectivity_deployment.template.json
varunpuranik: Why do we not have the OptimizeForPerformance flag?

PR #2720, opened 2020-03-20
Option 2: Use DeserializeSeed

lfitchett:


mqtt/mqtt3/src/proto/deserialize.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl PayloadDeDuper {
+    fn new() -> Self {
+        Self {
+            loaded_bytes: HashMap::new(),
+        }
+    }
+
+    fn de_dupe(&mut self, bytes: Vec<u8>) -> Bytes {
+        let hash = Self::calculate_hash(&bytes);
```
dmolokanov: why not just delegate it to hashmap itself?

dmolokanov: we propably can get away with HashSet
```rust
if let Some(bytes) = loaded_bytes.get(&bytes) {
  bytes.clone()
} else {
  loaded_bytes.insert(bytes);
  bytes.clone()
}
```

mqtt/mqtt3/src/proto/deserialize.rs
```diff
@@ -0,0 +1,140 @@
...
+        if let Some(payload) = self.loaded_bytes.get(&hash) {
+            if payload == &bytes {
```
dmolokanov: will that make byte-by-byte comparison?

mqtt/mqtt3/src/proto/deserialize.rs
```diff
@@ -0,0 +1,140 @@
...
+        payload
+    }
+
+    fn calculate_hash<T: Hash>(t: &T) -> u64 {
+        let mut s = DefaultHasher::new();
+        t.hash(&mut s);
+        s.finish()
+    }
+
+    fn reset(&mut self) {
```
dmolokanov: You can erase hash_[map|set] inside Drop implementation
```rust 
impl Drop for PayloadDeduper {
  fn drop(self) {
    self.loaded_bytes.clear();
}
```

mqtt/mqtt3/src/proto/deserialize.rs
```diff
@@ -0,0 +1,140 @@
...
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                write!(formatter, "an array of integers")
+            }
+
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: SeqAccess<'de>,
+            {
+                let topic_name = seq
+                    .next_element()?
```
dmolokanov: does serde provide any guarantees on the order of elements? I mean, can they have different layout depending on machine, OS, or some other random factor.

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -121,6 +124,53 @@ impl FileFormat for BincodeFormat {
...
+            .context(ErrorKind::Persist(ErrorReason::Serialize))?;
+        Ok(())
+    }
+}
+
+use mqtt3::proto::*;
+
+/// A simple format based on Bincode and serde
+#[derive(Clone, Debug, Default)]
+pub struct DeDupeFormat;
```
dmolokanov: Well, I think it's going to be the only production implementation. So no need to mention DeDuper part

mqtt/mqtt3/src/proto/deserialize.rs
lfitchett: It's based on the serialize function. If we do this, we should probably also implement the mirroring serialize functions

mqtt/mqtt3/src/proto/deserialize.rs
lfitchett: Yes

mqtt/mqtt3/src/proto/deserialize.rs
lfitchett: Since the hash is only u64, collision is fairly likely due to birthday paradox

mqtt/mqtt3/src/proto/deserialize.rs
lfitchett: This actually doesn't even need this function, this is just here from when I branched from option 1.

lfitchett: This code doesn't even build, I wanted to decide on an option before I spent more time on it. I just made a draft pr to make comparison easy

mqtt/mqtt3/src/proto/deserialize.rs
dmolokanov: yes, but exactly this HashSet implementation should do. If it is a collision - check for value equality

mqtt/mqtt3/src/proto/deserialize.rs
dmolokanov: Ah, I see

mqtt/mqtt3/src/proto/deserialize.rs
dmolokanov: What the doc says about ordering?

dmolokanov: https://serde.rs/deserialize-struct.html this example doesn't make it clear why it implements both `visit_seq` and `visit_map` for the same struct.

PR #2722, opened 2020-03-20
Test dotnet 3.1 build

and-rewsmith:


edge-util/test/Microsoft.Azure.Devices.Edge.Util.Test/workloadtestserver/Startup.cs
```diff
@@ -19,22 +20,24 @@ public Startup(IConfiguration configuration)
...
                 app.UseDeveloperExceptionPage();
             }
 
-            app.UseMvc();
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapControllers();
+            });
```
and-rewsmith: See:
https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-2.2&tabs=visual-studio#update-routing-startup-code

PR #2724, opened 2020-03-21
[mqtt] Consolidate Payloads

lfitchett:
This improves the persistence of the broker by consolidating payloads before it persists them. This results in both less disk used and a faster write speed.


mqtt/mqtt-broker/src/session.rs
```diff
@@ -279,12 +279,12 @@ impl DisconnectingSession {
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct SessionState {
-    client_id: ClientId,
-    subscriptions: HashMap<String, Subscription>,
+    pub client_id: ClientId,
```
dmolokanov: Let's not increase visibility of fields. They are used in the broker logic, so no need to make them more than private. Usually what people do - implement `into_parts(self)->(T1, T2, ...)`  method to deconstruct struct. You can find examples in the broker code

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -135,6 +138,209 @@ impl FileFormat for BincodeFormat {
...
+        fail_point!("bincodeformat.store.serialize_into", |_| {
+            Err(Error::from(ErrorKind::Persist(ErrorReason::Deserialize)))
+        });
+        bincode::serialize_into(encoder, &state)
+            .context(ErrorKind::Persist(ErrorReason::Serialize))?;
+        Ok(())
+    }
+}
+
+struct Consolidator {
```
dmolokanov: I believe you don't even need this struct. You can improve ergonomics of the code doing:
1. Implement `From<T> for U` trait for all parts of the consolidation state (state, sessions, publications)
2. Use `HashMap<Uuid, Publication>` from the ConsolidatedState to collect all payloads
3. Try to use as much as you can existing public API to construct state items
4. Because all ConsolidatedXXXX structs are persistor implementation details only, I suggest to rename them with the same Name as original structs.

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -135,6 +138,209 @@ impl FileFormat for BincodeFormat {
...
+        result.waiting_to_be_sent = session
+            .waiting_to_be_sent
+            .into_iter()
+            .map(|p| self.resolve_publication(p))
+            .collect();
+
+        result
+    }
+
+    fn get_id(&mut self, bytes: Bytes) -> u64 {
```
dmolokanov: Lets use uuid::new_v4() instead of hashing

mqtt/mqtt-broker/src/persist.rs
lfitchett: 1. I can't use from, since the state (hashmap) has to be passed in
1. Changed to let hashmap handle de-duping and hashing
1. Existing api? 
1. How would that work? `consolidate_state(state: mqtt::BrokerState) -> BrokerState` seems more confusing

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +145,211 @@ impl FileFormat for BincodeFormat {
...
+        Ok(())
+    }
+}
+
+struct Consolidator {
+    payloads: HashMap<Bytes, u64>,
+    curr_id: u64,
+}
+
+impl Consolidator {
```
vadim-kovalyov: it's hard to get quick understanding what is going on here. It would be great if we can add code docs that explain what is Consolidator and what problem it tries to solve.

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +145,211 @@ impl FileFormat for BincodeFormat {
...
+    fn consolidate_state(state: BrokerState) -> ConsolidatedState {
+        let mut this = Self {
```
vadim-kovalyov: nit: just curious if this is some sort of a pattern? why not just create local HashMap and local id counter? and pass it to `consolidate_publication` and `consolidate_session`? or you just don't want to move/return HashMap around?

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +145,211 @@ impl FileFormat for BincodeFormat {
...
+            .map(|(k, v)| (k, this.consolidate_publication(v)));
+        let retained = HashMap::from_iter(retained);
+
+        let sessions = sessions
+            .into_iter()
+            .map(|s| this.consolidate_session(s))
+            .collect();
+
+        let payloads = this.payloads.into_iter().map(|(k, v)| (v, k));
+        let payloads = HashMap::from_iter(payloads);
```
vadim-kovalyov: did not know that Rust allows you to shadow like that :) took me a while to realize that you actually flipping key/value

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +145,211 @@ impl FileFormat for BincodeFormat {
     }
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct ConsolidatedStateFormat;
```
dmolokanov: Why keep both formats?

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +145,211 @@ impl FileFormat for BincodeFormat {
...
+
+impl ConsolidatedStateFormat {
```
dmolokanov: I see that you did similar to FileFormat, but instead you can auto implement it with `#[derive(Default)]` feature

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +145,211 @@ impl FileFormat for BincodeFormat {
...
+}
+
+fn deserialize_payloads<'de, D>(deserializer: D) -> Result<HashMap<u64, Bytes>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let payloads = HashMap::<u64, Vec<u8>>::deserialize(deserializer)?
+        .into_iter()
+        .map(|(k, v)| (k, Bytes::from(v)));
+    let payloads = HashMap::from_iter(payloads);
```
dmolokanov: let payloads = payloads.collect::<HashMap<u64, Bytes>>();

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +145,211 @@ impl FileFormat for BincodeFormat {
...
+            .collect();
+
+        SessionState::from_parts(session.client_id, session.subscriptions, waiting_to_be_sent)
+    }
+
+    fn get_payload(&mut self, id: u64) -> Bytes {
+        self.payloads
+            .get(&id)
+            .expect("All payloads should be stored.")
+            .clone()
```
dmolokanov: In this case clone is cheap but we should avoid cloning data. you can use remove method to move publication ownership from HashMap

mqtt/mqtt-broker/benches/file_persist_bench.rs
```diff
@@ -0,0 +1,181 @@
...
+                    num_unique_messages,
+                    num_shared_messages,
+                    num_retained,
+                );
+
+                let tmp_dir = TempDir::new().unwrap();
+                let path = tmp_dir.path().to_owned();
+                let mut persistor = FilePersistor::new(path, format.clone());
+
+                Runtime::new()
```
dmolokanov: is it possible to write perf test with some macro like `#[tokio:test]` ?

mqtt/mqtt-broker/benches/file_persist_bench.rs
```diff
@@ -0,0 +1,181 @@
+use std::any::type_name;
+use std::collections::{HashMap, VecDeque};
+use std::iter::FromIterator;
+
+use bytes::Bytes;
+use criterion::*;
```
dmolokanov: Just curious why criterion?

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +145,211 @@ impl FileFormat for BincodeFormat {
...
+    }
+}
+
+#[derive(Deserialize, Serialize)]
+struct ConsolidatedState {
+    #[serde(serialize_with = "serialize_payloads")]
```
dmolokanov: why previously it worked without custom serialize methods?

mqtt/mqtt-broker/benches/file_persist_bench.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+fn make_random_payload(size: u32) -> Bytes {
+    Bytes::from_iter((0..size).map(|_| rand::random::<u8>()))
+}
+
+fn bench(c: &mut Criterion) {
+    let tests = vec![
+        (1, 1, 0, 0),
+        (10, 10, 10, 10),
```
dmolokanov: Do you think these numbers is sufficient enough for broker performance?

mqtt/mqtt-broker/src/persist.rs
dmolokanov: Please check code from my comment below

mqtt/mqtt-broker/src/persist.rs
lfitchett: I think we should keep it as a backup. The compiler will remove it anyways on release build.

mqtt/mqtt-broker/benches/file_persist_bench.rs
lfitchett: I wasn't able to find a way. I think async/await is new enough

mqtt/mqtt-broker/benches/file_persist_bench.rs
lfitchett: We didn't have a benching library and that was the most recommended one when I looked up benchmarking in rust.

mqtt/mqtt-broker/src/persist.rs
lfitchett: It still needed custom serialization methods, the custom methods were just located on the Publish class.

mqtt/mqtt-broker/benches/file_persist_bench.rs
lfitchett: No idea, I just threw those out there as an initial test. they're easy enough to change.

mqtt/mqtt-broker/src/persist.rs
lfitchett: yea, its actually really common to shadow in rust

mqtt/mqtt-broker/src/persist.rs
lfitchett: nope, denis recommended change :)

mqtt/mqtt-broker/src/persist.rs
lfitchett: denis's recommended changed should make it better

mqtt/mqtt-broker/src/persist.rs
dmolokanov: Lets not leave unused code. I may lead to confusion which variant we really use and why.

mqtt/mqtt-broker/benches/file_persist_bench.rs
dmolokanov: Ok then. Will wait until everyone caught up

mqtt/mqtt-broker/src/persist.rs
dmolokanov: Ah, thanks!

mqtt/mqtt-broker/benches/file_persist_bench.rs
dmolokanov: Could you please just play a bit to get a meaningful numbers? I have no clue what are they now.

mqtt/mqtt-broker/benches/file_persist_bench.rs
```diff
@@ -121,17 +121,19 @@ fn make_fake_state(
 
     let sessions = (0..num_clients)
         .map(|i| {
-            let mut session = SessionState::new(ClientId::from(format!("Session {}", i)));
             let waiting_to_be_sent = (0..num_unique_messages)
                 .map(|_| make_fake_publish(format!("Topic {}", i)))
                 .chain(shared_messages.clone());
```
dmolokanov:
```suggestion
                .chain(shared_messages.clone()).collect();
```
It will produce VecDeque and you don't need to specify exact type below

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -162,15 +162,15 @@ impl FileFormat for ConsolidatedStateFormat {
         fail_point!("bincodeformat.load.deserialize_from", |_| {
             Err(Error::from(ErrorKind::Persist(ErrorReason::Deserialize)))
         });
-        let state = bincode::deserialize_from(decoder)
+        let state: ConsolidatedState = bincode::deserialize_from(decoder)
             .context(ErrorKind::Persist(ErrorReason::Deserialize))?;
 
-        let state = Resolver::resolve_state(state);
+        let state: BrokerState = state.into();
```
dmolokanov: I think it's not necessary to specify type here.

mqtt/docs/persistance.md
```diff
@@ -0,0 +1,24 @@
+# Broker Persistance
+The broker's persistance strategy is modeled after mosquitto's. It saves the entire state to one compressed file. File saving is triggered by a configurable timer and on shutdown.
+
+## Requirements
+* Minimize space used on disk
+* If a valid state exists, memory constraints must not prevent it from being loaded
+* Write overhead is small
+
+## Stratagy
```
dmolokanov:
```suggestion
## Strategy
```

mqtt/docs/persistance.md
```diff
@@ -0,0 +1,24 @@
...
+### Persist
+The broker first consolidates all payloads in to a hashmap, replacing the paload with a referance id in the state. This removes duplicate payloads that are created when 2 or more devices are subscribed to the same topic. This new structure is then converted to a binary stream. This stream is passed to the GZip algorythm and the resulting compressed binary is written directly to disk. This file is called a snapshot.
```
dmolokanov: I worth mention that once broker receive signal to persist state it makes a **whole copy** of its state,
consolidates payloads, and then persist. It doesn't not directly manipulate with broker state. It helps to avoid data races but may lead to confusion from operator point of view if not stated.

mqtt/mqtt-broker/src/persist.rs
lfitchett: yea, but I think it makes it more readable

mqtt/docs/persistance.md
```diff
@@ -0,0 +1,31 @@
...
+## Strategy
+### Persist
+The broker first copies its state into a new representative object. It then consolidates all payloads in to a hashmap, replacing the payload with a reference id in the state. This removes duplicate payloads that are created when two or more devices are subscribed to the same topic. It also removes all transient data, such as currently in-flight messages and in-use packet identifiers.
+
+This new structure is then converted to a binary stream. This stream is passed to the GZip algorithm and the resulting compressed binary is written directly to disk. This file is called a snapshot.
+![InMemory](./images/InMemoryPersistance.png)
+
+![OnDisk](./images/OnDiskPersistance.png)
+
+If this is the first time persisting the state, a symlink is creating pointing to the new snapshot. If there is already a snapshot existing, the symlink is moved from the old snapshot's file to the new snapshot's file. This ensures the file the symlink points to is always valid, even if the broker unexpectedly shuts down while writing to disk.
```
vadim-kovalyov: created?
```suggestion
If this is the first time persisting the state, a symlink is created pointing to the new snapshot. If there is already a snapshot existing, the symlink is moved from the old snapshot's file to the new snapshot's file. This ensures the file the symlink points to is always valid, even if the broker unexpectedly shuts down while writing to disk.
```

mqtt/mqtt-broker/Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2018"
 [dependencies]
 async-trait = "0.1"
 bincode = "1.2"
-bytes = "0.5"
+bytes = "0.5.4"
```
dmolokanov: any specific reason for this change? other dependencies added without exact build number

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +142,156 @@ impl FileFormat for BincodeFormat {
...
 
+impl From<BrokerState> for ConsolidatedState {
+    fn from(state: BrokerState) -> Self {
+        let (retained, sessions) = state.into_parts();
+
+        #[allow(clippy::mutable_key_type)]
+        let mut payloads = HashMap::new();
+
+        let mut shrink_payload = |publication: Publication| {
+            let len = payloads.len() as u64;
```
dmolokanov, line 153:
```suggestion
            let next_id = payloads.len() as u64;
```

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -136,6 +142,156 @@ impl FileFormat for BincodeFormat {
...
+
+            let id = *payloads.entry(publication.payload).or_insert(len);
```
dmolokanov, line 155:
```suggestion
            let id = *payloads.entry(publication.payload).or_insert(next_id);
```

mqtt/mqtt-broker/Cargo.toml
lfitchett: just a merge change. We both independently added it and your change went in first. I'll take off the 4

PR #2725, opened 2020-03-23
[TEST ONLY] Limited Container Storage Test 

yophilav:
Introduce Limited Container Memory Tests
- Introduce `Benchmark` pipeline (based off of Connectivity pipeline)
- Test Limited Memory: EdgeHub {128MB, 192MB, 256MB} on {AMQP, MQTT, HTTP-MQTT, HTTP-AMQP} protocol heads
- Test Limited Memory EdgeAgent{64MB, 128MB}
- Includes two deployment files with protocol specific modules {MQTT, AMQP, HTTP}
- Includes Network Controller & Metric Collector


builds/e2e/benchmark.yaml
```diff
@@ -0,0 +1,232 @@
+trigger: none
```
philipktlin: I think this file should not be in e2e folder.  If you foresee more yamls for benchmark, we should create a new folder named "benchmark". if not, maybe move it to misc folder.

builds/e2e/benchmark.yaml
```diff
@@ -0,0 +1,232 @@
...
+pr: none
+
+variables:
+  images.artifact.name.linux: 'core-linux'
+  vsts.project: $(System.TeamProjectId)
```
philipktlin:
```suggestion
  vsts.project.Id: $(System.TeamProjectId)
```

builds/e2e/benchmark.yaml
```diff
@@ -0,0 +1,232 @@
...
+          itemPattern: |
+            $(images.artifact.name.linux)/IotEdgeQuickstart.linux-x64.tar.gz
+            $(images.artifact.name.linux)/e2e_deployment_files/$(deploymentFileName)
+            $(images.artifact.name.linux)/scripts/connectivityTest.sh
+            $(images.artifact.name.linux)/scripts/testHelper.sh
+            $(images.artifact.name.linux)/artifactInfo.txt
+      # Deploy connectivity test
+      - template: templates/connectivity-deploy.yaml
+        parameters:
+          release.label: 'ct$(agent.group)'
```
philipktlin: don't use "ct", it stands for connectivity test.

e2e_deployment_files/limited_container_storage_deployment.amqp.template.json
```diff
@@ -0,0 +1,439 @@
+{
+  "modulesContent": {
```
philipktlin: is it possible to combine limited_container_storage_deployment.amqp.template.json with limited_container_storage_deployment.mqtt.template.json?

test/connectivity/scripts/connectivityTest.sh
```diff
@@ -42,6 +42,11 @@ function prepare_test_from_artifacts() {
     sed -i -e "s@<Container_Registry>@$CONTAINER_REGISTRY@g" "$deployment_working_file"
```
philipktlin: In general, I think we should refactor this script. The problem previously we had in runE2ETests.sh is that we combine all the scenario into 1 single script file and therefore the input parameters are union of all scenarios.  It makes difficult to maintain.

To solve this problem, I started to create a testHelper.sh to extract common method/function for reuse.  What I would suggest:
1. Extract everything in connectivityTest.sh into small functions.  For example, process_args should be broken down into many methods, e.g. process_args_for_basic_parameters, process_args_for_metric, process_args_for_customBuild, etc; then move all these methods into testHelper.sh.  And move testHelper.sh to test folder for reuse.
2. create a new folder "benchmark" under test folder.
3. create a new "benchmark.sh" to use testHelper.sh to build the tests and uses exactly what that test really needs.  For example, I expect it will call prepare_test_from_artifacts_for_basic, prepare_test_from_artifacts_for_protocolHeads, but not prepare_test_from_artifacts_for_TwinTests.
4. With this new test script, a new yaml "benchmark-deploy.yaml" should be created.

philipktlin: This is not just to make your tests to work; this will help us to easily maintain our test script.  Check with Varun if he would like to do this refactoring later.  This is trade-off between code maintenance and deadline.

yophilav: I think this should be a task of its own. Would it be alright if I just do a setup of the pipeline + update the deployment file then I'll create a task to re-factor the bash files later?

philipktlin: Please confirm with Varun.

yophilav: I created this task to follow up.
https://msazure.visualstudio.com/One/_workitems/edit/6953217

e2e_deployment_files/limited_container_storage_deployment.amqp.template.json
yophilav: yup

builds/e2e/benchmark.yaml
```diff
@@ -0,0 +1,241 @@
...
+    timeoutInMinutes: 180
+    strategy:
+      matrix:
+        EdgeHub 128MB Storage AMQP:
+          testrun.config: 1
+          testrun.network.description: "Online"
+          testrun.network.frequencies: "00:00:00 00:00:00 0"
+          testrun.network.runProfile: "Online"
+          testrun.duration: "01:00:00"
+          edgeHub.storageSizeByte: "134217728"
```
philipktlin, line 23: use variable for "134217728"

PR #2727, opened 2020-03-23
[mqtt] added initial configuration

vipeller:


mqtt/mqtt-broker/src/config.json
```diff
@@ -0,0 +1,18 @@
+{
```
dmolokanov: Let's move this file with default settings to the `mqtt/mqtt-broker/config/default.json`

mqtt/mqtt-broker/src/config_bad_value_type.json
```diff
@@ -0,0 +1,6 @@
+{
```
dmolokanov: let's move all files that are for tests to the `mqtt/mqtt-broker/test/` folder

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+#[derive(Debug, Deserialize)]
+pub struct SessionMessages {
+    #[serde(deserialize_with = "humansize")]
+    max_message_size: Option<u64>,
+    max_count: Option<u32>,
+    #[serde(deserialize_with = "humansize")]
+    max_total_space: Option<u64>,
+    when_full: Option<String>,
```
dmolokanov: lets make it enum with all existing variants

mqtt/mqtt-broker/Cargo.toml
```diff
@@ -31,7 +35,9 @@ bytes = "0.5"
 fail = { version = "0.3", features = ["failpoints"] }
 matches = "0.1"
 proptest = "0.9"
+serde_json = "1.0"
 tempfile = "3"
+test-case = "1.0.0"
```
dmolokanov:
```suggestion
test-case = "1.0"
```

mqtt/mqtt-broker/src/config.json
myagley: Can we look at toml for this instead of json?

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+    struct Container {
+        #[serde(deserialize_with = "humansize")]
+        size: Option<u64>,
+    }
+
+    #[test_case( "123b",  123 ; "when using bytes")]
+    #[test_case( "123kb",  123*1024 ; "when using kilobytes")]
+    #[test_case( "123mb",  123*1024*1024 ; "when using megabytes")]
+    #[test_case( "123gb",  123*1024*1024*1024 ; "when using gigabytes")]
+    fn it_deserializes_different_multipliers(input: &str, expected: u64) {
```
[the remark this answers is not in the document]
myagley: Agree. If we are writing a parser, we should proptest it.

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+    pub fn new(path: Option<&Path>) -> Result<Self, ConfigError> {
+        let mut s = Config::new();
+        s.merge(File::from_str(DEFAULTS, FileFormat::Json))?;
+
+        if let Some(path) = path {
+            s.merge(File::from(path))?;
+        }
+
+        s.try_into()
+    }
```
myagley: I would probably make these two methods instead of using an option.

```rust
impl BrokerSettings {
    pub fn new() -> Result<Self, ConfigError> {
        // return defaults
    }

    pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self, ConfigError> {
        // merge file with defaults
    }
}
```
Passing an Option as a parameter is kind of an antipattern for situations like this.

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+pub struct RetainedMessages {
+    max_count: Option<u32>,
+    #[serde(with = "humantime_serde")]
+    expiration: Option<Duration>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct SessionMessages {
+    #[serde(deserialize_with = "humansize")]
+    max_message_size: Option<u64>,
```
myagley: Are there appropriate defaults for these sizes? for instance: `u64::MAX` This will make using this a lot less annoying because we won't need to unwrap the option.

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+    file_path: String,
+    #[serde(with = "humantime_serde")]
+    time_interval: Option<Duration>,
+    unsaved_message_count: Option<u32>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Session {
+    #[serde(with = "humantime_serde")]
+    expiration: Option<Duration>,
```
myagley: Same here: duration max

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+    messages: SessionMessages,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct BrokerSettings {
```
myagley: `BrokerConfig`?

mqtt/mqtt-broker/src/configuration.rs
myagley: For some of these fields we may want to look at making them `usize`. For instance, max message size seems like a good candidate, because a 32 bit platform couldn't allocate more than usize bytes.

mqtt/mqtt-broker/src/configuration.rs
myagley: The caller would do something like this:

```rust
let path: Option<OsString> = "myconfig";
let config = path.map_or(BrokerSetting::new(), BrokerSettings::from_file)?;
```

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    let s = String::deserialize(d)?;
+
+    let captures = match RE.captures(&s.as_str()) {
+        Some(s) => s,
+        None => return Err(error::<D>(&s.as_str(), &"256kb")),
+    };
+
+    let base = match captures[1].parse::<u64>() {
```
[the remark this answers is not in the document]
vipeller: If the regex fits, then because of the pattern, there are 2 capturing groups + the whole matching string, so "captures" should have 3 elements - if the regex library works well, and the pattern is right.

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+        Ok(n) => n,
+        Err(_) => return Err(error::<D>(&captures[1], &"256")),
```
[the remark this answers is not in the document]
vipeller: according to the doc, that is an "example" for right format. could be any valid number

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+    D: Deserializer<'de>,
+{
+    lazy_static! {
+        static ref RE: Regex = Regex::new(r"\s*(\d+)\s*([a-zA-Z]+)\s*")
+            .expect("failed to create new Regex from pattern");
+    }
+
+    let s = String::deserialize(d)?;
+
+    let captures = match RE.captures(&s.as_str()) {
```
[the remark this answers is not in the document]
vipeller: actually it is not an ok/err but some/none situation and the none part must return

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+        Some(s) => s,
+        None => return Err(error::<D>(&s.as_str(), &"256kb")),
```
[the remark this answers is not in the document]
vipeller: it cries for the type parameter (it doesn't need as_str(), though)

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,211 @@
...
+    };
+
+    let base = match captures[1].parse::<u64>() {
+        Ok(n) => n,
+        Err(_) => return Err(error::<D>(&captures[1], &"256")),
+    };
+
+    let multiplier = captures[2].to_lowercase();
```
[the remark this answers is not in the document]
vipeller: according to uncle bob... :)

mqtt/mqtt-broker/src/configuration.rs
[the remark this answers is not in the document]
vipeller: ok it works now the way you suggested

mqtt/mqtt-broker/src/configuration.rs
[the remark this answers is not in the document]
dmolokanov: I see

mqtt/mqtt-broker/src/config.json
vipeller: I thought we had a discussion about this that part of the settings would go to twins, and because of that going with json is a smoother transition?

mqtt/mqtt-broker/src/config_bad_value_type.json
vipeller: moved

mqtt/mqtt-broker/src/configuration.rs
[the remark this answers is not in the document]
vipeller: added prop test

mqtt/mqtt-broker/src/config.json
myagley: ok makes sense.

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,304 @@
+use config::{Config, ConfigError, File, FileFormat};
+use lazy_static::lazy_static;
+use regex::Regex;
+use serde::{Deserialize, Deserializer};
+use std::path::Path;
```
dmolokanov: `std::` use blocks should be at the top of the file

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,304 @@
...
+use std::time::Duration;
+
+pub const DEFAULTS: &str = include_str!("../config/default.json");
+
+#[derive(Debug, Deserialize)]
+pub enum QueueFullAction {
```
dmolokanov: Should we use `#[serde(rename_all = "snake_case")] ` on top of enum instead of specifying each value separately? https://serde.rs/container-attrs.html

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,304 @@
...
+    when_full: QueueFullAction,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct SessionPersistence {
+    file_path: String,
+
+    #[serde(with = "humantime_serde")]
+    time_interval: Duration,
+    unsaved_message_count: u32,
```
dmolokanov: Somewhere fields are separated with new line, somewhere not. Could you please make it more consistent?

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,304 @@
...
+pub struct BrokerConfig {
+    inflight_messages: InflightMessages,
+    retained_messages: RetainedMessages,
+    session: Session,
+    persistence: Option<SessionPersistence>,
+}
+
+pub fn humansize<'de, T, D>(deserializer: D) -> Result<T, D::Error>
+where
+    T: std::str::FromStr + std::ops::Mul<Output = T> + std::convert::From<u32>,
```
dmolokanov: Usually we try to import types and not use fully qualified name

mqtt/mqtt-broker/src/configuration.rs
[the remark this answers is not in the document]
dmolokanov: In this case we should go further. Define a separate type Volume with multiple variants for Bytes, Kilobytes, etc.. No? :)

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,304 @@
...
+
+fn error<'de, D>(unexpected: &str, expected: &str) -> D::Error
+where
+    D: Deserializer<'de>,
+{
+    serde::de::Error::invalid_value(serde::de::Unexpected::Str(&unexpected), &expected)
+}
+
+impl BrokerConfig {
+    pub fn new() -> Result<Self, ConfigError> {
```
dmolokanov: In what case we use this method and not from_file?

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,304 @@
...
+    proptest! {
+        #[test]
+        fn doesnt_crash(s in "\\PC*") {
+            parse_size_anything(&s);
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn it_parses_valid_input(p in (r"\s*",             // leading spaces if any
```
dmolokanov: Is it possible to move generator part in its own function? now it is a bit messy

dmolokanov: Propably, @myagley has more ideas on how to improve prop tests

mqtt/mqttd/src/main.rs
```diff
@@ -23,6 +23,16 @@ async fn main() -> Result<(), Error> {
         .nth(1)
         .unwrap_or_else(|| "0.0.0.0:1883".to_string());
 
+    // TODO pass it to broker
+    // TODO make it an argument to override defaul config
+    let path: Option<String> = None;
+    let config = path
```
dmolokanov: Interesting, for iotedged we used a different approach, where we have a method like `new` that takes `Option<Path>` rather than 2 almost identical methods. Any benefits to have 2 methods?

mqtt/mqtt-broker/Cargo.toml
```diff
@@ -31,7 +35,9 @@ bytes = "0.5"
 fail = { version = "0.3", features = ["failpoints"] }
 matches = "0.1"
 proptest = "0.9"
+serde_json = "1.0"
```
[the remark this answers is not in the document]
vipeller: I use it in tests to write json directly in the code then parse them immediately. I am not sure if we need it directly in normal code (at least right now)

mqtt/mqtt-broker/src/configuration.rs
vipeller: it was @myagley suggestion, see a bit above, also how to use it when no config file is defined to override defaults.

mqtt/mqtt-broker/Cargo.toml
```diff
@@ -10,11 +10,15 @@ edition = "2018"
 async-trait = "0.1"
 bincode = "1.2"
 chrono = "0.4"
+config = "0.10"
```
dmolokanov:
```suggestion
config = { version = "0.10", features = ["json"] }
```
to disable other file formats

mqtt/mqtt-broker/src/configuration.rs
dmolokanov: Ah, just found it. Still curious why it is better :)

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,298 @@
...
+impl BrokerConfig {
+    pub fn new() -> Result<Self, ConfigError> {
+        let mut s = Config::new();
+        s.merge(File::from_str(DEFAULTS, FileFormat::Json))?;
+
+        s.try_into()
+    }
+
+    pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self, ConfigError> {
+        let mut s = Config::new();
```
dmolokanov:
```suggestion
        let mut s = Self::new();
```
and you don't need to merge defaults

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    proptest! {
+        #[test]
+        fn it_parses_valid_input(lead in r"\s*",
+                                 num in any::<u64>(),
+                                 sep in r"\s*",
+                                 unit in r"(k|K|m|M|g|G)?(b|B)",
+                                 trail in r"\s*") {
+            let (input, expected) = get_input_and_expected(&lead, num, &sep, &unit, &trail);
+            parse_size_valid(&input, expected);
```
dmolokanov: I'd say lets move test body in the test function. Otherwise it's hard to read tests with usual structure Arrage, Act, Assert.
Here is a good example https://github.com/Azure/iotedge/blob/f3c7e9c8e7ace0db0390537ff5db1b8e324ffbf9/mqtt/mqtt-broker/src/subscription.rs#L297-L303

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,298 @@
...
+        let container_json = json!({ "size": input }).to_string();
+        let result = serde_json::from_str::<Container>(&container_json);
+
+        assert_matches!(result, Err(_err));
+    }
+
+    proptest! {
+        #[test]
+        fn doesnt_crash(s in "\\PC*") {
+            parse_size_anything(&s);
```
dmolokanov: Here as well test body should in the test function

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,298 @@
...
+    fn it_fails_deserializing_unknown_unit(input: &str) {
+        let container_json = json!({ "size": input }).to_string();
+        let result = serde_json::from_str::<Container>(&container_json);
+
+        assert_matches!(result, Err(_err));
+    }
+
+    proptest! {
+        #[test]
+        fn doesnt_crash(s in "\\PC*") {
```
dmolokanov:
```suggestion
        fn it_does_not_panic(s in "\\PC*") {
```

mqtt/mqtt-broker/src/configuration.rs
dmolokanov: You probably can 
1. create a custom type for all params. 
2. create a generator method
3. implement ToString for this type
4. almost copy-paste test from previous comment
....
the code will be much cleaner

vipeller: added a model structure for the test

PR #2728, opened 2020-03-23
[mqtt] Add basic authentication and authorization on connect

dmolokanov:
Implemented infrastructure for authentication and authorization: traits and basic implementation.
Authentication always returns identity that corresponds to anonymous client
Authorization always allows to connect to the broker.


mqtt/mqtt-broker/src/auth.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use crate::Error;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum AuthId {
+    Any,
+    Value(Identity),
+}
+
+pub type Identity = String;
```
lfitchett: Should you wrap this instead? 

https://github.com/rust-unofficial/patterns/blob/master/patterns/newtype.md

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -27,39 +28,77 @@ pub struct BrokerState {
...
+    N: Authenticator,
+    Z: Authorizer,
+{
+    pub fn new(authenticator: N, authorizer: Z) -> Self {
         Self {
-            sender,
-            messages,
-            sessions: HashMap::new(),
-            retained: HashMap::new(),
+            state: Option::default(),
```
lfitchett: why not just `None`?

mqtt/mqtt-broker/src/auth.rs
```diff
@@ -0,0 +1,85 @@
...
+            username,
+            password,
+            certificate,
+        }
+    }
+}
+
+pub trait Authenticator {
+    fn authenticate(&self, credentials: Credentials) -> Result<Option<AuthId>, Error>;
+}
```
vipeller: I wonder if the Option is needed around the AuthId. It is either authenticated (even if it becomes "any" which I see something like anonymous) or not, and if not, that is an error case. Because of that, I don't see a non-error coming back with None(). Is there any case I don't see?

mqtt/mqtt-broker/src/auth.rs
dmolokanov: It's not clear for now what type is going to be for Indentity 
just `struct Identity(String)`
or `enum Identity { Local(String), IotHub(String) }` 
so I decided to leave it as an alias for now

mqtt/mqtt-broker/src/broker.rs
dmolokanov: I think default() is more convenient if some decide to use another structure instead of Option. But it could be a None here, agree.

mqtt/mqtt-broker/src/auth.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Any => write!(f, "any"),
+            Self::Value(auth_id) => write!(f, "{}", auth_id),
+        }
+    }
+}
+
+#[derive(Debug, Fail, PartialEq)]
+pub enum AuthReason {
```
vadim-kovalyov: nit: 
```suggestion
pub enum AuthFailReason {
```

mqtt/mqtt-broker/src/auth.rs
dmolokanov: We want to distinguish cases
None - external component is not able to identify client, and
Err() - error occurred while making a call to an external component.

mqtt/mqtt-broker/src/auth.rs
dmolokanov: Yeah, should be just `ErrorReason` the same way as here: https://github.com/Azure/iotedge/blob/master/mqtt/mqtt-broker/src/persist.rs#L348

mqtt/mqtt-broker/src/auth.rs
```diff
@@ -0,0 +1,85 @@
+#![allow(dead_code)] // TODO @dmolokanov remove when implemented
+use std::fmt;
+
+use failure::Fail;
+
+use crate::Error;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum AuthId {
```
vadim-kovalyov: I know most of the things here are self-explanatory, but maybe still add some code docs in this file? for example we can explain what is AuthId, how it's different from any other identities, etc...

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -226,22 +283,50 @@ impl Broker {
...
+                return Err(e);
+            }
+        };
 
-            debug!("dropping connection due to invalid protocol level");
-            let message = Message::Client(client_id, ClientEvent::DropConnection);
-            try_send!(connreq.handle_mut(), message);
-            return Ok(());
+        // Check client permissions to connect
+        match self.authorize(auth_id).await {
```
varunpuranik: All messages should be authorized, right? Not just the connects.?

mqtt/mqtt-broker/src/auth.rs
```diff
@@ -0,0 +1,85 @@
...
+        Self {
+            username,
+            password,
+            certificate,
+        }
+    }
+}
+
+pub trait Authenticator {
+    fn authenticate(&self, credentials: Credentials) -> Result<Option<AuthId>, Error>;
```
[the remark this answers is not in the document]
dmolokanov: Ok(Some(id)) - component was able to obtain identity for given credentials
Ok(None) - component was **not**  able to obtain identity for given credentials
Err(e) - component returned an error while obtaining identity

mqtt/mqtt-broker/src/auth.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+pub trait Authorizer {
+    fn authorize(&self, auth_id: AuthId) -> Result<bool, Error>;
```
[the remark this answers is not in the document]
dmolokanov: Yes, this PR target authentication/authorization for client connection only. it will change soon to support other packet types

mqtt/mqtt-broker/src/broker.rs
dmolokanov: Correct. https://github.com/Azure/iotedge/pull/2728#discussion_r397339943

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -1023,26 +1376,66 @@ pub(crate) mod tests {
...
+    struct TestAuthenticator<F>(F);
+
+    impl<F> Authenticator for TestAuthenticator<F>
+    where
+        F: Fn() -> Result<Option<AuthId>, Error>,
+    {
+        fn authenticate(&self, _credentials: Credentials) -> Result<Option<AuthId>, Error> {
+            self.0()
+        }
     }
```
myagley: You should be able to impl `Authenticator` for functions (in the main code):

```rust
impl<F> Authenticator for F
where
    F: FnOnce(Credentials) -> Result<Option<AuthId>, Error>,
{
    fn authenticate(&self, credentials: Credentials) -> Result<Option<AuthId, Error> {
        self(credentials)
    }
} 
```

then you could pass a closure directly to the builder. (note I didn't compile this code, so it might need to be tweaked)

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -747,9 +789,59 @@ async fn publish_to(session: &mut Session, publication: &proto::Publication) ->
...
+    N: Authenticator,
+    Z: Authorizer,
+{
+    pub fn new(authenticator: N, authorizer: Z) -> Self {
+        Self {
+            state: Option::default(),
+            authenticator,
+            authorizer,
+        }
+    }
```
myagley: I don't think the authenticator and authorizer need to be required here. We could default them to something that noops or denies everything be default. For example:

```rust
BrokerBuilder::new()
    .authenticator(authN)
    .authorizer(authZ)
    .state(state)
    .build()
```

```rust
impl<N, Z> BrokerBuilder<N, Z>
where
    N: Authenticator,
    Z: Authorizer,
{
    pub fn new() -> Self {
        Self {
            state: None,
            authenticator: DefaultAuthenticator,
            authorizer: DefaultAuthorizer,
        }
    }

    pub fn authenticator<N1: Authenticate>(mut self, authenticator: N1) -> BrokerBuilder<N1, Z> {
        BrokerBuilder {
            state: self.state,
            authenticator,
            authorizer: self.authorizer,
        }
    }

    ...
```

mqtt/mqtt-broker/src/connection.rs
```diff
@@ -1,8 +1,8 @@
-use std::fmt;
 use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Duration;
 
+use derive_more::Display;
```
myagley: Are we only using this to get a `Display` derive? I'm not sure taking on another dependency is worth it?

mqtt/mqtt-broker/src/auth.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[display(fmt = "*")]
+    Any,
+
+    Value(Identity),
+}
+
+pub type Identity = String;
+
+#[derive(Clone, Debug)]
+pub struct Certificate(Vec<u8>);
```
myagley: I'm assuming this will eventually be some more sophisticated type than a byte array? Do the TLS libraries give us a parsed certificate?

mqtt/mqtt-broker/src/auth.rs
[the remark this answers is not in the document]
dmolokanov: Yes, this is the plan

mqtt/mqtt-broker/src/broker.rs
dmolokanov: Yes

mqtt/mqtt-broker/src/auth.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[allow(dead_code)]
+pub struct Credentials {
+    username: Option<String>,
+    password: Option<String>,
+    certificate: Option<Certificate>,
```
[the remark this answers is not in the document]
dmolokanov: Do you suggest I split them? Like this
```rust
pub enum Credentials { Basic(String, String), Certificate(Certificate) }
```

mqtt/mqtt-broker/src/broker.rs
dmolokanov: Nice Idea, for tests I can use the same!

mqtt/mqtt-broker/src/connection.rs
dmolokanov: I think yes, we could use it for something else. It has more features

mqtt/mqtt-broker/src/auth.rs
dmolokanov: TLS library provide it as a `Vec<u8>` we need to parse it with something else.

mqtt/mqtt-broker/src/auth.rs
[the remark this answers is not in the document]
myagley: If that makes sense, then yeah. Basically I would move all of the validation to the constructor of this type, so that the rest of the code can just assume everything is nice.

mqtt/mqtt-broker/src/auth.rs
myagley: bummer. :)

myagley: Wait, so does the `Authenticator` trait need to be async then?

dmolokanov: Because it will call EH via some kind of rpc?

myagley: yeah, that seems to imply the authentcator trait would need to be async then, if the rpc is going to happen in the `authenticate` call

myagley: We also probably don't need to call into the EdgeHub for cert authentication, just username/password? Hmm, I guess there is the whole thumbprint cert auth and scopes though...

PR #2730, opened 2020-03-24
Message priority metrics and debug logging

richma-ms:
This change adds priority to RoutingMessage and EdgeMessage, so that the values can be plumb down to the lower layers where we record message metrics.

The following metrics were modified with message priority information:
- Total messages sent
- Message time to process
- Message time to send

I've also added more detailed debug logging in the executor regarding priority changes, message enqueue, and message processing.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
@@ -135,7 +135,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             this.timer.Reset();
             try
             {
-                using (Metrics.TimeMessageSend(this.clientId, inputMessage.ProcessedPriority))
+                using (Metrics.TimeMessageSend(this.clientId))
                 {
                     Metrics.MessageProcessingLatency(this.clientId, inputMessage);
                     await this.client.SendEventAsync(message);
@@ -155,7 +155,6 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
         {
             IMessageConverter<Message> converter = this.messageConverterProvider.Get<Message>();
             string metricOutputRoute = null;
-            uint priority = uint.MaxValue;
             IList<Message> messages = Preconditions.CheckNotNull(inputMessages, nameof(inputMessages))
                 .Select(inputMessage =>
                 {
@@ -166,18 +165,19 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
                 .ToList();
             this.timer.Reset();
 
-            if (messages.Count > 0)
-            {
-                priority = inputMessages.First().ProcessedPriority;
-            }
-
             try
             {
-                using (Metrics.TimeMessageSend(this.clientId, priority))
+                using (Metrics.TimeMessageSend(this.clientId))
                 {
                     await this.client.SendEventBatchAsync(messages);
                     Events.SendMessage(this);
-                    Metrics.AddSentMessages(this.clientId, messages.Count, metricOutputRoute, priority);
+
+                    if (messages.Count > 0)
+                    {
+                        // Priority for messages in any given batch is the same, so we can
+                        // just use the priority from the first message in the collection
+                        Metrics.AddSentMessages(this.clientId, messages.Count, metricOutputRoute, inputMessages.First().ProcessedPriority);
+                    }
                 }
             }
             catch (Exception ex)
@@ -623,7 +623,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             static readonly IMetricsTimer MessagesTimer = Util.Metrics.Metrics.Instance.CreateTimer(
                 "message_send_duration_seconds",
                 "Time taken to send a message",
-                new List<string> { "from", "to", "priority" });
+                new List<string> { "from", "to" });
 
             static readonly IMetricsCounter SentMessagesCounter = Util.Metrics.Metrics.Instance.CreateCounter(
                 "messages_sent",
@@ -660,7 +660,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
                 "Time taken to process message in EdgeHub",
                 new List<string> { "from", "to", "priority" });
 
-            public static IDisposable TimeMessageSend(string id, uint priority) => MessagesTimer.GetTimer(new[] { id, "upstream", priority.ToString() });
+            public static IDisposable TimeMessageSend(string id) => MessagesTimer.GetTimer(new[] { id, "upstream" });
 
             public static void AddSentMessages(string id, int count, string fromRoute, uint priority) =>
                 SentMessagesCounter.Increment(count, new[] { id, "upstream", fromRoute, string.Empty, priority.ToString() });
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
@@ -135,12 +135,12 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             this.timer.Reset();
             try
             {
-                using (Metrics.TimeMessageSend(this.clientId))
+                using (Metrics.TimeMessageSend(this.clientId, inputMessage.ProcessedPriority))
                 {
                     Metrics.MessageProcessingLatency(this.clientId, inputMessage);
                     await this.client.SendEventAsync(message);
                     Events.SendMessage(this);
-                    Metrics.AddSentMessages(this.clientId, 1, inputMessage.GetOutput());
+                    Metrics.AddSentMessages(this.clientId, 1, inputMessage.GetOutput(), inputMessage.ProcessedPriority);
                 }
             }
             catch (Exception ex)
@@ -155,6 +155,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
         {
             IMessageConverter<Message> converter = this.messageConverterProvider.Get<Message>();
             string metricOutputRoute = null;
+            uint priority = uint.MaxValue;
             IList<Message> messages = Preconditions.CheckNotNull(inputMessages, nameof(inputMessages))
                 .Select(inputMessage =>
                 {
@@ -164,13 +165,19 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
                 })
                 .ToList();
             this.timer.Reset();
+
+            if (messages.Count > 0)
+            {
+                priority = inputMessages.First().ProcessedPriority;
+            }
+
             try
             {
-                using (Metrics.TimeMessageSend(this.clientId))
+                using (Metrics.TimeMessageSend(this.clientId, priority))
                 {
                     await this.client.SendEventBatchAsync(messages);
                     Events.SendMessage(this);
-                    Metrics.AddSentMessages(this.clientId, messages.Count, metricOutputRoute);
+                    Metrics.AddSentMessages(this.clientId, messages.Count, metricOutputRoute, priority);
                 }
             }
             catch (Exception ex)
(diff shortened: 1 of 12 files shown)


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
```diff
@@ -135,12 +135,12 @@ public async Task SendMessageAsync(IMessage inputMessage)
             this.timer.Reset();
             try
             {
-                using (Metrics.TimeMessageSend(this.clientId))
+                using (Metrics.TimeMessageSend(this.clientId, inputMessage.ProcessedPriority))
```
varunpuranik: Do we need Priority for the message process time?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
```diff
@@ -155,6 +155,7 @@ public async Task SendMessageBatchAsync(IEnumerable<IMessage> inputMessages)
         {
             IMessageConverter<Message> converter = this.messageConverterProvider.Get<Message>();
             string metricOutputRoute = null;
+            uint priority = uint.MaxValue;
```
lfitchett: should this be `RouteFactory.DefaultPriority`?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
```diff
@@ -473,7 +473,7 @@ static class Metrics
             static readonly IMetricsTimer MessagesTimer = Util.Metrics.Metrics.Instance.CreateTimer(
                 "message_send_duration_seconds",
                 "Time taken to send a message",
-                new List<string> { "from", "to" });
+                new List<string> { "from", "to", "priority" });
```
varunpuranik: I am not sure these metrics need priority.
The e2e latency metric does.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -536,6 +536,8 @@ public MessageWithOffset(IMessage message, long offset)
 
             public long Offset { get; }
 
+            public uint ProcessedPriority { get; set; }
```
varunpuranik: Hmm, is this a breaking change? All current messages will be Pri 0, as opposed to default, right?
And don't make the property settable

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/DeviceProxy.cs
```diff
@@ -164,7 +164,7 @@ static class Metrics
             static readonly IMetricsCounter SentMessagesCounter = Util.Metrics.Metrics.Instance.CreateCounter(
                 "messages_sent",
                 "Messages sent from edge hub",
-                new List<string> { "from", "to", "from_route_output", "to_route_input" });
+                new List<string> { "from", "to", "from_route_output", "to_route_input", "priority" });
```
lfitchett: Do you need this for device proxy? Don't priorities only apply to upstream messages? For local messages I think the tag should just be left off.

richma-ms: it's primarily used for upstream, but priority works for both cloud endpoint and brokered endpoint.

---
In reply to: [397351570](https://github.com/Azure/iotedge/pull/2730#discussion_r397351570) [](ancestors = 397351570)

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -1,143 +1,144 @@
-// Copyright (c) Microsoft. All rights reserved.
-namespace Microsoft.Azure.Devices.Routing.Core.Endpoints
-{
-    using System;
-    using System.Collections;
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Routing.Core.Endpoints
```
varunpuranik: Did some formatting change here? Can you try to revert it?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -1,143 +1,144 @@
...
+        readonly AtomicBoolean closed = new AtomicBoolean();
+        readonly IMessageStore messageStore;
+        readonly Task sendMessageTask;
+        readonly AsyncManualResetEvent hasMessagesInQueue = new AsyncManualResetEvent(true);
+        readonly AsyncEndpointExecutorOptions options;
+        readonly CancellationTokenSource cts = new CancellationTokenSource();
+        readonly ICheckpointerFactory checkpointerFactory;
+        readonly EndpointExecutorConfig config;
+        AtomicReference<SortedDictionary<uint, EndpointExecutorFsm>> prioritiesToFsms;
+        AtomicReference<EndpointExecutorFsm> lastUsedFsm;
```
varunpuranik: I suggest we don't bundle such changes together - they are hard to track, revert, etc.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms: I think VS overwrote the CLRF formatting, but a bunch of stuff actually changed in this file, lemme see what I can do to fix it.

richma-ms: okay, reverted this file and then manually re-applied the changes, it should now be fixed.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
richma-ms: I'm trying to signify an invalid value here, as opposed to the default value.  If there are no inputMessages, DefaultPriority has a good chance of being misleading.

---
In reply to: [397345757](https://github.com/Azure/iotedge/pull/2730#discussion_r397345757) [](ancestors = 397345757)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
```diff
@@ -164,13 +165,19 @@ public async Task SendMessageBatchAsync(IEnumerable<IMessage> inputMessages)
                 })
                 .ToList();
             this.timer.Reset();
+
+            if (messages.Count > 0)
```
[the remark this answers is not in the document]
richma-ms: this check was added here to make sure that inputMessages.First() doesn't throw.  Moving it to AddSentMessages would require passing in the messages collection to get the priority, which then further requires AddSentMessages() to have a split implementation for single vs batched messages.

---
In reply to: [397345699](https://github.com/Azure/iotedge/pull/2730#discussion_r397345699) [](ancestors = 397345699)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
richma-ms: hmm...I'm fine either way.  I just added it because it was relevant to messages, but one could argue that all messages are treated the same once they reach this point anyways.

---
In reply to: [397345125](https://github.com/Azure/iotedge/pull/2730#discussion_r397345125) [](ancestors = 397345125)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
richma-ms: I'm assuming you couldn't tell how this was used because StoringAsyncEndpointExecutor.cs got CRLF stomped.

But the set is necessary because the MessageStore doesn't understand priority when it re-constructs the message objects queue.  So the executor layer needs a way to tag the messages with their priority after the fact.

So this new property is only used for plumbing priority info for the purpose of metrics, and has no impact on any business logic.

---
In reply to: [397348395](https://github.com/Azure/iotedge/pull/2730#discussion_r397348395) [](ancestors = 397348395)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/EdgeMessage.cs
```diff
@@ -13,6 +14,15 @@ public EdgeMessage(byte[] body, IDictionary<string, string> properties, IDiction
             this.Body = Preconditions.CheckNotNull(body, nameof(body));
             this.Properties = Preconditions.CheckNotNull(properties, nameof(properties));
             this.SystemProperties = Preconditions.CheckNotNull(systemProperties, nameof(systemProperties));
+            this.ProcessedPriority = RouteFactory.DefaultPriority;
```
dylanbronson: Why the naming ProcessedPriority?

richma-ms:
talked offline, it's not technically broken because the queue length is dependent on queue ID, which now has format "endpoint_Pri0".  But that's not really ideal, we should be reporting the priority value as a different tag.

---
In reply to: [380553796](https://github.com/Azure/iotedge/pull/2730#issuecomment-380553796) [](ancestors = 380553796)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/EdgeMessage.cs
richma-ms: so that people don't confuse it with priority as when used during routing.  This property is not considered valid until the message is being processed, but since it's on the very generic and widespread IMessage interface, I don't want people to get the wrong idea and start thinking it can be referenced whenever and whereever.

you know what, now I think about I should prob leave a big fat comment on the interface decl to explain this.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/IMessage.cs
```diff
@@ -11,5 +11,11 @@ public interface IMessage : IDisposable
         IDictionary<string, string> Properties { get; }
 
         IDictionary<string, string> SystemProperties { get; }
+
+        // Used to report the priority at which the message
```
vadim-kovalyov: just curious why not doc style comments `///<summary>...` ?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/checkpointers/Checkpointer.cs
```diff
@@ -32,10 +32,40 @@ public class Checkpointer : ICheckpointer
             this.UnhealthySince = checkpointData.UnhealthySince;
             this.Proposed = checkpointData.Offset;
             this.closed = new AtomicBoolean(false);
+
+            // The endpoint ID and priority is encoded into the checkpointer ID
+            // using the following format:
+            //   {endpointId}_Pri{priority}
+            // We use "_Pri" as a delimiter to parse the endpoint ID and priority
+            // back out for metrics reporting
+            string[] tokens = id.Split(new string[] { "_Pri" }, System.StringSplitOptions.RemoveEmptyEntries);
```
dylanbronson: If a user enters an endpoint with a name "foo_Pribar" then the metrics won't get reported. Not a likely case, but something to keep in mind

richma-ms: Yea, to really fix that would involve a lot more plumbing though.  Let's come back in a different PR to address it.

6607970 - Plumb endpoint ID and priority to Checkpointer for accurate metrics reporting

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
varunpuranik: We keep all objects as immutable.. this breaks that pattern. Notice the same pattern with this object. We should keep the same pattern.

varunpuranik: Also, why can't we store the message priority and then load it as well?

PR #2735, opened 2020-03-24
Priority Queue End-To-End test

dylanbronson:
First end-to-end test for priority queues.
This test uses the new e2e test framework that deploys to an actual edge device that the user can specify. 

The test is fairly simple in concept. 
1. Start up a sender (I reused loadGen in this case) and have it send messages along 4 different routes, each with a different priority. It will pick randomly which route to send each message. 
2. When it's done sending, sort the results by priority and then by sequence number, and send that sorted list to TestResultCoordinator (TRC).
3. Start up a receiver module (I reused Relayer in this case) that will begin receiving the messages from the sender. Each time we receive a message, send it to the TRC.
4. After a bit of time, call the report generation endpoint on the TRC to receive a CountingReport. This report matches up messages from the Sender and the Receiver to make sure they are exact matches.
5. Verify the report passed. Since the Sender sent an expected list of the sorted values, the receiver will have to receive them in the exact right order to get a passing result.

This test is module-to-module right now. It is easier to get an e2e test up and running without involving IoTHub. It will also be easier to write the test that involves hub now that we have this one written.

I've reused TRC without having to change anything from its side.
This is fine for now, but in the future, we should clean up TRC to be more generic. Right now, it is designed for connectivity tests, so I had to fill in a lot of the parameters with garbage values. Also, when it starts up, its TestReportingService will fail to start up with an error message, but it still works. Changing that to be optional will also be a part of making TRC more generic.


philipktlin:
1. Tests need to run on all tier 1 platforms including linux amd64, arm32 and arm64; and windows10 x64, Server Core x64, IoT Core x64.  Currently new E2E test framework needs to expand.
2. The approach you used to queue up messages for testing is to not running receiver module.  This approach is fine. However, we should also test when edge hub is processing a large amount of messages; a new higher priority message should be processed ahead; to ensure that message processing will check priority queues not just startup, but during processing.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    [EndToEnd]
+    public class PriorityQueues : SasManualProvisioningFixture
+    {
+        [Test]
+        public async Task PriorityQueueModuleToModuleMessages()
+        {
+            CancellationToken token = this.TestToken;
+            string trcImage = Context.Current.TestResultCoordinatorImage.Expect(() => new ArgumentException("testResultCoordinatorImage parameter is required for Priority Queues test"));
+            string loadGenImage = Context.Current.LoadGenImage.Expect(() => new ArgumentException("loadGenImage parameter is required for TempFilter test"));
```
philipktlin:
```suggestion
            string loadGenImage = Context.Current.LoadGenImage.Expect(() => new ArgumentException("loadGenImage parameter is required for priority queue test"));
```

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+            string relayerImage = Context.Current.RelayerImage.Expect(() => new ArgumentException("relayerImage parameter is required for TempFilter test"));
```
philipktlin:
```suggestion
            string relayerImage = Context.Current.RelayerImage.Expect(() => new ArgumentException("relayerImage parameter is required for Priority Queues test"));
```

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+
+                       .WithDesiredProperties(new Dictionary<string, object>
+                       {
+                           ["reportMetadataList"] = new Dictionary<string, object>
+                           {
+                               ["reportMetadata1"] = new Dictionary<string, object>
+                               {
+                                   ["TestReportType"] = "CountingReport",
+                                   ["TestOperationResultType"] = "Messages",
+                                   ["ExpectedSource"] = "loadGenModule.send",
```
philipktlin: Update other parameter to refer to variable if possible.
```suggestion
                                   ["ExpectedSource"] = $"{loadGenModuleName}.send",
```

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+                    builder.AddModule(trcModuleName, trcImage)
+                       .WithEnvironment(new[]
+                       {
+                           ("trackingId", trackingId),
+                           ("eventHubConnectionString", "Unnecessary"),
+                           ("IOT_HUB_CONNECTION_STRING", this.iotHub.IoTHubConnectionString),
+                           ("logAnalyticsWorkspaceId", "Unnecessary"),
+                           ("logAnalyticsSharedKey", "Unnecessary"),
+                           ("logAnalyticsLogType", "Unnecessary"),
+                           ("testStartDelay", "00:00:00"),
```
philipktlin: why you don't have test start delay? what if edgeHub starts after loadGen module, loadGen module will fail, right?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+                    // This test uses the TestResultCoordinator. It was originally designed for connectivity tests, so many of the parameters
+                    // are unnecessary for the e2e tests.
+                    // TODO: Make TestResultCoordinator more generic, so we don't have to fill out garbage values in the e2e tests.
+                    builder.AddModule(trcModuleName, trcImage)
+                       .WithEnvironment(new[]
+                       {
+                           ("trackingId", trackingId),
+                           ("eventHubConnectionString", "Unnecessary"),
+                           ("IOT_HUB_CONNECTION_STRING", this.iotHub.IoTHubConnectionString),
+                           ("logAnalyticsWorkspaceId", "Unnecessary"),
```
philipktlin: I agreed these unnecessary parameters should not be provided; you should make it as optional in TRC settings and corresponding usages.  Are you going to handle in another PR?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+                                   ["TestDescription"] = "unnecessary"
+                               }
+                           }
+                       });
+                    builder.AddModule(loadGenModuleName, loadGenImage)
+                        .WithEnvironment(new[]
+                        {
+                            ("testResultCoordinatorUrl", trcUrl),
+                            ("senderType", "PriorityMessageSender"),
+                            ("trackingId", "e2eTestTrackingId"),
```
philipktlin: use same tracking id in TRC.
```suggestion
                            ("trackingId", trackingId),
```

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: Yes that was the plan - to change TRC in a separate PR

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: I probably should add it here anyway, but the way the e2e tests work currently, edgeHub and edgeAgent always come up first. Any deployment in this framework will be split into two deployments - edgeRuntime + anything else

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+                        });
+
+                    builder.GetModule(ModuleName.EdgeHub)
+                        .WithDesiredProperties(new Dictionary<string, object>
+                        {
+                            ["routes"] = new Dictionary<string, object>
+                            {
+                                ["LoadGenToRelayer1"] = new Dictionary<string, object>
+                                {
+                                    ["route"] = "FROM /messages/modules/" + loadGenModuleName + "/outputs/pri0 INTO BrokeredEndpoint('/modules/" + relayerModuleName + "/inputs/input1')",
```
philipktlin: suggestion: define a variable for this format and use string.format.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+                                    ["priority"] = 3
+                                }
+                            }
+                        });
+                });
+
+            EdgeDeployment deployment = await this.runtime.DeployConfigurationAsync(addInitialConfig, token);
+
+            // Wait for loadGen to send some messages
+            await Task.Delay(TimeSpan.FromSeconds(30));
```
philipktlin: Delay will be different in different platforms, is there any better approach to check deployment state?

test/modules/Relayer/Program.cs
```diff
@@ -52,6 +53,18 @@ static async Task Main(string[] args)
...
         }
 
+        static Task ExecuteWithRetry(Func<Task> func, Action<RetryingEventArgs> onRetry)
+        {
+            var transientRetryPolicy = RetryPolicy.DefaultExponential;
+            transientRetryPolicy.Retrying += (_, args) => onRetry(args);
+            return transientRetryPolicy.ExecuteAsync(func);
+        }
+
+        static void RetryingReportTestResult(RetryingEventArgs retryingEventArgs)
```
philipktlin: This method name should change; should be something related to logging retry event.

test/modules/Relayer/Program.cs
```diff
@@ -100,23 +113,30 @@ static async Task<MessageResponse> ProcessAndSendMessageAsync(Message message, o
...
                     SequenceNumber = sequenceNumber
                 };
-                await ModuleUtil.ReportTestResultAsync(testResultReportingClient, Logger, testResultReceived);
-                Logger.LogInformation($"Successfully received message: trackingid={trackingId}, batchId={batchId}, sequenceNumber={sequenceNumber}");
 
-                byte[] messageBytes = message.GetBytes();
-                var messageCopy = new Message(messageBytes);
-                messageProperties.ForEach(kvp => messageCopy.Properties.Add(kvp));
-                await moduleClient.SendEventAsync(Settings.Current.OutputName, messageCopy);
+                await ExecuteWithRetry(
```
philipktlin: why do we need retry for test result reporting?

test/modules/TestResultCoordinator/Controllers/ReportController.cs
```diff
@@ -39,13 +39,8 @@ public async Task<ContentResult> GetReportsAsync()
 
             return new ContentResult
             {
-                Content = this.AddManualRunReportingHeading(JsonConvert.SerializeObject(testResultReports, Formatting.Indented)) // explicit serialization needed due to the wrapping list
```
philipktlin: why do you remove this manual run reporting heading?  are you sure no one uses it?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+                {
+                    // This test uses the TestResultCoordinator. It was originally designed for connectivity tests, so many of the parameters
+                    // are unnecessary for the e2e tests.
+                    // TODO: Make TestResultCoordinator more generic, so we don't have to fill out garbage values in the e2e tests.
+                    builder.AddModule(trcModuleName, trcImage)
+                       .WithEnvironment(new[]
+                       {
+                           ("trackingId", trackingId),
+                           ("eventHubConnectionString", "Unnecessary"),
+                           ("IOT_HUB_CONNECTION_STRING", this.iotHub.IoTHubConnectionString),
```
damonbarry: Use `Context.Current.ConnectionString` here instead of creating a new property on `IotHub`.

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -0,0 +1,88 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Client;
+    using Microsoft.Extensions.Logging;
+
+    class PriorityMessageSender : SenderBase
+    {
+        readonly string[] outputs = new string[] { "pri0", "pri1", "pri2", "pri3" };
```
philipktlin: suggestion: just define the prefix "pri" and define the upper bound of priority instead of hard-coded it on line 37.

test/Microsoft.Azure.Devices.Edge.Test.Common/IotHub.cs
```diff
@@ -62,6 +62,8 @@ public IotHub(string iotHubConnectionString, string eventHubEndpoint, Option<Uri
                 });
         }
 
+        public string IoTHubConnectionString => this.iotHubConnectionString;
```
damonbarry: See my comment in `PriorityQueues.cs`...we don't need to add this property here.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+                                },
+                                ["LoadGenToRelayer3"] = new Dictionary<string, object>
+                                {
+                                    ["route"] = "FROM /messages/modules/" + loadGenModuleName + "/outputs/pri2 INTO BrokeredEndpoint('/modules/" + relayerModuleName + "/inputs/input1')",
+                                    ["priority"] = 2
+                                },
+                                ["LoadGenToRelayer4"] = new Dictionary<string, object>
+                                {
+                                    ["route"] = "FROM /messages/modules/" + loadGenModuleName + "/outputs/pri3 INTO BrokeredEndpoint('/modules/" + relayerModuleName + "/inputs/input1')",
+                                    ["priority"] = 3
```
richma-ms: >["priority"] = 3 [](start = 36, length = 16)

might be a good idea to have a least one priority value that's not sequential

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -0,0 +1,88 @@
...
+                (Settings.Current.TestDuration == TimeSpan.Zero || DateTime.UtcNow - testStartAt < Settings.Current.TestDuration))
+            {
+                try
+                {
+                    int priority = this.rng.Next(4);
+                    string output = this.outputs[priority];
+
+                    await this.SendEventAsync(messageIdCounter, output);
+
+                    // We need to set the first message because of the way priority queue logic works
```
philipktlin: @richma-ms we didn't mention this behavior in design spec.  How difficult to change this behavior (always retyr first message sent while offline)?

test/Microsoft.Azure.Devices.Edge.Test/helpers/Context.cs
```diff
@@ -157,6 +160,12 @@ public Context()
 
         public Option<string> MetricsValidatorImage { get; }
 
+        public Option<string> TestResultCoordinatorImage { get; }
+
+        public Option<string> LoadGenImage { get; }
+
+        public Option<string> RelayerImage { get; }
```
damonbarry: You need to document these new context.json parameters in `test/README.md`.

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -0,0 +1,88 @@
...
+            Guid batchId,
+            string trackingId)
+            : base(logger, moduleClient, batchId, trackingId)
+        {
+        }
+
+        public async override Task RunAsync(CancellationTokenSource cts, DateTime testStartAt)
+        {
+            bool firstMessageWhileOffline = true;
+            var priorityAndSequenceList = new List<(int, long)>();
```
philipktlin: use dictionary<int,List<long>>

test/modules/load-gen/PriorityMessageSender.cs
richma-ms: as per the other comment, we probably shouldn't assume sequentially incremental priority values in the testcase, since the product code doesn't make that assumption either.

---
In reply to: [397495161](https://github.com/Azure/iotedge/pull/2735#discussion_r397495161) [](ancestors = 397495161)

test/modules/load-gen/SenderBase.cs
```diff
@@ -0,0 +1,79 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Client;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    public abstract class SenderBase
```
philipktlin: be more specific
```suggestion
    public abstract class LoadGenSenderBase
```

test/modules/load-gen/SenderBase.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        public SenderBase(
+            ILogger logger,
+            ModuleClient moduleClient,
+            Guid batchId,
+            string trackingId)
+        {
+            this.Logger = logger;
```
philipktlin: Preconditions.CheckNotNull

test/modules/load-gen/PriorityMessageSender.cs
richma-ms: the behavior is not always retry the first message, but rather whichever message loop happens to be sending when connectivity went down.  Due to the way this test is setup, it ends up being the first message because connectivity is already down by the time the first message gets enqueued.

Changing this is very difficult, as it'd involve fundamentally changing how retry and message checkpointing works.  We can't promise 100% accurate priority order anyways, even if we reworked retry, there will still be race conditions that exists depending on where code execution is at when connectivity gets cut/resumed.  So what we have now is best effort for the least development cost, and as far as we know it's good enough for the customer scenarios that we have.

---
In reply to: [397497250](https://github.com/Azure/iotedge/pull/2735#discussion_r397497250) [](ancestors = 397497250)

test/modules/load-gen/SenderBase.cs
```diff
@@ -0,0 +1,79 @@
...
+            this.Client = Preconditions.CheckNotNull(moduleClient, nameof(moduleClient));
+            this.BatchId = batchId;
```
philipktlin: check not empty for batch id

test/modules/load-gen/SenderBase.cs
```diff
@@ -0,0 +1,79 @@
...
+        protected async Task ReportResult(long messageIdCounter)
+        {
+            await Settings.Current.TestResultCoordinatorUrl.ForEachAsync(
+                async trcUrl =>
+                {
+                    var testResultCoordinatorUrl = new Uri(trcUrl, UriKind.Absolute);
+                    var testResultReportingClient = new TestResultReportingClient { BaseUrl = testResultCoordinatorUrl.AbsoluteUri };
+                    var testResult = new MessageTestResult(Settings.Current.ModuleId + ".send", DateTime.UtcNow)
+                    {
+                        TrackingId = Settings.Current.TrackingId,
```
philipktlin:
```suggestion
                        TrackingId = this.TrackingId,
```

test/modules/load-gen/PriorityMessageSender.cs
richma-ms: I think SortedDictionary<int, List<long>> since the priorities need to be in ascending order for later below.  This way there's no need to re-sort later, and it just becomes list concatenation of each priority.

---
In reply to: [397497876](https://github.com/Azure/iotedge/pull/2735#discussion_r397497876) [](ancestors = 397497876)

test/modules/load-gen/SenderBase.cs
```diff
@@ -0,0 +1,79 @@
...
+        public SenderBase(
+            ILogger logger,
+            ModuleClient moduleClient,
+            Guid batchId,
+            string trackingId)
+        {
+            this.Logger = logger;
+            this.Client = Preconditions.CheckNotNull(moduleClient, nameof(moduleClient));
+            this.BatchId = batchId;
+            this.TrackingId = trackingId;
```
philipktlin: do you need to ensure tracking id is not null?

test/modules/load-gen/SenderType.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace LoadGen
+{
+    public enum SenderType
```
philipktlin:
```suggestion
    public enum LoadGenSenderType
```

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -0,0 +1,88 @@
...
+            this.Logger.LogInformation($"Sending finished. Now sending expected results to {Settings.Current.TestResultCoordinatorUrl}");
+
+            // Sort by priority then sequence number. Then, select just the sequence numbers
+            List<long> expectedSequenceNumberList = priorityAndSequenceList
+                .OrderBy(t => t.Item1)
+                .ThenBy(t => t.Item2)
+                .Select(t => t.Item2)
+                .ToList();
+
+            await this.ReportResult(1);
```
richma-ms: >await this.ReportResult(1); [](start = 12, length = 27)

nit: maybe add a comment referencing the above explanation of special-casing the first message

test/modules/load-gen/SenderType.cs
```diff
@@ -0,0 +1,9 @@
...
+    {
+        DefaultSender,
```
philipktlin:
```suggestion
        DefaultMessageSender,
```

test/modules/load-gen/PriorityMessageSender.cs
philipktlin: Sure; then please define a set of priority in Settings class and allow it to be passing in.

philipktlin: Sure; then you may need to define a list of priority in settings class and allow to pass in.

philipktlin:
Please test both connectivity and your new e2e test to make sure it works.

test/modules/Relayer/Program.cs
dylanbronson: I actually don't think we do anymore. Artifact from testing. Removed.

test/modules/TestResultCoordinator/Controllers/ReportController.cs
dylanbronson: It was only used while we were debugging to add a message before the content. But we don't need it, and it makes it harder to parse on the other side.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
philipktlin: How can you ensure that within 30 secs, all given priorities will be used at least one to send messages?

dylanbronson: DeployConfigurationAsync waits until all modules are up and running. Therefore, we know that loadGen is sending messages when we have this delay. The delay is a bit longer than the testDuration of LoadGen, so we know that LoadGen will be finished sending messages. We wait a bit of extra time for loadGen to send messages to the TRC. I think this should work on all platforms. If not, we can bump up the last delay time, which seems to be the only place where the bottleneck could be.

dylanbronson: To your second comment, we can't ensure that all priorities will be used at least once, but most probably they will be since they're chosen randomly. We can bump up the number of messages to be almost certain. 

But it's also not even essential that every route is used. The test will still pass, and the behavior will be correct.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: Nice I was looking for this

dylanbronson:
@philipktlin to answer your questions above:

1. As far as I know, we will be migrating our tests over to that framework. Including all platforms. So I don't want to be building on the old framework knowing that we will be migrating to the new
	
2. This was the simplest possible e2e test I could get running. After this, I plan on building the canonical case of the network going offline, messages piling up, and then messages sending. The one you mentioned seems to be one of the harder cases, so I will get to it after the other two

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,129 @@
...
+                                   ["TestReportType"] = "CountingReport",
+                                   ["TestOperationResultType"] = "Messages",
+                                   ["ExpectedSource"] = $"{loadGenModuleName}.send",
+                                   ["ActualSource"] = $"{relayerModuleName}.receive",
+                                   ["TestDescription"] = "unnecessary"
+                               }
+                           }
+                       });
+
+                    string priorityString = "182;0;8000;15";
```
philipktlin: you should also test for sending message without priority.  If it is difficult to cover in this PR, please make sure cover it in another test.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
philipktlin: Since you mentioned the relationship between "testDuration" and this delay value; please derive this delay value based on "testDuration".  For example. TimeSpan.Parse(testDuration).Add(TimeSpan.FromSeconds(10)).

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -0,0 +1,105 @@
...
+            ModuleClient moduleClient,
+            Guid batchId,
+            string trackingId)
+            : base(logger, moduleClient, batchId, trackingId)
+        {
+        }
+
+        public async override Task RunAsync(CancellationTokenSource cts, DateTime testStartAt)
+        {
+            string priorityString = Settings.Current.Priorities.Expect(() =>
```
philipktlin: Move this to constructor

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
philipktlin: nit:To be safe, you can throw an exception in PriorityMessageSender when only 1 priority is used.

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -0,0 +1,105 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Client;
+    using Microsoft.Extensions.Logging;
+
+    class PriorityMessageSender : LoadGenSenderBase
+    {
+        // readonly string[] outputs = new string[] { "pri0", "pri1", "pri2", "pri3" };
```
philipktlin: remove

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            bool firstMessageWhileOffline = true;
+            var priorityAndSequence = new SortedDictionary<int, List<long>>();
+            long messageIdCounter = 1;
+            while (!cts.IsCancellationRequested &&
+                (Settings.Current.TestDuration == TimeSpan.Zero || DateTime.UtcNow - testStartAt < Settings.Current.TestDuration))
+            {
+                try
+                {
+                    int choosePri = this.rng.Next(4);
```
philipktlin:
```suggestion
                    int choosePri = this.rng.Next(outputs.Length);
```

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+        }
+
+        public async override Task RunAsync(CancellationTokenSource cts, DateTime testStartAt)
+        {
+            string priorityString = Settings.Current.Priorities.Expect(() =>
+                new ArgumentException("PriorityMessageSender must have 'priorities' environment variable set to a valid list of string delimited by ';'"));
+            string[] outputs = priorityString
+                .Split(';')
+                .Select(x => "pri" + x)
```
philipktlin: Append "pri" when calling SendEventAsync

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+
+            this.Logger.LogInformation($"Sending finished. Now sending expected results to {Settings.Current.TestResultCoordinatorUrl}");
+
+            // Sort priority by sequence number
+            List<long> expectedSequenceNumberList = priorityAndSequence
+                .SelectMany(t =>
+                {
+                    t.Value.Sort();
```
philipktlin: Sort is not required; https://stackoverflow.com/questions/1043039/does-listt-guarantee-insertion-order.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,135 @@
...
+                                    ["route"] = "FROM /messages/modules/" + loadGenModuleName + "/outputs/pri1 INTO BrokeredEndpoint('/modules/" + relayerModuleName + "/inputs/input1')",
+                                    ["priority"] = 1
+                                },
+                                ["LoadGenToRelayer3"] = new Dictionary<string, object>
+                                {
+                                    ["route"] = "FROM /messages/modules/" + loadGenModuleName + "/outputs/pri2 INTO BrokeredEndpoint('/modules/" + relayerModuleName + "/inputs/input1')",
+                                    ["priority"] = 2
+                                },
+                                ["LoadGenToRelayer4"] = new Dictionary<string, object>
+                                {
```
varunpuranik: We should have a route with no/default priority

test/Microsoft.Azure.Devices.Edge.Test/Microsoft.Azure.Devices.Edge.Test.csproj
```diff
@@ -29,6 +29,7 @@
     <ProjectReference Include="..\..\edge-util\src\Microsoft.Azure.Devices.Edge.Util\Microsoft.Azure.Devices.Edge.Util.csproj" />
     <ProjectReference Include="..\..\edge-util\test\Microsoft.Azure.Devices.Edge.Util.Test.Common\Microsoft.Azure.Devices.Edge.Util.Test.Common.csproj" />
     <ProjectReference Include="..\Microsoft.Azure.Devices.Edge.Test.Common\Microsoft.Azure.Devices.Edge.Test.Common.csproj" />
+    <ProjectReference Include="..\modules\ModuleLib\Microsoft.Azure.Devices.Edge.ModuleUtil.csproj" />
```
damonbarry: I don't see where `ModuleUtils` is being used in the test project. Can it be removed?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Test
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Net.Http;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
```
damonbarry: See comment in csproj file...can this be removed?

test/Microsoft.Azure.Devices.Edge.Test/Microsoft.Azure.Devices.Edge.Test.csproj
dylanbronson: It's being used for TestConstants.PriorityQueues.Default as a way to have the source of truth for the "Default" keyword be shared across modules (e.g. this framework + loadGen module). If we continue bringing in other modules, ModuleUtil is a good place to put common logic like this

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
damonbarry: Ah...`ModuleUtils.TestConstants`...I see it now thanks

test/Microsoft.Azure.Devices.Edge.Test/Microsoft.Azure.Devices.Edge.Test.csproj
damonbarry: Nvm

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: Added

philipktlin:
Please get approval from Damon as well.

test/modules/load-gen/LoadGenSenderBase.cs
```diff
@@ -23,7 +23,7 @@ public abstract class LoadGenSenderBase
             this.Logger = Preconditions.CheckNotNull(logger, nameof(logger));
             this.Client = Preconditions.CheckNotNull(moduleClient, nameof(moduleClient));
             this.BatchId = Preconditions.CheckNotNull(batchId, nameof(batchId));
-            this.TrackingId = Preconditions.CheckNonWhiteSpace(trackingId, nameof(trackingId));
+            this.TrackingId = trackingId;
```
philipktlin: Even tracking id is not mandatory, it is better to not null.  btw, just a little sharing, do you know "??=" operator; read https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator if interested.
```suggestion
            this.TrackingId = trackingId ?? string.Empty;
```

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,143 @@
...
+                                   ["TestReportType"] = "CountingReport",
+                                   ["TestOperationResultType"] = "Messages",
+                                   ["ExpectedSource"] = $"{loadGenModuleName}.send",
+                                   ["ActualSource"] = $"{relayerModuleName}.receive",
+                                   ["TestDescription"] = "unnecessary"
+                               }
+                           }
+                       });
+
+                    string priorityString = "182;0;8000;15;Default";
```
philipktlin: what is "Default"? should it be "TestConstants.PriorityQueues.Default" constant?
```suggestion
                    string priorityString = $"182;0;8000;15;{TestConstants.PriorityQueues.Default}";
```

philipktlin: Should this be configurable? since Richard said that we are going to limit range of priorities.  your call.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/Microsoft.Azure.Devices.Edge.Util.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="App.Metrics.Formatters.Prometheus" Version="3.0.0" />
     <PackageReference Include="App.Metrics.Reporting.InfluxDB" Version="3.0.0" />
     <PackageReference Include="App.Metrics.Reporting.TextFile" Version="3.0.0" />
+    <PackageReference Include="Microsoft.AspNet.WebApi.Client" Version="5.2.7" />
```
philipktlin: why need this?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: I don't want anyone who's using these tests to have to worry about configuring the priorities. E2E tests have a lot of set up and are designed to be used with "one click"

I've changed it to have randomly generated priorities from 0-10

edge-util/src/Microsoft.Azure.Devices.Edge.Util/Microsoft.Azure.Devices.Edge.Util.csproj
dylanbronson: Not sure how that got there. Removing.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,168 @@
...
+                       .WithEnvironment(new[]
+                       {
+                           ("trackingId", trackingId),
+                           ("eventHubConnectionString", "Unnecessary"),
+                           ("IOT_HUB_CONNECTION_STRING", Context.Current.ConnectionString),
+                           ("logAnalyticsWorkspaceId", "Unnecessary"),
+                           ("logAnalyticsSharedKey", "Unnecessary"),
+                           ("logAnalyticsLogType", "Unnecessary"),
+                           ("testStartDelay", "00:00:00"),
+                           ("testDuration", "00:20:00"),
```
philipktlin:
```suggestion
                           ("testDuration", loadGenTestDuration),
```

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -0,0 +1,168 @@
...
+            CancellationToken token = this.TestToken;
+            string trcImage = Context.Current.TestResultCoordinatorImage.Expect(() => new ArgumentException("testResultCoordinatorImage parameter is required for Priority Queues test"));
+            string loadGenImage = Context.Current.LoadGenImage.Expect(() => new ArgumentException("loadGenImage parameter is required for Priority Queues test"));
+            string relayerImage = Context.Current.RelayerImage.Expect(() => new ArgumentException("relayerImage parameter is required for Priority Queues test"));
+
+            const string trcModuleName = "testResultCoordinator";
+            const string loadGenModuleName = "loadGenModule";
+            const string relayerModuleName = "relayerModule";
+            const string trcUrl = "http://" + trcModuleName + ":5001";
+            const string loadGenTestDuration = "00:00:20";
```
philipktlin: just to confirm 20 secs or 20 mins as I see this line 55 "("testDuration", "00:20:00"),".  Should they be the same? if so, uses this const in "testDuration" line 55.

dylanbronson: They should be different. TRC's is in order to make it very large so the reporting service never runs.

PR #2746, opened 2020-03-27
Manually run end-to-end pipeline without arguments

damonbarry:
This change updates the new end-to-end test pipeline to use pipeline triggers instead of (deprecated) build completion triggers. There are a few benefits:
- We're using the officially recommended/supported trigger
- It requires less code in the YAML definition than what's there now
- It still allows you to run the build manually and specify which "Build Images" and "Edgelet Packages" builds to grab artifacts from (e.g. if you've made changes to iotedged or edge agent, and you want to run end-to-end tests against your changes before you merge your PR), but you're no longer _required_ to supply build IDs. So for instance, if you want to try out the new end-to-end test you're writing, just run the build against your PR without any arguments and it will use the latest artifacts from master builds.

### DIFF
--- a/builds/e2e/e2e.yaml
+++ b/builds/e2e/e2e.yaml
@@ -4,13 +4,13 @@ pr: none
 resources:
   pipelines:
   - pipeline: images
-    source: ${{ variables.az.pipeline.images }}
+    source: $[variables.az.pipeline.images]
     trigger:
       branches:
       - master
       - release/*
   - pipeline: packages
-    source: ${{ variables.az.pipeline.packages }}
+    source: $[variables.az.pipeline.packages]
     trigger:
       branches:
       - master
--- a/builds/e2e/e2e.yaml
+++ b/builds/e2e/e2e.yaml
@@ -18,33 +18,33 @@ resources:
 
 jobs:
 
-# ################################################################################	
-#   - job: linux_arm32v7
-# ################################################################################	
-#     displayName: Linux arm32v7
+################################################################################	
+  - job: linux_arm32v7
+################################################################################	
+    displayName: Linux arm32v7
 
-#     pool:
-#       name: $(pool.name)
-#       demands: rpi3-e2e-tests
+    pool:
+      name: $(pool.name)
+      demands: rpi3-e2e-tests
 
-#     variables:
-#       os: linux
-#       arch: arm32v7
-#       artifactName: iotedged-debian9-arm32v7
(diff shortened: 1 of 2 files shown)


builds/e2e/e2e.yaml
```diff
@@ -1,38 +1,50 @@
...
-
-################################################################################	
-  - job: linux_arm32v7
-################################################################################	
-    displayName: Linux arm32v7
-
-    pool:
-      name: $(pool.name)
-      demands: rpi3-e2e-tests
+resources:
```
philipktlin: Just to share, currently there is a problem the e2e test is triggered by image build, but not edgelet build or both.  To solve this problem, should we use multi-stage build? https://docs.microsoft.com/en-us/azure/devops/pipelines/process/stages?view=azure-devops&tabs=yaml

damonbarry: @philipktlin This build has two build completion triggers: image build _and_ packages build. It runs twice for every commit in master.

philipktlin: Ideally we should just run once for each commit with these 2 artifacts from image build and edgelet build. right?

damonbarry: Yes that's what I was hoping to achieve in this draft PR. Although now I'm thinking it might still run twice per commit because build images and edgelet packages seem to both run on every commit, not just on commits that impact the components they build. We may need to improve build images and edgelet packages to reduce the number of times we run this pipeline.

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -19,69 +19,48 @@ steps:
...
-    buildVersionToDownload: $(whichImageBuild)
-    branchName: refs/heads/master
+    project: $(resources.pipeline.images.projectID)
+    pipeline: $(resources.pipeline.images.pipelineName)
+    buildVersionToDownload: specific
     buildId: $(imageBuildId)
-    downloadType: specific
+    downloadType: single
+    artifactName: $(az.pipeline.images.artifacts)
     itemPattern: $(az.pipeline.images.artifacts)/artifactInfo.txt
```
philipktlin: nit:can "$(az.pipeline.images.artifacts)" be removed?

damonbarry: If I removed it, I'd have to replace it with the hard-coded value `core-linux`. Is that what you're suggesting?

PR #2750, opened 2020-03-28
Add TimeToLive to PriorityQueue

dylanbronson:
Adding TimeToLive to MessageStore and adding integ test.

With the addition of priority queues, we also now have the TTL for each route, instead of for each MessageStore. 

This PR changes the MessageStore to be aware of the TTL for each message and use the TTL for each message as the source of truth for TTL. If a message has no TTL specified, it gets a default value of '0.' Messages with default value of 0 take from the global TTL of the MessageStore.

Note: Because message stores are per endpoint per priority, there could be multiple routes that use the same MessageStore. When CleanupProcessor runs, it assumes that the head of a MessageStore (you can think of a MessageStore as a queue) is the oldest message and that all the messages have the same TTL. Therefore, it assumes that it doesn't need to check past the first unexpired message it finds. Now, it is possible to have messages in the same queue with different TTL's. This PR just allows the store to pile up with expired messages, so long as the head is unexpired. In a future PR, we will allow another option - to go through the entire list each time to find expired messages. It's a tradeoff between I/O and DB space, so we can make it configurable.


edge-hub/src/Microsoft.Azure.Devices.Routing.Core/IMessageStore.cs
```diff
@@ -3,6 +3,7 @@ namespace Microsoft.Azure.Devices.Routing.Core
 {
     using System;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
```
richma-ms: >using Microsoft.Azure.Devices.Edge.Util; [](start = 4, length = 40)

is this needed? The only new type I see in this file is uint

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/storage/MessageStoreTest.cs
```diff
@@ -297,6 +297,88 @@ public async Task CleanupTestTimeoutUpdateTimeToLive()
...
+
+                module1Iterator = messageStore.GetMessageIterator("module1");
+                batch = await module1Iterator.GetNext(100);
+                Assert.Equal(100, batch.Count());
+
+                module2Iterator = messageStore.GetMessageIterator("module2");
+                batch = await module2Iterator.GetNext(100);
+                Assert.Equal(100, batch.Count());
+
+                await Task.Delay(TimeSpan.FromSeconds(100));
```
richma-ms: >await Task.Delay(TimeSpan.FromSeconds(100)); [](start = 16, length = 44)

prob worthwhile to make module1 and module2 messages have different TTL, then wait out and difference and see that only one set of messages get iterated.  This will likely be the most frequent scenario.

edge-hub/test/Microsoft.Azure.Devices.Routing.Core.Test/endpoints/StoringAsyncEndpointExecutorTest.cs
```diff
@@ -463,7 +463,7 @@ public void Dispose()
             {
             }
 
-            public async Task<IMessage> Add(string endpointId, IMessage message)
+            public async Task<IMessage> Add(string endpointId, IMessage message, uint timeToLiveSecs)
```
richma-ms: >timeToLiveSecs [](start = 86, length = 14)

I think the convention is to use "_" for discarded params?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/IMessageStore.cs
dylanbronson: Remnant. Removed.

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/storage/MessageStoreTest.cs
dylanbronson: Makes sense. I've made that change

PR #2760, opened 2020-03-30
Fix windows supportbundle

lfitchett:
The timestamp for start time wasn't properly respecting timezonse. It is now passed in using rfc3339

Note Windows uses https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmlconvert.todatetime?view=netframework-4.8 to parse and linux uses http://man7.org/linux/man-pages/man1/journalctl.1.html

### DIFF
--- a/edgelet/iotedge/src/support_bundle.rs
+++ b/edgelet/iotedge/src/support_bundle.rs
@@ -281,8 +281,8 @@ where
          let inspect = ShellCommand::new("powershell.exe")
             .arg("-NoProfile")
             .arg("-Command")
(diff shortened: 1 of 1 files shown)


edgelet/iotedge/src/support_bundle.rs
```diff
@@ -268,24 +268,24 @@ where
             NaiveDateTime::from_timestamp(state.log_options.since().into(), 0),
             Utc,
         );
-        let since = since_time.format("%F %T").to_string();
 
         #[cfg(unix)]
```
dmolokanov: not every unix has systemd :)

PR #2761, opened 2020-03-30
Give Pi more time to download images

damonbarry:
With the introduction of #2736, our self-hosted Pi agent has to download docker images from scratch. The current timeout (8 min) isn't enough for the first test to download all images and run the test. This change increases the timeouts (setup and test) to 10 minutes, which should be enough.


philipktlin:
did you check temp filter func image that can be downloaded within 10 mins?

damonbarry:
> did you check temp filter func image that can be downloaded within 10 mins?

I haven't, thanks for the heads up. I'll check it out

PR #2762, opened 2020-03-30
Remove timing from PriorityQueue E2E tests

dylanbronson:
This PR makes the Priority Queue E2E test Timing-Proof.

Now, load-gen and Relayer will both notify the test when they are complete. In that way, we don't have to wait with delays that would need to be configured for each platform. 

We do this by using a Direct Method on both modules. We poll every five seconds after the module starts up to see if it has completed yet.


test/modules/ModuleLib/TestResults/PriorityQueueTestStatus.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults
+{
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public class PriorityQueueTestStatus
+    {
+        public PriorityQueueTestStatus(bool isFinished, int resultCount)
+        {
+            this.IsFinished = Preconditions.CheckNotNull(isFinished, nameof(isFinished));
+            this.ResultCount = Preconditions.CheckNotNull(resultCount, nameof(resultCount));
```
damonbarry, line 11: `bool` and `int` are both primitives...they're value objects and can never be null.

test/modules/Relayer/Program.cs
```diff
@@ -129,5 +153,19 @@ static async Task<MessageResponse> ProcessAndSendMessageAsync(Message message, o
 
             return MessageResponse.Completed;
         }
+
+        private static async Task SetIsFinishedDirectMethodAsync(ModuleClient client)
+        {
+            await client.SetMethodHandlerAsync(
+                "IsFinished",
+                async (MethodRequest methodRequest, object _) => await Task.FromResult(IsFinished()),
```
damonbarry, line 161: This doesn't need to be `async`/`await`. The following should work:
```
(MethodRequest methodRequest, object _) => Task.FromResult(IsFinished())
```

test/modules/Relayer/Program.cs
```diff
@@ -121,6 +131,20 @@ static async Task<MessageResponse> ProcessAndSendMessageAsync(Message message, o
                     await ModuleUtil.ReportTestResultAsync(testResultReportingClient, Logger, testResultSent);
                     Logger.LogInformation($"Successfully sent message: trackingid={trackingId}, batchId={batchId}, sequenceNumber={sequenceNumber}");
                 }
+                else
+                {
+                    int uniqueResultsExpected = int.Parse(
+                        Settings.Current.UniqueResultsExpected.Expect<ArgumentException>(() => throw new ArgumentException("Must supply this value if in ReceiveOnly mode")));
```
damonbarry, line 137: I don't think you need to be explicit about the template argument to `Expect`. The compiler should be able to figure it out from your closure.

test/modules/Relayer/Settings.cs
```diff
@@ -58,6 +65,8 @@ static Settings Create()
 
         public bool ReceiveOnly { get; }
 
+        public Option<string> UniqueResultsExpected { get; }
```
damonbarry, line 68: Seems like this should be an `Option<int>`

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -4,15 +4,21 @@ namespace LoadGen
...
     using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults;
     using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
 
     class PriorityMessageSender : LoadGenSenderBase
     {
         readonly Random rng = new Random();
+        private bool isFinished;
+        private int resultsSent;
```
damonbarry, line 21: In relayer, these values were thread safe. Do they need to be thread safe here too?

dylanbronson: I made them thread safe in Relayer because relayer uses a callback on receiving input to send messages. So the messages can be handled asynchronously. It also needs to keep track of previous results and handle duplicates.

PriorityMessageSender is all synchronous, so we don't need to add any thread safety.

test/modules/Relayer/Program.cs
damonbarry: No, if it's complaining, might as well use the exact type.

PR #2763, opened 2020-03-31
[mqttd] Added Dockerfile for ARM

vadim-kovalyov:
For our performance tests we would need to have both AMD64 and ARM32 images. Adding a Dockerfile to build ARM image.

This image is based on https://github.com/messense/rust-musl-cross

### DIFF
--- a/mqtt/Dockerfile.armv7
+++ b/mqtt/Dockerfile.armv7
@@ -1,0 +1,30 @@
+# -*- mode: dockerfile -*-
+#
+# An example Dockerfile showing how to build a Rust executable using this
+# image, and deploy it with a tiny Alpine Linux container.
+
+# You can override this `--build-arg BASE_IMAGE=...` to use different
+# version of Rust or OpenSSL.
+ARG BASE_IMAGE=messense/rust-musl-cross:armv7-musleabihf
+
+# Our first FROM statement declares the build environment.
+FROM ${BASE_IMAGE} AS builder
+
+# Add our source code.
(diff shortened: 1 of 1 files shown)


mqtt/Dockerfile.armv7
```diff
@@ -0,0 +1,30 @@
+# -*- mode: dockerfile -*-
+#
+# An example Dockerfile showing how to build a Rust executable using this
+# image, and deploy it with a tiny Alpine Linux container.
+
+# You can override this `--build-arg BASE_IMAGE=...` to use different
+# version of Rust or OpenSSL.
+ARG BASE_IMAGE=messense/rust-musl-cross:armv7-musleabihf
```
dmolokanov: Could you consult with @arsing to resolve an issue you face with backtrace-sys crate for arm32 musl build?

mqtt/Dockerfile.armv7
```diff
@@ -0,0 +1,30 @@
+# -*- mode: dockerfile -*-
```
dmolokanov: This docker file looks very similar to one we have already. And it supports `--build-arg BASE_IMAGE=...`

mqtt/Dockerfile.armv7
arsing: I assume this ignores the `rust-toolchain` file?

vadim-kovalyov:
After some internal discussion we'll abandon this in favor of #2771

PR #2766, opened 2020-03-31
Support bundle fixes into 1.0.9

lfitchett:
This fixes 2 bugs. First, large enough logs weren't being written to the bundle b/c write was used instead of writre_all. Second, windows didn't respect time zones properly. This fixes that.

### DIFF
--- a/edgelet/iotedge/src/support_bundle.rs
+++ b/edgelet/iotedge/src/support_bundle.rs
@@ -232,8 +232,7 @@ where
          let inspect = ShellCommand::new("powershell.exe")
             .arg("-NoProfile")
             .arg("-Command")
-            .arg(&format!(r"$start=[Xml.XmlConvert]::ToDateTime('{}');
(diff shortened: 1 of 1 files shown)


edgelet/iotedge/src/support_bundle.rs
```diff
@@ -219,24 +219,24 @@ where
             NaiveDateTime::from_timestamp(state.log_options.since().into(), 0),
             Utc,
         );
-        let since = since_time.format("%F %T").to_string();
```
arsing: Is there a reason this fix was needed for `Get-WinEvent` but not for `Get-EventLog` used by `write_docker_log_to_file` ?

lfitchett: It might have been needed, I just didn't notice b/c the docker log has never been more than 1 or 2 lines from when it was first started. I'll go ahead and make a change for that too.

PR #2769, opened 2020-03-31
Content trust twin fields in module and module spec

ggjjj:
The design calls out for four fields of content trust information.
1. path to the root.json file
2. root ID 
3. path to the root CA file
4. flag to disableTOFU - Trust On First Use

The PR makes changes per module and the modulespec which gets used by the iotedged


ggjjj:
Closing this as the implemented design can be improved by changing DockerConfig instead of IModule

PR #2771, opened 2020-04-01
[mqtt] Added support for building and publishing MQTT broker images

vadim-kovalyov:
This is a stop-gap solution to enable us to build and publish MQTT broker images for performance tests. It will be reworked when MQTT will be integrated with EdgeHub.

### DIFF
--- a/mqtt/build/linux/buildImage.sh
+++ b/mqtt/build/linux/buildImage.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+# This script copies the iodedged executable files that go into the azureiotedge-iotedged image,
+# for each supported arch (x86_64, arm32v7, arm64v8).
+# It then publishes executable files along with their corresponding dockerfiles to the publish directory,
+# so that buildImage.sh can build the container image.
+
+set -e
+###############################################################################
+# Define Environment Variables
+###############################################################################
+
+ARCH=$1
+DOCKER_REGISTRY=$2
+DOCKER_USERNAME=$3
+DOCKER_PASSWORD=$4
+DOCKER_IMAGEVERSION=${BUILD_BUILDNUMBER:=$5}
+
+if [[ "$ARCH" == "amd64" ]]; then
+    ARCH="amd64"
+    DOCKER_IMAGENAME="azureiotedge/mqttd-linux-amd64"
+    TARGET="x86_64-unknown-linux-musl"
+    DOCKERFILE="../../docker/linux/amd64/Dockerfile"
+elif [[ "$ARCH" == "armv7" ]]; then
+    ARCH="arm32v7"
+    DOCKER_IMAGENAME="azureiotedge/mqttd-linux-arm32v7"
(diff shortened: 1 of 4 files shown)


mqtt/build/linux/buildImage.sh
```diff
@@ -0,0 +1,54 @@
...
+# Define Environment Variables
+###############################################################################
+
+ARCH=$1
+DOCKER_REGISTRY=$2
+DOCKER_USERNAME=$3
+DOCKER_PASSWORD=$4
+DOCKER_IMAGEVERSION=${BUILD_BUILDNUMBER:=$5}
+
+if [[ "$ARCH" == "amd64" ]]; then
```
dmolokanov, line 19: Please make sure scripts you added has similar structure, arguments and values similar across all projects https://github.com/Azure/iotedge/blob/master/scripts/linux/buildEdgelet.sh#L31-L43

mqtt/build/linux/buildImage.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -e
+###############################################################################
+# Define Environment Variables
+###############################################################################
+
+ARCH=$1
+DOCKER_REGISTRY=$2
+DOCKER_USERNAME=$3
+DOCKER_PASSWORD=$4
```
dmolokanov, line 16: Don't think you ever need a password in the script

mqtt/build/linux/buildImage.sh
```diff
@@ -0,0 +1,54 @@
...
+DOCKER_IMAGEVERSION=${BUILD_BUILDNUMBER:=$5}
+
+if [[ "$ARCH" == "amd64" ]]; then
+    ARCH="amd64"
+    DOCKER_IMAGENAME="azureiotedge/mqttd-linux-amd64"
```
dmolokanov, line 21: You could probably should stick with other images naming convention

mqtt/build/linux/buildImage.sh
```diff
@@ -0,0 +1,54 @@
...
+    TARGET="x86_64-unknown-linux-musl"
+    DOCKERFILE="../../docker/linux/amd64/Dockerfile"
```
dmolokanov, line 23: I believe you need to change this path. Scripts in the pipeline run from root directory that's why our scripts have several  ENV https://github.com/Azure/iotedge/blob/master/scripts/linux/buildEdgelet.sh#L22-L24

mqtt/build/linux/buildImage.sh
```diff
@@ -0,0 +1,54 @@
...
+# for each supported arch (x86_64, arm32v7, arm64v8).
+# It then publishes executable files along with their corresponding dockerfiles to the publish directory,
+# so that buildImage.sh can build the container image.
+
+set -e
+###############################################################################
+# Define Environment Variables
+###############################################################################
+
+ARCH=$1
```
dmolokanov, line 13: I think we may want to have the same structure for parsing script arguments and have them with named keys

PR #2772, opened 2020-04-01
Test Dashboard: Edit config file to allow edits while provisioned

and-rewsmith:
Added the following:

```
# <bool> allow updating provisioned dashboards from the UI
  allowUiUpdates: true
```


tools/IoTEdgeDevOps/TestDashboard/dashboards/Home-1585707775101.json
```diff
@@ -2032,7 +2035,7 @@
           "colors": [
             "rgba(245, 54, 54, 0.9)",
             "rgba(237, 129, 40, 0.89)",
-            "rgba(50, 172, 45, 0.97)"
+            "#73BF69"
```
philipktlin, line 2038: why changed to color hex?

and-rewsmith: It is generated, I was poking around and messed with the color wheel.

PR #2773, opened 2020-04-01
TRC: Cap number of unmatched results

and-rewsmith:
The TRC will sometimes fail to generate a report due to too many unmatched results. This PR caps the quantity of unmatched results to 10.


test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -174,6 +171,15 @@ internal static async Task UploadLogsAsync(string iotHubConnectionString, Uri bl
             logger.LogInformation($"Upload logs was started at {uploadLogStartAt} and completed at {uploadLogFinishAt}; and took {uploadLogFinishAt - uploadLogStartAt}.");
         }
 
+        internal static void EnqueueAndEnforceFixedSize<T>(Queue<T> q, T result)
+        {
+            q.Enqueue(result);
+            if (q.Count > Settings.Current.UnmatchedResultsMaxSize)
```
philipktlin, line 177: Settings.Current.UnmatchedResultsMaxSize should be a method parameter.

test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -174,6 +171,15 @@ internal static async Task UploadLogsAsync(string iotHubConnectionString, Uri bl
             logger.LogInformation($"Upload logs was started at {uploadLogStartAt} and completed at {uploadLogFinishAt}; and took {uploadLogFinishAt - uploadLogStartAt}.");
         }
 
+        internal static void EnqueueAndEnforceFixedSize<T>(Queue<T> q, T result)
```
philipktlin, line 174:
```suggestion
        internal static void EnqueueAndEnforceSize<T>(Queue<T> q, T result, int size)
```

test/modules/TestResultCoordinator/Reports/CountingReportGenerator.cs
```diff
@@ -136,7 +136,7 @@ public async Task<ITestResultReport> CreateReportAsync()
                 totalExpectCount,
                 totalMatchCount,
                 totalDuplicateResultCount,
-                unmatchedResults.AsReadOnly());
+                new List<TestOperationResult>(unmatchedResults).AsReadOnly());
```
philipktlin, line 139: do we need to keep using List? is it ok to use queue in CountingReport? will this look the same in serialized json?

test/modules/TestResultCoordinator/TestReportUtil.cs
philipktlin, line 174: Or you can make it as extension method for Queue<T>

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -111,6 +114,7 @@ static Settings Create()
                 configuration.GetValue<bool>("logUploadEnabled", true),
                 configuration.GetValue<string>("STORAGE_ACCOUNT_CONNECTION_STRING"),
                 configuration.GetValue<string>(TestConstants.NetworkController.RunProfilePropertyName),
+                configuration.GetValue<ushort>("UNMATCHED_RESULTS_MAX_SIZE", 10),
```
philipktlin, line 117:
```suggestion
                configuration.GetValue<ushort>("UNMATCHED_RESULTS_MAX_SIZE", DefaultUnmatchedResultsMaxSize),
```

test/modules/TestResultCoordinator/Reports/CountingReportGenerator.cs
and-rewsmith: We pass it in as a readonly list. We could instead pass the queue directly, but it wouldn't be readonly. We could find a readonly queue implementation, but also need to confirm that this would still serialize well.

test/modules/TestResultCoordinator/Reports/DeploymentTestReportGenerator.cs
```diff
@@ -106,7 +106,7 @@ public async Task<ITestResultReport> CreateReportAsync()
 
             while (hasExpectedResult)
             {
-                unmatchedResults.Add(this.ExpectedTestResults.Current);
+                TestReportUtil.EnqueueAndEnforceFixedSize(unmatchedResults, this.ExpectedTestResults.Current, Settings.Current.UnmatchedResultsMaxSize);
```
philipktlin, line 109: nit: seems the line swap looks weird to me; either break all arguments into its own line or all in the same line.

and-rewsmith: I think that is just the rendering. It's on the same line.

and-rewsmith:
The tests are failing. Will fix and then notify you when rdy.

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -66,6 +68,7 @@ class Settings
             this.LogUploadEnabled = logUploadEnabled;
             this.StorageAccountConnectionString = Preconditions.CheckNonWhiteSpace(storageAccountConnectionString, nameof(storageAccountConnectionString));
             this.NetworkControllerType = this.GetNetworkControllerType(networkControllerRunProfileName);
+            this.UnmatchedResultsMaxSize = unmatchedResultsMaxSize;
```
philipktlin, line 71:
```suggestion
            this.UnmatchedResultsMaxSize = Preconditions.CheckRange<ushort>(unmatchedResultsMaxSize, 1);
```

test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -174,6 +171,15 @@ internal static async Task UploadLogsAsync(string iotHubConnectionString, Uri bl
             logger.LogInformation($"Upload logs was started at {uploadLogStartAt} and completed at {uploadLogFinishAt}; and took {uploadLogFinishAt - uploadLogStartAt}.");
         }
 
+        internal static void EnqueueAndEnforceFixedSize<T>(Queue<T> q, T result, ushort maxSize)
```
philipktlin, line 174:
```suggestion
        internal static void EnqueueAndEnforceMaxSize<T>(Queue<T> q, T result, ushort maxSize)
```

PR #2775, opened 2020-04-01
Enable windows host metrics

lfitchett:
This re-enables host metrics for windows. They previously worked, but had to be disabled since they broke the windows arm-32 build. It also updates the sysinfo library


edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -636,85 +631,67 @@ impl ModuleRuntime for DockerModuleRuntime {
...
+            let mut info: libc::sysinfo = unsafe { mem::zeroed() };
+            let ret = unsafe { libc::sysinfo(&mut info) };
+            if ret == 0 {
+                info.uptime.try_into().unwrap_or_default()
+            } else {
+                0
+            }
+        };
+        #[cfg(windows)]
+        let uptime: u64 = unsafe { winapi::um::sysinfoapi::GetTickCount() }.into();
```
arsing, line 645: Use `GetTickCount64`. `GetTickCount` wraps after 50 days.

arsing, line 645: This is in milliseconds whereas the Linux one is in seconds. Should they not be consistent?

lfitchett: Changed

lfitchett: yes

PR #2776, opened 2020-04-01
[mqtt] Added support for building and publishing MQTT broker images

vadim-kovalyov:
This is a stop-gap solution to enable us to build and publish MQTT broker images for performance tests. It will be reworked when MQTT will be integrated with EdgeHub.


mqtt/docker/linux/amd64/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Fix permissions on source code.
+# RUN sudo chown -R rust:rust /home/rust
+
+# Build our application.
+RUN cargo build -p mqttd --release && strip /home/rust/src/target/x86_64-unknown-linux-musl/release/mqttd
+
+FROM scratch
+ENV RUST_LOG=info
+EXPOSE 1883/tcp
```
philipktlin, line 24: why need this?

philipktlin, line 24: why need this?

mqtt/docker/linux/amd64/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+EXPOSE 8883/tcp
+
+COPY --from=builder \
+    /home/rust/src/target/x86_64-unknown-linux-musl/release/mqttd \
+    /usr/local/bin/
+ENTRYPOINT ["/usr/local/bin/mqttd"]
```
philipktlin, line 30: should we use non-root user to run container like edgeHub?

mqtt/docker/linux/amd64/Dockerfile
vadim-kovalyov: @philipktlin  do you mean `EXPOSE`? it is just a convention to let consumers know that the container exposes these ports. Do you have any concerns about that?

mqtt/docker/linux/amd64/Dockerfile
vadim-kovalyov: good point, I'll investigate...

mqtt/docker/linux/amd64/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# version of Rust or OpenSSL.
+ARG BASE_IMAGE=ekidd/rust-musl-builder:stable
+
+# Our first FROM statement declares the build environment.
+FROM ${BASE_IMAGE} AS builder
+
+# Add our source code.
+ADD . ./
+
+# Fix permissions on source code.
```
dmolokanov, line 16: Is it still an issue? Remove it please if not.

builds/misc/images-mqtt.yaml
```diff
@@ -0,0 +1,57 @@
...
+################################################################################
+    displayName: Linux AMD64
+    pool:
+      vmImage: 'ubuntu-latest'
+
+    variables:
+      imageName: $(registry.address)/azureiotedge/mqttd-linux-amd64
+
+    steps:
+      - bash: 'docker login $(registry.address) --username $(registry.user) --password $(registry.password)'
```
philipktlin, line 20: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/docker?view=azure-devops#login

builds/misc/images-mqtt.yaml
```diff
@@ -0,0 +1,57 @@
...
+        displayName: 'Docker Login'
+
+      - task: Docker@2
```
philipktlin, line 23: this is nice. :)

builds/misc/images-mqtt.yaml
```diff
@@ -0,0 +1,57 @@
...
+        displayName: Build amd64 image
+        inputs:
+          repository: $(imageName)
+          command: build
+          Dockerfile: mqtt/docker/linux/amd64/Dockerfile
+          buildContext: mqtt/.
+
+      - bash: 'docker push $(imageName)'
```
philipktlin, line 31: I think you can use Docker@2 task to push as well.

mqtt/docker/linux/amd64/Dockerfile
philipktlin: Do we need to use 1883 port? For mqtt in edgeHub, we use port 8883 only.

builds/misc/images-mqtt.yaml
vadim-kovalyov: Good point. I followed what we had in other pipelines, but using the fit-for-purpose task is better. I also don't see an existing service connection in Azure DevOps for EdgeBuilds container registry (@philipktlin do you have one?). I'll create one if we don't have.

mqtt/docker/linux/amd64/Dockerfile
vadim-kovalyov: Yes we need, it is MQTT no TLS port, which we will support.

mqtt/docker/linux/amd64/Dockerfile
vadim-kovalyov: fixed

builds/misc/images-mqtt.yaml
philipktlin: please go ahead to create one. :)

mqtt/docker/linux/amd64/Dockerfile
philipktlin: check last sentence in overview section in https://en.wikipedia.org/wiki/MQTT.  8883 is used for MQTT over SSL (using TLS) and 1883 is MQTT over TCP.  And edgeHub is only listening to 8883.  Should MQTT broker also listen to 8883 only?

builds/misc/images-mqtt.yaml
```diff
@@ -0,0 +1,67 @@
...
+jobs:
+################################################################################
+  - job: linux_amd64
+################################################################################
+    displayName: Linux AMD64
+    pool:
+      vmImage: 'ubuntu-latest'
+
+    variables:
+      imageName: azureiotedge/mqttd-linux-amd64
```
dmolokanov, line 13: Once the plan changed and you are pushing this image to edgebuilds, could you please make it in a format the rest of images follow?
The name can be more descriptive and contain tag that corresponds to a build
`azureiotedge/iotedge-mqtt-broker:{VERSION}-linux-{ARCH}`

mqtt/.dockerignore
```diff
@@ -3,6 +3,7 @@
 .git
 .gitignore
 .vscode
+state
```
dmolokanov, line 6: what app generates this folder?

mqtt/docker/linux/amd64/Dockerfile
philipktlin: Just want to confirm that when release mqtt-broker, port 1883 should not be needed, right? why do we want to run perf and integration test using 1883? @dmolokanov could you clarify?

mqtt/.dockerignore
vadim-kovalyov: mqttd generates this folder to persist state.

dmolokanov: should we add it to a `.gitignore` as well?

vadim-kovalyov: it is already there - https://github.com/Azure/iotedge/blob/master/mqtt/.gitignore

mqtt/docker/linux/amd64/Dockerfile
dmolokanov: Actually it will be used for some clients. Some devices are not powerful enough to support any encryption. It'll special case, but we want to support them.

philipktlin: Do you mean mqtt broker will support both ports, but IoT Edge will support 8883 only?  And would you please give some example of clients will use port 1883?

PR #2777, opened 2020-04-01
Allow environment variables to be empty or whitespace.

darobs:
Environment variable values may be empty or whitespace.
Docker allows these values for environment variable:
"key1="
"key2=   "


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/KubernetesModule.cs
```diff
@@ -340,6 +343,10 @@ protected override void Load(ContainerBuilder builder)
                     })
                 .As<Task<IEnvironmentProvider>>()
                 .SingleInstance();
+
+            // SystemResourcesMetrics
+            builder.Register(c => new Edgelet.Docker.SystemResourcesMetrics(c.Resolve<IMetricsProvider>(), c.Resolve<IModuleManager>().GetSystemResourcesAsync, this.apiVersion, this.performanceMetricsUpdateFrequency))
+                .SingleInstance();
```
and-rewsmith, line 349: Is this change related to the environment variables null or whitespace?

darobs: No, not directly. This was done so I could test it in k8s mode. I can remove it and put it into a different PR, if you'd rather have the PR have a narrow focus.

and-rewsmith: The change is small so I'm fine with it. The risk is that if we need to undo this environment var change or port it somewhere, we need to remember to decouple these.

darobs: Good point - there's a decent chance this will be back-ported. Removed the commit from this PR and will put out another for the other changes.

PR #2779, opened 2020-04-02
[mqtt] added clap for command line parameters

vipeller:
and moved existing parameters into the config file


mqtt/mqttd/src/main.rs
```diff
@@ -19,30 +20,26 @@ async fn main() -> Result<(), Terminate> {
         .finish();
     let _ = tracing::subscriber::set_global_default(subscriber);
 
-    // TODO pass it to broker
-    // TODO make it an argument to override defaul config
-    let path: Option<String> = None;
-    let config = path
+    let matches = App::new("Edge MQTT Broker")
```
vadim-kovalyov, line 23: nit: iotedged has this App and config init login in a separate module (https://github.com/Azure/iotedge/blob/master/edgelet/iotedged/src/app.rs). Not super important, just nit suggestion to keep main.rs smaller.

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -1,16 +1,59 @@
...
+use native_tls::Identity;
 use regex::Regex;
 use serde::{Deserialize, Deserializer};
+use tracing::info;
 
 pub const DEFAULTS: &str = include_str!("../config/default.json");
 
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Transport {
```
vadim-kovalyov, line 20: nit: I was a bit confused b/w this `Transport` and `transport::Transport`. Maybe `TransportConfig`?

mqtt/mqtt-broker/config/default.json
```diff
@@ -1,4 +1,17 @@
...
+    "transports": [
+        { 
+            "tcp": {
+                 "address": "0.0.0.0:1883"
+            } 
+        },
+        { 
+            "tls": { 
+                "address": "0.0.0.0:8883", 
+                "identity_path": "identity.pfx"
```
varunpuranik, line 11: Nit: Can we call this something like server_cert or so? Identity seems confusing here.

mqtt/mqtt-broker/src/configuration.rs
vipeller: being in 'configuration.rs', do you think it is needed? all the rest of the config sections are defined without the 'config' postfix.

mqtt/mqtt-broker/config/default.json
vipeller: then cert_path, server_cert_path, srv_cert_path? i think the path should be there

vipeller: renamed srv_cert_path

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -1,16 +1,59 @@
...
+    Tcp {
+        address: String,
```
dmolokanov, line 22: Can we make it something with some type that less generic than string?

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -1,16 +1,59 @@
...
+            }
+        }
+    }
+}
+
+fn load_identity(path: &Path) -> Result<Identity, Error> {
+    info!("Loading identity from {:?}", path);
+    let cert_buffer = std::fs::read(&path).context(ErrorKind::LoadIdentity)?;
+
+    let cert_pwd = "";
```
dmolokanov, line 50: I think this variable don't bring too much value. Shouldn't we inline it instead?

mqtt/mqtt-broker/src/error.rs
```diff
@@ -69,6 +69,9 @@ pub enum ErrorKind {
     #[fail(display = "An error occurred loading configuration.")]
     LoadConfiguration,
 
+    #[fail(display = "Configured TLS without specified identity path.")]
+    InconsistentTlsConfiguration,
```
dmolokanov, line 73: Do we still need this error?

mqtt/mqtt-broker/config/default.json
dmolokanov: I think we need more consistent here: you have address (full form), but srv_cert_path (several short words).
I suggest we call it `certificate` and in rust code it will be `certificate: PathBuff`. So no confusion about what is certificate type

mqtt/mqtt-broker/src/configuration.rs
vadim-kovalyov: I'm fine with it, was just nitpicling.

mqtt/mqtt-broker/config/default.json
```diff
@@ -4,12 +4,6 @@
             "tcp": {
                  "address": "0.0.0.0:1883"
             } 
-        },
```
vadim-kovalyov: thanks :)

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -1,16 +1,59 @@
...
 
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Transport {
+    Tcp {
+        address: String,
+    },
+    Tls {
+        address: String,
+        srv_cert_path: PathBuf,
```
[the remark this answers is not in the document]
vipeller: I don't like that no _path in the name, but renamed to certificate

mqtt/mqtt-broker/config/default.json
vipeller: in the config it is not immediately clear that it should be a path, but "certificate_path" would be maybe long. renamed to "certificate"

mqtt/mqttd/src/main.rs
vipeller: moved out to a method

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -35,21 +35,17 @@ impl TryFrom<Transport> for TransportBuilder<String> {
...
-            }
+                certificate,
+            } => load_identity(certificate.as_path()).map(|identity| Self::Tls(address, identity)),
         }
     }
 }
 
 fn load_identity(path: &Path) -> Result<Identity, Error> {
     info!("Loading identity from {:?}", path);
     let cert_buffer = std::fs::read(&path).context(ErrorKind::LoadIdentity)?;
```
dmolokanov, line 46: Could you please add PathBuf to an ErrorKind
```suggestion
    let cert_buffer = std::fs::read(&path).context(ErrorKind::LoadIdentity(path.to_path_buf())?;
```

mqtt/mqttd/src/main.rs
```diff
@@ -120,3 +109,19 @@ async fn tick_snapshot(
         }
     }
 }
+
+#[allow(deprecated)]
```
dmolokanov, line 113: There is another macro `app_from_crate` that does what we need. And doesn't contain deprecated code!
```rust 
fn create_app() -> App<'static, 'static> {
    app_from_crate!().arg(
        Arg::with_name("config")
            .short("c")
            .long("config")
            .value_name("FILE")
            .help("Sets a custom config file")
            .takes_value(true),
    )
}
```

mqtt/mqttd/Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name = "mqttd"
+description = "mqttd is an MQTT broker to allow local communication"
```
dmolokanov, line 3:
```suggestion
description = "mqttd is a MQTT broker for IoT Edge platform"
```
?

mqtt/mqttd/src/main.rs
vipeller: it is the create_authors! which deprecated. I can remove that, doesn't hold much information anyway.

PR #2780, opened 2020-04-02
Longhaul: Add test start delay to direct methods and twins

and-rewsmith:
We are seeing test failures at the beginning of longhaul because the direct method sender module spins up before the receiver. This fix will prevent it by adding a test start delay to the sender modules.


scripts/linux/runE2ETest.sh
```diff
@@ -1073,6 +1079,7 @@ TRANSPORT_TYPE_4="${TRANSPORT_TYPE_4:-mqtt}"
 TWIN_UPDATE_FAILURE_THRESHOLD="${TWIN_UPDATE_FAILURE_THRESHOLD:-00:01:00}"
 METRICS_SCRAPE_FREQUENCY_IN_SECS="${METRICS_SCRAPE_FREQUENCY_IN_SECS:-300}"
 METRICS_UPLOAD_TARGET="${METRICS_UPLOAD_TARGET:-AzureLogAnalytics}"
+TEST_START_DELAY="${TEST_START_DELAY:-00:02:00}"
```
philipktlin, line 1082: Previously LH start right away and no start delay is required, why do we need to set it (for DMSender) in LH?

and-rewsmith: Currently, dmsender will receive "internal server error" status code if it comes up before the receiver. I'm not sure why we are starting to see this now, but we have made some changes to the direct method modules since it was stable.

I don't see a problem using test start delay as opposed to a retry. Connectivity also uses the test start delay.

philipktlin: It is ok to have test start delay.  I am concerned that the sequence number is increased if there is an exception.  In LH, the assumption is to increase sequence number if operation succeeds, otherwise retry with same sequence number.  If LH didn't get any result longer than 1s (not 100% sure the duration).  It will show a message in result like a warning.  If this behavior is changed, our LH test will fail if flaky network causes DM call failed once; then test analyzer will report missing records and failed test.  Correct me if my understanding is incorrect.

and-rewsmith: Two things:
1. There is no sequence number for direct method tests. It sends status reports with pass/fail.
2. This problem only occurs at the beginning of longhaul. I think that having this problem should fail the tests if it occurs when both modules are online.

philipktlin: 1. Sequence number is in directMethodCount variable. https://github.com/Azure/iotedge/blob/c3efe9185bc2a7063748e0cc07ee3b806cdbc416/test/modules/DirectMethodSender/DirectMethodSenderBase.cs#L35.  And it reports in test result, https://github.com/Azure/iotedge/blob/c3efe9185bc2a7063748e0cc07ee3b806cdbc416/test/modules/DirectMethodReceiver/src/DirectMethodReceiver.cs#L52

However test analyzer generating LH DM report doesn't care about sequence number.

Last question: do we expect LH to fail due to DM call failed causing by flaky network drop within very short period?

and-rewsmith: Ideally this shouldn't fail the test. I haven't seen a status code indicating network issues in the report. If we do see it, we should implement failure tolerance based on network exception type.

and-rewsmith:
I'm going split the start delay into one for each amd64/arm32

and-rewsmith:
This issue comes up sometimes after status code 0, so this PR won't solve the issue and I am closing. I think that the correct approach will be to handle this the same way we handle status code 0 (i.e. ignore) until the SDK backlog item allows us to fix this in edgehub.

and-rewsmith:
If we ever need a test start delay for longhaul we can work from this branch

builds/e2e/longhaul.yaml
```diff
@@ -14,6 +14,8 @@ variables:
   #container.registry: ''
   #snitch.storage.account: ''
   # Variables settable at queue time
+  #test.startDelay.arm
+  #test.startDelay.amd
```
philipktlin, line 18: Should we model startDelay to 2 value- 1 for 32 bits and 1 for 64 bits? or model 1 for short (test.startDelay.short) and 1 for long (test.startDelay.long)?  If you model use architecture, we will need to add arm64 in the future.

builds/e2e/longhaul.yaml
```diff
@@ -418,6 +424,7 @@ jobs:
           release.label: 'servercore-lh$(agent.group)'
           test.buildNumber: '$(Build.BuildNumber)'
           test.buildId: '$(Build.BuildId)'
+          test.startDelay: '$(test.startDelay)'
```
philipktlin, line 427: missing .amd

scripts/linux/runE2ETest.sh
```diff
@@ -1070,14 +1076,14 @@ TRANSPORT_TYPE_1="${TRANSPORT_TYPE_1:-amqp}"
...
 TRANSPORT_TYPE_4="${TRANSPORT_TYPE_4:-mqtt}"
-TWIN_UPDATE_FAILURE_THRESHOLD="${TWIN_UPDATE_FAILURE_THRESHOLD:-00:01:00}"
-METRICS_SCRAPE_FREQUENCY_IN_SECS="${METRICS_SCRAPE_FREQUENCY_IN_SECS:-300}"
-METRICS_UPLOAD_TARGET="${METRICS_UPLOAD_TARGET:-AzureLogAnalytics}"
 
 if [[ "${TEST_NAME,,}" == "longhaul" ]] ||
    [[ "${TEST_NAME,,}" == "stress"  ]]; then
     tracking_id=$(cat /proc/sys/kernel/random/uuid)
     TEST_INFO="$TEST_INFO,TestId=$tracking_id"
+    TWIN_UPDATE_FAILURE_THRESHOLD="${TWIN_UPDATE_FAILURE_THRESHOLD:-00:01:00}"
```
philipktlin, line 1084: good refactor

builds/e2e/longhaul.yaml
and-rewsmith: I think that it is fine unless there exist amd platforms that we want a long timeout for (or vice versa).

and-rewsmith:
Successful test run:
https://msazure.visualstudio.com/One/_build/results?buildId=30079521&view=results

PR #2781, opened 2020-04-02
Longhaul / Stress: Prune networks on deployment to avoid module start failure

and-rewsmith:
There is a moby issue (31643) where containers references fail to get removed from a network, leading to new modules with the same name as the last deployment failing to start.

This fixes it by pruning the networks on every deployment.

### DIFF
--- a/scripts/linux/runE2ETest.sh
+++ b/scripts/linux/runE2ETest.sh
@@ -22,6 +22,7 @@ function clean_up() {
     if [ "$CLEAN_ALL" = '1' ]; then
         echo 'Prune docker system'
         docker system prune -af --volumes || true
+        docker network prune -f || true
     else
         echo 'Remove docker containers'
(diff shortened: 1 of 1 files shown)


scripts/linux/runE2ETest.sh
```diff
@@ -22,6 +22,7 @@ function clean_up() {
     if [ "$CLEAN_ALL" = '1' ]; then
         echo 'Prune docker system'
         docker system prune -af --volumes || true
+        docker network prune -f || true
```
philipktlin, line 25: you don't need to call "docker network prune" as "docker system prune" will do.  The reason why it doesn't work now coz we need to remove all the containers first before calling docker system prune.  I have a PR out to fix it.

and-rewsmith:
Closing as @philipktlin has a PR out to fix the same issue

PR #2785, opened 2020-04-02
[mqtt] Basic authorization for MQTT broker

dmolokanov:
Basic authorization for MQTT broker supports 4 actions:
1. **Connect** - Represents a client attempt to connect to MQTT broker. Policy engine will be requested to grant permissions for each client connected to the broker. The Connect request contains following information necessary for policy engine to identify only known clients: `client_id`, `auth_id`, `WILL message`. MQTT Broker disconnects a client which doesn't have permissions to connect to the broker.
 
2. **Publish** - Represents a client attempt to publish a message to a certain topic. Policy engine will be requested to grant permissions for each PUBLISH message sent. The Publish request contains following information necessary for policy engine to grant corresponding permissions: `client_id`, `auth_id`, `PUBLISH message`.  MQTT Broker disconnects a client which doesn't have permissions to publish messages to the broker.

3. **Subscribe** -  Represents a client attempt to subscribe to a certain topic to receive messages. Policy engine will be requested to grant permissions for each SUBSCRIBE message sent. The Subscribe request contains following information necessary for policy engine to grant corresponding permissions: `client_id`, `auth_id`, `topic filter and corresponding QoS level`. MQTT Broker determines which topics from SUBSCRIBE message a client has permissions to subscribe to and returns corresponding SUBACK message. SUBACK qos `FAILURE(0x08)` is returned in case if client don't have permissions to subscribe to a topic with provided topic filter.

4. **Receive** - Represents a client attempt to receive a message from the broker. Policy engine will be requested to grant permission for each client the broker is dispatching message to. The Receive request contains following information necessary for policy engine to grant corresponding permissions: `client_id`, `auth_id`, `PUBLISH message`.


mqtt/mqtt-broker/src/auth/authorization.rs
```diff
@@ -0,0 +1,217 @@
...
+impl<F> Authorizer for F
+where
+    F: Fn(Activity) -> Result<bool, Error> + Sync,
+{
+    async fn authorize(&self, activity: Activity) -> Result<bool, Error> {
+        self(activity)
+    }
+}
+
+/// Default implementation that always denies any operation a client intends to perform.
```
vadim-kovalyov, line 24: thanks for comments!

mqtt/mqtt-broker/src/auth/authorization.rs
```diff
@@ -0,0 +1,217 @@
...
+/// This implementation will be used if custom authorization mechanism was not provided.
+pub struct DefaultAuthorizer;
+
+#[async_trait]
+impl Authorizer for DefaultAuthorizer {
+    async fn authorize(&self, _: Activity) -> Result<bool, Error> {
+        Ok(false)
```
lfitchett, line 31: Should the default authorize allow everything rather than deny? Presumably the default authorizer will only be used for tests, never in production?

mqtt/mqtt-broker/src/auth/authorization.rs
```diff
@@ -0,0 +1,217 @@
...
+impl From<proto::Connect> for Connect {
+    fn from(connect: proto::Connect) -> Self {
+        Self {
+            will: connect.will.map(Into::into),
+        }
+    }
+}
+
+/// Represents a publication description without payload to be used for authorization.
+pub struct Publication {
```
lfitchett, line 107: Maybe name this AuthorizationPublication?

mqtt/mqtt-broker/src/auth/authorization.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub topic_name: String,
+    pub qos: proto::QoS,
+    pub retain: bool,
+}
+
+impl From<proto::Publication> for Publication {
```
lfitchett, line 113: You won't be able to use the From interface anymore, but it might be better to create this publication from a reference of a proto::publication to avoid having to clone the payload. 

(I know cloning bytes only increments the ref count, but its still extra overhead)

mqtt/mqtt-broker/src/auth/authorization.rs
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+}
+
+/// Represents a client attempt to publish a new message on a specified MQTT topic.
+pub struct Publish {
+    publication: Publication,
+}
+
+impl From<proto::Publish> for Publish {
```
lfitchett, line 128: Same comment as above

mqtt/mqtt-broker/src/auth/authorization.rs
dmolokanov: Default authorizer will deny every request unless configured other way. In production code when we configure broker we set another authorizer that allows everything for now. Once we have edgehub integration we will set a new edgehub-authorizer instead.

mqtt/mqtt-broker/src/auth/authorization.rs
dmolokanov: This struct is a local for the module. I don't think it's ever be used in the code outside of the module

mqtt/mqtt-broker/src/auth/authorization.rs
dmolokanov: Not sure if will give us too much benefit... Comparing to serializing data to send to edgehub-authorizer it may cost almost nothing.

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -456,9 +472,39 @@ where
...
+        if let Some(session) = self.sessions.get_mut(&client_id) {
+            let activity = Activity::new(session.auth_id()?.clone(), client_id.clone(), operation);
+            match self.authorizer.authorize(activity).await {
+                Ok(true) => {
+                    debug!("client {} successfully authorized", client_id);
+                    Ok(true)
+                }
+                Ok(false) => {
+                    warn!("client {} not allowed to publish", client_id);
+                    self.drop_connection(client_id).await?;
```
myagley, line 493: I would probably drop the connection in `process_publish` not `can_publish`. I would expect a function named `can_publish` to not have any side effects.

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -783,13 +825,102 @@ where
...
+    publication: &proto::Publication,
+) -> Result<(), Error> {
+    if can_publish_to(authorizer, session, publication).await? {
+        if let Some(event) = session.publish_to(&publication)? {
+            session.send(event).await?
+        }
+    } else {
+        info!(
+            "client {} not allowed to receive messages",
+            session.client_id()
```
myagley, line 900: We may want to move this to `debug`. This could generate a bunch of log lines.

mqtt/mqtt-broker/src/session.rs
```diff
@@ -96,6 +112,7 @@ impl ConnectedSession {
         self.state.publish_to(publication)
     }
 
+    #[allow(dead_code)]
     pub fn subscribe(
```
myagley, line 116: Did we forget to remove this? Seem like `subscribe` should be updated to have the `subscribe_to` logic and one of these functions should be removed?

mqtt/mqtt-broker/src/session.rs
```diff
@@ -565,48 +609,58 @@ pub enum Session {
...
+            Self::Offline(offline) => offline.client_id(),
+            Self::Disconnecting(disconnecting) => disconnecting.client_id(),
+        }
+    }
+
+    pub fn auth_id(&self) -> Result<&AuthId, Error> {
+        match self {
+            Self::Transient(connected) => Ok(connected.auth_id()),
+            Self::Persistent(connected) => Ok(connected.auth_id()),
+            Self::Offline(_offline) => Err(Error::from(ErrorKind::SessionOffline)),
```
myagley, line 653: :+1:

mqtt/mqtt-broker/src/session.rs
dmolokanov: Yes, should remove `subscribe`. Thanks!

PR #2788, opened 2020-04-02
Fix build failure due to SA1135

philipktlin:
SA1135 throws exception when using .net core 3.1 or using Visual studio 16.5.  Our build is failing now. 
 Therefore need to suppress SA1135 for short term solution.


and-rewsmith:
For some reason the build pipeline started building using dotnet 3.1. We need to figure out why this is happening, but should merge this in order to fix local builds.

PR #2790, opened 2020-04-02
Longhaul: Don't fail tests for status code 0

and-rewsmith:
Longhaul direct method tests will often fail with status code 0 returned by edgehub to the sender module.

I talked to Priyanka on the SDK team who informed that the feature enabling the fix for SDK issue [1026](https://github.com/Azure/azure-iot-sdk-csharp/issues/1026) is still on the backlog. Until this is complete, we should not fail the tests for status code 0.


and-rewsmith:
I need to investigate if we can fix this in edgehub. Will re-open if we cannot.

and-rewsmith:
Reopening because we decided not to spend resources fixing in edgehub until the SDK enables us to fix the root cause (unsubscribing from iothub subscriptions).

test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
```diff
@@ -49,5 +49,19 @@ public class CloudOperationReport
...
+        bool IsPassedHelper()
+        {
+            bool noUnexpectedStatus = this.StatusCodes.All(
+                x => x.StatusCode.StartsWith("200") ||
+                x.StatusCode.Equals("OK") ||
+                (x.StatusCode.Equals("0") && this.ModuleId.Contains("DirectMethod")));
+
+            // The SDK does not allow edgehub to de-register from iothub subscriptions, which results in DirectMethod clients sometimes receiving status code 0.
+            // We expect to get this status sometimes, but if we receive too many we should fail the tests.
+            bool statusCodeZeroBelowThreshold = this.StatusCodes.Where(s => s.StatusCode.Equals("0")).Count() < (this.StatusCodes.Count / 100);
```
and-rewsmith, line 62: Status code 0 only happens on edgehub restart.

I picked a 1% threshold for the proportion of status code 0 to other responses. The main factor that determines if this threshold can be inadvertently exceeded is the edgehub restart interval. Seeing as how the default is every 4 hours, we would need a restart interval of < 5 min to be problematic (guess).

philipktlin: This change is a workaround for that git hub issue, right?  If so, please add a link pointing to that github issue.

test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
```diff
@@ -49,5 +49,19 @@ public class CloudOperationReport
...
         }
+
+        bool IsPassedHelper()
+        {
+            bool noUnexpectedStatus = this.StatusCodes.All(
+                x => x.StatusCode.StartsWith("200") ||
+                x.StatusCode.Equals("OK") ||
+                (x.StatusCode.Equals("0") && this.ModuleId.Contains("DirectMethod")));
+
+            // The SDK does not allow edgehub to de-register from iothub subscriptions, which results in DirectMethod clients sometimes receiving status code 0.
```
dylanbronson, line 60: nit: add "TODO:"

test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
philipktlin: Why don't put the github issue link there, I think it is ok to put github issue link, but not internal link e.g. task or PBI.

and-rewsmith: I had this before, but force pushed to resolve the issue we had on Friday. Will re-add.

test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
```diff
@@ -49,5 +49,21 @@ public class CloudOperationReport
 
             public DateTime LastReceivedAt { get; set; }
         }
+
+        bool IsPassedHelper()
```
philipktlin, line 53: why name it like this?

and-rewsmith: The IsPassed property needs more complex logic than what we had previously ([Expression Body Definition](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-operator#expression-body-definition)) to determine whether it is passed. The purpose of this helper is to provide a place where we can provide a multi-line implementation.

Seeing as how these fields' purposes are the same and the only difference was needing a mechanism to provide multi-line implementation, this name seemed ok. Do you have another suggestion?

philipktlin: I agreed to use a method to wrap the logic.  but why just name this method "IsPassed"?  Do you worry about other dev will move it back to inline/expression body if property and method name are the same?

and-rewsmith: It can't have the same name as the property

philipktlin: Sorry. my bads.

PR #2791, opened 2020-04-02
[mqtt] Deliver peer certificate to authorizer

dmolokanov:
In order to retrieve a peer certificate for authentication purposes we need a newer version of tokio wrapper around native-tls which is not released yet.
* Switched to use `tokio-native-tls` instead outdated `tokio-tls`
* Add code to extract certificate and pass it to authentication logic


mqtt/mqtt-broker/src/transport.rs
```diff
@@ -239,7 +239,31 @@ impl StreamSelector {
     pub fn peer_addr(&self) -> std::io::Result<SocketAddr> {
         match self {
             StreamSelector::Tcp(stream) => stream.peer_addr(),
-            StreamSelector::Tls(stream) => stream.get_ref().peer_addr(),
+            StreamSelector::Tls(stream) => stream.get_ref().get_ref().get_ref().peer_addr(),
```
vadim-kovalyov, line 242: 0_o, wow

mqtt/mqtt-broker/src/connection.rs
```diff
@@ -73,10 +74,10 @@ pub async fn process<I>(
     mut broker_handle: BrokerHandle,
 ) -> Result<(), Error>
 where
-    I: AsyncRead + AsyncWrite + Unpin,
+    I: AsyncRead + AsyncWrite + GetPeerCertificate<Certificate = Certificate> + Unpin,
 {
     // TODO read certificate from stream when ready
```
vadim-kovalyov, line 79: todo is no longer needed?

mqtt/mqtt-broker/src/transport.rs
dmolokanov: yep, not very pleasant API

mqtt/mqtt-broker/src/transport.rs
```diff
@@ -239,7 +239,31 @@ impl StreamSelector {
...
+    }
+}
+
+pub trait GetPeerCertificate {
+    type Certificate;
+
+    fn peer_certificate(&self) -> Result<Option<Self::Certificate>, Error>;
+}
+
+impl GetPeerCertificate for StreamSelector {
```
vadim-kovalyov, line 253: is it possible to unit test?

dmolokanov: what exact aspect you want to check?

mqtt/mqtt-broker/src/connection.rs
lfitchett: Looks like todo is still here

PR #2800, opened 2020-04-05
Consolidate the version of nuget packages in the solution

philipktlin:
Consolidate multiple versions of nuget packages in the solution.
1. Microsoft.Extensions.Configuration.Binder 2.2.0 and 2.2.4 -> 2.2.4
2. Microsoft.Extensions.Configuration.EnvironmentVariables 2.2.0 and 2.2.4 -> 2.2.4
3. nunit 2.11.0 and 3.12.0 -> 3.12.0
4. Microsoft.Extensions.Configuration.Json 2.0.0 and 2.2.0 -> 2.2.0
5.RunProcessAsTask 1.2.3.and 1.2.4 -> 1.2.4
6. YamlDotNet 5.4.0 and 6.0.0 -> 6.0.0


damonbarry:
I think it would also be good to investigate a centralized system for managing package versions in our repo, like this [MSBuild project SDK](https://github.com/Microsoft/MSBuildSdks/tree/master/src/CentralPackageVersions). I tried it in a simple project and it seems to work.

Note I also found a similar [Nuget effort](https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions), but it hasn't gone past the spec stage.

philipktlin:
> NUnit3TestAdapter

This library is only used in Microsoft.Azure.Devices.Edge.Test project.  Don't need to consolidate.

PR #2802, opened 2020-04-06
[mqtt] Compliance Tests

lfitchett:
This runs compliance tests against the broker


builds/ci/mqtt.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - release/*
+jobs:
+  ################################################################################
+  - job: compliance_amd64
+    ################################################################################
+    displayName: Compliance amd64
+    # dependsOn: check_run_pipeline
+    # condition: eq(dependencies.check_run_pipeline.outputs['check_files.RUN_PIPELINE'], 'true')
+    pool:
+      vmImage: "ubuntu-16.04"
```
dmolokanov, line 15:
```suggestion
      vmImage: "ubuntu-latest"
```
?

builds/ci/mqtt.yaml
```diff
@@ -0,0 +1,25 @@
...
+  branches:
+    include:
+      - master
+      - release/*
+jobs:
+  ################################################################################
+  - job: compliance_amd64
+    ################################################################################
+    displayName: Compliance amd64
+    # dependsOn: check_run_pipeline
```
dmolokanov, line 12: I dont think we need commented lines

builds/ci/mqtt.yaml
```diff
@@ -0,0 +1,25 @@
+trigger: none
```
dmolokanov, line 1: Not sure that this pipeline should be called mqtt... Do you expect we add more in this pipeline vs create a new one?

lfitchett: Talked with Vadim and we were thinking of making this the mqtt e2e pipeline. So once we add some e2e tests, it will run with this

dmolokanov: Ok, lets leave it mqtt then

mqtt/build/linux/certgen.sh
```diff
@@ -0,0 +1,3 @@
+openssl genrsa 2048 > private.pem
```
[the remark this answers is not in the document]
lfitchett: but the tests also check 8883, which will always need a cert I think?

builds/ci/mqtt.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - release/*
+jobs:
+  ################################################################################
+  - job: compliance_amd64
+    ################################################################################
+    displayName: Compliance amd64
+    pool:
+      vmImage: "ubuntu-latest"
+    steps:
+      - script: echo "##vso[task.setvariable variable=RUST_BACKTRACE;]1"
```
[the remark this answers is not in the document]
lfitchett: I just used the same as denis' build

builds/ci/mqtt.yaml
```diff
@@ -0,0 +1,23 @@
+trigger: none
+pr:
```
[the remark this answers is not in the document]
lfitchett: removed

mqtt/build/linux/certgen.sh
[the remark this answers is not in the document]
vadim-kovalyov: oh, ok, didn't know that.

builds/ci/mqtt.yaml
```diff
@@ -0,0 +1,23 @@
...
+  ################################################################################
+  - job: compliance_amd64
+    ################################################################################
+    displayName: Compliance amd64
+    pool:
+      vmImage: "ubuntu-latest"
+    steps:
+      - script: echo "##vso[task.setvariable variable=RUST_BACKTRACE;]1"
+        displayName: Set env variables
+      - bash: mqtt/build/linux/install.sh
```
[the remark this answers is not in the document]
lfitchett: its only about 7 minutes, and I think its worth keeping these self contained and independent from the build pipeline.

builds/ci/mqtt.yaml
dmolokanov: Any issues with `ubuntu-latest`? I see different vmImage value for mqtt-perf and this pipelines

mqtt/build/linux/compliance.sh
```diff
@@ -0,0 +1,40 @@
...
+# Download tests and make certificate
+###############################################################################
+
+$DIR/certgen.sh
+git clone https://github.com/lfitchett/paho.mqtt.testing.git
+
+###############################################################################
+# Start broker and run tests
+###############################################################################
+set +e
```
dmolokanov, line 31: Why `-e` here https://github.com/Azure/iotedge/pull/2802/files#diff-0f2e075fd2852867d4ab80bf4a1b4ea6R7
but `+e` in this line?

lfitchett: for the above lines, I want an error to kill the script. Below, I manually grab the error code of the tests `result=$?` because I want to kill the broker even if the tests fail.

lfitchett:
> LGTM! although, @dmolokanov Denis Molokanov FTE had a question about using `ubuntu-latest` instead.

It breaks on latest. The checkin tests use the same version

builds/ci/mqtt.yaml
lfitchett: It breaks on latest. The checkin tests use the same version

PR #2804, opened 2020-04-06
Added metrics to Kubernetes mode to prevent Agent crash.

darobs:
Edge Agent was crashing in Kubernetes mode, autofac could not resolve "SystemResourcesMetrics".  This will prevent the crash.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/modules/KubernetesModule.cs
```diff
@@ -340,6 +343,10 @@ protected override void Load(ContainerBuilder builder)
                     })
                 .As<Task<IEnvironmentProvider>>()
                 .SingleInstance();
+
+            // SystemResourcesMetrics
+            builder.Register(c => new Edgelet.Docker.SystemResourcesMetrics(c.Resolve<IMetricsProvider>(), c.Resolve<IModuleManager>().GetSystemResourcesAsync, this.apiVersion, this.performanceMetricsUpdateFrequency))
```
lfitchett, line 348: It might be better to make a ISystemResourcesMetrics interface and a NullSystemResourcesMetrics, since SystemResourcesMetrics makes http calls to docker. If the enable metrics flag is set to true it will try to start this in docker, and I'm not sure what would happen. 

Alternatively, you could replace `c.Resolve<IModuleManager>().GetSystemResourcesAsync` with `() => Task.FromResult(new Edgelet.Models.SystemResources(0, 0, 0, 0, 0, new Edgelet.Models.Disk[0], string.Empty))`. This will prevent the call to edgelet

darobs: Made an ISystemResourcesMetrics and an empty implementation for k8s.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/NullSystemResourcesMetrics.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker
+{
+    using System;
+    using Microsoft.Extensions.Logging;
+
+    public class NullSystemResourcesMetrics : ISystemResourcesMetrics
+    {
+        public NullSystemResourcesMetrics()
```
lfitchett, line 10: nit. no need to define a default constructor

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/Program.cs
```diff
@@ -140,6 +140,7 @@ public static async Task<int> MainAsync(IConfiguration configuration)
                 experimentalFeatures = ExperimentalFeatures.Create(configuration.GetSection("experimentalFeatures"), logger);
                 Option<ulong> storageTotalMaxWalSize = GetConfigIfExists<ulong>(Constants.StorageMaxTotalWalSize, configuration, logger);
                 Option<StorageLogLevel> storageLogLevel = GetConfigIfExists<StorageLogLevel>(Constants.StorageLogLevel, configuration, logger);
+                TimeSpan performanceMetricsUpdateFrequency = configuration.GetTimeSpan("PerformanceMetricsUpdateFrequency", TimeSpan.FromMinutes(5));
```
lfitchett, line 143: nit. does this still need to be moved?

darobs: Nope, missed that.

PR #2805, opened 2020-04-06
DirectMethodSender: Don't report internal server error for transient exceptions

and-rewsmith:
Every time the DirectMethodReceiver module's upstream connection is down (i.e. startup / edgehub restart), then DirectMethodSender will receive a DeviceNotFound exception when sending a direct method. 

Currently, we treat this as a internal server error when reporting results to the TestAnalyzer / TRC. Instead, we should return 404 and implement logic in the TestAnalyzer / TRC to treat this as a transient error.


test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -262,7 +262,7 @@ async Task<DirectMethodReportGeneratorMetadata> ProcessMismatchFailureCase()
             }
             else if (NetworkControllerStatus.Enabled.Equals(networkControllerStatus))
             {
-                if (HttpStatusCode.InternalServerError.Equals(statusCode))
+                if (HttpStatusCode.NotFound.Equals(statusCode))
```
philipktlin, line 265: The original intent is any server error will execute this if block.  But now it is changed to run only if 404 not found (DeviceNotFoundException).  This may not be correct, please confirm with @dylanbronson.

test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
```diff
@@ -33,7 +33,9 @@ public AggregateCloudOperationReport(string moduleId, IDictionary<string, Tuple<
 
         public IDictionary<string, string> TestInfo { get; }
 
-        public bool IsPassed => this.StatusCodes.All(x => x.StatusCode.StartsWith("200") || x.StatusCode.Equals("OK"));
+        public bool IsPassed => this.StatusCodes.All(x => x.StatusCode.StartsWith("200")
+                                                        || x.StatusCode.Equals("OK")
+                                                        || x.StatusCode.Equals("NotFound"));
```
philipktlin, line 38: It looks weird to me, why "NotFound" still make the report passed?

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
and-rewsmith: I changed the InternalServerError status to NotFound. The only risk I see is if the status code returned by the request (i.e. no exception happens) is NotFound. If we want to eliminate this chance, we will have to return a custom status code. What do you think?

test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
and-rewsmith: I agree that this looks weird, but this is a byproduct of how we defined the status code on DeviceNotFoundException. We could change it to a custom status code?

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
dylanbronson: I think this is mostly right.

When DMSender fails because the network is off, it fails with DeviceNotFoundException. This PR is now catching that specific exception and turning into a 404. Previously we just treat all exceptions as 500. I've never seen an exception while offline that is not DeviceNotFoundException in the DMSender logs.

If some other exception occurs while we are supposed to be offline, we probably want to know about it, so this report should fail.

dylanbronson, line 265: We probably want to fail the report, though, and not fail report generation, so we should change line 282 in this file to networkOffFailure++ if it is a 500

and-rewsmith: The risk is that if there is a product bug causing the dm response to actually return 404, we won't be able to differentiate between these. In this case we would want a test failure, but the test would succeed.

and-rewsmith: Oh, great call

and-rewsmith: @philipktlin I discussed with Dylan and we think it might be best to choose a custom status code. Do you agree? If so, we cannot use the HttpStatusCode type because it doesn't support custom codes. I don't know if this is desirable.

philipktlin: 1. First this change added to catch DeviceNotFoundException explicitly and return status code 404, and continue to return status code 500 for any other exceptions.
2. With Dylan's statement, currently we only see DeviceNotFoundException.  Therefore it is safe to change this line 265 to equals to 404 instead of 500.  We just need to have a wrapper function "IsExpectedError" to make the code more readable.
3. Http status code 404 is a standard to tell that resource not found.  If DM invoke call returns 404 instead of throwing DeviceNotFoundException in the future, it is still ok.  If DM invoke call returns 404 but device exists, then this is SDK bug.  Therefore I think it is safe to use status code 404.

test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
philipktlin: Forget about current implementation.  For connectivity, I expect it should report what the result is to TRC.  For LH, if there is any exception during DM invoke call, we should eat up the exception, log it and retry (without increasing sequence number [optional if this doesn't impact to Connectivity test]).  If that's the case, we don't need to make change here.

philipktlin: To implement this, you can keep the change in DMSenderBase.  You may need to create a new subclass from DMSenderBase for LH DM Cloud Sender.

and-rewsmith: I implemented it like this in order to keep the test-specific (longhaul vs connectivity) in the module that defines the test type (Analyzer vs TRC). Although I forgot that the TestAnalyzer expects the module to send test results rather than response codes. I think you are right and will make the change.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
and-rewsmith: For 1 and 2 I agree, but disagree for 3. Edgehub selects the status code on behalf of the module. Therefore I could make a breaking edgehub change to return 404 for all direct method responses and this would still pass the tests.

I think selecting a custom status code would make this risk negligible.

philipktlin: Quoted "Edgehub selects the status code on behalf of the module".  What do you mean?

If you are going to make "a breaking edgehub change to return 404 for all direct method responses", you are creating a bug in EdgeHub.  The test should not pass for calls when online network condition.  Do I understand correctly?

and-rewsmith: I agree that the tests should not pass, but if we eat up all 404 responses then the tests would pass.

and-rewsmith: I discussed with philip and the longhaul behavior ate exceptions before DirectMethodSender was refactored. At the very least we should fail the tests if a successful DM was not returned in some defined tolerance period (separate PR).

test/modules/DirectMethodSender/Program.cs
```diff
@@ -133,5 +137,10 @@ public static TestResultBase ConstructTestResult(DirectMethodResultType directMe
                     throw new NotImplementedException("Reporting Endpoint has an unknown type");
             }
         }
+
+        static bool ShouldReportResults(DirectMethodResultType resultType, HttpStatusCode statusCode)
```
philipktlin, line 141: I like the method name

test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
```diff
@@ -33,7 +33,8 @@ public AggregateCloudOperationReport(string moduleId, IDictionary<string, Tuple<
 
         public IDictionary<string, string> TestInfo { get; }
 
-        public bool IsPassed => this.StatusCodes.All(x => x.StatusCode.StartsWith("200") || x.StatusCode.Equals("OK"));
+        public bool IsPassed => this.StatusCodes.All(x => x.StatusCode.StartsWith("200")
```
and-rewsmith, line 36: I should undo this line space

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -262,11 +262,11 @@ async Task<DirectMethodReportGeneratorMetadata> ProcessMismatchFailureCase()
...
                 {
                     networkOffSuccess++;
                 }
-                else if (HttpStatusCode.OK.Equals(dmSenderTestResult.HttpStatusCode))
+                else if (HttpStatusCode.OK.Equals(statusCode))
```
philipktlin, line 269: good catch.

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
```diff
@@ -277,6 +277,10 @@ async Task<DirectMethodReportGeneratorMetadata> ProcessMismatchFailureCase()
                         networkOffFailure++;
                     }
                 }
+                else if (HttpStatusCode.InternalServerError.Equals(statusCode))
```
philipktlin, line 280: @dylanbronson, is this the right bucket?

test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
philipktlin: nit:revert this change

and-rewsmith: already did

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportDataWithSenderAndReceiverSource.cs
```diff
@@ -90,7 +90,7 @@ class DirectMethodReportDataWithSenderAndReceiverSource
                     // NetworkOnFailure test
                     Enumerable.Range(1, 7).Select(v => (ulong)v),
                     new[] { 1UL, 2UL, 3UL, 5UL, 6UL, 7UL },
-                    new List<HttpStatusCode> { HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.InternalServerError, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK },
+                    new List<HttpStatusCode> { HttpStatusCode.InternalServerError, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.NotFound, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK },
```
and-rewsmith, line 93: Network on can fail for two cases now so I am testing both

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportDataWithSenderAndReceiverSource.cs
```diff
@@ -101,25 +101,25 @@ class DirectMethodReportDataWithSenderAndReceiverSource
...
-                    10, 6, 0, 0, 0, 1, 0, 0, 0, false
+                    10, 5, 0, 0, 0, 2, 0, 0, 0, false
                 },
                 new object[]
                 {
                     // NetworkOffFailure test
                     Enumerable.Range(1, 7).Select(v => (ulong)v),
                     new[] { 1UL, 2UL, 3UL, 4UL, 5UL, 6UL, 7UL },
-                    new List<HttpStatusCode> { HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK },
+                    new List<HttpStatusCode> { HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.InternalServerError, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK },
```
and-rewsmith, line 111: Network off can fail for two cases now so I am testing both

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportDataWithSenderSourceOnly.cs
```diff
@@ -95,29 +95,29 @@ public class DirectMethodReportDataWithSenderSourceOnly
...
                     },
-                    10, 6, 0, 0, 0, 1, 0, 0, 0, false
+                    10, 5, 0, 0, 0, 2, 0, 0, 0, false
                 },
                 new object[]
                 {
                     // NetworkOffFailure test
                     Enumerable.Range(1, 7).Select(v => (ulong)v),
-                    new List<HttpStatusCode> { HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK },
+                    new List<HttpStatusCode> { HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.InternalServerError, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK },
```
and-rewsmith, line 104: See other comments for context on these changes in this file

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodReportGenerator.cs
dylanbronson: I think so.

With this change, if we get an InternalServerError while offline, that is unexpected and the report should fail.
The only Errors we should get while offline are NotFound Errors.

It changes the definition of NetworkOffFailure from "Call succeeded while offline" to "Something unexpected happened while offline - either success or some weird InternalServerError"

It could be argued that we should have another, more specific bucket, but both cases will need to be manually looked at anyway, so it's safe to put it in this bucket.

and-rewsmith: "It changes the definition of NetworkOffFailure from "Call succeeded while offline" to "Something unexpected happened while offline"

Call succeeded while offline will still fail. It changed the definition of NetworkOffFailure to also care about internal server error, while passing 404.

PR #2806, opened 2020-04-06
Fix build about SA1135

philipktlin:
Fix build due to SA1135 when build with .Net Core 3.1. Will re-enable it once https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3136 is resolved.


and-rewsmith:
Anyone know why some builds were passing while some were consistently failing?

philipktlin:
> Anyone know why some builds were passing while some were consistently failing?

Only Mqtt broker build will pass as it doesn't build iotedge solution.

PR #2815, opened 2020-04-08
[mqtt] Append a simple version to mqtt persistance

lfitchett:
~~This simply reserves 100 bytes for version information~~

This now has a string for version. It appends a header to the file that is 4 bytes representing the length of the version string, then the version string.

Later more sophisticated processing can happen after the string is read to enable legacy formats

### DIFF
--- a/mqtt/mqtt-broker/src/persist.rs
+++ b/mqtt/mqtt-broker/src/persist.rs
@@ -313,7 +313,7 @@ where
                         path.clone(),
                     ))))
                 });
-                let file = OpenOptions::new()
+                let mut file = OpenOptions::new()
                     .read(true)
                     .open(&path)
                     .context(ErrorKind::Persist(ErrorReason::FileOpen(path)))?;
@@ -321,6 +321,12 @@ where
                 fail_point!("filepersistor.load.format", |_| {
                     Err(Error::from(ErrorKind::Persist(ErrorReason::Serialize)))
                 });
+
+                let mut version = [0; 100];
+                debug!("reading version");
+                file.read_exact(&mut version)
+                    .map_err(|_| Error::from(ErrorKind::Persist(ErrorReason::ReadVersion)))?;
+
                 let state = format.load(file).map_err(Into::into)?;
                 Ok(Some(state))
             } else {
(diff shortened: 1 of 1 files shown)


mqtt/mqtt-broker/src/persist.rs
```diff
@@ -313,14 +313,20 @@ where
...
+                let mut file = OpenOptions::new()
                     .read(true)
                     .open(&path)
                     .context(ErrorKind::Persist(ErrorReason::FileOpen(path)))?;
 
                 fail_point!("filepersistor.load.format", |_| {
                     Err(Error::from(ErrorKind::Persist(ErrorReason::Serialize)))
                 });
+
+                let mut version = [0; 100];
```
vadim-kovalyov, line 325: `0u8`? and just curious why 100 byte? :)

lfitchett: the read trait requires a u8, so the type will be inferred. 100 bytes was completely arbitrary, it seemed like a good number 🤷‍♂

lfitchett:
How do we feel about having a size byte first, that represents the length of the header?

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -378,13 +397,22 @@ where
...
             });
-            let file = OpenOptions::new()
+            let mut file = OpenOptions::new()
                 .create(true)
                 .write(true)
                 .open(&path)
                 .context(ErrorKind::Persist(ErrorReason::FileOpen(path.clone())))?;
             debug!("{} opened.", path.display());
 
+            let version = bincode::serialize("Version 0.0.1")
```
dmolokanov, line 407: Seems a bit overkill I thought we can wrap `ConsolidatedState` into some struct/enum and just ser/de it:
```rust 
#[derive(Serialize, Deserialize)]
#[serde(tag = "version")]
enum FileFormat {
  #[serde(rename = 1)]
  V1(ConsolidatedState)
}
```
So no manual work involved, everything is done by serde

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -313,14 +314,32 @@ where
...
                 });
+
+                debug!("reading version");
+                let mut size = [0; std::mem::size_of::<u32>()];
+                file.read_exact(&mut size)
+                    .map_err(|_| Error::from(ErrorKind::Persist(ErrorReason::ReadVersion)))?;
+                let size: usize = u32::from_be_bytes(size).try_into().unwrap();
+
+                let mut version = Vec::with_capacity(size);
+                unsafe {
```
dmolokanov, line 333: I think we should avoid unsafe code as much as we can. Especially in such simple case

mqtt/mqtt-broker/src/persist.rs
varunpuranik: +1 on this

lfitchett: the problem is that the version has to be read before the processing begins on the state. Otherwise the persistor won't know which read function to call. 

For example, say we have 2 formats: simple binary and gzipped binary. The version information has to be separate from the state otherwise we won't know if it should be decompressed or not.

mqtt/mqtt-broker/src/persist.rs
lfitchett: The only safe way to do this I could find fills every value with 0, which seems inefficient. Is there a better way to get a mutable slice of a specified size?

lfitchett: I'll just use the safe func that fills it with 0. It's slightly inefficient, but its very minor

PR #2816, opened 2020-04-08
[mqtt] Remove failure crate dependency

dmolokanov:
The project uses `failure` crate to preserve the context of the error. While it works fine it seems like maintainer is planning to discontinue it soon. Another issue is that the crate is based on custom `Fail` trait which is not compatible with `std::error::Error` trait.
The idea is to get rid of `failure` crate in favor of `thiserror` which seems to provide almost everything we need and based on `std::error::Error`.


mqtt/mqtt-broker/src/auth/authentication.rs
```diff
@@ -13,38 +12,56 @@ pub enum Credentials {
...
+    fn from(certificate: Vec<u8>) -> Self {
+        Self(certificate)
+    }
+}
+
 /// A trait to authenticate a MQTT client with given credentials.
 #[async_trait]
 pub trait Authenticator {
+    /// Authentication error.
+    type Error;
```
myagley, line 28: should we force this to be `std::error::Error`? The broker below has a type bound on `Display` which is so close to being `std::error::Error` (just need to add `source`)

myagley:
```rust
type Error: std::error::Error;
```

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -32,7 +30,9 @@ macro_rules! try_send {
 pub struct Broker<N, Z>
 where
     N: Authenticator,
+    N::Error: Display + Into<Error>,
     Z: Authorizer,
+    Z::Error: Display + Into<Error>,
```
myagley, line 35: We only use this error for logging? Do we ever bubble this error out or take action on it?

mqtt/mqtt-broker/src/error.rs
```diff
@@ -1,135 +1,88 @@
...
+    SendConnectionMessage(#[source] tokio::sync::mpsc::error::SendError<Message>),
 
-    #[fail(display = "An error occurred sending a message to a snapshotter.")]
+    #[error("An error occurred sending a message to a snapshotter.")]
     SendSnapshotMessage,
 
-    #[fail(display = "An error occurred decoding a packet.")]
-    DecodePacket,
+    #[error("An error occurred decoding a packet.")]
+    DecodePacket(#[from] mqtt3::proto::DecodeError),
```
myagley, line 20: Does `#[from]` also set the `#[source]`?

mqtt/mqttd/src/main.rs
```diff
@@ -1,7 +1,6 @@
 use std::{convert::TryInto, env, io};
```
myagley, line 1: Just a general note, we should probably keep an eye on the length of `main`. I don't want what we have in `iotedged` to happen again where the `main` method is huge and not tested. Nothing required to change in this PR. Edits to `main` reminded me.

mqtt/mqtt-broker/src/error.rs
dmolokanov: yes, but not vice versa. `#[source]` dont generate `impl From`

dmolokanov: > The #[from] attribute always implies that the same field is #[source], so you don't ever need to specify both attributes.

PR #2817, opened 2020-04-08
TimeToLive E2E test

dylanbronson:
Adding first E2E test for TTL on Priority Routes.

It is a copy of the test for Priority Queues with a few changes.

We randomly generate TTL's that are above and below a threshold from a set of hard coded TTL's and assign them to the routes. These TTL's can also be -1, which represents the user not entering TTL and 0, which signifies the user putting in 0 as the TTL (which ends up being the default as well).

PriorityMessageSender will send messages choosing random priorities and selecting from the TTLs. It will then only report to the TRC the expected messages whose TTL's are above the TTLThreshold. 

In the test, we assume Default TTL to be above the threshold. Default TTL is higher than the longest possible time of the test. You could set the global TTL to be lower, but that is not the case in this test. Maybe another later test should have that case.


test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -124,32 +128,159 @@ public async Task PriorityQueueModuleToModuleMessages()
...
             }
 
             return routes;
         }
 
+        private string BuildTTLString(int numberOfTTLs)
+        {
+            Random rng = new Random();
+            // Choose from a set of 2 TTLs below the threshold, 2 above
+            var ttlSet = new string[] { "5", "10", "400", "1600" };
```
damonbarry, line 272: How long will this test take to complete (on average)?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -124,32 +128,159 @@ public async Task PriorityQueueModuleToModuleMessages()
             Assert.IsTrue(isPassed);
         }
 
-        private Dictionary<string, object> BuildRoutes(string[] priorities, string sendModule, string receiveModule)
+        [Test]
+
```
vadim-kovalyov, line 132: nit: extra line.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -124,32 +128,159 @@ public async Task PriorityQueueModuleToModuleMessages()
...
+                routes.Add($"LoadGenToRelayer{priority}", routeInfo);
             }
 
             return routes;
         }
 
+        private string BuildTTLString(int numberOfTTLs)
+        {
+            Random rng = new Random();
```
richma-ms, line 270: >Random rng = new Random(); [](start = 12, length = 26)

hmm the way this is setup can be pretty unintuitive, as TTLs are hardcoded and the threshold is implied.  Since you're trying to inject some amount of randomness into this anyways, maybe it'd be better to maintain a list of offsets instead, say from -15 to +50 or something.  Then randomly pick an offset and add it to the threshold to calculate the TTL.  That way the threshold will be tightly coupled, preventing people from accidentally breaking the test by changing the threshold without noticing how it's used.

dylanbronson: I like that idea

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: ~2-2.5 minutes on amd64

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -88,10 +90,12 @@ public async Task PriorityQueueModuleToModuleMessages()
...
                             ("testDuration", loadGenTestDuration),
                             ("messageFrequency", "00:00:00.5"),
-                            ("priorities", priorityString)
+                            ("priorities", priorityString),
+                            ("ttls", ttlString),
+                            ("ttlThresholdSecs", "100")
                         });
 
-                    Dictionary<string, object> routes = this.BuildRoutes(priorityString.Split(';'), loadGenModuleName, relayerModuleName);
+                    Dictionary<string, object> routes = this.BuildRoutes(priorityString.Split(';'), ttlString.Split(";"), loadGenModuleName, relayerModuleName);
```
philipktlin, line 98: should it be single quote?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -124,32 +128,159 @@ public async Task PriorityQueueModuleToModuleMessages()
...
+            const string trcModuleName = "testResultCoordinator";
+            const string loadGenModuleName = "loadGenModule";
+            const string relayerModuleName = "relayerModule";
+            const string trcUrl = "http://" + trcModuleName + ":5001";
+            const string loadGenTestDuration = "00:00:20";
+
+            string routeTemplate = $"FROM /messages/modules/{loadGenModuleName}/outputs/pri{0} INTO BrokeredEndpoint('/modules/{relayerModuleName}/inputs/input1')";
+
+            string trackingId = Guid.NewGuid().ToString();
+            string priorityString = this.BuildPriorityString(5);
```
philipktlin, line 149: I think you need to refactor to better manage priorityString, ttlString and routes.  It is loosely-coupled and easy to break.

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -28,16 +28,25 @@ class PriorityMessageSender : LoadGenSenderBase
...
 
         public string PriorityString { get; }
 
+        public string TtlString { get; }
+
+        public int TtlThresholdSecs { get; }
+
         public async override Task RunAsync(CancellationTokenSource cts, DateTime testStartAt)
         {
             string[] outputs = this.PriorityString.Split(';');
```
philipktlin, line 48: rename outputs to priorities?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -124,32 +128,159 @@ public async Task PriorityQueueModuleToModuleMessages()
...
-            foreach (string priority in priorities)
+            for (int i = 0; i < priorities.Length; i++)
             {
+                string ttl = ttls[i];
+                string priority = priorities[i];
                 // If we encounter "Default" in the priority list, don't add a priority - the default priority will automatically get picked up
-                if (priority.Contains(TestConstants.PriorityQueues.Default))
+                var routeInfo = new Dictionary<string, object>();
+                routeInfo["route"] = $"FROM /messages/modules/{sendModule}/outputs/pri{priority} INTO BrokeredEndpoint('/modules/{receiveModule}/inputs/input1')";
+                if (!priority.Contains(TestConstants.PriorityQueues.Default))
```
philipktlin, line 252: Use equals with string comparison option instead of Contains

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -28,16 +28,25 @@ class PriorityMessageSender : LoadGenSenderBase
...
+                new ArgumentException("PriorityMessageSender must have 'ttls' environment variable set to a valid list of strings delimited by ';'"));
+            this.TtlThresholdSecs = Settings.Current.TtlThresholdSecs.Expect(() =>
+                new ArgumentException("PriorityMessageSender must have 'ttlThresholdSecs' environment variable set to a valid int"));
             this.isFinished = false;
             this.resultsSent = 0;
         }
 
         public string PriorityString { get; }
 
+        public string TtlString { get; }
```
philipktlin, line 42: can it be private? same for PriorityString.

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -52,8 +61,10 @@ public async override Task RunAsync(CancellationTokenSource cts, DateTime testSt
                 {
                     int choosePri = this.rng.Next(outputs.Length);
```
philipktlin, line 62: Either you check in the constructor to ensure that both priorities and ttls has same number of items; or use minimum size of priorities and ttls, e.g. Math.Min(priorities.Length, ttls.Length).  If you choose to ensure both have same number of arguments and with previous comment suggest to convert these to collection type in settings class; you should do this validation in Settings constructor.

philipktlin: or you can think of another structure, for example 1 environment variable "MessagePriorityAndTTL" has value of <priority1>,<ttl1>;<priority2>,<ttl2>...

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -52,8 +61,10 @@ public async override Task RunAsync(CancellationTokenSource cts, DateTime testSt
...
                     string output = outputs[choosePri];
+                    int ttlForMessage = int.Parse(ttls[choosePri]);
```
philipktlin, line 64: rename choosePri

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -124,32 +128,159 @@ public async Task PriorityQueueModuleToModuleMessages()
...
                 {
-                    routes.Add($"LoadGenToRelayer{priority}", new Dictionary<string, object>
-                    {
-                        ["route"] = $"FROM /messages/modules/{sendModule}/outputs/pri{priority} INTO BrokeredEndpoint('/modules/{receiveModule}/inputs/input1')",
-                    });
+                    routeInfo["priority"] = int.Parse(priority);
                 }
-                else
+
+                if (int.Parse(ttl) != -1)
```
philipktlin, line 257: nit: int.Parse(ttl) >= 0

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -124,32 +128,159 @@ public async Task PriorityQueueModuleToModuleMessages()
...
+            const string loadGenModuleName = "loadGenModule";
+            const string relayerModuleName = "relayerModule";
+            const string trcUrl = "http://" + trcModuleName + ":5001";
+            const string loadGenTestDuration = "00:00:20";
+
+            string routeTemplate = $"FROM /messages/modules/{loadGenModuleName}/outputs/pri{0} INTO BrokeredEndpoint('/modules/{relayerModuleName}/inputs/input1')";
+
+            string trackingId = Guid.NewGuid().ToString();
+            string priorityString = this.BuildPriorityString(5);
+            string ttlString = this.BuildTTLString(5);
```
philipktlin, line 150: should BuildTTLString consider ttlThresholdSecs to random generate ttls?

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -28,16 +28,25 @@ class PriorityMessageSender : LoadGenSenderBase
             : base(logger, moduleClient, batchId, trackingId)
         {
             this.PriorityString = Settings.Current.Priorities.Expect(() =>
-                new ArgumentException("PriorityMessageSender must have 'priorities' environment variable set to a valid list of string delimited by ';'"));
+                new ArgumentException("PriorityMessageSender must have 'priorities' environment variable set to a valid list of strings delimited by ';'"));
+            this.TtlString = Settings.Current.Ttls.Expect(() =>
```
philipktlin, line 32: Both TtlString and PriorityString should be converted to collection type in Settings class.  Do we have use case that ttl string is better than collection type, e.g. array or list?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -124,32 +128,159 @@ public async Task PriorityQueueModuleToModuleMessages()
...
+                    Dictionary<string, object> routes = this.BuildRoutes(priorityString.Split(';'), ttlString.Split(";"), loadGenModuleName, relayerModuleName);
+                    builder.GetModule(ModuleName.EdgeHub).WithDesiredProperties(new Dictionary<string, object> { ["routes"] = routes });
+                });
+
+            EdgeDeployment deployment = await this.runtime.DeployConfigurationAsync(addInitialConfig, token);
+            PriorityQueueTestStatus loadGenTestStatus = await this.PollUntilFinishedAsync(loadGenModuleName, token);
+
+            // Wait long enough for TTL to expire for some of the messages
+            Log.Information("Waiting for 20 seconds for TTL's to expire");
+            await Task.Delay(20000);
```
philipktlin, line 214: should it be better to use ttlThresholdSecs?

philipktlin: I am worry that if relayer module takes long time to download and start up; messages may be dropped more than expected.  You may need to re-think about how to pick TTLs.

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -63,13 +74,9 @@ public async override Task RunAsync(CancellationTokenSource cts, DateTime testSt
...
                     }
-                    else
+                    else if (ttlForMessage <= 0 || ttlForMessage > this.TtlThresholdSecs)
                     {
-                        int priority = 2000000000; // Default priority
-                        if (!output.Contains(TestConstants.PriorityQueues.Default))
-                        {
-                            priority = int.Parse(output);
-                        }
+                        int priority = output.Contains(TestConstants.PriorityQueues.Default) ? 2000000000 : int.Parse(output);
```
philipktlin, line 79: Use equals instead of Contains

test/modules/load-gen/PriorityMessageSender.cs
```diff
@@ -97,17 +104,12 @@ public async override Task RunAsync(CancellationTokenSource cts, DateTime testSt
...
-                throw new InvalidOperationException($"Must send more than 1 priority for valid test results. Priorities sent: {priorityAndSequence.Keys.Count}");
-            }
-
             // Sort priority by sequence number
             List<long> expectedSequenceNumberList = priorityAndSequence
                 .SelectMany(t => t.Value)
                 .ToList();
 
-            this.resultsSent = expectedSequenceNumberList.Count;
+            this.resultsSent = expectedSequenceNumberList.Count + 1;
```
philipktlin, line 112: why add 1 here?

test/modules/load-gen/Settings.cs
```diff
@@ -71,7 +82,9 @@ static Settings Create()
                 Option.Maybe(testResultCoordinatorUrl),
                 configuration.GetValue<string>("IOTEDGE_MODULEID"),
                 configuration.GetValue("senderType", LoadGenSenderType.DefaultSender),
-                Option.Maybe(priorities));
+                Option.Maybe(priorities),
+                Option.Maybe(ttls),
+                ttlThresholdSecs);
```
philipktlin, line 87: ttlThresholdSecs should be Option<int> as it is applicable only for priority message test.

test/modules/load-gen/Settings.cs
```diff
@@ -96,6 +109,10 @@ static Settings Create()
 
         public Option<string> Priorities { get; }
 
+        public Option<string> Ttls { get; }
+
+        public Option<int> TtlThresholdSecs { get; }
```
philipktlin, line 114: Add new fields to ToString method including Priorites field.

test/modules/load-gen/PriorityMessageSender.cs
dylanbronson: To account for the first message, which must be sent and expected first regardless of its priority and TTL

test/modules/load-gen/Settings.cs
dylanbronson: It is already, it just doesn't need the extra Option.Maybe step

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: Used Richard's suggestion above. This could still be an issue, but much lesser chance now. And can always bump lowest TTLOffset up even higher

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: Used Richard's suggestion above

test/modules/load-gen/PriorityMessageSender.cs
dylanbronson: Refactored using TestInfo struct

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: Refactored using TestInfo struct

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -187,5 +266,19 @@ private async Task<PriorityQueueTestStatus> PollUntilFinishedAsync(string module
             while (!testStatus.IsFinished);
             return testStatus;
         }
+
+        private TestInfo InitTestInfo(int length, int ttlThreshold, bool defaultTtls = false)
```
philipktlin, line 270:
```suggestion
        private TestInfo InitTestInfo(int ttlSampleLength, int ttlThreshold, bool defaultTtls = false)
```

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -73,105 +148,109 @@ public async Task PriorityQueueModuleToModuleMessages()
...
-            if (!isPassed)
+                routes.Add($"LoadGenToRelayer{priority}", routeInfo);
+            }
+
+            return routes;
+        }
+
+        private List<int> BuildDefaultTtls(int numberOfTTLs)
+        {
+            var ttls = new List<int>();
```
philipktlin, line 205: return new List<int>(Enumerable.Repeat(0, numberOfTTLs));

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -187,5 +266,19 @@ private async Task<PriorityQueueTestStatus> PollUntilFinishedAsync(string module
             while (!testStatus.IsFinished);
             return testStatus;
         }
+
+        private TestInfo InitTestInfo(int length, int ttlThreshold, bool defaultTtls = false)
+        {
+            List<int> ttls = defaultTtls ? this.BuildDefaultTtls(length) : this.BuildTtls(length, ttlThreshold);
+            List<int> priorities = this.BuildPriorities(length);
+            return new TestInfo() { Ttls = ttls, Priorities = priorities, TtlThreshold = ttlThreshold };
```
philipktlin, line 274: if defaultTtls is true and ttlThreshold is not used in line 272, why return ttlThreshold here?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
philipktlin: should combine ttlThreshold and defaultTtls into Option<int> ttlThreshold?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
philipktlin: I think you can combine BuildDefaultTtls and BuildTtls method using Option<int> ttlThreshold.

test/modules/load-gen/Settings.cs
```diff
@@ -42,6 +44,8 @@ class Settings
             this.ModuleId = Preconditions.CheckNonWhiteSpace(moduleId, nameof(moduleId));
             this.SenderType = senderType;
             this.Priorities = priorities;
+            this.Ttls = ttls;
```
philipktlin, line 47: do we need to validate that ttls and priorities and ttlThresholdSecs should provide all or nothing? and if provided, size of ttls and priorites are the same (not allow empty list).

test/modules/load-gen/Settings.cs
```diff
@@ -56,9 +60,16 @@ static Settings Create()
...
+            List<int> priorities = string.IsNullOrWhiteSpace(configuration.GetValue<string>("priorities"))
                 ? null
-                : configuration.GetValue<string>("priorities");
+                : configuration.GetValue<string>("priorities").Split(';').Select(int.Parse).ToList();
+
+            List<int> ttls = string.IsNullOrWhiteSpace(configuration.GetValue<string>("ttls"))
+                ? null
+                : configuration.GetValue<string>("ttls").Split(';').Select(int.Parse).ToList();
+
+            int ttlThresholdValue = configuration.GetValue<int>("ttlThresholdSecs");
```
philipktlin, line 71: will this return default(int) or throw exception if this key is not defined?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -73,105 +148,109 @@ public async Task PriorityQueueModuleToModuleMessages()
...
 
-                    Dictionary<string, object> routes = this.BuildRoutes(priorityString.Split(';'), loadGenModuleName, relayerModuleName);
+                    Dictionary<string, object> routes = this.BuildRoutes(testInfo.Priorities, testInfo.Ttls, LoadGenModuleName, RelayerModuleName);
                     builder.GetModule(ModuleName.EdgeHub).WithDesiredProperties(new Dictionary<string, object> { ["routes"] = routes });
                 });
+        }
 
-            EdgeDeployment deployment = await this.runtime.DeployConfigurationAsync(addInitialConfig, token);
-            PriorityQueueTestStatus loadGenTestStatus = await this.PollUntilFinishedAsync(loadGenModuleName, token);
+        private Dictionary<string, object> BuildRoutes(List<int> priorities, List<int> ttls, string sendModule, string receiveModule)
```
philipktlin, line 176: Uses TestInfo instead
```suggestion
        private Dictionary<string, object> BuildRoutes(TestInfo testInfo, string sendModule, string receiveModule)
```

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: It's used in BuildAddInitialConfig. I didn't have it there at the time of you reviewing, but it should have been. Added it now.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: See answer to your other comment

test/modules/load-gen/Settings.cs
dylanbronson: Did some extra validation

test/modules/load-gen/Settings.cs
dylanbronson: Default. Now will throw exception if TTL and priorities both exist and TtlThreshold is <= 0

test/modules/load-gen/PriorityMessageSender.cs
philipktlin: Please add a comment here.

test/modules/load-gen/Settings.cs
```diff
@@ -56,9 +60,27 @@ static Settings Create()
...
+
+            if (!(ttls == null && priorities == null && ttlThresholdValue <= 0)
+                && !(ttls != null && priorities != null && ttlThresholdValue > 0))
+            {
+                throw new ArgumentException("TTLs, priorities, and TTLThreshold must either all be set or all not be set.");
+            }
+            else if (ttls.Count != priorities.Count && ttls.Count != 0)
```
philipktlin, line 78: if ttls is null, priorities is null and ttlThresholdValue is 0, it will evaluate this line.  And it will throw NullReferenceException when calls ttls.Count.
```suggestion
            else if (ttls?.Count != 0 && ttls?.Count != priorities?.Count)
```

dylanbronson: Good catch

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -39,34 +39,17 @@ public async Task PriorityQueueModuleToModuleMessages()
             string trackingId = Guid.NewGuid().ToString();
             TestInfo testInfo = this.InitTestInfo(5, 1000, true);
 
-            Action<EdgeConfigBuilder> addInitialConfig = this.BuildAddInitialConfig(trackingId, trcImage, loadGenImage, testInfo);
+            string trcIPAddress = Dns.GetHostEntry(TrcModuleName).AddressList[0].ToString();
```
philipktlin, line 42: Are you sure you can resolve container ip from the host?  Also when you are running this line, container hasn't been deployed yet, right?

PR #2826, opened 2020-04-09
[k8s][Test] Change last probe time to last transtion time

ggjjj:
The Last probe time in the Pod condition is usually null. This results in exceptions in logs. So switching to Last Transition time of the pod conditions object to track the accurate Pod status. Added unit tests.


edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesRuntimeInfoProviderTest.cs
```diff
@@ -295,13 +295,13 @@ public async Task ReturnModuleStatusWithPodConditionsWhenThereAreNoContainers()
...
                     Message = "Ready",
                     Reason = "Scheduling",
                 },
                 new V1PodCondition
                 {
-                    LastProbeTime = new DateTime(2020, 02, 05, 10, 10, 15),
+                    LastTransitionTime = new DateTime(2020, 02, 05, 10, 10, 15),
                     Message = "persistentvolumeclaim module-a-pvc not found",
                     Reason = "Unschedulable",
                 }
```
darobs, line 307: Should you also add a test where pod.Status.Conditions==null?

PR #2828, opened 2020-04-10
Restart edge agent when config is updated

damonbarry:
Currently, when a customer deploys changes to the Edge Agent module, e.g. updated environment variables or Docker `createOptions`, we don't detect them and therefore we don't update/restart the module. To change Edge Agent you must update its Docker image in the deployment, which is severe and cumbersome.

To fix this, we need to change the way we compare desired vs. current information about the Edge Agent module. Currently we override the default (and more comprehensive) module comparison with a lightweight test of the name (always "edgeAgent"), the type (always "docker" in this scenario) and the image. We could simply remove the override to detect other differences, but that would break a specific scenario: when the Edge Agent container is starting up for the first time.

We gather information about the current module environment from several places, including a cached copy of the last known edge configuration (as deployed from the cloud). Specifically, environment variables and Docker `createOptions` come from the cached config. When the Edge Agent container is starting up for the first time, its reconcile loop will compare desired vs. current modules at least once _before_ it can cache the config. During those first comparisons, information about the current module environment will be incomplete. If we do the more comprehensive comparison in this scenario we will falsely detect differences, and Edge Agent will restart endlessly.

To handle this case, we add a flag to EdgeAgentDockerRuntimeModule to signal that the config cache is empty. When we detect this flag we know to do the lightweight comparison to avoid false positives. Otherwise we can do the comprehensive comparison to detect _any_ changes to Edge Agent.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
@@ -4,26 +4,57 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker
     using System.Collections.Generic;
     using Microsoft.Azure.Devices.Edge.Agent.Core;
     using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
     using Newtonsoft.Json;
 
     public class EdgeAgentDockerModule : DockerModule, IEdgeAgentModule
     {
+        readonly string json;
+
         [JsonConstructor]
         public EdgeAgentDockerModule(string type, DockerConfig settings, ImagePullPolicy imagePullPolicy, ConfigurationInfo configuration, IDictionary<string, EnvVal> env, string version = "")
             : base(Core.Constants.EdgeAgentModuleName, version, ModuleStatus.Running, RestartPolicy.Always, settings, imagePullPolicy, Core.Constants.HighestPriority, configuration, env)
         {
             Preconditions.CheckArgument(type?.Equals("docker") ?? false);
+            this.json = JsonConvert.SerializeObject(this, Formatting.Indented);
         }
 
         public override bool Equals(IModule<DockerConfig> other)
         {
             if (ReferenceEquals(null, other))
                 return false;
+
+            void LogResult(string equalResult)
+            {
+                ILogger log = Logger.Factory.CreateLogger<EdgeAgentDockerModule>();
+                log.LogInformation(
+                     "\n=================================" +
+                    $"\nEdgeAgentDockerModule.Equals => {equalResult}" +
+                    $"\nTHIS:\n{this.json}" +
+                    $"\nOTHER:\n{JsonConvert.SerializeObject(other, Formatting.Indented)}" +
+                     "\n================================="); // +
+                    // $"\n{System.Environment.StackTrace}");
+            }
+
             if (ReferenceEquals(this, other))
+            {
+                LogResult("REFERENCE EQUALS");
                 return true;
-            return string.Equals(this.Name, other.Name) &&
+            }
+
+            bool result = string.Equals(this.Name, other.Name) &&
                 string.Equals(this.Type, other.Type) &&
                 string.Equals(this.Config.Image, other.Config.Image);
+
+            // Perform a full equality test in all cases except when a new Edge Agent container
+            // starts up and hasn't cached it's config yet
+            if (result && !(other is EdgeAgentDockerRuntimeModule otherRuntime && otherRuntime.ConfigIsEmpty))
+            {
+                result = base.Equals(other);
+            }
+
+            LogResult(result ? "EQUALS" : "NOT EQUALS");
+            return result;
         }
 
         public override int GetHashCode()
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
@@ -4,19 +4,15 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker
     using System.Collections.Generic;
     using Microsoft.Azure.Devices.Edge.Agent.Core;
     using Microsoft.Azure.Devices.Edge.Util;
-    using Microsoft.Extensions.Logging;
     using Newtonsoft.Json;
 
     public class EdgeAgentDockerModule : DockerModule, IEdgeAgentModule
     {
-        readonly string json;
-
         [JsonConstructor]
         public EdgeAgentDockerModule(string type, DockerConfig settings, ImagePullPolicy imagePullPolicy, ConfigurationInfo configuration, IDictionary<string, EnvVal> env, string version = "")
             : base(Core.Constants.EdgeAgentModuleName, version, ModuleStatus.Running, RestartPolicy.Always, settings, imagePullPolicy, Core.Constants.HighestPriority, configuration, env)
         {
             Preconditions.CheckArgument(type?.Equals("docker") ?? false);
-            this.json = JsonConvert.SerializeObject(this, Formatting.Indented);
         }
 
         public override bool Equals(IModule<DockerConfig> other)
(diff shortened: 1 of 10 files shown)


philipktlin:
I don't fully understand. What I am thinking in high level is EdgeAgent will get desired and current config, compare and generate deployment plan.  For desired config, it should get from IoT Hub twin or cached config.  For current config, it should always get from container API (Docker).  Why there is a problem during first time.

"During those first comparisons, information about the current module environment will be incomplete."  What is incomplete? Would you give me an example for this statement?

damonbarry:
> I don't fully understand. What I am thinking in high level is EdgeAgent will get desired and current config, compare and generate deployment plan. For desired config, it should get from IoT Hub twin or cached config.

Correct

>For current config, it should always get from container API (Docker). Why there is a problem during first time.
> 
> "During those first comparisons, information about the current module environment will be incomplete." What is incomplete?

The agent compiles the current config from a few different sources, including the docker API. Things like environment variables, `image`, and `createOptions`, however, do NOT come from docker. I don't know all the design/technical decisions that went into this code; for whatever reason that information comes from a cached copy of the last deployed config (i.e., the agent's last desired property update). So until you cache that information for the first time, you don't have it.

> Would you give me an example for this statement?

Here's an example comparison of edge agent desired vs current. Notice DESIRED has environment variables; CURRENT does not:
```
DESIRED:
{
  "version": "",
  "status": "running",
  "restartPolicy": "always",
  "imagePullPolicy": "on-create",
  "priority": 0,
  "type": "docker",
  "settings": {
    "image": "mcr.microsoft.com/azureiotedge-agent:1.0",
    "createOptions": "{}"
  },
  "env": {
    "hello": {
      "value": "world"
    }
  }
}
CURRENT:
{
  "exitCode": 0,
  "statusDescription": "running",
  "lastStartTimeUtc": "2020-04-08T23:31:44.9800362",
  "lastExitTimeUtc": "0001-01-01T00:00:00",
  "runtimeStatus": "running",
  "imagePullPolicy": "on-create",
  "priority": 0,
  "type": "docker",
  "settings": {
    "image": "mcr.microsoft.com/azureiotedge-agent:1.0",
    "imageHash": "sha256:cab1bd67d4d7d71b6eedcbfa5382e8e784881bef26aab1ed846974191381ef25",
    "createOptions": "{}"
  },
  "env": {}
}
```

philipktlin:
Quoted "We gather information about the current module environment from several places, including a cached copy of the last known edge configuration (as deployed from the cloud). Specifically, environment variables and Docker createOptions come from the cached config."

Can we get current module running all from docker API inclduing environment variables? if not, why?

darobs:
I think this basically works, and it will not negatively impact k8s mode.

The general risk here is still the [planner's CreateShutdownPlanAsync()](https://github.com/Azure/iotedge/blob/f72152408348caf81a7a83bf0436bc98c767a3ac/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/planners/HealthRestartPlanner.cs#L57) which will stop all active modules when edgeAgent is stopped.

damonbarry:
> Quoted "We gather information about the current module environment from several places, including a cached copy of the last known edge configuration (as deployed from the cloud). Specifically, environment variables and Docker createOptions come from the cached config."
> 
> Can we get current module running all from docker API inclduing environment variables? if not, why?

Again, I don't know all the design/technical reasons that went into the current architecture. And I think we _might_ be able to get that information from Docker. But I also think it would be very difficult to put that information back into the right format so you can easily compare it with the desired config.

For example, in a deployment you pass createOptions as stringized JSON. There are LOTS of different things the customer can put into that string, and if you do something like `docker container inspect` to get it back out again, it does not return a "createOptions" string; it returns structured output. You'd have to discover and collect the needed information from possibly a hundred different places in the structured output. You'd basically have to implement a full docker response parser. And even if you did I think you might miss corner cases.

Another example is the environment variables. Docker inserts the environment variables you passed in the deployment config, but it also inserts several others. If you query docker for the env vars, how do you know which ones are from the deployment?

I suspect that caching the config exactly as you received it from the agent's desired properties is much easier.

damonbarry:
> The general risk here is still the [planner's CreateShutdownPlanAsync()](https://github.com/Azure/iotedge/blob/f72152408348caf81a7a83bf0436bc98c767a3ac/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/planners/HealthRestartPlanner.cs#L57) which will stop all active modules when edgeAgent is stopped.

@darobs Isn't it a good thing that stopping the agent stops all modules, for security? What's the risk?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
```diff
@@ -19,11 +19,22 @@ public override bool Equals(IModule<DockerConfig> other)
         {
             if (ReferenceEquals(null, other))
                 return false;
+
             if (ReferenceEquals(this, other))
                 return true;
-            return string.Equals(this.Name, other.Name) &&
+
+            bool result = string.Equals(this.Name, other.Name) &&
```
philipktlin, line 26: Should we modify like this as name, type and config are already covered in base.Equals method:

// Perform a full equality test in all cases except when a new Edge Agent container
// starts up and hasn't cached it's config yet
if (other is EdgeAgentDockerRuntimeModule otherRuntime && !otherRuntime.ConfigIsEmpty)
{
    return base.Equals(other);
}

return string.Equals(this.Name, other.Name) &&
    string.Equals(this.Type, other.Type) &&
    string.Equals(this.Config.Image, other.Config.Image);

philipktlin:
I studied the code for a while and I think I understand the deployment flow now.

1. > The agent compiles the current config from a few different sources, including the docker API. Things like environment variables, image, and createOptions, however, do NOT come from docker.

Correct this statement, I think currently image is getting from docker and that's why image is used in equals method.

2. The current implementation is difficult to understand as Equals method is used for deployment comparison and deployment compares different fields if 1st time run or subsequent runs.  This is very odd for Equals method.

3.  This solution involves smallest change to fix the problem.  But I think it is better to let iotedged to store edgeAgent deployment config.

damonbarry:
> Correct this statement, I think currently image is getting from docker and that's why image is used in equals method.

You're right, image comes from docker, my mistake.

> 1. The current implementation is difficult to understand as Equals method is used for deployment comparison and deployment compares different fields if 1st time run or subsequent runs.  This is very odd for Equals method.

I agree it's not ideal. I also don't like the downcasting (which I assume requires reflection in .NET?) in this change and in `DockerRuntimeModule.Equals(IModule<DockerConfig> other)`.

> 2. This solution is good as it involves smallest change to fix the problem.  But I think it is better to let iotedged to store edgeAgent and other modules deployment config.  This will fix another issue that other module will redeploy if edgeAgent redeploy, even other modules' config doesn't change.

When edge agent restarts, it already executes a plan that stops, removes, creates, and starts all the other modules too.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
damonbarry: It's a choice between doing the downcast every time to ensure a single batch of string comparisons, or repeating the string comparisons occasionally (when agents with the same image--and represented by the same type--are being compared) and minimizing the downcasts. You'd really need to measure to decide which one is better, but I don't think this code is in a particularly hot path...

philipktlin:
I think it is good to go.  @varunpuranik any comments?

varunpuranik:
So if I understand this right - this will work most of the time, except of the case when there are 2 deployments back to back. In that case if EdgeAgent is restarting as the new deployment gets applied, EdgeAgent will miss that? (assuming the image has not changed but some other config such as environment variables or config has changed?)

damonbarry:
> So if I understand this right - this will work most of the time, except of the case when there are 2 deployments back to back. In that case if EdgeAgent is restarting as the new deployment gets applied, EdgeAgent will miss that? (assuming the image has not changed but some other config such as environment variables or config has changed?)

I think you're right...it is assumed that the first config a new agent module receives/caches is the "current" config. But if the received config hasn't been applied yet (because it replaced the _actual_ current config while edge agent was restarting) then it will effectively spoof the "current" config and no action will be taken. @varunpuranik Is that what you're thinking?

damonbarry:
@varunpuranik Given that this PR fixes the problem except in one corner case (new config deployed while agent is restarting), and given that the fix for the corner case will include changes to the daemon, I'd like to get this PR merged as-is, and address the corner case in a separate PR. Can you approve?

varunpuranik:
> @varunpuranik Given that this PR fixes the problem except in one corner case (new config deployed while agent is restarting), and given that the fix for the corner case will include changes to the daemon, I'd like to get this PR merged as-is, and address the corner case in a separate PR. Can you approve?

yes, sure.

damonbarry:
In talking to @varunpuranik about the solution for the corner case, it sounds like the work I do there will likely replace this rather than amend it. So I'll leave this open for now, but probably close it once I vet the other solution.

damonbarry:
Closing this PR in favor of #2889.

PR #2830, opened 2020-04-10
Update k8s-openapi to v0.7.1

arsing:
This version requires a lot less CPU and RAM to compile.

Since this version uses http:0.2 types, they need extra work to be converted
to the http:0.1 types that our codebase uses (via hyper 0.12).

Also update the nightly used by the Rust code coverage job because that seems
to fix the "ld crashes with exit code 1 and no error message" issue
that the code coverage job has been having for the last few days.

### DIFF
--- a/edgelet/kube-client/src/client.rs
+++ b/edgelet/kube-client/src/client.rs
@@ -18,7 +18,10 @@ use k8s_openapi::api::authorization::v1 as api_authorize;
 use k8s_openapi::api::core::v1 as api_core;
 use k8s_openapi::api::rbac::v1 as api_rbac;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1 as api_meta;
-use k8s_openapi::{http, DeleteOptional, ListOptional, Response as K8sResponse, ResponseBody};
+use k8s_openapi::{
+    http, CreateOptional, CreateResponse, DeleteOptional, DeleteResponse, List, ListOptional,
+    ListResponse, ReplaceOptional, ReplaceResponse, Response as K8sResponse, ResponseBody,
+};
 use log::{debug, trace};
 
 use crate::config::{Config, TokenSource};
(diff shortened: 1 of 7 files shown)


edgelet/kube-client/src/client.rs
```diff
@@ -809,27 +791,51 @@ where
                 Error::from(err.context(ErrorKind::UrlJoin(self.config.host().clone(), path)))
             })
             .and_then(|url| {
-                // set the full URL on the request including API path
-                *req.uri_mut() = url.as_str().parse::<Uri>().context(ErrorKind::Uri(url))?;
+                // req is an http 0.2 Request but hyper uses http 0.1, so destructure req and reassemble it.
```
massand, line 794: Question - Is there advantages to switching to hyper (which seems to use an older version of http - 0.1)?

arsing: No, it's just to have a reference to http 0.1. The alternative would've been to add a dep on http:0.1, remove the `k8s_openapi::http` import, and change all uses of the imported `http::` to fully-qualified `k8s_openapi::http::`

darobs:
Making an images build from this so I can test this out on a cluster....

edgelet/kube-client/src/client.rs
```diff
@@ -94,11 +97,10 @@ where
             },
             ..api_authorize::SelfSubjectAccessReview::default()
         };
-        let params = api_authorize::CreateSelfSubjectAccessReviewOptional::default();
 
         api_authorize::SelfSubjectAccessReview::create_self_subject_access_review(
             &subject_access_review,
-            params,
+            CreateOptional::default(),
```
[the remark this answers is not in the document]
arsing: I would personally have written `Default::default()` which is also how it's documented in the `k8s-openapi` docs. That would've worked even in 0.5.1 and been even more concise :)

PR #2835, opened 2020-04-10
Makes pubacks and pinqreq processed earlier

dmolokanov:


myagley:
What's the context for this change?

dmolokanov:
> What's the context for this change?
We did run perf tests and found out that the broker
1. is quite easily absorbs mqtt messages but quite slow at sending messages to a subscriber
2. often broker fails to answer ping response within a reasonable time that leads clients to disconnect.

So I had an idea to prioritize pubacks and pingreq via side channel to broker in order to mitigate slow-ish response.

All tests showed worse results than without side-channel :(

myagley:
Ok, we have to be careful with changes like this from a correctness stand point.

It's probably safe to side channel the `PingResp` but I'm nervous about the `PubAck`. I think the acks need to be processed by the broker thread before being sent?

myagley:
If we can prove this is still correct from a protocol perspective, I'm all for the change. Just want to make sure the optimization doesn't break us. Being correct is more important than being fast.

PR #2839, opened 2020-04-13
[mqtt] translation

lfitchett:
This is a poc for the broker legacy translation. It handles all the translation needed to request a twin. It doesn't handle notifying EdgeHub Core of the subscription.

Twin Update - Copy


mqtt/mqtt-broker/src/translation.rs
```diff
@@ -0,0 +1,81 @@
...
+        subscribe_to,
+    }
+}
+
+fn unsubscribe(client_id: &str, mut unsubscribe: proto::Unsubscribe) -> proto::Unsubscribe {
+    unsubscribe.unsubscribe_from = unsubscribe
+        .unsubscribe_from
+        .into_iter()
+        .map(|t| to_new_topic(client_id, t))
+        .collect();
```
lfitchett, line 43: @dmolokanov Is there a function instead of collect that will re-use the existing vector? I remember seeing something like that before

mqtt/mqtt-broker/src/translation.rs
```diff
@@ -0,0 +1,81 @@
...
+    publish.topic_name = to_legacy_topic(publish.topic_name);
+
+    publish
+}
+
+fn to_new_topic(client_id: &str, topic: String) -> String {
+    //TODO: make this cover all iothub topics
+    match topic.as_ref() {
+        "$iothub/twin/res" => format!("$edgehub/{}/twin/res", client_id),
+        "$iothub/twin/GET" => format!("$edgehub/{}/twin/get", client_id),
```
vipeller, line 64: I think the logic is a bit more complex than this: after the res and GET part some other information comes, e.g
$iothub/twin/GET/?$rid=1234   => $edgehub/client_id/twin/GET/?$rid=1234

mqtt/mqtt-broker/src/translation.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+fn to_legacy_topic(topic: String) -> String {
+    const DEVICE_ID: &str = r"[a-zA-Z-\.\+%_#\*\?!\(\),=@\$']";
+    lazy_static! {
+        static ref TWIN_PATTERN: Regex = Regex::new(&format!("\\$edgehub/{}/twin/get", DEVICE_ID))
+            .expect("failed to create new Regex from pattern");
+    }
+
```
vipeller, line 75: I have a feeling that because of the restricted form we need to match (starting with $edgehub) and the simple operation, (we just cut out the client id most of the time), regex might be not the fastest way to do this. As this runs at every publish, probably we should run some tests performance wise

mqtt/mqtt-broker/src/translation.rs
dmolokanov: Collect usually operates on iterators, not on vectors, and has corresponding semantic. If you want to reuse vector you need to iterate it with index and update in place. You may win some cycles, but it is subject to benchmark IMO

mqtt/mqtt-broker/src/translation.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+fn subscribe(client_id: &str, subscribe: proto::Subscribe) -> proto::Subscribe {
+    let proto::Subscribe {
+        packet_identifier,
+        mut subscribe_to,
+    } = subscribe;
+
+    for mut sub_to in &mut subscribe_to {
+        sub_to.topic_filter = to_new_topic(client_id, sub_to.topic_filter.clone());
```
dmolokanov, line 29: Unnecessary clone of the topic

mqtt/mqtt-broker/src/translation.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+pub fn translate_outgoing(packet: proto::Packet) -> proto::Packet {
+    match packet {
+        proto::Packet::Publish(p) => proto::Packet::Publish(publish_to(p)),
+        p => p,
+    }
+}
+
+fn subscribe(client_id: &str, subscribe: proto::Subscribe) -> proto::Subscribe {
```
dmolokanov, line 22: I'd say we need to introduce a notion of IothubTopicFilter here, so we can get rid of cloning topics and re-use elsewhere
```rust
fn subscribe(client_id: ClientId, subscribe: proto::Subscribe) -> proto::Subscribe {
    let subscribe_to = subscribe.subscribe_to.into_iter().map(|to| {
        proto::SubscribeTo {
            topic_filter: IotHubTopic::new(client_id.clone(), topic_filter).to_string(),
            qos: to.qos
        }
    }.collect();

    proto::Subscribe {
        packet_identifier:subscribe.packet_identifier,
        subscribe_to,
    }
}

struct IotHubTopic(String);

impl IotHubTopic {
    fn new(client_id:ClientId, topic_filter: String) -> Self {
        let topic = match &topic_filter {
            "$iothub/twin/res" => format!("$edgehub/{}/twin/res", client_id),
            "$iothub/twin/GET" => format!("$edgehub/{}/twin/get", client_id),
            _ => topic,
        };
    
        Self(topic)
    }
}

impl ToString for IotHubTopic {
    fn to_string(self) -> String {
        self.0
    }
}
```

mqtt/mqtt-broker/src/translation.rs
lfitchett: ok, that can be fixed with regex and captures, I'll look into that

mqtt/mqtt-broker/src/translation.rs
lfitchett: yea, an easy optimization is if starts with `$edgehub`, start doing regex, otherwise nothing. I didn't bother for the poc

mqtt/mqtt-broker/src/translation.rs
```diff
@@ -0,0 +1,81 @@
+use crate::ClientEvent;
```
dmolokanov, line 1: As a quick check of an idea, this may be good, but in order to check it in the master, this code is to be oxidized :)

myagley:
What's the background here? I want to make sure we aren't putting any IoT Hub specific things into the broker. Anything IoT Hub specific needs to go into the bridge.

varunpuranik:
> What's the background here? Does this stem from the fact that IoT Hub's topics don't include the device id?
> 
> I want to make sure we aren't putting any IoT Hub specific things into the broker. Anything IoT Hub specific needs to go into the bridge or into a separate `mqtt-iothub` crate. The `mqtt-broker` crate needs to remain pure. IoT Hub specific cruft needs to be isolated.

Yes, this is to allow us to keep track of "who" this subscription came from. 
And I agree with you - the root crate should not have any IoHub specific stuff. We need to create APIs so that we can put the IoTHub stuff in a derived crate (like we do in EdgeAgent or in edgelet for docker).
This is just a POC to give us an idea of how the code would look like.

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -985,6 +988,13 @@ pub struct BrokerHandle(Sender<Message>);
 
 impl BrokerHandle {
     pub async fn send(&mut self, message: Message) -> Result<(), Error> {
+        let message = match message {
+            Message::Client(client_id, event) => {
+                Message::Client(client_id, TRANSLATOR.translate_outgoing(event))
```
dmolokanov, line 993: This is not a channel from broker to client it is from client to broker

mqtt/Cargo.toml
```diff
@@ -3,6 +3,7 @@ members = [
     "mqtt3",
     "mqtt-broker",
     "mqttd",
+    "topic-translator",
```
dmolokanov, line 6: `mqtt-edgehub` maybe?

mqtt/mqtt-broker/tests/translation.rs
```diff
@@ -0,0 +1,264 @@
...
+        .authorizer(|_| Ok(true))
+        .build();
+
+    let mut server_handle = common::start_server(broker);
+
+    let mut edge_hub_core = TestClientBuilder::new(server_handle.address())
+        .client_id(ClientId::IdWithCleanSession("edge_hub_core".into()))
+        .build();
+    let mut device_1 = TestClientBuilder::new(server_handle.address())
+        .client_id(ClientId::IdWithCleanSession("device_1".into()))
```
vadim-kovalyov, line 26: I think we should start playing with more complex device names. Not necessary in this PR, but we need to test real prod-line device ids. Also, maybe add proptests for ids.

mqtt/topic-translator/src/translation.rs
```diff
@@ -0,0 +1,363 @@
...
+            );
+            publish.topic_name = new_topic;
+        }
+
+        publish
+    }
+}
+
+const DEVICE_ID: &str = r"(?P<device_id>.+)";
+macro_rules! translate_d2c {
```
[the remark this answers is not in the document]
lfitchett: It was the easiest way to do it without worrying about rust sizing. If I used vectors for example the lambdas would have to be boxed

mqtt/topic-translator/src/translation.rs
```diff
@@ -0,0 +1,363 @@
...
+use regex::Regex;
+use tracing::debug;
+
+lazy_static! {
+    pub static ref TRANSLATOR: Translator = Translator::new();
+}
+
+pub struct Translator {
+    d2c: TranslateD2C,
+    c2d: TranslateC2D,
```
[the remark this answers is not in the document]
lfitchett: It isn't that simple. 
d2c only translates (broker) incoming publishes to new. 
c2d translates (broker) incoming subscribes to new and (broker) outgoing publishes to old.

mqtt/mqtt-broker/tests/translation.rs
```diff
@@ -0,0 +1,264 @@
...
+        .build();
+
+    // Core subscribes
+    edge_hub_core
+        .subscribe("$edgehub/+/twin/get/#", QoS::AtLeastOnce)
+        .await;
+    edge_hub_core
+        .subscribe("$edgehub/subscriptions/device_1", QoS::AtLeastOnce)
```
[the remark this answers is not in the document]
lfitchett: Nope. eh core will subscribe to $edgehub/connections to get all connected devices. Then for each connected device that has an iothub identity (except itself) it will subscribe to $edgehub/subscriptions/{clientid}

mqtt/mqtt-broker/tests/translation.rs
```diff
@@ -0,0 +1,264 @@
...
+    // device requests twin update
+    device_1
+        .subscribe("$iothub/twin/res/#", QoS::AtLeastOnce)
+        .await;
+    device_1
+        .publish_qos1("$iothub/twin/GET/?rid=10", "", false)
+        .await;
+
+    // Core recieves request
+    recieve_with_topic(&mut edge_hub_core, "$edgehub/device_1/twin/get/?rid=10").await;
```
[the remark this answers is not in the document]
lfitchett: I think its harder to read, since the assert ends up split onto 5 or 6 lines. I find it easier to read with the function, especially since the same assert is used like 20 times

mqtt/mqtt-edgehub/src/translation.rs
```diff
@@ -0,0 +1,377 @@
...
+ ***********************************************************************************************************************
+ */
+
+use lazy_static::lazy_static;
+use mqtt3::proto;
+use regex::Regex;
+use tracing::debug;
+
+lazy_static! {
+    pub static ref TRANSLATOR: Translator = Translator::new();
```
vadim-kovalyov, line 18: nit: maybe you don't even need an exposed instance of the Translator? why not make `incoming_subscribe, etc, just a module funcs and call them from Connection?

mqtt/mqtt-broker/tests/translation.rs
vadim-kovalyov: I just tried this allowed device name and it breaks the test. I think we need a work item and a separate PR to add proptests and fix all corner cases.
```
fasdfsfd67iasdifu2341_(floor1)-=+*asfd!@#$%',.?das_1
```

mqtt/mqtt-edgehub/src/translation.rs
dmolokanov: I agree, Translator doesn't have any state, just functions. Now macros generate too many "unnecessary" internal structs, that just hold Regex. Could we try to simplify this a bit? Like, have a list of statically created Regexes for each required case and function? Probably you could make nested modules for groups in and out directions for the one topic? something like

```rust
pub mod publish {
  lazy_static! {
    REGEX_INCOMING = ,
    REGEX_OUTGOING = ,
}
  fn translate_incoming(...) -> Publish {
     if topic.starts_with("$iothub/") || topic.starts_with("devices/") {
        super::translate_incoming(REGEX_INCOMING, publish)
    }
  }

   fn translate_outgoing(...) -> Publish {
     if topic.starts_with("$edgehub/") {
        super::translate_outgoing(REGEX_INCOMING, publish)
    }
  }
}
  
```

mqtt/mqtt-edgehub/src/translation.rs
```diff
@@ -0,0 +1,361 @@
...
+        to_new {
+            "\\$iothub/twin/PATCH/properties/desired/(?P<params>.*)",
+            {|captures: regex::Captures<'_>, client_id| format!("$edgehub/{}/twin/desired/{}", client_id, &captures["params"])}
+        },
+        to_old {
+            format!("\\$edgehub/{}/twin/desired/(?P<params>.*)", DEVICE_ID),
+            {|captures: regex::Captures<'_>| format!("$iothub/twin/PATCH/properties/desired/{}", &captures["params"])}
+        }
+    },
+    twin_response_from_hub {
```
dmolokanov, line 229: missed comment?

mqtt/mqtt-edgehub/src/translation.rs
```diff
@@ -0,0 +1,361 @@
...
+    },
+
+    // Twin Translation
+    twin_send_update_to_hub {
+        to_new {
+            "\\$iothub/twin/PATCH/properties/reported/(?P<params>.*)",
+            {|captures: regex::Captures<'_>, client_id| format!("$edgehub/{}/twin/reported/{}", client_id, &captures["params"])}
+        }
+    },
+    twin_get_from_hub {
```
dmolokanov, line 189: missed comment?

mqtt/mqtt-edgehub/src/translation.rs
lfitchett: nope, its part of the "Twin Translation" group.

mqtt/mqtt-edgehub/src/translation.rs
lfitchett: same as above

mqtt/mqtt-edgehub/src/translation.rs
lfitchett: // Message Translation
c2d_message

// Twin Translation
twin_receive_update_from_hub 
twin_response_from_hub

// Direct Methods
receive_direct_method_request

PR #2841, opened 2020-04-14
disable edgelet code coverage

lfitchett:
There is a strange linker error

### DIFF
--- a/builds/checkin/edgelet.yaml
+++ b/builds/checkin/edgelet.yaml
@@ -140,6 +140,12 @@ jobs:
       - bash: edgelet/build/linux/clippy.sh
         displayName: Clippy
 
+# The CI VMs encounter
+#
+# >ld returned 1 exit status
+#
+# ... when running this job for no obvious reason.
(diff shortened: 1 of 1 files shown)


builds/checkin/edgelet.yaml
```diff
@@ -140,65 +140,65 @@ jobs:
...
-          codeCoverageTool: cobertura
-          summaryFileLocation: "edgelet/coverage.xml"
-      - task: mspremier.BuildQualityChecks.QualityChecks-task.BuildQualityChecks@5
-        displayName: "Check build quality"
-        inputs:
-          checkCoverage: true
-          coverageFailOption: fixed
-          coverageType: lines
-          coverageThreshold: $(coverage.goal)
+# ################################################################################
```
arsing, line 143: Add a comment here explaining why it's commented out. Something like

```
# The CI VMs encounter
#
# >ld returned 1 exit status
#
# ... when running this job for no obvious reason.
#
```

PR #2842, opened 2020-04-14
[mqtt] Convert broker task to a thread

myagley:
* Converts the broker processing from an async "task" to a proper thread.
* Converts the snapshotter processing from an async "task" to a proper thread.

This change leaves all of the async code in the networking code, where it is most beneficial.

In very early perf testing this has a perf benefit of ~25-30% over master.
Quick tests were run with the following load a single subscriber.

```sh
mqtt-benchmark  --broker tcp://127.0.0.1:1883 --count 10000 --size 1000 --clients 24 --qos 1 --format text -quiet true
```

**Current Task Model**
```
========= TOTAL (24) =========
Total Ratio:                 1.000 (240000/240000)
Total Runtime (sec):         4.626
Average Runtime (sec):       4.609
Msg time min (ms):           0.055
Msg time max (ms):           14.706
Msg time mean mean (ms):     0.459
Msg time mean std (ms):      0.001
Average Bandwidth (msg/sec): 2169.886
Total Bandwidth (msg/sec):   52077.253
```

**Thread Model (this PR)**
```
========= TOTAL (24) =========
Total Ratio:                 1.000 (240000/240000)
Total Runtime (sec):         3.564
Average Runtime (sec):       3.549
Msg time min (ms):           0.044
Msg time max (ms):           16.473
Msg time mean mean (ms):     0.353
Msg time mean std (ms):      0.001
Average Bandwidth (msg/sec): 2817.756
Total Bandwidth (msg/sec):   67626.151
```

**Mosquitto**
```
========= TOTAL (24) =========
Total Ratio:                 1.000 (240000/240000)
Total Runtime (sec):         3.469
Average Runtime (sec):       3.459
Msg time min (ms):           0.033
Msg time max (ms):           13.980
Msg time mean mean (ms):     0.345
Msg time mean std (ms):      0.001
Average Bandwidth (msg/sec): 2890.697
Total Bandwidth (msg/sec):   69376.717
```

Further testing is required. These are just quick comparisons on my laptop.

### DIFF
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -1007,7 +1007,7 @@ pub(crate) mod tests {
     use proptest::collection::{hash_map, vec};
     use proptest::prelude::*;
     use tokio::sync::mpsc::error::TryRecvError;
-    use tokio::sync::mpsc::{self, Receiver};
+    use tokio::sync::mpsc::{self, UnboundedReceiver};
     use uuid::Uuid;
 
     use mqtt3::{proto, PROTOCOL_LEVEL, PROTOCOL_NAME};
@@ -1035,7 +1035,7 @@ pub(crate) mod tests {
 
     fn connection_handle() -> ConnectionHandle {
         let id = Uuid::new_v4();
-        let (tx1, _rx1) = mpsc::channel(128);
+        let (tx1, _rx1) = mpsc::unbounded_channel();
         ConnectionHandle::new(id, tx1)
     }
 
@@ -1093,7 +1093,7 @@ pub(crate) mod tests {
             protocol_level: PROTOCOL_LEVEL,
         };
         let id = Uuid::new_v4();
-        let (tx1, mut rx1) = mpsc::channel(128);
+        let (tx1, mut rx1) = mpsc::unbounded_channel();
         let conn1 = ConnectionHandle::new(id, tx1.clone());
         let conn2 = ConnectionHandle::new(id, tx1);
         let client_id = ClientId::from("blah".to_string());
@@ -1106,14 +1106,12 @@ pub(crate) mod tests {
                 client_id.clone(),
                 ClientEvent::ConnReq(req1),
             ))
-            .await
             .unwrap();
         broker_handle
             .send(Message::Client(
                 client_id.clone(),
                 ClientEvent::ConnReq(req2),
             ))
-            .await
             .unwrap();
 
         assert_matches!(
@@ -1155,8 +1153,8 @@ pub(crate) mod tests {
             protocol_name: PROTOCOL_NAME.to_string(),
             protocol_level: PROTOCOL_LEVEL,
         };
-        let (tx1, mut rx1) = mpsc::channel(128);
-        let (tx2, mut rx2) = mpsc::channel(128);
+        let (tx1, mut rx1) = mpsc::unbounded_channel();
+        let (tx2, mut rx2) = mpsc::unbounded_channel();
         let conn1 = ConnectionHandle::from_sender(tx1);
         let conn2 = ConnectionHandle::from_sender(tx2);
         let client_id = ClientId::from("blah".to_string());
@@ -1169,14 +1167,12 @@ pub(crate) mod tests {
                 client_id.clone(),
                 ClientEvent::ConnReq(req1),
             ))
-            .await
             .unwrap();
         broker_handle
             .send(Message::Client(
                 client_id.clone(),
                 ClientEvent::ConnReq(req2),
             ))
-            .await
             .unwrap();
 
         assert_matches!(
@@ -1214,7 +1210,7 @@ pub(crate) mod tests {
             protocol_name: "AMQP".to_string(),
             protocol_level: PROTOCOL_LEVEL,
         };
-        let (tx1, mut rx1) = mpsc::channel(128);
+        let (tx1, mut rx1) = mpsc::unbounded_channel();
         let conn1 = ConnectionHandle::from_sender(tx1);
         let client_id = ClientId::from("blah".to_string());
         let req1 = ConnReq::new(client_id.clone(), connect1, None, conn1);
@@ -1224,7 +1220,6 @@ pub(crate) mod tests {
                 client_id.clone(),
                 ClientEvent::ConnReq(req1),
             ))
-            .await
             .unwrap();
 
         assert_matches!(
@@ -1253,7 +1248,7 @@ pub(crate) mod tests {
             protocol_name: PROTOCOL_NAME.to_string(),
             protocol_level: 0x3,
         };
-        let (tx1, mut rx1) = mpsc::channel(128);
+        let (tx1, mut rx1) = mpsc::unbounded_channel();
         let conn1 = ConnectionHandle::from_sender(tx1);
         let client_id = ClientId::from("blah".to_string());
         let req1 = ConnReq::new(client_id.clone(), connect1, None, conn1);
@@ -1263,7 +1258,6 @@ pub(crate) mod tests {
                 client_id.clone(),
                 ClientEvent::ConnReq(req1),
             ))
-            .await
             .unwrap();
 
         assert_matches!(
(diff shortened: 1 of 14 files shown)


mqtt/mqtt-broker/src/connection.rs
```diff
@@ -97,7 +96,7 @@ where
     match codec.next().await {
         Some(Ok(Packet::Connect(connect))) => {
             let client_id = client_id(&connect.client_id);
-            let (sender, events) = mpsc::channel(128);
+            let (sender, events) = mpsc::unbounded_channel();
```
dmolokanov, line 99: It can cause infinite growth of message queue, right?

mqtt/mqtt-broker/tests/persist_failpoints.rs
```diff
@@ -71,28 +69,23 @@ async fn test_persistor(count: usize, ops: Vec<Op>) {
...
-        .enable_all()
-        .build()
-        .unwrap()
-        .block_on(async {
-            fail::cfg("filepersistor.load.spawn_blocking", "return").unwrap();
 
-            let tmp_dir = TempDir::new().unwrap();
-            let path = tmp_dir.path().to_owned();
-            let mut persistor = FilePersistor::new(path, ConsolidatedStateFormat::default());
+    fail::cfg("filepersistor.store.fileopen", "return").unwrap();
```
dmolokanov, line 80: This line looks weird. I assume you define that each failpoint to behave as `return`, right? but why only one of failpoints defined here?

mqtt/mqtt-broker/src/snapshot.rs
```diff
@@ -13,11 +15,10 @@ enum Event {
 pub struct StateSnapshotHandle(Sender<Event>);
 
 impl StateSnapshotHandle {
-    pub async fn send(&mut self, state: BrokerState) -> Result<(), Error> {
+    pub fn try_send(&mut self, state: BrokerState) -> Result<(), Error> {
```
dmolokanov, line 18: Why `StateSnapshotHandle.try_send` but `BrokerHandle.send` ?

mqtt/mqtt-broker/src/snapshot.rs
```diff
@@ -62,13 +62,35 @@ impl<P> Snapshotter<P> {
 
 impl<P> Snapshotter<P>
 where
-    P: Persist,
+    P: Persist + Send + 'static,
 {
-    pub async fn run(mut self) -> P {
-        while let Some(event) = self.events.recv().await {
+    pub async fn run(self) -> Result<P, Error> {
+        let (tx, rx) = tokio::sync::oneshot::channel::<()>();
```
dmolokanov, line 68: I couldn't read this method. It's rather complicated involving oneshot channel, separate thread, async tasks, and external channel. Can we make it simpler or more clear?

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -107,48 +121,44 @@ where
...
                 Ok(())
             }
-            ClientEvent::Disconnect(_) => self.process_disconnect(client_id).await,
-            ClientEvent::DropConnection => self.process_drop_connection(client_id).await,
-            ClientEvent::CloseSession => self.process_close_session(client_id).await,
-            ClientEvent::PingReq(ping) => self.process_ping_req(client_id, ping).await,
+            ClientEvent::Disconnect(_) => self.process_disconnect(&client_id),
+            ClientEvent::DropConnection => self.process_drop_connection(&client_id),
+            ClientEvent::CloseSession => self.process_close_session(&client_id),
+            ClientEvent::PingReq(ref ping) => self.process_ping_req(&client_id, ping),
```
dmolokanov, line 135:
```suggestion
            ClientEvent::PingReq(ping) => self.process_ping_req(&client_id, &ping),
```
Everywhere?

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -42,33 +37,52 @@ where
...
+                let state = self.broker_loop();
+                if let Err(_e) = tx.send(()) {
+                    error!("failed to send broker state to event loop on shutdown");
+                }
+                state
+            })
+            .expect("failed to spawn broker thread");
+
+        // wait for the thread to exit
+        rx.await.map_err(Error::ThreadShutdown)?;
```
dmolokanov, line 61: Didn't quite understand, why we need oneshot channel to wait until the thread is finished, while we do the same with `handle.join()`?

mqtt/mqtt-broker/src/connection.rs
myagley: yes, but in this case we are bounding it with the session (Dequeue and number of inflight messages). Generally, I would be scared about an unbounded queue. However, in this case we have a different mechanism to bound the number of messages.

The key here is that we cannot block this loop. So, the send needs to not block.

mqtt/mqtt-broker/tests/persist_failpoints.rs
myagley: This test is to make sure that the failpoints are actually enabled. It's to protect against someone disabling the feature in the Config.toml, and us unknowingly switching off the failpoints. I want to make sure that the proptest test is actually enabling failures.

mqtt/mqtt-broker/src/broker.rs
myagley: It's to propagate any panics that happen on the broker thread.
I'm pushing a change that handles panics better.

mqtt/mqtt-broker/src/broker.rs
myagley: You don't like the `ref`?

mqtt/mqtt-broker/src/snapshot.rs
myagley: I don't think I can simplify it. I can add comments to explain what it is doing?

We want two things:
1. this `run` function needs to be `async` so that it can be joined with the other futures in the `server`
2. We need to handle and propagate any panics on the thread to the event loop thread so that we properly exit, instead of continuing to run with a dead thread.

The `oneshot` is to make this function `async`. The `handle.join()` is to propagate any panics.

myagley:
> Changes seem more or less mechanical. Have some concerns about the authenticate method that potentially will make a call to a edgehub and block the broker thread. It used to suspend a broker task and give the thread to do some other work while waiting the response.

Yeah, we can't make a network call on the broker task, regardless of whether this is a thread or an async task. This would block all other clients until the call is complete.

We either need to make the external call on the connection's task or do authN locally without communicating with the Edge Hub. Perhaps there is some other way of sharing the info between the two processes. (Also, this is only an issue for SAS keys?)

mqtt/mqtt-broker/src/broker.rs
myagley: The `oneshot` is used to signal the event loop that the thread has exited. The `handle.join()` is to propagate any panics from the broker thread onto the event loop thread (so that the process exits/crashes)

mqtt/mqtt-broker/src/connection.rs
dmolokanov: Yeah, that's what I figured. But worth to ask!

mqtt/mqtt-broker/tests/persist_failpoints.rs
dmolokanov: I wonder if there is a way to make it in a more obvious way...

mqtt/mqtt-broker/src/broker.rs
dmolokanov: Yeah, for some reason `ref` doesn't look as ergonomic as without `ref`. Feels like an additional mental burden.

Interesting, it's not only for me :)
https://rust-lang.github.io/rfcs/2005-match-ergonomics.html

myagley:
@dmolokanov what do we want to do with this PR?

dmolokanov:
> @dmolokanov what do we want to do with this PR?

I thought the plan was to merge it, right? I belived last time you wanted to check it with policy engine.

myagley:
No, we should keep it separate from the policy changes.

myagley:
> I thought the plan was to merge it, right? I belived last time you wanted to check it with policy engine.

Can I get a review please? :)

PR #2845, opened 2020-04-15
[mqtt] Fix bug when broker attempts to send message to offline session

dmolokanov:
Compliance test caught a bug when the broker errors when tries to publish a message to offline session. This error originated because of the wrong assumption that accessing offline session is somewhat illegal operation. It turns out that it's not: accessing a persistent session for a disconnected client is just as regular operation as any other.

There are 2 ways to solve this error. When accessing `auth_id` for offline session
1. return special marker that allows the broker to publish messages to the offline session
2. store `auth_id` along with session in the broker state and return `auth_id` for a previously connected client. 

This PR solves the issue with approach 2.

### DIFF
--- a/mqtt/mqtt-broker/src/auth/mod.rs
+++ b/mqtt/mqtt-broker/src/auth/mod.rs
@@ -1,3 +1,5 @@
+use serde::{Deserialize, Serialize};
+
 mod authentication;
 mod authorization;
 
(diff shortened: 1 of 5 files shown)


dmolokanov:
`auth_id` for offline sessions is TBD topic. Closing this PR for now

PR #2863, opened 2020-04-21
Add metrics upload jitter

lfitchett:
This adds a random interval for the metrics upload. This means that if 1000 devices are provisioned at once, they won't all try to upload at once.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsWorker.cs
```diff
@@ -48,7 +48,8 @@ public MetricsWorker(IMetricsScraper scraper, IMetricsStorage storage, IMetricsP
         public void Start(TimeSpan scrapingInterval, TimeSpan uploadInterval)
         {
             this.scrape = new PeriodicTask(this.Scrape, scrapingInterval, scrapingInterval, Log, "Metrics Scrape");
-            this.upload = new PeriodicTask(this.Upload, uploadInterval, uploadInterval, Log, "Metrics Upload");
+            TimeSpan uploadJitter = new TimeSpan((long)(uploadInterval.Ticks * new Random().NextDouble()));
```
CindyXing, line 51: Rather than creating a random number for each upload, how about a random number for each client? That way different clients can upload at different time slot; and same client always upload at the same slot.

varunpuranik: There is only one upload task per client, so this should do the same thing as you are suggesting, I believe.

varunpuranik: Only concern would be that now your initial start time could be too soon (if the NextDouble returns anything < 0.3). Ideally the jitter would be 20% +/- the final upload time. 
(something like new TImeSpan((long)(uploadInterval.Ticks * new Random().Next(8, 12) / 10)

lfitchett: Is is a concern to start too soon? If there are no scrapes, it just skips the upload. So if it rngs to .01, it would be functionally the same as 1.01. I like having an even distribution, it doesn't make sense to me to restrict it to a 40% band. Is there a reason to do so?

@CindyXing varun is correct, it is just the initial startup delay that is random, after that is is a consistent interval.

CindyXing: I see. Thanks for the clarification.

PR #2865, opened 2020-04-22
Fix route config change

philipktlin:
When route endpoint is removing, endpoint executor's cancellation token source will be canceled.  And endpoint executor is running inside a loop to process messages and wait for endpoint executor FSM to run sending command.  Inside FSM, it will keep retry sending messages until delivered or any exception not able to retry.  If an invalid/non-existing endpoint is defined, it will keep retrying forever and endpoint executor is awaiting the result from the sending command forever.  Therefore we need to have a way to tell FSM that endpoint is removing and should stop retrying.  Since endpoint executor is awaiting forever, it block the removal of endpoint which will block to add/update endpoints (in Dispatcher.ReplaceRoutes method).

In order to let FSM aware of removing endpoints, cancellation token source in endpoint executor will pass its token to FSM when FSM creation.  Therefore FSM sending message logic will consider if this token is canceled when endpoint is removing.

Step to reproduce:
1. Run iotedge with tempsenor and tempfilter; set up to route messages from tempsensor to tempfilterXXX, which is an invalid route.
2. start to run iotedged and check log of tempFilter, there is no message shown.
3. fix route to tempfilter and deploy.
4. expected tempfilter log should start showing up records, but it doesn't since the route can't be removed and new route can't be added.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/statemachine/EndpointExecutorFsm.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/statemachine/EndpointExecutorFsm.cs
@@ -81,6 +81,7 @@ namespace Microsoft.Azure.Devices.Routing.Core.Endpoints.StateMachine
         readonly Timer retryTimer;
         readonly AsyncLock sync = new AsyncLock();
         readonly ISystemTime systemTime;
+        readonly CancellationToken endpointExecutorToken;
 
         volatile State state;
         volatile int retryAttempts;
@@ -91,12 +92,12 @@ namespace Microsoft.Azure.Devices.Routing.Core.Endpoints.StateMachine
         volatile IProcessor processor;
         TimeSpan retryPeriod;
 
-        public EndpointExecutorFsm(Endpoint endpoint, ICheckpointer checkpointer, EndpointExecutorConfig config)
-            : this(endpoint, checkpointer, config, SystemTime.Instance)
+        public EndpointExecutorFsm(Endpoint endpoint, ICheckpointer checkpointer, EndpointExecutorConfig config, CancellationToken endpointExecutorToken)
+            : this(endpoint, checkpointer, config, SystemTime.Instance, endpointExecutorToken)
         {
         }
 
-        public EndpointExecutorFsm(Endpoint endpoint, ICheckpointer checkpointer, EndpointExecutorConfig config, ISystemTime systemTime)
+        public EndpointExecutorFsm(Endpoint endpoint, ICheckpointer checkpointer, EndpointExecutorConfig config, ISystemTime systemTime, CancellationToken endpointExecutorToken)
         {
             this.processor = Preconditions.CheckNotNull(endpoint).CreateProcessor();
             this.Checkpointer = Preconditions.CheckNotNull(checkpointer);
@@ -106,6 +107,7 @@ namespace Microsoft.Azure.Devices.Routing.Core.Endpoints.StateMachine
             this.retryPeriod = Timeout.InfiniteTimeSpan;
             this.lastFailedRevivalTime = checkpointer.LastFailedRevivalTime;
             this.unhealthySince = checkpointer.UnhealthySince;
+            this.endpointExecutorToken = endpointExecutorToken;
 
             if (checkpointer.LastFailedRevivalTime.HasValue)
             {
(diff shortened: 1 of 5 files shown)


edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/statemachine/EndpointExecutorFsm.cs
```diff
@@ -286,6 +290,15 @@ static async Task EnterSendingAsync(EndpointExecutorFsm thisPtr)
             ICollection<IMessage> messages = EmptyMessages;
             Stopwatch stopwatch = Stopwatch.StartNew();
             TimeSpan endpointTimeout = TimeSpan.FromMilliseconds(thisPtr.config.Timeout.TotalMilliseconds * thisPtr.Endpoint.FanOutFactor);
+
+            if (thisPtr.EndpointExecutorToken.IsCancellationRequested)
```
philipktlin, line 294: Do I still need to call Volatile.Read here?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/statemachine/EndpointExecutorFsm.cs
```diff
@@ -286,6 +290,15 @@ static async Task EnterSendingAsync(EndpointExecutorFsm thisPtr)
...
+            {
+                Events.SendCanceled(thisPtr);
+                next = Commands.Checkpoint(SinkResult<IMessage>.Empty);
```
philipktlin, line 297: is it ok to set next state to empty checkpoint? and do I still need to call RunInternalAsync?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/statemachine/EndpointExecutorFsm.cs
```diff
@@ -81,6 +81,7 @@ public class EndpointExecutorFsm : IDisposable
         readonly Timer retryTimer;
         readonly AsyncLock sync = new AsyncLock();
         readonly ISystemTime systemTime;
+        readonly CancellationToken endpointExecutorToken;
```
richma-ms, line 84: >endpointExecutorToken [](start = 35, length = 21)

no need to denote that this comes from the executor, right?  It can prob just be named cancellationToken

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/statemachine/EndpointExecutorFsm.cs
richma-ms: Something like Commands.Die is probably more appropriate, since at this point the endpoint is gone and these messages should just get dropped anyways

Either way I think you still do need to call RunInternalAsync, or the original command that the caller is waiting on won't get completed.

---
In reply to: [412575655](https://github.com/Azure/iotedge/pull/2865#discussion_r412575655) [](ancestors = 412575655)

myagley:
I'm not sure I understand what the issue is?

You should be able to remove an endpoint, by sending the endpoint fsm a "Close" command. Here's the relevant transition in the state machine: https://github.com/Azure/iotedge/blob/master/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/doc/src/failure.gv#L29

This is typically done by the EndpointExecutor when it is closed. For instance, this is the AsyncEndpointExecutor: https://github.com/Azure/iotedge/blob/master/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/AsyncEndpointExecutor.cs#L117

And this is where it is hooked up in the dispatcher (which manages the endpoints it is sending to): https://github.com/Azure/iotedge/blob/master/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/Dispatcher.cs#L249

This remove logic should be sound and we shouldn't need to ever "cancel" a send. The expectation is that the endpoint yields back to the executor if it can't deliver. Are there retries in the endpoint that is preventing this from happening? I'm cautious about changing the state machine behavior.

myagley:
Philip and I talked offline. I think we should just need to close all of the FSMs right before this line:
https://github.com/Azure/iotedge/blob/083381253b6def49bd144f574afe6ac9cad1e931/edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs#L96

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/statemachine/EndpointExecutorFsm.cs
varunpuranik: +1

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/statemachine/EndpointExecutorFsm.cs
varunpuranik: Yes, we probably need to discard all messages

PR #2867, opened 2020-04-22
Mqtt Broker Smoke Tests: Added TestClient wrapper and two basic tests.

vadim-kovalyov:
This is the beginning of the work for adding Mqtt Broker integration tests.
- This PR includes `TestClient` wrapper on `mqtt3::Client`
- Two basic smoke tests.


mqtt/mqtt-broker/tests/common.rs
```diff
@@ -0,0 +1,179 @@
...
+};
+use mqtt_broker::{Authenticator, Authorizer, Broker, BrokerState, Error, Server};
+
+/// A wrapper on the mqtt3::Client to help simplify client event loop management.
+#[derive(Debug)]
+pub struct TestClient {
+    publish_handle: PublishHandle,
+    subscription_handle: UpdateSubscriptionHandle,
+    shutdown_handle: ShutdownHandle,
+    pub events_receiver: Receiver<Event>,
```
dmolokanov, line 26: can we make TestClient to implement Stream? It will mean that we can read messages from a client like this:
`client.next().await` and moreover will give us some stream extensions:
`let all_events = client.collect().await` etc...

implementation itself should be pretty straitforward:
```rust
impl Stream for TestClient {
  fn poll_next(&mut self)  -> {
    self.event_receiver.poll_next();
  }
}

File: mqtt/mqtt-broker/tests/common.rs [hunk 2]
```diff
@@ -0,0 +1,179 @@
...
+
+        let shutdown_handle = client
+            .shutdown_handle()
+            .expect("couldn't get shutdown handle");
+
+        let (mut events_sender, events_receiver) = mpsc::channel(128);
+
+        let task = tokio::spawn(async move {
+            while let Some(event) = client.next().await {
+                let e = event.unwrap();
```
dmolokanov, line 140:
```suggestion
                let event = event.expect("event expected");
```

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -0,0 +1,179 @@
...
+where
+    N: Authenticator + Send + Sync + 'static,
+    Z: Authorizer + Send + Sync + 'static,
+{
+    lazy_static! {
+        static ref PORT: AtomicU32 = AtomicU32::new(5555);
+    }
+
+    let port = PORT.fetch_add(1, Ordering::SeqCst);
+    let server: String = format!("localhost:{}", port);
```
dmolokanov:
```suggestion
    let addr = format!("localhost:{}", port);
```

mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    client.events_receiver.recv().await; //skip connect event
+
+    matches::assert_matches!(
+        client.events_receiver.recv().await,
+        Some(Event::SubscriptionUpdates(_))
+    );
+    matches::assert_matches!(
+        client.events_receiver.recv().await,
+        Some(Event::Publication(ReceivedPublication{payload, ..})) if payload == bytes::Bytes::from("qos 0")
```
dmolokanov, line 96:
```suggestion
        Some(Event::Publication(ReceivedPublication{payload, ..})) if payload == "qos 0".into()
```

mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -0,0 +1,114 @@
...
+    broker_shutdown.send(()).expect("can't stop the broker");
+    broker_task
+        .await
+        .unwrap()
+        .expect("can't wait for the broker");
+}
+
+#[tokio::test]
+async fn basic_pub_sub() {
+    const TOPIC: &str = "topic/A";
```
dmolokanov, line 36: why const here?

mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    let broker = BrokerBuilder::default()
+        .authenticator(|_| Ok(Some(AuthId::Anonymous)))
+        .authorizer(|_| Ok(true))
+        .build();
+
+    let (broker_shutdown, broker_task, address) = common::start_server(broker);
+
+    let mut client = common::TestClientBuilder::default()
```
dmolokanov, line 45: I know in mqtt3 crate author used fully qualified names, but in other code we should utilized usings

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -0,0 +1,179 @@
...
+            self.will,
+            move || {
+                let address = address.clone();
+                let password = password.clone();
+                Box::pin(async move {
+                    let io = tokio::net::TcpStream::connect(&address).await;
+                    io.map(|io| (io, password))
+                })
+            },
+            std::time::Duration::from_secs(1),
```
dmolokanov, line 120:
```suggestion
            Duration::from_secs(1),
```

mqtt/mqtt-broker/tests/common.rs
vadim-kovalyov: It is a good idea, but I'd rather explore that in a separate PR for now.

mqtt/mqtt-broker/tests/smoke.rs
dmolokanov: I mean, what benefit of const vs &'static string here?

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -0,0 +1,179 @@
...
+        Self {
+            address,
+            client_id: None,
+            username: None,
+            password: None,
+            will: None,
+        }
+    }
+
+    pub fn address(mut self, address: T) -> Self {
```
dmolokanov, line 79: Do you still need this method since you have `new(address: T)`?

vadim-kovalyov: No need - removed.

PR #2871, opened 2020-04-23
update build triggers for iiot branch

philipktlin:
update build triggers for iiot branch, which is planned to contain all changes for MQTT broker and Nested Edge.


builds/misc/packages.yaml
```diff
@@ -3,6 +3,7 @@ trigger:
   branches:
     include:
       - master
+      - iiot
```
vadim-kovalyov, line 6: what if changes in `iiot` trigger a build that will not have up-to-date master changes? will it affect any other downstream pipelines/releases?

philipktlin: iiot branch (any branch) should always be complie-able and it should work all the time, even it doesn't have up-to-date changes from master.

PR #2872, opened 2020-04-23
Bring iiot branch up to date

vadim-kovalyov:


dmolokanov:
What do we do with this PR?

vadim-kovalyov:
> @vadim-kovalyov What do we do with this PR?

thanks for the reminder. I think with @philipktlin 's PR we don't need this one.

PR #2876, opened 2020-04-23
Fixed clippy warnings

vadim-kovalyov:
This PR fixes clippy warnings related to wildcard imports.

### DIFF
--- a/mqtt/mqtt-broker/src/auth/authorization.rs
+++ b/mqtt/mqtt-broker/src/auth/authorization.rs
@@ -193,17 +193,19 @@ mod tests {
 
     use matches::assert_matches;
 
+    use mqtt3::{proto, PROTOCOL_LEVEL, PROTOCOL_NAME};
+
     use crate::auth::{Activity, Authorizer, DefaultAuthorizer, Operation};
 
-    fn connect() -> mqtt3::proto::Connect {
-        mqtt3::proto::Connect {
+    fn connect() -> proto::Connect {
+        proto::Connect {
             username: None,
             password: None,
             will: None,
-            client_id: mqtt3::proto::ClientId::ServerGenerated,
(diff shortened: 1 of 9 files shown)


mqtt/mqtt-broker/src/broker.rs
```diff
@@ -1007,20 +1007,24 @@ pub struct NoSessionError;
 
 #[cfg(test)]
 pub(crate) mod tests {
-    use super::*;
+    use super::{
+        mpsc, proto, Activity, BrokerBuilder, BrokerHandle, BrokerState, ClientEvent, ClientId,
```
dmolokanov, line 1011: `mpsc` and `proto` is not a part of parent module. it happens to because parent module include `tokio::mpsc` and `mqtt3::proto` in its scope. 
It should be organized in this manner everywhere in the code instead of `super`
```rust
use tokio::mpsc;
use mqtt3::proto;
use crate::{...}
```

vadim-kovalyov: Make sense, fixed that.

mqtt/mqtt-broker/src/auth/authorization.rs
```diff
@@ -193,14 +193,14 @@ mod tests {
 
     use matches::assert_matches;
 
-    use super::*;
+    use crate::auth::{Activity, Authorizer, DefaultAuthorizer, Operation};
 
     fn connect() -> mqtt3::proto::Connect {
-        proto::Connect {
+        mqtt3::proto::Connect {
```
dmolokanov, line 199: nit: `use mqtt3::proto;`
to be consistent with the rest of changes.

vadim-kovalyov: fixed

PR #2878, opened 2020-04-23
Introduce ConnectionTimeout and MethodTimeout to CloudDirectMethodSender

yophilav:
Introduce ConnectionTimeout and MethodTimeout to CloudDirectMethodSender


test/modules/DirectMethodSender/DirectMethodCloudSender.cs
```diff
@@ -45,7 +45,12 @@ sealed class DirectMethodCloudSender : DirectMethodSenderBase
             ulong directMethodCount,
             CancellationToken none)
         {
-            CloudToDeviceMethod cloudToDeviceMethod = new CloudToDeviceMethod(methodName).SetPayloadJson($"{{ \"Message\": \"Hello\", \"DirectMethodCount\": \"{directMethodCount.ToString()}\" }}");
+            TimeSpan timeouts = Settings.Current.DirectMethodDelay;
+            CloudToDeviceMethod cloudToDeviceMethod = new CloudToDeviceMethod(
+                methodName,
+                Settings.Current.DirectMethodDelay,
```
varunpuranik, line 51: Why are we passing the DirectMethodDelay parameter to ConnectionTimeout and MethodTimeout?

yophilav: Its for this test bug: 
https://msazure.visualstudio.com/One/_sprints/taskboard/IoT-Platform-Edge/One/IoT/2004/2004.2?workitem=6418885

I think that DirectMethodDelay period should be enough for the re-auth to happen and pass the test.

PR #2879, opened 2020-04-24
Fix host and docker cpu stats

lfitchett:
This pr does several things to fix cpu stats for docker and host.
* For host cpu metrics, the measurement object was added as a member of the runtime instead of being newly constructed each time. This allows it to measure cpu impact between measurements instead of only measuring the instant it is polled.
* For docker metrics, previous use is stored in a dictionary. This also allows the cpu use to reflect the entirety of the time between measurements instead of the current use when the measurement occurs.
* This also fixes a math error when converting from ms to ns.

It also changes memory stats to more closely reflect docker stats. Before it reported to reserved RAM, regardless of how much was in use. The `docker stats` command subtracts the memory cache from this value, so the metrics were modified to do the same.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -215,16 +221,23 @@ Option<double> GetCpuUsage(DockerStats module)
...
+            module.Name.ForEach(name =>
+            {
+                currentTotal.ForEach(curr => previousModuleCpu[name] = curr);
+                currentSystem.ForEach(curr => previousSystemCpu[name] = curr);
+            });
+
             return moduleDelta.AndThen(moduleDif => systemDelta.AndThen(systemDif =>
             {
-                if (moduleDif > 0 && systemDif > 0)
+                if (moduleDif >= 0 && systemDif > 0)
```
CindyXing, line 240: What's the reason for adding the "="? I saw the docker golang version calculated the percpuusage. Should we consider it?

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -644,28 +652,30 @@ impl ModuleRuntime for DockerModuleRuntime {
         #[cfg(windows)]
         let uptime: u64 = unsafe { winapi::um::sysinfoapi::GetTickCount64() / 1000 };
 
-        let mut system_info = sysinfo::System::new();
-        system_info.refresh_all();
+        let mut system_resources = self.system_resources.as_ref().lock().unwrap();
```
CindyXing, line 655: Would there be any perf. side impact?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
lfitchett: If it only counts a value when cpu > 0, it skews the average b/c it skips all the 0 values. So  if the readings were 0, 1, 0, 0, 1, 1 the average would be 1 instead of .5.

edgelet/edgelet-docker/src/runtime.rs
lfitchett: There shouldn't be. This is the only consumer of the lock, its only needed b/c the runtime is clone.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -144,6 +144,7 @@ async Task Update()
             // edgelet sets used cpu to -1 on error
             if (systemResources.UsedCpu > 0)
             {
+                Console.WriteLine($"host cpu {systemResources.UsedCpu}");
```
damonbarry, line 147: Is this temporary debug output that can be removed? I ask because you used Console.WriteLine instead of the logger.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -158,11 +159,12 @@ async Task Update()
             }
 
             this.SetModuleStats(systemResources);
+            Console.WriteLine();
```
damonbarry, line 162: Remove?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -173,11 +175,19 @@ void SetModuleStats(SystemResources systemResources)
                 string name = module.Name.OrDefault().Substring(1); // remove '/' from start of name
                 var tags = new string[] { name, EdgeRuntimeModules.Contains(name).ToString() };
 
-                this.GetCpuUsage(module, name).ForEach(usedCpu => this.cpuPercentage.Update(usedCpu, tags));
+                this.GetCpuUsage(module).ForEach(usedCpu =>
+                {
+                    this.cpuPercentage.Update(usedCpu, tags);
+                    Console.WriteLine($"{name} cpu {usedCpu}");
```
damonbarry, line 181: Use a logger?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -195,82 +205,89 @@ void SetModuleStats(SystemResources systemResources)
...
-                }
+                return this.GetCpuLinux(module);
+            }
+            else
+            {
+                return this.GetCpuWindows(module);
+            }
+        }
+
+        // Modeled after https://github.com/docker/cli/blob/master/cli/command/container/stats_helpers.go#L166
```
damonbarry, line 220: This comment will go stale when docker cli code changes. Refer to a specific commit? If you want to keep the ugliness factor down, the URL only needs 7-10 digits of the commit ID...

PR #2881, opened 2020-04-24
[MQTT] Added authentication HTTP endpoint for external processes

vipeller:
MQTT Broker is running as an external process to EdgeHub, but relies on it to authenticate connecting clients. An HTTP endpoint is added to EdgeHub, which can receive different type of credentials (SAS token, certificates) and responses with yes/no, reflecting the authentication result.
The endpoint is designed for internal (in-container) use.

The main file is AuthAgentListener.cs, which handles the HTTP calls and dispatches the queries to the already implemented authentication machinery.


Microsoft.Azure.Devices.Edge.sln
```diff
@@ -219,7 +219,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MetricsValidator", "test\mo
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "EdgeHubRestartTester", "test\modules\EdgeHubRestartTester\EdgeHubRestartTester.csproj", "{DA073067-83EF-4F4C-8E58-51CA7870C1F6}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CloudToDeviceMessageTester", "test\modules\CloudToDeviceMessageTester\CloudToDeviceMessageTester.csproj", "{446F50F3-3E93-454B-8EBF-4830E92DFE5D}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CloudToDeviceMessageTester", "test\modules\CloudToDeviceMessageTester\CloudToDeviceMessageTester.csproj", "{446F50F3-3E93-454B-8EBF-4830E92DFE5D}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Devices.Edge.Hub.AuthAgent", "edge-hub\src\Microsoft.Azure.Devices.Edge.Hub.AuthAgent\Microsoft.Azure.Devices.Edge.Hub.AuthAgent.csproj", "{F040EA42-20FB-453E-8E09-9A1E04E80039}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Azure.Devices.Edge.Hub.AuthAgent.Test", "edge-hub\test\Microsoft.Azure.Devices.Edge.Hub.AuthAgent.Test\Microsoft.Azure.Devices.Edge.Hub.AuthAgent.Test.csproj", "{E5E341ED-571D-473F-B802-0CA885723C67}"
```
varunpuranik, line 226: Nit: Instead of AuthAgent - can we create a project MqttBrokerAdapter ot something like that to include all changes related to communication with the MQTT broker?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
```diff
@@ -0,0 +1,532 @@
...
+            this.listenerTask = null;
+
+            Events.Closed();
+        }
+
+        public void Dispose() => this.CloseAsync(CancellationToken.None).Wait();
+
+        private void Start()
+        {
+            var newListener = new HttpListener();
```
varunpuranik, line 79: HttpListener needs Admin on Windows. Better approach would be to use the existing HttpProtocol head. We can open up port 80 (or any other port we want), and limit traffic on that port to the Authentication controller. Of course, that port would not be exposed out.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
```diff
@@ -0,0 +1,532 @@
...
+    using Microsoft.Azure.Devices.Edge.Hub.Core;
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Identity;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    public class AuthAgentListener : IProtocolHead
+    {
+        private readonly IAuthenticator authenticator;
```
varunpuranik, line 22: Nit: Avoid private for fields (as per convention)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthRequest.cs
```diff
@@ -0,0 +1,23 @@
+namespace Microsoft.Azure.Devices.Edge.Hub.AuthAgent
+{
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Newtonsoft.Json;
+ 
+    public class AuthRequest
+    {
+        [JsonProperty("version", Required = Required.Always)]
```
varunpuranik, line 8: Please make the class immutable

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/UsernameParser.cs
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core.Identity
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public interface IUsernameParser
```
varunpuranik, line 9: Nit: Separate file please

varunpuranik, line 9: Nit: Suggest renaming to - MqttUsernameParser, or IoTHubIdentityMqttUsernameParser

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/AuthModule.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    using Microsoft.Azure.Devices.Edge.Hub.AuthAgent;
+    using Microsoft.Azure.Devices.Edge.Hub.Core;
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Identity;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Configuration;
+
+    class AuthModule : Module
+    {
+        private static readonly string defaultAddress = "http://localhost:7120/authenticate/";
```
varunpuranik, line 16: This should be config that can be passed in to both processes. 
Also, we should probably only include port and sub path (since it will only be available on localhost)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
```diff
@@ -0,0 +1,532 @@
...
+                    break;
+                }
+            }
+        }
+
+        private async Task SendResponseAsync(AuthRequest request, HttpListenerContext context)
+        {
+            try
+            {
+                (var isAuthenticated, var credentials) =
```
varunpuranik, line 148: Nit: Please use specific types where type is not obvious.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
```diff
@@ -0,0 +1,532 @@
...
+                    await GetIdsFromUsername(request)
```
varunpuranik, line 149: nit: this.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
```diff
@@ -0,0 +1,532 @@
...
+                            .FlatMap(ids => GetCredentials(request, ids.deviceId, ids.moduleId, ids.deviceClientType))
+                            .Match(
+                                async creds => (await AuthenticateAsync(creds), Option.Some(creds)),
+                                () => Task.FromResult((false, Option.None<IClientCredentials>())));
+
+                await SendAuthResultAsync(context, isAuthenticated, credentials);
```
varunpuranik, line 155: Nit: this.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
```diff
@@ -0,0 +1,532 @@
...
+
+            await this.listenerTask;
+            this.listenerTask = null;
+
+            Events.Closed();
+        }
+
+        public void Dispose() => this.CloseAsync(CancellationToken.None).Wait();
+
+        private void Start()
```
varunpuranik, line 77: Nit: Avoid private on methods

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
lfitchett: I had to make a very similar change for the metrics endpoint: https://github.com/Azure/iotedge/compare/master...lfitchett:asp.net_metrics

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
```diff
@@ -0,0 +1,532 @@
...
+        private readonly IUsernameParser usernameParser;
+        private readonly IClientCredentialsFactory clientCredentialsFactory;
+        private readonly string listeningAddress;
+
+        private HttpListener listener;
+
+        private Task listenerTask;
+        private bool exitListenerLoop;
+
+        private const int maxInBufferSize = 32 * 1024;
```
lfitchett, line 32: comment here reasoning as well? I dislike seemingly arbitrary consts with no explination.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
lfitchett: if stylecop wasn't forcing it, its not in the project. You'll need to add it as a referance to the csproj. Its a simple copy-paste from the other projects. you might also want to sync with andrew since the dotnet 3.1 changes will need to happen for this new project as well.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/Microsoft.Azure.Devices.Edge.Hub.AuthAgent.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
```
lfitchett, line 6: style cop will need to be added

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
vipeller: added an explanation how the buffer size was picked

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
vipeller: Changed using aspnet. Created a separate host, so other settings (authentications, maybe middlewares) will not interfere. Also the auth endpoint is extremely simple and now it can live its own life.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.AuthAgent/AuthAgentListener.cs
vipeller: added code analysis/checkinbuild to the new project

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/AuthModule.cs
vipeller: it comes from config but has a default value. changed the config structure, so now it is just port/base url

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentEventIds.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    public class AuthAgentEventIds
+    {
+        const int EventIdStart = 7000;
```
lfitchett, line 6: There's still no explanation for this. Why is it 7000?

vipeller: Listed all 'EventIdStart =' in the code and picked a range not being used. It seems like a convention I copied.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/IUsernameParser.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core.Identity
+{
+    public interface IUsernameParser
+    {
+        (string deviceId, string moduleId, string deviceClientType) Parse(string username);
```
varunpuranik, line 6: This is very MQTT Specific, right? Can we move it to the MQTT project? Or to Utils as MqttUsernameParser?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentEventIds.cs
varunpuranik: yup, thats the idea..

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHead.cs
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
```
varunpuranik, line 7: Nit: remove blank lines..

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentRequestHandler.cs
```diff
@@ -0,0 +1,377 @@
...
+
+        public AuthAgentRequestHandler(IAuthenticator authenticator, IUsernameParser usernameParser, IClientCredentialsFactory clientCredentialsFactory, AuthAgentProtocolHeadConfig config)
+        {
+            this.authenticator = Preconditions.CheckNotNull(authenticator, nameof(authenticator));
+            this.usernameParser = Preconditions.CheckNotNull(usernameParser, nameof(usernameParser));
+            this.clientCredentialsFactory = Preconditions.CheckNotNull(clientCredentialsFactory, nameof(clientCredentialsFactory));
+            this.config = Preconditions.CheckNotNull(config, nameof(config));
+        }
+
+        public static IWebHost CreateWebHostBuilder(
```
varunpuranik, line 51: Does this work? 2 WebHosts in the same project?

varunpuranik, line 51: Sorry for the churn but we are trying to move to dotnet 3.1 pretty soon.. maybe we should move this to the new pattern as well.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHead.cs
```diff
@@ -0,0 +1,93 @@
...
+            this.usernameParser = Preconditions.CheckNotNull(usernameParser, nameof(usernameParser));
+            this.clientCredentialsFactory = Preconditions.CheckNotNull(clientCredentialsFactory, nameof(clientCredentialsFactory));
+            this.config = Preconditions.CheckNotNull(config);
+        }
+
+        public async Task StartAsync()
+        {
+            Events.Starting();
+
+            var newHost = AuthAgentRequestHandler.CreateWebHostBuilder(this.authenticator, this.usernameParser, this.clientCredentialsFactory, this.config);
```
varunpuranik, line 39: Nit: Why is this static method on the handler class?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHead.cs
```diff
@@ -0,0 +1,93 @@
...
+    using Microsoft.Extensions.Logging;
+
+    public class AuthAgentProtocolHead : IProtocolHead
+    {
+        readonly IAuthenticator authenticator;
+        readonly IUsernameParser usernameParser;
+        readonly IClientCredentialsFactory clientCredentialsFactory;
+        readonly AuthAgentProtocolHeadConfig config;
+
+        private IWebHost host;
```
varunpuranik, line 23: Nit: remove private.

varunpuranik: Also, to avoid nulls, can we use Option and locking instead?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentRequestHandler.cs
```diff
@@ -0,0 +1,377 @@
...
+
+        public void Configure(IApplicationBuilder app) => app.Use(this.CallFilter).Run(this.Handler);
+
+        async Task Handler(HttpContext context)
+        {
+            try
+            {
+                Events.AuthRequestReceived();
+
+                var requestString = await ReadRequestBodyAsStringAsync(context.Request);
```
varunpuranik, line 76: Nit: please use actual types when the type is not clear..

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthRequest.cs
```diff
@@ -0,0 +1,38 @@
...
+            this.EncodedCertificate = encodedCertificate;
+
+            if (encodedCertificateChain != null)
+            {
+                this.EncodedCertificateChain = encodedCertificateChain.ToArray();
+            }
+        }
+
+        [JsonProperty("version", Required = Required.Always)]
+        public string Version { get; private set; }
```
varunpuranik, line 24: why do we need the private set?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/IUsernameParser.cs
vipeller: Moved the parser to the MQTT projectd, and added reference to Hub.MQTT for Hub.MqttBrokerAdapter

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentRequestHandler.cs
vipeller: Yes, they are independent. I didn't want to merry it with the other one because that has its own middlewares/setup, also this is a separate component and felt cleaner having its own host (also was easier to implement and will be easier to maintain them if they evolve)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHead.cs
vipeller: moved it to the protocol head

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHead.cs
vipeller: changed to Option - however using null/exchange was cleaner I think.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthRequest.cs
vipeller: left there by accident, removed

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentRequestHandler.cs
vipeller: added type (and destroyed the harmony of identification 😒 )

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHead.cs
vipeller: removed

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/IUsernameParser.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Mqtt
+{
+    public interface IUsernameParser
```
vadim-kovalyov, line 4: nit: interface comments?

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Mqtt.Test/UsernameParserTest.cs
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core.Test
+{
+    using System.Collections.Generic;
+    using Microsoft.Azure.Devices.Edge.Hub.Mqtt;
+    using Microsoft.Azure.Devices.Edge.Util.Test.Common;
+    using Xunit;
+
+    [Unit]
+    public class UsernameParserTest
```
vadim-kovalyov, line 10: nit: to be more precise, this file and tests should be named `MqttUsernameParserTest`

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/MqttUsernameParser.cs
```diff
@@ -0,0 +1,129 @@
...
+    using System.Linq;
+    using Microsoft.Azure.Devices.Edge.Hub.Core;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public class MqttUsernameParser : IUsernameParser
+    {
+        const string ApiVersionKey = "api-version";
+        const string DeviceClientTypeKey = "DeviceClientType";
+
+        public (string deviceId, string moduleId, string deviceClientType) Parse(string username)
```
vadim-kovalyov, line 15: nit: it might be out of scope of this change, but it would be nice to change return type to Option<> and get rid of exceptions in this method :)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentRequestHandler.cs
```diff
@@ -0,0 +1,355 @@
...
+
+            var mediaType = new MediaTypeHeaderValue("application/json");
+            mediaType.Encoding = Encoding.UTF8;
+            context.Response.ContentType = mediaType.ToString();
+            context.Response.ContentLength = buffer.Length;
+
+            await context.Response.Body.WriteAsync(buffer, 0, buffer.Length);
+        }
+
+        static async Task<Option<string>> ReadRequestBodyAsStringAsync(HttpRequest request)
```
vadim-kovalyov, line 204: can you please clarify why you need all this machinery? Isn't there already a way in asp.net core to read body buffered?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/IUsernameParser.cs
```diff
@@ -0,0 +1,8 @@
...
+    {
+        (string deviceId, string moduleId, string deviceClientType) Parse(string username);
```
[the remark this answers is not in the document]
vipeller: It used to be an internal method I factored out and wrapped into a class/interface. I can introduce a class to hold the components

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/AuthModule.cs
```diff
@@ -0,0 +1,49 @@
...
+    using Autofac;
+
+    using Microsoft.Azure.Devices.Edge.Hub.Core;
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Identity;
+    using Microsoft.Azure.Devices.Edge.Hub.Mqtt;
+    using Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Configuration;
+
+    class AuthModule : Module
```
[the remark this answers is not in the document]
vipeller: we usually don't have comment for classes (and in general)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/MqttUsernameParser.cs
vipeller: yes, but it was an already existing code I moved out for reuse.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentRequestHandler.cs
vipeller: The story of this handler that first it was built around a http listener, but then it turned out that it has permission problems on windows (needs admin), so I turned the same code into kestrel based, but kept the manual work which was already implemented. I can turn it into an "official" webapi with controllers and routing, but I felt unnecessary

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/IUsernameParser.cs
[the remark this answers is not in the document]
vipeller: I added the changes

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Mqtt.Test/UsernameParserTest.cs
vipeller: changed the class name

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentRequestHandler.cs
vipeller: changed it to controller based

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentController.cs
```diff
@@ -0,0 +1,228 @@
...
+    using Microsoft.Azure.Devices.Edge.Hub.Mqtt;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public class AuthAgentController : Controller
+    {
+        const int Authenticated = 200;
+        const int Unauthenticated = 403;
+
+        static readonly string apiVersion = "2020-04-20";
```
varunpuranik, line 21: Nit; why not Const? And should we move it to Constants?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentController.cs
```diff
@@ -0,0 +1,228 @@
...
+        [Produces("application/json")]
+        public async Task<JsonResult> HandleAsync([FromBody] AuthRequest request)
+        {
+            if (request == null)
+            {
+                Events.ErrorDecodingPayload();
+                return this.Json(GetErrorResult());
+            }
+
+            if (!string.Equals(request.Version, apiVersion))
```
varunpuranik, line 46: This is okay for now. But we need to think through how we want to manage the API version in the EdgeHub

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentController.cs
```diff
@@ -0,0 +1,228 @@
...
+            {
+                Events.ErrorBadVersion(request.Version ?? "(none)");
+                return this.Json(GetErrorResult());
+            }
+
+            try
+            {
+                var isAuthenticated = default(bool);
```
varunpuranik, line 54: Nit: Why do you need to initialize here?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHead.cs
```diff
@@ -0,0 +1,135 @@
...
+            {
+                if (this.host.HasValue)
+                {
+                    Events.StartedWhenAlreadyRunning();
+                    throw new InvalidOperationException("Cannot start AuthAgentProtocolHead twice");
+                }
+                else
+                {
+                    this.host = Option.Some(
+                                    CreateWebHostBuilder(
```
varunpuranik, line 55: Interesting that this works. So we essentially have 2 asp.net web servers in the same process. Any concerns about it conflicting or something like that?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHeadConfig.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public class AuthAgentProtocolHeadConfig
+    {
+        public AuthAgentProtocolHeadConfig(int port, string baseUrl)
+        {
+            this.Port = port;
+            this.BaseUrl = baseUrl;
+        }
+
+        public int Port { get; private set; }
```
varunpuranik, line 12: Nit: remove the private sets

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHeadConfig.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    public class AuthAgentProtocolHeadConfig
+    {
+        public AuthAgentProtocolHeadConfig(int port, string baseUrl)
+        {
+            this.Port = port;
```
varunpuranik, line 8: Precondition.Check
Same everywhere

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentStartup.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        public void Configure(IApplicationBuilder app)
+        {
+            app.UseMvc(routes =>
+            {
+                routes.MapRoute(
+                    "authenticate",
+                    this.config.BaseUrl,
+                    defaults: new { controller = "AuthAgent", action = "HandleAsync" });
```
varunpuranik, line 22: Should we use routing attribute instead?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentController.cs
vipeller: moved into a separate class

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentController.cs
vipeller: that is just to enable using 'var', so the variable names are aligned. also I didn't want to move the types into the deconstruction below, so it is easier to see what comes back. I thought it more readable because the variable names are not obscured by type definition, but the type is still there is one is interested

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentProtocolHead.cs
vipeller: my only concern was that they find each others controllers, but the new controller doesn't have routing attribute and the 'old' would use that for config, and the new controller is set manually in the new server, so that doesn't search for anything else.
I figured it would be both harder to combine the two (the existing has specialized middlewares for example)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/AuthAgentStartup.cs
vipeller: I wanted to be sure that the authentication doesn't handle any other requests (=to pick up classes with appropriate attributes automatically), also it uses configuration dependent parts (the baseurl) which cannot be done with routing attributes.

PR #2883, opened 2020-04-25
Improvements to the mqtt perf tests

vadim-kovalyov:
This will include:
- Change schedule from 4pm to 11pm PST/
- Remove conditional logic from Setup task
- Remove conditional logic from Cleanup task


vadim-kovalyov:
Did not work....

PR #2885, opened 2020-04-27
[mqtt] Add $edgehub topic notifier

lfitchett:
This sends retained messages to a sys topic whenever a connection/session is added or dropped, and when a subscription is added/removed.

Currently errors are logged to the console, since a notification error shouldn't stop a subscription or connection.

remaining todo:
* ~~throttling (if a client subscribes to 50 topics in the space of 10ms, we shouldn't notify 50 times)~~ We decided throttling was not necessary


mqtt/mqtt-broker/src/broker.rs
```diff
@@ -797,6 +814,99 @@ where
...
+            .join(r"\u{0000}");
+
+        let publication = proto::Publication {
+            topic_name: "$sys/connected".to_owned(),
+            qos: proto::QoS::AtMostOnce, //no ack
+            retain: true,
+            payload: Bytes::from(connected),
+        };
+
+        self.publish_all(publication).await.unwrap();
```
vadim-kovalyov, line 883: are you planning to implement todos (error handling specifically) in the same PR? if so, maybe move it to draft for now?

I think at least we need to write to stderr, but @dmolokanov can advice more.

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -797,6 +814,99 @@ where
...
+    }
+
+    async fn notify_disconnect(&mut self, client_id: &ClientId) {
+        if let None = self.sessions.get(client_id) { // If transient session is closed, remove its subscriptions
+            let publication = proto::Publication {
+                topic_name: format!("$sys/subscriptions/{}", client_id),
+                qos: proto::QoS::AtMostOnce, //no ack
+                retain: true,
+                payload: Bytes::from("".to_owned()),
```
vadim-kovalyov, line 854: nit:
```suggestion
                payload: "".into(),
```

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -355,6 +362,9 @@ where
             debug!("no session for {}", client_id);
         }
         debug!("drop connection handled.");
+
+        self.notify_disconnect(&client_id).await;
```
vadim-kovalyov, line 366: do we need to do this only if session was found and disconnected (ln:354)?

mqtt/mqtt-broker/src/broker.rs
lfitchett: Is that better? I think `.to_owned()` is more clear

mqtt/mqtt-broker/src/broker.rs
lfitchett: probably, but I erred on the side of caution. I'd rather send extra messages than miss one.

mqtt/mqtt-broker/src/broker.rs
lfitchett: Yea, I want to do it in this pr, its a simple change I just wasn't sure the approach.

mqtt/mqtt-broker/src/broker.rs
vadim-kovalyov: In my limited understanding they both will do the same thing in this case, but `.to_owned()` has a bit different purpose in general. If you need to convert one type into another (like here &str to Bytes) I would use `.into()`. Not a big deal, just nitpicking :)

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -1987,4 +2315,82 @@ pub(crate) mod tests {
...
+            .send(Message::Client(client_id.into(), event))
+            .await
+            .unwrap();
+    }
+
+    async fn send_subscribe(
+        handle: &mut BrokerHandle,
+        rx: &mut Receiver<Message>,
+        client_id: ClientId,
+        topics: &[&str],
```
dmolokanov, line 2332: I noticed that you create a static array in the test case and then pass it as a reference, while you can move and consume in some mehtods. It will reduce manipulations with pointers.

mqtt/mqtt-broker/src/broker.rs
lfitchett: oh, I forgot bytes impls From<&'static str>. I was using .to_owned to convert to a String, then From to turn the string into bytes. In this case it is better just to use .into()

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -797,6 +811,98 @@ where
...
+            .and_then(|session| match session {
+                // get session state
+                Session::Transient(connected) => Some(connected.state()),
+                Session::Persistent(connected) => Some(connected.state()),
+                Session::Offline(offline) => Some(offline.state()),
+                _ => None,
+            })
+            .map(|state| state.subscriptions().keys()) // get topic filters
+            .map(|topics| {
+                topics
```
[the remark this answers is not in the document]
lfitchett: I was going off of https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0d9e93ec8ddfcbf86f0294c3f011eedf from this thread https://stackoverflow.com/questions/36941851/whats-an-idiomatic-way-to-print-an-iterator-separated-by-spaces-in-rust

mqtt/mqtt-broker/src/broker.rs
lfitchett: Its a test, and passing by reference let me avoid making the functions take generics

mqtt/mqtt-broker/src/broker.rs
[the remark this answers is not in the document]
dmolokanov: Interesting... collect approach works faster!

mqtt/mqtt-broker/src/broker.rs
dmolokanov: Generics? I was talking about creating strings and move them instead of borrowing. Maybe I didn't understand your question

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -354,12 +355,19 @@ where
...
+        self.send_sys_message(get_sys_message(&StateChange::AddConnection(&self.sessions)))?;
+
         Ok(())
     }
 
     fn process_disconnect(&mut self, client_id: &ClientId) -> Result<(), Error> {
         debug!("handling disconnect...");
-        if let Some(mut session) = self.close_session(client_id) {
+        if let Some(mut session) = self.close_session(&client_id) {
+            self.send_sys_message(get_sys_message(&StateChange::RemoveConnection(
```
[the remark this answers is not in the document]
dmolokanov: actually this code can move to `self.close_session()` method and we dont need to duplicate notification everywhere

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -445,11 +461,14 @@ where
...
             .collect::<Vec<proto::Publication>>();
 
-        if let Some(session) = self.sessions.get_mut(client_id) {
+        if let Some(session) = self.sessions.get_mut(&client_id) {
             for mut publication in publications {
                 publication.retain = true;
                 publish_to(&self.authorizer, session, &publication)?;
             }
+            
+            let sys_message = get_sys_message(&StateChange::Subscriptions(&session));
```
dmolokanov, line 470:
```suggestion
            let change = StateChange::from_subscriptions(client_id, &session);
            self.publish_change(change)?;
```

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -464,8 +483,13 @@ where
     ) -> Result<(), Error> {
         match self.get_session_mut(client_id) {
             Ok(session) => {
-                let unsuback = session.unsubscribe(unsubscribe)?;
-                session.send(ClientEvent::UnsubAck(unsuback))
+                let unsuback = session.unsubscribe(&unsubscribe)?;
+                session.send(ClientEvent::UnsubAck(unsuback))?;
+
+                let sys_message = get_sys_message(&StateChange::Subscriptions(&session));
```
dmolokanov, line 489:
```suggestion
            let change = StateChange::from_subscriptions(client_id, &session);
            self.publish_change(change)?;
```

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -827,6 +851,20 @@ where
 
         Ok(())
     }
+
+    fn send_sys_message(&mut self, message: SysMessage) -> Result<(), Error> {
```
dmolokanov, line 855:
```rust 
fn publish_change(&mut self, change: StateChange) -> Result<(), Error> {
  let publication = state.into();
  self.publish_all(publication)
}
```
instead

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -1981,6 +2019,222 @@ pub(crate) mod tests {
         assert_matches!(sub_rx.try_recv(), Err(TryRecvError::Empty))
     }
 
+    #[tokio::test]
+    async fn test_notify_single_connection() {
```
dmolokanov, line 2023: some tests checking sessions to publication data gathering move to `sys_topic` module

mqtt/mqtt-broker/src/broker.rs
lfitchett: ah, I thought you meant pass the array by value, not the strings in the array. Since its a test, I'd rather have it cleaner than more efficient, and its much cleaner to pass in `&["foo", "bar"]` than `&["foo".to_owned(), "bar".to_owned()]`

mqtt/mqtt-broker/src/lib.rs
```diff
@@ -39,6 +40,7 @@ pub use crate::persist::{
 pub use crate::server::Server;
 pub use crate::session::SessionState;
 pub use crate::snapshot::{Snapshotter, StateSnapshotHandle};
+pub use crate::sys_topic::{get_sys_message, StateChange, SysMessage};
```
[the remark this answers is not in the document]
lfitchett: looking at the functions in this file, there isn't much to test. The ones in the broker are able to test that it actually sends a message when something subscribes. All this file really does is convert an array of sessions into an array of strings.

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,119 @@
+use std::collections::HashMap;
+
+use bytes::Bytes;
+use mqtt3::proto;
+
+use crate::session::Session;
+use crate::ClientId;
+
+pub enum StateChange<'a> {
```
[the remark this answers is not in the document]
lfitchett: I know this isn't exactly what you originally intended, but I think it makes a lot of sense as a conceptual breakdown of roles. The broker does 2 things. Sends this a state change, and publishes whatever messages come back. The sys_topics file takes changes, and creates the messages needed to notify. 

I don't think returning the state change is good, because some events can trigger multiple messages, which now the broker would have to know about. In addition, my way makes it easy to add more sys topics for things like number of messages sent, b/c it allows 1 event to trigger multiple sys messages.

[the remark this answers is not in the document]
dmolokanov: I don't think it is 2 things, it's only one: a broker which publishes state changes. Let's concentrate only on this task for now. And later we can make more agile solution if required.

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,102 @@
...
+                topic_name: format!("$edgehub/subscriptions/{}", client_id),
+                qos: QOS,
+                retain: true,
+                payload: get_message_body(subscriptions.into_iter()).into(),
+            },
+            StateChange::Connections(connections) => proto::Publication {
+                topic_name: "$edgehub/connected".to_owned(),
+                qos: QOS,
+                retain: true,
+                payload: get_message_body(connections.iter().map(|c| c.as_str())).into(),
```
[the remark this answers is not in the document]
dmolokanov:
```suggestion
                payload: serde_json::to_string(connections)?.into(),
```

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl<'a> StateChange<'a> {
+    pub fn get_subscriptions(client_id: &'a ClientId, session: &'a Session) -> Self {
+        Self::Subscriptions(
+            client_id,
+            session
+                .subscriptions()
+                .unwrap()
+                .keys()
+                .map(String::as_str)
```
dmolokanov, line 24: Why `as_str` here?

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,102 @@
...
+    Sessions(Vec<&'a ClientId>),
+}
+
+impl<'a> StateChange<'a> {
+    pub fn get_subscriptions(client_id: &'a ClientId, session: &'a Session) -> Self {
+        Self::Subscriptions(
+            client_id,
+            session
+                .subscriptions()
+                .unwrap()
```
dmolokanov, line 22: Production should not contain `unwrap`s

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,102 @@
...
+
+pub enum StateChange<'a> {
+    Subscriptions(&'a ClientId, Vec<&'a str>),
+    Connections(Vec<&'a ClientId>),
+    Sessions(Vec<&'a ClientId>),
+}
+
+impl<'a> StateChange<'a> {
+    pub fn get_subscriptions(client_id: &'a ClientId, session: &'a Session) -> Self {
+        Self::Subscriptions(
```
dmolokanov, line 18: It can be done in a more readable way:
```rust
let subscriptions = ...;
Self::Subscriptions(client_id, subscriptions)
```

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,102 @@
...
+            session
+                .subscriptions()
+                .unwrap()
+                .keys()
+                .map(String::as_str)
+                .collect(),
+        )
+    }
+
+    pub fn clear_subscriptions(client_id: &'a ClientId) -> Self {
```
dmolokanov, line 29: I think you can combine this case with previous case by making session Optional.

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use mqtt3::proto;
+use tracing::error;
+
+use crate::session::Session;
+use crate::ClientId;
+
+pub enum StateChange<'a> {
+    Subscriptions(&'a ClientId, Vec<&'a str>),
+    Connections(Vec<&'a ClientId>),
```
dmolokanov, line 12: Could you please clarify once again why store borrowing references instead of owning values? In the end we still make a copy when deserializing.

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,119 @@
+use std::collections::HashMap;
```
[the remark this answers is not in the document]
dmolokanov: This comment is still valid: here you should check that
1.  for incoming data you got correct enum Variant.
2. you filter sessions online/offline etc correctly
3. you convert each enum variant into expected publish packet

only things you need to check with broker unit tests is that for given state change you got an expected number of packets without checking all publication fields

mqtt/mqtt-broker/src/sys_topic.rs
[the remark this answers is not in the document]
lfitchett: Talked in person, any more comments should be its own thread

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+fn get_message_body<'a, P>(payload: P) -> String
+where
+    P: IntoIterator<Item = &'a str>,
+{
+    let payload: Vec<&str> = Vec::from_iter(payload);
+
+    if payload.is_empty() {
+        "".to_owned()
```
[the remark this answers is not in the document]
lfitchett: This was used for the clear subscriptions on no sub, but I guess I should change it to be "[]" if no subscriptions, and "" if no session

mqtt/mqtt-broker/src/sys_topic.rs
lfitchett: turn &String to &str

mqtt/mqtt-broker/src/sys_topic.rs
lfitchett: Vec<String> will be copied twice, once when moving into vec, and again when converting to json. Vec<&str> is only copied when converted to json

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,389 @@
...
+    use std::str::FromStr;
+    #[test]
+    fn test_subscriptions() {
+        let expected_id: ClientId = "Session".into();
+
+        // Session with no subscriptions
+        let session = make_session(expected_id.as_str(), Vec::<&str>::new(), true);
+        let no_subs = StateChange::new_subscription(&expected_id, &session);
+
+        if let StateChange::Subscriptions(stored_id, stored_subs) = &no_subs {
```
dmolokanov, line 121: isn't compiler smart enough to borrow here automatically?
```suggestion
        if let StateChange::Subscriptions(stored_id, stored_subs) = no_subs {
```

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,389 @@
...
+            assert_eq!(&&expected_id, stored_id);
+            assert_eq!(&Vec::<&str>::new(), stored_subs.as_ref().unwrap());
```
dmolokanov, line 123: Compiler definetely can infer type. And asserion looks better now, doesn't it?
```suggestion
            assert_eq!(Ok(Vec::new()), stored_subs);
```

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,389 @@
...
+    use std::collections::{HashMap, VecDeque};
+    use std::convert::TryInto;
+    use std::iter::FromIterator;
+    use std::str::FromStr;
+    #[test]
+    fn test_subscriptions() {
+        let expected_id: ClientId = "Session".into();
+
+        // Session with no subscriptions
+        let session = make_session(expected_id.as_str(), Vec::<&str>::new(), true);
```
dmolokanov, line 118:
```suggestion
        let session = make_session(expected_id.as_str(), Vec::new(), true);
```

dmolokanov: Lets put type only in places where it is really required. It doesn't add clarity but makes code harder to read

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -787,16 +791,13 @@ where
...
+        Ok(match self.sessions.remove(client_id) {
             Some(Session::Transient(connected)) => {
                 info!("closing transient session for {}", client_id);
-                self.publish_all(StateChange::get_connections(&self.sessions).into())
-                    .unwrap();
-                self.publish_all(StateChange::get_sessions(&self.sessions).into())
-                    .unwrap();
-                self.publish_all(StateChange::clear_subscriptions(client_id).into())
-                    .unwrap();
+                self.publish_all(StateChange::new_connection(&self.sessions).try_into()?)?;
```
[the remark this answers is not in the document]
lfitchett: It does. The connection is stored in the session. Transient and Persistent connections have the connection inside them.

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -1086,6 +1086,13 @@ pub enum SessionError {
     PacketIdentifiersExhausted,
     ProtocolViolation(Session),
     DuplicateSession(Session, proto::ConnAck),
+    Other(Error),
```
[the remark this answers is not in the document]
lfitchett: `open_session` uses it. When a new session is opened, it calls 
```
let subscription_change = StateChange::new_subscription(&client_id, &new_session).try_into()?;
...
self.publish_all(StateChange::new_session(&self.sessions).try_into()?)?;
self.publish_all(subscription_change)?;
```

mqtt/mqtt-broker/src/sys_topic.rs
lfitchett: I changed the enum variant to hold an option. I think it makes sense to have 2 different creation functions, since they are conceptually doing different things (updating subscriptions vs. removing the message), but I combined the enum variants (before there was a clear subscriptions variant)

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,389 @@
...
+            ],
+        );
+    }
+
+    fn make_session<I, S>(id: &str, subscriptions: I, online: bool) -> Session
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        let subscriptions = HashMap::from_iter(subscriptions.into_iter().map(|s| {
```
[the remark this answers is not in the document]
lfitchett: I just thought collect only worked for vecs. I'll change it, .collect looks better.

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -787,16 +791,13 @@ where
         }
     }
 
-    fn close_session(&mut self, client_id: &ClientId) -> Option<Session> {
-        match self.sessions.remove(client_id) {
+    fn close_session(&mut self, client_id: &ClientId) -> Result<Option<Session>, Error> {
+        Ok(match self.sessions.remove(client_id) {
```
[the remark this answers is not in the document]
lfitchett: I put the result of the match into a new_session var, then return Ok(new_session). This makes it more clear I think.

mqtt/mqtt-broker/src/sys_topic.rs
lfitchett:
```
error[E0382]: use of moved value: `many_subs`                                                     
   --> mqtt-broker\src\sys_topic.rs:154:43
    |
147 |         if let StateChange::Subscriptions(stored_id, stored_subs) = many_subs {
    |                                                      ----------- value moved here
...
154 |         let message: proto::Publication = many_subs.try_into().unwrap();
    |                                           ^^^^^^^^^ value used here after partial move      
    |
    = note: move occurs because value has type `std::option::Option<std::vec::Vec<&str>>`, which does not implement the `Copy` trait
```

mqtt/mqtt-broker/src/sys_topic.rs
lfitchett:
```
error[E0282]: type annotations needed
   --> mqtt-broker\src\sys_topic.rs:116:23
    |
116 |         let session = make_session(expected_id.as_str(), Vec::new(), true);
    |                       ^^^^^^^^^^^^ cannot infer type for type parameter `S` declared on the 
function `make_session`
```

lfitchett: This is b/c make_session takes an Iterator of AsRef<str>. I wanted it to take either &str or Strings

mqtt/mqtt-broker/src/broker.rs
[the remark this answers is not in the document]
dmolokanov: But anyone else as a reader when it goes further they still need to remember about this `Ok` wrapping on the top of the expression. It increases cyclomatic complexity and in general should be avoided. https://en.wikibooks.org/wiki/Computer_Programming/Coding_Style/Minimize_nesting

mqtt/mqtt-broker/src/sys_topic.rs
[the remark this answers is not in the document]
dmolokanov: `collect` is very powerful concept for accumulating collections. It works with all standard collections and even can help dealing with results and options.

mqtt/mqtt-broker/src/sys_topic.rs
dmolokanov: Yeah, I think you may want to use `Into<String>` in the make_session method instead.

mqtt/mqtt-broker/src/sys_topic.rs
dmolokanov: Ah, I see your confusion here: 
`StateChange` factory methods have nothing to do with with actual state changes happened in the broker.
`new_subscription` method doesn't reflect that new subscription was added, it just creates a StateChange variant with corrent data. So it's better to call method `new_subscriptions_change(client_id, Session)`
1. new subscription requested: `StateChange::new_subscriptions_change(client_id, Some(session))`
2. subscription removed: `StateChange::new_subscriptions_change(client_id, Some(session))`
3. session closed: `StateChange::new_subscription_change(client_id, None)`
etc..

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -1,102 +1,389 @@
...
-                qos: QOS,
-                retain: true,
-                payload: get_message_body(subscriptions.into_iter()).into(),
-            },
+    fn try_from(state: StateChange<'a>) -> Result<Self, Error> {
+        Ok(match state {
+            StateChange::Subscriptions(client_id, subscriptions) => {
+                let payload = if let Some(subscriptions) = subscriptions {
+                    serde_json::to_string(&subscriptions)
+                        .map_err(|_| Error::NotifyError)?
```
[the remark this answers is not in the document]
dmolokanov: What do you think about this?

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,399 @@
...
+    use std::convert::TryInto;
+    use std::str::FromStr;
+
+    use super::{StateChange, STATE_CHANGE_QOS};
+    use crate::broker::tests::{connection_handle, is_notify_equal, persistent_connect};
+    use crate::session::{Session, SessionState};
+    use crate::subscription::{Subscription, TopicFilter};
+    use crate::{AuthId, ClientId, ConnReq};
+    use mqtt3::proto;
+    #[test]
```
vadim-kovalyov, line 104: super nit: new line

mqtt/mqtt-broker/src/sys_topic.rs
dmolokanov: what about this?

lfitchett: I must not have refreshed github, b/c I didn't see this. I'll refactor the funcs

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,399 @@
...
+        let message: proto::Publication = many_connections.try_into().unwrap();
+        matches_connection_publication(message, &["Session 2", "Session 4", "Session 6"]);
+    }
+
+    #[test]
+    fn test_sessions() {
+        // No sessions
+        let sessions: HashMap<ClientId, Session> = HashMap::new();
+        let no_sessions = StateChange::new_session(&sessions);
+        if let StateChange::Connections(no_sessions) = &no_sessions {
```
vadim-kovalyov, line 243: Here and further in this test it should be `StateChange::Sessions(no_sessions)`. I would recommend try to avoid any `if` in the tests (otherwise who will test the correctness of the test? :) ). Instead, use assert_matches! if possible.

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -0,0 +1,399 @@
...
+                qos: STATE_CHANGE_QOS,
+                retain: true,
+                payload: serde_json::to_string(&sessions)?.into(),
+            },
+        })
+    }
+}
+
+#[cfg(test)]
+pub(crate) mod tests {
```
vadim-kovalyov, line 93: I would suggest going forward to make **unit tests** more granular. Here in every test you are testing both `new_***` and `try_into` together. Instead, I would split it into separate tests. This is not a big deal, and this is fine as well, but just as a recommendation going forward.

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -1969,6 +2001,227 @@ pub(crate) mod tests {
         assert_matches!(sub_rx.try_recv(), Err(TryRecvError::Empty))
     }
 
+    #[tokio::test]
```
vadim-kovalyov, line 2004: Just as a general comment, the more I look into the broker tests, the more I believe that tests that start the broker in a separate thread should be moved to integration tests. Not right now, but probably later I will move all of them.

mqtt/mqtt-broker/src/sys_topic.rs
lfitchett: that works for this test, but for similar tests further down I need to sort the returned vec, which a pattern match cannot do (afaik)

mqtt/mqtt-broker/src/sys_topic.rs
```diff
@@ -61,13 +61,10 @@ impl<'a> TryFrom<StateChange<'a>> for proto::Publication {
     fn try_from(state: StateChange<'a>) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
lfitchett: I made StateChange a variant of Error rather than its own enum.

mqtt/mqtt-broker/src/sys_topic.rs
lfitchett: cargo fmt removes it for some reason

mqtt/mqtt-broker/src/broker.rs
lfitchett: I agree, the tests in broker.rs are integration tests, however I think they access some internals and it might not be possible to move them

mqtt/mqtt-broker/src/sys_topic.rs
vadim-kovalyov: Sure, `assert_matches!` might not work everywhere, use `assert_eq!` then. And don't use `if`.
This line should look like below. Please note that `StateChange::new_session()` returns `StateChange::Sessions`, and not `StateChange::Connections`.
```
        assert_eq!(
            no_sessions,
            StateChange::Sessions(Vec::<&ClientId>::new())
        );
```

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -622,11 +629,14 @@ where
...
-                        let (state, events) = offline
-                            .into_online()
-                            .map_err(|_| SessionError::PacketIdentifiersExhausted)?;
-                        let new_session = Session::new_persistent(auth_id, connreq, state);
-                        (new_session, events, true)
+                        if let Ok((state, events)) = offline.into_online() {
+                            let new_session = Session::new_persistent(auth_id, connreq, state);
+                            (new_session, events, true)
+                        } else {
+                            panic!(
```
dmolokanov, line 636: there was no panic before. can we possibly revert this change back?

mqtt/mqtt-broker/src/state_change.rs
```diff
@@ -0,0 +1,413 @@
...
+                qos: STATE_CHANGE_QOS,
+                retain: true,
+                payload: serde_json::to_string(&sessions)?.into(),
+            },
+        })
+    }
+}
+
+#[cfg(test)]
+pub(crate) mod tests {
```
dmolokanov, line 91:
```suggestion
mod tests {
```

mqtt/mqtt-broker/src/error.rs
```diff
@@ -54,6 +55,9 @@ pub enum Error {
 
     #[error("Unable to start broker.")]
     InitializeBroker(#[from] InitializeBrokerError),
+
+    #[error("An error occurred when constructing state change: {0}")]
+    StateChange(#[from] JsonError),
```
dmolokanov, line 60: I'd rather keep it with a full path for error variants.

mqtt/mqtt-broker/src/state_change.rs
```diff
@@ -0,0 +1,413 @@
...
+impl<'a> TryFrom<StateChange<'a>> for proto::Publication {
+    type Error = Error;
+
+    fn try_from(state: StateChange<'a>) -> Result<Self, Error> {
+        Ok(match state {
+            StateChange::Subscriptions(client_id, subscriptions) => {
+                let payload = subscriptions
+                    .map(|subscriptions| serde_json::to_string(&subscriptions))
+                    .transpose()?
+                    .map_or_else(bytes::Bytes::new, |json| json.into());
```
dmolokanov, line 65:
```suggestion
                    .map_or_default(|json| json.into());
```

mqtt/mqtt-broker/src/state_change.rs
```diff
@@ -0,0 +1,413 @@
...
+    use std::collections::{HashMap, VecDeque};
+    use std::convert::TryInto;
+    use std::str::FromStr;
+
+    use super::{StateChange, STATE_CHANGE_QOS};
+    use crate::broker::tests::{connection_handle, is_notify_equal, persistent_connect};
+    use crate::session::{Session, SessionState};
+    use crate::subscription::{Subscription, TopicFilter};
+    use crate::{AuthId, ClientId, ConnReq};
+    use mqtt3::proto;
```
dmolokanov, line 101: missed new line

mqtt/mqtt-broker/src/state_change.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+}
+
+#[cfg(test)]
+pub(crate) mod tests {
+    use std::collections::{HashMap, VecDeque};
+    use std::convert::TryInto;
+    use std::str::FromStr;
+
+    use super::{StateChange, STATE_CHANGE_QOS};
```
dmolokanov, line 96:
```suggestion
    use crate::state_change::{StateChange, STATE_CHANGE_QOS};
```

mqtt/mqtt-broker/src/state_change.rs
```diff
@@ -0,0 +1,413 @@
...
+            },
+        })
+    }
+}
+
+#[cfg(test)]
+pub(crate) mod tests {
+    use std::collections::{HashMap, VecDeque};
+    use std::convert::TryInto;
+    use std::str::FromStr;
```
dmolokanov, line 94: Interesting I thought it is a part prelude...

mqtt/mqtt-broker/src/state_change.rs
dmolokanov, line 101: Can we make usings in 4 groups as usual?

mqtt/mqtt-broker/src/broker.rs
lfitchett: there was, it just moved

mqtt/mqtt-broker/src/state_change.rs
lfitchett: cargo fmt removes it for some reason

lfitchett: weird, it removes it when there's 1 newline, but leaves a new line when I add 2...

PR #2888, opened 2020-04-27
[mqtt]: Retained and Will messages integration tests

vadim-kovalyov:
This PR adds:
- Retained messages test
- Will message test
- TestClient implements Stream

### DIFF
--- a/mqtt/mqtt-broker/tests/common.rs
+++ b/mqtt/mqtt-broker/tests/common.rs
@@ -1,16 +1,20 @@
 use std::{
+    pin::Pin,
     sync::atomic::{AtomicU32, Ordering},
+    task::{Context, Poll},
     time::Duration,
 };
 
 use futures_util::{FutureExt, StreamExt};
 use lazy_static::lazy_static;
-use tokio::net::ToSocketAddrs;
-use tokio::sync::{
-    mpsc::{self, Receiver},
-    oneshot::{self, Sender},
+use tokio::{
+    net::ToSocketAddrs,
+    sync::{
+        mpsc::{self, UnboundedReceiver},
+        oneshot::{self, Sender},
+    },
+    task::{JoinError, JoinHandle},
 };
-use tokio::task::{JoinError, JoinHandle};
 
 use mqtt3::{
     proto::{Publication, SubscribeTo},
@@ -25,7 +29,7 @@ pub struct TestClient {
     publish_handle: PublishHandle,
     subscription_handle: UpdateSubscriptionHandle,
     shutdown_handle: ShutdownHandle,
-    pub events_receiver: Receiver<Event>,
+    events_receiver: UnboundedReceiver<Event>,
     task: JoinHandle<()>,
 }
 
@@ -45,8 +49,23 @@ impl TestClient {
         self.shutdown_handle.shutdown().await
     }
 
+    pub fn shutdown_handle(&mut self) -> ShutdownHandle {
+        self.shutdown_handle.clone()
+    }
+
     pub async fn join(self) -> Result<(), JoinError> {
         self.task.await
+    }
+}
+
+impl tokio::stream::Stream for TestClient
+where
+    Self: Unpin,
+{
+    type Item = Event;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.events_receiver.poll_recv(cx)
     }
 }
 
@@ -128,14 +147,13 @@ where
             .shutdown_handle()
             .expect("couldn't get shutdown handle");
 
-        let (mut events_sender, events_receiver) = mpsc::channel(128);
+        let (events_sender, events_receiver) = mpsc::unbounded_channel();
 
         let task = tokio::spawn(async move {
             while let Some(event) = client.next().await {
                 let event = event.expect("event expected");
                 events_sender
                     .send(event)
-                    .await
                     .expect("can't send an event to a channel");
             }
         });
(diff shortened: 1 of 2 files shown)


mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -22,7 +27,7 @@ async fn basic_connect_clean_session() {
         .build();
 
     assert_matches!(
-        client.events_receiver.recv().await,
+        client.next().await,
```
lfitchett, line 30: why did this change?

vadim-kovalyov: @dmolokanov suggested instead of exposing receiver, to impl `Stream` for `TestClient`. It allows us to use some helper methods like `.take()` `.collect()` `.filter_map()` etc.

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -41,12 +51,37 @@ impl TestClient {
...
+    }
+
+    /// Watis until client's event loop is finished.
     pub async fn join(self) -> Result<(), JoinError> {
-        self.task.await
+        self.event_loop_handle.await
+    }
+}
+
+impl tokio::stream::Stream for TestClient
```
dmolokanov, line 77:
```suggestion
impl Stream for TestClient
```

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -41,12 +51,37 @@ impl TestClient {
...
+
+    /// Terminates client w/o sending Disconnect packet.
+    pub async fn terminate(self) -> Result<(), JoinError> {
+        self.termination_handle
+            .send(())
+            .expect("unable to send termination signal");
+        self.event_loop_handle.await
+    }
+
+    /// Watis until client's event loop is finished.
```
dmolokanov, line 71:
```suggestion
    /// Waits until client's event loop is finished.
```

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -41,12 +51,37 @@ impl TestClient {
...
+}
+
+impl tokio::stream::Stream for TestClient
+where
+    Self: Unpin,
+{
+    type Item = Event;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.events_receiver.poll_recv(cx)
```
dmolokanov, line 84:
```suggestion
        self.events_receiver.poll_next(cx)
```

mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -113,3 +115,153 @@ async fn basic_pub_sub() {
...
+        .authorizer(|_| Ok(true))
+        .build();
+
+    let (broker_shutdown, broker_task, address) = common::start_server(broker);
+
+    let mut client = TestClientBuilder::new(address)
+        .client_id("mqtt-smoke-tests")
+        .build();
+
+    client
```
dmolokanov, line 136: wouldn't it be nice if we have these methods:
```rust
client.publish_qos0("topic/a", "payload ", true).await;
client.publish_qos2("topic/a", "payload ", true).await;
client.publish_qos2("topic/a", "payload ", true).await;
```

mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -113,3 +115,153 @@ async fn basic_pub_sub() {
...
+
+    client_b
+        .subscribe(SubscribeTo {
+            topic_filter: topic.into(),
+            qos: QoS::ExactlyOnce,
+        })
+        .await
+        .expect("couldn't subscribe to a topic");
+
+    client_b.next().await; //skip connect event
```
dmolokanov, line 247: I'd be nice if we can avoid these events from tests

mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -113,3 +115,153 @@ async fn basic_pub_sub() {
         .unwrap()
         .expect("can't wait for the broker");
 }
+
+#[tokio::test]
+async fn retained_messages() {
```
dmolokanov, line 120: can we make test case names with more context in it? not just it relates to the retained messages

dmolokanov: This comment applicable to all test cases

mqtt/mqtt-broker/tests/smoke.rs
dmolokanov: just wonder: if we spit one stream into several streams by event type, will that work? like one stream for Connect event, one for SubscriptionUpdate, and one for Publish events? 
probably it can help use to await for connect when we need and just don't care about it when we dont need in tests. Also we can extract Event data to those streams...

mqtt/mqtt-broker/tests/common.rs
vadim-kovalyov: There is no `.poll_next()` on `UnboundedReceiver`, I think it should be `poll_recv()`

mqtt/mqtt-broker/tests/smoke.rs
vadim-kovalyov: As discussed offline: we'll add test comments that explain the scenario instead.

mqtt/mqtt-broker/tests/smoke.rs
vadim-kovalyov: I like the idea, I'll explore it in the following PRs.

PR #2889, opened 2020-04-27
Restart edge agent when config is updated

damonbarry:
Currently, when a customer deploys changes to the Edge Agent module, e.g. updated environment variables or Docker `createOptions`, we don't detect them and therefore we don't update/restart the module. To change Edge Agent you must update its Docker image in the deployment, which is severe and cumbersome.

We don't detect most changes in Edge Agent's config because `EdgeAgentDockerModule`'s implementation of `Equals` doesn't even look for them. It could be updated to compare the `env` and `createOptions` from a cached copy of the last known edge configuration--that's how all other modules do it, but that won't work in one important scenario: a brand new Edge Agent container will have an empty config cache until it can download the config from the cloud. If `Equals` is allowed to compare a desired config against the empty config cache, Edge Agent will restart itself endlessly.

The fix is to store Edge Agent's config somewhere else, in a way that's accessible from the _moment_ the container starts. In this change, we save the config as labels on the new Edge Agent container when we create it, right before the old Agent container is stopped/removed. Then when we need to compare a running container's config against desired config from the cloud, we'll use the labels rather than the config cache.

Labels:
`net.azure-devices.edge.create-options`
`net.azure-devices.edge.env`

Notes:
- This is for Edge Agent only; the config cache is still used for all other modules.
- Environment variables are unordered. A new config that specifies the same environment variables in a different order will not trigger a restart of the Agent.
- `createOptions` is stringized JSON, and is therefore compared strictly as a string. A new config that changes `createOptions` in any way _will_ trigger a restart of the Agent.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
```diff
@@ -21,9 +23,24 @@ public override bool Equals(IModule<DockerConfig> other)
...
-                string.Equals(this.Type, other.Type) &&
-                string.Equals(this.Config.Image, other.Config.Image);
+
+            if (!string.Equals(this.Name, other.Name) ||
+                !string.Equals(this.Type, other.Type) ||
+                !string.Equals(this.Config.Image, other.Config.Image))
+                return false;
+
+            IDictionary<string, string> labels = other.Config.CreateOptions?.Labels ?? new Dictionary<string, string>();
+            if (!labels.TryGetValue(Core.Constants.Labels.CreateOptions, out string createOptions) ||
```
philipktlin, line 33: In what situation, it can't get create options from label?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
```diff
@@ -21,9 +23,24 @@ public override bool Equals(IModule<DockerConfig> other)
...
+                !labels.TryGetValue(Core.Constants.Labels.Env, out string env))
+                return true;
+
+            // If the 'net.azure-devices.edge.create-options' and 'net.azure-devices.edge.env' labels exist
+            // on the other IModule, compare them to this IModule
+            if (JsonConvert.SerializeObject(this.Config.CreateOptions) != createOptions)
```
philipktlin, line 39:
```suggestion
            if (!string.Equals(JsonConvert.SerializeObject(this.Config.CreateOptions), createOptions))
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
philipktlin: It will be missing when 1st time start up by iotedged, right?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/CombinedEdgeletConfigProvider.cs
```diff
@@ -29,8 +29,13 @@ public override CombinedDockerConfig GetCombinedConfig(IModule module, IRuntimeI
         {
             CombinedDockerConfig combinedConfig = base.GetCombinedConfig(module, runtimeInfo);
 
-            // if the workload URI is a Unix domain socket then volume mount it into the container
             CreateContainerParameters createOptions = CloneOrCreateParams(combinedConfig.CreateOptions);
+
+            // before making any other modifications to createOptions, save edge agent's createOptions + env as
+            // container labels so they're available as soon as it loads
+            InjectEdgeAgentLabels(module, createOptions);
```
philipktlin, line 36: Should we name it as "InjectContainerLabels"? Currently I am confused why we need to call InjectEdgeAgentLabels even it is creating other modules.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
```diff
@@ -21,9 +23,24 @@ public override bool Equals(IModule<DockerConfig> other)
                 return false;
             if (ReferenceEquals(this, other))
                 return true;
-            return string.Equals(this.Name, other.Name) &&
-                string.Equals(this.Type, other.Type) &&
-                string.Equals(this.Config.Image, other.Config.Image);
+
+            if (!string.Equals(this.Name, other.Name) ||
+                !string.Equals(this.Type, other.Type) ||
+                !string.Equals(this.Config.Image, other.Config.Image))
```
philipktlin, line 29: Existing code doesn't compare Config.NotaryContentTrust, do you know why?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
```diff
@@ -21,9 +23,24 @@ public override bool Equals(IModule<DockerConfig> other)
                 return false;
             if (ReferenceEquals(this, other))
                 return true;
-            return string.Equals(this.Name, other.Name) &&
-                string.Equals(this.Type, other.Type) &&
-                string.Equals(this.Config.Image, other.Config.Image);
+
+            if (!string.Equals(this.Name, other.Name) ||
+                !string.Equals(this.Type, other.Type) ||
```
philipktlin, line 28: I think we previously missed "Version" property.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
```diff
@@ -36,6 +53,8 @@ public override int GetHashCode()
                 hashCode = (hashCode * 397) ^ (this.Version != null ? this.Version.GetHashCode() : 0);
                 hashCode = (hashCode * 397) ^ (this.Type != null ? this.Type.GetHashCode() : 0);
                 hashCode = (hashCode * 397) ^ (this.Config?.Image != null ? this.Config.Image.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (this.Config?.CreateOptions != null ? this.Config.CreateOptions.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (this.Env != null ? this.Env.GetHashCode() : 0);
```
philipktlin, line 57: Content in dictionaries can be the same, but their hash code can be different.  I think we don't need to consider this.Env in GetHashCode.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/CombinedEdgeletConfigProvider.cs
```diff
@@ -62,6 +67,31 @@ static string BindPath(Uri uri)
                 : uri.AbsolutePath;
         }
 
+        static void InjectEdgeAgentLabels(IModule module, CreateContainerParameters createOptions)
```
philipktlin, line 70: do you know what is the difference between CreateContainerParameters.Env (IList<string>) and IModule.Env (IDictionary<string, EnvVal>)?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
damonbarry: Correct, the first time agent starts it won't have the labels, because the "bootstrap" container is created independently by iotedged. And you raise a good point...

Normally the config.yaml will point to the vanilla 1.0 agent or some other stable agent that won't change. The agent specified in the deployment will be different, so the old agent (if it has my changes) will tell iotedged to create the new container with the right labels. The unlabeled container will only run for a very short time and everything will be fine.

But if the config.yaml and the deployment specify the same agent image, then the agent won't be updated when it reads the deployment config, and the unlabeled container will continue to run. This is a problem. To fix it, I think I'll need to make an additional change to iotedged (separate PR) so that it labels the bootstrap container too. Both label values can probably just be empty JSON objects ("{}").

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
damonbarry: It makes sense to do this so I can add the `StringComparison.OrdinalIgnoreCase` parameter.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/CombinedEdgeletConfigProvider.cs
damonbarry: Because it only injects the labels for edgeAgent. See the implementation of the method:
```
if (module.Name.Equals(Constants.EdgeAgentModuleName, StringComparison.OrdinalIgnoreCase))...
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
damonbarry: I think that was intentional. @varunpuranik?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
damonbarry: Yeah I wasn't quite sure of the best thing to do here. You think we don't need `Env`, but we _do_ need `CreateOptions`?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/CombinedEdgeletConfigProvider.cs
damonbarry: Potentially the same information, just different structures for different parts of the pipeline. `CreateContainerParameters` is Docker-specific, while `IModule` is module type-agnostic.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
philipktlin: We can't use this.Config.CreateOptions.GetHashCode() as it doesn't override GetHashCode() method and use default object GetHashCode method, which means the content of 2 CreateOptions can be same but their GetHashCode will return different values.  Same case as Dictionary.

damonbarry: Ah right...I'm mixing reference- and value-type hashing semantics here. I'll remove...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
damonbarry: Actually I don't want to ignore case, so just `StringComparison.Ordinal`. That's equivalent to what I have now, but I'll change it to be explicit.

damonbarry: Done

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
damonbarry: Removed

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
philipktlin: Talked to @ggjjj she will add NotaryContentTrust after this PR.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeAgentDockerModule.cs
varunpuranik: Yup, we don't want to compare version. It is not used.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerEnvironment.cs
```diff
@@ -97,16 +98,21 @@ public async Task<ModuleSet> GetModulesAsync(CancellationToken token)
                         break;
 
                     case Core.Constants.EdgeAgentModuleName:
+                        var labels = dockerRuntimeInfo.Config.CreateOptions?.Labels ?? new Dictionary<string, string>();
```
varunpuranik, line 101: CreateOptions has labels?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/CombinedEdgeletConfigProvider.cs
```diff
@@ -62,6 +67,31 @@ static string BindPath(Uri uri)
...
+            // member of Microsoft.Azure.Devices.Edge.Agent.Core.Agent) and we could get them there, but we'd miss the
+            // bootstrap scenario where edge agent is starting for the first time and therefore has no config store.
+            // That's why edge agent is treated differently: its config is persisted in the container's labels before
+            // it ever starts.
+            if (module.Name.Equals(Constants.EdgeAgentModuleName, StringComparison.OrdinalIgnoreCase))
+            {
+                // get createOptions JSON before making any updates to labels
+                string createOptionsJson = JsonConvert.SerializeObject(createOptions);
+
+                var moduleWithDockerConfig = (IModule<DockerConfig>)module; // cast is safe; base impl already checked it
```
varunpuranik, line 84: This seems a bit scary.. I agree this should work in our cases, but is it guaranteed to be true?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Constants.cs
```diff
@@ -111,6 +111,8 @@ public static class Labels
             public const string DesiredStatus = "net.azure-devices.edge.desiredStatus";
             public const string NormalizedCreateOptions = "net.azure-devices.edge.normalizedCreateOptions";
             public const string ConfigurationId = "net.azure-devices.edge.configurationId";
+            public const string CreateOptions = "net.azure-devices.edge.create-options";
+            public const string Env = "net.azure-devices.edge.env";
```
varunpuranik, line 115: Can you pls remind me again why we need Env vars as a label, as opposed to just checking the env vars available in the container? 
(I remember discussing this, but don't remember why we decided we need this).

damonbarry: We need to know which vars the customer set in the deployment, not the larger set of vars on the container (including the IOTEDGE_* variables we add, anything docker adds, etc). I think it would be difficult to know exactly which vars came from the deployment.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/CombinedEdgeletConfigProvider.cs
damonbarry: What's the scary part exactly? The cast? The caller of this helper first calls `base.GetCombinedConfig` which throws if `module` is not `IModule<DockerConfig>`.

varunpuranik: Its fine for now.. but we should not take a dependency on checks in callers. What if this code path is called from somewhere else?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Constants.cs
varunpuranik: Yes, and we need to track the environment variables that were removed, which becomes tricky. So this looks good. Thanks.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/CombinedEdgeletConfigProvider.cs
damonbarry: Agreed. Originally this code was written directly into the calling method. Later, when I moved it into a helper routine, this assumption became even more tenuous. Actually, why didn't I just make this helper take `IModule<DockerConfig>` as an argument instead of `IModule`? 🙄

PR #2891, opened 2020-04-28
[mqtt]: Retained messages zero payload integration tests

vadim-kovalyov:
This PR adds:
- Retained messages w/ zero-length payload test
- Some TestClient improvements

### DIFF
--- a/mqtt/mqtt-broker/tests/common.rs
+++ b/mqtt/mqtt-broker/tests/common.rs
@@ -10,17 +10,18 @@ use futures_util::{FutureExt, StreamExt};
 use lazy_static::lazy_static;
 use tokio::{
     net::ToSocketAddrs,
+    stream::Stream,
     sync::{
         mpsc::{self, UnboundedReceiver},
         oneshot::{self, Sender},
     },
-    task::{JoinError, JoinHandle},
+    task::JoinHandle,
 };
 
 use mqtt3::{
-    proto::{Publication, SubscribeTo},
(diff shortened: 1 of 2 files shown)


mqtt/mqtt-broker/tests/common.rs
```diff
@@ -44,37 +45,90 @@ impl TestClient {
         self.publish_handle.publish(publication).await
     }
 
+    pub async fn publish_qos0(&mut self, topic: &str, payload: &str, retain: bool) {
```
dmolokanov, line 48:
```suggestion
    pub async fn publish_qos0(&mut self, topic: impl Into<String>, payload: impl Into<Bytes>, retain: bool) {
```
because https://docs.rs/bytes/0.5.4/bytes/struct.Bytes.html#impl-From%3C%26%27static%20%5Bu8%5D%3E

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -44,37 +45,90 @@ impl TestClient {
...
+            topic_name: topic.into(),
+            qos: QoS::ExactlyOnce,
+            retain,
+            payload: payload.to_owned().into(),
+        })
+        .await
+        .expect("couldn't publish")
+    }
+
     pub async fn subscribe(
```
dmolokanov, line 81:
```suggestion
    pub async fn subscribe(topic: impl Into<String>, qos: QoS)
```
?

PR #2893, opened 2020-04-28
Alternate cleanup processor for MessageStore

dylanbronson:
With the addition of PriorityQueues, the CleanupProcessor assumptions change slightly: Previously, we could always assume that if a message at the head of the queue should not be deleted, then none of the other messages in the queue should be either. Now, because we can have different TTL's for messages within the same queue, there can be messages that have expired in the queue after the head. 

The checkEntireQueueOnCleanup flag is an environment variable for edgeHub. If it is set to true, we will check the entire queue every time cleanup processor runs. If it is set to false, we just remove the oldest items in the queue until we get to one that is not expired.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -232,96 +234,124 @@ void EnsureCleanupTask(object state)
...
+                    throw;
+                }
+            }
+
+            private async Task CleanQueue(bool checkEntireQueueOnCleanup)
             {
                 long totalCleanupCount = 0;
                 long totalCleanupStoreCount = 0;
-
-                try
```
dylanbronson, line 239: The diff looks a bit funky here because I moved the queue cleaning logic into a method

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -232,96 +234,124 @@ void EnsureCleanupTask(object state)
...
+                int batchSize = 10;
```
richma-ms, line 253: >int batchSize = 10; [](start = 16, length = 19)

constant?

PR #2894, opened 2020-04-28
fix clippy errors

philipktlin:
fix clippy errors


darobs:
Looks like windows clippy failed with this:
iotedged\src\signal.rs:62:5
   |
62 |     use tokio_signal;
   |     ^^^^^^^^^^^^^^^^^ help: remove it entirely

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -1012,7 +1010,13 @@ where
 
 #[cfg(test)]
 mod tests {
-    use super::*;
```
darobs, line 1015: What's the clippy rule here?  What are we gaining by spelling out dozens of types and functions we use in a unit test?

edgelet/dps/src/registration.rs
```diff
@@ -3,7 +3,6 @@
 use std::sync::{Arc, RwLock};
 use std::time::{Duration, Instant};
 
-use base64;
```
dmolokanov, line 6: what happened? why has the build started to fail all of a sudden? Some problems are related to new rust update. But some - not

edgelet/edgelet-kube/src/convert/to_k8s.rs
```diff
@@ -540,8 +550,17 @@ mod tests {
     use edgelet_docker::DockerConfig;
     use edgelet_test_utils::cert::TestCert;
 
-    use crate::constants::env::*;
-    use crate::constants::*;
+    use crate::constants::env::{
```
dmolokanov, line 553: nit: I'd say we use this constants via `env::CONSTANT_NAME` instead of importing all of them

edgelet/dps/src/registration.rs
```diff
@@ -593,13 +591,14 @@ mod tests {
...
 
     use edgelet_core::crypto::{MemoryKey, MemoryKeyStore};
-    use http;
     use hyper::{self, Body, Request, Response, StatusCode};
-    use serde_json;
-    use tokio;
     use url::Url;
 
-    use super::*;
+    use super::{
```
dmolokanov, line 597: Although this compiles and works it is confusing now. `super` module (parent) which is `registration` does not have a child `future` it just imports `future` from futures crate.
It'd be really nice if we can avoid this confusion by importing with appropriate crate: not `super::future` but `futures::future`.

dmolokanov: This comment applies to all files

edgelet/hsm-rs/src/lib.rs
```diff
@@ -13,7 +13,22 @@
     clippy::use_self
 )]
 
-use hsm_sys::*;
```
darobs, line 16: I hope you were able to just copy&paste this from the clippy output.

edgelet/dps/src/registration.rs
philipktlin: I needed to fix clippy errors as this PR https://github.com/Azure/iotedge/pull/2871 failed check-in build.  I think we use latest clippy (nightly?) which causes checkin build failed.

edgelet/dps/src/registration.rs
philipktlin: doesn't this the same problem of existing code?  And should the test always use whatever the parent uses?  This recalls 1 thing I read from "Rust programming language" book, it depends on how do you want to organize your tests.  If you want tests stick with implementation in the same file, uses super seems more appropriate.  If we plan to move tests to its own folder, then state references explicitly.

Since this doesn't change existing behavior, can this be handled in another PR? and we should make sure this is what our team is going to follow.

edgelet/edgelet-kube/src/convert/to_k8s.rs
philipktlin: I think all these constants are used in this file.  Do you mean that we should state full qualified name like "crate::constants::env::<constant name>" in every usage?

edgelet/dps/src/registration.rs
darobs: I agree with Denis - usage of external crates shouldn't be imported from "super".  I would suggest creating an issue and assign owners as appropriate for each section.

edgelet/dps/src/registration.rs
dmolokanov: we use stable clippy. in the latest update, the new wildcard check was added. but it doesn't explain why unused imports passed previously

edgelet/edgelet-kube/src/convert/to_k8s.rs
dmolokanov: it's a nit, so you can leave it here. it's just constants that may have the same name because they live if different modules, but importing it like this can lead to a name collision.
it's not the case here though

```rust
// I meant
use crate::contants::env;

...

// later in the code use it like

foo(env::CONSTANT_NAME);
```

edgelet/edgelet-docker/src/runtime.rs
dmolokanov: I agree with clippy that using wildcard leads to confusion. Receiver trait exist in std, in crossbeam, in tokio: oneshot, bounded, unbounded.... I really want to know without IDE what this Receiver came from!!!
The only exception is `prelude::*` case.

edgelet/dps/src/registration.rs
philipktlin: created a bug https://dev.azure.com/msazure/One/_workitems/edit/7123592 to track.

edgelet/edgelet-docker/src/runtime.rs
philipktlin: support link: https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_imports

PR #2895, opened 2020-04-28
Edgelet build fix for Mac OS

massand:
A quick PR for fixing MacOS edgelet build. The last commit on that file did some refactor that ignored non-Windows, non-Linux. If we don't want zero uptime reporting, we could use libc I guess.


arsing:
>If we don't want zero uptime reporting, we could use libc I guess.

No, it's fine. It's a niche request so it's not worth the effort.

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -631,6 +631,9 @@ impl ModuleRuntime for DockerModuleRuntime {
                 })
             });
 
+        #[cfg(not(any(windows, target_os = "linux")))]
+        let uptime: u64 = 0;
+
         #[cfg(target_os = "linux")]
```
dmolokanov, line 637: will this work if you simply change to this?
```suggestion
        #[cfg(target_os = "unix")]
```

massand: target_os = "unix" doesn't build. There's actually a "macos" target_os. But, libc doesn't have sysinfo for "macos" platform.

dmolokanov: `#[cfg(unix)]` its not a target_os, but family. but it won't work anyways due to sysinfo limitations...

dmolokanov:
You don't like Kodiak?

PR #2897, opened 2020-04-29
[mqtt] Benchmark broker message dispatching

dmolokanov:
In order to verify how quickly broker can dispatch messages we need some benchmark tests.
`dispatch_message` benches start broker with empty state and no network connections. Firstly, each bench test creates a given number of publishers, subscribers and connects them to a broker. Then it makes a series of exercises to send messages to a broker and measures the time between a message successfully sent to a broker and broker processed this message. For this case broker extended with a `on_publish` "callback" which has to be exposed only for benchmark tests.

### 1-to-1 
Each publisher sends messages to one corresponding subscriber:
```
pub1 pub topic/1            sub1 sub topic/1
pub2 pub topic/2            sub2 sub topic/2
pub2 pub topic/N            sub2 sub topic/N
```
### fan-in
All publishers send message to a randomly chose topic
```
pub1 pub topic/2            sub1 sub topic/1
pub2 pub topic/2            sub2 sub topic/2
pubN pub topic/2            sub2 sub topic/N
```
### fan-out 
Multiple publishers sends messages to a topic all subscribers subscribed to
```
pub1 pub topic/foo            sub1 sub topic/foo
pub2 pub topic/foo            sub2 sub topic/foo
pubN pub topic/foo            sub2 sub topic/foo
```

## How to run benches
```bash 
cd mqtt 
cargo bench --bench dispatch_messages \
    --features="benches" \
    --manifest-path mqtt-broker/Cargo.toml
```
or 
```bash
cd mqtt/mqtt-broker
cargo bench --bench dispatch_messages --features="benches"
```
to run all benches
```bash 
cd mqtt/mqtt-broker
cargo bench --all --features="benches"
```


mqtt/mqtt-broker/benches/dispatch_messages.rs
```diff
@@ -0,0 +1,331 @@
...
+                        ClientEvent::PubRel(_) => None,
+                        ClientEvent::PubComp(_) => None,
+                        ClientEvent::DropConnection
+                        | ClientEvent::CloseSession
+                        | ClientEvent::Disconnect(_) => {
+                            stop = true;
+                            None
+                        }
+                        ClientEvent::PingReq(_) => Some(ClientEvent::PingResp(proto::PingResp)),
+                        ClientEvent::PingResp(_) => Some(ClientEvent::PingReq(proto::PingReq)),
```
vadim-kovalyov, line 242: Maybe I'm missing something, but are you sure you need to reply with PingReq on PingResp? won't you get an "infinite loop" of pings?

mqtt/mqtt-broker/benches/dispatch_messages.rs
```diff
@@ -0,0 +1,331 @@
...
+                proto::PacketIdentifierDupQoS::AtLeastOnce(next, false)
+            }
+            proto::QoS::ExactlyOnce => todo!(),
+        };
+
+        proto::Publish {
+            packet_identifier_dup_qos: packet_id,
+            retain: false,
+            topic_name,
+            payload: Bytes::from_iter((0..payload_size.into()).map(|_| rand::random::<u8>())),
```
vadim-kovalyov, line 301: nit: I'm not sure if randomness does matter here. Won't `vec![0u8, size]` spare some cycles?

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -38,6 +38,8 @@ where
     retained: HashMap<String, proto::Publication>,
     authenticator: N,
     authorizer: Z,
+
+    pub on_publish: Option<Sender<()>>,
```
vadim-kovalyov, line 42: q: I'm not sure, but is it possible to guard this with `[cfg(test)]`?

mqtt/mqtt-broker/benches/dispatch_messages.rs
dmolokanov: The idea was the client responds to the broker ping request. But I wonder if the broker actively checks clients. Probably not...

mqtt/mqtt-broker/benches/dispatch_messages.rs
dmolokanov: 1. data preparation doesn't participate in execution measurements., only message publishing 
2. randomization help to make cache missed when loading that "emulates" real behavior.

mqtt/mqtt-broker/src/broker.rs
dmolokanov: not with tests, I'm going to add special feature and run benches with this feature enabled

mqtt/mqtt-broker/Cargo.toml
```diff
@@ -38,9 +36,10 @@ atty = "0.2"
 bytes = "0.5"
 criterion = "0.3"
 fail = { version = "0.3", features = ["failpoints"] }
+itertools = "0.9"
 matches = "0.1"
 proptest = "0.9"
-rand = "0.3"
+rand = "0.7"
```
myagley, line 42: We should probably put these dependencies behind the feature flag as well.

myagley: Not blocking an approval on this. But, something we might want to consider to keep it clean.

dmolokanov: this is a good idea, but it seems like it doesn't work with stable rust now https://github.com/rust-lang/cargo/issues/5954

mqtt/mqtt-broker/benches/dispatch_messages.rs
```diff
@@ -0,0 +1,444 @@
...
+    init_logging();
+
+    let mut group = c.benchmark_group("fan_out");
+    for (qos, count, payload_size) in scenarios() {
+        let strategy = Strategy::new_fan_out(count, count);
+        dispatch_messages(&mut group, strategy, count, payload_size, qos);
+    }
+    group.finish();
+}
+
```
vipeller, line 61: these three methods look exactly the same, except the strategy they create and the name. Can it be made into a single method, and fan_in()/fan_out()/one_to_one() just call that single method with a "strategy creator" function injected?

dmolokanov: Yeah, I thought about that. I still wonder if number of publishers should not be equal number of subscribers. it may vary depending on strategy...

mqtt/mqtt-broker/benches/dispatch_messages.rs
```diff
@@ -0,0 +1,444 @@
...
+                            }
+                            None
+                        }
+                        ClientEvent::PubRec(_) => None,
+                        ClientEvent::PubRel(_) => None,
+                        ClientEvent::PubComp(_) => None,
+                        ClientEvent::DropConnection
+                        | ClientEvent::CloseSession
+                        | ClientEvent::Disconnect(_) => {
+                            stop = true;
```
vadim-kovalyov, line 271: nit: log(...); break;

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -181,10 +184,10 @@ where
                 Ok(())
             }
             ClientEvent::PubAck0(id) => self.process_puback0(&client_id, id),
-            ClientEvent::PubAck(ref puback) => self.process_puback(&client_id, puback),
-            ClientEvent::PubRec(ref pubrec) => self.process_pubrec(&client_id, pubrec),
-            ClientEvent::PubRel(ref pubrel) => self.process_pubrel(&client_id, pubrel),
-            ClientEvent::PubComp(ref pubcomp) => self.process_pubcomp(&client_id, pubcomp),
+            ClientEvent::PubAck(puback) => self.process_puback(&client_id, &puback),
```
vadim-kovalyov, line 187: nit: since you've touched this, why do we even need refs here? we are deconstructing the event anyways, why not move into `process_puback`. Just a question, no need to be fixed in this PR :)

mqtt/mqtt-broker/Cargo.toml
dmolokanov: Fixed with `benches` feature

mqtt/mqtt-broker/benches/dispatch_messages.rs
dmolokanov: it logged bellow

mqtt/mqtt-broker/src/broker.rs
dmolokanov: good catch!

dmolokanov: Actually it's a clippy that forces to borrow values that were not consumed:
https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value

PR #2899, opened 2020-04-29
E2E test for PriorityQueue Offline case

dylanbronson:
This E2E test does the following:

1. Deploys edgeAgent and edgeHub with multiple routes set from LoadGen -> IotHub (specifically EventHub) with different priorities
2. Deploys LoadGen, TestResultCoordinator and NetworkController modules
3. NetworkController shuts off the network from the cloud
4. LoadGen begins sending messages through edge to the cloud (but since the network is shut off, edge cannot forward them to the cloud)
5. LoadGen finishes sending messages and sends the expected order of those messages to TestResultCoordinator
6. NetworkController turns the network back on
7. The test reads data from EventHub until we have received all the expected messages.
8. The test sends the actual order of the messages received to TestResultCoordinator
9. TestResultCoordinator matches actual to expected messages and returns the result to the test
10. Test passes if all messages received in correct order. Fails otherwise. Duplicates are accounted for (and are legal).

This only works for Linux, as NetworkController hasn't been implemented for Windows yet.


test/connectivity/modules/NetworkController/LinuxTrafficControlController.cs
```diff
@@ -102,16 +102,22 @@ async Task<bool> AddDropRule(CancellationToken cs)
 
                 // Adding rules to filter packages by iotHub IP
                 // Details about how the rules work https://wiki.archlinux.org/index.php/Advanced_traffic_control
+                Log.LogInformation($"Executing: {LinuxTrafficControllerHelper.GetRootRule(this.networkInterfaceName)}");
```
dylanbronson, line 105: These logs were originally just for testing, but they seem helpful, so I kept them in

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -84,6 +126,31 @@ public async Task PriorityQueueTimeToLive()
...
+        {
+            // Use a set to account for duplicates
+            HashSet<int> results = new HashSet<int>();
+            await this.iotHub.ReceiveEventsAsync(
+                this.runtime.DeviceId,
+                startTime,
+                data =>
+                {
+                    int sequenceNumber = int.Parse(data.Properties["sequenceNumber"].ToString());
+                    Log.Information($"Received message from IoTHub with sequence number: {sequenceNumber}");
```
damonbarry, line 139: This should be `Log.Verbose`

test/Microsoft.Azure.Devices.Edge.Test/helpers/Context.cs
```diff
@@ -161,6 +162,8 @@ public Context()
 
         public Option<string> RelayerImage { get; }
 
+        public Option<string> NetworkControllerImage { get; }
```
damonbarry, line 165: Update `test/README.md` to document this new test parameter.

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -54,6 +58,44 @@ public async Task PriorityQueueModuleToModuleMessages()
             await this.ValidateResultsAsync();
         }
 
+        [Test]
+        public async Task PriorityQueueModuleToHubMessages()
+        {
+            // TODO: Add Windows and ARM32. Windows won't be able to work for this test until we add NetworkController Windows implementation
+            if (OsPlatform.IsWindows() || !OsPlatform.Is64Bit())
```
varunpuranik, line 65: Is there a reason to block it in code, in this test?
Since we don't even run these tests on Windows, it should not matter, right? And when we enable NetworkController on Windows, it can start running automatically?

dylanbronson: The E2E test suite runs in Windows, just not this specific test. Other tests in this file will also run on Windows very soon - just a matter of debugging.
So when these run on the pipeline (using dotnet test), this needs to be there to skip the test for Windows. Unless there's some other way of skipping them.

test/connectivity/modules/NetworkController/Program.cs
```diff
@@ -73,10 +80,35 @@ static async Task Main()
...
+        {
+            Environment.SetEnvironmentVariable("IOTEDGE_GATEWAYHOSTNAME", string.Empty);
+            ModuleClient moduleClient = await ModuleUtil.CreateModuleClientAsync(TransportType.Amqp_Tcp_Only, ModuleUtil.DefaultTimeoutErrorDetectionStrategy, ModuleUtil.DefaultTransientRetryStrategy);
+            await moduleClient.SetMethodHandlerAsync("toggleConnectivity", ToggleConnectivity, new Tuple<string, CancellationToken>(networkInterfaceName, token));
+        }
+
+        private static async Task<MethodResponse> ToggleConnectivity(MethodRequest methodRequest, object userContext)
+        {
+            Log.LogInformation("Direct method toggleConnectivity has been invoked.");
+            var networkInterfaceNameAndToken = (Tuple<string, CancellationToken>)userContext;
```
ancaantochi, line 98: nit: see named tuples so you can use the names instead of Item1, Item2

test/connectivity/modules/NetworkController/Program.cs
```diff
@@ -31,7 +37,7 @@ static async Task Main()
                         var offline = new OfflineController(name, Settings.Current.IotHubHostname, Settings.Current.NetworkRunProfile.ProfileSetting);
                         var satellite = new SatelliteController(name, Settings.Current.IotHubHostname, Settings.Current.NetworkRunProfile.ProfileSetting);
                         var cellular = new CellularController(name, Settings.Current.IotHubHostname, Settings.Current.NetworkRunProfile.ProfileSetting);
-                        var controllers = new List<INetworkController>() { offline, satellite, cellular };
+                        controllers = new List<INetworkController>() { offline, satellite, cellular };
```
ancaantochi, line 40: nit: add controllers to the list instead of creating new list

test/connectivity/modules/NetworkController/Program.cs
```diff
@@ -73,10 +80,35 @@ static async Task Main()
...
             await cts.Token.WhenCanceled();
+            await CleanupControllingRulesOnShutdown(controllers, CancellationToken.None);
             completed.Set();
             handler.ForEach(h => GC.KeepAlive(h));
         }
 
+        private static async Task SetToggleConnectivityMethod(string networkInterfaceName, CancellationToken token)
+        {
+            Environment.SetEnvironmentVariable("IOTEDGE_GATEWAYHOSTNAME", string.Empty);
+            ModuleClient moduleClient = await ModuleUtil.CreateModuleClientAsync(TransportType.Amqp_Tcp_Only, ModuleUtil.DefaultTimeoutErrorDetectionStrategy, ModuleUtil.DefaultTransientRetryStrategy);
```
ancaantochi, line 91: Read the TransportType from env variable

test/connectivity/modules/NetworkController/Program.cs
```diff
@@ -73,10 +80,35 @@ static async Task Main()
...
+            await moduleClient.SetMethodHandlerAsync("toggleConnectivity", ToggleConnectivity, new Tuple<string, CancellationToken>(networkInterfaceName, token));
+        }
+
+        private static async Task<MethodResponse> ToggleConnectivity(MethodRequest methodRequest, object userContext)
+        {
+            Log.LogInformation("Direct method toggleConnectivity has been invoked.");
+            var networkInterfaceNameAndToken = (Tuple<string, CancellationToken>)userContext;
+
+            // true for network on (restriction disabled), false for network off (restriction enabled)
+            bool networkOnValue = (bool)JObject.Parse(methodRequest.DataAsJson)["networkOnValue"];
```
ancaantochi, line 101: please use Boolean.TryParse

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -54,6 +58,42 @@ public async Task PriorityQueueModuleToModuleMessages()
...
+            string networkControllerImage = Context.Current.NetworkControllerImage.Expect(() => new ArgumentException("networkControllerImage parameter is required for Priority Queues test"));
+            string trackingId = Guid.NewGuid().ToString();
+            TestInfo testInfo = this.InitTestInfo(5, 1000, true);
+
+            var testResultReportingClient = new TestResultReportingClient { BaseUrl = "http://localhost:5001" };
+
+            Action<EdgeConfigBuilder> addInitialConfig = this.BuildAddInitialConfig(trackingId, "hubtest", trcImage, loadGenImage, testInfo, true);
+            Action<EdgeConfigBuilder> addNetworkControllerConfig = this.BuildAddNetworkControllerConfig(trackingId, networkControllerImage);
+            EdgeDeployment deployment = await this.runtime.DeployConfigurationAsync(addInitialConfig + addNetworkControllerConfig, token);
+            await this.ToggleConnectivity(false, NetworkControllerModuleName, token);
```
ancaantochi, line 83: It's not easy to understand what false/true means for ToggleConnectivity. Maybe sending networkOn/Off or online/offline is more clear.

PR #2904, opened 2020-04-30
Fix route config change bug

philipktlin:
When route endpoint is removing, endpoint executor's cancellation token source will be canceled. And endpoint executor is running inside a loop to process messages and wait for endpoint executor FSM to run sending command. Inside FSM, it will keep retry sending messages until delivered or any exception not able to retry. If an invalid/non-existing endpoint is defined, it will keep retrying forever and endpoint executor is awaiting the result from the sending command forever. Therefore we need to have a way to tell FSM that endpoint is removing and should stop retrying. Since endpoint executor is awaiting forever, it block the removal of endpoint which will block to add/update endpoints (in Dispatcher.ReplaceRoutes method).

In order to let FSM aware of removing endpoints, endpoint executor need to call FSM CloseAsync to complete current command in FSM.  This will unblock sendMessageTask and allow to completely remove endpoint executor.  In previous PR https://github.com/Azure/iotedge/pull/2865, Mike commented to use ImmutableDictionary and close all FSMs to unblock sendMessageTask in EndpointExecutor.

Step to reproduce:

Run iotedge with tempsenor and tempfilter; set up to route messages from tempsensor to tempfilterXXX, which is an invalid route.
start to run iotedged and check log of tempFilter, there is no message shown.
fix route to tempfilter and deploy.
expected tempfilter log should start showing up records, but it doesn't since the route can't be removed and new route can't be added.


edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -1,153 +1,162 @@
-// Copyright (c) Microsoft. All rights reserved.
```
varunpuranik, line 1: Is there some formatting issue here? Why are all lines showing up as diffs? Can you pls fix that.. hard to tell what has actually changed.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -44,7 +44,7 @@ public class StoringAsyncEndpointExecutor : IEndpointExecutor
             this.options = Preconditions.CheckNotNull(options);
             this.messageStore = messageStore;
             this.sendMessageTask = Task.Run(this.SendMessagesPump);
-            this.prioritiesToFsms = new AtomicReference<SortedDictionary<uint, EndpointExecutorFsm>>(new SortedDictionary<uint, EndpointExecutorFsm>());
+            this.prioritiesToFsms = new AtomicReference<ImmutableDictionary<uint, EndpointExecutorFsm>>(ImmutableDictionary<uint, EndpointExecutorFsm>.Empty);
```
myagley, line 47: Were we relying on the sorted-ness of the `SortedDictionary`?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -144,10 +151,11 @@ public async Task UpdatePriorities(IList<uint> priorities, Option<Endpoint> newE
             // Update priorities by merging the new ones with the existing.
             // We don't ever remove stale priorities, otherwise stored messages
             // pending for a removed priority will never get sent.
-            SortedDictionary<uint, EndpointExecutorFsm> snapshot = this.prioritiesToFsms;
+            ImmutableDictionary<uint, EndpointExecutorFsm> snapshot = this.prioritiesToFsms;
+            Dictionary<uint, EndpointExecutorFsm> updatedSnapshot = new Dictionary<uint, EndpointExecutorFsm>(snapshot);
```
varunpuranik, line 155: Why do we need this?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -159,22 +167,22 @@ public async Task UpdatePriorities(IList<uint> priorities, Option<Endpoint> newE
...
                     // Update the existing FSM with the new endpoint
-                    EndpointExecutorFsm fsm = snapshot[priority];
+                    EndpointExecutorFsm fsm = updatedSnapshot[priority];
                     await newEndpoint.ForEachAsync(e => fsm.RunAsync(Commands.UpdateEndpoint(e)));
                 }
             }
 
-            this.prioritiesToFsms.Value = snapshot;
-            Events.UpdatePrioritiesSuccess(this, snapshot.Keys.ToList());
+            this.prioritiesToFsms.CompareAndSet(snapshot, updatedSnapshot.ToImmutableDictionary());
```
varunpuranik, line 180: What happens if this compare fails?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
varunpuranik: Yeah, to process messages in priority order.. but he is sorting the dict before using below, so that should work.

varunpuranik:
Can we add some tests for this?

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
philipktlin: we need to update dictionary and use immutable dictionary will create many copies.  Therefore I created a local copy of dictionary for update and convert to immutable dictionary once at the end.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
philipktlin: There are 2 places calling prioritiesToFsms.CompareAndSet in this class, which are UpdatePriorites and Dispose methods.  Dispose method should be called by GC once CloseAsync method is called and free up all the references and all other methods are completed.  Therefore Dispose should not be called at the same time of UpdatePriorities.  So what happens if 2 threads calling UpdatePriorites method?  Currently there are 2 places call to UpdatePriorities method, 1) from StoringAsyncEndpointExecutorFactory.CreateAsync method, this is create scenario.  Therefore it should not have 2 threads to try calling UpdatePriorities. 2) from its SetEndpoint method, this means 2 config change requests sending to EdgeHub almost at the same time and one is updated successfully but not the other.  This line will throw exception and will catch in SetEndpoint method try-catch block and report SetEndpointFailure, then rethrow and caught again in ConfigUpdater, which logs an error.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
richma-ms, line 155: >snapshot [](start = 110, length = 8)

can this just be prioritiesToFsms and have the snapshot taken implicitly?

richma-ms, line 155: >Dictionary [](start = 12, length = 10)

why not use a SortedDictionary as the updated snapshot?  You can still convert it to an ImmutableDictionary at the end when writing back to the AtomicReference.  This way the sorting is happening on insertion, and we wouldn't have to re-sort on every iteration of the message pump loop.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -214,11 +222,12 @@ async Task SendMessagesPump()
                     {
                         await this.hasMessagesInQueue.WaitAsync(this.options.BatchTimeout);
 
-                        SortedDictionary<uint, EndpointExecutorFsm> snapshot = this.prioritiesToFsms;
+                        ImmutableDictionary<uint, EndpointExecutorFsm> snapshot = this.prioritiesToFsms;
                         bool haveMessagesRemaining = false;
 
+                        uint[] orderedPriorities = snapshot.Keys.OrderBy(k => k).ToArray();
```
richma-ms, line 228: >uint[] orderedPriorities = snapshot.Keys.OrderBy(k => k).ToArray(); [](start = 24, length = 67)

see above comment, we're re-sorting here on inner loop, which seems unnecessary if the priorities haven't changed

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
philipktlin: It is because ImmutableDictionary is not guaranteed to be sorted, and we should not depend on internal implementation how immutableDictionary persist the data.  Therefore uses Dictionary here and do sorting during consumption.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
philipktlin: see reply above.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
philipktlin: Yes. I will log an event if CompareAndSet returns false.  I traced up the usages and all methods changing endpoints in dispatcher used async lock to avoid 2 threads to calling any executor at the same time.

varunpuranik: In that case, why do you need CompareAndSet? Just use the Set.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
```diff
@@ -159,22 +167,28 @@ public async Task UpdatePriorities(IList<uint> priorities, Option<Endpoint> newE
...
                     // Update the existing FSM with the new endpoint
-                    EndpointExecutorFsm fsm = snapshot[priority];
+                    EndpointExecutorFsm fsm = updatedSnapshot[priority];
                     await newEndpoint.ForEachAsync(e => fsm.RunAsync(Commands.UpdateEndpoint(e)));
                 }
             }
 
-            this.prioritiesToFsms.Value = snapshot;
-            Events.UpdatePrioritiesSuccess(this, snapshot.Keys.ToList());
+            if (this.prioritiesToFsms.CompareAndSet(snapshot, updatedSnapshot.ToImmutableDictionary()))
```
philipktlin, line 180: @varunpuranik  I think it is more safe to use CompareAndSet and write an error log if happens.

edge-hub/src/Microsoft.Azure.Devices.Routing.Core/endpoints/StoringAsyncEndpointExecutor.cs
philipktlin: It is more preventive.  Even caller get changed and causes race condition, at least we will report it in log.

philipktlin:
> Can we add some tests for this?

Done.  Add 1 test to ensure that in-processing FSM will get closed if Endpoint executor close method is called.

PR #2906, opened 2020-04-30
Update window package versioning on Window Edgelet pipeline for master branch

yophilav:
Update window package versioning on Window Edgelet build pipeline for master branch.
The current way of version concatenation can produce a version string that violates `FFFF.FFFF.FFFF.FFFF` format.


builds/misc/packages.yaml
```diff
@@ -135,7 +135,13 @@ jobs:
     steps:
       - powershell: |
           $base_version = Get-Content -Path "$(Build.SourcesDirectory)\edgelet\version.txt"
-          $version = ("{0}{1}" -f $base_version, $(Build.BuildNumber))
+          if ($base_version -like '*~*') {
+            $version = ("{0}{1}" -f $base_version, $(Build.BuildNumber))
+          } else {
+            $version = $base_version
+          }
+          echo "base_version=$base_version"
```
arsing, line 143: Are you planning to leave these in or are they just for debugging? If you're planning to leave them in, use `Write-Host` instead of `echo`.

arsing:
>Fixing window package version on master

The build is fine on master, right? It's only broken on `release/*` ?

yophilav:
> > Fixing window package version on master
> 
> The build is fine on master, right? It's only broken on `release/*` ?

That is correct. The build is fine because we are using `1.0.10~dev` version.

builds/misc/packages.yaml
yophilav: Done.

arsing:
>>>Fixing window package version on master
>>
>>The build is fine on master, right? It's only broken on `release/*` ?
>
>That is correct. The build is fine because we are using `1.0.10~dev` version.

Right, so please fix the commit message before you merge it.

Also, I assume you tested this on the `release/1.0.9` branch already?

yophilav:
> > > > Fixing window package version on master
> > > 
> > > 
> > > The build is fine on master, right? It's only broken on `release/*` ?
> > 
> > 
> > That is correct. The build is fine because we are using `1.0.10~dev` version.
> 
> Right, so please fix the commit message before you merge it.
> 
> Also, I assume you tested this on the `release/1.0.9` branch already?

Yes, I did.

PR #2911, opened 2020-04-30
Limit route priority to 0-9

richma-ms:
For the initial iteration of prioritized messages, we're limiting the usage to 10 different customizable values, 0-9 inclusive.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/RouteConfiguration.cs
```diff
@@ -22,8 +23,15 @@ public RouteConfiguration(string route)
         public RouteConfiguration(string route, uint priority, uint timeToLiveSecs)
         {
             this.Route = Preconditions.CheckNonWhiteSpace(route, nameof(route));
-            this.Priority = priority;
             this.TimeToLiveSecs = timeToLiveSecs;
+
+            // Verify the route, this must be either between [0-9], or the default value
+            if ((priority < 0 || priority > 9) && priority != RouteFactory.DefaultPriority)
```
varunpuranik, line 29: Do we have tests for this?

varunpuranik: What happens when I set a route to 20? Does EH log the error and continue chugging along? Gets default priority?

richma-ms: no, the exception is caught and sent up as reported status by the Agent

PR #2916, opened 2020-05-04
added authenticator for edgehub

vipeller:


mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
...
+    password: Option<String>,
+    certificate: Option<String>,
+    certificate_chain: Option<Vec<String>>,
+}
+
+impl From<(Option<String>, Credentials)> for EdgeHubAuthRequest {
+    fn from(item: (Option<String>, Credentials)) -> Self {
+        let (password, certificate) = match item.1 {
+            Credentials::Password(p) => (p, None),
+            Credentials::ClientCertificate(c) => (None, Some(encode(c))),
```
dmolokanov, line 73:
```suggestion
            Credentials::ClientCertificate(c) => (None, Some(base64::encode(c))),
```

mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
+use async_trait::async_trait;
+use base64::encode;
+use failure::ResultExt;
+use serde::{Deserialize, Serialize};
+use std::fmt;
+
+use crate::auth::{AuthId, Authenticator, Credentials};
+use crate::{Error, ErrorKind};
+
+pub struct EdgeHubAuthenticator {
```
dmolokanov, line 10: Usually it easier to read when code for a struct structured this way
```rust
// struct goes first
pub struct Foo{ ... }

// impl with constructor and other methods goes next
impl Foo {
  // struct lifecycle usually starts with constructor
  pub fn new() { }
  ...
  // other function goes next
  pub fn do_work() {
    do_step1();
    do_step2();
  }
  
  // each step of algorithm `do_work` goes right after `do_work` method 
  fn do_step1() { .. }
  fn do_step2() { .. }

  // another method
  pub fn do_some_thing_else() { .. }
}

// impl traits after impl Foo
impl TraitBar for Foo {
  fn bar(&self) -> A { .. }
}
```

mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
...
+}
+
+struct EdgeHubResultCodeVisitor;
+impl<'de> serde::de::Visitor<'de> for EdgeHubResultCodeVisitor {
+    type Value = EdgeHubResultCode;
+    fn expecting(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("200 or 403")
+    }
+
+    impl_for!(visit_i32, i32);
```
dmolokanov, line 133: I think we can use existing solution https://github.com/dtolnay/serde-repr

mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    use base64::decode;
+    use matches::assert_matches;
+    use mockito::{mock, Matcher};
+
+    use crate::auth::{AuthId, Authenticator, Certificate, Credentials};
+    use crate::edgehub_auth::EdgeHubAuthenticator;
+    use crate::Error;
+
+    const CERT: &str = "MIIBLjCB1AIJAOTg4Zxl8B7jMAoGCCqGSM49BAMCMB8xHTAbBgNVBAMMFFRodW1i\
```
dmolokanov, line 156: I hope it's not really cert and url below don't exist. For unit tests you can use fake values, right?

mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
...
+        .with_body(r#"{"result": 200, "identity":"vikauthtest.azure-devices.net/ca_root","version": "2020-04-20"}"#)
+        .create();
+
+        let authenticator = EdgeHubAuthenticator {
+            url: mockito::server_url() + "/authenticate/",
+        };
+        let cert_as_string = CERT.to_string();
+        let raw_cert_result = decode(cert_as_string);
+        let raw_cert = raw_cert_result.unwrap();
+        let cert: Certificate = Certificate::from(raw_cert);
```
dmolokanov, line 177: if you use from you dont need to put type for definition
```suggestion
        let cert = Certificate::from(raw_cert);
```

mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
...
+        let _m = mock("POST", "/authenticate/")
+        .with_status(200)
+        .with_body(r#"{"result": 200, "identity":"vikauthtest.azure-devices.net/ca_root","version": "2020-04-20"}"#)
+        .create();
+
+        let authenticator = EdgeHubAuthenticator {
+            url: mockito::server_url() + "/authenticate/",
+        };
+        let cert_as_string = CERT.to_string();
+        let raw_cert_result = decode(cert_as_string);
```
dmolokanov, line 175:
```suggestion
        let raw_cert = base64::decode(CERT.to_string()).expect("cert");
```

mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
...
+    async fn auth_request(
+        username: &str,
+        credentials: Credentials,
+    ) -> Result<Option<AuthId>, Error> {
+        let authenticator = EdgeHubAuthenticator {
+            url: mockito::server_url() + "/authenticate/",
+        };
+
+        authenticator
+            .authenticate(Some(username.to_string()), credentials)
```
dmolokanov, line 363: Usually I put method under test in the test method body and not hide it in some other function, and extract setup step in the separate functions instead.

mqtt/mqtt-broker/src/error.rs
```diff
@@ -9,6 +9,18 @@ pub struct Error {
 
 #[derive(Debug, Fail, PartialEq)]
 pub enum ErrorKind {
+    #[fail(display = "Authentication: failed to execute.")]
+    AuthExecution,
```
dmolokanov, line 13: It's better to move this as a inner error for Auth error.
`Error::Authenticate(AuthenticateError::Execution(...))`

mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
...
+            EdgeHubResultCode::Unauthenticated => Ok(None),
+        }
+    }
+}
+
+macro_rules! impl_for{
+    ($fun_name:ident, $typ:ident) => {
+        fn $fun_name<E: serde::de::Error>(self, v: $typ) -> Result<EdgeHubResultCode, E> {
+            match v {
+                200 => Ok(EdgeHubResultCode::Authenticated),
```
lfitchett, line 118: Is there a way to avoid double defining this relation? The enum already defines Authenticated as 200 and Unauthenticated as 403, can you use that?

mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
...
+#[derive(Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct EdgeHubAuthResponse {
+    version: String,
+    result: EdgeHubResultCode,
+    identity: Option<String>,
+}
+
+impl EdgeHubAuthResponse {
+    fn into_auth_id(self) -> Result<Option<AuthId>, Error> {
```
vadim-kovalyov, line 102: why not also implement From? were there any issues with that?

mqtt/mqtt-broker/src/edgehub_auth.rs
```diff
@@ -0,0 +1,370 @@
...
+    username: Option<String>,
+    password: Option<String>,
+    certificate: Option<String>,
+    certificate_chain: Option<Vec<String>>,
+}
+
+impl From<(Option<String>, Credentials)> for EdgeHubAuthRequest {
+    fn from(item: (Option<String>, Credentials)) -> Self {
+        let (password, certificate) = match item.1 {
+            Credentials::Password(p) => (p, None),
```
vadim-kovalyov, line 72: as far as I can see, this line will always match. Maybe you wanted `Credentials::Password(**Some(p)**) => (p, None),`

PR #2917, opened 2020-05-04
Prepare for release 1.0.9

yophilav:
Prepare for release 1.0.9
* Update [Windows IoT Core ARM32 version compatibility issue](https://support.microsoft.com/en-us/help/4542617/you-might-encounter-issues-when-using-windows-server-containers-with-t) to OS version 10.0.17763.1158 ([3f42583](https://github.com/Azure/iotedge/commit/3f425836720c07b522da19bf582351e1afba4bf6))
* Update [Windows Server 2019 AMD64 version compatibility issue](https://support.microsoft.com/en-us/help/4542617/you-might-encounter-issues-when-using-windows-server-containers-with-t) to tag version 2.1.17-nanoserver-1809 ([9552413](https://github.com/Azure/iotedge/commit/9552413925b370abb74ac6a35ebf54556842fb3f))
* Hotfix [Linux AMD64 security vulnerability](https://people.canonical.com/~ubuntu-security/cve/2020/CVE-2020-1967.html) by updating the dotnet core runtime image to use 2.1.17-alpine3.11 ([9c3d240](https://github.com/Azure/iotedge/commit/9c3d240eb733ae5d6b3fc3dd1a5a6b2b94275b96))


edgelet/contrib/config/windows/config.yaml
```diff
@@ -160,7 +160,7 @@ agent:
   type: "docker"
   env: {}
   config:
-    image: "mcr.microsoft.com/azureiotedge-agent:1.0.9"
+    image: "mcr.microsoft.com/azureiotedge-agent:1.0.9.1"
```
philipktlin, line 163: @varunpuranik I found that in release/1.0.8, we specify exact version as release in Windows config yaml, but we used 1.0 for Linux config yaml.  Do you know why?

CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
+# 1.0.9.1 (2020-05-04)
```
and-rewsmith, line 1: Are we releasing this today?

yophilav: I'm starting the process today. 
Since it's May 4th I'd like to take the opportunity. 🛰️

and-rewsmith: Haha ok. I think we should change it to be accurate though when we know the release date

edgelet/contrib/config/windows/config.yaml
dylanbronson: To me it makes sense that for RC's we should be explicit about it, but for releases that will be tagged with 1.0 (aka automatically downloaded when we ask for 1.0), then we shouldn't need to add it explicitly.

So I think it should "1.0" in this case

and-rewsmith: @philipktlin I asked this question to arnav a long long time ago (like October), and I think he told me that the linux one didn't get used so no one kept it up to date. 

@dylanbronson Isn't what you are referring to a part of the images release job? This information isn't contained in the yamls, unless I am misunderstanding something.

dylanbronson: @and-rewsmith I was under the impression it worked like this: when we have the image set to 1.0, we expect it to automatically pull the latest version of the image (whatever is tagged with 1.0). Since we will tag 1.0.9.1 with 1.0 (in the release job), we expect 1.0.9.1 to be pulled down when we ask for 1.0.

So having 1.0 or 1.0.9.1 explicitly in this config.yaml should pull down the 1.0.9.1 image once 1.0.9.1 is released. 

If you install Edge through normal installation process on Linux right now, the config.yaml will have "mcr.microsoft.com/azureiotedge-agent:1.0"
If you do the same but on Windows, the config.yaml will have  "mcr.microsoft.com/azureiotedge-agent:1.0.9"

Both will pull 1.0.9, and I think it's better to just have 1.0, since that signals "latest," but I don't think it matters too much.

Unless there's some reason that I don't know about that having 1.0 for Windows doesn't work like that

philipktlin: Refer to @and-rewsmith 's comment "linux one didn't get used so no one kept it up to date". 
Now I understand.  To explain more, currently when upgrade we expect customers will keep existing config.yaml.  Even we release a config.yaml with latest image tag, no one will use it.

@dylanbronson your description is correct for new installation.  If I installed iotedged on my linux device right after 1.0.8.5 is released.  Therefore my linux device is currently running iotedged version 1.0.8 and 1.0 image tag in config.yaml which installed docker image is version 1.0.8.5.  Now version 1.0.9 is released.  Customer upgraded iotedged to 1.0.9 and 1.0 image tag in config.yaml (customer generally will keep config.yaml).  After upgrade and start up iotedged, it will pull edge agent image tag 1.0 (actual version is 1.0.8.5) and it exists as the docker image of version 1.0.8.5 is downloaded by 1.0 tag before.  Therefore it will try to create a container with existing image.  This may fail refer to backward compatibility (https://docs.microsoft.com/en-us/azure/iot-edge/support#releases).

For windows it is different as upgrade installation will require to reconfigure config.yaml.

CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
...
+## Edge Agent
+### Bug Fixes
+* Hotfix [Window IoT Core ARM32 security vulnerability](https://support.microsoft.com/en-us/help/4542617/you-might-encounter-issues-when-using-windows-server-containers-with-t) by updating the base operating system images to version 10.0.17763.1158 ([3f42583](https://github.com/Azure/iotedge/commit/3f425836720c07b522da19bf582351e1afba4bf6))
```
varunpuranik, line 4: Just say "Update base image to xx to address Windows security issue"

edgelet/contrib/config/windows/config.yaml
dylanbronson: @philipktlin and I talked offline and did a manual test. When a new docker image is pushed to docker under the same tag (1.0 in this case), the hash will change and docker will recognize this. So when a docker container is removed, the new image will be pulled down, and the image will be updated correctly. Docker will **not** try to to create the container with the existing image, it will instead go and grab the updated image.

This means it looks like it should be 1.0 here, not 1.0.9.1 explicitly. For RC's it should be explicit, but for anything that will be tagged with 1.0, we should have 1.0 in the config.yaml.

@arsing @myagley 
Please correct us if we are wrong in any of this understanding.

arsing: >For windows it is different as upgrade installation will require to reconfigure config.yaml.

This is not true.

>So when a docker container is removed

Containers aren't removed as part of running `Update-IoTEdge`

arsing: But anyway, yes you can make this `1.0` too, since it'll only get used for new installs just like with Linux packages. That is, in both Windows and Linux, updating the package leaves the user's existing `config.yaml` untouched, so there's no point bumping it in every release.

edgelet/contrib/config/windows/config.yaml
```diff
@@ -160,7 +160,7 @@ agent:
   type: "docker"
   env: {}
   config:
-    image: "mcr.microsoft.com/azureiotedge-agent:1.0.9"
+    image: "mcr.microsoft.com/azureiotedge-agent:1.0"
```
varunpuranik, line 163: Did we test this? Might be better to keep this 1.0.9.1 (to avoid unnecessary changes)

philipktlin: Dylan did the test with me and it shows that a new container will get the latest version.  Also Edge agent will re-deploy if it detects different version registered in azure portal.  I think it should be safe to make this change.

PR #2918, opened 2020-05-05
[mqtt] Suspend sender task not block a thread

dmolokanov:
Performance tests showed that we have performance degradation introduced with #2842 PR. This changed introduced `crossbeam_channel` dependency that implements sync blocking operations for both sender and receiver parts.
Synchronous blocking operations allows broker to run in separate native thread with receiver. Under a high-load when broker channel got filled with events and **all threads** in the tokio thread-pool got blocked on `channel.send()` operation. 

This PR reverts some changes introduced #2842 for work with broker channel. It uses `tokio::sync::mpsc` channel instead that blocks **tasks** that run on tokio thread pool instead of whole thread. Broker main loop is spawned as a separate task but all operations within broker iteration and sync now.

Performance tests show improvements with new changes.

### DIFF
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -36,7 +36,7 @@ pub struct Broker<N, Z> {
     authorizer: Z,
 
     #[cfg(feature = "__internal_broker_callbacks")]
-    pub on_publish: Option<Sender<std::time::Duration>>,
+    pub on_publish: Option<tokio::sync::mpsc::UnboundedSender<std::time::Duration>>,
 }
 
 impl<N, Z> Broker<N, Z>
@@ -1001,10 +1001,7 @@ pub struct BrokerHandle(Sender<Message>);
 
 impl BrokerHandle {
     pub async fn send(&mut self, message: Message) -> Result<(), Error> {
-        self.0
-            .send(message)
-            .await
-            .map_err(|e| Error::SendBrokerMessage(e.into()))
+        self.0.send(message).await.map_err(Error::SendBrokerMessage)
     }
 }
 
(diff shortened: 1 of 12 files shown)


mqtt/mqtt-broker/src/broker.rs
```diff
@@ -14,6 +13,8 @@ use crate::session::{ConnectedSession, Session, SessionState};
 use crate::{
     subscription::Subscription, AuthId, ClientEvent, ClientId, ConnReq, Error, Message, SystemEvent,
 };
+use futures::StreamExt;
```
vadim-kovalyov, line 16: nit: rearrange

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -35,7 +36,7 @@ pub struct Broker<N, Z> {
     authorizer: Z,
 
     #[cfg(feature = "__internal_broker_callbacks")]
-    pub on_publish: Option<Sender<std::time::Duration>>,
+    pub on_publish: Option<tokio::sync::mpsc::UnboundedSender<std::time::Duration>>,
```
vadim-kovalyov, line 39: nit: remove namespaces

mqtt/mqtt-broker/src/snapshot.rs
```diff
@@ -43,7 +42,7 @@ pub struct Snapshotter<P> {
 
 impl<P> Snapshotter<P> {
     pub fn new(persistor: P) -> Self {
-        let (sender, events) = crossbeam_channel::bounded(5);
+        let (sender, events) = mpsc::channel(1024);
```
vadim-kovalyov, line 45: nit: why 1024?

mqtt/mqtt-broker/src/broker.rs
dmolokanov: I did it intentionally because the code is hidden behind feature

mqtt/mqtt-broker/src/snapshot.rs
dmolokanov: copy paste :(

PR #2920, opened 2020-05-05
Fix windows e2e PQ tests

dylanbronson:
The PQ e2e tests fail for windows currently. 
This is because whenever the modules try to talk to each other, they are unable because they can't find the URI. 
When I was investigating, I tried Pinging from inside the docker container to see what the issue was. Simply by pinging the testresultcoordinator from inside the docker container kickstarted the container to be able to talk to the testresultcoordinator with its name. It probably resolved the hostname, and updated a cache. We've seen DNS caching issues on Windows before for hostnames of docker containers.

So I added a Ping workaround in the e2e tests that will ping TRC on windows before reporting any results. This makes the tests pass. Less than ideal, so we should eventually figure out the root cause of why DNS resolution isn't working correctly on Windows containers.

### DIFF
--- a/test/modules/Relayer/Program.cs
+++ b/test/modules/Relayer/Program.cs
@@ -6,6 +6,7 @@ namespace Relayer
     using System.Collections.Generic;
     using System.Linq;
     using System.Net;
+    using System.Net.NetworkInformation;
     using System.Runtime.CompilerServices;
     using System.Text;
     using System.Threading;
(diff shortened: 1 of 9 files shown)


test/modules/Relayer/Program.cs
```diff
@@ -155,6 +167,27 @@ static async Task<MessageResponse> ProcessAndSendMessageAsync(Message message, o
             return MessageResponse.Completed;
         }
 
+        static void SendPing()
+        {
+            Ping ping = new Ping();
+            string uri = Settings.Current.TestResultCoordinatorUrl;
+            string pingMe = uri.Substring(uri.LastIndexOf('/') + 1).Split(':')[0];
```
damonbarry, line 174: This is ok for test code, but probably better to let .NET do URI parsing.

```csharp
    var uri = new Uri(Settings.Current.TestResultCoordinatorUrl);
    string pingMe = uri.Host;
```

PR #2934, opened 2020-05-07
E2E tests - Add networkController to yaml and update schema version 

dylanbronson:
Adding networkController to the e2e test pipeline yaml. Also updating the schema version that the E2E tests use so that priorityQueue tests work in edgeHub.


test/Microsoft.Azure.Devices.Edge.Test.Common/config/HubModuleConfigBuilder.cs
```diff
@@ -15,7 +15,7 @@ public HubModuleConfigBuilder(Option<string> image, bool optimizeForPerformance)
             this.WithDesiredProperties(
                 new Dictionary<string, object>
                 {
-                    ["schemaVersion"] = "1.0",
+                    ["schemaVersion"] = "1.1",
```
and-rewsmith, line 18: So schema 1.1 supports priority?

dylanbronson: Yes, and 1.0 does not

PR #2938, opened 2020-05-09
[TEST-ONLY] Control group testing for base image update of 1.0.9

yophilav:
[TEST-ONLY] Control group testing for base image update of 1.0.9


edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/EdgeAgentDockerModuleTest.cs
```diff
@@ -121,7 +121,7 @@ public void MixedIModuleImplEqualsTest()
 
             var fullRuntimeModule = CreateEdgeAgentDockerRuntimeModule(new DockerConfig(
                 "Foo",
-                createOptions: "{\"ignore\": \"me\"}",
```
philipktlin, line 124: revert it.

test/modules/DirectMethodSender/DirectMethodCloudSender.cs
```diff
@@ -45,12 +45,7 @@ sealed class DirectMethodCloudSender : DirectMethodSenderBase
             ulong directMethodCount,
             CancellationToken none)
         {
-            TimeSpan timeouts = Settings.Current.DirectMethodDelay;
```
philipktlin, line 48: do we have a bug for this? if so, please update that bug with this proposed fix in what file and line number.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/EdgeAgentDockerModuleTest.cs
philipktlin: Oh.  master already have this change.  I don't understand why this change shows up?

PR #2945, opened 2020-05-11
[mqtt] added edgehub-based authentication

vipeller:


mqtt/mqtt-broker/src/auth/authentication.rs
```diff
@@ -53,36 +71,51 @@ pub struct AuthenticateError;
...
+    async fn authenticate(
+        &self,
+        _: Option<String>,
+        _: Credentials,
+    ) -> Result<Option<AuthId>, Self::Error> {
         Ok(None)
     }
 }
 
+impl Clone for DefaultAuthenticator {
```
lfitchett, line 87: can you not #[derive(Clone)] ?

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
```diff
@@ -0,0 +1,338 @@
...
+    type Error = Error;
+
+    async fn authenticate(
+        &self,
+        username: Option<String>,
+        credentials: Credentials,
+    ) -> Result<Option<AuthId>, Self::Error> {
+        self.authenticate_client((username, credentials).into())
+            .await
+            .map_err(Error::Authenticate)?
```
lfitchett, line 25: Error impls From<AuthenticationError>. You shouldn't need to map this, ? will call .into() for you

mqtt/mqtt-broker/src/error.rs
```diff
@@ -80,3 +83,19 @@ pub enum InitializeBrokerError {
...
 }
+
+/// Authentication errors.
+#[derive(Debug, Error)]
+pub enum AuthenticationError {
+    #[error("failed to execute.")]
+    Execution,
+
+    #[error("failed to send request.")]
+    SendRequest,
```
lfitchett, line 94: should you have this wrap the underlying reqwest error?

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
```diff
@@ -0,0 +1,338 @@
...
+
+        if !response.version.eq(API_VERSION) {
+            return Err(AuthenticationError::ApiVersion(response.version));
+        }
+
+        Ok(response)
+    }
+}
+
+impl Clone for EdgeHubAuthenticator {
```
lfitchett, line 59: use derive instead

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
```diff
@@ -0,0 +1,338 @@
+use async_trait::async_trait;
+use serde::{Deserialize, Serialize};
+use serde_repr::Deserialize_repr;
+
+use crate::auth::{AuthId, Authenticator, Credentials};
+use crate::{AuthenticationError, Error};
+
+const API_VERSION: &str = "2020-04-20";
+
+pub struct EdgeHubAuthenticator {
```
vadim-kovalyov, line 10: struct coments

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
```diff
@@ -0,0 +1,338 @@
...
+        if response.status() != reqwest::StatusCode::OK {
+            return Err(AuthenticationError::ProcessResponse);
+        }
+
+        let response = response
+            .json::<EdgeHubAuthResponse>()
+            .await
+            .map_err(|_| AuthenticationError::ProcessResponse)?;
+
+        if !response.version.eq(API_VERSION) {
```
vadim-kovalyov, line 51: just curious why this check?

mqtt/mqtt-broker/src/server.rs
```diff
@@ -163,14 +168,16 @@ where
     }
 }
 
-async fn incoming_task<A, F>(
+async fn incoming_task<A, F, N>(
     transport: TransportBuilder<A>,
     handle: BrokerHandle,
     mut shutdown_signal: F,
+    authenticator: N,
```
vadim-kovalyov, line 175: nit: you can pass a ref `&N` to avoid extra clone.

vipeller: this is how it works with the clones:
- at line 56 it creates its own clone for every transport , which the will be passed to their respective incoming_task (e.g. if there is 2 transports specified, now both incoming_task will have its own copy.
- then the given incoming_task has a loop around incoming connections, and every new connection has its own copy of authorizator.

I don't see how reference could work?

mqtt/mqtt-broker/src/auth/authentication.rs
```diff
@@ -13,14 +15,21 @@ pub enum Credentials {
...
 impl From<Vec<u8>> for Certificate {
     fn from(certificate: Vec<u8>) -> Self {
         Self(certificate)
     }
 }
 
 /// A trait to authenticate a MQTT client with given credentials.
-pub trait Authenticator {
+#[async_trait]
+pub trait Authenticator: Clone {
```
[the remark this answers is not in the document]
vipeller: I made this change, however I don't see the benefit to write the type spec at every place (3 at this moment) instead of bringing it with the trait via "inheritance"?

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
vipeller: I added version to the record structure, and the server also checks it (and sends it back). The check is for avoid and update at server side but not following at client side.

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
vipeller: Im not sure I understand:
- the map_err() is there so it wrap the incoming AuthenticationError (a separate enum to list authentication related errors) int an Error::Authenticate(e)
- the into() is there because "authenticate_client" returns an EdgeHubAuthResponse structure, which is the mirror of the json structure coming back in the http response, and it gets turned (by the into() ) into an Option<AuthId> which is the return type of the authenticators.

lfitchett: The `Authenticate(#[from] AuthenticationError)` will automatically make 
```
impl From<AuthenticationError> for Error {
    fn from(e: AuthenticationError) -> Error {
        Error::Authenticate(e)
}
```
This means you can just do 
```
self.authenticate_client((username, credentials).into())
            .await?
            .into()
```

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -226,28 +222,21 @@ where
...
-        let credentials = connreq.certificate().map_or(
-            Credentials::Basic(
-                connreq.connect().username.clone(),
-                connreq.connect().password.clone(),
-            ),
-            |certificate| Credentials::ClientCertificate(certificate.clone()),
-        );
-        let auth_id = match self.authenticator.authenticate(credentials) {
-            Ok(Some(auth_id)) => {
+        let auth_id = match connreq.auth_result() {
```
dmolokanov, line 225:
```suggestion
        let auth_id = match connreq.auth() {
```

mqtt/mqtt-broker/src/error.rs
```diff
@@ -80,3 +83,19 @@ pub enum InitializeBrokerError {
...
 }
+
+/// Authentication errors.
+#[derive(Debug, Error)]
+pub enum AuthenticationError {
+    #[error("failed to execute.")]
+    Execution,
+
+    #[error("failed to send request: {0}.")]
+    SendRequest(#[source] reqwest::Error),
```
dmolokanov, line 94:
```suggestion
    SendRequest(#[from] reqwest::Error),
```
I'll give you From reqwest::Error to AuthenticationError for free

mqtt/mqtt-broker/src/lib.rs
```diff
@@ -61,25 +61,32 @@ impl std::fmt::Display for ClientId {
...
+    Successful(Option<AuthId>),
+    Failed,
+}
+
 #[derive(Debug)]
 pub struct ConnReq {
     client_id: ClientId,
     connect: proto::Connect,
-    certificate: Option<Certificate>,
+    auth_result: AuthResult,
```
dmolokanov, line 75:
```suggestion
    auth: AuthResult,
```

mqtt/mqtt-broker/src/lib.rs
```diff
@@ -96,8 +103,8 @@ impl ConnReq {
         &self.handle
     }
 
-    pub fn certificate(&self) -> Option<&Certificate> {
-        self.certificate.as_ref()
+    pub fn auth_result(&self) -> &AuthResult {
```
dmolokanov, line 106:
```suggestion
    pub fn auth(&self) -> &AuthResult {
```

mqtt/mqtt-broker/src/server.rs
dmolokanov: we can probably wrap it with Arc and clone references?
```suggestion
    authenticator: Arc<N>,
```

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
```diff
@@ -0,0 +1,331 @@
...
+            .into()
+    }
+}
+
+impl EdgeHubAuthenticator {
+    async fn authenticate_client(
+        &self,
+        request: EdgeHubAuthRequest,
+    ) -> Result<EdgeHubAuthResponse, AuthenticationError> {
+        let response = reqwest::Client::new()
```
dmolokanov, line 36: Can we store http client as a field instead of url?

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
```diff
@@ -0,0 +1,331 @@
+use async_trait::async_trait;
```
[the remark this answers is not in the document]
myagley: :+1:

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
vipeller: that worked, cool!

mqtt/mqtt-broker/src/auth/authentication_edgehub.rs
```diff
@@ -0,0 +1,331 @@
...
+const API_VERSION: &str = "2020-04-20";
+
+#[derive(Clone)]
+pub struct EdgeHubAuthenticator {
+    url: String,
+}
+
+#[async_trait]
+impl Authenticator for EdgeHubAuthenticator {
+    type Error = Error;
```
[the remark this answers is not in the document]
vipeller: done

mqtt/mqtt-broker/src/auth/authentication.rs
```diff
@@ -30,17 +39,26 @@ pub trait Authenticator {
...
+        username: Option<String>,
+        credentials: Credentials,
+    ) -> Result<Option<AuthId>, Self::Error>;
 }
 
+#[async_trait]
 impl<F> Authenticator for F
 where
-    F: Fn(Credentials) -> Result<Option<AuthId>, AuthenticateError> + Sync,
+    F: Fn(Option<String>, Credentials) -> Result<Option<AuthId>, AuthenticateError> + Sync + Clone,
```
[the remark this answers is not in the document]
vipeller: Arc worked

vipeller:
obsolete, merged from another PR

PR #2946, opened 2020-05-12
Longhaul: Fix direct method status reporting and don't fail for old messages

and-rewsmith:
This PR contains two changes. 

First, there is a bug with the way we calculated whether direct methods were pass or fail. We would do a check to make sure that status code 0 happened less than 1% of all statuses. The problem with our logic is that StatusCode.Count returns the number of unique status codes. What we want is the count of each unique status code summed.

Secondly, we are failing tests when the analyzer gets backed up reading messages. I have tried to debug this problem and think there is a significant time investment until we get a fix. Therefore I think we should pass tests with this well-known issue so as not to mask potential new issues.


test/modules/TestAnalyzer/AggregateCloudOperationReport.cs
```diff
@@ -61,7 +61,8 @@ bool IsPassedHelper()
             // Github issue: https://github.com/Azure/iotedge/issues/681
             // We expect to get this status sometimes because of edgehub restarts, but if we receive too many we should fail the tests.
             // TODO: When the SDK allows edgehub to de-register from subscriptions and we make the fix in edgehub, then we can fail tests for any status code 0.
-            bool statusCodeZeroBelowThreshold = this.StatusCodes.Where(s => s.StatusCode.Equals("0")).Count() < ((double)this.StatusCodes.Count / 100);
+            double allStatusCount = this.StatusCodes.Select(s => s.Count).Sum();
```
philipktlin, line 64: why use double?
```suggestion
            int totalNumberOfResults = this.StatusCodes.Select(s => s.Count).Sum();
```

and-rewsmith: This is intentional in order to get the division in the next line to return a double. This is certainly not obvious though so I'll make it explicit.

test/modules/TestAnalyzer/ModuleMessagesReport.cs
```diff
@@ -42,7 +42,7 @@ public ModuleMessagesReport(string moduleId, StatusCode statusCode, long receive
 
         public IList<MissedMessagesDetails> MissedMessages { get; }
 
-        public bool IsPassed => this.StatusCode == StatusCode.AllMessages;
+        public bool IsPassed => this.StatusCode == StatusCode.AllMessages || this.StatusCode == StatusCode.OldMessages;
```
dylanbronson, line 45: Can we make a new status code, something like OldMessagesNoneMissing and OldMessagesWithMissing?

This will currently pass even with missing messages as long as they are also old messages. I think we should fail if there are missing messages, but succeed on Old messages as long as there are none missing

and-rewsmith: Oh wow great catch! I thought it checked for old messages after missing. Do we need a new status code, or can we just re-order the checks?

dylanbronson: Yep, ordering works even better

and-rewsmith: I'm glad we caught this before 1.0.10. This has the potential to completely mask message loss. Luckily I'm pretty confident about our prior releases as arm32 never showed message loss.

and-rewsmith: @dylanbronson Once I started refactoring I liked your idea of a custom status code for missing and old messages together. I implemented it this way.

philipktlin: Do we need status code?  If doing it again, I will store last result timestamp (maybe also threshold) and some counts.

and-rewsmith: If we didn't have a status code we would need similar logic in order to determine IsPassed for the report. I like the status code because it provides more information than just true / false.

philipktlin: In TestResultReporter https://github.com/Azure/iotedge/blob/a5d4eff8da25ab37b931fdd17ca1c79d419e7f44/test/modules/TestAnalyzer/TestResultReporter.cs#L105, it returns statusCode.OldMessages even missingCounter > 0.  This means we will pass the test even there is missing message.  Am I missing anything?

and-rewsmith: I'm confused. What you linked to is the current implementation in master. You are correct that this is not what we want, so we reordered the checks in this PR.

test/modules/TestAnalyzer/TestResultReporter.cs
```diff
@@ -101,16 +101,46 @@ static ModuleMessagesReport GetReceivedMessagesReport(string moduleId, double to
                 }
             }
 
-            // check if last message is older
-            if (DateTime.Compare(lastMessageDateTime.AddMilliseconds(toleranceInMilliseconds), endDateTime) < 0)
+            bool areMessagesMissing = missingCounter > 0;
```
philipktlin, line 104: As talked, suggest to move these new logic into a function and returns status code and status message.

test/modules/TestAnalyzer/TestResultReporter.cs
```diff
@@ -101,16 +101,46 @@ static ModuleMessagesReport GetReceivedMessagesReport(string moduleId, double to
...
+            bool areLatestMessagesTooOld = DateTime.Compare(lastMessageDateTime.AddMilliseconds(toleranceInMilliseconds), endDateTime) < 0;
+            string missingMessagesStatus = $"Missing messages: {missingCounter}. ";
```
philipktlin, line 106: Should remove the space at the end here; and add space when you construct status message below.

test/modules/TestAnalyzer/ModuleMessagesReport.cs
philipktlin: Sorry, I forget to refresh PR page and shows the old version.  You are good and comments are provided.

test/modules/TestAnalyzer/TestResultReporter.cs
```diff
@@ -101,16 +101,46 @@ static ModuleMessagesReport GetReceivedMessagesReport(string moduleId, double to
...
+            if (areLatestMessagesTooOld)
             {
                 Logger.LogInformation($"Module {moduleId}: last message datetime={lastMessageDateTime} and end datetime={endDateTime}");
-                return new ModuleMessagesReport(moduleId, StatusCode.OldMessages, totalMessagesCounter, $"Missing messages: {missingCounter}. No messages received for the past {toleranceInMilliseconds} milliseconds.", lastMessageDateTime, missedMessages, Settings.Current.TestInfo);
             }
 
-            return missingCounter > 0
-                ? new ModuleMessagesReport(moduleId, StatusCode.SkippedMessages, totalMessagesCounter, $"Missing messages: {missingCounter}.", lastMessageDateTime, missedMessages, Settings.Current.TestInfo)
-                : new ModuleMessagesReport(moduleId, StatusCode.AllMessages, totalMessagesCounter, "All messages received.", lastMessageDateTime, Settings.Current.TestInfo);
+            StatusCode statusCode;
```
philipktlin, line 120: General comments: We put more than 1 information in status code.  If we needed to consider 1 more information to determine status code, it will make status code assignment logic more complicated.  We go for this quick approach now as more changes are required in iotedge dashboard if we don't provide status code.  Good for now.

test/modules/TestAnalyzer/TestResultReporter.cs
```diff
@@ -101,16 +101,54 @@ static ModuleMessagesReport GetReceivedMessagesReport(string moduleId, double to
                 }
             }
 
-            // check if last message is older
-            if (DateTime.Compare(lastMessageDateTime.AddMilliseconds(toleranceInMilliseconds), endDateTime) < 0)
+            bool areMessagesMissing = missingCounter > 0;
+            bool areLatestMessagesTooOld = DateTime.Compare(lastMessageDateTime.AddMilliseconds(toleranceInMilliseconds), endDateTime) < 0;
```
philipktlin, line 105: nit: you are checking last message.  Should name it as "isLastMessageTooOld"?

test/modules/TestAnalyzer/TestResultReporter.cs
```diff
@@ -101,16 +101,54 @@ static ModuleMessagesReport GetReceivedMessagesReport(string moduleId, double to
...
-                ? new ModuleMessagesReport(moduleId, StatusCode.SkippedMessages, totalMessagesCounter, $"Missing messages: {missingCounter}.", lastMessageDateTime, missedMessages, Settings.Current.TestInfo)
-                : new ModuleMessagesReport(moduleId, StatusCode.AllMessages, totalMessagesCounter, "All messages received.", lastMessageDateTime, Settings.Current.TestInfo);
+            (StatusCode statusCode, string statusMessage) = GetStatus(areMessagesMissing, areLatestMessagesTooOld, missingCounter, toleranceInMilliseconds);
+
+            return new ModuleMessagesReport(moduleId, statusCode, totalMessagesCounter, statusMessage, lastMessageDateTime, missedMessages, Settings.Current.TestInfo);
+        }
+
+        static (StatusCode, string) GetStatus(bool areMessagesMissing, bool areLatestMessagesTooOld, long missingCounter, double toleranceInMilliseconds)
+        {
+            string missingMessagesStatus = $"Missing messages: {missingCounter}. ";
```
philipktlin, line 124: nit: remove last space character at the end and add space in below usage, for example, line 133.

PR #2949, opened 2020-05-12
PacketStream + few more granular integration tests

vadim-kovalyov:
In this PR I have:
- Added `PacketStream` - a simple wrapper on top of TcpStream + PacketCodec. I will allow us to test more granular scenarios
- Added few more tests.
- Some refactoring of existing integration tests.


mqtt/mqtt3/src/proto/packet.rs
```diff
@@ -170,11 +185,6 @@ impl PacketMeta for Connect {
         }
 
         let protocol_level = src.try_get_u8()?;
-        if protocol_level != crate::PROTOCOL_LEVEL {
```
vadim-kovalyov, line 173: While testing, I found a bug that violated [MQTT-3.1.2-2].
```
The Server MUST respond to the CONNECT Packet with a CONNACK return code 0x01 (unacceptable protocol level) and then disconnect the Client if the Protocol Level is not supported by the Server [MQTT-3.1.2-2].
```

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -284,11 +287,135 @@ where
...
+/// Used to control server lifetime during tests. Implements
+/// Drop to cleanup resources after every test.
+pub struct ServerHandle {
+    address: String,
+    shutdown: Option<Sender<()>>,
+    task: Option<JoinHandle<Result<BrokerState, Error>>>,
+}
+
+impl ServerHandle {
+    pub fn address(&self) -> String {
```
dmolokanov, line 388:
```suggestion
    pub fn address(&self) -> &str {
```
seems like it makes unneccessary string allocations..

mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -451,9 +566,181 @@ async fn overlapping_subscriptions() {
...
+        .authenticator(|_| Ok(Some(AuthId::Anonymous)))
+        .authorizer(|_| Ok(true))
+        .build();
+
+    let server_handle = common::start_server(broker);
+
+    let mut client = PacketStream::open(server_handle.address()).await;
+    client.send_packet(Packet::PingReq(PingReq)).await;
+
+    assert_matches!(client.next().await, None); // None means stream closed.
```
dmolokanov, line 583: You don't make match here and below, so you may want to use `assert_eq!` instead

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -284,11 +287,135 @@ where
...
+    pub async fn open(server_addr: impl ToSocketAddrs) -> Self {
+        // broker may not be available immediately in the test,
+        // so we'll try to connect for some time.
+        let mut result = TcpStream::connect(&server_addr).await;
+        let start_time = Instant::now();
+        while let Err(_) = result {
+            if start_time.elapsed() > *DEFAULT_TIMEOUT {
+                break;
+            }
+            result = TcpStream::connect(&server_addr).await;
```
lfitchett, line 313: How long does TcpStream::connect take? should you put like a 200ms sleep here to prevent it from spamming reconnect s?

mqtt/mqtt-broker/tests/common.rs
```diff
@@ -284,11 +287,135 @@ where
...
+        username: Option<String>,
+        password: Option<String>,
+    ) -> Self {
+        let mut client = Self::open(server_addr).await;
+        client
+            .send_connect(Connect {
+                username,
+                password,
+                client_id,
+                ..Default::default()
```
lfitchett, line 340: what is the  `..` doing? I've never seen this syntax before

mqtt/mqtt3/src/proto/packet.rs
```diff
@@ -137,6 +138,20 @@ pub struct Connect {
     pub protocol_level: u8,
 }
 
+impl Default for Connect {
```
dmolokanov, line 141: I don't think it is a good idea to add a default for proto::Connect only for testing purposes. Semantically connect packet suppose to have data that identifies a client.
Add the factory method in tests instead if you need to

mqtt/mqtt3/src/proto/packet.rs
dmolokanov: You just removed this check? Could you please point me to a unit test that covers this case?

mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -382,11 +427,81 @@ async fn offline_messages() {
...
+    let mut client_b = TestClientBuilder::new(server_handle.address())
+        .client_id(ClientId::IdWithCleanSession("mqtt-smoke-tests-b".into()))
+        .build();
+
+    client_b.publish_qos0(topic_a, "o qos 0", false).await;
+    client_b.publish_qos1(topic_b, "o qos 1", false).await;
+    client_b.publish_qos2(topic_c, "o qos 2", false).await;
+
+    // need to wait till all messages are processed.
+    tokio::time::delay_for(Duration::from_secs(1)).await;
```
dmolokanov, line 462: :(

mqtt/mqtt-broker/tests/smoke.rs
```diff
@@ -382,11 +427,81 @@ async fn offline_messages() {
...
+        .build();
+
+    let server_handle = common::start_server(broker);
+
+    let mut client_a = TestClientBuilder::new(server_handle.address())
+        .client_id(ClientId::IdWithExistingSession("mqtt-smoke-tests-a".into()))
+        .build();
+
+    // expects existing session.
+    assert_matches!(
```
dmolokanov, line 481:
```suggestion
    assert_eq!(
```
?

mqtt/mqtt-broker/tests/common.rs
dmolokanov: https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax

mqtt/mqtt3/src/proto/packet.rs
vadim-kovalyov: Test named `wrong_protocol_version_rejected` that found it.

mqtt/mqtt-broker/tests/common.rs
vadim-kovalyov: I thought about that, constant retry it doesn't hurt for tests, we just need to timeout at some point. But I can add the delay.

mqtt/mqtt3/src/proto/packet.rs
vadim-kovalyov: Agreed, will add the factory method.

mqtt/mqtt-broker/tests/common.rs
vadim-kovalyov: There are some complications with lifetime of &str for spawning broker thread. Will leave this as is.

mqtt/mqtt-broker/tests/common.rs
lfitchett: cool, good to know

PR #2950, opened 2020-05-12
[mqtt] Add versioned file format

lfitchett:
This wraps the consolidated file format in a version enum. We decided that compression and encoding would be handled by file extension, so this only encapsulates the brokerstate object.

### DIFF
--- a/mqtt/mqtt-broker/src/persist.rs
+++ b/mqtt/mqtt-broker/src/persist.rs
@@ -108,10 +108,29 @@ impl<F> FilePersistor<F> {
     }
 }
 
-#[derive(Clone, Debug, Default)]
-pub struct ConsolidatedStateFormat;
+#[derive(Deserialize, Serialize)]
+enum VersionedState {
+    V1(ConsolidatedState),
+}
 
-impl FileFormat for ConsolidatedStateFormat {
+impl From<BrokerState> for VersionedState {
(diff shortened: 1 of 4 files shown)


mqtt/mqtt-broker/src/persist.rs
```diff
@@ -571,7 +590,7 @@ pub(crate) mod tests {
     async fn filepersistor_smoketest() {
         let tmp_dir = TempDir::new().unwrap();
         let path = tmp_dir.path().to_owned();
-        let mut persistor = FilePersistor::new(path, ConsolidatedStateFormat::default());
+        let mut persistor = FilePersistor::new(path, VersionedFileFormat::default());
 
         persistor.store(BrokerState::default()).await.unwrap();
```
vadim-kovalyov, line 595: nit: btw, it is better to use some test data in BorkerState, instead of `default()`. When I played with versioned state, I had a peculiar issue where I was writing into a wrong file and reading from the old one. They both returned `default()` state, so the test would pass, but the logic was broken:) using some actual test data revealed the problem.

lfitchett: Yea, I think once we have a second version there needs to be a test added for reading an old version, But that can;t be made until there's 2 versions

lfitchett:
I think a test should be added when we have a second version to test with

PR #2953, opened 2020-05-13
[mqtt] Model-based tests to verify broker subscriptions management

dmolokanov:
Model-based tests contain a simplified model of MQTT broker that handles some MQTT packets. Randomly generated MQTT packets processed by both: broker and broker model. Later tests verify that the broker and its model processed packet in a similar way.

Currently, there is only on a test case that handles a randomly generated sequence of CONNECT, DISCONNECT, SUBSCRIBE, UNSUBSCRIBE packets with pseudo-random data. In the end, the test verifies that both the broker and its model contain the same number of sessions and same number of subscriptions.

### DIFF
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -103,6 +103,7 @@ where
         BrokerState { retained, sessions }
     }
 
+    #[cfg(any(test, feature = "proptest"))]
     pub fn clone_state(&self) -> BrokerState {
         let retained = self.retained.clone();
         let sessions = self
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -103,7 +103,27 @@ where
         BrokerState { retained, sessions }
     }
 
-    fn process_message(&mut self, client_id: ClientId, event: ClientEvent) -> Result<(), Error> {
+    pub fn clone_state(&self) -> BrokerState {
+        let retained = self.retained.clone();
+        let sessions = self
+            .sessions
+            .values()
+            .filter_map(|session| match session {
+                Session::Transient(session) => Some(session.state().clone()),
+                Session::Persistent(session) => Some(session.state().clone()),
+                Session::Offline(session) => Some(session.state().clone()),
+                _ => None,
+            })
+            .collect();
+
+        BrokerState { retained, sessions }
+    }
+
+    pub fn process_message(
+        &mut self,
+        client_id: ClientId,
+        event: ClientEvent,
+    ) -> Result<(), Error> {
         debug!("incoming: {:?}", event);
         let result = match event {
             ClientEvent::ConnReq(connreq) => self.process_connect(client_id, connreq),
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -1042,34 +1042,18 @@ pub(crate) mod tests {
     use bytes::Bytes;
     use futures_util::future::FutureExt;
     use matches::assert_matches;
-    use proptest::collection::{hash_map, vec};
-    use proptest::prelude::*;
-    use tokio::sync::mpsc::error::TryRecvError;
-    use tokio::sync::mpsc::{self, UnboundedReceiver};
+    use tokio::sync::mpsc::{self, error::TryRecvError, UnboundedReceiver};
     use uuid::Uuid;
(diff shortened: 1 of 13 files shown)


mqtt/mqtt-broker/src/broker.rs
```diff
@@ -103,7 +103,27 @@ where
         BrokerState { retained, sessions }
     }
 
-    fn process_message(&mut self, client_id: ClientId, event: ClientEvent) -> Result<(), Error> {
+    pub fn clone_state(&self) -> BrokerState {
```
vadim-kovalyov, line 106: do we want this in release bits? can we guard this with `#[cfg(any(test, feature = "proptest"))]`

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -523,9 +523,10 @@ pub(crate) mod tests {
     use proptest::prelude::*;
     use tempfile::TempDir;
 
-    use crate::broker::{tests::arb_broker_state, BrokerState};
-    use crate::persist::{
-        ConsolidatedState, ConsolidatedStateFormat, FileFormat, FilePersistor, Persist,
+    use crate::{
```
vadim-kovalyov, line 526: :+1:

mqtt/mqtt-broker/src/proptest.rs
```diff
@@ -0,0 +1,268 @@
...
+    collection::{hash_map, hash_set, vec, vec_deque},
+    num,
+    prelude::*,
+};
+
+use crate::{
+    session::{IdentifiersInUse, PacketIdentifiers},
+    BrokerState, ClientId, Publish, Segment, SessionState, Subscription, TopicFilter,
+};
+use bytes::Bytes;
```
vadim-kovalyov, line 15: nit: rearrange :)

mqtt/mqtt-broker/src/proptest.rs
```diff
@@ -0,0 +1,268 @@
...
+prop_compose! {
+    pub fn arb_broker_state()(
+        retained in hash_map(arb_topic(), arb_publication(), 0..20),
+        sessions in vec(arb_session_state(), 0..10),
+    ) -> BrokerState {
+        BrokerState::new(retained, sessions)
+    }
+}
+
+prop_compose! {
```
vadim-kovalyov, line 26: :+1:  nice to have all prop_compose in a separate module!

mqtt/mqtt-broker/tests/broker_model.rs
```diff
@@ -0,0 +1,240 @@
...
+    tokio::runtime::Builder::new()
+        .basic_scheduler()
+        .enable_all()
+        .build()
+        .unwrap()
+        .block_on(test_broker_manages_sessions(events));
+    }
+}
+
+async fn test_broker_manages_sessions(events: impl IntoIterator<Item = BrokerEvent>) {
```
vadim-kovalyov, line 27: I think this module would benefit from code comments. It would be easier to follow for those who new or unfamiliar with model-base testing. Even for those who is familiar with it, it would be easier to understand how your model works and how events flow in the model.

mqtt/mqtt-broker/tests/broker_model.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut broker = BrokerBuilder::default()
+        .authenticator(|_| Ok(Some(AuthId::Anonymous)))
+        .authorizer(|_| Ok(true))
+        .build();
+
+    let mut model = BrokerModel::default();
+
+    let mut clients = Vec::new();
```
vadim-kovalyov, line 35: I'm not sure I understand `clients` vector and what it is used for? Is it just to keep receivers from dropping?

mqtt/mqtt-broker/src/broker.rs
dmolokanov: Something is telling me that we may want to use this method in integration tests maybe. Not 100% sure though

mqtt/mqtt-broker/src/proptest.rs
dmolokanov: Yeah, basically I borrowed approach from https://github.com/bodil/im-rs crate

mqtt/mqtt-broker/tests/broker_model.rs
dmolokanov: yep, you're right. when we include message dispatching it will collect messages for each client.

PR #2956, opened 2020-05-13
Fixing arm32 new e2e tests

dylanbronson:
Previously arm32 new e2e tests assume that the repo is already there. Somehow our pi got into a bad state, and the git command failed, failing the whole pipeline. 

This change changes the command to do the same thing that we wanted to do (delete any file that was owned by root in the current folder) but without making any assumptions.


builds/e2e/e2e.yaml
```diff
@@ -35,8 +35,9 @@ jobs:
     timeoutInMinutes: 120
 
     steps:
-    - pwsh: |
-        sudo git clean -ffdx
+    - task: CmdLine@2
+      inputs:
+        script: 'sudo find . -type f -group root -execdir rm {} + && sudo find . -type d -group root -execdir rm -r {} \;'
```
damonbarry, line 40: I worry that the `&& ...` part of the command would be misinterpreted by bash (part of `execdir` rather than a new command)...you confirmed that this works as expected?

dylanbronson: Running this fixed the pipeline, as it was stuck in a bad state. But now it fails with a few "No such file or directory found" errors. I'll investigate

dylanbronson: It fails when it can't find anything, but we should allow that to happen, since there's a possibility that there is nothing

So I redirected stderr so that it will ignore error for this step
I wanted to do this by setting Azure pipeline's failOnStdErr: false (https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/command-line?view=azure-devops&tabs=yaml), but that didn't work... Not sure why not, though

damonbarry: I think `failOnStdErr` is an _additional_ way to get something to fail in the pipeline. The inverse is not true: you can't set it to `false` to ignore failures. To do that, just append ` || true` to the end of the command whose exit code you want to ignore.

I can't figure out why that `sudo find . -type d -group root -execdir rm -r {} \;` is failing...it is actually removing directories, then reporting it can't find them (because it removed them)...weird. But this seems to do the same thing without failing:

```
sudo find . -type d -group root -print0 | xargs -0 /bin/rm -rf
```

PR #2957, opened 2020-05-13
Longhaul / Stress / E2E / Connectivity: Mitigate docker network bug by restarting docker

and-rewsmith:
There is a moby bug affecting longhaul and connectivity tests where a container reference fails to get removed from the azure-iotedge network stop the iotedge service, reinstall, and start back up. This results in some containers not being able to start up on the next deployment.

This PR mitigates this Moby bug by restarting docker every time the tests start.


dylanbronson:
Can you leave a comment there, letting people know we have to do this because of the moby bug?

PR #2965, opened 2020-05-15
[mqtt] Moving authentication to connection from the broker

vipeller:
Moving the authentication step to connection - this calls the authentication module and stores the authentication result in the ConnReq event. The broker has the same logic as before, except it uses the already acquired result, instead of calling the authentication module directly

### DIFF
--- a/mqtt/mqtt-broker/src/server.rs
+++ b/mqtt/mqtt-broker/src/server.rs
@@ -1,4 +1,5 @@
 use std::future::Future;
+use std::sync::Arc;
 
 use futures_util::future::{self, Either, FutureExt};
 use futures_util::pin_mut;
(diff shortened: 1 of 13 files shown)


mqtt/mqtt-broker/src/server.rs
```diff
@@ -10,46 +11,54 @@ use tracing_futures::Instrument;
...
         F: Future<Output = ()> + Unpin,
         I: IntoIterator<Item = TransportBuilder<A>>,
+        N: Authenticator + Send + Sync + 'static,
+        N::Error: Into<AuthenticationError>,
     {
         let Server { broker } = self;
         let mut handle = broker.handle();
         let broker_task = tokio::spawn(broker.run());
 
+        let authenticator = Arc::new(authenticator);
```
vadim-kovalyov, line 48: just a question: It might be a good idea to use Arc instead of cloning. Just to confirm - when we actually call `authenticator.authenticate()`, it is not a blocking call, right?

PR #2966, opened 2020-05-15
Fix new E2E pipeline

dylanbronson:
A previous change in EdgeAgent (https://github.com/Azure/iotedge/commit/11d4e3a66f1986aa47c4ee8f4fdb70550a435d3a) allowed EdgeAgent to be restarted when the config is updated. This, however, broke the new e2e test pipeline.

The change in EA stores environment variables as labels on the new EA container before running it and deleting the old container. However, this change requires that both versions of EA know about the change to work.

So EA version with the change (let's call it change x) is incompatible with an EA version without change x.

In the new e2e pipeline, we start up bootstrapping 1.0 -> which pulls 1.0.9.1 currently. 1.0.9.1 does not have change x, and will not be receiving change x.  But in development and in the pipeline, the bootstrap pulls down the latest version of EA in master, so this test will fail since these are incompatible. 

This PR defines a new bootstrap image for EA for the pipeline. This is temporary. Once 1.0 knows about the change OR once we make the EA image backwards compatible, we can remove this fix.

This PR also contains a change to the E2E module comparison logic when a module gets started up. If the CreateOptions is filled out in an image with the prefix 'net.azure-devices.edge' then all of the CreateOptions will be ignored in comparison of reported and desired properties. Otherwise the test will never verify that EA module got deployed, since EA (after change x) "magically" inserts environment variables into the Labels portion of CreateOptions. The test doesn't specify these to be filled out in DesiredProperties, so the test doesn't know about them. 

Ignoring the entirety of CreateOptions might be too crude, so if anyone has an idea to make it more specific we can change that (maybe we could make it for just EA and hardcode EA name in the comparison to check first). I think it is fine for now, though, because that prefix shouldn't get into labels any other way, so no other module should run into this issue. And EA never changes CreateOptions in any of the new E2E tests, so it should be safe for now.


test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
```diff
@@ -166,7 +166,9 @@ async Task WaitForStatusAsync()
                             if (v.Path.EndsWith("settings.createOptions"))
                             {
                                 // normalize stringized JSON inside "createOptions"
-                                v.Value = JObject.Parse((string)v.Value).ToString(Formatting.None);
+                                // But ignore the value if it starts with 'net.azure-devices.edge'
+                                string value = JObject.Parse((string)v.Value).ToString(Formatting.None);
+                                v.Value = value.Contains("net.azure-devices.edge") ? "{}" : value;
```
damonbarry, line 171: Yeah this should work but it is pretty coarse. 😉 

Instead of filtering certain values _out of_ the _actual_ twin json, what if we insert it _into_ the _expected_ json? I mean, we know exactly what values we expect to find there.

In `EdgeConfiguration.cs`, in the method `Create`, you could replace the following line:

```csharp
            var reported = new Dictionary<string, object>
            {
                ["systemModules"] = desired
                    .Value<JObject>("systemModules")
                    .Children<JProperty>()
                    .ToDictionary(
                        p => p.Name,
-                        p => CreateExpectedModuleConfig((JObject)p.Value))
+                        p => p.Name == ModuleName.EdgeAgent
+                            ? CreateExpectedAgentModuleConfig((JObject)p.Value)
+                            : CreateExpectedModuleConfig((JObject)p.Value))
            };
```

...then create the new method `CreateExpectedAgentModuleConfig` that does this (untested):

```csharp
        static object CreateExpectedAgentModuleConfig(JObject source)
        {
            JToken createOptions = source.SelectToken("settings.createOptions") ?? new JObject();
            JToken env = source.SelectToken("env") ?? new JObject();

            string createOptionsLabel = JsonConvert.SerializeObject(createOptions);

            createOptions.Value<JObject>().TryAdd("Labels", new JObject());
            JObject labels = createOptions.Value<JObject>("Labels");
            labels.TryAdd("net.azure-devices.edge.owner", new JValue("Microsoft.Azure.Devices.Edge.Agent"));
            labels.TryAdd("net.azure-devices.edge.create-options", new JValue(createOptionsLabel));
            labels.TryAdd("net.azure-devices.edge.env", env);

            return CreateExpectedModuleConfig(source);
        }
```

test/Microsoft.Azure.Devices.Edge.Test.Common/DaemonConfiguration.cs
```diff
@@ -12,11 +12,24 @@ public class DaemonConfiguration
         readonly string configYamlFile;
         readonly YamlDocument config;
 
-        public DaemonConfiguration(string configYamlFile)
+        public DaemonConfiguration(string configYamlFile, (string image, string address, string username, string password) bootstrapAgentInfo)
```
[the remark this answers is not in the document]
damonbarry: This also means that `CreateEdgeDaemon` and the constructors for linux and windows `EdgeDaemon` classes would take two new params instead of one, and both would be optional with defaults.

test/README.md
```diff
@@ -45,6 +45,7 @@ The end-to-end tests take several parameters, which they expect to find in a fil
 | `tempFilterFuncImage` | * | Azure temperature filter function to be used. Required when running the test 'TempFilterFunc', ignored otherwise.|
 | `tempFilterImage` | * | Docker image to pull/use for the temperature filter module. Required when running the test 'TempFilter', ignored otherwise.|
 | `tempSensorImage` || Docker image to pull/use for the temperature sensor module (see the test 'TempSensor'). If not given, `mcr.microsoft.com/azureiotedge-simulated-temperature-sensor:1.0` is used.|
+| `edgeAgentBootstrapImage` || Docker image to pull/use for the initial startup of edgeAgent. This is a temporary parameter - only here now because the 1.0 version is incompatible with master's edgeAgent. If you want to specify your own, the container registry used is the first registry in the list of registries given by the 'registries' parameter|'
```
damonbarry, line 48: I'd make it clear this is the Edge Agent image used in config.yaml, as opposed to the image given to deployments.

damonbarry, line 48: There's a note about config.yaml in the description of `edgeAgentImage` above that needs to be corrected, now that you've added `edgeAgentBootstrapImage`.

damonbarry, line 48: Also, "incompatible" is a little misleading since the two versions of Edge Agent aren't incompatible with each other from a production perspective. More accurately, the _way the test framework verifies deployments_ depends on behavior that was introduced into Edge Agent during development of 1.0.10, so we can't use 1.0 images until 1.0.10 is released.

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
```diff
@@ -66,7 +66,9 @@ public static EdgeConfiguration Create(string deviceId, IEnumerable<ModuleConfig
                             .Children<JProperty>()
                             .ToDictionary(
                                 p => p.Name,
-                                p => CreateExpectedModuleConfig((JObject)p.Value))
+                                p => p.Name == ModuleName.EdgeAgent
+                                ? CreateExpectedAgentModuleConfig((JObject)p.Value)
+                                : CreateExpectedModuleConfig((JObject)p.Value))
```
damonbarry, line 71: Nit: indent

test/README.md
dylanbronson: Changed it around to make it clearer

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
dylanbronson: Thanks for the suggestion! Had to alter a bit to get it to work, but it works

test/Microsoft.Azure.Devices.Edge.Test.Common/config/EdgeConfiguration.cs
```diff
@@ -103,6 +105,32 @@ static object CreateExpectedModuleConfig(JObject source)
...
+            string createOptionsLabel = JsonConvert.SerializeObject(createOptions);
+
+            createOptions.Value<JObject>().TryAdd("Labels", new JObject());
+            JObject labels = createOptions.Value<JObject>("Labels");
+            labels.TryAdd("net.azure-devices.edge.create-options", new JValue(createOptionsLabel));
+            labels.TryAdd("net.azure-devices.edge.env", JsonConvert.SerializeObject(env));
+            labels.TryAdd("net.azure-devices.edge.owner", new JValue("Microsoft.Azure.Devices.Edge.Agent"));
+            createOptions["Labels"] = labels;
+            source["settings"]["createOptions"] = createOptions;
+            if (source["settings"] != null)
```
damonbarry, line 122: `source["settings"]` will never be `null` here; you used it in the previous line to access `["createOptions"]`. 😉 

Since C# uses references you shouldn't need to update parent objects with pointers to their children. But it looks like my suggested code didn't handle the case where `source` doesn't have "settings" and/or "createOptions". a cleaner way to handle that is to use the TryAdd/Value pattern that you're already using for `labels`:

```csharp
            source.TryAdd("settings", new JObject());
            JObject settings = source.Value<JObject>("settings");

            settings.TryAdd("createOptions", new JObject());
            JObject createOptions = settings.Value<JObject>("createOptions");
            string createOptionsLabel = JsonConvert.SerializeObject(createOptions);

            createOptions.TryAdd("Labels", new JObject());
            JObject labels = createOptions.Value<JObject>("Labels");

            JToken env = source.SelectToken("env") ?? new JObject();

            labels.TryAdd("net.azure-devices.edge.create-options", new JValue(createOptionsLabel));
            labels.TryAdd("net.azure-devices.edge.env", JsonConvert.SerializeObject(env));
            labels.TryAdd("net.azure-devices.edge.owner", new JValue("Microsoft.Azure.Devices.Edge.Agent"));

            return CreateExpectedModuleConfig(source);
```

PR #2967, opened 2020-05-15
Recognize changes to bootstrap agent config

damonbarry:
With recent changes (#2889), you can now deploy changes to Edge Agent's config (e.g., modify environment variables or Docker create options) and it will detect these changes and update itself accordingly. But there's still one corner case where changes aren't detected: when the "bootstrap" agent container (the one created independently by iotedged when it starts up) is running and a deployment comes in with the exact same agent image, but with different environment variables or create options. That's because the bootstrap container doesn't have the labels we use (`net.azure-devices.edge.create-options`, `net.azure-devices.edge.env`) to detect changes; right now agent containers only get those labels when their creation was initiated by Edge Agent.

This commit updates iotedged so that it, too, will add the needed labels to the agent container it creates.


edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -337,6 +337,18 @@ impl ModuleRuntime for DockerModuleRuntime {
                     .unwrap_or_else(HashMap::new);
                 labels.insert(LABEL_KEY.to_string(), LABEL_VALUE.to_string());
```
damonbarry, line 339: Doing this in `create` (which gets called for all containers) isn't great IMO. A better option would have been the watchdog, which has code dedicated solely to creating Edge Agent in the bootstrap scenario. But unfortunately that code doesn't know about Docker; this is the first (or best) place I could find that actually knows about Docker and things like container labels.

damonbarry: If anyone has a better idea I'd love to hear it...

arsing: Just like https://github.com/Azure/iotedge/pull/2305#issuecomment-574785186 , can you not do it in `init_agent_spec` ?

damonbarry: 🙄 I completely forgot about that PR. Yes, I bet that will work nicely...

damonbarry:
Alright @arsing, it looks like the `diff_with_*` unit tests are passing now.

PR #2971, opened 2020-05-17
Add more logs in C2D message receiver for investigation

philipktlin:
From connectivity test log, we got NRE in CloudToDeviceMessageReceiver. Therefore adding code to detect NRE and write warning log instead.

[2020-05-07 19:07:52.574 ERR] Error occurred while receiving message.
System.NullReferenceException: Object reference not set to an instance of an object.
   at CloudToDeviceMessageTester.CloudToDeviceMessageReceiver.StartAsync(CancellationToken ct) in /home/vsts/work/1/s/test/modules/CloudToDeviceMessageTester/CloudToDeviceMessageReceiver.cs:line 89


test/modules/CloudToDeviceMessageTester/CloudToDeviceMessageReceiver.cs
```diff
@@ -86,10 +88,28 @@ public async Task StartAsync(CancellationToken ct)
...
                     {
                         Message message = await this.deviceClient.ReceiveAsync();
+
+                        if (message == null)
+                        {
+                            this.logger.LogWarning("Received message is null");
+                            continue;
+                        }
+
+                        if (!(message?.Properties.ContainsKey(TestConstants.Message.SequenceNumberPropertyName) ?? false) ||
```
yophilav, line 98: Since the message cannot be null here, I don't think we need the `?.`

philipktlin: It is possible to be null as operation timeout happens or edgeHub restart.

PR #2976, opened 2020-05-19
Improve Twin Tester

philipktlin:
Refactor TwinState to use AtomicLong type to allow access from multiple threads (1 writer and N readers).  Update logic to skip callback validation due to edge hub restart.


test/modules/TwinTester/DesiredPropertiesValidator.cs
```diff
@@ -122,24 +114,29 @@ async Task RemovePropertiesFromTwin(Dictionary<string, string> propertiesToRemov
...
-        bool ShouldExpectDesiredPropertyCallback(TwinState twinState, DateTime twinUpdateTime)
+        bool SkipCallbackValidationDueToEdgeHubRestart(TwinTestState twinTestState, DateTime twinUpdateTime)
         {
-            return (twinUpdateTime - twinState.LastTimeOfEdgeRestart).Duration() > Settings.Current.EdgeHubRestartFailureTolerance;
+            DateTime edgeHubRestartTolerancePeriodLowerBound =
+                twinTestState.EdgeHubLastStopped == DateTime.MinValue ? DateTime.MinValue : twinTestState.EdgeHubLastStopped.Add(-Settings.Current.EdgeHubRestartFailureTolerance);
+            DateTime edgeHubRestartTolerancePeriodUpperBound =
+                twinTestState.EdgeHubLastStarted == DateTime.MinValue ? DateTime.MinValue : twinTestState.EdgeHubLastStarted.Add(Settings.Current.EdgeHubRestartFailureTolerance);
+
+            return edgeHubRestartTolerancePeriodLowerBound <= twinUpdateTime && twinUpdateTime <= edgeHubRestartTolerancePeriodUpperBound;
```
and-rewsmith, line 139: We will also need to apply this failure tolerance when operations to ModuleClient.GetTwin() fail. We shouldn't report these errors if caused by a long edgeHub restart.

This will logic change will affect making reported property updates and the call to GetTwin() in this DesiredPropertiesValidator.cs (line 37).

philipktlin: Talked to Andrew that we should agree to wrap module client with edgehub restart tolerance logic.  But we also agreed to do it in different PR.

test/modules/TwinTester/DesiredPropertiesValidator.cs
```diff
@@ -65,20 +56,21 @@ public async Task ValidateAsync()
             foreach (KeyValuePair<string, DateTime> desiredPropertyUpdate in desiredPropertiesUpdated)
             {
                 bool hasTwinUpdate = propertyUpdatesFromTwin.Contains(desiredPropertyUpdate.Key);
-                bool hasReceivedPotentialCallback = desiredPropertiesReceived.ContainsKey(desiredPropertyUpdate.Key) || !this.ShouldExpectDesiredPropertyCallback(this.twinState, desiredPropertyUpdate.Value);
+                bool hasCallbackReceived = desiredPropertiesReceived.ContainsKey(desiredPropertyUpdate.Key);
                 string status;
-                if (hasTwinUpdate && hasReceivedPotentialCallback)
+
+                if (hasTwinUpdate && (hasCallbackReceived || this.SkipCallbackValidationDueToEdgeHubRestart(this.twinTestState, desiredPropertyUpdate.Value)))
```
and-rewsmith, line 62: I think it is better to consolidate these fields. I feel it makes the logic below (line 69, 73) awkward because hasCallbackReceived might not be needed.

The current implementation will work, but this seems error prone to me.

test/modules/TwinTester/TwinAllOperationsInitializer.cs
```diff
@@ -17,28 +18,47 @@ class TwinAllOperationsInitializer : ITwinTestInitializer
...
+                Logger.LogInformation($"Detected change in connection status:{Environment.NewLine}Changed Status: {status} Reason: {reason}");
+                if (status == ConnectionStatus.Disconnected_Retrying)
+                {
+                    this.twinTestState.EdgeHubLastStopped = DateTime.UtcNow;
+                }
+                else if (status == ConnectionStatus.Connected)
+                {
+                    this.twinTestState.EdgeHubLastStarted = DateTime.UtcNow;
+                }
+            });
```
and-rewsmith, line 54: Good move 👍

philipktlin: :star:

PR #2977, opened 2020-05-19
Test Agent Log Rotation

yophilav:
- Enable log rotation for test agents - 1MB per file, up to 3 log file

Note:
Connectivity Test = 24 modules * 1MB * 3 files = 72MB
LongHaul = 17 modules * 1 MB * 3 files = 51MB
E2E = 4 modules * 1 MB * 3 files = 12MB


e2e_deployment_files/long_haul_deployment.template.json
```diff
@@ -22,7 +22,7 @@
             "type": "docker",
             "settings": {
               "image": "<Container_Registry>/microsoft/azureiotedge-agent:<Build.BuildNumber>-linux-<Architecture>",
-              "createOptions": ""
+              "createOptions": "{\"HostConfig\":{\"LogConfig\":{\"Type\":\"json-file\",\"Config\":{\"max-size\":\"1m\",\"max-file\":\"3\"}}}}"
```
philipktlin, line 25: do you know how big the log file size after 1 day of LH test?  I think ideally we should define 7 number of log file with the size of normal log file size for a day.

yophilav: It looks like LoadGen2 from LH output the most log at about 14MB per day. 
While edgeHub log takes about 3.5MB per day, and edgeAgent takes 0.07MB per day

yophilav: That's why I thought the 3MB per module should be about just enough.

PR #2981, opened 2020-05-20
[Draft] Adding InvalidOperationException to list of unretryable exceptions

dylanbronson:
We received a GH issue that showed many EdgeHub InvalidOperationException (429 from IoTHub) errors. It looked like EdgeHub was using the same sdk moduleClient to connect, when the moduleClient connection had been broken. 

If we add InvalidOperationException to our list of NonRetryableExceptions, we will make a new moduleClient every time we receive that exception.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
@@ -23,7 +23,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
         static readonly Type[] NonRecoverableExceptions =
         {
             typeof(NullReferenceException),
-            typeof(ObjectDisposedException)
+            typeof(ObjectDisposedException),
+            typeof(InvalidOperationException)
         };
 
         readonly IClient client;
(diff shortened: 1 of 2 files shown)


philipktlin:
I found that status 429 means throttling error in IoT hub, https://docs.microsoft.com/en-us/azure/virtual-machines/troubleshooting/troubleshooting-throttling-errors#throttling-by-azure-resource-manager-vs-resource-providers.  Please trace down and confirm if "429" in the log is coming from IoT hub.  If so, I think we should understand what customer scenario is.  Also can we reproduce this bug?  Does customer have this problem again?

Since this GH issue is about 29 days ago, I think the Kusto log may be removed.

philipktlin:
Talked to Dylan and he shared https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs#L94 which is the line converting Amqp.NotAllowed to InvalidOperationException.

Since the exception message "System.InvalidOperationException: Service rejected the message with status: 429" contains status 429 which I found that it means throttling error in IoT hub, https://docs.microsoft.com/en-us/azure/virtual-machines/troubleshooting/troubleshooting-throttling-errors#throttling-by-azure-resource-manager-vs-resource-providers.  I tried to trace up the code path to confirm that this message is coming from IoT hub.  And I looked at  Microsoft.Azure.Amqp github repo, but i can't confirm it is from IoT hub (it is if I understand their code correctly).  Since we are depending on c# sdk.  I think we should ask c# sdk team to confirm it is correct to throw InvalidOperationException rather than IotHubThrottledException even status 429 in exception message?

For this PR, it is better to ask @varunpuranik to review.  thanks.

dylanbronson:
Going to close this. 
It's related to a GH issue we had, but I think this fix would just be fixing a symptom of a greater issue. 
The logs are gone for the event, so we can't exactly see what was going on on the IoTHub side, but it looks like the issue comes from there.

Hub sends 429 -> SDK turns it into InvalidOperationException -

InvalidOperationException shouldn't be "NonRecoverable." Using "retry" here wasn't exactly correct, since edgeHub isn't "retrying," but rather it's doing the same connectivity check every 5 minutes. 

So this PR is fundamentally wrong, but so is the behavior of IoTHub.

PR #2985, opened 2020-05-20
Add cleanup threshold for reported properties validation

philipktlin:
Add cleanup threshold for reported properties validation to see if there is any reported property received after failure threshold and before cleanup threshold.

### DIFF
--- a/test/modules/TwinTester/ReportedPropertiesValidator.cs
+++ b/test/modules/TwinTester/ReportedPropertiesValidator.cs
@@ -112,14 +112,14 @@ namespace TwinTester
                     propertiesToRemove[reportedPropertyUpdate.Key] = null; // will later be serialized as a twin update
                 }
                 else if (!this.reportedPropertyKeysFailedToReceiveWithinThreshold.Contains(reportedPropertyUpdate.Key) &&
-                        this.ExceedFailureThreshold(this.twinState, reportedPropertyUpdate.Key, reportedPropertyUpdate.Value))
+                        this.DoesExceedFailureThreshold(this.twinState, reportedPropertyUpdate.Key, reportedPropertyUpdate.Value))
                 {
                     this.reportedPropertyKeysFailedToReceiveWithinThreshold.Add(reportedPropertyUpdate.Key);
                     status = $"{(int)StatusCode.ReportedPropertyUpdateNotInCloudTwin}: Failure receiving reported property update";
                     await this.HandleReportStatusAsync(status);
                     Logger.LogInformation($"{status} for reported property update {reportedPropertyUpdate.Key}");
                 }
-                else if (this.ExceedCleanupThreshold(this.twinState, reportedPropertyUpdate.Key, reportedPropertyUpdate.Value))
+                else if (this.DoesExceedCleanupThreshold(this.twinState, reportedPropertyUpdate.Key, reportedPropertyUpdate.Value))
                 {
                     status = $"{(int)StatusCode.ReportedPropertyUpdateNotInCloudTwinAfterCleanUpThreshold}: Failure receiving reported property update after cleanup threshold";
                     await this.HandleReportStatusAsync(status);
@@ -131,7 +131,7 @@ namespace TwinTester
             return propertiesToRemove;
         }
 
-        bool ExceedFailureThreshold(TwinState twinState, string reportedPropertyKey, DateTime reportedPropertyUpdatedAt)
+        bool DoesExceedFailureThreshold(TwinState twinState, string reportedPropertyKey, DateTime reportedPropertyUpdatedAt)
         {
             DateTime comparisonPoint = reportedPropertyUpdatedAt > twinState.LastTimeOffline ? reportedPropertyUpdatedAt : twinState.LastTimeOffline;
             bool exceedFailureThreshold = DateTime.UtcNow - comparisonPoint > Settings.Current.TwinUpdateFailureThreshold;
@@ -146,7 +146,7 @@ namespace TwinTester
             return exceedFailureThreshold;
         }
 
-        bool ExceedCleanupThreshold(TwinState twinState, string reportedPropertyKey, DateTime reportedPropertyUpdatedAt)
+        bool DoesExceedCleanupThreshold(TwinState twinState, string reportedPropertyKey, DateTime reportedPropertyUpdatedAt)
         {
             TimeSpan cleanUpThreshold = TimeSpan.FromMinutes(5);
             DateTime comparisonPoint = reportedPropertyUpdatedAt > twinState.LastTimeOffline ? reportedPropertyUpdatedAt : twinState.LastTimeOffline;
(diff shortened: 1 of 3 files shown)


test/modules/TwinTester/StatusCode.cs
```diff
@@ -8,10 +8,12 @@ enum StatusCode
         DesiredPropertyReceived = 202,
         ReportedPropertyReceived = 203,
         ReportedPropertyUpdated = 204,
+        ReportedPropertyReceivedAfterThreshold = 205,
```
and-rewsmith, line 11: Seems weird to me that this is a success, considering the update didn't come in time. If you want to separate this from the errors I think it is good to use another leading number (i.e. 3 or 4)

test/modules/TwinTester/TwinAllOperationsResultHandler.cs
```diff
@@ -72,11 +72,11 @@ public Task HandleReportedPropertyUpdateExceptionAsync(string failureStatus)
             return this.CallAnalyzer(failureStatus);
         }
 
-        async Task CallAnalyzer(string failureStatus)
+        async Task CallAnalyzer(string result)
```
and-rewsmith, line 75: good change this would be very confusing

test/modules/TwinTester/StatusCode.cs
philipktlin: I am open to use "5". 😃

test/modules/TwinTester/ReportedPropertiesValidator.cs
```diff
@@ -52,14 +53,14 @@ public async Task ValidateAsync()
                 return;
             }
 
-            TwinCollection propertiesToRemoveFromTwin = await this.ValidatePropertiesFromTwinAsync(receivedTwin);
-            await this.RemovePropertiesFromStorage(propertiesToRemoveFromTwin);
-            await this.RemovePropertiesFromTwin(propertiesToRemoveFromTwin);
+            TwinCollection propertiesToRemove = await this.ValidatePropertiesFromTwinAsync(receivedTwin);
+            await this.RemovePropertiesFromStorage(propertiesToRemove);
+            await this.RemovePropertiesFromTwin(propertiesToRemove);
```
and-rewsmith, line 58: If you want to consolidate these methods I think it is fine

test/modules/TwinTester/ReportedPropertiesValidator.cs
```diff
@@ -89,46 +90,76 @@ async Task<TwinCollection> ValidatePropertiesFromTwinAsync(Twin receivedTwin)
...
 
-                    status = $"{(int)StatusCode.ValidationSuccess}: Successfully validated reported property update";
-                    Logger.LogInformation(status + $" {reportedPropertyUpdate.Key}");
+                    await this.HandleReportStatusAsync(status);
+                    Logger.LogInformation($"{status} for reported property update {reportedPropertyUpdate.Key}");
+                    propertiesToRemove[reportedPropertyUpdate.Key] = null; // will later be serialized as a twin update
                 }
-                else if (this.ExceedFailureThreshold(this.twinState, reportedPropertyUpdate.Value))
+                else if (!this.reportedPropertyKeysFailedToReceiveWithinThreshold.Contains(reportedPropertyUpdate.Key) &&
+                        this.ExceedFailureThreshold(this.twinState, reportedPropertyUpdate.Key, reportedPropertyUpdate.Value))
```
and-rewsmith, line 115: super nit: I try to always make my bool names ask a question (i.e. DoesExceedFailureThreshold)

test/modules/TwinTester/ReportedPropertiesValidator.cs
philipktlin: not for now. minimize the change. add a bug https://dev.azure.com/msazure/One/_workitems/edit/7272973

test/modules/TwinTester/TwinAllOperationsResultHandler.cs
philipktlin: ✌️

test/modules/TwinTester/ReportedPropertiesValidator.cs
```diff
@@ -89,46 +90,76 @@ async Task<TwinCollection> ValidatePropertiesFromTwinAsync(Twin receivedTwin)
...
+                    Logger.LogInformation($"{status} for reported property update {reportedPropertyUpdate.Key}");
+                    propertiesToRemove[reportedPropertyUpdate.Key] = null; // will later be serialized as a twin update
                 }
+            }
+
+            return propertiesToRemove;
+        }
 
-                propertiesToRemoveFromTwin[reportedPropertyUpdate.Key] = null; // will later be serialized as a twin update
-                await this.HandleReportStatusAsync(status);
```
and-rewsmith, line 122: We need to handle report status in every case so I think this can stay here, right?

test/modules/TwinTester/ReportedPropertiesValidator.cs
```diff
@@ -89,46 +90,76 @@ async Task<TwinCollection> ValidatePropertiesFromTwinAsync(Twin receivedTwin)
...
+                    await this.HandleReportStatusAsync(status);
+                    Logger.LogInformation($"{status} for reported property update {reportedPropertyUpdate.Key}");
+                    propertiesToRemove[reportedPropertyUpdate.Key] = null; // will later be serialized as a twin update
                 }
+            }
+
+            return propertiesToRemove;
+        }
 
-                propertiesToRemoveFromTwin[reportedPropertyUpdate.Key] = null; // will later be serialized as a twin update
```
and-rewsmith, line 121: Good change. It is cleaner to remove the "else continue" and make this happen only where needed. We should also change the DesiredPropertiesValidator at some point.

and-rewsmith:
Why not make the same change for desired properties?

test/modules/TwinTester/ReportedPropertiesValidator.cs
philipktlin: 🚀

test/modules/TwinTester/ReportedPropertiesValidator.cs
philipktlin: since I remove "else continue", I need to move this line into each if block.

and-rewsmith: What if we added after the if/else-if chain, ```if status != "": ReportStatus()```?

philipktlin: I think it is simple to put that line in each if blocks, so dev just read only the if-block they care about.  we can extract it out if we have more common calls across these if blocks.

philipktlin:
> Why not make the same change for desired properties?
I added cleanup threshold to reported properties validator as I would like to see if the reported properties still get updated.  It is a workaround.

To completely fix this, I think we should not add cleanup threshold, but implement similar logic to consider edgeHub restart period in desired properties validator.  However, I would like to minimize the changes before 1.0.10 rc1.

test/modules/TwinTester/ReportedPropertiesValidator.cs
philipktlin: 👌

PR #2987, opened 2020-05-20
EdgeHub recursively pulls scopes for nested devices

richma-ms:
This change updates DeviceScopeApiClient to hit the new GetModulesAndDevicesInTargetScope API to retrieve scopes for nested Edge devices, and reworks DeviceScopeIdentityCache to construct an in-memory tree to represent the nested hierarchy.

The existing and new unit test pass, but I suspect E2E tests will fail since they're still target the production deployments, and the new API is only available in Anca's private deployment.  I'll do manual E2E testing against the private deployment before merging.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -133,6 +150,56 @@ async Task<ScopeResult> GetIdentitiesInScope(Uri uri)
...
+                    this.retryStrategy);
+            }
+            catch (Exception e)
+            {
+                Events.ErrorInScopeResult(e);
+                throw;
+            }
+        }
+
+        async Task<ScopeResult> GetIdentitiesInTargetScope(Uri uri, Option<string> continuationLink)
```
vadim-kovalyov, line 169: nit: 
```suggestion
        async Task<ScopeResult> GetIdentitiesInTargetScope(Uri uri, Option<string> continuationToken)
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -133,6 +150,56 @@ async Task<ScopeResult> GetIdentitiesInScope(Uri uri)
...
+                {
+                    // TODO: RICHMA - pass through the AuthChain
+                    AuthChain = "edge2;edge1;edge0",
+                    PageSize = this.batchSize,
+                    ContinuationLink = continuationLink.GetOrElse(() => string.Empty)
+                };
+
+                string token = await this.edgeHubTokenProvider.GetTokenAsync(Option.None<TimeSpan>());
+                msg.Headers.Add(HttpRequestHeader.Authorization.ToString(), token);
+                msg.Content = new StringContent(JsonConvert.SerializeObject(payload).ToString(), Encoding.Default, "application/json");
```
vadim-kovalyov, line 186: nit:
```suggestion
                msg.Content = new StringContent(JsonConvert.SerializeObject(payload), Encoding.Default, "application/json");
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedScopeRequest.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// This is an object to encapsulate JSON serialization of the
+    /// GetDevicesAndModulesInTargetScope POST request.
+    /// </summary>
+    class NestedScopeRequest
+    {
+        [JsonProperty(PropertyName = "pageSize")]
+        public int PageSize { get; set; }
+
+        [JsonProperty(PropertyName = "continuationLink", NullValueHandling = NullValueHandling.Ignore)]
+        public string ContinuationLink;
```
vadim-kovalyov, line 16: nit: ContinuationToken?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Devices" Version="1.19.0" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.19.0-richma" />
```
vadim-kovalyov, line 11: I don't have much context, but just curious why are you changing this?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -175,6 +177,60 @@ public static void BadRequestResult(string id, HttpStatusCode statusCode)
             }
         }
 
+        /// <summary>
+        /// This is a helper class used to pull down the devices/modules in scope for every nested Edge Device.
+        /// It maintains a queue of ServiceIdentitiesIterator, each representating a child Edge device
+        /// in the subtree, and uses these iterators to perform breadth-first-expansion on the tree.
+        /// </summary>
```
vadim-kovalyov, line 184: thanks for great comments! helps a lot to get context.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -175,6 +177,60 @@ public static void BadRequestResult(string id, HttpStatusCode statusCode)
...
+                if (this.remainingEdgeNodes.Count == 0)
+                {
+                    return Task.FromResult(Enumerable.Empty<ServiceIdentity>());
+                }
+
+                // Get the head without popping, as each iterator could potentially require
+                // multiple calls to GetNext() to fully drain, we can't remove it from
+                // the queue until it's fully completed
+                IServiceIdentitiesIterator currentNode = this.remainingEdgeNodes.Peek();
+                IList<ServiceIdentity> results = currentNode.GetNext().Result.ToList();
```
vadim-kovalyov, line 212: q: why blocking call? Do we support C# 8, and can we implement IAsyncEnumerator?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,240 @@
...
+            {
+                // Auth chain starts with the originating device
+                authChain = treeNode.Identity.Id;
+
+                // Walk the parents up to the root
+                while (treeNode.Parent.HasValue)
+                {
+                    // Accumulate the auth chain per parent visited in the format:
+                    // "leaf1;edge1;edge2;...;rootEdge"
+                    ServiceIdentityTreeNode parent = treeNode.Parent.Expect(() => new InvalidOperationException());
```
vadim-kovalyov, line 95: nit: I would add a message to the exception.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,240 @@
...
+                    ? Option.Some(treeNode.Identity)
+                    : Option.None<ServiceIdentity>();
+        }
+
+        public IList<string> GetAllIds()
+        {
+            return this.nodes.Select(kvp => kvp.Value.Identity.Id).ToList();
+        }
+
+        public bool TryGetAuthChain(string id, out string authChain)
```
vadim-kovalyov, line 79: nit: maybe return Option instead of TryGet...?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -133,6 +150,56 @@ async Task<ScopeResult> GetIdentitiesInScope(Uri uri)
...
+        async Task<ScopeResult> GetIdentitiesInTargetScope(Uri uri, Option<string> continuationLink)
+        {
+            HttpClient client = this.proxy
+                .Map(p => new HttpClient(new HttpClientHandler { Proxy = p }, disposeHandler: true))
+                .GetOrElse(() => new HttpClient());
+            using (var msg = new HttpRequestMessage(HttpMethod.Post, uri))
+            {
+                var payload = new NestedScopeRequest()
+                {
+                    // TODO: RICHMA - pass through the AuthChain
```
richma-ms, line 178: >// TODO: RICHMA - pass through the AuthChain [](start = 20, length = 44)

As I'm doing manual E2E testing, I quickly realized that I forgot to implement this...will have another iteration out soon.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -20,8 +21,12 @@ public class DeviceScopeApiClient : IDeviceScopeApiClient
 
         const string InScopeTargetIdentityUriFormat = "/devices/{0}/modules/{1}/deviceAndModuleInDeviceScope?targetDeviceId={2}&targetModuleId={3}&api-version={4}";
 
+        const string GetDevicesAndModulesInTargetScopeUriFormat = "/devices/{0}/modules/{1}/devicesAndModulesInTargetDeviceScope?api-version={2}";
+
         const string ApiVersion = "2018-08-30-preview";
 
+        const string NestedApiVersion = "2020-06-30-preview";
```
philipktlin, line 29: I expected the new api version should be backward compatible for existing API calls, and only new api version should be used.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -55,28 +60,40 @@ public class DeviceScopeApiClient : IDeviceScopeApiClient
             this.retryStrategy = retryStrategy ?? TransientRetryStrategy;
         }
 
+        public DeviceScopeApiClient(string deviceId, DeviceScopeApiClient client)
+        {
+            this.deviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
+            this.iotHubBaseHttpUri = client.iotHubBaseHttpUri;
```
philipktlin, line 69: Preconditions.CheckNonWhiteSpace

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,240 @@
...
+            // this include placing the new node as the new root.
+            List<ServiceIdentityTreeNode> danglingChildren =
+                this.nodes
+                .Select(kvp => kvp.Value)
+                .Where(s => s.Identity.ParentScopes.HasValue && s.Identity.ParentScopes.Equals(device.DeviceScope))
+                .ToList();
+
+            foreach (ServiceIdentityTreeNode child in danglingChildren)
+            {
+                child.Parent = Option.Some(newNode);
```
huguesBouvier, line 197: Np: Code is when parent comes in late is reapeate: see lines [150, 151], [124,125], [164,165].
Maybe encapsulate in a function to make sure that if one area is modified they are modified the same way.
ie: The operation that would have been done if the parent was there need to be done when the parent might finally coming. Someone changing the code in the year or 2 might not be aware that change need to be made similarly in those 3 areas.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,240 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core
```
huguesBouvier, line 2: Is it possible to create cycles by mistake in the configuration? If that's the case, code might crash in recursively looping

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -55,28 +63,41 @@ public class DeviceScopeApiClient : IDeviceScopeApiClient
...
+            return this.GetIdentityWithRetry(this.GetServiceUri(targetDeviceId, targetModuleId));
         }
 
-        internal Uri GetServiceUri(Option<string> continuationToken) =>
-            continuationToken
-                .Map(c => new Uri(this.iotHubBaseHttpUri, c))
-                .GetOrElse(
-                    () =>
-                    {
-                        string relativeUri = InScopeIdentitiesUriTemplate.FormatInvariant(this.deviceId, this.moduleId, this.batchSize, null, ApiVersion);
```
philipktlin, line 76: I can't find any usage of InScopeIdentitiesUriTemplate.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -133,6 +150,56 @@ async Task<ScopeResult> GetIdentitiesInScope(Uri uri)
...
+        {
+            HttpClient client = this.proxy
+                .Map(p => new HttpClient(new HttpClientHandler { Proxy = p }, disposeHandler: true))
+                .GetOrElse(() => new HttpClient());
+            using (var msg = new HttpRequestMessage(HttpMethod.Post, uri))
+            {
+                var payload = new NestedScopeRequest()
+                {
+                    // TODO: RICHMA - pass through the AuthChain
+                    AuthChain = "edge2;edge1;edge0",
```
huguesBouvier: Is it a placeholder?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -21,6 +21,8 @@ public ServiceProxy(IDeviceScopeApiClient securityScopesApiClient)
 
         public IServiceIdentitiesIterator GetServiceIdentitiesIterator() => new ServiceIdentitiesIterator(this.securityScopesApiClient);
```
huguesBouvier, line 22: Do we need to keep this part and all the code attached to it? Seems that nested is a super set of this one.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -175,6 +177,60 @@ public static void BadRequestResult(string id, HttpStatusCode statusCode)
...
+                        var childClient = this.rootClient.CreateOnBehalfOfDeviceScopeClient(identity.DeviceId);
+                        this.remainingEdgeNodes.Enqueue(new ServiceIdentitiesIterator(childClient));
+                    }
+                }
+
+                return Task.FromResult(results.AsEnumerable());
+            }
+        }
+
         class ServiceIdentitiesIterator : IServiceIdentitiesIterator
```
huguesBouvier, line 234: See above, if it is possible to remove it, it would simplify the code a lot.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
philipktlin: Use Encoding.UTF8

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -55,28 +63,41 @@ public class DeviceScopeApiClient : IDeviceScopeApiClient
...
+        public DeviceScopeApiClient(string deviceId, DeviceScopeApiClient client)
+        {
+            this.deviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
+            this.iotHubBaseHttpUri = client.iotHubBaseHttpUri;
+            this.moduleId = client.moduleId;
+            this.batchSize = client.batchSize;
+            this.edgeHubTokenProvider = client.edgeHubTokenProvider;
+            this.proxy = client.proxy;
+            this.retryStrategy = client.retryStrategy;
+            this.AuthChainProvider = client.AuthChainProvider;
```
philipktlin, line 75: need to call Preconditions.CheckNotNull. And should make it readonly, right?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -133,6 +154,60 @@ async Task<ScopeResult> GetIdentitiesInScope(Uri uri)
...
+                if (!this.AuthChainProvider.TryGetAuthChain(this.deviceId, out string authChain))
+                {
+                    throw new ArgumentException($"No valid authentication chain for {this.deviceId}");
+                }
+
+                var payload = new NestedScopeRequest()
+                {
+                    AuthChain = authChain,
+                    PageSize = this.batchSize,
+                    ContinuationLink = continuationLink.GetOrElse(() => string.Empty)
```
philipktlin, line 189: Rename "ContinuationLink" to "ContinuationToken"?
```suggestion
                    ContinuationToken = continuationToken.GetOrElse(() => string.Empty)
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -133,6 +154,60 @@ async Task<ScopeResult> GetIdentitiesInScope(Uri uri)
...
+                string content = await response.Content.ReadAsStringAsync();
+                if (response.IsSuccessStatusCode)
+                {
+                    var scopeResult = JsonConvert.DeserializeObject<ScopeResult>(content);
+                    Events.GotValidResult();
+                    return scopeResult;
+                }
+                else
+                {
+                    throw new DeviceScopeApiException("Error getting device scope result from IoTHub", response.StatusCode, content);
```
philipktlin, line 206: should we change the word "IoTHub", as it can be nested parent edge device? or even better, if we have different message for iot hub and parent edge device (including parent edge device id in the message).
```suggestion
                    throw new DeviceScopeApiException("Error getting device scope result from upstream", response.StatusCode, content);
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/IDeviceScopeApiClient.cs
```diff
@@ -2,9 +2,12 @@
 namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
 {
     using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Hub.Core;
 
     public interface IDeviceScopeApiClient
     {
+        IDeviceScopeApiClient CreateOnBehalfOfDeviceScopeClient(string deviceId);
```
philipktlin, line 9: just want to confirm, the device id argument can be leaf device id or edge device id including both child and root. do I understand correctly?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/IDeviceScopeApiClient.cs
```diff
@@ -2,9 +2,12 @@
...
+
         Task<ScopeResult> GetIdentitiesInScope();
```
philipktlin, line 11: should we fix the naming as this method returns Task<ScopeResult>, Async suffix should be used.  Same for GetNext method.
```suggestion
        Task<ScopeResult> GetIdentitiesInScopeAsync();
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -29,31 +29,41 @@ public sealed class DeviceScopeIdentitiesCache : IDeviceScopeIdentitiesCache
...
             TimeSpan refreshRate)
         {
             this.serviceProxy = serviceProxy;
             this.encryptedStore = encryptedStorage;
-            this.serviceIdentityCache = initialCache;
             this.refreshRate = refreshRate;
             this.refreshCacheTimer = new Timer(this.RefreshCache, null, TimeSpan.Zero, refreshRate);
+
+            // Populate the ServiceIdentityTree
+            this.serviceIdentityTree = new ServiceIdentityTree(edgeDeviceId);
```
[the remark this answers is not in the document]
richma-ms: yea actually that's a good idea, if the caller creates the tree instance and passes in an interface ref, then we have an easier time hooking up the authchain provider to the devicescopeApiClient.

---
In reply to: [428739558](https://github.com/Azure/iotedge/pull/2987#discussion_r428739558) [](ancestors = 428739558)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
philipktlin: It seems this method is no longer used in DeviceScopeIdentitiesCache line158, which is replaced by GetNestedServiceIdentitiesIterator method.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,275 @@
...
+    /// Edge device in a nested Edge environment.
+    /// </summary>
+    class ServiceIdentityTree : IAuthenticationChainProvider
+    {
+        public event EventHandler<string> ServiceIdentityRemoved;
+
+        readonly string rootDeviceId;
+        AtomicReference<ImmutableDictionary<string, ServiceIdentityTreeNode>> nodes;
+
+        public ServiceIdentityTree(string rootDeviceId)
```
philipktlin, line 24: is this possible for a subtree's root device? if so, we should use a different name to avoid confusion of root device of the entire tree.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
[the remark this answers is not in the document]
philipktlin: Should we register  IAuthenticationChainProvider as singleton in dependency injection?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -55,28 +63,41 @@ public class DeviceScopeApiClient : IDeviceScopeApiClient
             this.retryStrategy = retryStrategy ?? TransientRetryStrategy;
         }
 
+        public DeviceScopeApiClient(string deviceId, DeviceScopeApiClient client)
```
varunpuranik, line 66: Where do we need to use this?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -210,22 +217,22 @@ async Task<Option<ServiceIdentity>> GetServiceIdentityInternal(string id)
             Preconditions.CheckNonWhiteSpace(id, nameof(id));
             using (await this.cacheLock.LockAsync())
```
philipktlin, line 218: do we still need lock here since you changed to use AtomicReference?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -133,6 +159,61 @@ async Task<ScopeResult> GetIdentitiesInScope(Uri uri)
...
+
+        async Task<ScopeResult> GetIdentitiesInTargetScope(Uri uri, Option<string> continuationLink)
+        {
+            HttpClient client = this.proxy
+                .Map(p => new HttpClient(new HttpClientHandler { Proxy = p }, disposeHandler: true))
+                .GetOrElse(() => new HttpClient());
+            using (var msg = new HttpRequestMessage(HttpMethod.Post, uri))
+            {
+                // Get the auth-chain for the target device
+                if (!this.AuthChainProvider.TryGetAuthChain(this.targetEdgeDeviceId, out string authChain))
```
richma-ms, line 186: >f (!this.AuthChainProvider.TryGetAuthChain(this.targetEdgeDeviceId, out string authChain)) [](start = 17, length = 90)

From David: Consider accumulating the authchain through the nested iterator, as opposed to re-calculating it at runtime everytime we make a request.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,275 @@
...
+    /// <summary>
+    /// This is a tree implementation that uses the ParentScopes property
+    /// to reconstruct the hierarchy of modules/devices under the current
+    /// Edge device in a nested Edge environment.
+    /// </summary>
+    class ServiceIdentityTree : IAuthenticationChainProvider
+    {
+        public event EventHandler<string> ServiceIdentityRemoved;
+
+        readonly string rootDeviceId;
```
richma-ms, line 21: >rootDeviceId [](start = 24, length = 12)

maybe actorDeviceId?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -30,12 +36,15 @@ public class DeviceScopeApiClient : IDeviceScopeApiClient
         readonly RetryStrategy retryStrategy;
 
         readonly Uri iotHubBaseHttpUri;
-        readonly string deviceId;
+        readonly string rootEdgeDeviceId;
```
richma-ms, line 39: >rootEdgeDeviceId [](start = 24, length = 16)

maybe actorDeviceId?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,275 @@
...
+                // Auth chain starts with the originating device
+                authChain = treeNode.Identity.Id;
+
+                // Walk the parents up to the root
+                while (treeNode.Parent.HasValue)
+                {
+                    // Accumulate the auth chain per parent visited in the format:
+                    // "leaf1;edge1;edge2;...;rootEdge"
+                    ServiceIdentityTreeNode parent = treeNode.Parent.Expect(() => new InvalidOperationException());
+                    authChain += ";" + parent.Identity.Id;
```
philipktlin, line 128: should we use local variable and assign authChain just before return true statement?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -175,6 +177,61 @@ public static void BadRequestResult(string id, HttpStatusCode statusCode)
...
+                IServiceIdentitiesIterator currentNode = this.remainingEdgeNodes.Peek();
+                IList<ServiceIdentity> results = currentNode.GetNext().Result.ToList();
+
+                if (!currentNode.HasNext)
+                {
+                    // We're done with this iterator, it's safe to pop it now
+                    this.remainingEdgeNodes.Dequeue();
+                }
+
+                foreach (ServiceIdentity identity in results)
```
richma-ms, line 220: >foreach (ServiceIdentity identity in results) [](start = 16, length = 45)

From Anca: check for max level depth here? (5)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,275 @@
...
+            if (!this.nodes.CompareAndSet(snapshot, writableSnapshot.ToImmutableDictionary()))
+            {
+                // There's only ever one thread calling InsertOrUpdate, so this should never fail
+                throw new InvalidOperationException("Multi-threaded insert/update detected on ServiceIdentityTree");
+            }
+        }
+
+        public void Remove(string id)
+        {
+            // In the case of an explicit removal, the device/module is
```
philipktlin, line 67: should we introduce lock mechanism here rather than in the caller class, DeviceScopeIdentitiesCache?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -175,6 +177,61 @@ public static void BadRequestResult(string id, HttpStatusCode statusCode)
...
 
+        /// <summary>
+        /// This is a helper class used to pull down the devices/modules in scope for every nested Edge Device.
+        /// It maintains a queue of ServiceIdentitiesIterator, each representating a child Edge device
+        /// in the subtree, and uses these iterators to perform breadth-first-expansion on the tree.
+        /// </summary>
+        class NestedServiceIdentitiesIterator : IServiceIdentitiesIterator
+        {
+            IDeviceScopeApiClient rootClient;
+            Queue<IServiceIdentitiesIterator> remainingEdgeNodes;
```
richma-ms, line 188: >IServiceIdentitiesIterator [](start = 18, length = 26)

From David: Maybe instead of having a queue of iterators that deals with the continuation token, just maintain a queue of ApiClients.  Then every time we need to use the continuation token, enqueue a new ApiClient.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,275 @@
...
+            }
+            else
+            {
+                // It's okay if we can't find the parent device, it's possible that the identity
+                // for the parent device hasn't came in yet. In this case we'll just leave the
+                // module as a dangling node, and the parent should come in at a later point.
+            }
+        }
+
+        void InsertDeviceIdentity(IDictionary<string, ServiceIdentityTreeNode> nodes, ServiceIdentity device)
```
philipktlin, line 168:
```suggestion
        void InsertDeviceIdentity(IDictionary<string, ServiceIdentityTreeNode> nodes, ServiceIdentity deviceServiceIdentity)
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,275 @@
...
+            ImmutableDictionary<string, ServiceIdentityTreeNode> snapshot = this.nodes;
+            var writableSnapshot = new Dictionary<string, ServiceIdentityTreeNode>(snapshot);
+
+            this.RemoveRecursive(writableSnapshot, id);
+
+            // Write back to the atomic field
+            if (!this.nodes.CompareAndSet(snapshot, writableSnapshot.ToImmutableDictionary()))
+            {
+                // There's only ever one thread calling Remove(), so this should never fail
+                throw new InvalidOperationException("Multi-threaded insert/update detected on ServiceIdentityTree");
```
philipktlin, line 79: Does identity cache get corrupted? How to recover this?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        /// <summary>
+        /// This is an object to wrap ServiceIdentity so that it can
+        /// point to a parent and act as a node in a tree.
+        /// </summary>
+        internal class ServiceIdentityTreeNode
+        {
+            public ServiceIdentity Identity { get; private set; }
+            public Option<ServiceIdentityTreeNode> Parent;
```
richma-ms, line 264: >public Option<ServiceIdentityTreeNode> Parent; [](start = 12, length = 46)

From David: maybe carry a static string for the authchain in each node, so that we don't ever need to traverse upwards?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
richma-ms: I think you're right, but based on David's suggestion for removing IAuthenticationChainProvider, maybe we can get rid of the AtomicReference in ServiceIdentityTree, which means wel'l need this lock again.

---
In reply to: [428953358](https://github.com/Azure/iotedge/pull/2987#discussion_r428953358) [](ancestors = 428953358)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -155,7 +162,7 @@ async Task RefreshCache()
                 {
                     Events.StartingRefreshCycle();
                     var currentCacheIds = new List<string>();
-                    IServiceIdentitiesIterator iterator = this.serviceProxy.GetServiceIdentitiesIterator();
+                    IServiceIdentitiesIterator iterator = this.serviceProxy.GetNestedServiceIdentitiesIterator();
                     while (iterator.HasNext)
                     {
                         IEnumerable<ServiceIdentity> batch = await iterator.GetNext();
```
richma-ms, line 168: >IEnumerable<ServiceIdentity> batch = await iterator.GetNext(); [](start = 24, length = 62)

From David: maybe instead of blocking on individual network calls, we collate a collection of tasks, and try to resolve them all in parallel.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,275 @@
...
+                nodes
+                .Select(kvp => kvp.Value)
+                .Where(serviceIdentity => serviceIdentity.Identity.IsEdgeDevice && serviceIdentity.Identity.DeviceScope.Contains(scopeId))
+                .ToList();
+
+            if (devices.Count() > 0)
+            {
+                // There shouldn't be more than one device, but even if there is
+                // for some reason, we can just use the first one
+                return Option.Some(devices.First());
```
philipktlin, line 251: I think it is better to have a data structure to protect adding a service identity with duplicate device scope.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -20,8 +22,12 @@ public class DeviceScopeApiClient : IDeviceScopeApiClient
 
         const string InScopeTargetIdentityUriFormat = "/devices/{0}/modules/{1}/deviceAndModuleInDeviceScope?targetDeviceId={2}&targetModuleId={3}&api-version={4}";
 
+        const string GetDevicesAndModulesInTargetScopeUriFormat = "/devices/{0}/modules/{1}/devicesAndModulesInTargetDeviceScope?api-version={2}";
```
varunpuranik, line 25: It might be cleaner to have a new class that implements IDeviceScopeApiClient

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedScopeRequest.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
+{
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// This is an object to encapsulate JSON serialization of the
+    /// GetDevicesAndModulesInTargetScope POST request.
+    /// </summary>
+    class NestedScopeRequest
+    {
```
varunpuranik, line 11: Pls make this immutable

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/IDeviceScopeApiClient.cs
varunpuranik, line 9: If we split it out, we don't need this method?

varunpuranik: Maybe use a Provider pattern will be better

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -175,6 +177,61 @@ public static void BadRequestResult(string id, HttpStatusCode statusCode)
...
+                if (this.remainingEdgeNodes.Count == 0)
+                {
+                    return Task.FromResult(Enumerable.Empty<ServiceIdentity>());
+                }
+
+                // Get the head without popping, as each iterator could potentially require
+                // multiple calls to GetNext() to fully drain, we can't remove it from
+                // the queue until it's fully completed
+                IServiceIdentitiesIterator currentNode = this.remainingEdgeNodes.Peek();
+                IList<ServiceIdentity> results = currentNode.GetNext().Result.ToList();
```
varunpuranik, line 212: This .Result is dangerous, and can deadlock..

varunpuranik: await the result instead

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
varunpuranik: I wanted to use AsyncEnumerator when I coded this.. it was not available at that time! :)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -210,22 +217,22 @@ async Task<Option<ServiceIdentity>> GetServiceIdentityInternal(string id)
...
-                this.serviceIdentityCache[id] = storedServiceIdentity;
-                await this.SaveServiceIdentityToStore(id, storedServiceIdentity);
+                bool hasValidServiceIdentity = this.serviceIdentityTree.Get(id)
+                    .Filter(s => s.Status == ServiceIdentityStatus.Enabled)
+                    .HasValue;
+
+                // Remove the target identity and all its children from the tree,
+                // this will invoke a callback for each deleted identity, so we
+                // can also delete them from storage.
+                this.serviceIdentityTree.Remove(id);
```
varunpuranik, line 235: We need to keep in mind -ve caching

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,244 @@
...
+            else
+            {
+                this.InsertDeviceIdentity(writableSnapshot, identity);
+            }
+
+            // Write back to the atomic field
+            if (!this.nodes.CompareAndSet(snapshot, writableSnapshot.ToImmutableDictionary()))
+            {
+                // There's only ever one thread calling InsertOrUpdate, so this should never fail
+                throw new InvalidOperationException("Multi-threaded insert/update detected on ServiceIdentityTree");
```
varunpuranik, line 57: How will throwing work here? Do you expect the caller to re-try? WHy not add that logic here?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
varunpuranik: Yes, we should definitely handle cycles

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,244 @@
...
+            this.nodes = new AtomicReference<ImmutableDictionary<string, ServiceIdentityTreeNode>>(ImmutableDictionary<string, ServiceIdentityTreeNode>.Empty);
+        }
+
+        public void InsertOrUpdate(ServiceIdentity identity)
+        {
+            // There should always be a valid ServiceIdentity
+            Preconditions.CheckNotNull(identity, nameof(identity));
+
+            ImmutableDictionary<string, ServiceIdentityTreeNode> snapshot = this.nodes;
+            var writableSnapshot = new Dictionary<string, ServiceIdentityTreeNode>(snapshot);
```
varunpuranik, line 36: This snapshotting of ImmutableDictionary seems expensive in terms of allocations and time.. esp as the tree could get updated relatively frequently. Maybe a readerwriter lock is better?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,244 @@
...
+                    {
+                        // We've reached the root device, we're done
+                        return true;
+                    }
+
+                    // Move on to the next parent
+                    treeNode = parent;
+                }
+
+                // If we walked the parent chain up to a dead-end, i.e. topmost node does not
```
varunpuranik, line 143: Can you pls explain again how dangling trees are possible? If a middle device gets deleted? Should we handle that as part of the remove then?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,244 @@
...
+
+            if (nodes.TryGetValue(module.DeviceId, out ServiceIdentityTreeNode parentDeviceNode))
+            {
+                // Hook up the module to the parent device
+                newNode.Parent = Option.Some(parentDeviceNode);
+                parentDeviceNode.Children.Add(newNode);
+            }
+            else
+            {
+                // It's okay if we can't find the parent device, it's possible that the identity
```
varunpuranik, line 165: This happens if the child connects before we have had a chance to sync the tree?

varunpuranik: My concern here is - we are not actively doing anything to complete the tree.. in that case GetAuthChain will not be able to create the chain and we will not do anything about it.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
varunpuranik: Updates become tricky in that case. But it certainly seems efficient to keep a cache of the auth chains. Maybe in a different map that can be invalidated?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,244 @@
...
+
+        /// <summary>
+        /// This is an object to wrap ServiceIdentity so that it can
+        /// point to a parent and act as a node in a tree.
+        /// </summary>
+        internal class ServiceIdentityTreeNode
+        {
+            public ServiceIdentity Identity { get; private set; }
+            public Option<ServiceIdentityTreeNode> Parent;
+            public List<ServiceIdentityTreeNode> Children;
```
varunpuranik, line 234: Where is this list being used?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,244 @@
...
+            return snapshot.Select(kvp => kvp.Value.Identity.Id).ToList();
+        }
+
+        public bool TryGetAuthChain(string id, out string authChain)
+        {
+            Preconditions.CheckNonWhiteSpace(id, nameof(id));
+            authChain = string.Empty;
+
+            // Auth-chain for root
+            if (this.rootDeviceId == id)
```
varunpuranik, line 112: Nit: It may be possible to remove this special case by making this the while loop condition.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
varunpuranik: It feels like we should either fill up the tree up-front (and in updates), or fill it up on demand on GetAuthChain

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,244 @@
...
+    using Microsoft.Azure.Devices.Edge.Util.Concurrency;
+
+    /// <summary>
+    /// This is a tree implementation that uses the ParentScopes property
+    /// to reconstruct the hierarchy of modules/devices under the current
+    /// Edge device in a nested Edge environment.
+    /// </summary>
+    public class ServiceIdentityTree :
+        IServiceIdentityTree,
+        IAuthenticationChainProvider
```
varunpuranik, line 19: Do we need this another interface? IServiceIdentityTree seems to already implement GetAuthTree

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedScopeRequest.cs
richma-ms: yea "Link" is technically incorrect terminology for POSTs.  But service side names it ContinuationLink due to legacy, as do our existing request/response classes, so I kept it as such for consistency.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.csproj
richma-ms: nested Edge need SDK changes that are not released yet, so I had to create my own SDK version for the interim.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/IDeviceScopeApiClient.cs
richma-ms: Yea good idea

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
richma-ms: The tree impl is now threadsafe via locks, and I removed the locking at the caching layer.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
richma-ms: yea we're already protected by using a dictionary

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,326 @@
...
+            }
+
+            public void AddChild(ServiceIdentityTreeNode childNode)
+            {
+                if (!this.Identity.IsEdgeDevice)
+                {
+                    throw new ArgumentException($"{this.Identity.Id} is not an Edge device, only Edge devices can have children");
+                }
+
+                if (this.currentDepth >= MaxNestingDepth && childNode.Identity.IsEdgeDevice)
```
ancaantochi, line 242: I think MaxNestingDepth should be set to 5 and here should add 1 to currentDepth because it starts at 0

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
ancaantochi: We discussed offline and when loading from storage there is no order so that would be the case when children are added before parent

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,326 @@
...
+            public void RemoveChild(ServiceIdentityTreeNode childNode)
+            {
+                this.children.Remove(childNode);
+                childNode.Parent = Option.None<ServiceIdentityTreeNode>();
+                childNode.RemoveAuthChain();
+            }
+
+            public void RemoveAllChildren()
+            {
+                var snapshot = new List<ServiceIdentityTreeNode>(this.children);
```
ancaantochi, line 266: can it just loop through the list and update the child and then set this.children to empty list?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -73,14 +74,14 @@ public Task<ScopeResult> GetIdentity(string targetDeviceId, string targetModuleI
...
-                        string relativeUri = InScopeIdentitiesUriTemplate.FormatInvariant(this.deviceId, this.moduleId, this.batchSize, null, ApiVersion);
+                        string relativeUri = InScopeIdentitiesUriTemplate.FormatInvariant(this.TargetEdgeDeviceId, this.moduleId, this.batchSize, null, ApiVersion);
                         var uri = new Uri(this.iotHubBaseHttpUri, relativeUri);
                         return uri;
                     });
 
         internal Uri GetServiceUri(string targetDeviceId, string targetModuleId)
         {
-            string relativeUri = InScopeTargetIdentityUriFormat.FormatInvariant(this.deviceId, this.moduleId, targetDeviceId, targetModuleId, ApiVersion);
+            string relativeUri = InScopeTargetIdentityUriFormat.FormatInvariant(this.TargetEdgeDeviceId, this.moduleId, targetDeviceId, targetModuleId, ApiVersion);
```
ancaantochi, line 84: it is a little confusing here using TargetEdgeDeviceId and targetDeviceId, either renamse TargetEdgeDeviceId  to edgeOnBehalfOfDeviceId or similar or targetDeviceId is more like requested identity.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            var transientRetryPolicy = new RetryPolicy(TransientErrorDetectionStrategy, retryStrategy);
+            transientRetryPolicy.Retrying += (_, args) => onRetry(args);
+            return transientRetryPolicy.ExecuteAsync(func);
+        }
+
+        async Task<ScopeResult> GetIdentitiesInTargetScopeWithRetry(Uri uri, Option<string> continuationLink)
+        {
+            // If the caller supplied a continuation token, then it overrides the cached token
+            Option<string> continuation = continuationLink.HasValue ? continuationLink : this.continuationLink;
```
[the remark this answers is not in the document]
richma-ms: right now none, but I left it as an option to conform to the existing interface definition.  When we remove the legacy DeviceScopeApiClient, we can then update the interface to fit the nested scenario only and remove this param.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -0,0 +1,326 @@
...
+                this.nodes.Remove(id);
+            }
+        }
+
+        Option<ServiceIdentityTreeNode> FindDeviceByScopeId(string scopeId)
+        {
+            Preconditions.CheckNonWhiteSpace(scopeId, nameof(scopeId));
+
+            // Look for Edge devices with a matching scope ID
+            List<ServiceIdentityTreeNode> devices =
```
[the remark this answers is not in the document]
richma-ms: the code will prob be more bug prone, generally I'd like to avoid parsing anything if the alternative isn't very expensive.  In this case, I don't think the linear search is that bad since we filter by Edge devices first, and that's just a boolean check.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/IServiceIdentityTree.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Identity.Service;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public interface IServiceIdentityTree
```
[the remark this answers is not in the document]
richma-ms: any suggestions? I feel like IServiceIdentityCache would be way to similar to IDeviceScopeIdentitiesCache...

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/IDeviceScopeApiClientProvider.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
+{
+    using Azure.Devices.Edge.Util;
+
+    public interface IDeviceScopeApiClientProvider
```
ancaantochi, line 6: this interface should have 1 method CreateDeviceScopeClient() and implementation based on the feature if it is turned on or not.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -0,0 +1,224 @@
...
+        async Task<ScopeResult> GetIdentitiesInTargetScope(Uri uri, Option<string> continuationToken)
+        {
+            HttpClient client = this.proxy
+                .Map(p => new HttpClient(new HttpClientHandler { Proxy = p }, disposeHandler: true))
+                .GetOrElse(() => new HttpClient());
+            using (var msg = new HttpRequestMessage(HttpMethod.Post, uri))
+            {
+                // Get the auth-chain for the target device
+                Option<string> maybeAuthChain = await this.serviceIdentityTree.GetAuthChain(this.TargetEdgeDeviceId);
+                string authChain = maybeAuthChain.Expect(() => new ArgumentException($"No valid authentication chain for {this.TargetEdgeDeviceId}"));
```
ancaantochi, line 138: InvalidOperationException

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -12,21 +12,25 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
...
-        readonly IDeviceScopeApiClient securityScopesApiClient;
+        readonly IDeviceScopeApiClientProvider securityScopesApiClientProvider;
+        readonly IDeviceScopeApiClient defaultSecurityScopesApiClient;
 
-        public ServiceProxy(IDeviceScopeApiClient securityScopesApiClient)
+        public ServiceProxy(IDeviceScopeApiClientProvider securityScopesApiClientProvider)
         {
-            this.securityScopesApiClient = Preconditions.CheckNotNull(securityScopesApiClient, nameof(securityScopesApiClient));
+            this.securityScopesApiClientProvider = Preconditions.CheckNotNull(securityScopesApiClientProvider, nameof(securityScopesApiClientProvider));
+            this.defaultSecurityScopesApiClient = this.securityScopesApiClientProvider.CreateDeviceScopeClient();
```
ancaantochi, line 21: if IDeviceScopeApiClientProvider knows only to CreateDeviceScopeClient and based in implementation return one or the other we don't need this

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -12,21 +12,25 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
...
         }
 
-        public IServiceIdentitiesIterator GetServiceIdentitiesIterator() => new ServiceIdentitiesIterator(this.securityScopesApiClient);
+        public IServiceIdentitiesIterator GetServiceIdentitiesIterator() => new ServiceIdentitiesIterator(this.defaultSecurityScopesApiClient);
```
ancaantochi, line 24: same, it should only have GetServiceIdentitiesIterator and maybe modify the IDeviceScopeApiClientProvider  to also create iterator. I see the iterator implementation is not public here, but I think it could be made public

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -13,47 +13,58 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
...
         readonly IServiceProxy serviceProxy;
         readonly IKeyValueStore<string, string> encryptedStore;
-        readonly AsyncLock cacheLock = new AsyncLock();
-        readonly IDictionary<string, StoredServiceIdentity> serviceIdentityCache;
+        readonly IServiceIdentityTree serviceIdentityTree;
         readonly Timer refreshCacheTimer;
         readonly TimeSpan refreshRate;
         readonly AsyncAutoResetEvent refreshCacheSignal = new AsyncAutoResetEvent();
         readonly object refreshCacheLock = new object();
+        readonly bool nestedEdgeEnabled;
```
ancaantochi, line 26: deviceidentititescache should not have nestedEdgeEnabled, serviceProxy internally should return the iterator

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/IServiceIdentityTree.cs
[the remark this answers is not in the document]
ancaantochi: it would be confusing to have both with cache, but maybe IDeviceScopeIdentitiesCache should use Manager or CacheManager?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
richma-ms: I don't think that saves us anything, right?  It'd still need to go through and recursively remove the authchain on each grandchildren, etc.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/service/ServiceIdentity.cs
```diff
@@ -48,9 +52,19 @@ public class ServiceIdentity : IEquatable<ServiceIdentity>
         [JsonConverter(typeof(OptionConverter<string>))]
         public Option<string> ModuleId { get; }
 
+        [JsonProperty("deviceScope")]
+        [JsonConverter(typeof(OptionConverter<string>))]
+        public Option<string> DeviceScope { get; }
+
+        [JsonProperty("parentScopes")]
+        public ISet<string> ParentScopes { get; }
```
ancaantochi, line 60: IotHub returns a IList, it should be in theory a set, but just in case that changes should use a list here too

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/IServiceIdentityTree.cs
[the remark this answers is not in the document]
richma-ms: IServiceIdentityHierarchy

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/DeviceScopeApiClient.cs
```diff
@@ -18,7 +18,7 @@ public class DeviceScopeApiClient : IDeviceScopeApiClient
 
         const string InScopeIdentitiesUriTemplate = "/devices/{0}/modules/{1}/devicesAndModulesInDeviceScope?deviceCount={2}&continuationToken={3}&api-version={4}";
 
-        const string InScopeTargetIdentityUriFormat = "/devices/{0}/modules/{1}/deviceAndModuleInDeviceScope?targetDeviceId={2}&targetModuleId={3}&api-version={4}";
+        const string InScopeTargetIdentityUriFormat = "/devices/{0}/modules/{1}/deviceAndModuleInDeviceScope?requestedDeviceId={2}&requestedModuleId={3}&api-version={4}";
```
ancaantochi, line 21: query parameter name shouldn't be changed here because IotHub uses targetDeviceId

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
richma-ms: talked offline, moved this to ServiceProxy

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        public NestedDeviceScopeApiClient(
+            string iotHubHostName,
+            string deviceId,
+            string targetDeviceId,
+            string moduleId,
+            Option<string> continuationToken,
+            int batchSize,
+            ITokenProvider edgeHubTokenProvider,
+            IServiceIdentityHierarchy serviceIdentityTree,
```
ancaantochi, line 66: nit: rename to serviceIdentityHierarchy

PR #2988, opened 2020-05-21
Fix PriorityQueue e2e tests from dotnet 3.1 update 

dylanbronson:
This PR does 4 things. I feel it's best that it's put together, though, since they all work together. 

1. Changes TRC to take specific settings for its two services that it uses on Startup. The tests broke with the dotnet 3.1 update because now if an ASPNET service fails, the whole TRC will fail, whereas previously one failed service didn't matter for the rest of the TRC. So it has specific settings for each service, which also makes the settings class cleaner. I can now remove unnecessary env vars from PriorityQueue tests as well.
2. Prevent nulls from the Hub. The PriQ tests were failing on receiving a message that was missing expected fields during development. I'm not sure why this happened, and I've never seen it happen in the pipeline, so it must have just been something weird during development, but it's still helpful to log those messages as a warning and not crash.
3. Update start delay time for Module to Hub test. I need to do more investigation into making this timing-proof, but for now this fixes the problem and is better than a broken test.
4. I needed a ForEach method from Option that didn't take a parameter. I could have used the other ones that were there and just used an empty parameter, but it felt cleaner to have a dedicated method in Option.cs for this purpose, so I added it. It can easily be removed if anyone disagrees


edge-util/src/Microsoft.Azure.Devices.Edge.Util/Option.cs
```diff
@@ -140,6 +140,14 @@ public void ForEach(Action<T> action)
...
 
+        public void ForEach(Action action)
+        {
+            if (this.HasValue)
+            {
+                action();
+            }
+        }
+
         public void ForEach(Action<T> action, Action none)
```
philipktlin, line 151: is it possible that 2nd parameter be optional with default value "empty action"?

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -136,16 +136,25 @@ async Task ReceiveEventsFromIotHub(DateTime startTime, ConcurrentQueue<MessageTe
...
+                                });
+                                results.Add(sequenceNumber);
+                            }
+                            else
                             {
-                                TrackingId = data.Properties["trackingId"].ToString(),
-                                BatchId = data.Properties["batchId"].ToString(),
-                                SequenceNumber = data.Properties["sequenceNumber"].ToString()
-                            });
+                                Log.Verbose("Message is missing information. Needs to have trackingId, batchId, and sequenceNumber. Not enqueuing.");
```
damonbarry, line 155: I think there's a `Log.Warning()` method...is that a better choice here?

edge-util/src/Microsoft.Azure.Devices.Edge.Util/Option.cs
dylanbronson: "Action none = () => { }" doesn't work because it needs to be compile-time constant, but I don't think that even fits the specific case here - the case where I need to perform an action if Option.HasValue, but I don't need the value as the parameter.

PR #2991, opened 2020-05-22
Use BTreeMap instead of HashMap for config.yaml's hash

damonbarry:
`iotedged` detects changes to `config.yaml` by creating an hash of it's in-memory representation, saving it to disk, and comparing it against the latest next time it starts up. The problem is that several structs which make up our in-memory representation of `config.yaml` use `HashMap` as a dictionary. `HashMap` cannot guarantee that different maps with the same keys will serialize in the same order, so two semantically equivalent `HashMap`s might not compare equal. The impact is that `iotedged` might think that `config.yaml` changed (and remove/restart all modules as a result) when it didn't.

This change moves the offending types to `BTreeMap`, which has the guarantee we need. It also removes the custom `serialize_ordered` attribute on `HashMap`, which just copies the `HashMap` to a `BTreeMap` anyway so using `BTreeMap` to begin with is simpler.

### DIFF
--- a/edgelet/docker-rs/src/models/container_create_body.rs
+++ b/edgelet/docker-rs/src/models/container_create_body.rs
@@ -23,7 +23,7 @@ use serde_json::Value;
 //
 // To avoid having to do this for effectively the whole crate, instead we've just commented out the fields we don't use in our code.
 //
-// Note: We're using BTreeMap instead of HashMap because iotedged stores an HMAC of its local config (whose object representation uses this struct)
+// Note: We're using BTreeMap instead of HashMap because iotedged stores a hash of its local config (whose object representation uses this struct)
 // to detect changes. Since different HashMaps with the same keys aren't guaranteed to serialize in the same order (and thus won't compare equal),
 // we need to use another map type that can provide that guarantee.
 //
--- a/edgelet/docker-rs/src/models/container_create_body.rs
+++ b/edgelet/docker-rs/src/models/container_create_body.rs
@@ -16,12 +16,16 @@ use serde_json::Value;
 // We do not want to restrict the properties that the user can set in their create options, because future versions of Docker can add new properties
 // that we don't define here.
 //
-// So this type has a `#[serde(flatten)] HashMap` field to collect all the extra properties that we don't have a struct field for.
+// So this type has a `#[serde(flatten)] BTreeMap` field to collect all the extra properties that we don't have a struct field for.
 //
 // But if an existing field references another type under `crate::models::`, then that would still be parsed lossily, so we would have to also add
-// a `#[serde(flatten)] HashMap` field there. And if that type has fields that reference types under `crate::models::` ...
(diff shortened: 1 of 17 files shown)


damonbarry:
Test failures in #2989 should resolve once this change goes in.

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -717,8 +717,13 @@ impl ModuleRuntime for DockerModuleRuntime {
...
                                 DockerConfig::new(
                                     container.image().to_string(),
-                                    ContainerCreateBody::new()
-                                        .with_labels(container.labels().clone()),
+                                    ContainerCreateBody::new().with_labels(
+                                        container
+                                            .labels()
+                                            .iter()
+                                            .map(|(k, v)| (k.to_string(), v.to_string()))
+                                            .collect::<BTreeMap<_, _>>(),
```
arsing, line 725: Don't need the turbofish. `collect()`'s return type will be inferred from the `with_labels` signature.

edgelet/docker-rs/src/models/container_create_body.rs
```diff
@@ -16,20 +16,24 @@ use serde_json::Value;
...
-// So this type has a `#[serde(flatten)] HashMap` field to collect all the extra properties that we don't have a struct field for.
+// So this type has a `#[serde(flatten)] BTreeMap` field to collect all the extra properties that we don't have a struct field for.
 //
 // But if an existing field references another type under `crate::models::`, then that would still be parsed lossily, so we would have to also add
-// a `#[serde(flatten)] HashMap` field there. And if that type has fields that reference types under `crate::models::` ...
+// a `#[serde(flatten)] BTreeMap` field there. And if that type has fields that reference types under `crate::models::` ...
 //
 // To avoid having to do this for effectively the whole crate, instead we've just commented out the fields we don't use in our code.
 //
+// Note: We're using BTreeMap instead of HashMap because iotedged stores an HMAC of its local config (whose object representation uses this struct)
```
arsing, line 26: `s/an HMAC of/a hash of/g`

HMAC takes a hash and signs it with a key. We just do a hash.

arsing:
@ancaantochi FYI. This removes the function you added in https://github.com/Azure/iotedge/commit/0e6a402fecf9f11c3f8afff7713352ddc165a234 to sort `agent.env` at serialization time, by changing `ModuleSpec::env` to be a `BTreeMap` directly.

edgelet/docker-rs/src/models/container_create_body.rs
damonbarry: Oops, thanks

edgelet/edgelet-docker/src/runtime.rs
damonbarry: Fixed.

edgelet/docker-rs/src/models/container_create_body.rs
damonbarry: Fixed everywhere (including PR title/description).

PR #2993, opened 2020-05-22
EdgeHub & EdgeAgent would shutdown if Workload API goes stale

yophilav:
This is an improvement to module reliability when the workload API goes stale. 
Currently both edgeAgent and edgeHub would get stuck if the iotedged got killed unexpected. 
The two module would forever hand trying to communicate via a "stale" iotedged socket. 
This PR fixes the problem by having both edgeAgent and edgeHub exit upon receiving a specific, configurable error code. This will allow the two module to be brought up with a new iotedged socket. 

Reproduce steps: 
1. Rename and kill the iotedged
- `sudo mv /usr/bin/iotedged /usr/bin/iotedged1`
- `ps aux | grep iotedge`
- `sudo kill -9 <PIDofIoTEdge>`
NOTE: edgeAgent should be dead soon after this.
2. Disconnect the internet
```
sudo tc qdisc add dev  <dockerNetDev> root handle 1: prio priomap 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 && \
sudo tc qdisc add dev  <dockerNetDev> parent 1:2 handle 20: netem loss 100% delay 0ms 0ms rate 0mbit && \
sudo tc filter add dev  <dockerNetDev> parent 1:0 protocol ip u32 match ip src <IoTHubIpAddr> flowid 1:2 && \
sudo tc qdisc show dev  <dockerNetDev>
```
3. `docker logs edgeHub` make sure message and things are still being recorded
4. `docker restart tempSensor`
5. Wait for  (>1min) a bit for the IoTHub proxy to die of old age
6. Reconnect the internet
- `sudo tc qdisc delete dev  <dockerNetDev> root`
Note: 
- This should trigger the edgeHub to try to SignAsync for the tempSensor with IoTHub. The process uses the old iotedged socket -- edgeHub realizes that it is using the old socket and decided to jump off the cliff -- This is where edgeHub dies.
7. `sudo mv /usr/bin/iotedged1 /usr/bin/iotedged && sudo systemctl start iotedge`
NOTE: 
- This should bring up edgeAgent with a new iotedged socket
- edgeAgent revives edgeHub to use a new iotedged socket.
8. docker logs edgeHub


edge-util/src/Microsoft.Azure.Devices.Edge.Util/edged/version_2018_06_28/WorkloadClient.cs
```diff
@@ -11,9 +11,6 @@ namespace Microsoft.Azure.Devices.Edge.Util.Edged.Version_2018_06_28
 
     class WorkloadClient : WorkloadClientVersioned
     {
-        static readonly RetryStrategy TransientRetryStrategy =
```
philipktlin, line 14: good cleanup 👍

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -23,14 +23,18 @@ abstract class ModuleManagementHttpClientVersioned
         const string LogsUrlTailParameter = "tail";
         const string LogsUrlSinceParameter = "since";
 
+        const int MaxRetryCount = 4;
```
philipktlin, line 26: why 4?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -103,6 +108,12 @@ public virtual async Task<Stream> GetModuleLogs(string module, bool follow, Opti
 
         protected abstract void HandleException(Exception ex, string operation);
 
+        bool IsModuleManagementSocketStale(Exception ex)
+        {
+            return (string.Compare(ex.GetType().ToString(), "System.Net.Internals.SocketExceptionFactory+ExtendedSocketException") == 0)
```
philipktlin, line 113: catch socket exception first and check exception message. https://stackoverflow.com/questions/56909680/how-can-i-catch-an-extendedsocketexception. Also please ignore case when string comparison.

Also does .net core 3.1 throw the same exception?  Andrew's .net core 3.1 upgrade PR will be committed today.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -103,6 +108,12 @@ public virtual async Task<Stream> GetModuleLogs(string module, bool follow, Opti
...
+                && ex.Message.ToString().Contains("Connection refused");
```
philipktlin, line 114:
```suggestion
                && ex.Message.Contains("Connection refused");
```

philipktlin: add case insensitive contains in stringex.cs https://dotnetthoughts.net/how-to-make-string-contains-case-insensitive/, and use here.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -117,14 +128,27 @@ protected internal async Task<T> Execute<T>(Func<Task<T>> func, string operation
             try
             {
                 Events.ExecutingOperation(operation, this.ManagementUri.ToString());
-                T result = await ExecuteWithRetry(func, r => Events.RetryingOperation(operation, this.ManagementUri.ToString(), r), this.transientErrorDetectionStrategy)
+                T result = await ExecuteWithRetry(
+                    func,
+                    (r) => Events.RetryingOperation(operation, this.ManagementUri.ToString(), r),
+                    this.transientErrorDetectionStrategy)
                     .TimeoutAfter(this.operationTimeout);
```
philipktlin, line 135: check with Varun what happens if timeout happens, should we do the same?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -117,14 +128,27 @@ protected internal async Task<T> Execute<T>(Func<Task<T>> func, string operation
...
             }
             catch (Exception ex)
             {
                 Events.ErrorExecutingOperation(ex, operation, this.ManagementUri.ToString());
+                if (this.IsModuleManagementSocketStale(ex))
+                {
+                    this.failedSocketCount++;
+                    if (this.failedSocketCount >= MaxRetryCount)
+                    {
+                        Environment.Exit(0);
```
philipktlin, line 148: Add Log.Error with specific event id and message.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/edged/WorkloadClientVersioned.cs
```diff
@@ -8,10 +8,13 @@ namespace Microsoft.Azure.Devices.Edge.Util.Edged
...
     abstract class WorkloadClientVersioned
     {
+        const int MaxRetryCount = 3;
+        int failedSocketCount = 0;
+
         static readonly TimeSpan DefaultOperationTimeout = TimeSpan.FromMinutes(5);
 
         static readonly RetryStrategy TransientRetryStrategy =
-            new ExponentialBackoff(retryCount: 2, minBackoff: TimeSpan.FromSeconds(1), maxBackoff: TimeSpan.FromSeconds(3), deltaBackoff: TimeSpan.FromSeconds(2));
+            new ExponentialBackoff(retryCount: MaxRetryCount, minBackoff: TimeSpan.FromSeconds(1), maxBackoff: TimeSpan.FromSeconds(3), deltaBackoff: TimeSpan.FromSeconds(2));
```
philipktlin, line 17: I think ExponentialBackoff in workload client and module management client should initialize with same values.  Please sync up with @ancaantochi what values should be used (e.g. minBackoff, maxBackoff).

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -41,6 +45,7 @@ abstract class ModuleManagementHttpClientVersioned
             this.Version = Preconditions.CheckNotNull(version, nameof(version));
             this.transientErrorDetectionStrategy = transientErrorDetectionStrategy;
             this.operationTimeout = operationTimeout.GetOrElse(DefaultOperationTimeout);
+            this.failedSocketCount = 0;
```
philipktlin, line 48: Not needed; already set to 0 in line 36.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/edged/WorkloadClientVersioned.cs
```diff
@@ -30,6 +33,7 @@ abstract class WorkloadClientVersioned
             this.ModuleGenerationId = Preconditions.CheckNonWhiteSpace(moduleGenerationId, nameof(moduleGenerationId));
             this.transientErrorDetectionStrategy = transientErrorDetectionStrategy;
             this.operationTimeout = operationTimeout.GetOrElse(DefaultOperationTimeout);
+            this.failedSocketCount = 0;
```
philipktlin, line 36: Not needed.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/edged/WorkloadClientVersioned.cs
```diff
@@ -55,13 +59,27 @@ protected internal async Task<T> Execute<T>(Func<Task<T>> func, string operation
...
+                    this.failedSocketCount++;
+                    Events.ErrorExecutingOperation(ex, operation, this.WorkloadUri.ToString());
+                    if (this.failedSocketCount >= MaxRetryCount)
+                    {
+                        Environment.Exit(0);
+                    }
+                }
+
                 this.HandleException(ex, operation);
                 Events.SuccessfullyExecutedOperation(operation, this.WorkloadUri.ToString());
```
philipktlin, line 84: Remove it to fix this legacy bug and call Events.ErrorExecutingOperation first right after caught instead of line76 same as what did in Module management client.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
yophilav: I'm just using whatever the value was there and then +1 it.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
yophilav: I don't know what exception Dotnet3.1 is going to throw. I'll have to test it out.. Can you point me to the PR please? I'd like to get the testing going asap. 

I certainly can also added the `if(ex is SocketException)`

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
yophilav: I think it should be the exact match, error message won't simply change the casing for itself.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
yophilav: Sure. I can use `110` error code that it was throwing. I thought that we capture and handle the error correctly; hence, it should just shutdown with error code `0`

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
yophilav: I'm paranoid.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
yophilav: ` Log.Error` is done by `Events.ErrorExecutingOperation`

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -103,6 +108,13 @@ public virtual async Task<Stream> GetModuleLogs(string module, bool follow, Opti
 
         protected abstract void HandleException(Exception ex, string operation);
 
+        bool IsModuleManagementSocketStale(Exception ex)
+        {
+            return (ex is SocketException)
+                && (string.Compare(ex.GetType().ToString(), "System.Net.Internals.SocketExceptionFactory+ExtendedSocketException") == 0)
```
varunpuranik, line 114: This is not a good idea. We should just use the most derived public exception. 
We should not take a dependency on internal things like Internal Exception type or message.

yophilav: the most derived I can get would be `(ex is SocketException)` but that feels a little too generic.

yophilav:
The test with `netcoreapp3.1` build is passing.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
philipktlin: Please follow https://stackoverflow.com/questions/56909680/how-can-i-catch-an-extendedsocketexception to catch SocketException and check exception message.

yophilav: I think that's what I'm doing here. 
1. Check the exception is type of `SocketException`
2. Verify the messages
2.1 Make sure the type is of `System.Net.Internals.SocketExceptionFactory+ExtendedSocketException`
2.2 Observe the message that it is `Connection refused`

philipktlin: As what Varun said you don't need to do 2.1.

philipktlin: @varunpuranik please confirm if 2.2 is ok.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -117,14 +129,27 @@ protected internal async Task<T> Execute<T>(Func<Task<T>> func, string operation
...
             }
             catch (Exception ex)
             {
                 Events.ErrorExecutingOperation(ex, operation, this.ManagementUri.ToString());
+                if (this.IsModuleManagementSocketStale(ex))
+                {
+                    this.failedSocketCount++;
+                    if (this.failedSocketCount >= MaxRetryCount)
+                    {
+                        Environment.Exit((ex as SocketException)?.ErrorCode ?? 110);
```
philipktlin, line 149: You need to log error message to make sure we see the reason why edgeHub exits.

yophilav: It is already taken care by the `Event.Error()` as the first thing exception is thown.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -117,14 +129,27 @@ protected internal async Task<T> Execute<T>(Func<Task<T>> func, string operation
...
+                T result = await ExecuteWithRetry(
+                    func,
+                    (r) => Events.RetryingOperation(operation, this.ManagementUri.ToString(), r),
+                    this.transientErrorDetectionStrategy)
                     .TimeoutAfter(this.operationTimeout);
                 Events.SuccessfullyExecutedOperation(operation, this.ManagementUri.ToString());
+                this.failedSocketCount = 0;
                 return result;
             }
             catch (Exception ex)
```
philipktlin, line 141: Add a new catch block for SocketException

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -117,14 +129,27 @@ protected internal async Task<T> Execute<T>(Func<Task<T>> func, string operation
...
             {
                 Events.ErrorExecutingOperation(ex, operation, this.ManagementUri.ToString());
+                if (this.IsModuleManagementSocketStale(ex))
+                {
+                    this.failedSocketCount++;
+                    if (this.failedSocketCount >= MaxRetryCount)
```
varunpuranik, line 147: How will this be true? The Execute method is called only once, to the previous step will get executed at most once. 
The retry logic is in the ExecuteWithRetry call, right?

yophilav: The Execute method is called only once -- yup, but periodically. 

The retry logic is in the ExecuteWithRetry call, right?
Not really. The stale socker is consider a non-transient and the retry routine is never called.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
yophilav: I think the error message check should be fine. Given an example across 2.1 to 3.1, the error message didn't change but the statusCode changed.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -23,14 +24,18 @@ abstract class ModuleManagementHttpClientVersioned
...
         static readonly TimeSpan DefaultOperationTimeout = TimeSpan.FromMinutes(5);
 
         static readonly RetryStrategy TransientRetryStrategy =
-            new ExponentialBackoff(retryCount: 3, minBackoff: TimeSpan.FromSeconds(2), maxBackoff: TimeSpan.FromSeconds(30), deltaBackoff: TimeSpan.FromSeconds(3));
+            new ExponentialBackoff(retryCount: MaxRetryCount, minBackoff: TimeSpan.FromSeconds(2), maxBackoff: TimeSpan.FromSeconds(30), deltaBackoff: TimeSpan.FromSeconds(3));
 
         readonly ITransientErrorDetectionStrategy transientErrorDetectionStrategy;
         readonly TimeSpan operationTimeout;
 
+        int failedSocketCount = 0;
```
ancaantochi, line 37: why is this needed? I tough it can not recover from this exception that's why it needs to restart the module. If needed it should be atomic because methods from this class can be called from different threads. There is an AtomicLong implementation in Utils that you can use.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/edged/WorkloadClientVersioned.cs
```diff
@@ -55,21 +59,51 @@ protected internal async Task<T> Execute<T>(Func<Task<T>> func, string operation
...
+                T result = await ExecuteWithRetry(
+                    func,
+                    r => Events.RetryingOperation(operation, this.WorkloadUri.ToString(), r),
+                    this.transientErrorDetectionStrategy)
                     .TimeoutAfter(this.operationTimeout);
                 Events.SuccessfullyExecutedOperation(operation, this.WorkloadUri.ToString());
+                this.failedSocketCount = 0;
                 return result;
             }
+            catch (SocketException ex)
```
ancaantochi, line 71: catch (SocketException ex) when (ex.Message.Contains("Connection refused") so IsWorkloadSocketStale is not needed

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
yophilav: I could show you my test logs that it is behaving as expected.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
yophilav: Sure I can make it failed right away. 
I wasn't sure if there is other type of Socket failure that would cause the "connection refused' that may recover.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -117,11 +118,20 @@ protected internal async Task<T> Execute<T>(Func<Task<T>> func, string operation
             try
             {
                 Events.ExecutingOperation(operation, this.ManagementUri.ToString());
-                T result = await ExecuteWithRetry(func, r => Events.RetryingOperation(operation, this.ManagementUri.ToString(), r), this.transientErrorDetectionStrategy)
+                T result = await ExecuteWithRetry(
+                    func,
+                    (r) => Events.RetryingOperation(operation, this.ManagementUri.ToString(), r),
+                    this.transientErrorDetectionStrategy)
```
varunpuranik, line 124: Looks like the error detection strategy won't retry on SocketException..

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/versioning/ModuleManagementHttpClientVersioned.cs
```diff
@@ -117,11 +118,20 @@ protected internal async Task<T> Execute<T>(Func<Task<T>> func, string operation
...
                     .TimeoutAfter(this.operationTimeout);
                 Events.SuccessfullyExecutedOperation(operation, this.ManagementUri.ToString());
                 return result;
             }
+            catch (SocketException ex) when (ex.Message.Contains("Connection refused"))
```
varunpuranik, line 129: We should also not use the exception message, since we are doing a rather drastic step.. so we need to be more careful

philipktlin: Suggest to use SocketException.ErrorCode to compare a configurable value (int?) and add default value to appsettings json.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/edged/WorkloadClientVersioned.cs
```diff
@@ -11,7 +12,7 @@ abstract class WorkloadClientVersioned
         static readonly TimeSpan DefaultOperationTimeout = TimeSpan.FromMinutes(5);
 
         static readonly RetryStrategy TransientRetryStrategy =
-            new ExponentialBackoff(retryCount: 2, minBackoff: TimeSpan.FromSeconds(1), maxBackoff: TimeSpan.FromSeconds(3), deltaBackoff: TimeSpan.FromSeconds(2));
+            new ExponentialBackoff(retryCount: 3, minBackoff: TimeSpan.FromSeconds(1), maxBackoff: TimeSpan.FromSeconds(3), deltaBackoff: TimeSpan.FromSeconds(2));
```
philipktlin, line 15: TransientRetryStrategy should be configured with same values in here and ModuleManagementHttpClientVersioned.cs.

yophilav: Yup. I checked with Anca. Both of the are now set to 3

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Constants.cs
```diff
@@ -99,6 +99,8 @@ public static class Constants
 
         public const string WorkloadApiVersion = "2019-01-30";
 
+        public const string WorkloadStaleSocketExceptionKey = "WorkloadStaleSocketExceptionCode";
```
varunpuranik, line 102: Where is this coming from?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/appsettings_agent.json
```diff
@@ -25,5 +25,6 @@
   "EdgeModuleVolumeName": "",
   "EdgeHubVolumePath": "",
   "EdgeModuleVolumePath": "",
-  "ConfigRefreshFrequencySecs": "3600"
+  "ConfigRefreshFrequencySecs": "3600",
+  "WorkloadStaleSocketExceptionCode": 111
```
varunpuranik, line 29: Where is the value 111 coming from? 
How do we know iotedged will throw this error code?

philipktlin:
LGTM 👍 please confirm with Varun.

PR #3000, opened 2020-05-23
Identity and key service APIs

massand:
- Add identity service and key service API implementations
- Launch both services from iotedged
- UTs
- Configuration for new identity and key services
- Packaging changes


edgelet/api/keyServiceVersion_2020_06_01.yaml
```diff
@@ -0,0 +1,88 @@
...
+            $ref: '#/definitions/ErrorResponse'
+
+definitions:
+  SignRequest:
+    type: object
+    properties:
+      keyHandle:
+        type: string
+        description: Key handle returned by Identity Service.
+      algo:
```
arsing, line 44: Let's be consistent and use full words for everything. So `algorithm`

edgelet/api/keyServiceVersion_2020_06_01.yaml
```diff
@@ -0,0 +1,88 @@
...
+        type: string
+        description: Sign algorithm to be used.
+        enum:
+          - HMACSHA256
```
arsing, line 48: `'HMAC-SHA256'`

edgelet/api/identityServiceVersion_2020_06_01.yaml
```diff
@@ -0,0 +1,102 @@
...
+      genId:
+        type: string
+        description: The generation ID of the device workload assigned by IoT Hub.
+        example: "12345"
+      auth:
+        $ref: '#/definitions/Credentials'
+    required:
+      - hubName
+      - deviceId
+      - managedBy
```
CindyXing, line 70: Why this one is required? What can callers use the "managedBy" for?

edgelet/api/identityServiceVersion_2020_06_01.yaml
```diff
@@ -0,0 +1,102 @@
+swagger: '2.0'
+schemes:
+  - http
+info:
+  title: Identity Service API
+  version: '2020-02-02'
+
+paths:
+  '/identities/identity':
```
CindyXing, line 9: What do you think about /current-identity as the path?

edgelet/api/identityServiceVersion_2020_06_01.yaml
```diff
@@ -0,0 +1,102 @@
+swagger: '2.0'
+schemes:
+  - http
```
CindyXing, line 3: What's the reason not using https?

edgelet/api/identityServiceVersion_2020_06_01.yaml
```diff
@@ -0,0 +1,102 @@
...
+  ErrorResponse:
+    type: object
+    properties:
+      message:
+        type: string
+    required:
+      - message
+
+parameters:
+  api-version:
```
CindyXing, line 96: In general, what's our thought to include api version in the base path vs parameter?

edgelet/api/identityServiceVersion_2020_06_01.yaml
massand: This is required in the future. So, we are including it now so callers don't need to change their code for this.

edgelet/api/identityServiceVersion_2020_06_01.yaml
massand: /identities/identity is a more REST-like name for the identity resource. It is also the name for the future IS API. Also, /current-identity has no meaning because it is always the identity of the caller. 

We are preserving /identities in the URL because we know there will be a backwards-compatible API that will be exposed for iotedged in the future.

edgelet/api/identityServiceVersion_2020_06_01.yaml
massand: I believe this is because we don't do TLS over UDS. It is simply HTTP over UDS with peer cred auth. @arsing can correct me if I am wrong here.

edgelet/api/identityServiceVersion_2020_06_01.yaml
massand: It is a convention in existing APIs for backwards compat, in case we decide to change API on server. The query parameter includes the intended API version the caller is targeting.

arsing: I think Cindy is asking about `/v1/foo/bar` vs `/foo/bar/?api-version=1`. If so, we already do the latter for existing Edge API, and I assume we do those for consistency with the Azure cloud services' URIs. So I think we should stay consistent with those.

edgelet/api/identityServiceVersion_2020_06_01.yaml
arsing: Yes, there's no point in doing TLS over UDS.

The only time TLS comes into play is for Kubernetes, because IS/KS/CS will have to bind to TCP, just like iotedged does today for the management and workload servers. But I don't think we need to represent that here in the spec. The specs are used for generating clients, and the clients don't care whether the underlying transport is UDS or TLS-over-TCP or whatever. The existing management and workload specs also just use `http`.

edgelet/api/identityServiceVersion_2020_06_01.yaml
CindyXing: Can you clarify what we use it for in the future?

massand: When module identities are generated on startup for host processes by identityservice, they will be having this field set as the current 'identityservice' enum. When module runtime requests for identities to be created for edge, they will be tagged as being managedby = 'edgeruntime' or 'moduleruntime'. This is used by 'moduleruntime' to delete identities it created, but it can still query for identities created by 'identityservice'.

edgelet/api/identityServiceVersion_2020_06_01.yaml
CindyXing: It doesn't seem to me a valid reason. 

First of all, ADU doesn't plan to use UDS. Secondly we are assuming processes run on the same host, which is not the case in distributed scenarios: cluster or nested. For example, a child edge asks its parent for local identity. 

From client's perspective, it does need to know if the API server supports http only or http/https. And the schemes is for that purpose.

edgelet/api/identityServiceVersion_2020_06_01.yaml
CindyXing: /identities/identity is not a standard REST like path. /identities/{id} is. 

To understand this API, aren't we returning result based on the caller?

edgelet/api/identityServiceVersion_2020_06_01.yaml
CindyXing: Yes. I meant the API version scheme. 

It is true that azure ARM uses query parameter for api version. I am seeing microsoft graph rest API is using the url path for API versioning. The benefit is to enable routing when talking about distributed system. I am ok for us to keep the version parameter and we should keep this in mind when designing cluster and nested.

edgelet/api/identityServiceVersion_2020_06_01.yaml
arsing: >First of all, ADU doesn't plan to use UDS.

I don't understand. They don't have a choice but to use UDS.

>Secondly we are assuming processes run on the same host, which is not the case in distributed scenarios: cluster or nested. For example, a child edge asks its parent for local identity.

I covered the Kubernetes cluster case in my comment. For nested edge, I'm not sure downstream devices are going to connect to upstream device's IS? I've only heard of them connecting to upstream device's Edge Hub.

>From client's perspective, it does need to know if the API server supports http only or http/https.

It doesn't matter in the clients that we generate from the spec. As I said, we already have the management and workload swagger specs, and they only use `http`, which makes no difference because the clients are generic over the transport. We can add `https` here if you insist but it's not going to make any difference to the generated clients code.

edgelet/api/identityServiceVersion_2020_06_01.yaml
CindyXing: Ok. Given module runtime is the one driving the container module identity provisioning/lifetime; and IS for host processes, what's the benefit of having this field and mark it as required? Do I miss anything?

massand: So that we can setup policies for different CRUD operations on identities based on their owner. Managedby provides that context for each identity. It is just metadata and every system that automates identity management should have a clear owner for identity lifecycle. Module runtime being an external entity to identity service needs to have authorization and context to perform CRUD operations on identities created by identity service. We don't wish to hardcode these in code and couple the two services.

edgelet/api/identityServiceVersion_2020_06_01.yaml
massand: The caller doesn't know what {id} to pass in. It is querying for that {id} with /identities/identity.
How is /identities/identity not a valid REST path? If it has a documented contract for the resource - an identity spec - that is being returned, then any canonical name (i.e. identity) is valid here. Could you point to some example you have in mind when you say it is not standard?

edgelet/api/identityServiceVersion_2020_06_01.yaml
massand: As @arsing said, UDS is the only available method of accessing this API. And we have been following the same pattern (and we don't plan to break that for now) in the new iotedged APIs.

edgelet/api/identityServiceVersion_2020_06_01.yaml
```diff
@@ -0,0 +1,102 @@
+swagger: '2.0'
+schemes:
+  - http
+info:
+  title: Identity Service API
+  version: '2020-02-02'
```
arsing, line 6: The filename says `2020-06-01`, this says `2020-02-02` and the `api-version` parameter says `2020-06-28`. They should all be consistent. Same for the KS file.

massand: Copy pasta errors from another swagger. Thanks for catching this.

edgelet/api/identityServiceVersion_2020_06_01.yaml
CindyXing: For policy, shouldn't it be applied based on identity itself? Let's add it when it is clearer especially not having it as "required". Thanks.

edgelet/api/identityServiceVersion_2020_06_01.yaml
CindyXing: REST assumes resource for each API path. It encourages either plural or single. In general, folks use plural. 

Here the "/identities/identity" means the current identity of the caller. I see /currentIdentity is clearer to the caller and know what he/she is getting. 

Some of the examples:
https://docs.oracle.com/cd/E69533_01/Platform.11-3/ATGWSFrameGuide/html/s0519usingswaggerannotations01.html

https://app.swaggerhub.com/apis/Floktu/Floktu-API/3.0

massand: I understand the point. The current end of the path (/identity) is still a valid resource in the current proposal. The nesting of identity resource under /identities is the main concern in this discussion. I will change it to just /identity because current implies it is a current state (which cannot be verified by identity service alone because it doesn't actually connect - DPS tells it during provisioning and the client will need to connect to the cloud endpoint (Hub) with a token signed). Just authenticating with IS isn't enough for client to consider it a valid current identity. I will make this change and check in. That work?

CindyXing: Sounds good. Thanks.

edgelet/api/identityServiceVersion_2020_06_01.yaml
massand: Removed this for now. It can be revisited later I agree.

edgelet/edgelet-http-keyservice/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+    ) -> impl Future<Item = Self, Error = Error>
+    where
+        K: 'static + KeyStore + Clone + Sync + Send,
+        K::Key: Sign,
+        M: ModuleRuntime + Authenticator<Request = Request<Body>> + Clone + Send + Sync + 'static,
+        W: WorkloadConfig + Clone + Send + Sync + 'static,
+        <M::AuthenticateFuture as Future>::Error: Fail,
+    {
+        let router = router!(
+            get   Version2020_06_01 runtime Policy::Anonymous => "/sign" => SignHandler::new(key_store.clone()),
```
massand, line 47: Fixed - changed get to post

edgelet/iotedged/src/lib.rs
```diff
@@ -1478,14 +1482,29 @@ where
 
     let workload = start_workload::<_, _, _, _, M>(
         settings,
-        key_store,
+        &key_store.clone(),
```
CindyXing, line 1485: What does the '&' mean? Why not do the same for cert_manager and workload_config.clone()?

edgelet/edgelet-http-keyservice/src/error.rs
```diff
@@ -0,0 +1,107 @@
...
+use hyper::header::{CONTENT_LENGTH, CONTENT_TYPE};
+use log::error;
+
+#[derive(Debug)]
+pub struct Error {
+    inner: Context<ErrorKind>,
+}
+
+#[derive(Clone, Debug, Fail)]
+pub enum ErrorKind {
```
CindyXing, line 18: Give IS, KS, CS and MR are REST servers, can some of the code in error.rs and lib.rs be reused?

edgelet/iotedged/src/lib.rs
massand: Without clone, the program can no longer borrow that variable in future calls.

edgelet/edgelet-http-keyservice/src/error.rs
massand: It isn't clear what code could be reused. The error messages are different and the libs have different routes defined.

edgelet/iotedged/src/lib.rs
arsing: `&key_store.clone()` is redundant with `key_store` (no `&` and no `.clone()`). `&` borrows are already `Copy`. Your code compiles fine with just `key_store` as it was originally. It's a different matter for `cert_manager` and `workload_config` that are being taken by value and thus the first usage (ie `start_workload`) does need a `.clone()`

---

There's also the case of why `start_workload` takes a `&` of the key store when it eventually calls `WorkloadService::new` with it and that one clones it so as to have an owned value. That is a silly thing to do - if the callee needs an owned value, it should take an owned value, rather than taking a borrow and cloning it. That allows the caller to know that there are no hidden copies, and the caller could avoid the unnecessary clone if they don't need the original value any more and can just offer it directly. But anyway, it was already wrong before you change so you don't have to fix that.

edgelet/iotedged/src/lib.rs
```diff
@@ -359,6 +359,7 @@ where
                     $id_cert_thumprint,
                 )?;
 
+                //let provisioning_result = $provisioning_result.clone();
```
massand, line 362: will remove

edgelet/Makefile
```diff
@@ -104,6 +104,8 @@ deb: release
 	$(INSTALL_PROGRAM) -D $(TARGET)/iotedge $(TARGET)/$(PACKAGE)/iotedge
 	$(INSTALL_DATA) -D $(srcdir)/contrib/systemd/debian/iotedge.service $(TARGET)/$(PACKAGE)/debian/iotedge.service
 	$(INSTALL_DATA) -D $(srcdir)/contrib/systemd/debian/iotedge.socket $(TARGET)/$(PACKAGE)/debian/iotedge.socket
+	$(INSTALL_DATA) -D $(srcdir)/contrib/systemd/debian/idservice.socket $(TARGET)/$(PACKAGE)/debian/idservice.socket
```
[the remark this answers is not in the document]
massand: I will refactor this in another PR. Created a task to finalize on package and service naming.

edgelet/contrib/config/linux/config.yaml
```diff
@@ -288,6 +288,8 @@ connect:
 #     management_uri - used by the Edge Agent and 'iotedge' CLI to start,
 #                      stop, and manage modules
 #     workload_uri   - used by modules to retrieve tokens and certificates
+#     identityservice_uri   - used by host-level processes to retrieve identity information
+#     keyservice_uri - used by host-level processes to perform identity key operations
```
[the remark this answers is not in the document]
massand: Yeah these are definitely placeholder names. I will follow convention once I have final names. BTW, the 'identity_uri' really threw me off, which is why I started adding 'service'. I liked your earlier suggestion of iotid and iotkey.

PR #3002, opened 2020-05-23
Fix service deployment pipeline

damonbarry:
The service deployment pipeline is used by the IoT Hub team to try out IoT Edge against new versions of the service before they release. It uses the released versions of our bits (so, 1.0.9 currently), but it uses the end-to-end test code from `master`.

The problem is that we recently made some changes to the end-to-end tests that rely on new behavior slated for the next release. This caused the service deployment pipeline to fail against our release bits.

The fix is to use test code that matches product bits.

I also updated the pipeline YAML so that is doesn't hardcode `netcoreapp3.1`--it programmatically determines the binary path instead.

### DIFF
--- a/builds/service/service-deployment.yaml
+++ b/builds/service/service-deployment.yaml
@@ -7,7 +7,7 @@ schedules:
   displayName: Nightly build
   branches:
     include:
-    - master
+    - release/1.0.9 # TODO: Return to 'master' once 1.0.10 is released
   always: true
 
 pool:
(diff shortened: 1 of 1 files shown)


damonbarry:
According to the [Azure pipelines docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/scheduled-triggers?view=azure-devops&tabs=yaml#example-of-scheduled-triggers-for-multiple-branches), this change would need to be checked into `master` AND `release/1.0.9` before the service deployment pipeline would start running nightly on `release/1.0.9` instead of `master`. It's easier to just temporarily override with a scheduled trigger in the pipeline settings UI. Closing this PR...

PR #3009, opened 2020-05-26
Edge Hub metrics go through asp.net

lfitchett:
This solves the issue that requires running as container administrator on windows


doc/BuiltInMetrics.md
```diff
@@ -11,9 +11,11 @@ As of release 1.0.9, metrics are exposed as an experimental feature available at
 | `ExperimentalFeatures__Enabled`          | `true` |
 | `ExperimentalFeatures__EnableMetrics`    | `true` |
 
-### Windows Note
+### Windows Note (1.0.9 only)
 Metrics on Windows are not fully supported the 1.0.9 experimental release. Host metrics (cpu, memory and disk usage) will all show as 0. Moby metrics are supported. In addition, edgeHub must be started as a container administrator to expose metrics. 
 
+This is no longer required in the 1.0.10 release
```
CindyXing, line 17: Given the title is for 1.0.9 only, do we still need this line?

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
```diff
@@ -42,6 +42,12 @@ protected override async Task Test(CancellationToken cancellationToken)
                 expected.Remove("edgeAgent_created_pids_total");
             }
 
+            if (OsPlatform.IsWindows())
+            {
+                // Docker doesn't return this on windows
+                expected.Remove("edgeAgent_created_pids_total");
```
CindyXing, line 48: With it, will we miss a field or upload as 0? I wonder if there is any potential issue when scope script is calculated against.

doc/BuiltInMetrics.md
lfitchett: I think it still adds clarity, and it can't hurt

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
lfitchett: There won't be a reported metric. There are no scope scripts currently that use  this. I'm going to add this to the documentation and documentation test in my test additions

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
```diff
@@ -130,14 +130,18 @@ public CpuUsage(double? total_usage, double? usage_in_kernelmode, double? usage_
     public class MemoryStats
     {
         [JsonConstructor]
-        public MemoryStats(double? limit, double? max_usage, double? usage, MemoryStatsExtended stats)
+        public MemoryStats(double? privateworkingset, double? limit, double? max_usage, double? usage, MemoryStatsExtended stats)
```
philipktlin, line 133:
```suggestion
        public MemoryStats(double? privateWorkingSet, double? limit, double? max_usage, double? usage, MemoryStatsExtended stats)
```

philipktlin: btw what is private working set?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
```diff
@@ -179,12 +179,19 @@ void SetModuleStats(SystemResources systemResources)
                 });
                 module.MemoryStats.ForEach(ms =>
                 {
-                    ms.Limit.ForEach(limit => this.totalMemory.Set(limit, tags));
-                    ms.Usage.ForEach(usage =>
+                    if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
philipktlin, line 182: should we consider to separate in different classes for different platforms? do we need to change this code to support Yocto?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
lfitchett: this is the json constructor, the parameter name must exactly match the json it's parsing
```
        "memory_stats": {
            "commitbytes": 124702720,
            "commitpeakbytes": 128036864,
            "privateworkingset": 92790784
        },
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
lfitchett: It's likely docker would have a different json output for something like Yocto, so for now I'm just worried about tier 1

doc/BuiltInMetrics.md
and-rewsmith: I agree. We could either have release specific information in each release branch, or aggregate everything in master. I like the latter because then we don't have to manage multiple branches of docs and it is simpler for customers.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Startup.cs
```diff
@@ -47,6 +48,12 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
 
         public void Configure(IApplicationBuilder app)
         {
+            // Seperate metrics endpoint from https authentication
```
and-rewsmith, line 51: nit: typo should be separate

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
and-rewsmith: @philipktlin If you had the same question I did, I didn't know what a working set was in the context of memory. I found an answer [here](https://stackoverflow.com/questions/5405693/help-understanding-windows-memory-working-set#:~:text=Simply%20said%2C%20the%20working%20set,out%20(i.e.%20in%20RAM).&text=If%20your%20program%20uses%20more,are%20not%20necessarily%20%22gone%22.)

@lfitchett What does private imply here?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Microsoft.Azure.Devices.Edge.Hub.Service.csproj
```diff
@@ -33,6 +33,8 @@
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.3" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.3" />
     <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="3.1.3" />
+    <PackageReference Include="prometheus-net" Version="3.4.0" />
+	  <PackageReference Include="prometheus-net.AspNetCore" Version="3.4.0" />
```
varunpuranik, line 37: nit: spacing

varunpuranik: Why do we need a direct ref to prometheus here?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Startup.cs
```diff
@@ -47,6 +48,12 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
+            app.Map("/metrics", metricsApp =>
```
varunpuranik, line 52: Does this mean we still allow other paths on 9600 port (like Twin, etc?)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Hosting.cs
```diff
@@ -75,18 +75,21 @@ public class Hosting
...
                         options.Listen(
                             !Socket.OSSupportsIPv6 ? IPAddress.Any : IPAddress.IPv6Any,
-                            port,
+                            mainPort,
                             listenOptions =>
                             {
                                 listenOptions.UseHttps(connectionAdapterOptions);
                             });
+
+                        options.Listen(!Socket.OSSupportsIPv6 ? IPAddress.Any : IPAddress.IPv6Any, metricsPort);
```
and-rewsmith, line 92: How was it listening on the metrics port before?

lfitchett: It started its own server

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Microsoft.Azure.Devices.Edge.Hub.Service.csproj
lfitchett: UseMetricServer(String); is an extension function added by prometheus-net.AspNetCore, which depends on prometheus-net

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Startup.cs
lfitchett: I think it might. Is that an issue? It will still require the same authentication

and-rewsmith: If I am understanding this correctly, ASP.NET is listening on two ports and we are concerned about duplicating the existing endpoints (i.e. twin) on this new metrics port? I feel like we shouldn't duplicate even if it requires the same authentication.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Startup.cs
```diff
@@ -47,6 +48,12 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
 
         public void Configure(IApplicationBuilder app)
         {
+            // Separate metrics endpoint from https authentication
+            app.Map("/metrics", metricsApp =>
+            {
+                metricsApp.UseMetricServer(string.Empty);
```
and-rewsmith, line 54: From the stub in Prometheus.AspNetCore I can see string.empty maps to this parameter:
```
[NullableAttribute(2)] string? url = "/metrics"
```

Is your intention here to let it take the default /metrics? If so, your arg isn't null so won't this not work? Or will string.empty be treated the same as a null?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
philipktlin: do we enable/test metrics in either new or old e2e tests?

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/Metrics.cs
```diff
@@ -12,7 +12,7 @@ public static class Metrics
 
         public static IMetricsListener Listener { get; private set; } = new NullMetricsListener();
 
-        public static void Init(IMetricsProvider metricsProvider, IMetricsListener metricsListener, ILogger logger)
+        public static void InitWithServer(IMetricsProvider metricsProvider, IMetricsListener metricsListener, ILogger logger)
```
philipktlin, line 15: I didn't find any usage of this method.

philipktlin:
I found 2 MetricsListener classes implemented IMetricsListener in Edge.Util project.  However I can't find usage of iotedge\edge-util\src\Microsoft.Azure.Devices.Edge.Util\metrics\MetricsListener.cs.  Do we still need this?

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/Metrics.cs
```diff
@@ -24,5 +24,16 @@ public static void Init(IMetricsProvider metricsProvider, IMetricsListener metri
                 Listener.Start(logger);
             }
         }
+
+        public static void InitWithAspNet(IMetricsProvider metricsProvider, ILogger logger)
```
philipktlin, line 28: This looks weird to me as there are none related to AspNet inside this method. why not just name it as "Init"?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Startup.cs
philipktlin: Technically, you use MapWhen to filter out requests with path other than "/metrics" on 9600 port.  But this is same as before since request of any path can send to edgeHub.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Startup.cs
philipktlin: if you want to go for default, you can just call UseMetricServer() without arguments.

test/modules/MetricsValidator/src/Program.cs
```diff
@@ -38,6 +38,8 @@ static async Task<int> MainAsync()
                     .Build();
 
                 var transportType = configuration.GetValue("ClientTransportType", Microsoft.Azure.Devices.Client.TransportType.Mqtt);
+
+                Logger.LogInformation("Make Client");
```
philipktlin, line 42: More specific, e.g. Metrics Validator create module client instance. etc.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/models/DockerStats.cs
lfitchett: private is reserved for that docker container's use. So if that's eh's container, its how much RAM eh is using

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet.Docker/SystemResourcesMetrics.cs
lfitchett: Yep. There's tests in the new e2e framework, which I'm currently expanding

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Startup.cs
lfitchett: It needs to be empty string since `app.Map("/metrics", metricsApp =>` already routes from /metrics. If I passed in null, it would listen to `edgehub:9600/metrics/metrics`. Whatever I pass in there will be appended to `edgehub:9600/metrics`, so I "append" empty string.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/Metrics.cs
lfitchett: It's not used now since edgeAgent uses dependency injection instead of the global variables. If we want another module to use this library the same way as edgeHub, it would use this if it doesn;t have asp.net

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/Metrics.cs
lfitchett: This only works if the asp.net plugin is used

test/modules/MetricsValidator/src/Program.cs
lfitchett: This will only ever show up in MetricsValidator logs, do I really need to be that specific? It's like 2 lines down from "Starting Metrics Validator" in the logs

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Microsoft.Azure.Devices.Edge.Hub.Service.csproj
philipktlin:
```suggestion
    <PackageReference Include="prometheus-net.AspNetCore" Version="3.4.0" />
```

test/modules/MetricsValidator/src/Program.cs
philipktlin: at least put "Use module client" sth, "Client" is too general.

edge-util/src/Microsoft.Azure.Devices.Edge.Util/metrics/Metrics.cs
lfitchett: I could try that, it might let me remove the dependency from eh as well

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Hosting.cs
and-rewsmith: From the way UseHttps() is used above, I would think that https isn't configured for the metrics port. From other conversations though, you think it is? Didn't you say that https worked for direct methods on 9600 /twin?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/Startup.cs
and-rewsmith: If we can filter out non metrics requests on 9600 port, then why not do so? Before we were only listening on the main port (and technically 443). Now we are introducing another with the customer expectation as only-used-for-metrics. 

Also, I don't see how https should work considering the UseHttps() logic in Hosting.cs. Even if this works, it would be nice to eliminate the confusion by restricting the paths. I'm guessing I don't have full context here though.

PR #3013, opened 2020-05-27
Nested edge config related change in EdgeHub

philipktlin:
Get parent edge from environment variable and provide it as gateway hostname when creating module client or device client.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ClientTokenCloudConnection.cs
```diff
@@ -115,7 +117,7 @@ public async Task<ICloudProxy> UpdateTokenAsync(ITokenCredentials newTokenCreden
                                 // complete the tokenGetter
                                 if (this.tokenGetter.HasValue)
                                 {
-                                    if (TokenHelper.IsTokenExpired(this.Identity.IotHubHostName, newTokenCredentials.Token))
+                                    if (TokenHelper.IsTokenExpired(this.Identity.IotHubHostname, newTokenCredentials.Token))
```
philipktlin, line 120: We should continue to use IoT hub hostname for any token related calls, right?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/DeviceIdentity.cs
```diff
@@ -1,27 +1,59 @@
...
+            : base(iotHubHostName, gatewayHostname, deviceId)
         {
             this.DeviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
-            this.asString = new Lazy<string>(() => $"DeviceId: {this.DeviceId} [IotHubHostName: {this.IotHubHostName}]");
         }
 
         public string DeviceId { get; }
 
-        public override string Id => this.DeviceId;
+        protected bool Equals(DeviceIdentity other) =>
```
philipktlin, line 20: @varunpuranik would you tell me why we don't override Equality members in DeviceIdentity and ModuleIdentity before?  thanks.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ClientProvider.cs
```diff
@@ -18,12 +18,12 @@ public IClient Create(IIdentity identity, IAuthenticationMethod authenticationMe
 
             if (identity is IModuleIdentity)
             {
-                ModuleClient moduleClient = ModuleClient.Create(identity.IotHubHostName, authenticationMethod, transportSettings);
+                ModuleClient moduleClient = ModuleClient.Create(identity.IotHubHostname, identity.GatewayHostname.OrDefault(), authenticationMethod, transportSettings);
```
varunpuranik, line 21: Using OrDefault is not a good idea. You can map/match on the identity.GatewayHostname

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/IIdentity.cs
```diff
@@ -1,10 +1,14 @@
...
+    using Microsoft.Azure.Devices.Edge.Util;
+
     public interface IIdentity
     {
         string Id { get; }
 
-        string IotHubHostName { get; }
+        string IotHubHostname { get; }
+
+        Option<string> GatewayHostname { get; }
```
varunpuranik, line 12: I actually feel we can take this out of the Identity. I know IotHubHostname is here, and this is a similar pattern. But maybe we can take the opportunity to simplify. 
I don't think we need to move IotHubHostname out, but we can just keep GatewayHostname at a global config value.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/DeviceIdentity.cs
varunpuranik: coz we havne't needed to. We just compare Ids.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ClientTokenCloudConnection.cs
varunpuranik: For this call, this is correct.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/IIdentity.cs
philipktlin: From data modeling point of view, I think it is more appropriate to have gateway hostname inside identity instead of a global config value.  I think adding gateway hostname is still simple and straight-forward.  What is your concern?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/DeviceIdentity.cs
```diff
@@ -1,27 +1,57 @@
...
+            : base(iotHubHostName, deviceId)
         {
             this.DeviceId = Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
-            this.asString = new Lazy<string>(() => $"DeviceId: {this.DeviceId} [IotHubHostName: {this.IotHubHostName}]");
         }
 
         public string DeviceId { get; }
 
-        public override string Id => this.DeviceId;
+        protected bool Equals(DeviceIdentity other) =>
```
varunpuranik, line 19: Why do we need these changes?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/IIdentity.cs
```diff
@@ -1,10 +1,12 @@
 // Copyright (c) Microsoft. All rights reserved.
 namespace Microsoft.Azure.Devices.Edge.Hub.Core.Identity
 {
+    using Microsoft.Azure.Devices.Edge.Util;
```
varunpuranik, line 4: ?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/identity/DeviceIdentity.cs
varunpuranik: We can probably do this cleanup in another PR.

PR #3020, opened 2020-05-29
[mqtt] Move authentication logic to separate crate

dmolokanov:
* Refactored error handling for `Authenticator` and `Authorizer`
* Moved common traits and types to `mqtt-broker-core`
* Added `EdgeHubAuthenticator` into `mqtt-edgehub` crate


mqtt/mqtt-broker/tests/common.rs
```diff
@@ -445,3 +445,41 @@ where
...
+
+pub struct DummyAuthorizer(bool);
+
+impl DummyAuthorizer {
+    pub fn allow() -> Self {
+        Self(true)
+    }
+}
+
+#[async_trait]
```
vadim-kovalyov, line 478: q: why it's async trait?

dmolokanov: good catch

PR #3026, opened 2020-06-01
[mqtt] Translation prop tests

lfitchett:
This adds a prop test for the translation. It currently fails, since # and +, which are allowed in device and module ids are mqtt wildcards and break the routing. 

A separate change will fix this.


mqtt/mqtt-broker/src/proptest.rs
```diff
@@ -144,7 +144,7 @@ pub fn arb_client_id() -> impl Strategy<Value = proto::ClientId> {
     ]
 }
 pub fn arb_clientid() -> impl Strategy<Value = ClientId> {
-    "[a-zA-Z0-9]{1,23}".prop_map(Into::into)
+    "[a-zA-Z0-9_()-=*!@$%',.?+#]{1,23}".prop_map(Into::into)
```
dmolokanov, line 147: Not sure that this is correct change. From the doc
> The Server MUST allow ClientIds which are between 1 and 23 UTF-8 encoded bytes in length, and that contain only the characters
>
>"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ" [MQTT-3.1.3-5].

Such CONNECT packet with such client_id should not be parsed

lfitchett: We need to allow at least everything except + and # for now

PR #3028, opened 2020-06-01
Fix DeviceId collision in LH and Stress tests

yophilav:
Fix DeviceId collision in LH and Stress tests


scripts/linux/runE2ETest.sh
```diff
@@ -698,7 +698,10 @@ function run_longhaul_test() {
     print_highlighted_message "Run Long Haul test for $image_architecture_label"
     test_setup
 
-    local device_id="$RELEASE_LABEL-Linux-$image_architecture_label-longhaul"
+    # Import test-related functions
+    . $(dirname "$0")/testHelper.sh
```
philipktlin, line 702: testHelper.sh is not in the same folder as runE2ETest.sh in the iotedge repo.  It is in the same folder only in the artifact.  To simplify, just clone that method and put a TODO to share.

PR #3034, opened 2020-06-02
Let Debug logs be configurable for connectivity test pipeline

dylanbronson:
Changing connectivity tests to be able to configure debug logs from the pipeline variable.


builds/e2e/connectivity.yaml
```diff
@@ -316,3 +318,4 @@ jobs:
           edgeRuntime.buildNumber: '$(edgeRuntime.buildNumber)'
           customEdgeAgent.image: '$(customEdgeAgent.image)'
           customEdgeHub.image: '$(customEdgeHub.image)'
+          runtimeLogLevel: '$(runtimeLogLevel)'
```
yophilav, line 321: For the consistency, please use `test.runtimeLogLevel` instead of `runtimeLogLevel`

PR #3042, opened 2020-06-04
[k8s] Azure stack issues.

darobs:
Two major bug fixes in this PR:
1. Long IoTHub names, like those in Azure Stack were causing edgeAgent to crash.

**FIX**: remove the "hubname" label on objects created by Edge on K8s.  We can still select on device name label. Affects iotedged and edgeagent containers.

2. Unable to make TLS connection to IoT Hub.  The iotedged and edgeAgent containers were not getting the trusted CA certificates installed in such a way that they could create a TLS  connection.

**FIX**: **iotedged**: Helm chart changes to iotedged deployment, allowing the user to mount the Host OS ceretificate store.  (`.Values.iotedged.data.useHostTrust`) **edgeAgent**: use the trust bundle provided by iotedged and install it in a user trusted certificate store. create a user in edgeAgent container to have a place to put the store.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentController.cs
```diff
@@ -290,13 +288,12 @@ bool IsCreatedByController(V1PersistentVolumeClaim claim)
                 return false;
             }
 
-            if (!labels.ContainsKey(KubernetesConstants.K8sEdgeDeviceLabel) || !labels.ContainsKey(KubernetesConstants.K8sEdgeHubNameLabel))
+            if (!labels.ContainsKey(KubernetesConstants.K8sEdgeDeviceLabel))
```
CindyXing, line 291: What was the reason to have the hubnamelabel earlier?

darobs: `devicename + hubname` is unique across all IoT Hubs. This would guarantee that if we selected for these labels, we would only get objects for that unique device. You could distinguish `device1 + hub1` and `device1 + hub2` at a cluster level-query by label selection. That was the original idea.

But these labels are defined by Edge on K8s and used by Edge on K8s. We didn't make any guarantees on these labels. For our own purposes, we are using these label selections as a way to limit queries to objects Edge on K8s created.

PR #3043, opened 2020-06-04
Make sure the manifest is executed

yophilav:
Make sure the manifest is executed


scripts/linux/buildManifest.sh
```diff
@@ -159,11 +157,11 @@ cat $manifest
 echo "Done Building And Pushing Docker Images"
 
 
-if [ $WIN_ARM32_ENABLED -eq 1 ]; then
-    curl -Lo /tmp/manifest-tool 'https://edgebuild.blob.core.windows.net/iotedge-win-arm32v7-tools/manifest-tool-linux-amd64' &&
-        chmod +x /tmp/manifest-tool &&
-        /tmp/manifest-tool --debug push from-spec $IGNORE_MISSING $manifest
-fi
+# Download and execute the manifest tool
```
philipktlin, line 160: Why do we enable windows arm32 in manifest?

scripts/linux/buildManifest.sh
```diff
@@ -159,11 +157,11 @@ cat $manifest
 echo "Done Building And Pushing Docker Images"
```
and-rewsmith, line 157: Can we move this log down below the push? I was the one that caused this problem, and I think this log contributed to my mistake.

and-rewsmith:
I have a duplicate PR here, but I will close:
https://github.com/Azure/iotedge/pull/3039/files

scripts/linux/buildManifest.sh
and-rewsmith: So we don't. This tool is for all platforms. I think it is named this way because it was made to work with win arm32 also. 

I made a mistake when adding the win-arm32-enabled flag. I thought this was an winarm32 specific tool because of the naming / log message above, but should have looked more carefully.

yophilav: @philipktlin We actually use `manifest-tool-linux-amd64` executable that stores in `iotedge-win-arm32v7-tools` blob.

philipktlin: ok. please fix this path of "manifest-tool-linux-amd64" in blob storage.  I suggest to move "manifest-tool-linux-amd64" to "iotedge-linux-amd64-tools".

and-rewsmith: This might be a dangerous thing to do if other programs depend on it, right? Also, what is the motivation for this name change?

damonbarry: We'd need to understand the history of this tool--why it was named this way (or why it was placed in this path in our blob storage)--before we change it.

As of early 2018 this functionality is available in the docker CLI as `docker manifest ...` so maybe we can switch to that. Although because of windows arm32v7 maybe the CLI won't work for us.

In any case, I think this PR should be evaluated in its present scope. We can make things better in a separate PR.

scripts/linux/buildManifest.sh
```diff
@@ -156,14 +154,13 @@ sed "s/__REGISTRY__/${DOCKER_REGISTRY}/g; s/__VERSION__/${DOCKER_IMAGEVERSION}/g
 echo "Build image with following manifest:"
 cat $manifest
 
-echo "Done Building And Pushing Docker Images"
+# Download and execute the manifest tool
+curl -Lo /tmp/manifest-tool 'https://edgebuild.blob.core.windows.net/iotedge-win-arm32v7-tools/manifest-tool-linux-amd64' &&
+    chmod +x /tmp/manifest-tool &&
+    /tmp/manifest-tool --debug push from-spec $IGNORE_MISSING $manifest
 
+echo "Done Building And Pushing Docker Images"
```
damonbarry, line 162: Thanks for pushing down this log statement. I thought it was odd the log was emitted _before_ the tool was invoked. That may be part of what led to the confusion... 👍

and-rewsmith: yep

PR #3049, opened 2020-06-05
Some perf improvements for QoS 0 messages.

vadim-kovalyov:
**Context:**
When `outgoing_task` receives a QoS0 message on `ConnectionHandle`, it sends back PubAck0 into `BrokerHandle` to push message processing further. If we have a lot of fast/noisy publishers, they can quickly fill `BrokerHandle` queue and as a result `outgoing_task` will wait (be blocked) for the queue to free up and overall outgoing messages rate goes down.

**Solution:**
This PR introduces Semaphore for incoming publications (`ClientEvent::PublishFrom`) to limit the number of incoming "in-flight" messages per client. This allows us to put backpressure on fast/noisy publishers and to avoid blocking broker incoming queue (`BrokerHandle`).


mqtt/mqtt-broker/src/lib.rs
```diff
@@ -146,7 +146,9 @@ pub enum ClientEvent {
     UnsubAck(proto::UnsubAck),
 
     /// PublishFrom - publish packet from a client
-    PublishFrom(proto::Publish),
+    /// Contains optional permit for managing max number of
+    /// incoming messages per publisher.
+    PublishFrom(proto::Publish, Option<tokio::sync::OwnedSemaphorePermit>),
```
dmolokanov, line 151:
```suggestion
    PublishFrom(proto::Publish, Option<OwnedSemaphorePermit>),
```

mqtt/mqtt-broker/src/connection.rs
```diff
@@ -239,6 +242,12 @@ async fn incoming_task<S>(
 where
     S: Stream<Item = Result<Packet, DecodeError>> + Unpin,
 {
+    // We limit the number of incoming publications (PublishFrom) per client
+    // in order to avoid (a single) publisher to occupy whole BrokerHandle queue.
+    // This helps with QoS 0 messages throughput, due to the fact that outgoing_task
+    // also uses sends PubAck0 for QoS 0 messages to BrokerHandle queue.
+    let inflight_guard = Arc::new(Semaphore::new(10));
```
dmolokanov, line 249: it's now inflight, right? for me, inflight is when the broker sends messages to the client.
```suggestion
    let incoming_publications_limit = Arc::new(Semaphore::new(10));
```
?

dmolokanov, line 249: did you had a chance to play with this number? should we make it configurable via config file or ENV?

varunpuranik, line 249: Changes look good. But is the 10 number heuristics based? What is the thinking around the right number here?

vadim-kovalyov: As discussed offline, this number was optimal in local testing.

vadim-kovalyov: As discussed offline, we don't need to configure it for now.

PR #3050, opened 2020-06-05
[mqtt] Add additional transport for local MQTT communication

dmolokanov:
Add additional transport for local MQTT communication on `localhost:1882`.
Run additional transport only when `edgehub` feature activated.
Add authenticator that allows any connection on localhost.


mqtt/mqtt-broker-core/src/auth/authentication.rs
```diff
@@ -43,34 +34,89 @@ pub trait Authenticator {
...
+        Self {
+            client_id,
+            peer_addr,
+            username: None,
+            password: None,
+            certificate: None,
+        }
+    }
+
+    pub fn with_username(&mut self, username: impl Into<String>) -> &mut Self {
```
vadim-kovalyov, line 62: q: nit: so for builder pattern should we use "with_xxxx" everywhere?

mqtt/mqtt-broker-core/src/auth/authentication.rs
```diff
@@ -82,44 +128,50 @@ pub struct DefaultAuthenticator;
...
 
 #[cfg(test)]
 mod tests {
+    use std::net::SocketAddr;
+
     use matches::assert_matches;
 
-    use crate::auth::{
-        authenticate_fn_ok, AuthId, Authenticator, Credentials, DefaultAuthenticator,
+    use super::{
```
vadim-kovalyov, line 142: I thought we were not in favor of `super` :)

mqtt/mqtt-edgehub/src/auth/edgehub.rs
```diff
@@ -57,35 +61,31 @@ impl Authenticator for EdgeHubAuthenticator {
...
 #[derive(Serialize)]
 #[serde(rename_all = "snake_case")]
-pub struct EdgeHubAuthRequest {
-    version: String,
-    username: Option<String>,
-    password: Option<String>,
+pub struct EdgeHubAuthRequest<'a> {
+    version: &'a str,
+    username: Option<&'a str>,
+    password: Option<&'a str>,
```
vadim-kovalyov, line 76: q: you didn't want to clone values?

mqtt/mqtt-edgehub/src/auth/local.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct LocalAuthenticator;
+
+impl LocalAuthenticator {
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+#[async_trait]
+impl Authenticator for LocalAuthenticator {
```
vadim-kovalyov, line 17: can you please add some comments on why we need this LocalAuthenticator?

mqtt/mqtt-broker-core/src/auth/authentication.rs
dmolokanov: this specific case is different: it an object that optional fields set in "builder-like" style. usually the builder doesn't expose getters.

But, seems like a good idea to have similar semantics across all the code

mqtt/mqtt-edgehub/src/auth/edgehub.rs
dmolokanov: correct

mqtt/mqtt-edgehub/src/auth/local.rs
dmolokanov: There is a comment when we actually use this authenticator.

mqtt/mqtt-broker-core/src/auth/authentication.rs
dmolokanov: Yeah, I am moving code across modules for the last few weeks, and changed my opinion:
1. when in unit tests we use **parent** module items: struct, enums, functions - use `super`
2. whatever outside of parent module, use `crate::module::Struct`

PR #3051, opened 2020-06-05
Hard-coded the rust toolchain to version 1.43.1 for master branch

yophilav:
Hard-coded the rust toolchain to version 1.43.1 for master branch
- MQTT
- Edgelet


arsing:
Why? Then it's just going to rot because no one is going to update it.

PR #3054, opened 2020-06-08
[mqtt] Print detailed error messages in the log when failed to start

dmolokanov:
Print whole stack of inner error messages in the log when broker failed to start working.

### DIFF
--- a/mqtt/mqtt-broker/src/auth.rs
+++ b/mqtt/mqtt-broker/src/auth.rs
@@ -27,6 +27,6 @@ mod imp {
     use mqtt_broker_core::auth::{authenticate_fn_ok, AuthId, Authenticator};
 
     pub(super) fn authenticator() -> impl Authenticator<Error = Box<dyn StdError>> {
-        authenticate_fn_ok(|_, _| Some(AuthId::Anonymous))
+        authenticate_fn_ok(|_| Some(AuthId::Anonymous))
     }
 }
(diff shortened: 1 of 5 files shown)


mqtt/mqtt-broker/src/auth.rs
```diff
@@ -27,6 +27,6 @@ mod imp {
     use mqtt_broker_core::auth::{authenticate_fn_ok, AuthId, Authenticator};
 
     pub(super) fn authenticator() -> impl Authenticator<Error = Box<dyn StdError>> {
-        authenticate_fn_ok(|_, _| Some(AuthId::Anonymous))
+        authenticate_fn_ok(|_| Some(AuthId::Anonymous))
```
vadim-kovalyov, line 30: :+1:

mqtt/mqtt-broker/src/error.rs
```diff
@@ -86,3 +87,17 @@ pub enum InitializeBrokerError {
     #[error("An error occurred  bootstrapping TLS")]
     Tls(#[source] native_tls::Error),
 }
+
+pub struct DetailedErrorValue<'a, E>(pub &'a E);
+
+impl<'a, E: StdError> std::fmt::Display for DetailedErrorValue<'a, E> {
```
vadim-kovalyov, line 93: nit: use std::fmt::{...}?

PR #3057, opened 2020-06-08
Enable nested OnBehalfOf connection

richma-ms:
This change consumes a new Devices.Client SDK package that I push to MyGet, and sets the authchain for the OnBehalfOf device on Amqp/MQTT transport settings and uses the updated IoTHub API version.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Microsoft.Azure.Devices.Edge.Agent.Diagnostics.csproj
```diff
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.23.1" />
+    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.23.1-NestedEdge" />
```
dylanbronson, line 12: Your changes in the SDK will go in their next release, right? aka we won't have to maintain a separate sdk version for too long?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
```diff
@@ -101,6 +101,14 @@ async Task<(bool isAuthenticated, bool shouldFallback)> AuthenticateInternalAsyn
 
         async Task<(bool isAuthenticated, bool serviceIdentityFound)> AuthenticateWithServiceIdentity(T credentials, string serviceIdentityId, bool syncServiceIdentity)
         {
+            Option<string> authChain = await this.deviceScopeIdentitiesCache.GetAuthChain(serviceIdentityId);
+
+            if (!authChain.HasValue)
+            {
+                // The identity is no within our nested hierarchy
```
dylanbronson, line 108: nit: typo - not

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Microsoft.Azure.Devices.Edge.Agent.Diagnostics.csproj
richma-ms: I don't think the timing is locked down yet, since I've updated the REST API version in the SDK, they can't release this change until the IoT Hub changes from Anca and David go public.

varunpuranik: We should rebase the private SDK bits with the changes from 1.26.0 version of the SDK

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -95,11 +95,16 @@ public async Task<Option<string>> GetAuthChain(string id)
         {
             Preconditions.CheckNonWhiteSpace(id, nameof(id));
 
-            // Auth-chain for the root (when there aren't any nodes in the tree yet)
+            // Auth-chain for the acting device (when there aren't any nodes in the tree yet)
             if (id == this.actorDeviceId)
             {
                 return Option.Some(this.actorDeviceId);
             }
+            else if (id == this.actorDeviceId + "/$edgeHub")
```
varunpuranik, line 103: Why do you need to treat EdgeHub on the device differently? (Why can't it be handled by the loop below?) Is it because we need this for EdgeHub even if it hasn't been populated yet?

Nit: We should use a const for $edgeHub

richma-ms: yes, it's because we make the initial upstream call to get the scope as $edgeHub, and the tree isn't populated yet at that point.

PR #3059, opened 2020-06-09
Update long haul log size for debug

yophilav:
Update long haul log size for debug
If the RunTimeLogLevel is set to `debug` the log rotation is set to 96MB for edgeHub and edgeAgent. Otherwise the log rotaiton is set to 4MB. The reason behind this is to facilitate that the debugging log can be fully captured for 7 days without being rotated out for a newer log on the two modules. 

Our most memory constrains test agent is ARM32 with 8GB of storage. With ARM32 with 8GB storage ( ~5GB actually available for logging ), we have 96M x 7 days x 2 modules (edgeHub & edgeAgent) = ~1.4GB.  And the rest of the modules takes up = 4M x 7 days x 16 modules = 448MB. Total = ~2GB. We should ideally have about 3GB left on ARM32 device storage.

### DIFF
--- a/scripts/linux/runE2ETest.sh
+++ b/scripts/linux/runE2ETest.sh
@@ -159,6 +159,7 @@ function prepare_test_from_artifacts() {
                     cp "$long_haul_deployment_artifact_file" "$deployment_working_file"
                     sed -i -e "s@<DesiredModulesToRestartCSV>@$DESIRED_MODULES_TO_RESTART_CSV@g" "$deployment_working_file"
                     sed -i -e "s@<RestartIntervalInMins>@$RESTART_INTERVAL_IN_MINS@g" "$deployment_working_file"
+                    sed -i -e "s@<MaxLogSize>@$max_log_size@g" "$deployment_working_file"
                 else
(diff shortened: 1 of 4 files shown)


scripts/linux/runE2ETest.sh
```diff
@@ -1118,6 +1119,12 @@ if [[ "${TEST_NAME,,}" == "longhaul" ]]; then
     TWIN_UPDATE_SIZE="${TWIN_UPDATE_SIZE:-1}"
     TWIN_UPDATE_FREQUENCY="${TWIN_UPDATE_FREQUENCY:-00:00:15}"
     TEST_START_DELAY="${TEST_START_DELAY:-00:00:00}"
+
+    # if Run log level is debug, we need ~96M for 24hr of edgeHub/edgeAgent log
+    if [[ "${RUNTIME_LOG_LEVEL,,}" == "debug" ]]; then
```
philipktlin, line 1124: I think you want to use different log file size for different architecture, rather than log level.  I don't agree that we can only use 4m x 7 on Linux ARM32 device.  And why 96M for AMD64? did you check the average LH log file size per day? and why we don't turn on debug log level for Linux arm32?

PR #3063, opened 2020-06-09
Added session config and message limits

vadim-kovalyov:
**Context:**
In order to help with memory consumption for offline or slow subscribers, we need to introduce limits on message queue size and inflight message queue size.

The main complexity here is that queues live in `SessionState`, so we better put `SessionConfig` there as well. But `SessionState` also participates in persistence, but we don't want to persist `SessionConfig`, since it can change b/w broker restarts.

**Solutions:**
There were two ways to work this out:
1. Add an extra layer (data structure) `SessionSnapshot` that would not have `SessionConfig` and be used for persistence. In `BrokerBuilder`, on broker load, convert `SessionSnapshot` into "runtime" `SessionState` by adding `SessionConfig` to it. This is more complex approach.
2. Load `SessionState` from persisted state with default config values, but then ensure to set actual `SessionConfig` in the `BrokerBuilder`. This is less complex but less "rusty" approach.

In this PR I'm attempting approach number 1.


mqtt/mqtt-broker/src/broker.rs
```diff
@@ -991,15 +998,26 @@ where
         self
     }
 
+    pub fn with_config(mut self, config: BrokerConfig) -> Self {
+        self.config = Some(config);
+        self
+    }
+
     pub fn build(self) -> Broker<Z> {
+        let config = match self.config {
```
dmolokanov, line 1007: lets put BrokerConfig::default() in the constructor instead?

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -651,7 +652,9 @@ where
                         }
                     } else {
                         info!("cleaning offline session for {}", client_id);
-                        let new_session = Session::new_transient(auth_id, connreq);
+                        let state =
+                            SessionState::new(client_id.clone(), self.config.session().clone());
```
dmolokanov, line 656: since we will not make separate settings per session, it may worthwhile to share config between sessions via Rc instead of cloning

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -23,64 +23,76 @@ pub enum Transport {
...
     expiration: Duration,
 }
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Clone, PartialEq)]
 pub struct SessionMessages {
     #[serde(deserialize_with = "humansize")]
     max_message_size: u64,
-    max_count: u32,
+    max_inflight_messages: u32,
```
dmolokanov, line 44:
```suggestion
    max_inflight_messages: usize,
```

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -23,64 +23,76 @@ pub enum Transport {
...
+    max_queued_messages: u32,
```
dmolokanov, line 45:
```suggestion
    max_queued_messages: usize,
```

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -152,6 +157,19 @@ impl BrokerConfig {
...
 
+impl Default for BrokerConfig {
+    fn default() -> Self {
+        let mut s = Config::new();
+
+        // Returning Self instead of Result simplifies the code significantly.
+        // It is guaranteed that next two calls must not fail,
+        // otherwise we have a bug in the code or in ../config/default.json file.
+        // It is guarded by a unit test as well.
+        s.merge(File::from_str(DEFAULTS, FileFormat::Json)).unwrap();
```
dmolokanov, line 168: lets make expect with meaningful message instead unwrap?

mqtt/mqtt-broker/src/session.rs
```diff
@@ -350,7 +358,19 @@ impl SessionState {
...
                 Ok(Some(event))
             } else {
-                self.waiting_to_be_sent.push_back(publication);
+                if self.waiting_to_be_sent.len() > self.config.max_queued_messages() {
+                    match self.config.when_full() {
+                        QueueFullAction::DropNew => {
+                            return Ok(None);
+                        }
+                        QueueFullAction::DropOld => {
+                            drop(self.waiting_to_be_sent.pop_front());
```
dmolokanov, line 367: I think in general we should avoid calling `drop` manually
```suggestion
                            let _ = self.waiting_to_be_sent.pop_front();
```

mqtt/mqtt-broker/src/session.rs
```diff
@@ -320,6 +323,11 @@ impl SessionState {
         &self.subscriptions
     }
 
+    pub fn set_config(mut self, config: SessionConfig) -> SessionState {
```
dmolokanov, line 326:
```suggestion
    pub fn set_config(&mut self, config: SessionConfig) {
```
?

mqtt/mqtt-broker/src/session.rs
vadim-kovalyov: Sure, I thought it would make it more clear that we want to drop the message.

mqtt/mqtt-broker/src/broker.rs
vadim-kovalyov: I think cloning session config is fine. For Rc<SessionConfig>, I would have to keep a filed in the broker and it does not give any big advantage.

mqtt/mqtt-broker/src/broker.rs
vadim-kovalyov: I think that this is not a big deal and would prefer to keep `default()` as a constructor for BrokerBuilder.

vadim-kovalyov: I guess I didn't get what you meant the first time. Fixed it now.

mqtt/mqtt-broker/src/session.rs
```diff
@@ -458,10 +487,32 @@ impl SessionState {
...
+        }
+
         let num_inflight = self.waiting_to_be_acked.len()
             + self.waiting_to_be_acked_qos0.len()
             + self.waiting_to_be_completed.len();
-        num_inflight < MAX_INFLIGHT_MESSAGES
+        num_inflight < self.config.max_inflight_messages()
+    }
+
+    fn add_to_queue(&mut self, publication: proto::Publication) {
```
dmolokanov, line 500:
```suggestion
    fn enqueue(&mut self, publication: proto::Publication) {
```
 ?

mqtt/mqtt-broker/src/session.rs
```diff
@@ -458,10 +487,32 @@ impl SessionState {
...
+                QueueFullAction::DropNew => {
+                    // do nothing
+                }
+                QueueFullAction::DropOld => {
+                    let _ = self.waiting_to_be_sent.pop_front();
+                    self.waiting_to_be_sent.push_back(publication);
+                }
+            }
+        } else {
+            self.waiting_to_be_sent.push_back(publication);
```
dmolokanov, line 514: When `self.config.max_queued_messages()` == 0 it will always push messages into the queue. Is that what we want?

dmolokanov: I assume mosquitto works this way, but I'm not sure we need to do the same. If user wants to opt-out from `max_queued_messages` we have Option for this case. `Option<NonZeroUsize>` ?
https://doc.rust-lang.org/std/num/struct.NonZeroUsize.html

mqtt/mqtt-broker/src/session.rs
```diff
@@ -598,8 +632,7 @@ pub enum Session {
 }
 
 impl Session {
-    pub fn new_transient(auth_id: AuthId, connreq: ConnReq) -> Self {
-        let state = SessionState::new(connreq.client_id().clone());
+    pub fn new_transient(auth_id: AuthId, connreq: ConnReq, state: SessionState) -> Self {
```
dmolokanov, line 635: It looks weird that for the transient session you need to create a session state first. It seems like should be encapsulated in the `new_transient` method

dmolokanov: Or if you want to move it out and make it similar to other methods, you don't need to pass ConnReq inside, pass its parts instead

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -23,64 +23,89 @@ pub enum Transport {
...
+    #[serde(deserialize_with = "humansize")]
+    max_message_size: u64,
+    max_inflight_messages: usize,
+    max_queued_messages: usize,
+    #[serde(deserialize_with = "humansize")]
+    max_queued_size: u64,
+    when_full: QueueFullAction,
+}
+
+impl SessionConfig {
```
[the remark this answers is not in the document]
vadim-kovalyov: Yes, I was thinking about that as well. I'll try to figure something out in the next PR.

mqtt/mqtt-broker/src/session.rs
```diff
@@ -530,39 +581,10 @@ impl SessionState {
...
-            waiting_to_be_acked_qos0: HashMap::new(),
-            waiting_to_be_released: HashMap::new(),
-            waiting_to_be_completed: HashSet::new(),
-        }
-    }
 }
 
 #[cfg(any(test, feature = "proptest"))]
-#[allow(clippy::too_many_arguments)]
+#[allow(clippy::too_many_arguments, dead_code)]
```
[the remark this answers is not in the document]
vadim-kovalyov: good catch

mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -23,64 +23,89 @@ pub enum Transport {
...
+
+    pub fn max_inflight_messages(&self) -> usize {
+        self.max_inflight_messages
+    }
+
+    pub fn max_queued_messages(&self) -> usize {
+        self.max_queued_messages
+    }
+
+    pub fn when_full(&self) -> &QueueFullAction {
```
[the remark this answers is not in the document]
vadim-kovalyov: Yeah, I've actually done that for queue size limits. Will publish in the next PR.

mqtt/mqtt-broker/src/session.rs
vadim-kovalyov: I wanted to move it out since `SessionState` now requires config. Since `Broker` owns the config, I wanted to consolidate `SessionState` construction in the broker, instead of passing config into `new_transient`.

mqtt/mqtt-broker/config/default.json
```diff
@@ -1,25 +1,21 @@
...
     },
     "session": {
         "expiration": "60d",
-        "messages": {
-            "max_message_size": "256kb",
-            "max_count": 1000,
-            "max_total_space": "16kb",
-            "when_full": "drop_new"
-        }                
+        "max_message_size": "256kb",
```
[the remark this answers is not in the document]
vipeller: Except that I'd prefer an embedded structure vs group1_settingA / group2_settingA style (e.g. message_expiration), I assume it can work. I think embedded structure is also better for extension. But I am not against the flat structure.

mqtt/mqtt-broker/src/session.rs
vadim-kovalyov: Yes, in our case 0 in the config.json would mean "unlimited" queue. But I also agree with using `NonZeroUsize` internally. As discussed offline, let's make that change in the next PR to keep this one shorter.

PR #3074, opened 2020-06-11
Handle ServiceClient timeout on DirectMethodSender

yophilav:
In some rare occasion, the IoTHub would take too long to response causing ServiceClient to operation time out which translates to InternalServerError500 by Direct Method Sender. Frankly, we don't care how long IoTHub takes to response. Direct Method Sender should simply retry to send the direct method again by triggering the ServiceClient with the same directMethodSequence number.

### DIFF
--- a/test/modules/DirectMethodSender/DirectMethodSenderBase.cs
+++ b/test/modules/DirectMethodSender/DirectMethodSenderBase.cs
@@ -37,7 +37,30 @@ namespace DirectMethodSender
             logger.LogInformation($"{this.GetType().ToString()} : Calling Direct Method on device {this.deviceId} targeting module [{this.targetModuleId}] with count {this.directMethodCount}.");
             try
             {
-                int resultStatus = await this.InvokeDeviceMethodAsync(this.deviceId, this.targetModuleId, methodName, this.directMethodCount, CancellationToken.None);
+                const int maxRetry = 3;
+                int resultStatus = (int)HttpStatusCode.InternalServerError;
+                int transientRetryCount = 0;
+
+                while (transientRetryCount <= maxRetry)
+                {
+                    try
+                    {
+                        transientRetryCount++;
+                        resultStatus = await this.InvokeDeviceMethodAsync(this.deviceId, this.targetModuleId, methodName, this.directMethodCount, CancellationToken.None);
+                    }
+                    catch (IotHubCommunicationException e) when (e.IsTransient)
+                    {
+                        if (transientRetryCount < maxRetry)
+                        {
+                            logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {this.directMethodCount}.");
+                        }
+                        else
+                        {
+                            logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {this.directMethodCount}. Retry Exceeded.");
+                            return new Tuple<HttpStatusCode, ulong>(HttpStatusCode.RequestTimeout, this.directMethodCount);
+                        }
+                    }
+                }
 
                 string statusMessage = $"Calling Direct Method with count {this.directMethodCount} returned with status code {resultStatus}";
                 if (resultStatus == (int)HttpStatusCode.OK)
@@ -56,12 +79,6 @@ namespace DirectMethodSender
             {
                 logger.LogInformation(e, $"Transient exception caught with count {this.directMethodCount}");
                 return new Tuple<HttpStatusCode, ulong>(HttpStatusCode.NotFound, this.directMethodCount);
-            }
-            catch (IotHubCommunicationException e) when (e.Message.Contains("operation timed out"))
-            {
-                logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {this.directMethodCount}. Retry.");
-                this.directMethodCount--;
-                return new Tuple<HttpStatusCode, ulong>(HttpStatusCode.RequestTimeout, this.directMethodCount + 1);
             }
             catch (Exception e)
             {
--- a/test/modules/DirectMethodSender/DirectMethodSenderBase.cs
+++ b/test/modules/DirectMethodSender/DirectMethodSenderBase.cs
@@ -57,6 +57,12 @@ namespace DirectMethodSender
                 logger.LogInformation(e, $"Transient exception caught with count {this.directMethodCount}");
                 return new Tuple<HttpStatusCode, ulong>(HttpStatusCode.NotFound, this.directMethodCount);
             }
+            catch (IotHubCommunicationException e) when (e.Message.Contains("operation timed out"))
+            {
+                logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {this.directMethodCount}. Retry.");
+                this.directMethodCount--;
+                return new Tuple<HttpStatusCode, ulong>(HttpStatusCode.RequestTimeout, this.directMethodCount + 1);
+            }
             catch (Exception e)
             {
                 logger.LogError(e, $"Exception caught with count {this.directMethodCount}");
--- a/test/modules/DirectMethodSender/DirectMethodSenderBase.cs
+++ b/test/modules/DirectMethodSender/DirectMethodSenderBase.cs
@@ -79,12 +79,14 @@ namespace DirectMethodSender
             {
                 try
                 {
+                    logger.LogDebug($"InvokeDirectMethodWithRetryAsync with count {directMethodCount}; Retry {transientRetryCount}");
                     transientRetryCount++;
                     resultStatus = await this.InvokeDeviceMethodAsync(deviceId, targetModuleId, methodName, directMethodCount, CancellationToken.None);
+                    return resultStatus;
                 }
                 catch (IotHubCommunicationException e) when (e.IsTransient)
                 {
-                    logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {this.directMethodCount}. Retry: {transientRetryCount}");
+                    logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {directMethodCount}. Retry: {transientRetryCount}");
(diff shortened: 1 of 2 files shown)


test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -57,6 +57,12 @@ abstract class DirectMethodSenderBase : IDisposable
                 logger.LogInformation(e, $"Transient exception caught with count {this.directMethodCount}");
                 return new Tuple<HttpStatusCode, ulong>(HttpStatusCode.NotFound, this.directMethodCount);
             }
+            catch (IotHubCommunicationException e) when (e.Message.Contains("operation timed out"))
```
philipktlin, line 60: Don't use e.Message.

Microsoft.Azure.Devices.Common.Exceptions.IotHubCommunicationException: The POST operation timed out.
 ---> System.Threading.Tasks.TaskCanceledException: The operation was canceled.
 ---> System.IO.IOException: Unable to read data from the transport connection: Operation canceled.
 ---> System.Net.Sockets.SocketException (125): Operation canceled

Is it possible to check the inner exception is SocketException?

philipktlin: I would suggest to use a while loop for line 40 and has another try catch block and catch only for specific timeout case in order to retry instead of minus count.

test/modules/DirectMethodSender/Program.cs
```diff
@@ -140,7 +140,8 @@ public static TestResultBase ConstructTestResult(DirectMethodResultType directMe
 
         static bool ShouldReportResults(DirectMethodResultType resultType, HttpStatusCode statusCode)
         {
-            return !(resultType == DirectMethodResultType.LegacyDirectMethodTestResult && statusCode == HttpStatusCode.NotFound);
+            return !(resultType == DirectMethodResultType.LegacyDirectMethodTestResult && statusCode == HttpStatusCode.NotFound) &&
```
dylanbronson, line 143: Can you add a brief comment saying what we're not reporting and why

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -37,7 +37,30 @@ abstract class DirectMethodSenderBase : IDisposable
             logger.LogInformation($"{this.GetType().ToString()} : Calling Direct Method on device {this.deviceId} targeting module [{this.targetModuleId}] with count {this.directMethodCount}.");
             try
             {
-                int resultStatus = await this.InvokeDeviceMethodAsync(this.deviceId, this.targetModuleId, methodName, this.directMethodCount, CancellationToken.None);
+                const int maxRetry = 3;
+                int resultStatus = (int)HttpStatusCode.InternalServerError;
+                int transientRetryCount = 0;
+
+                while (transientRetryCount <= maxRetry)
```
philipktlin, line 44: nit: since it is named as maxRetry.
```suggestion
                while (transientRetryCount < maxRetry)
```

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -37,7 +37,30 @@ abstract class DirectMethodSenderBase : IDisposable
...
+                {
+                    try
+                    {
+                        transientRetryCount++;
+                        resultStatus = await this.InvokeDeviceMethodAsync(this.deviceId, this.targetModuleId, methodName, this.directMethodCount, CancellationToken.None);
+                    }
+                    catch (IotHubCommunicationException e) when (e.IsTransient)
+                    {
+                        if (transientRetryCount < maxRetry)
```
philipktlin, line 53: nit: suggest to make it simple, just log retry count and direct method count.

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -37,7 +37,30 @@ abstract class DirectMethodSenderBase : IDisposable
...
+                        {
+                            logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {this.directMethodCount}.");
```
philipktlin, line 55:
```suggestion
                            logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {this.directMethodCount} and retry count {transientRetryCount}.");
```

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -37,7 +37,30 @@ abstract class DirectMethodSenderBase : IDisposable
             logger.LogInformation($"{this.GetType().ToString()} : Calling Direct Method on device {this.deviceId} targeting module [{this.targetModuleId}] with count {this.directMethodCount}.");
             try
             {
-                int resultStatus = await this.InvokeDeviceMethodAsync(this.deviceId, this.targetModuleId, methodName, this.directMethodCount, CancellationToken.None);
+                const int maxRetry = 3;
```
philipktlin, line 40: try to wrap this block in another method.

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -37,7 +37,30 @@ abstract class DirectMethodSenderBase : IDisposable
...
+                    catch (IotHubCommunicationException e) when (e.IsTransient)
+                    {
+                        if (transientRetryCount < maxRetry)
+                        {
+                            logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {this.directMethodCount}.");
+                        }
+                        else
+                        {
+                            logger.LogInformation(e, $"Transient IotHubCommunicationException caught with count {this.directMethodCount}. Retry Exceeded.");
+                            return new Tuple<HttpStatusCode, ulong>(HttpStatusCode.RequestTimeout, this.directMethodCount);
```
philipktlin, line 60: Remove this line and just set resultStatus  = HttpStatusCode.RequestTimeout inside catch block.

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -87,6 +64,41 @@ abstract class DirectMethodSenderBase : IDisposable
...
+            while (transientRetryCount <= maxRetry)
+            {
+                try
+                {
+                    transientRetryCount++;
+                    resultStatus = await this.InvokeDeviceMethodAsync(deviceId, targetModuleId, methodName, directMethodCount, CancellationToken.None);
+                }
+                catch (IotHubCommunicationException e) when (e.IsTransient)
+                {
+                    if (transientRetryCount < maxRetry)
```
philipktlin, line 87: Remove this if block and just keep line 89 and 94; and add transientRetryCount into log message in line 89.

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -87,6 +64,41 @@ abstract class DirectMethodSenderBase : IDisposable
...
+            string deviceId,
+            string targetModuleId,
+            string methodName,
+            ulong directMethodCount)
+        {
+            const int maxRetry = 3;
+            int resultStatus = (int)HttpStatusCode.InternalServerError;
+            int transientRetryCount = 0;
+
+            while (transientRetryCount <= maxRetry)
```
philipktlin, line 78: If this if-condition has equal sign, it will run maxRetry+1 times.  Therefore remove equal sign from if-condition.

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -64,6 +64,41 @@ abstract class DirectMethodSenderBase : IDisposable
...
 
+        async Task<int> InvokeDirectMethodWithRetryAsync(
+            ILogger logger,
+            string deviceId,
+            string targetModuleId,
+            string methodName,
+            ulong directMethodCount)
+        {
+            const int maxRetry = 3;
+            int resultStatus = (int)HttpStatusCode.InternalServerError;
```
philipktlin, line 75: Should we initialize to zero instead?

yophilav: I prefer to have a test failed by default and if the send success, the value should be change to other value.

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
yophilav: OOOH I see what you mean. nice!

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
philipktlin: This value is never used, why not use zero?

test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -79,12 +79,14 @@ abstract class DirectMethodSenderBase : IDisposable
             {
                 try
                 {
+                    logger.LogDebug($"InvokeDirectMethodWithRetryAsync with count {directMethodCount}; Retry {transientRetryCount}");
                     transientRetryCount++;
                     resultStatus = await this.InvokeDeviceMethodAsync(deviceId, targetModuleId, methodName, directMethodCount, CancellationToken.None);
+                    return resultStatus;
```
philipktlin, line 85: shouldn't it return on line 94?

yophilav: It would but that's not the expected behavior. 
if the DM can successfully be send, it should only send 1 dm, not 3 dms.

philipktlin: I prefer not using return inside a while loop, would you use break statement or have a proper exit condition for while clause? or you still prefer use return here?

yophilav: break; sounds good. Fixed!

PR #3075, opened 2020-06-11
Phlin/edge agent get twin fix

philipktlin:
1. When Edge agent calling GetTwinAsync method throws exception after retry policy, if the exception is not timeout exception it will trigger to close module client and retry again.
2. Remove cloud connection in connection manager when remove device connection for deployment scenario that a module is removed and added it back in a short period of time.  This will prevent cloud connection is in invalid state.
3. Upgrade Microsoft.Azure.Devices.Client NuGet package from 1.23.1 to 1.23.2, which contains AMQP Safely close fix.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
@@ -184,14 +184,16 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub
                 });
         }
 
-        async Task<Option<Twin>> GetTwinFromIoTHub(bool forceNewModuleClient = false)
+        async Task<Option<Twin>> GetTwinFromIoTHub(bool retry = false)
         {
+            IModuleClient moduleClient = null;
+
             try
             {
                 async Task<Twin> GetTwinFunc()
                 {
-                    Events.GettingModuleClient(forceNewModuleClient);
-                    IModuleClient moduleClient = await this.moduleConnection.GetOrCreateModuleClient(forceNewModuleClient);
+                    Events.GettingModuleClient(retry);
+                    moduleClient = await this.moduleConnection.GetOrCreateModuleClient();
                     Events.GotModuleClient();
                     return await moduleClient.GetTwinAsync();
                 }
@@ -211,8 +213,9 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub
             {
                 Events.ErrorGettingTwin(e);
 
-                if (!forceNewModuleClient && this.RetryWithNewModuleClient(e))
+                if (!retry && !(e is TimeoutException))
                 {
+                    await moduleClient?.CloseAsync();
                     return await this.GetTwinFromIoTHub(true);
                 }
 
@@ -365,9 +368,9 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub
                 Log.LogDebug((int)EventIds.ErrorUpdatingReportedProperties, ex, "Error updating reported properties in IoT Hub");
             }
 
-            public static void GettingModuleClient(bool forceRenew)
+            public static void GettingModuleClient(bool retry)
             {
-                Log.LogDebug((int)EventIds.GettingModuleClient, $"Getting module client to refresh the twin with forceRenew set to {forceRenew}");
+                Log.LogDebug((int)EventIds.GettingModuleClient, $"Getting module client to refresh the twin with retry set to {retry}");
             }
 
             public static void GotModuleClient()
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
@@ -3,6 +3,8 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub
 {
     using System;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Amqp;
+    using Microsoft.Azure.Amqp.Encoding;
     using Microsoft.Azure.Devices.Client;
     using Microsoft.Azure.Devices.Edge.Agent.Core;
     using Microsoft.Azure.Devices.Edge.Agent.Core.ConfigSources;
@@ -24,7 +26,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub
         static readonly ITransientErrorDetectionStrategy AllButFatalErrorDetectionStrategy = new DelegateErrorDetectionStrategy(ex => ex.IsFatal() == false);
 
         static readonly RetryStrategy TransientRetryStrategy =
-            new ExponentialBackoff(5, TimeSpan.FromSeconds(2), TimeSpan.FromSeconds(60), TimeSpan.FromSeconds(4));
+            new ExponentialBackoff(3, TimeSpan.FromSeconds(2), TimeSpan.FromSeconds(60), TimeSpan.FromSeconds(4));
 
         readonly AsyncLock twinLock = new AsyncLock();
         readonly ISerde<DeploymentConfig> desiredPropertiesSerDe;
@@ -184,14 +186,14 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub
                 });
         }
 
-        async Task<Option<Twin>> GetTwinFromIoTHub()
+        async Task<Option<Twin>> GetTwinFromIoTHub(bool forceNewModuleClient = false)
         {
             try
             {
                 async Task<Twin> GetTwinFunc()
                 {
-                    Events.GettingModuleClient();
-                    IModuleClient moduleClient = await this.moduleConnection.GetOrCreateModuleClient();
+                    Events.GettingModuleClient(forceNewModuleClient);
+                    IModuleClient moduleClient = await this.moduleConnection.GetOrCreateModuleClient(forceNewModuleClient);
                     Events.GotModuleClient();
                     return await moduleClient.GetTwinAsync();
                 }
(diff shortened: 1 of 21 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -210,10 +210,21 @@ async Task<Twin> GetTwinFunc()
             catch (Exception e)
             {
                 Events.ErrorGettingTwin(e);
+
+                if (!forceNewModuleClient && this.RetryWithNewModuleClient(e))
+                {
+                    return await this.GetTwinFromIoTHub(true);
```
varunpuranik, line 216: If this throws, will the exception bubble up? I beleive we don't want that.

philipktlin: since this method already caught everything, it will not have any exception thrown from this method.  When exception happens, it will catch and return Option.None<Twin>().

varunpuranik: ah yes

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -210,10 +210,21 @@ async Task<Twin> GetTwinFunc()
...
+                }
+
                 return Option.None<Twin>();
             }
         }
 
+        bool RetryWithNewModuleClient(Exception e)
```
varunpuranik, line 223: Nit: remove this method, not very useful

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -184,14 +184,14 @@ async Task RefreshTwinAsync()
...
+        async Task<Option<Twin>> GetTwinFromIoTHub(bool forceNewModuleClient = false)
         {
             try
             {
                 async Task<Twin> GetTwinFunc()
                 {
-                    Events.GettingModuleClient();
-                    IModuleClient moduleClient = await this.moduleConnection.GetOrCreateModuleClient();
+                    Events.GettingModuleClient(forceNewModuleClient);
+                    IModuleClient moduleClient = await this.moduleConnection.GetOrCreateModuleClient(forceNewModuleClient);
```
varunpuranik, line 194: This will cause the module client to be created every time in the retry operation (up to 3 times). 
We should do it only once. Maybe it may be easier to close the moduleClient in the code here and let GetOrCreateModuleClient create a new one.

philipktlin: If we come to this point, it means something happens to the module client other than connection timeout.  If that is the case, I expect it should be succeeded in 1st call and no retry should be needed.  If retry really happens, it is no harmful to re-create a module client.  I think this is more simple, isn't it?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
philipktlin: Sure. I will do inline.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -210,6 +212,21 @@ async Task<Twin> GetTwinFunc()
             catch (Exception e)
             {
                 Events.ErrorGettingTwin(e);
+
+                if (!retrying && moduleClient != null && !(e is TimeoutException))
```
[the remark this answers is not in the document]
philipktlin: Yes.  you are right.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
```diff
@@ -75,7 +75,7 @@ await currentDeviceConnection
         public Task RemoveDeviceConnection(string id)
         {
             return this.devices.TryGetValue(Preconditions.CheckNonWhiteSpace(id, nameof(id)), out ConnectedDevice device)
-                ? this.RemoveDeviceConnection(device, false)
+                ? this.RemoveDeviceConnection(device, true)
```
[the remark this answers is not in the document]
philipktlin: will do.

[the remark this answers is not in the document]
varunpuranik: Can we separate the 2 fixes out?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/Microsoft.Azure.Devices.Edge.Hub.Core.csproj
```diff
@@ -30,7 +30,7 @@
   <ItemGroup>
     <PackageReference Include="App.Metrics" Version="3.0.0" />
     <PackageReference Include="JetBrains.Annotations" Version="2018.3.0" />
-    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.23.1" />
+    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.23.2" />
```
varunpuranik, line 33: Let's also separate out this into another PR as well.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
varunpuranik, line 78: Let's remove the flag altogether..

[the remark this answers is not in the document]
philipktlin: Checked with Dylan. His run for connectivity test shows negative due to another setting issue.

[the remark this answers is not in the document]
philipktlin: will split it.

philipktlin: will do in another PR.

varunpuranik: Actually, let's add a setting (env var) to EdgeHub close the cloud connection when device is gone or not. Default will be to close the cloud connection, but the flag lets you turn it off.

philipktlin: 1. In what case we would like to keep the cloud connection even the module is removed from deployment?
2. There are totally 5 usage references and 2 of them passed false in 2nd argument.  One is this usage.  @varunpuranik  I would like to clarify, does env var setting to control all 5 usages or only those usages currently passing false as 2nd argument?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/Microsoft.Azure.Devices.Edge.Hub.Core.csproj
philipktlin: ok.

test/Microsoft.Azure.Devices.Edge.Test.Common/Microsoft.Azure.Devices.Edge.Test.Common.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
philipktlin, line 1: Tried to manually revert this, but no difference.  Don't know why it shows up.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -354,9 +372,9 @@ public static void ErrorUpdatingReportedProperties(Exception ex)
                 Log.LogDebug((int)EventIds.ErrorUpdatingReportedProperties, ex, "Error updating reported properties in IoT Hub");
             }
 
-            public static void GettingModuleClient()
+            public static void GettingModuleClient(bool retrying)
             {
-                Log.LogDebug((int)EventIds.GettingModuleClient, "Getting module client to refresh the twin");
+                Log.LogDebug((int)EventIds.GettingModuleClient, $"Getting module client to refresh the twin with retrying set to {retrying}");
```
varunpuranik, line 377: This is probably not useful anymore? Maybe we should remove

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/ModuleConnection.cs
```diff
@@ -134,7 +134,7 @@ enum EventIds
                 ReceivedMethodCallback,
                 DisposingModuleConnection,
                 DisposedModuleConnection,
-                ErrorDisposingModuleConnection
+                ErrorDisposingModuleConnection,
```
varunpuranik, line 137: ? revert?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
philipktlin: it is good to display in the log to indicate that this call is retrying or not.   I think we should keep this.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/ModuleConnection.cs
philipktlin: good catch.

PR #3076, opened 2020-06-11
Broker / EdgeHub Packaging and Build Infra

and-rewsmith:


edge-hub/docker/linux/amd64/Dockerfile.backup
```diff
@@ -0,0 +1,36 @@
+ARG base_tag=3.1.4-alpine3.11
```
and-rewsmith, line 1: reminder: delete

PR #3095, opened 2020-06-15
Added bridge component to collaborate with a generic MQTT broker

vipeller:
As the MQTT endpoint is being replaced by a generic broker, a bridge is needed between EdgeHub and the Broker. The purpose of this bridge is to transform EdgeHub events/messages/operations into MQTT messages, and to transfrom MQTT messages into EdgeHub events.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IConnectionRegistry.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    using System.Threading.Tasks;
+
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Device;
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Identity;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public interface IConnectionRegistry
+    {
+        Task<Option<IDeviceListener>> GetUpstreamProxyAsync(IIdentity identity);
```
varunpuranik, line 12: Nit: suggested names - GetDeviceListener/GetDeviceProxy to be inline with the current naming

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessorBase.cs
```diff
@@ -21,16 +21,34 @@ protected SubscriptionProcessorBase(IConnectionManager connectionManager)
         public Task AddSubscription(string id, DeviceSubscription deviceSubscription)
         {
             Events.AddingSubscription(id, deviceSubscription);
-            this.ConnectionManager.AddSubscription(id, deviceSubscription);
-            this.HandleSubscriptions(id, new List<(DeviceSubscription, bool)> { (deviceSubscription, true) });
+            var hasChanged = this.ConnectionManager.AddSubscription(id, deviceSubscription);
+
+            if (hasChanged)
```
varunpuranik, line 26: Maybe we should make this change in Master?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Authentication/AuthAgentController.cs
```diff
@@ -44,6 +44,12 @@ public async Task<JsonResult> HandleAsync([FromBody] AuthRequest request)
                 return this.Json(GetErrorResult());
             }
 
+            // FIXME: this can be removed, one the broker skips authentication for special clients
+            if (this.IsSystemComponent(request))
```
varunpuranik, line 48: That change has been merged I believe?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/BridgeConnection/MqttBridgeProtocolHead.cs
```diff
@@ -0,0 +1,83 @@
...
+    using Microsoft.Azure.Devices.Edge.Hub.Core;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public class MqttBridgeProtocolHead : IProtocolHead
+    {
+        readonly MqttBridgeProtocolHeadConfig config;
+        readonly IMqttBridgeConnector connector;
+
+        public string Name => "EH-BRIDGE";
```
varunpuranik, line 16: This is printed in the logs - can we just call it MQTT Broker Head?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/BridgeConnection/MqttBridgeProtocolHead.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        public MqttBridgeProtocolHead(MqttBridgeProtocolHeadConfig config, IMqttBridgeConnector connector)
+        {
+            this.config = Preconditions.CheckNotNull(config);
+            this.connector = connector;
```
varunpuranik, line 21: Nit; null check?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/BridgeConnection/MqttBridgeConnector.cs
```diff
@@ -0,0 +1,444 @@
...
+            {
+                producer.SetConnector(this);
+            }
+        }
+
+        public async Task ConnectAsync(string serverAddress, int port)
+        {
+            Events.Starting();
+
+            var client = default(MqttClient);
```
varunpuranik, line 48: nit: remove?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/BridgeConnection/MqttBridgeConnector.cs
```diff
@@ -0,0 +1,444 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Channels;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using uPLibrary.Networking.M2Mqtt;
```
varunpuranik, line 12: Couple of questions - 
1. How did you choose this lib? 
2. Any licensing concerns?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/BridgeConnection/MqttBridgeConnector.cs
```diff
@@ -0,0 +1,444 @@
...
+                {
+                    Events.StartedWhenAlreadyRunning();
+                    throw new InvalidOperationException("Cannot start mqtt-bridge connector twice");
+                }
+                else
+                {
+                    this.mqttClient = Option.Some(new MqttClient(serverAddress, port, false, MqttSslProtocols.None, null, null));
+                }
+
+                client = this.mqttClient.Expect(() => new Exception("No mqtt-bridge connector instance found to setup"));
```
varunpuranik, line 62: A better pattern might be - 
client = new MqttClient()
this.mqttClient = Option.Some(client)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/BridgeConnection/MqttBridgeConnector.cs
varunpuranik: We should hide all the client related details behind interfaces, so that we can swap it out for a different lib at a later point, if needed.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/BridgeConnection/IMqttBridgeConnector.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using System.Threading.Tasks;
+
+    public interface IMqttBridgeConnector
```
varunpuranik, line 6: Super nit: Can we call these classes/interfaces MqttBroker* instead of MqttBridge*. MqttBridge has a chance to confuse with an actual Mqtt Bridge

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/BridgeConnection/IMqttBridgeConnector.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
```
varunpuranik, line 2: Super nit: for consistency, can we pls camelCase the folder names (bridgeConnection)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/TwinHandler.cs
```diff
@@ -0,0 +1,350 @@
...
+                            var message = new EdgeMessage.Builder(publishInfo.Payload).Build();
+                            await proxy.UpdateReportedPropertiesAsync(message, rid);
+                        },
+                        match,
+                        publishInfo);
+        }
+
+        async Task<bool> HandleUpstreamRequest(Func<IDeviceListener, string, Task> action, Match match, MqttPublishInfo publishInfo)
+        {
+            var id1 = match.Groups["id1"];
```
varunpuranik, line 199: This seems very loosely coupled. I think a better pattern is - when you do the match, you get the parameters into an object, and pass that object instead.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IMessageConsumer.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    public interface IMessageConsumer
```
varunpuranik, line 7: So the pattern you have here is that all "handlers" are consumers, and they consume the message that is relevant to them?
Would it be easier (for testing and for understanding) that there is one component that figures out the "type" of the message and posts on the right channel for a particular "handler"?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IMessageConsumer.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        Task<bool> HandleAsync(MqttPublishInfo publishInfo);
+        void ProducerStopped();
```
varunpuranik, line 10: Can the producer restart? If so, how can it signal to the consumer it has restarted? If not, then maybe we can just use the dispose pattern.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/TwinHandler.cs
```diff
@@ -0,0 +1,350 @@
...
+            public static void BadPayloadFormat(Exception e) => Log.LogError((int)EventIds.BadPayloadFormat, e, "Bad payload format: cannot deserialize subscription update");
+            public static void ProcessingLoopStarted() => Log.LogInformation((int)EventIds.ProcessingLoopStarted, "Processing loop started");
+            public static void ProcessingLoopStopped() => Log.LogInformation((int)EventIds.ProcessingLoopStopped, "Processing loop stopped");
+            public static void ErrorEnqueueingNotification() => Log.LogError((int)EventIds.ErrorEnqueueingNotification, "Error enqueueing notification");
+            public static void ErrorProcessingNotification(Exception e) => Log.LogError((int)EventIds.ErrorProcessingNotification, e, "Error processing [Twin] notification");
+        }
+
+        class ProcessingInfo
+        {
+            public enum GetOrSet
```
varunpuranik, line 332: Nit: Should we call it Action - GetDesiredProperties/SetReportedProperties for clarity?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/TwinHandler.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                default:
+                    Events.BadIdentityFormat(identity.Id);
+                    throw new Exception($"cannot decode identity {identity.Id}");
+            }
+        }
+
+        static string GetDesiredPropertiesUpdateTopic(IIdentity identity, string version)
+        {
+            var identityComponents = identity.Id.Split(HandlerUtils.IdentitySegmentSeparator, StringSplitOptions.RemoveEmptyEntries);
```
varunpuranik, line 274: if((identity is IModuleIdentity moduleIdentity)
{}
else if(identity is IDeviceIdentity deviceIdentity)
{
}
else
{
}

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/ConnectionHandler.cs
```diff
@@ -0,0 +1,294 @@
...
+
+        readonly Channel<MqttPublishInfo> notifications;
+        readonly Task processingLoop;
+
+        AsyncLock guard = new AsyncLock();
+
+        // Normal dictionary would be sufficient because of the locks, however we need AddOrUpdate()
+        ConcurrentDictionary<IIdentity, IDeviceListener> knownConnections = new ConcurrentDictionary<IIdentity, IDeviceListener>();
+
+        // this class is auto-registered so no way to implement an async activator.
```
varunpuranik, line 37: What do you mean auto-registered? We have a pattern to do this with Autofac using a Create method..

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/ConnectionHandler.cs
```diff
@@ -0,0 +1,294 @@
...
+        // IConnectionProvider is registered
+        public ConnectionHandler(Task<IConnectionProvider> connectionProvider, IIdentityProvider identityProvider, ISystemComponentIdProvider systemComponentIdProvider, DeviceProxy.Factory deviceProxyFactory)
+        {
+            if (!connectionProvider.IsCompleted)
+            {
+                // if this leads to a dead-lock, at least it gets logged.
+                Events.BlockingDependencyInjection();
+            }
+
+            this.connectionProvider = connectionProvider.Result;
```
varunpuranik, line 48: Pls don't do this.. we use only async/await everywhere in the code as a principle.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/ConnectionHandler.cs
```diff
@@ -0,0 +1,294 @@
...
+            this.processingLoop = this.StartProcessingLoop();
+        }
+
+        public IReadOnlyCollection<string> Subscriptions => subscriptions;
+
+        public async Task<Option<IDeviceListener>> GetUpstreamProxyAsync(IIdentity identity)
+        {
+            using (await this.guard.LockAsync())
+            {
+                if (this.knownConnections.TryGetValue(identity, out var result))
```
varunpuranik, line 69: Nit: don't use var where the type if not obvious

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/ConnectionHandler.cs
```diff
@@ -0,0 +1,294 @@
...
+
+        async Task HandleDeviceConnectedAsync(MqttPublishInfo mqttPublishInfo)
+        {
+            var updatedIdentities = this.GetIdentitiesFromUpdateMessage(mqttPublishInfo);
+            await updatedIdentities.ForEachAsync(async i => await this.ReconcileConnectionsAsync(new HashSet<IIdentity>(i)));
+        }
+
+        async Task ReconcileConnectionsAsync(HashSet<IIdentity> updatedIdentities)
+        {
+            var identitiesAdded = default(HashSet<IIdentity>);
```
varunpuranik, line 127: Nit: no need to define outside the lock

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/GenericSubscriptions.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using System.Collections.Generic;
+
+    public class GenericSubscriptions : ISubscriber
```
varunpuranik, line 6: How is this used?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ISubscriber.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using System.Collections.Generic;
+
+    public interface ISubscriber
```
varunpuranik, line 6: What is difference between a ISubscriber and IMessageConsumer?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IMessageProducer.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    public interface IMessageProducer
+    {
+        void SetConnector(IMqttBridgeConnector connector);
```
varunpuranik, line 6: Do we need this? Why can't the connector be passed in as part of initialization?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ISystemComponentIdProvider.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    public interface ISystemComponentIdProvider
+    {
+        string EdgeHubBridgeId { get; }
```
varunpuranik, line 6: Maybe we can make this a const and remove this interface?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/MqttPublishInfo.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    public class MqttPublishInfo
+    {
+        public MqttPublishInfo(string topic, byte[] payload)
+        {
+            this.Topic = topic;
```
varunpuranik, line 8: Preconditions checks?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/SubscriptionPattern.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Device;
+
+    public class SubscriptionPattern
+    {
+        public SubscriptionPattern(string pattern, DeviceSubscription subscription)
+        {
+            this.Pattern = pattern;
```
varunpuranik, line 10: Preconditions checks?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/TwinHandler.cs
```diff
@@ -0,0 +1,350 @@
...
+                FailedToSendTwinUpdateMessage,
+                FailedToSendDesiredPropertiesUpdateMessage,
+                BadPayloadFormat,
+                ProcessingLoopStarted,
+                ProcessingLoopStopped,
+                ErrorEnqueueingNotification,
+                ErrorProcessingNotification
+            }
+
+            public static void TwinUpdate(string id, string statusCode, string correlationId, int messageLen) => Log.LogDebug((int)EventIds.TwinUpdate, $"Twin Update sent to client: [{id}], status: [{statusCode}], rid: [{correlationId}], msg len: [{messageLen}]");
```
varunpuranik, line 312: Nit: We don't use the [] pattern elsewhere. I feel coming up with a new pattern for logs will make the EdgeHub logs look very inconsistent.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/SubscriptionChangeHandler.cs
```diff
@@ -0,0 +1,119 @@
...
+                        if (subscriptionMatch.IsMatchWithIds(id1, id2))
+                        {
+                            subscribes = true;
+                            break;
+                        }
+                    }
+
+                    try
+                    {
+                        await proxy.AddOrRemoveSubscription(subscribes, subscriptionPattern.Subscrition);
```
varunpuranik, line 90: So, here it seems like we will handle all subscriptions on a subscription change? Can we instead only handle changed subscriptions?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ISubscriptionWatcher.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using System.Collections.Generic;
+
+    public interface ISubscriptionWatcher
+    {
+        IReadOnlyCollection<SubscriptionPattern> WatchedSubscriptions { get; }
```
varunpuranik, line 8: 2 questions - 
1. Should the consumer/subscriber/SubscriptionWatcher be the same interface?
2. If we move all the subscription matching to one class as I suggested above, then we don't really need this?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/SubscriptionPattern.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Device;
+
+    public class SubscriptionPattern
+    {
+        public SubscriptionPattern(string pattern, DeviceSubscription subscription)
```
varunpuranik, line 8: Feels like this can be a simple map?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/ConnectionHandler.cs
vipeller: By auto-registered I mean that the ConnectionHandler is captured by its interfaces among other types (implementing the same interfaces), like the following (it is from MqttBridgeModule.cs, around line 38):

var componentTypes = MqttBridgeComponentDiscovery.GetCandidateTypes().ToArray();
builder.RegisterTypes(componentTypes)
                   .AsSelf()
                   .As(t => GetNonStandardBridgeInterfaces(t))
                   .SingleInstance();

Because of this, it cannot have that specific activator than many types in the xxxModule.cs files, with manually resolve their dependencies using 'await'.

One thing I can do is to change the 'auto-register' to a 'manual-register', meaning that I list all the classes manually, registering them - and in case of ConnectionHandler, I can create an async activator to avoid the x.Result in the code.

Just a side note, that using that many custom activators (basically manually resolve the dependencies of an instance) doesn't seem a good pattern - and it comes from a few Task<T> registrations. Later, every type using those Task<T>s needs a custom activator, and then the types using these later types need too, making a cascading need of custom activators. This decreases the power of dependency injection frameworks - just like in this current example. Maybe the long term solution would be to eliminate the Task<T> registrations.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/GenericSubscriptions.cs
vipeller: The purpose of this class was originally that because more handlers were interested in "$edgehub/+/subscriptions", to avoid repeated subscriptions, there were to options:
- the framework that collects subscriptions, after collecting all, it eliminates the duplications
- move the subscriptions that more handlers interested are into a separate class.

this GenericSubscriptions is a solution of the second option. Since then the handler of "$edgehub/+/subscriptions" was moved into a single class (instead of different handlers handling their of subscription changes), so maybe this class can be removed (and the "$edgehub/+/subscriptions" be moved into that singe class, I will check this option)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/SubscriptionChangeHandler.cs
vipeller: There was an option to maintain a state, so at this level we would know that when the broker sends an updated subscription list, that is a change for a given subscription or not - but this state is also maintained by SubscriptionProcessor/ConnectionManager. I didn't want to duplicate this state (afraid of getting out of sync and causing problems). Instead I wanted ConnectionManager to be the source of truth. There is a related change that ConnectionManager reports back that it was an actual change or not, so the subscription processor will not do unnecessary updates at IoT Hub. With that change, the "unnecessary" calls don't seem that big performance hit, also subscriptions don't happen often.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IMessageConsumer.cs
vipeller: I was considering this and thought that the "figuring out" part would be a big 'if' with trying to match different patterns to see if that matches to a pattern of handler. If I walk that way, there are some options:
- the regular expression for matching (or the operation to do for matching) is in the 'smart' class (e.g. the  BridgeConnector), a whole list of them. In this case there is a handler-related string (or function) in the 'smart' class, that I didn't like, the code is not cohesive (related parts are separated)
- the regular expression (or a function for match) is in the handler class, and the 'smart' class asks for it. In this case the cohesion is better. But if I do this, it is not too far from the current solution:
- Just ask all handlers I know about (which was discovered by a marker interface of IMessageConsumer), and send them all the information, so they can handle it if they want, but if not, it is not a 'smart' class concern (and that way it is not a smart class anymore)

I think this solution is more flexible, supports the scenario when different handlers handle different aspect of the same message - maybe there is no such scenario right now, but there was before the subscription watcher class was created. Before SubscriptionWathcer, every handler took care of their own subscription changes, so multiple handlers processed the subscription change message.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IMessageProducer.cs
vipeller: That is because a circular dependency between the instances. The BridgeConnector has a reference to all IMessageProducers, and every message producer has a reference to the BridgeConnector.  It is a bit like the BindDeviceProxy() of the DeviceMessageHandler class

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ISubscriber.cs
vipeller: Logically it is two separate things to decide what messages should come from the broker and who should process those messages. One example was for that the 'subscription change' which was processed by multiple consumers, but there was a separate subscriber for the messages itself.
With the introduction of subscription watcher, this separation doesn't seem important right now, and the two interfaces could be merges, as right now the same class processes the message (consumes) than it subscribed to it.
I am not sure if it will not be handy to have the separation later.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ISubscriptionWatcher.cs
vipeller: The Consumer/Subscriber could be merged as today they go in pairs (all classes implementing 'subscriber' also implements 'consumer' and it consumes the topic 'subscriber' subscribed to)
SubscriptionWatcher is a different thing and it is not related to 'Subscriber' even if the name implies: in case of a 'Subscriber', the implementor class has needs to see a specific topic (so then later it can receive those messages). In case of SubscriptionWatcher, the implementor class needs to no if a connected client subscribed to a specific topic. 
Basically it is the opposite direction: if I am a subscriber, I need the messages. If I am a SubscriptionWatcher, I watch who subscribed to a topic, so I can provide them messages.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ISystemComponentIdProvider.cs
vipeller: It has the device id in it, so the content is dynamic. I was considering simple indexed/keyed string parameter, and register the current id with autofac, but then a key attribute needed to add in MqttBrokerAdapter, so I would have needed to add an autofac dependency to the project. That is how I chose to create a separate interface, so I didn't need keyed string for injection

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/SubscriptionPattern.cs
vipeller: Yes, but it always iterates through the entire collection it gathers from the subscription watcher instances - 'pattern' is a regular expression it needs to match, not a simple string for lookup.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IMessageConsumer.cs
varunpuranik: Yes, this solution is certainly most flexible, but I wonder if we need this flexibility, since we have a fixed (small) set of scenarios only, and the list is not likely to grow/change. The concerns being - now we have loops in each handler, and now we need to make sure each task/loop works correctly, etc. Also, as I said, it might be easier to make sure the match works or log it if it doesn't work. 

In any case, this is not a huge concern - it is just an opinion. I am happy to discuss with you if you like this approach better, and we can figure out.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/ConnectionHandler.cs
varunpuranik: I agree with your last point, we need to look into how to make DI work with async creators. But passing in a Task in the ctor and then getting the Result doesn't seem like a good idea. So maybe we can just go with the current pattern for now..

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IMessageProducer.cs
varunpuranik: How is the producer used in the MqttBridgeConnector? I don't see any usage other than to set the connector.

vipeller: They just come with in the same packet with the discovery object (discovery contains all components). I review that pattern if this can be eliminated

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/TwinHandler.cs
vipeller: removed the helper function, so these are not passed as parameters anymore

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IMessageConsumer.cs
vipeller: It cannot restart, however the consumers are injected services, so calling Dispose on them by the user class (the connector in this case) doesn't seem right to me - they don't own the instance. From the practical side, Dispose would be called by autofac, too - it doesn't cause problem (double dispose call), but to me it shows that that maybe Dispose() is not the right place to signal the consumer.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/TwinHandler.cs
vipeller: I also wanted to call it 'Action' but then the compiler made me realize that this word is already used. I can try "operation" or similar

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/TwinHandler.cs
vipeller: ended up with switch:
switch (identity)
            {
                case IModuleIdentity moduleIdentity: ..
                case IDeviceIdentity deviceIdentity: ..
                default:

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/GenericSubscriptions.cs
vipeller: Made an attempt to remove the GenericSubscription, however it didn't work:
- The handler class for the topics the GenericSubscription subscribes is the 'SubscriptionChangeHandler'
- The 'SubscriptionChangeHandler' is not a normal handler class. It was created because there were three different handler classes handling their own subscription changes, but the related code was almost the same in all 3 classes. That code was moved out into 'SubscriptionChangeHandler'
- The 'SubscriptionChangeHandler' needs to know about other edgehub-bridge components (because it needs to find whose subscriptions it needs to handle), so it has a dependency to ComponentDiscovery.
- Because of that dependency, it cannot be 'discovered' by ComponentDiscovery, (that would be a circular dependency)
- Because if this, SubscriptionChangeHandler is injected directly into MqttBrokerConnector, and the connector uses it apart from the rest of the handlers (it asks SubscriptionChangeHandler first, then the rest of the handlers)

because all of that, SubscriptionChangeHandler cannot just implement ISubscriber listing its subscriptions, so I need to keep a separate class for that. For the same reason I will not be able to merge ISubscriber and IMessageConsumer.

I keep thinking if there is a better way to implement these interfaces

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IMessageProducer.cs
vipeller: I made an attempt to remove the SetConnector call, but it is not that simple. Certain classes implement both IMessageConsumer and IMessageProducer. The connector needs to know all the message consumers (because it has a loop asking all of them to process incoming messages). On the other hand MessageProducers need to know about the connector, so they can send out messages. If a class implements both interfaces, then it has a circular dependency with the connector

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/ConnectionHandler.cs
vipeller: I fixed it the same than TwinController already does: it still gets a Task<T>, but instead of waiting in the constructor for the task to complete, it stores it as a "getter", and in the method it needs it (which is already an async menthod), it gets the connectionProvider with provider = await connectionProviderGetter (which is the Task<ConnectionProvider>)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/ConnectionHandler.cs
vipeller: fixed by storing the Task<x> and using with 'await' when connectionProvider is necessary

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/TwinHandler.cs
vipeller: removed, however it is a great pattern - important bits catch the eyes and easier to search (e.g. for [12] vs 12 where the latter stops on every time stamp :) )

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ISubscriptionWatcher.cs
vipeller: Refactored this the following way:
- ISubscriber/IMessageConsumer is merged at a price that SubscriptionChangeHandler cannot use ComponentDiscovery to find the classes (that is to eliminate circular dependencies), but it gets all the interested components injected directly in its constructor
- Removed the interface ISubscriptionWatcher at the price that its single method is added to the interface of all three original implementor.

Now
- Have only two interfaces from the original four (IMessageConsumer/Producer)
- The main loop in the connector is simpler (it does not have SubscriptionChangeHandler as a special case, but it works as a normal MessageConsumer)

But
- the SubscriptionChangeHandler cannot pick up automatically the classes that needs subscription changes to be handled, but needs to add to the parameter list of SubscriptionChangeHandler. More over, that class does not need the instances for other reason just to get the matching patterns. It could have receive a list, but that requires either keyed parameters (-> adding dependency to autofac), or manual registration (-> does not work with ComponentDiscovery)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/Handlers/GenericSubscriptions.cs
vipeller: With a recent bigger change, managed to remove GenericSubcriptions

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ISubscriber.cs
vipeller: At the end, merged ISubscriber into IMessageConsumer

vipeller:
obsolete, merged the changes from another request

PR #3103, opened 2020-06-15
[mqtt] MQTT broker edgehub aware authorization

dmolokanov:
Introduces authorization logic to enforce basic authz rules integrating with EdgeHub.

## LocalAuthorizer
LocalAuthorizer allows to perform any `CONNECT`, `SUBSCRIBE`, `PUBLISH` operations for clients connected from the localhost. Intended to use to allow internal communication between EdgeHub and Broker.
When the client connected not from localhost it delegates authorization rules enforcement to `EdgeHubAuthorizer`.

## EdgeHubAuthorizer
EdgeHubAuthorizer identifies what topics client requests access to a validates whether the operation is allowed.
#### CONNECT
- Anonymous clients cannot connect to the broker
- Only clients that were authenticated by EdgeHub can connect to the broker. `client_id` should be equal `auth_id`

#### SUBSCRIBE
1. A client **cannot** subscribe to
  - receive everything (`#`)
  - any special ($-prefixed) topics (e.g. `$edgehub/#`, `$iothub/#`, `$upstream/#`, `$downstream/#`, `$SYS/#`, etc)

2. A client **can** subscribe to some IoTHub topics to receive
  - _its own_ twin object
  - C2D messages
  - direct method requests

3. A client **can** subscribe to any non-special MQTT topic

#### PUBLISH
1. A client **cannot** publish to
  - any special ($-prefixed) topics (e.g. `$edgehub/foo`, `$iothub/foo`, `$upstream/foo`, `$downstream/foo`, `$SYS/foo`, etc..)

2. A client **can** publish to some IoTHub topics
  - request _its own_ twin object
  - send telemetry event
  - send desired property update
  - send reported property update

3. A client **can** publish to any non-special MQTT topic


mqtt/build/linux/clippy.sh
```diff
@@ -61,3 +61,4 @@ echo "Running clippy..."
 $CARGO clippy --all
 $CARGO clippy --all --tests --all-features
 $CARGO clippy --all --examples
+$CARGO clippy --all --benches
```
vadim-kovalyov, line 64: :+1:

mqtt/mqtt-broker/src/lib.rs
```diff
@@ -183,3 +192,12 @@ pub enum Message {
     Client(ClientId, ClientEvent),
     System(SystemEvent),
 }
+
+#[cfg(test)]
+pub(crate) mod tests {
```
vadim-kovalyov, line 197: is this needed?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -0,0 +1,307 @@
...
+    auth::{Activity, AuthId, Authorization, Authorizer, Connect, Operation, Publish, Subscribe},
+    ClientId, ClientInfo,
+};
+
+#[derive(Debug, Default)]
+pub struct EdgeHubAuthorizer;
+
+#[allow(clippy::unused_self)]
+impl EdgeHubAuthorizer {
+    pub fn new() -> Self {
```
vadim-kovalyov, line 13: just curious why parameterless `new` if we have `default`?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -0,0 +1,307 @@
...
+    ) -> Authorization {
+        match client_info.auth_id() {
+            // forbid anonymous clients to connect to the broker
+            AuthId::Anonymous => Authorization::Forbidden("anonymous client".to_string()),
+            // allow only those clients whose auth_id and client_id identical
+            AuthId::Identity(identity) => {
+                if identity == client_id.as_str() {
+                    Authorization::Allowed
+                } else {
+                    Authorization::Forbidden(format!("unknown client: {}", client_id))
```
vadim-kovalyov, line 31:
```suggestion
                    Authorization::Forbidden(format!("client id {} does not match registered iothub identity id.", client_id))
```

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn authorize_iothub_topic(
+        &self,
+        client_id: &ClientId,
+        client_info: &ClientInfo,
+        topic: &str,
+    ) -> Authorization {
+        match client_info.auth_id() {
+            // forbid anonymous clients to subscribe to restricted topics
+            AuthId::Anonymous => {
+                Authorization::Forbidden("IoTHub topic: anonymous client".to_string())
```
vadim-kovalyov, line 78:
```suggestion
                Authorization::Forbidden("Anonymous clients do not have access to iothub topics".to_string())
```

mqtt/mqtt-broker/src/lib.rs
dmolokanov: nope, it is rust_analyzer generated snippet.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
dmolokanov: during the active development phase, it is convenient to have new method. usually, it tends to accept some parameters over time. in this case, it may not require anything

varunpuranik:
The approach looks good overall. 
2 points -
1. System topics: Should we block all system topics and only allow specific $iothub topics 
2. Telemetry over non-system topics: For now, I think we should allow the following topics only for that identity - 
devices/{deviceId}/messages/events
devices/{deviceId}/modules/{moduleId}/messages
devices/{deviceId}/messages/devicebound

dmolokanov:
> The approach looks good overall.
> 2 points -
> 
>     1. System topics: Should we block all system topics and only allow specific $iothub topics
Yeah, probably we can be more restrictive at this moment.
> 
>     2. Telemetry over non-system topics: For now, I think we should allow the following topics only for that identity -
>        devices/{deviceId}/messages/events
>        devices/{deviceId}/modules/{moduleId}/messages
>        devices/{deviceId}/messages/devicebound

@varunpuranik authorization rules take place **after** topic converted into an internal format. So at that point in time broker and authorizer handles converted topics. That's why you don't see rules for telemetry topics in "old" format.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    fn authorize_iothub_topic(
+        &self,
+        client_id: &ClientId,
+        client_info: &ClientInfo,
+        topic: &str,
+    ) -> Authorization {
+        match client_info.auth_id() {
+            // forbid anonymous clients to subscribe to restricted topics
+            AuthId::Anonymous => Authorization::Forbidden(
```
varunpuranik, line 88: Do we need this explicitly (since we don't let anon clients to connect)?

dmolokanov: Correct, but I'd like to check in place. We may want to let anon clients connect to access generic MQTT topics but forbid iothub related topics.

varunpuranik:
> @varunpuranik authorization rules take place **after** topic converted into an internal format. So at that point in time broker and authorizer handles converted topics. That's why you don't see rules for telemetry topics in "old" format.

Got it, makes sense. 
So will we block client1 trying to publish on "devices/client2/messages/events"?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -0,0 +1,336 @@
...
+    topic_filter.starts_with('$')
+}
+
+fn is_iothub_topic(topic: &str) -> bool {
+    topic.starts_with("$edgehub/") || topic.starts_with("$iothub/")
+}
+
+fn allowed_iothub_topic(client_id: &ClientId) -> Vec<String> {
+    vec![
+        format!("$edgehub/clients/{}/messages/events", client_id),
```
varunpuranik, line 128: what will be the client_id for modules? deviceId/moduleId?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -0,0 +1,336 @@
...
+
+fn is_forbidden_topic(topic_filter: &str) -> bool {
+    topic_filter.starts_with('$')
+}
+
+fn is_iothub_topic(topic: &str) -> bool {
+    topic.starts_with("$edgehub/") || topic.starts_with("$iothub/")
+}
+
+fn allowed_iothub_topic(client_id: &ClientId) -> Vec<String> {
```
varunpuranik, line 126: Will this allocate for every pub? Can we do better? 
(regex or caching come to mind)

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -0,0 +1,336 @@
...
+        .iter()
+        .any(|forbidden_topic| topic_filter.starts_with(forbidden_topic))
+}
+
+fn is_forbidden_topic(topic_filter: &str) -> bool {
+    topic_filter.starts_with('$')
+}
+
+fn is_iothub_topic(topic: &str) -> bool {
+    topic.starts_with("$edgehub/") || topic.starts_with("$iothub/")
```
varunpuranik, line 123: nit: const?
Also, we will add $upstream when we use it (for the bridge)?

dmolokanov: Not sure we will add $up[down]stream here for the bridge. This authorizer is for non-local MQTT clients which should not be aware of $up[down]stream topics

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
dmolokanov: Yeah, that's what I think. Any concerns it won't work?

dmolokanov:
> So will we block client1 trying to publish on "devices/client2/messages/events"?

yes. `EdgeHubAuthorizer` should block accessing IoTHub topics for another client:
`devices/client2/messages/events` converted to `$edgehub/client2/messages/events` and then access (subscrive or publish) be blocked if the `client_id != "client2"`

dmolokanov:
Merging this PR. Will add fixes in the next one.

PR #3100, opened 2020-06-15
Remove device connection will always remove cloud connection

philipktlin:

PR #3109, opened 2020-06-16
Fix ConnectionManager tests

philipktlin:
Fix ConnectionManager tests since RemoveDeviceConnection will remove cloud connection as well in https://github.com/Azure/iotedge/pull/3100

### DIFF
--- a/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/ConnectionManagerTest.cs
+++ b/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/ConnectionManagerTest.cs
@@ -132,7 +132,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Test
             await connectionManager.RemoveDeviceConnection(deviceCredentials2.Identity.Id);
 
             returnedValue = await connectionManager.GetCloudConnection(deviceCredentials2.Identity.Id);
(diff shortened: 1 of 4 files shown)


edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/ConnectionManagerTest.cs
```diff
@@ -132,7 +132,7 @@ public async Task CloudConnectionTest()
             await connectionManager.RemoveDeviceConnection(deviceCredentials2.Identity.Id);
 
             returnedValue = await connectionManager.GetCloudConnection(deviceCredentials2.Identity.Id);
-            Assert.True(returnedValue.HasValue);
+            Assert.False(returnedValue.HasValue);
```
yophilav, line 135: Should `deviceCredentials1` be removed at the end of the test as well?

philipktlin: agreed that this is nice to clean up, but this is breaking anything and clean up may need for other tests in this test class as well.  It should be done in a separate PR and in master branch only.

philipktlin: Add more assert statement for deviceCredentials1 cloud proxy.

PR #3110, opened 2020-06-16
Added CI pipeline for mqtt

vadim-kovalyov:
In this PR:
- Added CI build pipeline for Amd64 (linux), Arm32 (Linux), Amd64 (Windows)
- Removed 3rd party compliance tests. We have moved most of them to our integration tests already.
- Enabled the pipeline to run on every commit to master/iiot with precondition check.


mqtt/build/linux/cross-test.sh
```diff
@@ -0,0 +1,66 @@
...
+                * ) usage;;
+            esac
+        fi
+    done
+}
+
+process_args "$@"
+
+if [[ -z ${RELEASE} ]]; then
+    cd "$PROJECT_ROOT" && IOTEDGE_HOMEDIR=/tmp cross test --all --target "$TARGET"
```
dmolokanov, line 63: we just figured that `--all` option deprecated in favor of `--workspace` should we fix that everywhere in scripts?

dmolokanov, line 63: I think it can be quite convenient to run benches as well. What do you think?

mqtt/build/linux/cross-test.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
dmolokanov, line 1: Do you think it is possible to combine test.sh and cross-test.sh into one?

mqtt/build/linux/cross.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
dmolokanov, line 1: Do you think it is possible to combine build.sh and cross.sh into one? I noticed we have `CARGO` variable in the script 
https://github.com/Azure/iotedge/blob/master/mqtt/build/linux/build.sh#L23
so I assume we can use it like this
```bash
CARGO=cross ./build.sh
```

builds/ci/mqtt.yaml
```diff
@@ -1,19 +1,127 @@
...
 jobs:
-  ################################################################################
-  - job: compliance_amd64
-    ################################################################################
-    displayName: Compliance amd64
+
+################################################################################
+  - job: check_run_pipeline
+################################################################################
+    displayName: Check pipeline preconditions (changes ARE in builds or mqtt)
```
philipktlin, line 12: why "ARE" in capital?

builds/ci/mqtt.yaml
```diff
@@ -1,19 +1,127 @@
...
+      - task: Bash@3
         displayName: Install Rust
-      - bash: mqtt/build/linux/build.sh -r 1
+        inputs:
+          filePath: mqtt/build/linux/install.sh
+      - script: cargo install cross --version 0.1.16
+        displayName: Install cross
+      - task: Bash@3
         displayName: Build
-      - bash: mqtt/build/linux/compliance.sh
```
philipktlin, line 17: do we still need to run this script?

vadim-kovalyov: Not anymore. all 3rd party tests were moved into rust.

builds/ci/mqtt.yaml
vadim-kovalyov: Don't know :) copy-paste!

mqtt/build/linux/cross.sh
vadim-kovalyov: Good point. I combined build.sh and test.sh

mqtt/build/linux/cross-test.sh
vadim-kovalyov: Yes, good point.

mqtt/build/linux/cross-test.sh
vadim-kovalyov: I'm not sure about that. Just adding benches won't be beneficial. We need to think about how we collect and validate results of benchmarks. I would suggest to tackle it in a separate task.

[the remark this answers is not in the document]
vadim-kovalyov: copy-paste :) removed

vadim-kovalyov:
> I assume we need to make similar changes to run checkins for all supported platforms, right?

@dmolokanov, do you want to add arm32 into checkin builds? We do windows and linux amd64 right now. It might be beneficial, but at the same time the probability of platform-dependent build issues is very low. I think windows/linux coverage is fine for checkin. The rest would be caught in CI. Thoughts?

builds/ci/mqtt.yaml
dmolokanov: I did it distinguish when task started but not supposed to.

builds/ci/mqtt.yaml
```diff
@@ -92,7 +92,7 @@ jobs:
         inputs:
           filePath: mqtt/build/windows/install.ps1
       - task: PowerShell@2
-        displayName: Clippy for windows
+        displayName: Clippy for Windows
```
dmolokanov, line 95: 🤣

mqtt/mqtt-broker/src/auth.rs
```diff
@@ -29,14 +29,14 @@ mod imp {
     use std::error::Error as StdError;
 
     use mqtt_broker_core::auth::{
-        authenticate_fn_ok, authorize_fn_ok, AuthId, Authenticator, Authorizer,
+        authenticate_fn_ok, authorize_fn_ok, AuthId, Authenticator, Authorization, Authorizer,
```
dmolokanov, line 32: Oops. Second time in a row. We probably need to add another step to build/check with `--no-default-features` flag

PR #3114, opened 2020-06-17
Update Device.Client private package to 1.26.0

richma-ms:
Re-built the private package for Device.Client SDK by applying Nested Edge changes on top of the 1.26.0 commit.  This change just updates our project references to match the new package.


edge-modules/MetricsCollector/MetricsCollector.csproj
```diff
@@ -1,46 +1,46 @@
...
-  <ItemGroup>
-    <AdditionalFiles Include="..\..\stylecop.json" Link="stylecop.json" />
-  </ItemGroup>
-  <PropertyGroup>
-    <CodeAnalysisRuleSet>..\..\stylecop.ruleset</CodeAnalysisRuleSet>
-  </PropertyGroup>  
-  <Import Project="..\..\stylecop.props" />
-
-</Project>
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
varunpuranik, line 1: (encoding causing this diff to look like this?)

richma-ms: ah crap, not sure why this happened again, I'll redo it

dylanbronson: This always happens for MetricsCollector when we update sdk - probably whoever wrote it had different default encoding from everyone else. No biggie either way

PR #3115, opened 2020-06-17
[PnP] Parse modelID from username in Mqtt connect packet and stamp every message 

dylanbronson:
The Azure IoT SDK passes the modelID on the MQTT CONNECT packet when a Plug and Play (pnp) device connects to hub. In order to integrate pnp with edge, edge needs to parse this modelID to pass along to the hub when edge's deviceClient connects to hub.

We cache the modelID in DeviceListener for all devices that connected with a modelID and stamp every message that comes from those devices with the associated modelID. Every message from a device with a modelID will have a ModelID in the SystemProperties. This will be used to create/recreate the cloud connection depending on if the modelID has changed.

Currently, this PR only handles MQTT. But we will add the modelID for amqp as well in a separate PR.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/ClientConnectionHandler.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/ClientConnectionHandler.cs
@@ -47,7 +47,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Amqp
                                 .GetOrElse(
                                     async () =>
                                     {
-                                        IDeviceListener dl = await this.connectionProvider.GetDeviceListenerAsync(this.identity);
+                                        // TODO: Implement plug and play for AMQP
+                                        IDeviceListener dl = await this.connectionProvider.GetDeviceListenerAsync(this.identity, Option.None<string>());
                                         var deviceProxy = new DeviceProxy(this, this.identity);
                                         dl.BindDeviceProxy(deviceProxy);
                                         this.deviceListener = Option.Some(dl);
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionProvider.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionProvider.cs
@@ -20,9 +20,9 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
             this.messageAckTimeout = messageAckTimeout;
         }
 
-        public Task<IDeviceListener> GetDeviceListenerAsync(IIdentity identity)
+        public Task<IDeviceListener> GetDeviceListenerAsync(IIdentity identity, Option<string> modelId)
         {
-            IDeviceListener deviceListener = new DeviceMessageHandler(Preconditions.CheckNotNull(identity, nameof(identity)), this.edgeHub, this.connectionManager, this.messageAckTimeout);
+            IDeviceListener deviceListener = new DeviceMessageHandler(Preconditions.CheckNotNull(identity, nameof(identity)), this.edgeHub, this.connectionManager, this.messageAckTimeout, modelId);
             return Task.FromResult(deviceListener);
         }
 
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
@@ -189,10 +189,13 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Device
 
         public Task ProcessDeviceMessageBatchAsync(IEnumerable<IMessage> messages)
         {
-            foreach (IMessage message in messages)
+            this.modelId.ForEach(modelId =>
             {
-                this.modelId.ForEach(modelId => message.SystemProperties[SystemProperties.ModelId] = modelId);
-            }
+                foreach (IMessage message in messages)
+                {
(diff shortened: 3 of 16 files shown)


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/ClientConnectionHandler.cs
```diff
@@ -47,7 +47,8 @@ public Task<IDeviceListener> GetDeviceListener()
                                 .GetOrElse(
                                     async () =>
                                     {
-                                        IDeviceListener dl = await this.connectionProvider.GetDeviceListenerAsync(this.identity);
+                                        // TODO: Implement plug and play for AMQP
```
richma-ms, line 50: > // TODO: Implement plug and play for AMQP [](start = 39, length = 42)

Put the task/PBI ID, otherwise if plans changes then chances are we won't clean up this stub.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionProvider.cs
```diff
@@ -20,9 +20,9 @@ public ConnectionProvider(IConnectionManager connectionManager, IEdgeHub edgeHub
             this.messageAckTimeout = messageAckTimeout;
         }
 
-        public Task<IDeviceListener> GetDeviceListenerAsync(IIdentity identity)
+        public Task<IDeviceListener> GetDeviceListenerAsync(IIdentity identity, Option<string> modelId)
```
philipktlin, line 23: if this modelId is pnp-domain specific, i think it is better to name it as pnpModelId

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
```diff
@@ -179,9 +181,24 @@ public async Task SendGetTwinRequest(string correlationId)
...
+            return this.edgeHub.ProcessDeviceMessage(this.Identity, message);
+        }
+
+        public Task ProcessDeviceMessageBatchAsync(IEnumerable<IMessage> messages)
+        {
+            this.modelId.ForEach(modelId =>
+            {
+                foreach (IMessage message in messages)
+                {
+                    message.SystemProperties[SystemProperties.ModelId] = modelId;
```
richma-ms, line 196: >message.SystemProperties[SystemProperties.ModelId] = modelId; [](start = 20, length = 61)

I think for edge-specific metadata, we avoid putting it as an actual message header.  Instead it gets wrapped into the various IMessage implementations.

The danger of putting the ID as a header is that we're modifying the original message, and it'll be difficult to maintain on the sending end the list of headers that we need to strip back out.

dylanbronson: So you believe we should add it as a field in the IMessage implementations and just deal with whatever conversions need to be done as it flows through routing?

I was planning on adding logic on the sending side that would just strip out any SystemProperty with SystemProperties.ModelID in it before sending. What difficulty do you foresee?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/ClientConnectionHandler.cs
philipktlin: Normally we don't want to put internal links in iotedge repo as it is an open source project.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
varunpuranik: IMessage doesn't let you wrap another IMessage. You would do that only to make the SystemProperties immutable. But I think this is fine for now. This is an existing pattern. 

Also the System properties have to be explicitly set when sending the messages out, and this is not a property that can be set by the client.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/DeviceIdentityProvider.cs
```diff
@@ -198,7 +200,16 @@ internal static (string deviceId, string moduleId, string deviceClientType) Pars
                 deviceClientType = string.Empty;
             }
 
-            return (deviceId, moduleId, deviceClientType);
+            Option<string> modelIdOption = Option.None<string>();
+            if (queryParameters.TryGetValue(ModelIdKey, out string modelId))
```
varunpuranik, line 204: Nit: combine the ifs

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/DeviceMessageHandlerTest.cs
```diff
@@ -34,14 +34,37 @@ public async Task ForwardsGetTwinOperationToEdgeHub()
...
+            var listener = new DeviceMessageHandler(identity, edgeHub.Object, connMgr, DefaultMessageAckTimeout, Option.None<string>());
             listener.BindDeviceProxy(deviceProxy.Object);
             await listener.SendGetTwinRequest("cid");
 
             edgeHub.Verify(x => x.GetTwinAsync(identity.Id), Times.Once);
             Assert.Same(expectedMessage, actualMessage);
         }
 
+        [Fact]
+        public async Task ProcessMessageBatchAsync_WithModelIdAsyncTest()
```
varunpuranik, line 46: We need a test for non-batched process message as well.

PR #3117, opened 2020-06-18
Make histogram store data for 1 hour

lfitchett:


lfitchett:
no, there's 2 values, maxAge and ageBuckets. More buckets is more precise, but costs more. maxAge just makes each bucket stick around longer before being replaced with a new bucket

PR #3118, opened 2020-06-18
Fix for MQTT Image CI pipeline.

vadim-kovalyov:
**Context:**
If we try to override `rust-toolchain` version, our build images fail, b/c they don't have specific version installed.

**Fix:**
Added a statement in Dockerfiles to force add specific rustup target, in case it has been overriden in `rust-toolchain` file

### DIFF
--- a/builds/misc/images-mqtt.yaml
+++ b/builds/misc/images-mqtt.yaml
@@ -33,9 +33,9 @@ jobs:
           containerRegistry: iotedge-edgebuilds-acr
           Dockerfile: mqtt/docker/$(os)/$(arch)/Dockerfile
           buildContext: mqtt/.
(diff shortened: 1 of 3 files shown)


builds/misc/images-mqtt.yaml
```diff
@@ -33,9 +33,9 @@ jobs:
           containerRegistry: iotedge-edgebuilds-acr
           Dockerfile: mqtt/docker/$(os)/$(arch)/Dockerfile
           buildContext: mqtt/.
-          tags: $(Build.BuildId)-$(os)-$(arch)
+          tags: $(Build.BuildNumber)-$(os)-$(arch)
```
dmolokanov, line 36: 👍

PR #3124, opened 2020-06-19
[mqtt] Disable max_queued_size by default

dmolokanov:
Default settings we have right now `max_queued_size: "10mb:` and `max_queued_messages: 1000`. In order to disable the message queue limit, one has to assign 0 to **both** settings which rather confusing.

The proposed change is to set `max_queued_size: 0` by default. It will disable check for the maximum amount of space message queue allocates. This way to remove any message queue limits it is required to set only **one** setting `max_queued_messages: 0` (still 1000 by default).

Makes size value to be parsed from a human-readable size (e.g 256kb) or number 32 (="32b").


mqtt/mqtt-broker/src/configuration/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+    unsaved_message_count: u32,
+}
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct SessionConfig {
+    #[serde(with = "humantime_serde")]
+    expiration: Duration,
+    max_message_size: Option<HumanSize>,
+    max_inflight_messages: usize,
+    max_queued_messages: usize,
```
varunpuranik, line 57: What if this is not specified in the config? It will default to 0 (= no limit)?

dmolokanov: we have `"max_queued_messages": 1000` and `"max_queued_size": null` (=no limits) by default

mqtt/mqtt-broker/src/configuration/size.rs
```diff
@@ -0,0 +1,187 @@
...
+
+use lazy_static::lazy_static;
+use regex::Regex;
+use serde::{Deserialize, Deserializer};
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct HumanSize(usize);
+
+impl HumanSize {
+    pub const fn new(bytes: usize) -> Self {
```
vadim-kovalyov, line 11: let's call it `new_bytes`

mqtt/mqtt-broker/src/configuration/size.rs
```diff
@@ -0,0 +1,187 @@
...
+        Self(bytes)
+    }
+
+    pub fn new_kilobytes(kilobytes: usize) -> Option<Self> {
+        kilobytes.checked_mul(1024).map(Self)
```
vadim-kovalyov, line 16: nit: I understand your urge to use combinators everywhere, but isn't it simpler to read and understand `Self( kilobytes.checked_mul(1024))`?

vadim-kovalyov: nvm, you still need to handle Option...

dmolokanov:
> @dmolokanov, if a customer wants to disable, let's say, `max_message_size`, what should they set in the config? `max_message_size: null`? will that override the default 256kb setting while merging default configs?

Good question, I didn't change any logic related to `max_message_size` because we don't support this setting right now. But IMHO it should use the same approach: 
When customer deletes some field from config broker should use the default value. If the default value is not set in `defaults.json`, `None` should be used.
In your example, customer should be able either just delete `max_message_size` or put `max_message_size: null`. The result will be the same. Right now we don't have ability to disable `max_message_size` check.

mqtt/mqtt-broker/src/configuration/size.rs
```diff
@@ -0,0 +1,249 @@
...
+use regex::Regex;
+use serde::{
+    de::{Error as SerdeError, Visitor},
+    Deserialize, Deserializer,
+};
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct HumanSize(usize);
+
+impl HumanSize {
```
vadim-kovalyov, line 13: nit: do you think we should name the file `human_size.rs`, by the name of the main struct in there?

mqtt/mqtt-broker/src/configuration/size.rs
```diff
@@ -0,0 +1,249 @@
...
+        let captures = SIZE_PATTERN
+            .captures(src)
+            .ok_or_else(|| ParseHumanSizeError::InvalidFormat(src.to_string()))?;
+
+        let value = captures[1]
+            .parse()
+            .map_err(|_| ParseHumanSizeError::Value(captures[1].to_string()))?;
+
+        match captures[2].to_lowercase().as_str() {
+            "" | "b" => Ok(Some(Self::new_bytes(value))),
```
vadim-kovalyov, line 53: :+1:

mqtt/mqtt-broker/src/session/state.rs
```diff
@@ -371,7 +371,7 @@ impl BoundedQueue {
     pub fn dequeue(&mut self) -> Option<proto::Publication> {
         match self.inner.pop_front() {
             Some(publication) => {
-                self.current_size -= publication.payload.len() as u64;
```
vadim-kovalyov, line 374: :+1:

mqtt/mqtt-broker/test/config_override_with_zero.json
```diff
@@ -0,0 +1,8 @@
+{
+    "session": {
```
vadim-kovalyov, line 2: nit: `no_session_limits.json`, btw, where is this used?

PR #3127, opened 2020-06-20
Edgehub E2E tests fix

richma-ms:
This disables authchain check for ConnectionProvider by default, which should fix the bulk of the E2E automation failures.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
@@ -33,6 +33,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
         readonly IIdentity edgeHubIdentity;
         readonly TimeSpan operationTimeout;
         readonly IProductInfoStore productInfoStore;
+        readonly bool nestedEdgeEnabled;
         Option<IEdgeHub> edgeHub;
 
         public CloudConnectionProvider(
@@ -49,7 +50,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             TimeSpan operationTimeout,
             bool useServerHeartbeat,
             Option<IWebProxy> proxy,
-            IProductInfoStore productInfoStore)
+            IProductInfoStore productInfoStore,
+            bool nestedEdgeEnabled = false)
         {
             this.messageConverterProvider = Preconditions.CheckNotNull(messageConverterProvider, nameof(messageConverterProvider));
             this.clientProvider = Preconditions.CheckNotNull(clientProvider, nameof(clientProvider));
@@ -66,6 +68,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             this.edgeHubIdentity = Preconditions.CheckNotNull(edgeHubIdentity, nameof(edgeHubIdentity));
             this.operationTimeout = operationTimeout;
             this.productInfoStore = Preconditions.CheckNotNull(productInfoStore, nameof(productInfoStore));
+            this.nestedEdgeEnabled = nestedEdgeEnabled;
         }
 
         public void BindEdgeHub(IEdgeHub edgeHubInstance)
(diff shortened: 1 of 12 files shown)


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -83,14 +86,20 @@ public async Task<Try<ICloudConnection>> Connect(IClientCredentials clientCreden
...
+
                 // Get the transport settings
-                Option<string> authChain = await this.deviceScopeIdentitiesCache.GetAuthChain(clientCredentials.Identity.Id);
                 ITransportSettings[] transportSettings = GetTransportSettings(
                     this.upstreamProtocol,
                     this.connectionPoolSize,
                     this.proxy,
                     this.useServerHeartbeat,
-                    authChain.Expect(() => new InvalidOperationException($"No auth chain for the client identity: {clientCredentials.Identity.Id}")));
+                    authChain);
```
varunpuranik, line 102: So this will set the authChain to empty string? It seems like it will be cleaner to pass the option and set the authChain on the transport settings object only if it is not empty.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -147,14 +156,19 @@ public async Task<Try<ICloudConnection>> Connect(IIdentity identity, Action<stri
                 Option<ServiceIdentity> serviceIdentity = (await this.deviceScopeIdentitiesCache.GetServiceIdentity(identity.Id))
                     .Filter(s => s.Status == ServiceIdentityStatus.Enabled);
 
-                // Get the transport settings
-                Option<string> authChain = await this.deviceScopeIdentitiesCache.GetAuthChain(identity.Id);
+                string authChain = string.Empty;
```
philipktlin, line 159: should we extract a method for reuse?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
```diff
@@ -103,7 +106,7 @@ async Task<(bool isAuthenticated, bool serviceIdentityFound)> AuthenticateWithSe
         {
             Option<string> authChain = await this.deviceScopeIdentitiesCache.GetAuthChain(serviceIdentityId);
 
-            if (!authChain.HasValue)
+            if (this.nestedEdgeEnabled && !authChain.HasValue)
```
philipktlin, line 109: Previously it will return as long as authChain doesn't have value.  why only return here if nested edge is enabled?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -159,6 +159,7 @@ void RegisterMqttModule(ContainerBuilder builder, (bool isEnabled, bool usePersi
             bool checkEntireQueueOnCleanup = this.configuration.GetValue("CheckEntireQueueOnCleanup", false);
             Option<string> gatewayHostname = Option.Maybe(this.configuration.GetValue<string>(Constants.ConfigKey.GatewayHostname));
             bool closeCloudConnectionOnDeviceDisconnect = this.configuration.GetValue("CloseCloudConnectionOnDeviceDisconnect", true);
+            bool nestedEdgeEnabled = this.configuration.GetValue<bool>(Constants.ConfigKey.NestedEdgeEnabled);
```
philipktlin, line 162: Are we going to have this flag permanently or just for short term?

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/ClientTokenCloudConnectionTest.cs
```diff
@@ -439,7 +439,8 @@ IClient GetMockedDeviceClient()
                 TimeSpan.FromSeconds(20),
                 false,
                 Option.None<IWebProxy>(),
-                productInfoStore);
+                productInfoStore,
+                true);
```
philipktlin, line 443: we should use true only for new tests recently added for nested edge scenario. right?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
richma-ms: This check only applies in nested edge scenarios, previously it assumed nested edge is always enabled.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
richma-ms: short term until all the dependent pieces are released (iothub / sdk).

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Test/ClientTokenCloudConnectionTest.cs
richma-ms: the end goal is to only have the nested edge logic in product, at which point all the tests need to run in this configuration.  I've already previously updated these tests to mock out the new nested edge behavior, so might as well have them run through nested edge codepath now.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
richma-ms: for the next phase (updating identities connection-failure), we'll need to check the authchain here anyways.  So pushing that logic lower into the static helpers will actually make it more awkward.

The empty string path is strictly temporary until our service side and SDK dependencies come in, at which point this flag wouldn't be needed and we won't have pass in an empty string anymore.

PR #3132, opened 2020-06-23
[TEST] Broker / EdgeHub Packaging and Build Infra

and-rewsmith:
The MQTT Broker is designed to run in the same container as EdgeHub Core. Currently, we do not have the packaging logic or build infrastructure to do this.  

This PR contains the following:
- Packaging logic for edge hub and the mqtt broker to run inside the same container.We made our own watchdog for these two processes (edgehub and broker) implemented in rust.
- New build logic for the mqtt broker which will cross compile and lock down necessary dependencies (rust toolchain, openssl, cmake) on a per-platform basis. This is modeled after the edgelet build process defined in package.sh.
- Pipeline integration that will build and publish combined edgehub/broker images for each target platform.

**Packaging logic**
I created a watchdog rust program that will launch two child processes for edgehub and broker. This is the big picture:
- Both processes run together in the same container 
- If either of these processes stop, the other is shut down gracefully and the container exits 
- If the docker container is stopped, both processes shut down gracefully 
- Handle processes appropriately (i.e. prevention of zombie processes) 
- Redirect output of both processes to stdout 

Since the watchdog is a component running inside edgehub, I have refactored the directory structure of edgehub to contain subfolders: (docker, core, watchdog). This change bloated the size of the PR as I had to change many .cs and .csproj files.

**New Build Logic**
We needed to come up with a new build process for the broker and the watchdog. The current approach did not provide flexibility in selecting rust toolchain, openssl, cmake versions. 

We modeled this new build logic after edgelet. See edgelet's package.sh for context on how edgelet cross compiles to lock down dependencies. This can mostly be replicated, but since there was no existing logic for alpine, we had to come up with something new. 

Regarding alpine, we don't need to build directly on alpine. Building on bionic will also work as long as it produces a musl binary. In order to build musl artifacts on bionic, I ripped out the relevant parts of [this dockerfile](https://github.com/emk/rust-musl-builder/blob/master/Dockerfile#L1) (line 38-70, 110-126, 145-155), which then get copied into the alpine image.

All the other build logic has precedent in edgelet.

**Pipeline Integration**
Since we decided not to build windows images for the broker, I have removed windows from the build and manifest steps. I then used the new build logic to build the watchdog/broker artifacts, copied them into the build context (edgehub's dotnet publish folder), then set the watchdog as the image entry point.

This change will significantly increase the build time of our build images pipeline (60+ mins). We can slightly mitigate this by reducing the 6 rust build steps to 3 (consolidate broker and watchdog cross compilations), but the cross compilation of the broker is the most costly. The best way to mitigate this is to have a base image that gets regularly uploaded to our build container registry, but then we need to deal with maintaining compatibility for all feature branches. We can handle this outside this PR.

**Thoughts and lingering things**
- Our current edge hub image build approach is to copy all the contents of the edge hub publish folder into the image. The broker and watchdog artifacts are now inside this publish folder (to reduce build context size). There is currently precedent for this behavior (i.e. the dockerfile folder), however I think this is a dangerous habit because it can lead to undesirable files bloating the container size.  This would be acceptable imo if we could provide some sort of exclude list, however that is not supported. [Relevant issue](https://github.com/moby/moby/issues/15771). I tried using a rm -rf command to remove the watchdog, mqtt, and docker folders but since the build agent is amd64, it can't run this command on the arm base images. We need to either find a way around this issue or concede to a larger build context containing all the published dotnet apps.
- We still have windows docker files checked into the repo. I think it is best to remove them, but was hesitant due to the already large size of the PR.
- Need to handle build images release. We will end up replicating the logic in our build pipeline, but how are we going to handle the code sign? I think this might be fine since there is no signing of executables on linux but need to double check. 
- I am still doing manual testing on the broker, but there is likely some configuration that I could be missing. Let me know if you see something wrong.


edge-hub/docker/linux/arm32v7/Dockerfile
```diff
@@ -1,13 +1,21 @@
 ARG base_tag=1.0.5-linux-arm32v7
 FROM azureiotedge/azureiotedge-hub-base:${base_tag}
 
+ADD ./watchdog/arm32/watchdog /usr/local/bin/watchdog
+ADD ./mqtt/arm32/mqttd /usr/local/bin/mqttd
```
and-rewsmith, line 5: I think we should strip these executables

builds/misc/templates/copy-broker-watchdog-artifacts.yaml
```diff
@@ -0,0 +1,20 @@
+# TODO: var for base paths
+# TODO: strip in dockerfile
+
+# This template will move the mqtt and watchdog artifacts into the EdgeHub dotnet publish folder
+# EdgeHub's dotnet target folder will serve as a minimal build context for the image
+steps:
+  - bash: |
+      mkdir -p $(Build.BinariesDirectory)/publish/Microsoft.Azure.Devices.Edge.Hub.Service/watchdog/amd64
```
and-rewsmith, line 8: nit: We can extract the path publish/Microsft...Service to a var

edge-hub/docker/linux/amd64/Dockerfile
```diff
@@ -18,18 +21,25 @@ RUN adduser -Ds /bin/sh -u ${EDGEHUBUSER_ID} edgehubuser
...
 EXPOSE 5671/tcp	
 EXPOSE 443/tcp	
 
 USER edgehubuser	
 
-CMD echo "$(date --utc +"%Y-%m-%d %H:%M:%S %:z") Starting Edge Hub" && \	
-    exec /usr/bin/dotnet Microsoft.Azure.Devices.Edge.Hub.Service.dll
+# TODO: fix broker state persistence
+CMD echo "$(date --utc +"%Y-%m-%d %H:%M:%S %:z") Starting Edge Hub" && \
+    /usr/local/bin/watchdog
```
and-rewsmith, line 45: The broker needs to be configured to dump its state where it has permission to do so.

edge-hub/watchdog/linux/Makefile
```diff
@@ -0,0 +1,14 @@
+TARGET=target/release
+CARGOFLAGS=--manifest-path=$(SRCDIR)/Cargo.toml
+CARGO=cargo
+SRCDIR?=.
+
+# TODO: should we have a make target that doesn't build release?
+all:
+	$(CARGO) build $(CARGOFLAGS)
```
and-rewsmith, line 8: Should we remove this? It might be safer to only build release since the only place this makefile is used is in the build pipeline.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+use std::process::{Command, Stdio, Child};
+use std::sync::{Arc};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::Duration;
+use nix::unistd::Pid;
+use nix::sys::signal::{self, Signal};
+
+// TODO: should we set up logging more formally with dependency
+// TODO: if not, then we can add info / error
+const LOG_HEADER: &'static str = "[WATCHDOG]:";
```
and-rewsmith, line 15: Should we use a dependency for logging? If not, it would be helpful to add INFO, ERROR labels.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+    let mut broker = Command::new("/usr/local/bin/mqttd")
+            .stdout(Stdio::inherit())
+            .spawn()
+            .expect("failed to execute MQTT broker process");
+   
+    // TODO: confirm desirable to wait for edgehub and broker processes to be created first
+    let signals = Signals::new(&[SIGTERM])?;
+    let has_received_sigterm = Arc::new(AtomicBool::new(true));
+    let sigterm_listener = has_received_sigterm.clone();
+    thread::spawn(move || {
```
and-rewsmith, line 35: We should probably start this thread at the very beginning so no sent SIGTERMS are ignored.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+        for _sig in signals.forever() {
+            println!("{:?} Received SIGTERM for watchdog", LOG_HEADER);
+            sigterm_listener.store(false, Ordering::Relaxed); // TODO: figure out best Ordering
```
and-rewsmith, line 38: We should pick the correct ordering here.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    println!("{:?} launched Edge Hub process with pid {:?}", LOG_HEADER, edgehub.id());
+    println!("{:?} launched MQTT Broker process with pid {:?}", LOG_HEADER, broker.id());
+    let mut is_edgehub_running = is_child_process_running(&mut edgehub);
+    let mut  is_broker_running = is_child_process_running(&mut broker);
+    while has_received_sigterm.load(Ordering::Relaxed) && is_edgehub_running && is_broker_running {
+        is_edgehub_running = is_child_process_running(&mut edgehub);
+        is_broker_running = is_child_process_running(&mut broker);
+
+        thread::sleep(Duration::from_secs(4)); // TODO: configure wait time for poll
```
and-rewsmith, line 50: We need to determine how often the watchdog should poll the processes. Any ideas?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+    if is_edgehub_running {
+        println!("{:?} Sending SIGTERM to Edge Hub", LOG_HEADER);
+        signal::kill(Pid::from_raw(edgehub.id() as i32), Signal::SIGTERM).unwrap();
+    }
+    if is_broker_running {
+        println!("{:?} Sending SIGTERM to MQTT Broker", LOG_HEADER);
+        signal::kill(Pid::from_raw(broker.id() as i32), Signal::SIGTERM).unwrap();
+    }
+
+    thread::sleep(Duration::from_secs(10)); // TODO: configure wait time
```
and-rewsmith, line 78: We need to determine how long we will give EdgeHub and Broker to both shut down cleanly before issuing SIGKILL. Any ideas?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    shutdown(&mut edgehub, &mut broker);
+
+    println!("Exiting");
+    Ok(())
+}
+
+fn is_child_process_running(child_process: &mut Child) -> bool
+{
+    return !child_process.try_wait().unwrap().is_some();
```
and-rewsmith, line 61: Need to determine how to handle unwrap if fails.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+fn shutdown(mut edgehub: &mut Child, mut broker: &mut Child) {
+    println!("{:?} Initiating shutdown of MQTT Broker and Edge Hub", LOG_HEADER);
+
+    let mut is_edgehub_running = is_child_process_running(&mut edgehub);
+    let mut is_broker_running = is_child_process_running(&mut broker);
+    if is_edgehub_running {
+        println!("{:?} Sending SIGTERM to Edge Hub", LOG_HEADER);
+        signal::kill(Pid::from_raw(edgehub.id() as i32), Signal::SIGTERM).unwrap();
```
and-rewsmith, line 71: Need to determine how to handle unwrap if fails.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    thread::sleep(Duration::from_secs(10)); // TODO: configure wait time
+
+    // TODO: we need to unwrap signal::kill at least to verify if end sigkill succeeded. How do we do this guaranteeing no panic from already killed process (i.e. from earlier slow sigterm)
+    is_edgehub_running = is_child_process_running(&mut edgehub);
+    is_broker_running = is_child_process_running(&mut broker);
+    if is_edgehub_running {
+        println!("Killing Edge Hub");
+        edgehub.kill().unwrap();
```
and-rewsmith, line 85: Need to determine how to handle unwrap if fails.

scripts/linux/cross-platform-rust-build.sh
```diff
@@ -0,0 +1,315 @@
...
+        # Here are the places to check for new releases:
+        #
+        # - https://github.com/rust-lang/mdBook/releases
+        # - https://github.com/EmbarkStudios/cargo-about/releases
+        # - https://github.com/EmbarkStudios/cargo-deny/releases
+        SETUP_COMMAND=$'
+            OPENSSL_VERSION=1.1.1g
+            MDBOOK_VERSION=0.3.7
+            CARGO_ABOUT_VERSION=0.2.2
+            CARGO_DENY_VERSION=0.6.7
```
and-rewsmith, line 120: We need to find a way to handle these important dependencies since they will get updated. Any ideas? We can elevate to the top of file for clarity, but that will separate it from the alpine logic.

edge-hub/docker/linux/amd64/Dockerfile
```diff
@@ -18,18 +21,23 @@ RUN adduser -Ds /bin/sh -u ${EDGEHUBUSER_ID} edgehubuser
...
 
 # Install RocksDB	
-COPY --from=builder publish/* /usr/local/lib/	
+COPY --from=hub-builder publish/* /usr/local/lib/	
 
 WORKDIR /app	
 
-COPY $EXE_DIR/ ./	
+# Copy edge hub artifacts and remove unnecessary items, leaving just edge hub core dlls
+COPY $EXE_DIR/ ./
```
and-rewsmith, line 29: We need to remove docker files, mqtt binary, watchdog binary. I tried adding a rm step, but since the builder agent is amd64 this won't work on arm. Any ideas?

builds/misc/templates/copy-broker-watchdog-artifacts.yaml
```diff
@@ -0,0 +1,20 @@
...
+      mkdir -p $(Build.BinariesDirectory)/$(edgehub.build.path)/mqtt/amd64
+      mkdir -p $(Build.BinariesDirectory)/$(edgehub.build.path)/mqtt/arm32
+      mkdir -p $(Build.BinariesDirectory)/$(edgehub.build.path)/mqtt/arm64
+      cp edge-hub/watchdog/linux/target/x86_64-unknown-linux-musl/release/watchdog $(Build.BinariesDirectory)/$(edgehub.build.path)/watchdog/amd64
+      cp edge-hub/watchdog/linux/target/armv7-unknown-linux-gnueabihf/release/watchdog $(Build.BinariesDirectory)/$(edgehub.build.path)/watchdog/arm32
+      cp edge-hub/watchdog/linux/target/aarch64-unknown-linux-gnu/release/watchdog $(Build.BinariesDirectory)/$(edgehub.build.path)/watchdog/arm64
+      cp mqtt/target/x86_64-unknown-linux-musl/release/mqttd $(Build.BinariesDirectory)/$(edgehub.build.path)/mqtt/amd64
+      cp mqtt/target/armv7-unknown-linux-gnueabihf/release/mqttd $(Build.BinariesDirectory)/$(edgehub.build.path)/mqtt/arm32
+      cp mqtt/target/aarch64-unknown-linux-gnu/release/mqttd $(Build.BinariesDirectory)/$(edgehub.build.path)/mqtt/arm64
+    displayName: Copy watchdog and broker artifacts to dockerfile build context
```
and-rewsmith, line 20: nit: we could clean this up further

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: Also, it is logging the quotes for some reason.

scripts/linux/cross-platform-rust-build.sh
```diff
@@ -0,0 +1,295 @@
+#!/bin/bash
```
and-rewsmith, line 1: I don't think we should consolidate this with the script in edgelet because there is a lot of edgelet specific logic.

and-rewsmith:
I'm opening 3 smaller PRs so please don't review this.

PR #3133, opened 2020-06-23
Introduce CentOs7 to the new E2E test framework

yophilav:
Introduce CentOs7 to the new E2E test framework
The tests suit includes: 
- QuickstartCerts()
- TempSensor()
- TempFilter()
- ModuleToModuleDirectMethod(Amqp)
- ModuleToModuleDirectMethod(AmqpWs)
- ModuleToModuleDirectMethod(Mqtt)
- ModuleToModuleDirectMethod(MqttWs)
- DpsSymmetricKey()
	
Work in-progress tests (Upcoming PRs):
- TempFilterFunc()
- PriorityQueueModuleToHubMessages()
- PriorityQueueModuleToModuleMessages()
- PriorityQueueTimeToLive()
- ValidateMetrics()


test/Microsoft.Azure.Devices.Edge.Test.Common/linux/OsPlatform.cs
```diff
@@ -24,7 +24,7 @@ public async Task<string> CollectDaemonLogsAsync(DateTime testStartTime, string
             return daemonLog;
         }
 
-        public IEdgeDaemon CreateEdgeDaemon(Option<string> _, Option<string> bootstrapAgentImage, Option<Registry> bootstrapRegistry) => new EdgeDaemon(bootstrapAgentImage, bootstrapRegistry);
+        public IEdgeDaemon CreateEdgeDaemon(Option<string> _, Option<string> bootstrapAgentImage, Option<Registry> bootstrapRegistry) => EdgeDaemon.CreateAsync(bootstrapAgentImage, bootstrapRegistry).Result;
```
damonbarry, line 27: This needs to become `CreateEdgeDaemonAsync`, instead of using `Task.Result`.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Test.Common.Linux
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public class PackageManagement
```
damonbarry, line 8: This class is a great solution for handling linux variations in `EdgeDaemon`. But rather than exposing several public properties it should expose methods that use the (private) properties: `InstallAsync` and `UninstallAsync`. And I guess you would still need one property for the list of services that gets passed to start/stop.

This will make `EdgeDaemon.cs` a lot cleaner.

builds/e2e/templates/e2e-run.yaml
```diff
@@ -15,7 +15,15 @@ steps:
     }
     else
     {
-      sudo --preserve-env dotnet vstest $testFile --logger:trx
+      # Unfortunately CentOs has some failing test that need to be worked on.
```
damonbarry, line 18: This works, but we already have a pattern for disabling tests. Search for tests that call `Assert.Ignore("...")`, e.g. in `PriorityQueue.cs`. I'd prefer that you not introduce another pattern.

builds/e2e/e2e.yaml
```diff
@@ -131,3 +131,29 @@ jobs:
...
+      artifactName: iotedged-centos7-amd64
+
+    steps:
+    - task: CmdLine@2
+      inputs:
+        script: 'sudo find . -type f -group root -execdir rm {} + && sudo find . -type d -group root -print0 | sudo xargs -0 /bin/rm -rf' 
+      displayName: Clean up files (as sudo)
+
+    - template: templates/e2e-setup.yaml
+    - template: templates/e2e-run.yaml
```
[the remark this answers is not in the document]
yophilav: Got it.

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -86,6 +86,11 @@ steps:
       '$(System.ArtifactsDirectory)/$(az.pipeline.images.artifacts)/artifactInfo.txt'
     $imageId = ($imageId -split '=')[1]
     $imageTag = "$imageId-$(os)-$(arch)"
+
+    echo "edgeAgentImage = $imagePrefix-agent:$imageTag";
+    echo "edgeHubImage = $imagePrefix-hub:$imageTag";
+    echo "edgeAgentBootstrapImage = $imagePrefix-agent-bootstrap-e2e-$(os)-$(arch)";
```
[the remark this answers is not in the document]
yophilav: Removed.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
yophilav: Thank you for the advice! I love the idea. I'm working on it.

builds/e2e/templates/e2e-run.yaml
yophilav: I see what you mean. Though it looks like there is no easy way for a test to detect if is running on CentOs from within the dotnet. I'm thinking to use the category as a work around to disable test until I bring all the test up on CentOs. 
And for the sake of consistency, I could convert the aforementioned PriorityQueue Tests into using Category-blacklist. Would that work for you?

damonbarry: Just keep what you have here. Don't convert any of the existing tests.

With `Assert.Ignore`, disabled tests show as "skipped", which is a nice, visible reminder, so I'd like to keep that.

damonbarry: I suspect something could be added to `OsPlatform`--we already have methods there like `IsWindows()` and `Is64Bit()` so it seems like a natural fit. But if you need to call `lsb_release` or something similar, then you have to handle async, etc...not worth the effort IMO. The `CentOsSafe` category is temporary, so no biggie.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -0,0 +1,187 @@
...
+            this.iotedgeServices = null;
+            this.uninstallCmd = null;
+            this.isInitialized = false;
+        }
+
+        async Task CheckInitializationAsync()
+        {
+            if (!this.isInitialized)
+            {
+                var tokenSource = new CancellationTokenSource();
```
damonbarry, line 47: Don't create a new token. These new `*Async` methods should take a `CancellationToken` argument, and should use the token that's passed in. That way if the test setup is taking too long it will time out.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -0,0 +1,187 @@
...
+                string[] platformInfo = await Process.RunAsync("lsb_release", "-sir", tokenSource.Token);
+                this.InitializeParameters(platformInfo);
+                this.isInitialized = true;
+            }
+        }
+
+        void InitializeParameters(string[] platformInfo)
+        {
+            if (platformInfo.Length == 1)
```
damonbarry, line 56: Is `platformInfo` sometimes not 1 line?

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -0,0 +1,187 @@
...
+                            $"{this.packageTool} updateinfo",
+                            $"{this.packageTool} install --yes iotedge",
+                            "pathToSystemdConfig=$(systemctl cat iotedge | head -n 1); sed 's/=on-failure/=no/g' ${pathToSystemdConfig#?} > ~/override.conf; sudo mv -f ~/override.conf ${pathToSystemdConfig#?}; sudo systemctl daemon-reload;"
+                        };
+                default:
+                    throw new NotImplementedException($"Don't know how to install daemon on for '.{this.packageExtension}'");
+            }
+        }
+
+        public async Task InternalStopAsync(CancellationToken token)
```
damonbarry, line 172: This should be called `StopAsync` since it's a public method.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -0,0 +1,187 @@
...
+                        throw new NotImplementedException($"Don't know how to install daemon on operating system '{this.os} {this.version}'");
+                    }
+
+                    break;
+                default:
+                    throw new NotImplementedException($"Don't know how to install daemon on operating system '{os}'");
+            }
+        }
+
+        public async Task<string[]> GetInstallCommandsFromLocalAsync(string path)
```
damonbarry, line 107: There are four public methods in this class. Two return an array of command strings that the caller must pass to `Process.RunAsync`, and the other two run the commands directly. I think you should choose one approach or the other for all four methods.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
yophilav: nope, unfortunately. But they are returning in the same format though.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
yophilav: done

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
yophilav: yup. consolidating

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/LinuxStandardBase.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Test.Common.Linux
+{
+    using Microsoft.Azure.Devices.Edge.Util;
+    using SupportedPackageExtension = PackageManagement.SupportedPackageExtension;
```
damonbarry, line 5: Nit: The higher level class (`PackageManagement`) should depend on the lower level class (`LinuxStandardBase`), not the other way around.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/OsPlatform.cs
yophilav: This got refactored into using lazy. :D

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -0,0 +1,285 @@
...
+                            break;
+                        default:
+                            throw new NotImplementedException($"Don't know how to install daemon on operating system '{os}'");
+                    }
+
+                    return new LinuxStandardBase(os, version, packageExtension);
+                });
+
+        string forceInstallConfigCmd = null;
+        public string ForceInstallConfigCmd
```
damonbarry, line 60: Most of these properties are only used internally, so they shouldn't be public.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
yophilav: With the lazy<> lsbTask, passing in a token can be tricky. The currently solution is to have a hard-coded timeout of 10 seconds. (for ARM32).

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -0,0 +1,285 @@
...
+                            throw new NotImplementedException($"Undefined forceInstallConfigCmd for '{lsb.PackageExtension}'");
+                    }
+                }
+
+                return this.forceInstallConfigCmd;
+            }
+        }
+
+        string installCmd = null;
+        public string InstallCmd
```
damonbarry, line 85: There are 2 problems with these properties:
1. Properties can't be async so you're using `.Result` instead, which is what we were trying to avoid (mixing sync/async patterns).
2. Each property is a lot of overhead, considering what it does. It has to await the LSB result, then switch on the package extension to produce a string. But most of these properties are used in `GetInstallCommandsFromLocal` or `GetInstallCommandsFromMicrosoftProd`, which also await the LSB result and then switches on the package extension to produce a string. Why not get rid of these properties and insert the strings directly into the `GetInstallCommandsFrom*` methods?

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
```diff
@@ -35,47 +35,14 @@ public async Task InstallAsync(Option<string> packagesPath, Option<Uri> proxy, C
                     properties = properties.Append(p).ToArray();
                 });
 
-            string[] commands = await packagesPath.Match(
-                p =>
+            string[] commands = packagesPath.Match(
+                (p) =>
```
damonbarry, line 39: Nit: you only need parenthesis if you don't have any arguments

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -0,0 +1,285 @@
...
+    using Serilog;
+
+    public class PackageManagement
+    {
+        Lazy<Task<LinuxStandardBase>> lsbTask =
+            new Lazy<Task<LinuxStandardBase>>(
+                async () =>
+                {
+                    var cts = new CancellationTokenSource();
+                    cts.CancelAfter(10000);
```
damonbarry, line 19: We need to use the token generated in `SetupFixture`. Which means it needs to be passed to the constructor. Then you'll initialize `lsbTask` in the constructor instead of here.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
```diff
@@ -153,8 +120,7 @@ public async Task UninstallAsync(CancellationToken token)
                     async () =>
                     {
                         string[] output =
-                            await Process.RunAsync("apt-get", "purge --yes libiothsm-std iotedge", token);
-
+                            await Process.RunAsync($"{this.packageManagement.PackageTool}", $"{this.packageManagement.UninstallCmd} libiothsm-std iotedge", token);
```
damonbarry, line 123: Given my other comment about the problems with these `PackageManagement` properties, I'd create a `PackageManagement.GetUninstallCommandsAsync` method instead. Also, a `PackageManagement.GetIotedgeServicesAsync` method.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
damonbarry: See my latest comments. We *need* to honor the cancellation tokens created in the fixtures.

smoke/IotEdgeQuickstart/details/Process.cs
```diff
@@ -7,6 +7,7 @@ namespace IotEdgeQuickstart.Details
     using System.Threading;
     using System.Threading.Tasks;
     using RunProcessAsTask;
+    using Serilog;
```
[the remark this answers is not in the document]
yophilav: Yup, you're absolutely correct.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/LinuxStandardBase.cs
yophilav: Ok.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
yophilav: got it

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
yophilav: Sure!

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
yophilav: How about a property that has a private set; but public get;

damonbarry: A property can't be async so you'd still have the problem of mixing sync and async code.

yophilav: Yeah.. I revisit my approach there.
For the properties that required outside of the class, I create a getter for them.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
yophilav: there there.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
yophilav: private variable they are!

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
yophilav: Yup. I got that figured out now. Sorry about that.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -0,0 +1,100 @@
...
+            SupportedPackageExtension.Deb => new[]
+            {
+                $"curl https://packages.microsoft.com/config/{this.os}/{this.version}/multiarch/prod.list > /etc/apt/sources.list.d/microsoft-prod.list",
+                "curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > /etc/apt/trusted.gpg.d/microsoft.gpg",
+                $"apt-get update",
+                $"apt-get install --yes iotedge"
+            },
+            // Based on instructions at:
+            // https://github.com/MicrosoftDocs/azure-docs/blob/058084949656b7df518b64bfc5728402c730536a/articles/iot-edge/how-to-install-iot-edge-linux.md
+            // TODO: 8/30/2019 support curl behind a proxy
```
damonbarry, line 73: Whoops...can you move these comments up to the Deb-arm of the switch? That's where they belong...

PR #3136, opened 2020-06-24
Add registry api proxy in EdgeHub

philipktlin:
Add registry api proxy in EdgeHub, which is calling from iotedged on child device.  Currently it supports Get, Post, Put, Delete of registry API endpoint (e.g. /deivces/{deviceId}/modules/{moduleId}).  Add unit test and manually tested with postman.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/RegistryProxyMiddleware.cs
```diff
@@ -0,0 +1,119 @@
...
+            HttpRequest request = context.Request;
+
+            // This pattern is used to match registry API for create, update, get, delete of a module, and list modules.
+            Regex registryApiUriPattern = new Regex(@"\/devices\/.+\/modules(\/.+)?$");
+
+            if (context.WebSockets.IsWebSocketRequest ||
+                !request.IsHttps ||
+                !registryApiUriPattern.IsMatch(request.Path))
+            {
+                return null;
```
varunpuranik, line 111: Don't use null pls.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/RegistryProxyMiddleware.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            await responseMessage.Content.CopyToAsync(response.Body);
+        }
+
+        internal Uri BuildDestinationUri(HttpContext context)
+        {
+            HttpRequest request = context.Request;
+
+            // This pattern is used to match registry API for create, update, get, delete of a module, and list modules.
+            Regex registryApiUriPattern = new Regex(@"\/devices\/.+\/modules(\/.+)?$");
```
varunpuranik, line 105: What is the pattern here? 
just /devices/deviceId/modules/moduleId ?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Startup.cs
```diff
@@ -93,6 +81,28 @@ public void Configure(IApplicationBuilder app)
             });
         }
 
+        private (string iotHubHostname, string edgeDeviceId, string gatewayHostname) GetStartupParameters()
```
varunpuranik, line 84: Remove private.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/RegistryProxyMiddleware.cs
philipktlin: You can find all references of iotedged calling to registry api in this file, https://github.com/Azure/iotedge/blob/4425533aed41cd3eed4780612af214a075b51093/edgelet/iothubservice/src/device.rs#L91.  This pattern is for GET, PUT, DELETE of /devices/{deviceId}/modules/{moduleId} and GET of /devices/{deviceId}/modules.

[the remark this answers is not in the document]
philipktlin: can't make it as static, it use class field in line 115.  Or you want me to move this class field as method argument and make it static.  Btw asp.net will use single instance of middleware to process multiple request, do we still have benefit to make it static?  Let me know if you still prefer to use static here.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/RegistryProxyMiddleware.cs
philipktlin: Option is used.

philipktlin:
> Changes look okay. But how will it work once we want to support on-behalf of authentication from EdgeHub.

Let's chat about this offline.  I would like to understand the details.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/RegistryProxyMiddleware.cs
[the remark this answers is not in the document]
varunpuranik: I am talking about the Regex.. that can be static right?

PR #3139, opened 2020-06-24
Refactor edge hub folder structure to move source code under "core/"

and-rewsmith:
The edge hub directory structure we are eventually going for is the following:
- core
- mqtt bridge
- authentication
- authentication wrapper

EdgeHub will then ref the mqtt broker and the policy engine as separate crates. I am making this change now (only for core) for two reasons:
- We will eventually need to do this, and I already did the work to do so (which is pretty tedious)
- It made sense considering that the watchdog is now in the edgehub folder


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Mqtt.Test/DeviceIdentityProviderTest.cs
```diff
@@ -95,7 +95,6 @@ public static IEnumerable<object[]> GetIdentityProviderInputs()
             };
         }
 
-
```
and-rewsmith, line 98: Fix stylecop

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/DeviceIdentityProvider.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Mqtt
     using static System.FormattableString;
 
     public class DeviceIdentityProvider : IDeviceIdentityProvider
-    {
```
and-rewsmith, line 17: Fix stylecop

PR #3140, opened 2020-06-24
Add watchdog process that will oversee child processes for edgehub and broker

and-rewsmith:
I created a watchdog rust program that will launch two child processes for edgehub and broker. This is the big picture:
- Both processes run together in the same container
- If either of these processes stop, the other is shut down gracefully and the container exits
- If the docker container is stopped, both processes shut down gracefully
- Handle processes appropriately (i.e. prevention of zombie processes)
- Redirect output of both processes to stdout


edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+        .with_max_level(tracing::Level::INFO)
+        .finish();
+    let _ = tracing::subscriber::set_global_default(subscriber);
+
+    info!("Starting watchdog");
+}
+
+fn is_child_process_running(child_process: &mut Child) -> bool
+{
+    return !child_process.try_wait().unwrap().is_some();
```
and-rewsmith, line 66: We should handle this unwrap here and elsewhere.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::sync::{Arc};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::Duration;
+use nix::unistd::Pid;
+use nix::sys::signal::{self, Signal};
+use signal_hook::{iterator::Signals, SIGTERM};
+use tracing::{info};
+use tracing_subscriber;
+
+fn main() -> Result<(), Box<dyn Error>> {
```
and-rewsmith, line 14: I think this Box<dyn Error> is necessitated by Signals, but need to double check.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+    init_logging();
+
+    let signals = Signals::new(&[SIGTERM])?;
+    let has_received_sigterm = Arc::new(AtomicBool::new(true));
+    let sigterm_listener = has_received_sigterm.clone();
+    thread::spawn(move || {
+        for _sig in signals.forever() {
+            info!("Received SIGTERM for watchdog");
+            sigterm_listener.store(false, Ordering::Relaxed);
```
and-rewsmith, line 23: Relaxed is fine because there is no ordering constraints.
https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Relaxed

edge-hub/watchdog/linux/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "watchdog"
+version = "0.1.0"
```
and-rewsmith, line 3: I replicated the pattern in the broker here.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+    shutdown(&mut edgehub, &mut broker);
+
+    info!("Exiting");
+    Ok(())
+}
+
+fn init_logging() {
+    let subscriber = tracing_subscriber::fmt::Subscriber::builder()
+        .with_max_level(tracing::Level::INFO)
+        .finish();
```
and-rewsmith, line 51: We do this differently in the broker, but I took this sample from the docs. Let me know if you recommend changes.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+fn wait_for_sigterm_or_processes_to_shutdown(mut edgehub: &mut Child, mut broker: &mut Child, has_received_sigterm: Arc<std::sync::atomic::AtomicBool>) {
+    let mut is_edgehub_running = is_child_process_running(&mut edgehub);
+    let mut is_broker_running = is_child_process_running(&mut broker);
+    while has_received_sigterm.load(Ordering::Relaxed) && is_edgehub_running && is_broker_running {
+        is_edgehub_running = is_child_process_running(&mut edgehub);
+        is_broker_running = is_child_process_running(&mut broker);
+
+        thread::sleep(Duration::from_secs(4));
```
and-rewsmith, line 78: We need to configure this time to wait until checking the child processes again.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+    if is_edgehub_running {
+        info!("Sending SIGTERM to Edge Hub");
+        sigterm_child_process(&mut edgehub);
+    }
+    if is_broker_running {
+        info!("Sending SIGTERM to MQTT Broker");
+        sigterm_child_process(&mut broker);
+    }
+
+    thread::sleep(Duration::from_secs(10));
```
and-rewsmith, line 101: We need to pick an appropriate duration of time to give the child processes to gracefully shut down. It definitely should not be 10 right now. For a reference point, I think the default right now for iotedged to shut down edge agent is 60.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    };
+
+    let mut edgehub = Command::new("dotnet")
+            .arg("/app/Microsoft.Azure.Devices.Edge.Hub.Service.dll")
+            .stdout(Stdio::inherit())
+            .spawn()
+            .expect("Failed to execute Edge Hub process");
+    
+    let mut broker = Command::new("/usr/local/bin/mqttd")
```
and-rewsmith, line 32: Not directly related to this PR, but are we should we instead put this binary in /app similar to the Edge Hub dll?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    });
+
+    Ok(has_received_sigterm)
+}
+
+fn wait_for_sigterm_or_processes_to_shutdown(mut edgehub: &mut Child, mut broker: &mut Child, has_received_sigterm: Arc<std::sync::atomic::AtomicBool>) {
+    let mut is_edgehub_running = is_child_process_running(&mut edgehub);
+    let mut is_broker_running = is_child_process_running(&mut broker);
+    while has_received_sigterm.load(Ordering::Relaxed) && is_edgehub_running && is_broker_running {
```
and-rewsmith, line 74: Ordering::Relaxed is the appropriate ordering since we only care about atomic operations.
https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Relaxed

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+use std::io::Error;
+use std::process::{Command, Stdio, Child};
+use std::sync::{Arc};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::Duration;
+use nix::unistd::Pid;
+use nix::sys::signal::{self, Signal};
+use signal_hook::{iterator::Signals, SIGTERM};
+use tracing::{info, error};
+use tracing_subscriber;
```
and-rewsmith, line 13: Let me know if I should order these in a specific way.

vadim-kovalyov: thanks for asking this. Usually we split into 4 groups divided by a new line. 
- std::
- 3rd party crates (like nix, tracing_
- 1st party crates (like mqtt_broker)
- local (everything crate::...)

Also, it is not needed (anymore) to declare `extern crate`

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+fn init_logging() {
+    let subscriber = tracing_subscriber::fmt::Subscriber::builder()
+        .with_max_level(tracing::Level::INFO)
+        .finish();
+    let _ = tracing::subscriber::set_global_default(subscriber);
+
+    info!("Starting watchdog");
+}
+
+fn register_sigterm_listener() -> Result<Arc<std::sync::atomic::AtomicBool>, Error> {
```
vadim-kovalyov, line 57:
```suggestion
fn register_sigterm_listener() -> Result<Arc<AtomicBool>, Error> {
```
we usually try not to use FQN. Only for enums we usually do `EnumType::EnumVariant`

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+    let signals = Signals::new(&[SIGTERM])?;
```
vadim-kovalyov, line 58: Should we handle SIGINT as well?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+    if !is_broker_running {
+        info!("Killing MQTT Broker");
+        kill_child_process(broker);
+    }
+}
+
+fn sigterm_child_process(child: &mut Child)
+{
+    match signal::kill(Pid::from_raw(child.id() as i32), Signal::SIGTERM) {
```
vadim-kovalyov, line 117: you could use `if let` in such cases.
```rust
    if let Err(e) = signal::kill(Pid::from_raw(child.id() as i32), Signal::SIGTERM) {
        error!("Failed to send SIGTERM signal to child process. {:?}", e);
    }
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+use nix::unistd::Pid;
+use nix::sys::signal::{self, Signal};
+use signal_hook::{iterator::Signals, SIGTERM};
+use tracing::{info, error};
+use tracing_subscriber;
+
+fn main() -> Result<(), Error> {
+    init_logging();
+
+    let has_received_sigterm = match register_sigterm_listener() {
```
vadim-kovalyov, line 18: nit: I would call it `shutdown_signal` and make it false by default. Otherwise it's a bit confusing.

edge-hub/watchdog/linux/src/main.rs
vadim-kovalyov: Sure, why not. I assume it would be just Dockerfile/script change?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        Ok(has_received_sigterm) => has_received_sigterm,
+        Err(e) => {
+            error!("Failed to register sigterm listener. Shutting down.");
+            return Err(e)
+        }
+    };
+
+    let mut edgehub = Command::new("dotnet")
+            .arg("/app/Microsoft.Azure.Devices.Edge.Hub.Service.dll")
+            .stdout(Stdio::inherit())
```
vadim-kovalyov, line 28: How would combined logs look like? Would that be readable?

edge-hub/watchdog/linux/src/main.rs
vadim-kovalyov: Also, use `cargo fmt` and `cargo clippy` - they will be helpful.

edge-hub/watchdog/linux/src/main.rs
vadim-kovalyov: How is CPU consumption if you set it to 1sec?

edge-hub/watchdog/linux/src/main.rs
vadim-kovalyov: I think it is fine to do what Iotedged does - 60s. After that, the container will be shut down anyway :)

edge-hub/watchdog/linux/Makefile
```diff
@@ -0,0 +1,13 @@
+TARGET=target/release
```
vadim-kovalyov, line 1: just curious why do we need makefile?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+    let mut is_broker_running = is_child_process_running(&mut broker);
+    while has_received_sigterm.load(Ordering::Relaxed) && is_edgehub_running && is_broker_running {
+        is_edgehub_running = is_child_process_running(&mut edgehub);
+        is_broker_running = is_child_process_running(&mut broker);
+
+        thread::sleep(Duration::from_secs(4));
+    }
+}
+
+fn is_child_process_running(child_process: &mut Child) -> bool
```
vadim-kovalyov, line 82: nit: `is_running(child_process: &mut Child)`

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+{
+    return !child_process.try_wait().unwrap().is_some();
```
vadim-kovalyov, line 84: no need for `return`.

edge-hub/watchdog/linux/src/main.rs
vadim-kovalyov: We just need to answer the question: is this is a recoverable error, if we can't get the status of a child process? should we handle that and try to gracefully shutdown the other process? or should we just blow up with `.expect()` and hard-restart the container?

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: oh yes, this always gets me

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: Oh yes it should definitely be false by default. I am guessing I inverted the name but not the behavior.

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: I think a good argument for including SIGINT is that if a container is started in non-detached mode then they can ctrl c to gracefully stop (correct me pls if wrong). Other than that, is there another way SIGINT can be triggered? I think we may as well add it.

and-rewsmith:
@dmolokanov 

Regarding the supervisor, while this option now is more appealing due to the windows situation, what advantages does it give us? The way I see it is that we have a simple use case, we don't expect the use case to change, we can easily handle zombie processes with custom watchdog, we can inject our own logic if needed, and it is more approachable than a supervisor config (imo). The downside is that we could benefit from standard supervisor features in the future if the use case changes. Writing something custom for a more complicated case is certainly not desirable. Do we really expect the use case to change though? 

You talked above about potentially restarting these child processes rather than restarting the container. That seems like something that a supervisor would be better for. I'm also thinking that this could go the other way though. If we pick a supervisor and want to implement something not supported then we are hosed. What are your thoughts?

edge-hub/watchdog/linux/Makefile
and-rewsmith: There is a script in the build pipeline that will cross compile rust artifacts. It is modeled after the build process for edgelet, which calls make in a certain directory. I chose to keep it consistent.

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: I'm totally willing to cede to your judgement here, but I always prefer longer names with more context. Is there a reason why you disagree with this?

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: I'll check

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: yep

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: If this fails, I think we should act as if the child is not running so that everything gets shut down cleanly.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+fn shutdown(mut edgehub: &mut Child, mut broker: &mut Child) {
+    info!("Initiating shutdown of MQTT Broker and Edge Hub");
+
+    let mut is_edgehub_running = is_child_process_running(&mut edgehub);
+    let mut is_broker_running = is_child_process_running(&mut broker);
```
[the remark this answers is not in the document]
and-rewsmith: I thought about moving the logging in, but thought it was nice to be able to log what entity is being targeted for SIGTERM/SIGKILL. The child process doesn't have state about the name of the process. It has the id, but I don't think that's very useful.
https://doc.rust-lang.org/std/process/struct.Child.html

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: Actually, the argument name makes this obvious so I agree. I've changed it.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        info!("Killing Edge Hub");
+        kill_child_process(edgehub);
+    }
+    if !is_broker_running {
+        info!("Killing MQTT Broker");
+        kill_child_process(broker);
+    }
+}
+
+fn sigterm_child_process(child: &mut Child)
```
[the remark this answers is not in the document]
and-rewsmith: Actually, the argument name makes this obvious so I agree. I've changed it.

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: In order to do this I have to make the child process an option type which will also change the function definitions throughout the file. @vadim-kovalyov What are your thoughts here?

and-rewsmith: In the case where one process fails to start, the other might be started and we would still want to gracefully shut it down. My idea for implementing this logic involves converting the edgehub, broker vars to Option<Child>. Do you agree with this?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+            error!("Failed to register sigterm listener. Shutting down.");
+            return Err(e)
+        }
+    };
+
+    let mut edgehub = Command::new("dotnet")
+            .arg("/app/Microsoft.Azure.Devices.Edge.Hub.Service.dll")
+            .stdout(Stdio::inherit())
+            .spawn()
+            .expect("Failed to execute Edge Hub process");
```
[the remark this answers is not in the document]
and-rewsmith: In the case where one process fails to start, the other might be started and we would still want to gracefully shut it down. My idea for implementing this logic involves converting the edgehub, broker vars to Option. Do you agree with this?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+    Ok(())
+}
+
+fn init_logging() {
+    let subscriber = tracing_subscriber::fmt::Subscriber::builder()
+        .with_max_level(tracing::Level::INFO)
+        .finish();
+    let _ = tracing::subscriber::set_global_default(subscriber);
+
+    info!("Starting watchdog");
```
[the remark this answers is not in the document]
and-rewsmith: I wanted to do it as soon as logging was initialized. I can move it though.

edge-hub/watchdog/linux/src/main.rs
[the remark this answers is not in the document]
and-rewsmith: I think I should add a custom type to give the readable name for this process. It would allow for better logs elsewhere too.

edge-hub/watchdog/linux/src/main.rs
vadim-kovalyov: I'm fine with it.

edge-hub/watchdog/linux/src/main.rs
vadim-kovalyov: Yeah, exactly. By looking at the argument names you can easily infer the meaning. Coming from .net world I can totally understand the urge for longer names, I would do the same in C# :) but it feels like it rust people tend to use concise identificators.

edge-hub/watchdog/linux/src/main.rs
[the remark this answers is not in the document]
vadim-kovalyov: I don't mind `.expect` here as long as it produces good logs with error message from `.spawn()`. This is really unrecoverable situation if edgehub fails to start. 

If we don't see an error message from `.spawn()`, then we should change that to a proper `match` or `if let Err(e)`, and log the error.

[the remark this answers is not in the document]
and-rewsmith: Okay I will find out what the logs look like if expect fails.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::thread;
+use std::time::Duration;
+use tracing::{error, info, subscriber, Level};
+use tracing_subscriber::fmt::Subscriber;
+
+fn main() -> Result<(), Error> {
+    block_on(async_main())
+}
```
and-rewsmith, line 17: Is this the standard approach?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+    // start edgehub and blow up if can't start
+    let mut edgehub = Command::new("dotnet")
+        .arg("/app/Microsoft.Azure.Devices.Edge.Hub.Service.dll")
+        .stdout(Stdio::inherit())
+        .spawn()
+        .expect("Failed to start Edge Hub process");
+    info!("Launched Edge Hub process with pid {:?}", edgehub.id());
+
+    // unwrap broker if started, else shutdown edgehub and exit
+    let maybe_broker = match Command::new("/usr/local/bin/mqttd")
```
and-rewsmith, line 40: is this var name ok?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+        Err(e) => {
+            error!("Failed to start MQTT Broker process. {:?}", e);
+            None
+        }
+    };
+    let mut broker = maybe_broker.unwrap_or_else(|| {
+        info!("Broker process not started, so shutting down EdgeHub and exiting");
+        let shutdown = shutdown(&mut edgehub);
+        block_on(shutdown);
+        exit(1);
```
and-rewsmith, line 54: assuming exit code 1 is fine

edge-hub/watchdog/linux/src/main.rs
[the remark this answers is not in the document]
dmolokanov: Yeah, I kinda opposite opinion here. If it is expected failure, code should not panic, but handle error and cause program shutdown with expected exit code. It is basically the same problem in C# - misuse of exception to control program flow. At first glance they are handy, but down the road causes unpredictable behavior.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
+use nix::sys::signal::{self, Signal};
```
dmolokanov, line 3: seems like it is still not sorted

edge-hub/watchdog/linux/src/main.rs
dmolokanov: actually not. let's discuss async stuff offline? It adds complexity without clear benefits

edge-hub/watchdog/linux/src/main.rs
dmolokanov: it's not an Option, so `maybe_` is not very applicable here.
you can do a bit different:
```rust
let broker = Command::new("/usr/local/bin/mqttd")
        .stdout(Stdio::inherit())
        .spawn() ;
match broker 
{
  Ok(broker )  => broker,
  Err(e) => { warn!(...); exit(1); }
}

File: edge-hub/watchdog/linux/src/main.rs (hunk 25 above)
Comment (dmolokanov) [reply edge-hub/watchdog/linux/src/main.rs:54]: let's make a different exit code when edgehub and broker and watchdog exits with error?

File: edge-hub/watchdog/linux/src/main.rs [hunk 27]
```diff
@@ -0,0 +1,131 @@
...
+        for _sig in signals.forever() {
+            info!("Received shutdown signal for watchdog");
+            sigterm_listener.store(true, Ordering::Relaxed);
+        }
+    });
+
+    Ok(should_shutdown)
+}
+
+fn is_running(child_process: &mut Child) -> bool {
```
dmolokanov, line 97: lets make function fallible, other words return an Result<bool, Err>  and propagate error to the caller?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+    match child_process.try_wait() {
+        Ok(status) => status.is_none(),
+        Err(e) => {
+            error!("Error while polling child process. {:?}", e);
+            false
+        }
+    }
+}
+
+async fn shutdown(mut child: &mut Child) {
```
dmolokanov, line 107: I dont see any async code in this function. SHould we remove async than?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        Ok(child) => Some(child),
+        Err(e) => {
+            error!("Failed to start MQTT Broker process. {:?}", e);
+            None
+        }
+    };
+    let mut broker = maybe_broker.unwrap_or_else(|| {
+        info!("Broker process not started, so shutting down EdgeHub and exiting");
+        let shutdown = shutdown(&mut edgehub);
```
dmolokanov, line 52: Do we really need to gracefully stop edgehub if broker failed? can we rely on container restart instead?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+        if !is_running(&mut broker) {
+            break;
+        }
+
+        thread::sleep(Duration::from_secs(1));
+    }
+
+    let shutdown_edgehub = shutdown(&mut edgehub);
+    let shutdown_broker = shutdown(&mut broker);
+    join!(shutdown_edgehub, shutdown_broker);
```
dmolokanov, line 72: async by itself doesn't make function to run in a separate task. right now code execute separately but will block threads and not tasks. more changes required if you want to implement it properly.

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: Ok I think this is good,  but the caller would still behave the same though right?

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: Yes so I brought this up with Vadim. This runs synchronously, but we were considering replacing the call to thread::sleep() with some async thing that yielded back to the caller. Since we are removing async though I can remove.

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: I think we should still gracefully stop edge hub because the amqp protocol heads will still be open and potentially accepting telemetry (and writing it to disk or something else that we would want to gracefully stop)

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: You can explain this when we pair program.

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: Does cargo fmt not sort? I need to check

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+    let subscriber = Subscriber::builder().with_max_level(Level::INFO).finish();
+    let _ = subscriber::set_global_default(subscriber);
+}
+
+// TODO: figure out how to log shutdown signal
+fn register_shutdown_listener() -> Result<Arc<AtomicBool>, Error> {
+    info!("Registering shutdown signal listener");
+    let should_shutdown = Arc::new(AtomicBool::new(false));
+    flag::register(SIGTERM, Arc::clone(&should_shutdown))?;
+    flag::register(SIGINT, Arc::clone(&should_shutdown))?;
```
and-rewsmith, line 183: Using this signature we cannot log when the signal is received and of what type. I am going to look for a way to do this.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        if has_edgehub_shutdown.load(Ordering::Relaxed) {
+            break;
+        }
+
+        thread::sleep(Duration::from_secs(1));
+    }
+
+    should_shutdown.store(true, Ordering::Relaxed); // tell the threads to shut down their child process
```
and-rewsmith, line 155: We use a shared bool to signal the other threads to shut down the child processes. The threads also share bool that indicates their running/stopped status, so we can wait on this before returning from main.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+                process: child,
+            };
+
+            while child_process.is_running() && !has_parent_started_shutdown.load(Ordering::Relaxed)
+            {
+                let poll_interval: Duration = Duration::new(PROCESS_POLL_INTERVAL_SECS, 0);
+                thread::sleep(poll_interval);
+            }
+
+            child_process.shutdown_if_running();
```
and-rewsmith, line 58: I did not want to check the running/stopped status here and log because, even with this check here, we should still check before sending SIGTERM. So there is no reason to do the check twice.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+        let child = match command.spawn() {
+            Ok(child) => {
+                info!("Launched {:?} process with pid {:?}", name, child.id());
+                child
+            }
+            Err(e) => {
+                error!("Failed to start {:?} process. {:?}", name, e);
+                exit(2);
+            }
+        };
```
and-rewsmith, line 42: Is having this child process created before the thread spawns a good idea?

edge-hub/watchdog/linux/sh/blocking-operation.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# The most simple script i could come up with
+# for simulating a blocking thing - could be
+# `nginx -g 'daemon off;' or anything like that.
+for i in `seq 1 1000`; do
+  echo "$i\t `date +%s`"
+  sleep 1
+done
```
and-rewsmith, line 9: This is just for testing. I can delete this later if we think we won't need it again.

edge-hub/watchdog/linux/src/main.rs
vadim-kovalyov: I don't see a problem with that.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+    };
+
+    let has_broker_shutdown = ChildProcess::run_child_process(
+        "MQTT Broker".to_string(),
+        Command::new("/usr/local/bin/mqttd").stdout(Stdio::inherit()),
+        Arc::clone(&should_shutdown),
+    );
+
+    let has_edgehub_shutdown = ChildProcess::run_child_process(
```
vadim-kovalyov, line 134: I believe we want to start edgehub first, don't we?

and-rewsmith: It gets started inside run_child_process() then returns a bool that will tell main() if it is still running.

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: I thought that we would easily be able to filter based on the log headers, but that is not the case. Here is an example log of edgehub starting up:

```
Jul 06 21:47:12.324  INFO watchdog: Starting watchdog
Jul 06 21:47:12.402  INFO watchdog: Launched Edge Hub process with pid 8
Jul 06 21:47:12.403  INFO watchdog: Launched MQTT Broker process with pid 9
Jul 06 21:47:12.996  INFO mqttd: Loading state...
Jul 06 21:47:13.003  INFO mqtt_broker::persist: no state file found at /tmp/mqtt/state/state.dat.
Jul 06 21:47:13.004  INFO mqttd: state loaded.
Jul 06 21:47:13.004  INFO mqttd: Starting server...
Jul 06 21:47:13.004  INFO mqttd: Persisting state every 300s
Jul 06 21:47:13.004  INFO mqttd::snapshot::imp: Setup to persist state on USR1 signal
Jul 06 21:47:13.545  INFO mqtt_broker::transport: Loading identity from /tmp/mqtt/config/identity.pfx
Jul 06 21:47:13.620  INFO server{listener=0.0.0.0:1883}: mqtt_broker::server: Listening on address 0.0.0.0:1883
Jul 06 21:47:13.622  INFO server{listener=0.0.0.0:1883}:server{listener=0.0.0.0:8883}: mqtt_broker::server: Listening on address 0.0.0.0:8883
Jul 06 21:47:13.623  INFO server{listener=0.0.0.0:1883}:server{listener=0.0.0.0:8883}:server{listener=127.0.0.1:1882}: mqtt_broker::server: Listening on address 127.0.0.1:1882
2020-07-06 21:47:14.624 +00:00 Edge Hub Main()
<7> 2020-07-06 21:47:15.941 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Edged.WorkloadClient] - Making a Http call to unix:///var/run/iotedge/workload.sock to CreateServerCertificateAsync
<7> 2020-07-06 21:47:16.764 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Connecting socket /var/run/iotedge/workload.sock
<7> 2020-07-06 21:47:16.791 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Connected socket /var/run/iotedge/workload.sock
<7> 2020-07-06 21:47:16.803 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Sending request http://workload.sock/modules/%24edgeHub/genid/637244011215002617/certificate/server?api-version=2019-01-30
<7> 2020-07-06 21:47:16.927 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Response received Created
<7> 2020-07-06 21:47:16.972 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Edged.WorkloadClient] - Received a valid Http response from unix:///var/run/iotedge/workload.sock for CreateServerCertificateAsync
<7> 2020-07-06 21:47:17.359 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Edged.WorkloadClient] - Making a Http call to unix:///var/run/iotedge/workload.sock to TrustBundleAsync
<7> 2020-07-06 21:47:17.363 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Connecting socket /var/run/iotedge/workload.sock
<7> 2020-07-06 21:47:17.364 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Connected socket /var/run/iotedge/workload.sock
<7> 2020-07-06 21:47:17.364 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Sending request http://workload.sock/trust-bundle?api-version=2019-01-30
<7> 2020-07-06 21:47:17.365 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Response received OK
<7> 2020-07-06 21:47:17.367 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Edged.WorkloadClient] - Received a valid Http response from unix:///var/run/iotedge/workload.sock for TrustBundleAsync
<6> 2020-07-06 21:47:17.378 +00:00 [INF] [EdgeHub] - Installing certificates [CN=iotedged workload ca:08/18/2020 16:01:01],[CN=Test Edge Device CA:08/18/2020 16:01:01],[CN=Test Edge Owner CA:08/18/2020 16:01:01] to Root
<6> 2020-07-06 21:47:17.417 +00:00 [INF] [EdgeHub] - Installing certificates [CN=Test Edge Owner CA:08/18/2020 16:01:01] to Root
<6> 2020-07-06 21:47:17.430 +00:00 [INF] [EdgeHub] - Enabling SSL protocols: Tls, Tls11, Tls12
<6> 2020-07-06 21:47:18.216 +00:00 [INF] [EdgeHub] - Experimental features configuration: {"Enabled":false,"DisableCloudSubscriptions":false,"DisableConnectivityCheck":false}
<6> 2020-07-06 21:47:19.841 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Hub.Service.Modules.RoutingModule] - Created persistent store at /tmp/edgeHub
<6> 2020-07-06 21:47:19.862 +00:00 [INF] [EdgeHub] - Initializing Edge Hub

```

I would like to find a way to prepend appropriate log headers to each line so that one can effectively filter by watchdog, broker, or edgehub logs. Something like [Edge Hub], [Broker], [Watchdog]. I'm not sure if this is possible from the watchdog or we will have to do it from the edgehub/broker themselves. I'll check. What do you think of this idea?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+        "Edge Hub".to_string(),
+        Command::new("dotnet")
+            .arg("/app/Microsoft.Azure.Devices.Edge.Hub.Service.dll")
+            .stdout(Stdio::inherit()),
+        Arc::clone(&should_shutdown),
+    );
+
+    while !should_shutdown.load(Ordering::Relaxed) {
```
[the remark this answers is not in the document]
and-rewsmith: So we are sharing the watchdog's sigterm/sigint listener bool across threads, so you are right that there is no need for ```while !should_shutdown.load(Ordering::Relaxed)```. 

But ```has_broker_shutdown``` and ```has_edgehub_shutdown``` are not shared between threads. Currently, if either broker / edgehub stop then this while loop exits and we use the ```should_shutdown``` shared bool to signal the child threads to shutdown. How would this work with join()? If a join() call is waiting then it needs to be able to determine when the other child process shut down, so that it can shut itself down. Should I use channels to communicate across child threads directly? Not sure how this would work, but it would solve this problem right?

I will look into channels.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+            }
+            Err(e) => {
+                error!("Failed to start {:?} process. {:?}", name, e);
+                exit(2);
+            }
+        };
+
+        let mut has_shutdown = Arc::new(AtomicBool::new(false));
+        let has_shutdown_listener = Arc::clone(&mut has_shutdown);
+        thread::spawn(move || {
```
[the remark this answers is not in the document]
and-rewsmith: See my response to your other comment about the handle.

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: @dmolokanov This is what I tried to do initially, but since the command that I pass into this ```run_child_process function``` is freed at the end of the call to this same function, I had difficulty giving it to another thread. I thought that it was better to avoid the lifetimes and do what was needed with the command immediately.

edge-hub/watchdog/linux/src/main.rs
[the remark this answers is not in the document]
dmolokanov: I don't think you need channels here. Let's discuss it offline?

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+        "Edge Hub".to_string(),
+        "dotnet".to_string(),
+        "/app/Microsoft.Azure.Devices.Edge.Hub.Service.dll".to_string(),
+        Arc::clone(&should_shutdown),
+    );
+
+    match broker_handle.join() {
+        Ok(()) => info!("Successfully stopped broker process"),
+        Err(e) => {
+            should_shutdown.store(true, Ordering::Relaxed);
```
and-rewsmith, line 50: If one thread panics then make the other shut down.

edge-hub/watchdog/linux/src/child_process.rs
```diff
@@ -0,0 +1,62 @@
...
+};
+
+use std::{process::Child, thread::sleep, time::Duration};
+use tracing::{error, info};
+
+const PROCESS_SHUTDOWN_TOLERANCE_SECS: u64 = 5;
+pub const PROCESS_POLL_INTERVAL_SECS: u64 = 1;
+
+pub struct ChildProcess {
+    pub name: String,
```
dmolokanov, line 13: lets make a constructor function `pub fn run(name: String, process: Child) -> Self` instead of public members. it is usually a bad practice to the coupling on internal implementations

edge-hub/watchdog/linux/src/child_process.rs
```diff
@@ -0,0 +1,62 @@
...
+                );
+                false
+            }
+        }
+    }
+
+    fn wait_for_exit(&mut self) {
+        let mut elapsed_secs = 0;
+        while elapsed_secs < PROCESS_SHUTDOWN_TOLERANCE_SECS && self.is_running() {
+            let poll_interval: Duration = Duration::from_secs(PROCESS_POLL_INTERVAL_SECS);
```
dmolokanov, line 57:
```suggestion
            let poll_interval = Duration::from_secs(PROCESS_POLL_INTERVAL_SECS);
```
?

edge-hub/watchdog/linux/src/child_process.rs
```diff
@@ -0,0 +1,62 @@
+use nix::{
+    sys::signal::{self, Signal},
+    unistd::Pid,
+};
+
+use std::{process::Child, thread::sleep, time::Duration};
+use tracing::{error, info};
+
+const PROCESS_SHUTDOWN_TOLERANCE_SECS: u64 = 5;
+pub const PROCESS_POLL_INTERVAL_SECS: u64 = 1;
```
dmolokanov, line 10: Should it still be pub?

edge-hub/watchdog/linux/src/child_process.rs
```diff
@@ -0,0 +1,62 @@
+use nix::{
+    sys::signal::{self, Signal},
+    unistd::Pid,
+};
+
+use std::{process::Child, thread::sleep, time::Duration};
+use tracing::{error, info};
+
+const PROCESS_SHUTDOWN_TOLERANCE_SECS: u64 = 5;
```
dmolokanov, line 9: I'd rather made these constants as statics with `Duration` type

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+    shutdown_listener: Arc<AtomicBool>,
+) -> JoinHandle<()> {
+    spawn(move || {
+        let child = match Command::new(program)
+            .arg(args)
+            .stdout(Stdio::inherit())
+            .spawn()
+        {
+            Ok(child) => {
+                info!("Launched {:?} process with pid {:?}", name, child.id());
```
dmolokanov, line 92:
```suggestion
                info!("Launched {} process with pid {}", name, child.id());
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+
+fn register_shutdown_listener() -> Result<Arc<AtomicBool>, Error> {
+    info!("Registering shutdown signal listener");
+    let should_shutdown = Arc::new(AtomicBool::new(false));
+    register(SIGTERM, Arc::clone(&should_shutdown))?;
+    register(SIGINT, Arc::clone(&should_shutdown))?;
+    Ok(should_shutdown)
+}
+
+pub fn run_child_process(
```
dmolokanov, line 79: Can we make it to return `Result<JoinHandle<()>, anyhow::Error>` ? It will simplify code a bit and propagate an error and cause an error on an upper level.
On the upper level, you can make it different error codes for broker and edgehub.
```rust 
pub fn run( name: String,
    program: String,
    args: String) -> anyhow::Result<JoinHandle<()>> {

   let child = match Command::new(program)
            .arg(args)
            .stdout(Stdio::inherit())
            .spawn()
            .with_context(|| format!("Failed to start {:?} process.", name)?;

    info!("Launched {} process with pid {:?}", name, child.id());
    ..
}

File: edge-hub/watchdog/linux/src/main.rs [hunk 46]
```diff
@@ -0,0 +1,114 @@
...
+            .arg(args)
+            .stdout(Stdio::inherit())
+            .spawn()
+        {
+            Ok(child) => {
+                info!("Launched {:?} process with pid {:?}", name, child.id());
+                child
+            }
+            Err(e) => {
+                error!("Failed to start {:?} process. {:?}", name, e);
```
dmolokanov, line 96:
```suggestion
                error!("Failed to start {} process. {}", name, e);
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    let edgehub_handle = run_child_process(
+        "Edge Hub".to_string(),
+        "dotnet".to_string(),
+        "/app/Microsoft.Azure.Devices.Edge.Hub.Service.dll".to_string(),
+        Arc::clone(&should_shutdown),
+    );
+
+    match broker_handle.join() {
+        Ok(()) => info!("Successfully stopped broker process"),
```
dmolokanov, line 48: let's move this log message to the end of `run` function so the code will be a bit simpler
```rust
if let Err(e) = broker_handle.join() {
  should_shutdown.store(true, Ordering::Relaxed);
  error!("Failure while running broker process. {}", e)
}
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+        Err(e) => {
+            should_shutdown.store(true, Ordering::Relaxed);
+            error!("Failure while running broker process. {:?}", e)
+        }
+    };
+    match edgehub_handle.join() {
```
dmolokanov, line 54: Same comment as for broker join() call

edge-hub/watchdog/linux/src/main.rs
dmolokanov, line 79: it feels like this function place in the child_process module.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
+mod child_process;
```
dmolokanov, line 1:
```suggestion
mod child;
```
Coming from a c# it is hard to abandon the habit to make VeryLongAndDetailed types and module names even nit: where it is not necessary. What I noticed working with rust codebases, rules are much more relaxed. 
IMO watchdog is always will monitor processes only, so maybe "_process" part in  `child_process` is not necessary. 
Same for `fn run_child_process()` it the function is inside `child` modules it can be called only `run` without loosing any meanings.

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+    time::Duration,
+};
+
+use child_process::ChildProcess;
+use child_process::PROCESS_POLL_INTERVAL_SECS;
+use signal_hook::{flag::register, SIGINT, SIGTERM};
+use tracing::{error, info, subscriber, Level};
+use tracing_subscriber::fmt::Subscriber;
+
+fn main() -> Result<(), Error> {
```
dmolokanov, line 18:
```suggestion
fn main() -> anyhow::Result<()> {
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+    init_logging();
+    info!("Starting watchdog");
+
+    let should_shutdown = match register_shutdown_listener() {
```
dmolokanov, line 22:
```suggestion
    let should_shutdown = register_shutdown_listener().context("Failed to register sigterm listener. Shutting down.")?;
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+        Err(e) => {
+            should_shutdown.store(true, Ordering::Relaxed);
+            error!("Failure while running broker process. {:?}", e)
+        }
+    };
+    match edgehub_handle.join() {
+        Ok(()) => info!("Successfully stopped edgehub process"),
+        Err(e) => {
+            should_shutdown.store(true, Ordering::Relaxed);
+            error!("Failure while running edgehub process. {:?}", e)
```
dmolokanov, line 58:
```suggestion
            error!("Failure while running edgehub process. {}", e)
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+        "dotnet".to_string(),
+        "/app/Microsoft.Azure.Devices.Edge.Hub.Service.dll".to_string(),
+        Arc::clone(&should_shutdown),
+    );
+
+    match broker_handle.join() {
+        Ok(()) => info!("Successfully stopped broker process"),
+        Err(e) => {
+            should_shutdown.store(true, Ordering::Relaxed);
+            error!("Failure while running broker process. {:?}", e)
```
dmolokanov, line 51:
```suggestion
            error!("Failure while running broker process. {}", e)
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+    Ok(should_shutdown)
+}
+
+pub fn run_child_process(
+    name: String,
+    program: String,
+    args: String,
+    shutdown_listener: Arc<AtomicBool>,
+) -> JoinHandle<()> {
+    spawn(move || {
```
vadim-kovalyov, line 85: nit: I believe it is better to use prefix for external func `thread::spawn`.
```suggestion
    thread::spawn(move || {
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+                exit(2);
+            }
+        };
+        let mut child_process = ChildProcess {
+            name,
+            process: child,
+        };
+
+        while child_process.is_running() && !shutdown_listener.load(Ordering::Relaxed) {
+            let poll_interval: Duration = Duration::from_secs(PROCESS_POLL_INTERVAL_SECS);
```
vadim-kovalyov, line 106:
```suggestion
            let poll_interval = Duration::from_secs(PROCESS_POLL_INTERVAL_SECS);
```

edge-hub/watchdog/linux/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+    register(SIGTERM, Arc::clone(&should_shutdown))?;
+    register(SIGINT, Arc::clone(&should_shutdown))?;
+    Ok(should_shutdown)
+}
+
+pub fn run_child_process(
+    name: String,
+    program: String,
+    args: String,
+    shutdown_listener: Arc<AtomicBool>,
```
vadim-kovalyov, line 83:
```suggestion
    should_shutdown: Arc<AtomicBool>,
```

edge-hub/watchdog/linux/src/main.rs
and-rewsmith: Oh I just realized I misread this. Do you think we should switch the order?

edge-hub/watchdog/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+    let shutdown_listener = Arc::new(AtomicBool::new(false));
+    let should_shutdown = shutdown_listener.clone();
+    let signals = Signals::new(&[SIGTERM, SIGINT])?;
+    thread::spawn(move || {
+        for signal in signals.forever() {
+            let signal_name = match signal {
+                SIGTERM => "SIGTERM",
+                SIGINT => "SIGINT",
+                _ => "Unexpected",
+            };
```
and-rewsmith, line 78: I tried to map the int to a name automatically but didn't find a way to do that.

edge-hub/watchdog/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+    };
+
+    if let Err(e) = broker_handle.join() {
+        should_shutdown.store(true, Ordering::Relaxed);
+        error!("Failure while running broker process. {:?}", e)
+    }
+    info!("Successfully stopped broker process");
+
+    edgehub_handle.map(|handle| {
```
dmolokanov:
```suggestion
    edgehub_handle.map(|handle| handle.join()).with_context(|| {
        should_shutdown.store(true, Ordering::Relaxed);
        format!("Failure while running edgehub process. {:?}", e)
    }?;
    info!("Successfully stopped edgehub process");
```

edge-hub/watchdog/src/child.rs
```diff
@@ -0,0 +1,101 @@
...
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    thread,
+    thread::{sleep, JoinHandle},
+    time::Duration,
+};
+use tracing::{error, info};
+
+const PROCESS_SHUTDOWN_TOLERANCE_SECS: Duration = Duration::from_secs(45);
```
and-rewsmith, line 18: This is coupled to iotedged's shutdown tolerance for the edgehub container. The watchdog shutdown tolerance needs to be a bit shorter than iotedged's to make sure it can reap the child pids. 

I now see a problem where someone configures iotedged's shutdown tolerance to be super low. Iotedged will kill root docker process then the kernel will kill orphaned children leaving them as zombies. The same thing would happen if we were using a supervisor. I'm going to look for a way around this.

edge-hub/watchdog/src/main.rs
dmolokanov: I'd say you don't need to change  shutdown_listener into true for unexpected signals. What you can do it to apply https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map on the iterator
```rust
signals.forever()
.filter_map(|signal| match signal {
  SIGTERM => Some("SIGTERM"),
  SIGINT => Some("SIGINT"),
  _ => None})
.for_each(|signal| {
  info!("Watchdog received {} signal", signal);
  shutdown_listener.store(true, Ordering::Relaxed);
});
```

edge-hub/watchdog/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+
+fn main() -> Result<()> {
+    init_logging();
+    info!("Starting watchdog");
+
+    let should_shutdown = register_shutdown_listener()
+        .context("Failed to register sigterm listener. Shutting down.")?;
+
+    let edgehub_handle = run(
+        "Edge Hub".to_string(),
```
dmolokanov, line 24: somewhere in logs, you put "edgehub", somewhere else "Edge Hub". lets add some consistency across watchdog logs? same comment for broker

edge-hub/watchdog/src/child.rs
and-rewsmith: So this isn't a super big deal because we can just make this time configurable. @dmolokanov Do you agree?

edge-hub/watchdog/src/main.rs
and-rewsmith: The way I meant to do this is "Edge Hub" and "MQTT Broker". I'll fix the offending instances

edge-hub/watchdog/src/child.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn wait_for_exit(&mut self) {
+        let mut elapsed_secs = 0;
+        while elapsed_secs < PROCESS_SHUTDOWN_TOLERANCE_SECS.as_secs() && self.is_running() {
+            sleep(PROCESS_POLL_INTERVAL_SECS);
+            elapsed_secs += PROCESS_POLL_INTERVAL_SECS.as_secs();
+        }
+    }
+}
+
+pub fn run(
```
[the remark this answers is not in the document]
and-rewsmith: Cool trick, but every time you want to use name you will also need into(). There is now a lot more noise than before imo. Do you still think it is cleaner this way?

edge-hub/watchdog/src/main.rs
and-rewsmith: This wouldn't compile because handle.join() returns (), for which std::ops::try is not implemented.

```
the `?` operator can only be applied to values that implement `std::ops::Try`
the trait `std::ops::Try` is not implemented for `()`
required by `std::ops::Try::into_result`
```

I am not sure why though because handle.join() should return Result<()> which should be fine?

edge-hub/watchdog/src/child.rs
and-rewsmith: Reading through this thread I am not sure if it is an issue. The poster says that the processes are removed from the process table, but a commenter seems to think it is still possible for these to become zombies.

https://stackoverflow.com/questions/39739658/what-happens-to-other-processes-when-a-docker-containers-pid1-exits

and-rewsmith: Crisis averted:
https://stackoverflow.com/questions/49162358/docker-init-zombies-why-does-it-matter/49162867

_Next, a container is a cgroup of processes running in their own PID namespace. This cgroup is cleaned up when the container is stopped. Any zombies that are in a container are removed on stop. They don't reach the hosts init._

_Third is the different ways containers are used. Most run one main process and nothing else. If there is another process spawned it is usually a child of that main process. So until the parent exits, the zombie will exist. Then see point 2 (the zombies will be cleared on container exit_

edge-hub/watchdog/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    if let Err(e) = edgehub_handle.join() {
+        should_shutdown.store(true, Ordering::Relaxed);
+        error!("Failure while running MQTT Broker process. {:?}", e)
+    }
+    info!("Successfully stopped Edge Hub process");
+
+    if let Some(Err(e)) = broker_handle.map(|handle| handle.join()) {
+        should_shutdown.store(true, Ordering::Relaxed);
+        error!("Failure while running MQTT Broker process. {:?}", e);
```
dmolokanov, line 52: I think you need to stop here. otherwise, log messages will confuse.
```suggestion
        bail!("Failure while running MQTT Broker process. {:?}", e);
```

and-rewsmith: I think I would rather fix this with else

and-rewsmith: The case on line 44 is easier to handle. For this block I thought I could add an ```else if let Some(_) = broker_handle``` but I don't think this is true because ```broker_handle``` is moved in the initial map() call

and-rewsmith: I used a pattern shown here:
https://doc.rust-lang.org/stable/rust-by-example/flow_control/if_let.html

See ```// Provide an altered failing condition```

edge-hub/watchdog/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "watchdog"
+version = "0.1.0"
+authors = ["Azure IoT Edge Devs"]
+edition = "2018"
+
+[dependencies]
+nix = "0.17.0"
```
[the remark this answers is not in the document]
and-rewsmith: Is this what we have determined is best practice? Not specifying minor version?

edge-hub/watchdog/src/child.rs
```diff
@@ -74,11 +74,15 @@ impl ChildProcess {
...
-    name: String,
-    program: String,
-    args: String,
+    name: impl Into<String>,
+    program: impl Into<String>,
+    args: impl Into<String>,
     should_shutdown: Arc<AtomicBool>,
 ) -> Result<JoinHandle<()>> {
+    let name = name.into();
+    let program = program.into();
```
dmolokanov, line 83: nit: for those that you used only once you can inline `xxx.into()` calls without hurting readability.

edge-hub/watchdog/src/main.rs
dmolokanov: yeah, the second block is a bit harder to read though?
let's go back with 
```rust
if let Some(handle) = broker_handle {
  if let Err(e) = handle.join() {
    ...
  }
} else {
  ...
}
```
another fancy way to solve nesting problem is `transpose` operator
```rust
if let Some(_) = broker_handle
  .map(|handle|handle.join().map_err(|e| anyhow!("Failure while running MQTT Broker process. {:?}", e)) // makes Option<Result<()>>
  .transpose()? { // makes Result<Option<()>>
  info!("Successfully stopped MQTT Broker process");
}  
```

and-rewsmith: I was playing around with the transpose, and it was autoformatting to more nested levels than your first block suggestion above :).

I'm going to go with the second suggestion, but I think we want the ending else statement to be inside the top-level if block.

edge-hub/watchdog/src/child.rs
and-rewsmith: I personally think this is more readable. I didn't like how we were putting the into() function call into other logic.

edge-hub/watchdog/Cargo.toml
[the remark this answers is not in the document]
dmolokanov: yeah. it is expected that there should not be breaking changes within major version. and it helps to update crates

edge-hub/watchdog/src/child.rs
```diff
@@ -0,0 +1,113 @@
...
+            self.send_signal(Signal::SIGKILL);
+        }
+
+        self.wait_for_exit();
+    }
+
+    fn send_signal(&mut self, signal: Signal) {
+        info!("Sending {} signal to {} process", signal, self.name);
+
+        /* Casting process.id() from u32 to i32 cannot wrap. The max pid for linux is ~4 million. The max value for i32 is 2,147,483,647.
```
dmolokanov, line 63: 👍

edge-hub/watchdog/src/child.rs
dmolokanov: You'll get used to it soon 🤣

edge-hub/watchdog/src/child.rs
```diff
@@ -0,0 +1,113 @@
...
+        .spawn()
+        .with_context(|| format!("Failed to start {:?} process.", name))?;
+
+    let handle = thread::spawn(move || {
+        info!("Launched {} process with pid {}", name, child.id());
+
+        let mut child_process = ChildProcess::new(name, child);
+
+        while child_process.is_running() && !should_shutdown.load(Ordering::Relaxed) {
+            sleep(PROCESS_POLL_INTERVAL_SECS);
```
dmolokanov, line 104: nit
```suggestion
            thread::sleep(PROCESS_POLL_INTERVAL_SECS);
```

edge-hub/watchdog/src/child.rs
dmolokanov, line 63: nit: I don't think it is a rule, but if you could change the comment line width to something around 80 symbols, @vadim-kovalyov will really appreciate.
Also could you switch to `//` type of comment event for multiline please?

edge-hub/watchdog/src/child.rs
```diff
@@ -0,0 +1,113 @@
+use std::{
+    process::{Child, Command, Stdio},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    thread,
```
dmolokanov, line 7: nit
```suggestion
   thread::{self, JoinHandle},
```

PR #3142, opened 2020-06-24
update docker file for vulnerability issue

philipktlin:
update Linux arm32 and arm64 docker base image for vulnerability issue. And fix docker base image/tag in EdgeHubRestartTester windows arm32 and TwinTester.

### DIFF
--- a/edge-agent/docker/linux/arm32v7/Dockerfile
+++ b/edge-agent/docker/linux/arm32v7/Dockerfile
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.5-linux-arm32v7
+ARG base_tag=1.0.6-linux-arm32v7
 FROM azureiotedge/azureiotedge-agent-base:${base_tag}
 
 ARG EXE_DIR=.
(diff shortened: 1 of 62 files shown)


edge-agent/docker/linux/arm32v7/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.5-linux-arm32v7
+ARG base_tag=1.0.6-linux-arm32v7
```
philipktlin, line 1: change to 1.1.0

edge-agent/docker/windows/arm32v7/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.5-windows-arm32v7
+ARG base_tag=1.1.0-windows-arm32v7
```
philipktlin, line 1: I changed windows to be consistent that upgrade to netcore3.1 should start from 1.1.x.  I think we don't need to build Windows ARM32 base image for now, although image build still includes windows arm32 docker image but edgelet build doesn't.  Please confirm with Varun, if it is ok to remove Windows ARM32 from image build.  If so, no need to build Windows ARM32 base image, otherwise please build new docker image.

PR #3146, opened 2020-06-25
Allow Clippy's similar_names warning

yophilav:
Allow Clippy's similar_names warning

### DIFF
--- a/edgelet/edgelet-hsm/tests/crypto_get_random.rs
+++ b/edgelet/edgelet-hsm/tests/crypto_get_random.rs
@@ -2,6 +2,7 @@
 
 #![deny(unused_extern_crates, warnings)]
 #![deny(clippy::all, clippy::pedantic)]
+#![allow(clippy::similar_names)]
 
 use lazy_static::lazy_static;
 use std::sync::Mutex;
--- a/edgelet/iotedge-proxy/src/proxy/client.rs
+++ b/edgelet/iotedge-proxy/src/proxy/client.rs
@@ -1,7 +1,5 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-#![allow(clippy::similar_names)]
-
 use std::convert::AsRef;
 
 use failure::{Fail, ResultExt};
--- a/edgelet/iotedge-proxy/src/proxy/client.rs
+++ b/edgelet/iotedge-proxy/src/proxy/client.rs
@@ -1,5 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
+#![allow(clippy::similar_names)]
+
 use std::convert::AsRef;
 
 use failure::{Fail, ResultExt};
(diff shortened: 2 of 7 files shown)


edgelet/edgelet-hsm/tests/crypto_get_random.rs
```diff
@@ -2,6 +2,7 @@
 
 #![deny(unused_extern_crates, warnings)]
 #![deny(clippy::all, clippy::pedantic)]
+#![allow(clippy::similar_names)]
```
philipktlin, line 5: from https://rust-lang.github.io/rust-clippy/master/ and filter by "similar_names", it shows that similar_names rule is "Checks for names that are very similar and thus confusing".  I think we should fix it instead of disable it.  Does this issue happen in master.  I think Mike asked in teams, but no one responds.  Please reply in teams and confirm no one is currently working on it.

yophilav: Arnav suggested to suppress.  There are already a few of allowing the ` similar_names` in the master.

arsing:
There are still failures, in `iotedge-proxy/src/proxy/service.rs` now.

yophilav:
working on it.

edgelet/edgelet-hsm/tests/crypto_get_random.rs
philipktlin: @arsing since we disable at file level, it will not detect when any new changes fail with this rule.  If so, should we just disable at global level instead of at file level?

arsing: This is the global level. Files under `tests/` are integration tests. Each file is compiled as a crate.

edgelet/iotedge-proxy/src/proxy/client.rs
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
+#![allow(clippy::similar_names)]
```
arsing, line 3: This should be in `src/lib.rs`

edgelet/iotedge-proxy/src/proxy/service.rs
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
+#![allow(clippy::similar_names)]
```
arsing, line 3: This should be in `src/lib.rs`

edgelet/iotedge/src/support_bundle.rs
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
+#![allow(clippy::similar_names)]
```
arsing, line 3: This should be in `src/lib.rs`

PR #3151, opened 2020-06-26
Add manifest step for release image build

and-rewsmith:
When converting the pipeline to yaml I missed the manifest step, as I mistakenly thought it happened in the subsequent release job. Adding it here.

### DIFF
--- a/builds/misc/images-release.yaml
+++ b/builds/misc/images-release.yaml
@@ -608,7 +608,7 @@ jobs:
 ################################################################################
   - job: manifest
 ################################################################################
-    displayName: Manifest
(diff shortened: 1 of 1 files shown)


builds/misc/images-release.yaml
```diff
@@ -605,3 +605,21 @@ jobs:
         inputs: 
           ArtifactName: publish-win
           PathtoPublish: $(Build.BinariesDirectory)\publish
+################################################################################
+  - job: manifest
+################################################################################
+    displayName: Manifest
```
yophilav, line 611: Nit:
``` suggestion
    displayName: Image Manifest Publication
```

PR #3152, opened 2020-06-27
Correct Image Tagging & Manifest Publication on "Azure-IoT-Edge-Core Images Release YAML" Pipeline

yophilav:
Correct Image Tagging & Manifest Publication on "Azure-IoT-Edge-Core Images Release YAML" Pipeline. 

Currently the "Azure-IoT-Edge-Core Images Release YAML" Pipeline incorrectly tags a releasing image with a BuildNumber causing the manifest publication step to fail when it tries to pull the published image to generate a manifest.

### DIFF
--- a/builds/misc/images-release.yaml
+++ b/builds/misc/images-release.yaml
@@ -617,9 +617,9 @@ jobs:
     steps:
     - script: scripts/linux/buildManifest.sh -r '$(registry.address)' -u '$(registry.user)' -p '$(registry.password)' -v '$(version)' -t '$(System.DefaultWorkingDirectory)/edgelet/iotedge-diagnostics/docker/manifest.yaml.template' -n '$(namespace)' --tags '$(tags)'
       displayName: 'Publish azureiotedge-diagnostics Manifest'
-    - script: scripts/linux/buildManifest.sh -r $(registry.address) -u $(registry.user) -p $(registry.password) -v $(version) -t $(System.DefaultWorkingDirectory)/edge-agent/docker/manifest.yaml.template -n $(namespace) --tags "$(tags)"
+    - script: scripts/linux/buildManifest.sh -r '$(registry.address)' -u '$(registry.user)' -p '$(registry.password)' -v '$(version)' -t '$(System.DefaultWorkingDirectory)/edge-agent/docker/manifest.yaml.template' -n '$(namespace)' --tags '$(tags)'
       displayName: 'Publish Edge Agent Manifest'
-    - script: scripts/linux/buildManifest.sh -r $(registry.address) -u $(registry.user) -p $(registry.password) -v $(version) -t $(System.DefaultWorkingDirectory)/edge-hub/docker/manifest.yaml.template -n $(namespace) --tags "$(tags)"
+    - script: scripts/linux/buildManifest.sh -r '$(registry.address)' -u '$(registry.user)' -p '$(registry.password)' -v '$(version)' -t '$(System.DefaultWorkingDirectory)/edge-hub/docker/manifest.yaml.template' -n '$(namespace)' --tags '$(tags)'
       displayName: 'Publish Edge Hub Manifest'
-    - script: scripts/linux/buildManifest.sh -r $(registry.address) -u $(registry.user) -p $(registry.password) -v $(version) -t $(System.DefaultWorkingDirectory)/edge-modules/SimulatedTemperatureSensor/docker/manifest.yaml.template -n $(namespace) --tags "$(tags)"
+    - script: scripts/linux/buildManifest.sh -r '$(registry.address)' -u '$(registry.user)' -p '$(registry.password)' -v '$(version)' -t '$(System.DefaultWorkingDirectory)/edge-modules/SimulatedTemperatureSensor/docker/manifest.yaml.template' -n '$(namespace)' --tags '$(tags)'
       displayName: 'Publish Temperature Sensor Manifest'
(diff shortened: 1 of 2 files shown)


varunpuranik:
Can you add some more details on this fix?

and-rewsmith:
@varunpuranik The release build isn't working because we switched to using existing templates when converting to yaml. The existing template for linux does not use the pipeline's version parameter. We can fix this by changing the template to accept this parameter and passing it into the script.

builds/misc/images-release.yaml
```diff
@@ -617,9 +620,9 @@ jobs:
...
       displayName: 'Publish Edge Agent Manifest'
-    - script: scripts/linux/buildManifest.sh -r $(registry.address) -u $(registry.user) -p $(registry.password) -v $(version) -t $(System.DefaultWorkingDirectory)/edge-hub/docker/manifest.yaml.template -n $(namespace) --tags "$(tags)"
+    - script: scripts/linux/buildManifest.sh -r '$(registry.address)' -u '$(registry.user)' -p '$(registry.password)' -v '$(version)' -t '$(System.DefaultWorkingDirectory)/edge-hub/docker/manifest.yaml.template' -n '$(namespace)' --tags '$(tags)'
       displayName: 'Publish Edge Hub Manifest'
-    - script: scripts/linux/buildManifest.sh -r $(registry.address) -u $(registry.user) -p $(registry.password) -v $(version) -t $(System.DefaultWorkingDirectory)/edge-modules/SimulatedTemperatureSensor/docker/manifest.yaml.template -n $(namespace) --tags "$(tags)"
+    - script: scripts/linux/buildManifest.sh -r '$(registry.address)' -u '$(registry.user)' -p '$(registry.password)' -v '$(version)' -t '$(System.DefaultWorkingDirectory)/edge-modules/SimulatedTemperatureSensor/docker/manifest.yaml.template' -n '$(namespace)' --tags '$(tags)'
```
and-rewsmith, line 627: Why this change?

builds/misc/templates/image-linux.yaml
```diff
@@ -3,21 +3,22 @@ parameters:
...
   project: ''
+  version: ''
 
 steps:
   - task: Bash@3
     displayName: Build Image - ${{ parameters.name }} - amd64
     inputs:
       filePath: scripts/linux/buildImage.sh
-      arguments: -r $(registry.address) -u $(registry.user) -p $(registry.password) -i ${{ parameters.imageName }} -n ${{ parameters.namespace }} -P ${{ parameters.project }}
+      arguments: -r "$(registry.address)" -u "$(registry.user)" -p "$(registry.password)" -i "${{ parameters.imageName }}" -n "${{ parameters.namespace }}" -P "${{ parameters.project }}" -v "${{ parameters.version }}"
```
and-rewsmith, line 13: We changed to single quotes in images-release yaml but double quotes here. Was this intentional?

builds/misc/templates/image-linux.yaml
```diff
@@ -3,21 +3,22 @@ parameters:
   imageName: ''
   namespace: 'microsoft'
   project: ''
+  version: ''
```
and-rewsmith, line 6: The default value here should be $(Build.BuildNumber) or it will break the regular image build

builds/misc/images-release.yaml
```diff
@@ -299,16 +299,19 @@ jobs:
           imageName: azureiotedge-agent
           name: "Edge Agent"
           project: Microsoft.Azure.Devices.Edge.Agent.Service
+          version: $(version)
```
and-rewsmith, line 302: Does this need to be ${{ parameters.version }} ?

builds/misc/images-release.yaml
yophilav: It is a better practice for shell scripting. By putting the `'` around an argument make sure the string-substitution of a variable is not confused as a flag.

builds/misc/templates/image-linux.yaml
yophilav: Yes, we are using `task: Bash@3` syntax here. While above we used `script:` syntax

builds/misc/templates/image-linux.yaml
yophilav: No it won't. The script has the logic to figure out if a string is empty or not. If the string is empty, it will use the `$(Build.BuildNumber)`. I made sure of it by code inspection + running the test to verify.

For Image build pipeline: 

https://dev.azure.com/msazure/One/_build/results?buildId=32358519&view=logs&j=cb472997-8ad7-5b8a-089b-78e3f1db5255

For Image release pipeline

https://msazure.visualstudio.com/One/_build/results?buildId=32358553&view=logs&j=7b5daa7d-adff-5805-f451-ad7193802469&t=efbf70ad-10b8-5ad3-3a7c-d6ad7449acae

builds/misc/images-release.yaml
yophilav: That depending on where the `$(version)` is being referred.
In this spot, the `${{ parameters.version }}` is actually being set to the value of `$(version)` from the build pipeline's variable. (They are 2 different variables).

builds/misc/images-release.yaml
philipktlin: Technically I can use double quote instead of single quote.  Why use single quote instead of double quote?

yophilav: I chose `'` instead of `"` because it looks better with the VSTS syntax.

PR #3153, opened 2020-06-27
[mqtt] Remove PUBACK0

dmolokanov:
TODO: limit numbers of all messages in-transit from broker to connection.

### DIFF
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -235,7 +235,7 @@ where
                 Ok(())
             }
             ClientEvent::PublishFrom(publish, _) => self.process_publish(&client_id, publish),
-            ClientEvent::PublishTo(_publish) => {
+            ClientEvent::PublishTo(_publish, _) => {
                 info!("broker received a PublishTo, ignoring");
                 Ok(())
             }
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -1,8 +1,13 @@
-use std::collections::HashMap;
-use std::convert::TryInto;
-use std::panic;
+use std::{collections::HashMap, convert::TryInto, panic};
 
-use tokio::sync::mpsc::{self, Receiver, Sender};
+use futures_util::{
+    future::{self, Either},
+    pin_mut,
+};
+use tokio::sync::{
+    mpsc::{self, Receiver, Sender},
+    Notify,
+};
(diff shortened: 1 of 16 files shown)


mqtt/mqtt-broker/src/broker.rs
```diff
@@ -29,9 +34,64 @@ macro_rules! try_send {
     }};
 }
 
-pub struct Broker<Z> {
+pub struct BrokerRunner<Z> {
```
vadim-kovalyov, line 37: can you please clarify why do we need this separate struct?

dmolokanov: Broker used to have 2 "parts" messages receiving and messages processing logic. method `run` deconstructs broker instance so we could different functions can own different parts of the broker. So the separation I made is somewhat intuitive IMO.
The problem I ran into: `Receiver::recv()` takes `&mut self`, then I make `pin_mut` its future and pass ownership to `future::select()`. So it makes it impossible to have another borrow of mutable reference to `self` when I need to process a message which also requires mutable reference to `self`. If I split messages receiving and messages processing between different instance there are no such problem.

vadim-kovalyov: I see. The second part make sense. Nit naming suggestion `BrokerLoop`?

dmolokanov: `BrokerRunner` has `run` method that's why it is runner :)

mqtt/mqtt-broker/src/session/state/mod.rs
```diff
@@ -226,36 +224,30 @@ impl SessionState {
...
-            Some(limit) => {
-                let num_inflight = self.waiting_to_be_acked.len()
-                    + self.waiting_to_be_acked_qos0.len()
-                    + self.waiting_to_be_completed.len();
-                num_inflight < limit.get()
-            }
-            None => true,
-        }
+    pub(super) fn allowed_to_send(&self) -> Option<PublishPermit> {
+        self.config
```
vadim-kovalyov, line 238: :crying_cat_face: this is unreadable... can we make it simple with `match` please?

mqtt/mqtt-broker/src/lib.rs
```diff
@@ -193,6 +196,50 @@ pub enum Message {
     System(SystemEvent),
 }
 
+#[derive(Debug, Clone)]
+pub struct PublishPermits {
```
vadim-kovalyov, line 200: :+1:

PR #3154, opened 2020-06-28
Add logic for image configuration in image-linux.sh

and-rewsmith:
The release build isn't working because we switched to using existing templates when converting to yaml. The existing template for linux does not use the pipeline's version parameter. We can fix this by changing the template to accept this parameter and passing it into the script.


and-rewsmith:
Closing because duplicate #3152

PR #3157, opened 2020-06-29
add ms_telemetry to edgehub metrics

lfitchett:
Add ms_telemetry tag to edgehub metrics. This tag tells edgeAgent whether it should upload the metric through internal channels or not.

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/linkhandlers/ModuleMessageLinkHandler.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/linkhandlers/ModuleMessageLinkHandler.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Amqp.LinkHandlers
             static readonly IMetricsCounter MessagesMeter = Util.Metrics.Metrics.Instance.CreateCounter(
                 "messages_sent",
                 "Messages sent to module",
-                new List<string> { "from", "to", "from_route_output", "to_route_input", "priority" });
+                new List<string> { "from", "to", "from_route_output", "to_route_input", "priority", "ms_telemetry" });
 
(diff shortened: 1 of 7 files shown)


and-rewsmith:
Can you elaborate on:
_upload the metric through internal channels or not_

 In what case would we be uploading metrics through these channels vs not?

and-rewsmith:
Checkin was failing but looked transient so i triggered a rerun

lfitchett:
> Can you elaborate on:
> _upload the metric through internal channels or not_
> 
> In what case would we be uploading metrics through these channels vs not?

This is the distinction for our internal pipeline. Customers can scrape all metrics regardless of this tag. This just lets edgeAgent know which metrics to collect for p360.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/linkhandlers/ModuleMessageLinkHandler.cs
```diff
@@ -46,7 +46,7 @@ static class Metrics
             static readonly IMetricsCounter MessagesMeter = Util.Metrics.Metrics.Instance.CreateCounter(
                 "messages_sent",
                 "Messages sent to module",
-                new List<string> { "from", "to", "from_route_output", "to_route_input", "priority" });
+                new List<string> { "from", "to", "from_route_output", "to_route_input", "priority", "ms_telemetry" });
```
and-rewsmith, line 49: I'm guessing you have considered an alternate approach where edge agent maintains the state to let it know what metrics it should scrape?

The downsides of this are that this list would need to be maintained, but then we wouldn't have to include this ms_telemetry data with every metric.

and-rewsmith:
This is for all edge hub metrics right? Are you porting this to iiot? If so, Viktor is replacing the mqtt protocol head so you will have to sync with him for how to cleanly port this over.

PR #3160, opened 2020-06-30
Introduce Azure-IoT-Edge-Core Images Release to 1.0.9 branch

yophilav:
- Build Images Release: Convert to yaml (9e04b67)
- Add manifest step for release image build (bee08ed)
- Correct Image Tagging & Manifest Publication on "Azure-IoT-Edge-Core Images Release YAML" Pipeline (d1bd62d)
- Correct BuildNumber to be a version for Image Release Yaml pipeline (e8249557)


philipktlin:
I think we can't merge this commit: Build Images Release: Convert to yaml (9e04b67), as 1.0.9 still uses dotnet 2.1 and you can't combine all architecture in a single job.  And I think we should continue to use Build Images Release for 1.0.9 branch and 1.0.10 or later will use Build Images Release (YAML).

PR #3163, opened 2020-06-30
Nested edge demo

richma-ms:
A branch to compile all in-flight changes for the nested edge demo


richma-ms:
Closing this PR, any changes here that need to be merged should have its own separate bug/WI and corresponding PR.

PR #3164, opened 2020-06-30
fix sas token generation

philipktlin:
Always use iot hub name for token generation and use parent hostname as upstream gateway for registry api call if it is provided, otherwise use iot hub name.

### DIFF
--- a/edgelet/contrib/config/linux/debian/config.yaml
+++ b/edgelet/contrib/config/linux/debian/config.yaml
@@ -230,9 +230,8 @@ hostname: "<ADD HOSTNAME HERE>"
 # Optional. This entry should only be specified for the host name of the 
 # parent edge device for nested edge sceanrio. This value is injected into 
 # Edge Agent and Edge Hub modules as environment value 'IOTEDGE_GATEWAYHOSTNAME',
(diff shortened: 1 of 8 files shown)


edgelet/contrib/config/linux/debian/config.yaml
```diff
@@ -223,6 +223,20 @@ agent:
...
 hostname: "<ADD HOSTNAME HERE>"
 
+###############################################################################
+# Nested Edge parent hostname
+###############################################################################
+#
+# Optional. This entry should only be specified for the host name of the 
+# parent edge device for nested edge sceanrio. This value is injected into 
+# Edge Agent and Edge Hub modules as environment value 'IOTEDGE_GATEWAYHOSTNAME',
+# and injected into custom module as environment value 'IOTEDGE_PARENTHOSTNAME'.
```
varunpuranik, line 233: Nit: other modules as

edgelet/contrib/config/linux/debian/config.yaml
```diff
@@ -223,6 +223,20 @@ agent:
...
+# Regardless of case the parent_hostname is specified below, a lower case value
```
varunpuranik, line 234: Nit: The parent_hostname value is converted to lower case before it is used.

PR #3165, opened 2020-06-30
[mqtt] Unbounded messages channel + separate acks channel

dmolokanov:
There are some perf improvements include:
1. to avoid blocking on `PubAck0` events to the broker, change a channel type from `bounded(1024)` to `unbounded`. Broker channel is unbounded by implementation but it is bounded by the number of clients. Max length of the channel is `max_len = num_publishers * publish_permits + c`. By default `publish_permits = 10`, `c = small constant` - how many other messages in the system`.
2. all acknowledgments go into the separate channel, which is unbounded. It will help to speed up the processing of sent items. Instead of going to the end of the broker channel, it goes to a separate channel.

Acknowledgments channel has a priority and all Ack messages will be processed **before** any other message types.


mqtt/mqtt-broker/src/broker.rs
```diff
@@ -1003,11 +1007,19 @@ where
 }
 
 #[derive(Clone, Debug)]
-pub struct BrokerHandle(Sender<Message>);
+pub struct BrokerHandle(UnboundedSender<Message>, UnboundedSender<Message>);
```
vadim-kovalyov, line 1010: nit: would it be more clear to make it struct with named fileds "ack_sender", "message_sender"?

PR #3172, opened 2020-07-02
Fix edge agent connection 1.0.9

ancaantochi:
Removed the check for TimeoutException
Call dispose instead of close on the module sdk

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
@@ -213,7 +213,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub
             {
                 Events.ErrorGettingTwin(e);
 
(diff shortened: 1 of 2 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -213,7 +213,7 @@ async Task<Twin> GetTwinFunc()
             {
                 Events.ErrorGettingTwin(e);
 
-                if (!retrying && moduleClient != null && !(e is TimeoutException))
+                if (!retrying && moduleClient != null)
```
varunpuranik, line 216: Test?

ancaantochi: running  connectivity tests, do you mean adding unit tests?

varunpuranik: yes

PR #3174, opened 2020-07-03
Fix edge agent connection (#3172)

ancaantochi:
* Recreate connection after every exception

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
@@ -264,7 +264,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub
             {
                 Events.ErrorGettingTwin(e);
 
-                if (!retrying && moduleClient != null && !(e is TimeoutException))
+                if (!retrying && moduleClient != null)
                 {
                     try
                     {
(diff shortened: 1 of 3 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -264,7 +264,7 @@ async Task<Twin> GetTwinFunc()
             {
                 Events.ErrorGettingTwin(e);
 
-                if (!retrying && moduleClient != null && !(e is TimeoutException))
+                if (!retrying && moduleClient != null)
```
varunpuranik, line 267: unit tests?

ancaantochi: just added a test for TimeoutException

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/sdkClient/WrappingSdkModuleClient.cs
```diff
@@ -14,7 +15,18 @@ public class WrappingSdkModuleClient : ISdkModuleClient
...
-        public Task OpenAsync() => this.sdkModuleClient.OpenAsync();
+        public Task OpenAsync()
+        {
+            try
+            {
+                return this.sdkModuleClient.OpenAsync();
+            }
+            catch (Exception)
+            {
+                this.sdkModuleClient?.Dispose();
```
philipktlin, line 26: do we need to close first before calling dispose?

ancaantochi: no need to close, confirmed with SDK team that we should use Dispose instead as is cleaning everything and Close has some gaps.

PR #3180, opened 2020-07-07
Update CbsNode to use correct audience for nested Edge

richma-ms:
CbsNode was previously parsing the audience from the authentication token.  This is incorrect for nested Edge, where we'll receive the token from one of the parent Edge devices instead of the original connecting device.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
@@ -148,13 +148,15 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Amqp
 
             try
             {
-                string audience = message.ApplicationProperties.Map[CbsConstants.PutToken.Audience] as string;
-                return (token, audience);
+                SharedAccessSignature.Parse(iotHubHostName, token);
             }
             catch (Exception e)
             {
                 throw new InvalidOperationException("Cbs message does not contain a valid token", e);
             }
+
+            string audience = message.ApplicationProperties.Map[CbsConstants.PutToken.Audience] as string;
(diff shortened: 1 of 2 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -148,8 +148,8 @@ internal static (string token, string audience) ValidateAndParseMessage(string i
 
             try
             {
-                SharedAccessSignature sharedAccessSignature = SharedAccessSignature.Parse(iotHubHostName, token);
-                return (token, sharedAccessSignature.Audience);
+                string audience = message.ApplicationProperties.Map[CbsConstants.PutToken.Audience] as string;
```
varunpuranik, line 151: How does IoTHub do it? 
Is this fully backward compatible?

varunpuranik, line 151: Tests?

richma-ms: We've checked for the audience properties previously as well (but didn't use it).  It's set by the AMQP library when the SDK sends the authentication token, so I don't see a problem with backward compat.
https://github.com/Azure/azure-amqp/blob/master/src/Cbs/AmqpCbsLink.cs

IoTHub uses the authchain to figure out the connecting vs actor device, and use the API version to prevent compatbility issues.  Since EdgeHub doesn't do the same, this is the next best I could think of.

richma-ms: I updated the existing test collateral to use the audience ApplicationProperty.

PR #3181, opened 2020-07-07
[mqtt] Make edgehub authorizer to be aware of edge modules

dmolokanov:
Forces following patterns in topic translation and authorization
- $edgehub/{device_id}/{module_id}/{path}
- $edgehub/{device_id}/{path}


mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -140,34 +140,14 @@ fn is_iothub_topic(topic: &str) -> bool {
...
-        ),
-        format!("$edgehub/clients/{}/twin/get", client_id),
-        format!("$iothub/clients/{}/twin/get", client_id),
-        format!("$edgehub/clients/{}/twin/res", client_id),
-        format!("$iothub/clients/{}/twin/res", client_id),
-        format!("$edgehub/clients/{}/methods/post", client_id),
-        format!("$iothub/clients/{}/methods/post", client_id),
-        format!("$edgehub/clients/{}/methods/res", client_id),
-        format!("$iothub/clients/{}/methods/res", client_id),
+        format!("$edgehub/{}/messages/events", client_id),
```
varunpuranik, line 143: Just to confirm: we were using $iothub/$edgeHub to distinguish which topics need to be "translated" and which don't. We don't have $iothub anymore because we don't have any clients using those topics, right? 
Once we support the nested scenario with the bridge, we will have to support that?

mqtt/mqtt-edgehub/src/topic/translation.rs
```diff
@@ -176,6 +179,13 @@ macro_rules! translate_c2d {
 
 translate_d2c! {
     // Message Translation
+    module_send_message {
+        to_new_topic {
+            format!("devices/{}/modules/{}/messages/events(?P<path>.*)", DEVICE_ID, MODULE_ID),
```
varunpuranik, line 184: Instead of inlining the format, it seems like it will be cleaner to define the map in one central place.
Doesn't have to be done now though..

mqtt/mqtt-broker-core/src/auth/mod.rs
```diff
@@ -44,4 +49,29 @@ impl<T: Into<Identity>> From<T> for AuthId {
 }
 
 /// Non-anonymous client identity.
-pub type Identity = String;
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct Identity(Arc<String>);
```
varunpuranik, line 53: just curious: why this change?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
dmolokanov: Correct. Initially, I went ahead and "implemented" checks for a nested scenario which turned out to be done prematurely. Will return to it once mqtt topics are stabilized

mqtt/mqtt-edgehub/src/topic/translation.rs
dmolokanov: Yeah, I don't like implementation too much. Still could not come up with anything better.

mqtt/mqtt-broker-core/src/auth/mod.rs
dmolokanov: We are cloning `AuthId` a bunch of times during the client life time, so it should be don earlier.

PR #3185, opened 2020-07-07
New Build Images pipeline to support packaged edgehub

and-rewsmith:
The new edge hub will contain three components in the same container: Watchdog, Broker, and EdgeHub. Currently, our build images pipeline does not produce packaged images. This PR assumes that the watchdog rust application is present. As of now this is in final stages of review in another PR.

This PR contains the following:
- New build logic for the mqtt broker which will cross compile and lock down necessary dependencies (rust toolchain, openssl, cmake) on a per-platform basis. This is modeled after the edgelet build process defined in package.sh.
- Pipeline integration that will build and publish combined edgehub/broker images for each target platform.

**Pipeline Integration**
Since we decided not to build windows images for the broker, I have removed windows from the build and manifest steps.

For the broker and watchdog build I do the following:
- cross compile
- create a new folder to contain edgehub's published artifacts (watchdog, broker, c# dlls)
- copy these artifacts into this folder
- alter the edgehub dockerfile to add these artifacts to the image and start the watchdog.

This change will significantly increase the build time of our build images pipeline (60+ mins). We can later mitigate this by reducing the 6 rust build steps to 3 (consolidate broker and watchdog cross compilations), but it will still be long. The best way to mitigate this is to have a base image that gets regularly uploaded to our build container registry, but then we need to deal with maintaining compatibility for all feature branches. We can handle this outside this PR.

**New Build Logic for Broker/Watchdog**
We needed to come up with a new build process for the broker and the watchdog. The current approach did not provide flexibility in selecting rust toolchain, openssl, cmake versions. 

We modeled this new build logic after edgelet. See edgelet's package.sh for context on how edgelet cross compiles to lock down dependencies. This can mostly be replicated, but since there was no existing logic for alpine, we had to come up with something new. 

Regarding alpine, we don't need to build directly on alpine. Building on bionic will also work as long as it produces a musl binary. In order to build musl artifacts on bionic, I ripped out the relevant parts of [this dockerfile](https://github.com/emk/rust-musl-builder/blob/master/Dockerfile#L1) (line 38-70, 110-126, 145-155), which then get copied into the alpine image.

All the non-musl build logic has precedent in edgelet.


scripts/linux/cross-platform-rust-build.sh
```diff
@@ -0,0 +1,293 @@
+#!/bin/bash
```
and-rewsmith, line 1: Arnav recommended a separate script because there is a lot of edgelet specific logic in the edgelet build packaging.sh file.

scripts/linux/cross-platform-rust-build.sh
```diff
@@ -0,0 +1,293 @@
...
+
+case "$PACKAGE_OS.$PACKAGE_ARCH" in
+    ubuntu18.04.amd64)
+        # The below SETUP was copied from https://github.com/emk/rust-musl-builder/blob/master/Dockerfile.
+        # The beginning of the below setup specifies versions for other dependencies.
+        # Here are the places to check for new releases:
+        #
+        # - https://github.com/rust-lang/mdBook/releases
+        # - https://github.com/EmbarkStudios/cargo-about/releases
+        # - https://github.com/EmbarkStudios/cargo-deny/releases
```
and-rewsmith, line 105: We need to find a way to handle these important dependencies since they will get updated. Any ideas? We can elevate to the top of file for clarity, but that will separate it from the alpine logic.

builds/misc/templates/consolidate-edgehub-artifacts.yaml
```diff
@@ -0,0 +1,24 @@
...
+      mkdir -p $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/mqttd
+      mkdir -p $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/watchdog
+
+      cp  -r ${{ parameters['watchdog.artifacts.path'] }} $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/watchdog
+
+      cp  -r ${{ parameters['mqttd.artifacts.path'] }} $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/mqttd
+      cp ${{ parameters['mqttd.config.path'] }} $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/mqttd
+
+      mv  $(Build.BinariesDirectory)/${{ parameters['dotnet.artifacts.path'] }}/docker $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}
+      cp  -r $(Build.BinariesDirectory)/${{ parameters['dotnet.artifacts.path'] }} $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}
```
and-rewsmith, line 22: I used whitespace here to separate watchdog, mqtt, and edgehub logic.

scripts/linux/cross-platform-rust-build.sh
```diff
@@ -0,0 +1,293 @@
...
+
+set -e
+
+# Get directory of running script
+DIR="$(cd "$(dirname "$0")" && pwd)"
+
+BUILD_REPOSITORY_LOCALPATH="$(realpath "${BUILD_REPOSITORY_LOCALPATH:-$DIR/../..}")"
+
+CMAKE_ARGS='-DCMAKE_BUILD_TYPE=Release'
+CMAKE_ARGS="$CMAKE_ARGS -DBUILD_SHARED=On -Drun_unittests=Off -Duse_default_uuid=On -Duse_emulator=Off -Duse_http=Off"
```
darobs, line 63: You probably don't need `-DBUILD_SHARED=On` or `-Duse_default_uuid=On -Duse_emulator=Off` unless you are building the HSM library. You also don't want to set "BUILD_SHARED=On" if you are targeting MUSL.

edge-hub/docker/linux/amd64/Dockerfile
```diff
@@ -3,7 +3,11 @@ FROM azureiotedge/azureiotedge-runtime-base:1.2-linux-amd64 as builder
 
 FROM mcr.microsoft.com/dotnet/core/aspnet:${base_tag}
 
-ARG EXE_DIR=.	
+ADD ./watchdog/target/x86_64-unknown-linux-musl/release/watchdog /usr/local/bin/watchdog
+ADD ./mqttd/target/x86_64-unknown-linux-musl/release/mqttd /usr/local/bin/mqttd
+ADD ./mqttd/production.json /tmp/mqtt/config/production.json
```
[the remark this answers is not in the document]
and-rewsmith: I created /tmp/mqttd because the tmp directory is where the edgehub user has permission to write. The broker needs to write its state to this directory. Because I already had an mqttd folder, I chose to put the configuration here. Perhaps there is a better place?

builds/misc/templates/build-broker-watchdog.yaml
```diff
@@ -0,0 +1,33 @@
+steps:
```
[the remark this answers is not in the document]
and-rewsmith: Great idea I will look into this in a separate PR

builds/misc/images.yaml
```diff
@@ -22,7 +23,11 @@ jobs:
 
       - script: scripts/linux/buildBranch.sh -c $(Build.Configuration) --no-rocksdb-bin
         name: build
-        displayName: Build ($(Build.Configuration))
+        displayName: Build ($(Build.Configuration)) dotnet artifacts
+
+      - template: templates/build-broker-watchdog.yaml
```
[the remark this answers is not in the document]
and-rewsmith: I have a few higher priority tasks to finish then I will look into this next.

scripts/linux/cross-platform-rust-build.sh
and-rewsmith: Wait so these CMAKE_ARGS can be removed altogether. This build script doesn't need to use cmake for the broker. I should have removed it.

builds/misc/images.yaml
```diff
@@ -190,177 +195,6 @@ jobs:
...
-    workspace:
-      clean: all
-    steps:
-      - template: ../templates/install-dotnet2.yaml
-      - template: ../templates/install-dotnet3.yaml
-
-      - script: echo $(registry.password)|docker login "edgebuilds.azurecr.io" -u "$(registry.user)" --password-stdin
-        displayName: Docker Login
-
-      - powershell: scripts/windows/build/Publish-Branch.ps1 -Configuration:"$(Build.Configuration)" -PublishTests:$False -UpdateVersion
```
and-rewsmith, line 212: I will delete all the yaml and ps1 windows files in another PR. This will mean we also need to disable the checkin / CI for tests on windows. 

I need to confirm, but think our E2E tests should probably stay because we will be running linux containers on windows. We will need to alter the test pipeline to run these linux containers.

dylanbronson: Why did we decide not to build the broker for windows? Is this part of a greater agenda to move away from Windows and towards linux containers on windows?

edge-hub/docker/linux/amd64/Dockerfile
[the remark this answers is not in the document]
darobs: It's more up to the people who are working on the broker.  If you all are OK with it.

builds/misc/templates/consolidate-edgehub-artifacts.yaml
```diff
@@ -0,0 +1,24 @@
...
+  watchdog.artifacts.path: 'edge-hub/watchdog/target'
+  all.artifacts.path: 'publish/edge-hub'
+
+steps:
+  - bash: |
+      mkdir -p $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}
+      mkdir -p $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/mqttd
+      mkdir -p $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/watchdog
+
+      cp  -r ${{ parameters['watchdog.artifacts.path'] }} $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/watchdog
```
vadim-kovalyov, line 16: nit: not sure if it is possible or if it would work, but for readability you can try to define `variables` block after params and define vars such as `watchdog.dest.path` and similar.
```
variables:
  watchdog.dest.path: $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/watchdog
 ***
```

edge-hub/docker/linux/amd64/Dockerfile
```diff
@@ -22,14 +26,19 @@ COPY --from=builder publish/* /usr/local/lib/
 
 WORKDIR /app	
 
-COPY $EXE_DIR/ ./	
+COPY $EXE_DIR/ ./
+
+# Set broker log level
+ENV RUST_LOG=debug
```
vadim-kovalyov, line 32: this is temporary, right?

edge-hub/docker/linux/amd64/Dockerfile
```diff
@@ -22,14 +26,19 @@ COPY --from=builder publish/* /usr/local/lib/
...
 EXPOSE 5671/tcp	
 EXPOSE 443/tcp	
 
 USER edgehubuser	
 
-CMD echo "$(date --utc +"%Y-%m-%d %H:%M:%S %:z") Starting Edge Hub" && \	
-    exec /usr/bin/dotnet Microsoft.Azure.Devices.Edge.Hub.Service.dll
+CMD echo "$(date --utc +"%Y-%m-%d %H:%M:%S %:z") Starting Edge Hub" && \
+    cd /tmp/mqtt && \
+    /usr/local/bin/watchdog -c /tmp/mqtt/config/config.json
```
vadim-kovalyov, line 44: As discussed offline, we probably should not pass any mqttd related params to watchdog, instead have mqttd look for default config location.

scripts/linux/cross-platform-rust-build.sh
```diff
@@ -0,0 +1,278 @@
...
+esac
+
+if [ -n "$RUST_TARGET" ]; then
+    RUST_TARGET_COMMAND="rustup target add $RUST_TARGET &&"
+fi
+
+
+case "$PACKAGE_OS.$PACKAGE_ARCH" in
+    ubuntu18.04.amd64)
+        # The below SETUP was copied from https://github.com/emk/rust-musl-builder/blob/master/Dockerfile.
```
vadim-kovalyov, line 96: I think we should mention this in THIRDPARTYNOTICE file, since that repo is under MIT license.

scripts/linux/cross-platform-rust-build.sh
vadim-kovalyov: We don't need MDBOOK as far as I know, so it can be safely removed. I'm also not sure if we use Cargo About/Deny plugins. @dmolokanov any thoughts?

We also don't need `graphviz` package, as far as I know.

builds/misc/images.yaml
and-rewsmith: Yep that is the reason.

edge-hub/docker/linux/amd64/Dockerfile
and-rewsmith: I switched it to info

scripts/linux/cross-platform-rust-build.sh
and-rewsmith: Why does the musl-builder install these?

builds/misc/templates/consolidate-edgehub-artifacts.yaml
and-rewsmith: I tried adding a variables section before and got an error. I think it was because you can't use variables in a template?

scripts/linux/cross-platform-rust-build.sh
and-rewsmith: I added it. Do you know why though we have three different THIRDPARTYNOTICES files? I feel it would be better to just have one?

edge-hub/docker/linux/amd64/Dockerfile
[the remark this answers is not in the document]
and-rewsmith: I have changed this to read the config from /app

scripts/linux/cross-platform-rust-build.sh
vadim-kovalyov: I believe musl-builder tends to be a generic builder image that tries to satisfy many popular dependencies. We may not need all of them...

PR #3188, opened 2020-07-07
[mqtt] Ensure module-to-module topics translation and authz

dmolokanov:
Add topics translation rules for module-to-module communication 
Add corresponding authz rules

### DIFF
--- a/mqtt/mqtt-edgehub/src/topic/translation.rs
+++ b/mqtt/mqtt-edgehub/src/topic/translation.rs
@@ -78,7 +78,9 @@ pub fn translate_outgoing_publish(mut publish: proto::Publish) -> proto::Publish
 
 const DEVICE_ID: &str = r"(?P<device_id>[^/]+)";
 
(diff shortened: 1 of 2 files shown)


mqtt/mqtt-edgehub/src/topic/translation.rs
```diff
@@ -24,7 +24,7 @@ pub fn translate_incoming_subscribe(
     mut subscribe: proto::Subscribe,
 ) -> proto::Subscribe {
     for mut sub_to in &mut subscribe.subscribe_to {
-        if let Some(new_topic) = TRANSLATE_C2D.to_new_topic(&sub_to.topic_filter, client_id) {
+        if let Some(new_topic) = TRANSLATE_C2D.to_internal(&sub_to.topic_filter, client_id) {
```
varunpuranik, line 27: Will this to_internal / to_external be valid when we support the bridge scenario?

dmolokanov: My hope with new topics we don't need to convert anything

PR #3189, opened 2020-07-07
Enable twin encrypt by default

ancaantochi:
Enable twin encryption by default.
TwinStore is using EdgeTwin as entity name when it is not encrypted and underlyingEdgeTwin for the entity name when is encrypted. Because of that even if we it has implemented a way to update from uncrypted to encrypted data is actually not doing that because the store is different. I updated the code to use underlyingEdgeTwin  if it already exist and use EdgeTwin if underlyingEdgeTwin  was never created, that way data from unencrypted store is updated to encrypted while it is accessed.

Please note: if customer changes from encrypted to unencrypted data won't be recovered. In this case if edgeHub is set  to unencrypted and immediately after it goes offline it won't be able to get the configuration from store.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/twin/TwinStoreEntity.cs
```diff
@@ -27,7 +27,7 @@ public TwinStoreEntity(TwinCollection reportedPropertiesPatch)
         public TwinStoreEntity(Twin twin, TwinCollection reportedPropertiesPatch)
         {
             this.Twin = Option.Maybe(twin);
-            this.ReportedPropertiesPatch = reportedPropertiesPatch?.Count != 0
+            this.ReportedPropertiesPatch = reportedPropertiesPatch != null && reportedPropertiesPatch.Count != 0
```
varunpuranik, line 30: Do you have tests that cover this logic?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/StoreProvider.cs
```diff
@@ -25,10 +25,22 @@ public StoreProvider(IDbStoreProvider dbStoreProvider, TimeSpan operationTimeout
...
-            IKeyValueStore<TK, TV> dbStoreMapper = new KeyValueStoreMapper<TK, byte[], TV, byte[]>(entityDbStore, new BytesMapper<TK>(), new BytesMapper<TV>());
-            IEntityStore<TK, TV> entityStore = new EntityStore<TK, TV>(dbStoreMapper, entityName, 12);
-            IEntityStore<TK, TV> timedEntityStore = new TimedEntityStore<TK, TV>(entityStore, this.operationTimeout);
-            return timedEntityStore;
+            return this.GetEntityStore<TK, TV>(entityName, entityDbStore);
+        }
+
+        public IEntityStore<TK, TV> GetEntityStore<TK, TV>(string backwardCompatibleEntityName, string entityName)
+        {
+            Preconditions.CheckNonWhiteSpace(backwardCompatibleEntityName, nameof(backwardCompatibleEntityName));
```
varunpuranik, line 33: Do we need a test for this logic?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/StoreProvider.cs
```diff
@@ -25,10 +25,22 @@ public StoreProvider(IDbStoreProvider dbStoreProvider, TimeSpan operationTimeout
...
         {
             IDbStore entityDbStore = this.dbStoreProvider.GetDbStore(Preconditions.CheckNonWhiteSpace(entityName, nameof(entityName)));
-            IKeyValueStore<TK, TV> dbStoreMapper = new KeyValueStoreMapper<TK, byte[], TV, byte[]>(entityDbStore, new BytesMapper<TK>(), new BytesMapper<TV>());
-            IEntityStore<TK, TV> entityStore = new EntityStore<TK, TV>(dbStoreMapper, entityName, 12);
-            IEntityStore<TK, TV> timedEntityStore = new TimedEntityStore<TK, TV>(entityStore, this.operationTimeout);
-            return timedEntityStore;
+            return this.GetEntityStore<TK, TV>(entityName, entityDbStore);
+        }
+
+        public IEntityStore<TK, TV> GetEntityStore<TK, TV>(string backwardCompatibleEntityName, string entityName)
```
varunpuranik, line 31: I wonder if we should make this logic specific to the TwinStore..

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/StoreProvider.cs
ancaantochi: The tests from EntityStoreTestBase cover this, I added: BackwardCompatibleStoreName_ShouldReturnOlderEntityStore and BackwardCompatibleStoreName_ShouldReturnNewEntity

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/StoreProvider.cs
ancaantochi: I think the store provider should handle returning the store so it looks like more of a provider logic.  On the other way it's true that this is specific for twin because we need to get around having 2 names.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/twin/TwinStoreEntity.cs
ancaantochi: just added tests

PR #3190, opened 2020-07-08
Fix OnBehalfOf auth in EdgeHub

richma-ms:
With the change to get the correct audience, authentication now has the audience for the originating device, but the token for the actor device.  This change adds extra logic to validate the token against the actor device if we fail to authenticate against the originating device.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
```diff
@@ -82,12 +83,12 @@ async Task<(bool isAuthenticated, bool shouldFallback)> AuthenticateInternalAsyn
                 }
 
                 bool syncServiceIdentity = this.syncServiceIdentityOnFailure && !reauthenticating;
-                (bool isAuthenticated, bool valueFound) = await this.AuthenticateWithServiceIdentity(tCredentials, tCredentials.Identity.Id, syncServiceIdentity);
+                (bool isAuthenticated, bool valueFound) = await this.AuthenticateWithAuthChain(tCredentials, tCredentials.Identity.Id, syncServiceIdentity);
```
varunpuranik, line 86: Is this backward compatible?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
```diff
@@ -114,10 +114,16 @@ bool TryGetSharedAccessSignature(string token, IIdentity identity, out SharedAcc
...
 
-        bool ValidateCredentials(SharedAccessSignature sharedAccessSignature, ServiceIdentity serviceIdentity, IIdentity identity) =>
-            this.ValidateTokenWithSecurityIdentity(sharedAccessSignature, serviceIdentity) &&
-            this.ValidateAudience(sharedAccessSignature.Audience, identity) &&
-            this.ValidateExpiry(sharedAccessSignature, identity);
+        bool ValidateCredentials(SharedAccessSignature sharedAccessSignature, ServiceIdentity serviceIdentity, IIdentity identity, bool isOnBehalfOf)
+        {
+            // No need to check audience against the identity if it's an
+            // OnBehalfOf connection, they'll always be mismatched
+            bool hasValidAudience = isOnBehalfOf ? true : this.ValidateAudience(sharedAccessSignature.Audience, identity);
```
varunpuranik, line 121: We don't need to check audience for nested scenarios?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
```diff
@@ -102,25 +103,50 @@ async Task<(bool isAuthenticated, bool shouldFallback)> AuthenticateInternalAsyn
...
-            if (this.nestedEdgeEnabled && !authChain.HasValue)
+            // For nested Edge, we need to check that an authchain exists
+            Option<string> authChain = await this.deviceScopeIdentitiesCache.GetAuthChain(serviceIdentityId);
+            if (!authChain.HasValue)
             {
                 // The identity is not within our nested hierarchy
                 return (false, false);
             }
 
+            var authChainIds = authChain.OrDefault().Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries).ToList();
```
varunpuranik, line 125: Can you add some comments about the auth logic here...

varunpuranik:
So did on-behalf-of scenario never work?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
richma-ms, line 86: >(bool isAuthenticated, bool valueFound) = await this.AuthenticateWithAuthChain(tCredentials, tCredentials.Identity.Id, syncServiceIdentity); [](start = 16, length = 140)

maybe detect that this is an OnBehalfOf connection by checking the token audience against the identity, and fork the logic early

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
```diff
@@ -102,25 +103,50 @@ async Task<(bool isAuthenticated, bool shouldFallback)> AuthenticateInternalAsyn
...
+            if (!isAuthenticated && authChainIds.Count() > 2)
```
richma-ms, line 126: we should be getting the actor from the token's audience, then verifying that the actor is in the authchain.

philipktlin: I don't understand "authChainIds.Count() > 2"? why 2?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
```diff
@@ -102,25 +103,50 @@ async Task<(bool isAuthenticated, bool shouldFallback)> AuthenticateInternalAsyn
...
+            {
+                // Failed to authenticate against the originating identity, but the daisy-chained
+                // nature of nested Edge means the connection could be coming from our immediate
+                // child Edge, which acts as a parent to the originating identity. This "acting"
+                // Edge would be the second to last identity in the authchain.
+                string actorEdgeHubId = authChainIds[authChainIds.Count() - 2] + "/" + Constants.EdgeHubModuleId;
```
philipktlin, line 132: I don't understand why is "authChainIds.Count() - 2"?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
richma-ms: because there should always be at least 3 ids in the authchain for OnBehalfOf, [original]/[actor]/[parent].

But this doesn't apply anymore, I'm now just getting the actor out of the credential.

---
In reply to: [451892190](https://github.com/Azure/iotedge/pull/3190#discussion_r451892190) [](ancestors = 451892190)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
```diff
@@ -31,62 +31,158 @@ public class DeviceScopeTokenAuthenticator : DeviceScopeAuthenticator<ITokenCred
...
+                Events.InvalidAuthChain(targetId, authChain);
+                return false;
+            }
+
+            // The actor identity should be in the authChain
+            string actorId = actorDeviceId + actorModuleId.Map(id => "/" + id).GetOrElse(string.Empty);
+            bool targetAuthChainHasActor = false;
+            foreach (string id in authChainIds)
+            {
+                if (id == actorDeviceId)
```
richma-ms, line 96: >actorDeviceId [](start = 26, length = 13)

this should be actorId

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
```diff
@@ -102,25 +134,31 @@ async Task<(bool isAuthenticated, bool shouldFallback)> AuthenticateInternalAsyn
...
+        async Task<(bool isAuthenticated, bool serviceIdentityFound)> AuthenticateWithAuthChain(T credentials, string actorEdgeHubId, bool syncServiceIdentity)
         {
-            Option<string> authChain = await this.deviceScopeIdentitiesCache.GetAuthChain(serviceIdentityId);
-
-            if (this.nestedEdgeEnabled && !authChain.HasValue)
-            {
-                // The identity is not within our nested hierarchy
+            // For nested Edge, we need to check that we have
+            // a valid authchain for the target identity
+            Option<string> authChain = await this.deviceScopeIdentitiesCache.GetAuthChain(credentials.Identity.Id);
```
richma-ms, line 141: >GetAuthChain [](start = 77, length = 12)

We should check for Disabled state

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
```diff
@@ -81,13 +84,42 @@ async Task<(bool isAuthenticated, bool shouldFallback)> AuthenticateInternalAsyn
...
+                    tCredentials.Identity is IModuleIdentity moduleIdentity &&
+                    !actorModuleId.HasValue)
                 {
-                    // Module can use the Device key to authenticate
+                    // We're trying to authenticate a module, but the token is for a device.
+                    // This is the scenario where a module can use its parent device's token
+                    // to authenticate.
                     Events.AuthenticatingWithDeviceIdentity(moduleIdentity);
-                    (isAuthenticated, valueFound) = await this.AuthenticateWithServiceIdentity(tCredentials, moduleIdentity.DeviceId, syncServiceIdentity);
+                    (isAuthenticated, valueFound) = await this.AuthenticateWithServiceIdentity(tCredentials, moduleIdentity.DeviceId, syncServiceIdentity, Option.None<string>());
```
richma-ms, line 117: this needs to be failover still because when a module auth using the device's key, it's still using a module token.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
richma-ms: actually actorDeviceId is correct here, the authchain won't have the acting module ($edgeHub), only the deviceId.  The actorId local var is extraneous.

---
In reply to: [452482339](https://github.com/Azure/iotedge/pull/3190#discussion_r452482339) [](ancestors = 452482339)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
```diff
@@ -31,62 +31,157 @@ public class DeviceScopeTokenAuthenticator : DeviceScopeAuthenticator<ITokenCred
...
+            }
+
+            return true;
+        }
+
+        internal bool ValidateAudienceIds(string audienceDeviceId, Option<string> audienceModuleId, IIdentity identity)
+        {
+            string audienceId = audienceDeviceId + audienceModuleId.Map(id => "/" + id).GetOrElse(string.Empty);
+
+            if (!audienceModuleId.HasValue)
```
ancaantochi, line 115: nit: use Match

PR #3191, opened 2020-07-08
Update the base image for 1.0.9 arm32

yophilav:
Update the base image for 1.0.9 arm32 for security vulnerability [USN-4377-1](https://lists.ubuntu.com/archives/ubuntu-security-announce/2020-June/005457.html)
2.1.19-bionic-arm32v7 -> 1.0.4.1-linux-arm32v7

### DIFF
--- a/edge-agent/docker/linux/arm32v7/Dockerfile
+++ b/edge-agent/docker/linux/arm32v7/Dockerfile
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.4-linux-arm32v7
+ARG base_tag=1.0.6-linux-arm32v7
 FROM azureiotedge/azureiotedge-agent-base:${base_tag}
 
 ARG EXE_DIR=.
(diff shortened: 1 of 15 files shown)


edge-agent/docker/linux/arm32v7/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.4-linux-arm32v7
+ARG base_tag=1.0.6-linux-arm32v7
```
philipktlin, line 1: Just want to confirm that 1.0.5 is used in master and 1.0.10 branch when upgrade netcore to 3.1, right?

yophilav: That's correct.

yophilav:
> Did you do a global search for "-arm32v7" without any version information?

I did `base_tag` and manually goes through all of them making sure there is none left behind.

PR #3193, opened 2020-07-08
Remove windows testing from Image Builds, Checkin, CI, E2E

and-rewsmith:
We have made the choice to only support linux containers. The future for iotedge on windows platform will be linux containers on mariner, which is essentially linux containers on linux on windows. I talked to Varun discussing removal of all the windows checkin, ci, and e2e. We might later move to testing linux containers on windows, but this is undetermined. We might not need explicit windows platform tests to determine if this mariner scenario works.

I have removed all files supporting builds and tests for our windows containers, with the exception of the following:
- Windows [proxy deployment configuration](https://github.com/Azure/iotedge/blob/iiot/builds/e2e/proxy-deployment-template.json). There is a bit of coupling here with our proxy setup for linux and additional work is needed to uncouple these. We could do so here, but I chose to prioritize the work for MQTT Broker private preview. If any of these files are needed later we can pull them out of VC. 
- Iotedgectl related windows [files](https://github.com/Azure/iotedge/blob/iiot/scripts/windows/deploy/Deploy-Bootstrap.ps1). My understanding is that this tool is deprecated so these should have been removed already. There might be a reason that is not the case, so I have acted cautiously by not removing. If you disagree and think this and relevant files are safe to remove then I can do so.

For anyone that runs checkin tests locally on windows (guessing not due to the time this would take), this no longer works as we have removed the windows checkin gate and relevant files. I am going to open another PR after this to remove the windows dockerfiles.


doc/devguide.md
```diff
@@ -16,32 +18,20 @@ Make sure the following dependencies are installed in your environment before yo
...
 Besides using Visual Studio in Windows, you can build by running the build script:
 
-### Linux
 ```sh
 scripts/linux/buildBranch.sh
 ```
 
-### Windows
-```powershell
-scripts\windows\buildBranch.bat
```
vadim-kovalyov, line 26: Do we want to keep this for local dev on windows? How would the dev story look like on windows when we remove all windows support? does it mean that going forward we have to use WSL(2) for local dev?

and-rewsmith: I talked to Denis about this. People can still dev on windows as they do now. They just cannot run the checkins locally (did anyone even do this?).

I brought up keeping this and other windows files, but feel that it would be inconsistent with the new checkins we are adding (i.e. watchdog) which only have linux supported builds.

and-rewsmith:
@dmolokanov I am going to open another PR after this to remove the windows dockerfiles. No changes are required to edgelet windows builds/scripts as edgelet will still need to be built and tested on windows. We are removing the files in this PR because they relate to the windows container.

builds/misc/images-release.yaml
```diff
@@ -6,7 +6,7 @@ jobs:
     # This job is run on a hosted linux agent, which needs dotnet 2 and 3 installed as a manual step. 
     # The code sign steps will fail unless we explicitly say to use dotnet 2 before.
     # This means we have to toggle back and forth between primary dotnet installations for the sign and build.
-    # TODO: Investigate why we have to toggle primary installs on linux, but not on windows.
+    # TODO: Investigate why we have to toggle primary installs on linux, when we dind't have to do this on windows (now removed).
```
dylanbronson, line 9: nit: spelling

dylanbronson:
What kind of testing have you done on this PR? We should probably run all of our pipelines that used Windows to make sure we're not breaking anything.

Also as we talked about offline, we should do the same in master soon after doing it in iiot branch if no future releases will have Windows

But this is a great change

and-rewsmith:
@dylanbronson I ran new e2e, ci, longhaul, stress and they deployed like before except without windows. I will cherry-pick this to master once merged.

PR #3194, opened 2020-07-08
[k8s public preview] Experimental Service Options (#3179)

darobs:
Added another `k8s-experimental` section called `serviceOptions`. I has 2 fields, `type` and `LoadBalancerIP`.

`type` will be used as Service type if a Service is created.

`loadBalancerIP` is put into the `loadBalancerIP` field in the Service, if a service is created.

This allows individual modules to override default Service settings.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/service/KubernetesServiceMapper.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/service/KubernetesServiceMapper.cs
@@ -75,14 +75,21 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment.Service
                 //
                 // If the user wants to expose the ClusterIPs port externally, they should manually create a service to expose it.
                 // This gives the user more control as to how they want this to work.
-                var serviceType = onlyExposedPorts
-                    ? PortMapServiceType.ClusterIP
-                    : this.defaultMapServiceType;
+                var serviceType = this.GetServiceType(module, onlyExposedPorts);
+                var loadBalancerIP = module.Config.CreateOptions.ServiceOptions.FlatMap(so => so.LoadBalancerIP).OrDefault();
 
-                return Option.Some(new V1Service(metadata: serviceMeta, spec: new V1ServiceSpec(type: serviceType.ToString(), ports: servicePorts.Values.ToList(), selector: labels)));
+                return Option.Some(new V1Service(metadata: serviceMeta, spec: new V1ServiceSpec(type: serviceType.ToString(), loadBalancerIP: loadBalancerIP, ports: servicePorts.Values.ToList(), selector: labels)));
             }
 
             return Option.None<V1Service>();
+        }
+
+        PortMapServiceType GetServiceType(KubernetesModule module, bool onlyExposedPorts)
+        {
+            var serviceType = module.Config.CreateOptions.ServiceOptions.FlatMap(so => so.Type);
+            return serviceType.GetOrElse(() => onlyExposedPorts
+                    ? PortMapServiceType.ClusterIP
+                    : this.defaultMapServiceType);
         }
     }
 }
--- a/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesExperimentalPodParametersTest.cs
+++ b/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesExperimentalPodParametersTest.cs
@@ -4,6 +4,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test
     using System.Collections.Generic;
     using System.Linq;
     using k8s.Models;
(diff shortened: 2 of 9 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/service/KubernetesServiceMapper.cs
```diff
@@ -75,14 +75,21 @@ Option<V1Service> PrepareService(IModuleIdentity identity, KubernetesModule modu
...
+                var loadBalancerIP = module.Config.CreateOptions.ServiceOptions.FlatMap(so => so.LoadBalancerIP).OrDefault();
 
-                return Option.Some(new V1Service(metadata: serviceMeta, spec: new V1ServiceSpec(type: serviceType.ToString(), ports: servicePorts.Values.ToList(), selector: labels)));
+                return Option.Some(new V1Service(metadata: serviceMeta, spec: new V1ServiceSpec(type: serviceType.ToString(), loadBalancerIP: loadBalancerIP, ports: servicePorts.Values.ToList(), selector: labels)));
             }
 
             return Option.None<V1Service>();
         }
+
+        PortMapServiceType GetServiceType(KubernetesModule module, bool onlyExposedPorts)
```
CindyXing, line 87: What's the thought process behind this? Why can't we just let k8s handle it?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesExperimentalPodParametersTest.cs
```diff
@@ -338,9 +340,84 @@ public void ParsesSomeDeploymentStrategyExperimentalOptions()
...
+            parameters.ServiceOptions.ForEach(options => Assert.False(options.Type.HasValue));
+            parameters.ServiceOptions.ForEach(options => Assert.False(options.LoadBalancerIP.HasValue));
+        }
+
+        [Fact]
+        public void ParsesServiceOptions()
+        {
+            var experimental = new Dictionary<string, JToken>
+            {
+                ["k8s-experimental"] = JToken.Parse(@"{ serviceOptions: { type: ""nodeport"", loadbalancerip: ""100.1.2.3"" } }")
```
CindyXing, line 367: loadbalancerip is only valid when servicetype is loadbalancer. Why we honor it when service type is nodeport?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/service/KubernetesServiceMapper.cs
darobs: We have to choose a type for the service.  The translation logic in line 78 used to be:
                var serviceType = onlyExposedPorts
                    ? PortMapServiceType.ClusterIP
                    : this.defaultMapServiceType;

But with the addition of the service Options, that became more complicated to read, so I pulled this out into this subroutine. Now, the logic is, if the user specifies a service type, we use that. If the user does not, we use the old logic.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesExperimentalPodParametersTest.cs
darobs: Was taking the assumption that we take the data given to us "as is."  There does not seem to be a negative consequence to assigning it even if it is not appropriate, at least in my testing on AKS.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/service/KubernetesServiceMapper.cs
CindyXing: I see. We are facing the same situation as the volume.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesExperimentalPodParametersTest.cs
CindyXing: I see. What do you see with kubectl describe  the service? Do you see loadbalancerip field when servicetype is nodeport?

darobs: I have confirmed that a `NodePort` node works like a node port (was able to connect to service through assigned port). The `loadBalancerIP` field is ignored for `NodePort` services. Yes, it shows up in the spec:

```yaml
spec:
  clusterIP: 10.0.150.44
  externalTrafficPolicy: Cluster
  loadBalancerIP: 192.168.1.230
  ports:
  - name: exposedport-6379-tcp
    nodePort: 30570
    port: 6379
    protocol: TCP
    targetPort: 6379
  selector:
    net.azure-devices.edge.deviceid: device1
    net.azure-devices.edge.module: redisedge
  sessionAffinity: None
  type: NodePort
```

PR #3197, opened 2020-07-09
Adding Digest field to modules

ggjjj:
- Added digest field to Docker Module config
- Removed previous design of NotaryContentTrust
- updated tests cases to reflect the changes


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/CombinedDockerConfig.cs
```diff
@@ -9,11 +9,11 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker
 
     public class CombinedDockerConfig
     {
-        public CombinedDockerConfig(string image, CreateContainerParameters createOptions, Option<NotaryContentTrust> notaryContentTrust, Option<AuthConfig> authConfig)
+        public CombinedDockerConfig(string image, CreateContainerParameters createOptions, Option<string> digest, Option<AuthConfig> authConfig)
         {
             this.Image = Preconditions.CheckNonWhiteSpace(image, nameof(image)).Trim();
             this.CreateOptions = Preconditions.CheckNotNull(createOptions, nameof(createOptions));
-            this.NotaryContentTrust = notaryContentTrust;
+            this.Digest = digest;
```
CindyXing, line 16: What hash algorithm is to be used? Are we assuming SHA2?

ggjjj: DockerConfig is the desired properties of the modules. It gets populated with whatever is in the deployment manifest. So what the customer sets the value of digest of the image to be. usually docker container images use SHS256

CindyXing: Ok. We'll need to use the corresponding algorithm to calculate a digest against the downloaded payload; and then compare with this value. So the hash algorithm needs to be aligned. 

If we assume SHA2, then it should be the same. Or specify the algorithm in the config as well.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker
     using Microsoft.Azure.Devices.Edge.Util;
     using Newtonsoft.Json;
     using Newtonsoft.Json.Linq;
+    using Org.BouncyCastle.Asn1.X509;
```
varunpuranik, line 11: Where is this being used?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/CombinedDockerConfig.cs
ggjjj: The digest field would include the SHA2 algorithm along with it. For Example : The complete name would be **image-name@sha256:54f12e6b6ce**, then the digest that would be specified would be **sha256:54f12e6b6ce**
We wont calculate the digest in the device. This field will be used just to compare the digest received via the notary lookup command. So need to specify a separate field in config.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
ggjjj: thanks. will remove.

PR #3200, opened 2020-07-09
Add watchdog to checkin and consolidate replicated files

and-rewsmith:
Since we added the watchdog we need to enable the build, format, and clippy in the checkin gate. I have discussed with Vadim and we think this is a good opportunity to consolidate build scripts for mqtt / edgelet that have nearly identical logic. Thus, I have consolidated the rust install, build, and format scripts under the path /scripts/linux/generic-rust/[name].sh.

The watchdog checkin is not enabled right now because the yaml needs to exist in the iiot branch first. Once it is merged I will create the checkin pipeline, verify it works, then add it to the gate.


builds/checkin/watchdog.yaml
```diff
@@ -0,0 +1,56 @@
...
+  ################################################################################
+    displayName: Style Check
+    dependsOn: check_run_pipeline
+    condition: eq(dependencies.check_run_pipeline.outputs['check_files.RUN_PIPELINE'], 'true')
+    pool:
+      vmImage: "ubuntu-16.04"
+    steps:
+      - bash: scripts/linux/generic-rust/install.sh --project-root "edge-hub/watchdog"
+        displayName: Install Rust
+      - bash: scripts/linux/generic-rust/format.sh --project-root "edge-hub/watchdog"
```
and-rewsmith, line 53: reminder: delete other format / install scripts when we have verified new ones work

builds/checkin/mqtt.yaml
```diff
@@ -34,10 +34,12 @@ jobs:
...
         displayName: Set env variables
-      - bash: mqtt/build/linux/install.sh
+      - bash: scripts/linux/generic-rust/install.sh --project-root "mqtt"
         displayName: Install Rust
-      - bash: mqtt/build/linux/build.sh
-        displayName: Build
+      - bash: scripts/linux/generic-rust/build.sh --project-root "mqtt" --packages "mqttd/Cargo.toml" --manifest-path --no-default-features
+        displayName: Build with no default features
+      - bash: scripts/linux/generic-rust/build.sh --project-root "mqtt" --packages "mqttd/Cargo.toml" --manifest-path
+        displayName: Build with default features
```
and-rewsmith, line 42: two builds to cover default features

builds/checkin/watchdog.yaml
```diff
@@ -0,0 +1,54 @@
+trigger: none
+pr:
+  branches:
+    include:
+      - iiot
```
and-rewsmith, line 5: We will add triggers when we port the change to master.

edge-hub/watchdog/build/clippy.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
```
and-rewsmith, line 1: note: This diff is confusing. This is not a rename.

builds/checkin/watchdog.yaml
```diff
@@ -0,0 +1,54 @@
...
+    steps:
+      - script: echo "##vso[task.setvariable variable=RUST_BACKTRACE;]1"
+        displayName: Set env variables
+      - bash: scripts/linux/generic-rust/install.sh --project-root "edge-hub/watchdog"
+        displayName: Install Rust
+      - bash: scripts/linux/generic-rust/build.sh --project-root "edge-hub/watchdog" --packages "watchdog"
+        displayName: Build
+
+  ################################################################################
+  - job: style_check
```
vadim-kovalyov, line 41: Btw, do we need to build this for arm? I think a separate style check job made sense when we build for multiple platforms (linux/windows, since there could be conditional code paths). Otherwise you can combine everything in one job.

edge-hub/watchdog/rust.toolchain
```diff
@@ -0,0 +1 @@
+1.43.1
```
vadim-kovalyov, line 1: it should be fine for watchdog to be on `stable` toolchain, b/c it does not use async. We use 1.43 for mqttd only to work around clippy issue with .await.

builds/checkin/watchdog.yaml
and-rewsmith: The checkin gates usually just build on amd64 since that is the arch of the agent. I think combining these is a good idea and will do so.

and-rewsmith: Unfortunately, I cannot consolidate the jobs in the mqtt / edgelet / dotnet checkins because my other PR to remove all windows files is still under review. I will do so only for watchdog here then open a separate PR to clean up the others.

PR #3201, opened 2020-07-09
Get cloud connection if adding device with subscriptions

dylanbronson:
We are seeing an issue where a module using the nodeJS moduleClient with MQTT and only subscribing to DirectMethods (no telemetry) is unable to receive any direct methods after ~45 minutes. 

This happens because the module gets reauthenticated. With MQTT, the device connection must be dropped momentarily during reauth. When it goes back online, the ConnectionManager adds the DeviceConnection, but does not add the CloudConnection. Because there is no telemetry or twin actions, the cloud connection never gets established, and subscriptions for the device are never processed.

This PR will get the cloud connection if the device has any subscriptions when the device is re-added after reauth. 
I've tested it with the nodeJS SDK with success. 

I assume that the reason we weren't seeing this with C# SDK is that the C# SDK (incorrectly) resubscribes in this reauth case. It shouldn't need to resubscribe, though.


edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/ConnectionManagerTest.cs
```diff
@@ -733,6 +761,8 @@ public async Task KeepSubscriptionsOnDeviceRemoveTest()
             Assert.Equal(2, subscriptions.Count);
             Assert.True(subscriptions[DeviceSubscription.Methods]);
             Assert.True(subscriptions[DeviceSubscription.C2D]);
+            Option<ICloudProxy> cloudProxy = await connectionManager.GetCloudConnection(deviceId);
+            Assert.True(cloudProxy.HasValue);
```
dylanbronson, line 765: I would love to also prove that after the initial AddDeviceConnection on line 727, we DON'T have a CloudConnection for that device. I'm not so sure how to do that, since GetCloudConnection will create one.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
```diff
@@ -68,6 +68,14 @@ public async Task AddDeviceConnection(IIdentity identity, IDeviceProxy devicePro
             Preconditions.CheckNotNull(deviceProxy, nameof(deviceProxy));
             ConnectedDevice device = this.GetOrCreateConnectedDevice(identity);
             Option<DeviceConnection> currentDeviceConnection = device.AddDeviceConnection(deviceProxy);
+            currentDeviceConnection.ForEach(async c =>
```
ancaantochi, line 71: I think this won't work if edgeHub restarts because device connection is not stored.

dylanbronson: If edgeHub restarts, I believe we will resubscribe anyway through a different path, so we won't need to do this explicitly

dylanbronson: We see this does work on edgeHub restarts. I added a log statement in SessionStatePersistenceProvider where we see the subscription getting kicked off after an EH restart

varunpuranik, line 71: Change looks good. Can you add a comment for why we are doing this..

PR #3202, opened 2020-07-10
Base image update for ARM32 and ARM64 platforms for security vulnerability [USN-4377-1]

yophilav:
Base image update for ARM32 and ARM64 platforms for security vulnerability [USN-4377-1](https://lists.ubuntu.com/archives/ubuntu-security-announce/2020-June/005457.html) with iotedge non-root user
3.1.5-bionic-arm32v7 -> 1.0.6.0-linux-arm32v7
3.1.5-bionic-arm64v8 -> 1.0.6.0-linux-arm64v8

(Cherry-pick from https://github.com/Azure/iotedge/commit/a9ce4ce641c87abad18fa85ef5f1538542b76b1e)


test/connectivity/modules/NetworkController/docker/linux/arm64v8/Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG base_tag=1.0.5-linux-arm64v8
-FROM azureiotedge/azureiotedge-module-base:${base_tag}
```
varunpuranik, line 2: Was this never working before?

test/connectivity/modules/NetworkController/docker/linux/arm64v8/Dockerfile
```diff
@@ -1,5 +1,5 @@
...
+ARG base_tag=1.0.5.1-linux-arm64v8
+FROM edgebuilds.azurecr.io/microsoft/azureiotedge-module-base-full:${base_tag}
```
varunpuranik, line 2: WHy this change?

yophilav: We publish images to 2 different places. I talked with Anca and it seemed like the NetworkController pull its base image from a different CR than the rest of the test modules. This fix consolidates that.

yophilav:
> We use azureiotedge/azureiotedge-module-base:{basetag} in some places and edgebuilds.azurecr.io/microsoft/azureiotedge-module-base-full:${base_tag} in other places. What is the logic?

Currently on the NetworkController is the only one that uses `azureiotedge/azureiotedge-module-base:{basetag}`. That is by mistake. This PR fixes the issue.

philipktlin:
> > We use azureiotedge/azureiotedge-module-base:{basetag} in some places and edgebuilds.azurecr.io/microsoft/azureiotedge-module-base-full:${base_tag} in other places. What is the logic?
> 
> Currently on the NetworkController is the only one that uses `azureiotedge/azureiotedge-module-base:{basetag}`. That is by mistake. This PR fixes the issue.

NetworkController ues azureiotedge-module-base-full in Linux OS (Test Analyzer and TestResultCoordinator as well).  There are many modules using azureiotedge/azureiotedge-module-base, for example, tempSensor, metricCollector, CloudToDeviceMessageTester, DeploymentTester, DirectMethodRecevier, DirectMethodSender, EdgeHubRestartTester, LoadGen, MetricValidator, ModuleRestarter, Relayer, and TwinTeser.

test/connectivity/modules/NetworkController/docker/linux/arm64v8/Dockerfile
philipktlin: This image is not used in any place as Connectivity test is configured to run on Linux amd64 and arm32.   Agreed, this should be fixed.

and-rewsmith:
@varunpuranik @philipktlin @yophilav The difference between the module-base images is that azureiotedge base images are publicly available, whereas edgebuilds base images are not. I was the one to do this last october because we needed base images only for our tests, and I felt it didn't make sense to put these test-specific images on the public facing dockerhub.

yophilav:
> @varunpuranik @philipktlin @yophilav The difference between the module-base images is that azureiotedge base images are publicly available, whereas edgebuilds base images are not. I was the one to do this last october because we needed base images only for our tests, and I felt it didn't make sense to put these test-specific images on the public facing dockerhub.

We publish our base images to both CR be default. Ideally no matter which CR you use, it should result in the same final output. Just for the sake of consistency, which CR should be picked for our build pipeline to pull to create these images?

yophilav:
Given the current situation with our base image CR security scan, I'd like to merge this PR solely just for the ARM32 & ARM64 base image. The source of CR will be addressed in an upcoming PR.

test/modules/load-gen/docker/linux/arm32v7/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.5-linux-arm32v7
+ARG base_tag=1.0.6.0-linux-arm32v7
 FROM azureiotedge/azureiotedge-module-base:${base_tag}
```
philipktlin, line 2: Please confirm with Varun and Damon if it is ok to move all base images in EdgeBuilds CR; if so, create another PR.  thanks.

PR #3205, opened 2020-07-10
Add edgeAgentUser to ARM32 & ARM64

yophilav:
Add edgeAgentUser to ARM32 & ARM64

### DIFF
--- a/edge-agent/docker/linux/arm32v7/base/Dockerfile
+++ b/edge-agent/docker/linux/arm32v7/base/Dockerfile
@@ -7,4 +7,8 @@ RUN apt-get update && \
     ln -s /lib/arm-linux-gnueabihf/libc.so.6 /usr/lib/arm-linux-gnueabihf/libc.so && \
     rm -rf /var/lib/apt/lists/*
 
+# create a user to allow agent to optionally run as non-root
+ARG EDGEAGENTUSER_ID=1000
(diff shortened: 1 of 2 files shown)


darobs:
> Can you add some details on why this is needed now?

We may not need this for 1.0.10. 

This is required for Edge on Kubernetes - admission policies on locked down clusters require the containers run as non-root.  Creating this user allows the edgeAgent to successfully run as a non-root user.  It shouldn't have an effect on single device edge runtimes.

yophilav:
> Can you add some details on why this is needed now?

It is a subset of changes that @darobs introduced to master here: 
https://github.com/Azure/iotedge/commit/10cf7779f6999da5a466e6f87bc1736ed9ad19ed#diff-845ddd11d19616c8462d5016e1ced74d

yophilav:
> > Can you add some details on why this is needed now?
> 
> We may not need this for 1.0.10.
> 
> This is required for Edge on Kubernetes - admission policies on locked down clusters require the containers run as non-root. Creating this user allows the edgeAgent to successfully run as a non-root user. It shouldn't have an effect on single device edge runtimes.

@darobs Should 1.0.10 and master branch share the same base image with the non-root user? 
Or do you think 1.0.10 and master should have their own base image version granted 1.0.10 won't have the non-root user, but the master version will.

PR #3210, opened 2020-07-13
Added bridge functionality to connect edgeHub and MQTT Broker

vipeller:


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/MqttBrokerModule.cs
```diff
@@ -0,0 +1,92 @@
...
+            // 2) by every interface name that is not the following:
+            //      IMessageConsumer/IProducer/ISubscriber/ISubscriptionWatcher.
+            //    That is because some of the producers/consumers are also special services
+            //    and components will need those injected by custom interfaces.
+            builder.RegisterTypes(componentTypes)
+                   .AsSelf()
+                   .As(t => GetNonStandardBridgeInterfaces(t))
+                   .SingleInstance();
+
+            builder.RegisterType<DeviceProxy>()
```
varunpuranik, line 43: Why do we need this?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/MqttComponentDiscovery.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public MqttBridgeComponentDiscovery(ILogger logger)
+        {
+            this.logger = Preconditions.CheckNotNull(logger);
+
+            this.Producers = new IMessageProducer[0];
+            this.Consumers = new IMessageConsumer[0];
+        }
+
+        public void Discover(IComponentContext context)
```
varunpuranik, line 28: Make this static.. so that we can make the class immutable.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/MqttBrokerModule.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                            return new MqttBrokerProtocolHead(config, connector);
+                        })
+                    .As<MqttBrokerProtocolHead>()
+                    .SingleInstance();
+
+            base.Load(builder);
+        }
+
+        static IEnumerable<Type> GetNonStandardBridgeInterfaces(Type type) => type.GetInterfaces().Where(t => !MqttBridgeComponentDiscovery.CandidateInterfaces.Contains(t));
```
varunpuranik, line 90: THis seems a bit hacky. What types implement multiple interfaces that need to be discovered?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/MqttBrokerModule.cs
```diff
@@ -0,0 +1,92 @@
...
+        public MqttBrokerModule(IConfiguration config)
+        {
+            this.config = Preconditions.CheckNotNull(config, nameof(config));
+        }
+
+        protected override void Load(ContainerBuilder builder)
+        {
+            var componentTypes = MqttBridgeComponentDiscovery.GetCandidateTypes().ToArray();
+
+            // The classes will be registered by two types:
```
varunpuranik, line 30: This logic seems interesting to me. We should discuss this.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/MqttComponentDiscovery.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Autofac;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public class MqttBridgeComponentDiscovery : IComponentDiscovery
```
varunpuranik, line 11: Nit: align the file name and type name

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+        Option<MqttClient> mqttClient;
+
+        public MqttBrokerConnector(IComponentDiscovery components, ISystemComponentIdProvider systemComponentIdProvider)
+        {
+            this.components = Preconditions.CheckNotNull(components);
+            this.systemComponentIdProvider = Preconditions.CheckNotNull(systemComponentIdProvider);
+
+            // because of the circular dependency between MqttBridgeConnector and the producers,
+            // in this loop the producers get the IMqttBridgeConnector reference:
+            foreach (var producer in components.Producers)
```
varunpuranik, line 36: So, is the circular dependency because of IComponentDiscovery? Why can't the producer types take in a IMqttBrokerConnector? We don't seem to be using the producers anywhere else here.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+            var client = this.mqttClient.Expect(() => new Exception("No mqtt-bridge connector instance found to send messages."));
+
+            var added = default(bool);
+            var tcs = new TaskCompletionSource<bool>();
+
+            // need the lock, otherwise it can happen the the ACK comes back sooner as the id is
+            // put into the dictionary next line, causeing the ACK being unknown.
+            lock (this.guard)
+            {
+                var messageId = client.Publish(topic, payload, 1, false);
```
varunpuranik, line 151: This is sync?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Channels;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using uPLibrary.Networking.M2Mqtt;
```
varunpuranik, line 12: We should hide the external dependency behind interfaces.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/MqttBrokerModule.cs
vipeller: The registrations happen for the following uses:
- the component discovery will collect all types implementing IMessageConsumer, etc, then resolves them by type (e.g. it has its in list that TwinHandler implements IMessageConsumer, so it will ask autofac for a TwinHandler instance. That is why these types are registered "AsSelf" (e.g. as TwinHandler)
- there are types that use these classes (like TwinHandler) as special services. They know them from an interface, like ITwinHandler (e.g the DeviceProxy class gets ITwinHander injected. That is why every class is registered all of their implemented interfaces, which is not the usual "IMessageConsumer", etc. A class cannot be registered by "IMessageConsumer" because there are several classes implementing them and it would be ambiguous to resolve by them. What happens here is the the "GetNosStandardBridgeInterfaces()" filters out "IMessageConsumer", but leaves there eg "ITwinHandler"

As a result, every "normal" dependency injection use cases (like a constructor parameter receiving an ITwinHandler) can be resolved, but the discovery class also can resolve them by specific type (e.g TwinHandler, because discovery will see that the TwinHandler class is  one implementing IMessageConsumer, so it will request for TwinHandler to put it into its "consumers" list)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/MqttBrokerModule.cs
vipeller: There are a couple of them. As I implemented handlers as separate classes, on of the 'not-so-good' side effect is that from the point of the DeviceProxy class, every operation is in a different class. Because of that, there are 4 classes injected to it (different handlers). These classes are registered by interfaces like "ITwinHandler or "ICloud2DeviceMessageHandler". In the same time, they must not be registered as "IMessageConsumer/Producer" - that is the filtering happening in the line above.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: Some classes are producers and consumers in the same time. The MqttBrokerConnector needs the consumers (as it calls them with "handle()"), so the consumers need to be injected to the MqttBrokerConnector class (that happens trough the discovery class). These consumer classes, when they are also producers (e.g TwinHandler), need to get a reference to MqttBrokerConnector.

Simpler - TwinHandler needs a reference to MqttBrokerConnector as it needs to send messages out, but MqttBrokerConnector also needs a reference to TwinHandler, so it can call its "HandleAsync" method when a message comes in.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: Yes, the library has sync interface, but in the background it just puts it into a queue, so it is pretty fast

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: I was considering it, but coming up a "general" library just to hide the concrete type did not seem worth it. First, the library has a quite narrow interface to use, second probably I could not create an interface that will fit exactly to the next library. So with the next library some changes need to be done anyway, and I thought it means less work to make the change in this class. (also meaning that the broker connector is a sort of abstraction toward the rest of the classes - supporting send/receive operations)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+                                            Events.FailedToForward(e);
+                                            continue;
+                                        }
+
+                                        foreach (var consumer in this.components.Consumers)
+                                        {
+                                            try
+                                            {
+                                                var accepted = await consumer.HandleAsync(publishInfo);
+                                                Events.MessageForwarded(consumer.GetType().Name, accepted, publishInfo.Topic, publishInfo.Payload.Length);
```
varunpuranik, line 264: Won't this flood your logs?

varunpuranik: Why not break the loop if a message is accepted

vipeller: I can change it to list only those accepted - or if it was not accepted by anything

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+    {
+        const int ReconnectDelayMs = 2000;
+
+        readonly IComponentDiscovery components;
+        readonly ISystemComponentIdProvider systemComponentIdProvider;
+        readonly Dictionary<ushort, TaskCompletionSource<bool>> pendingAcks = new Dictionary<ushort, TaskCompletionSource<bool>>();
+
+        readonly object guard = new object();
+
+        Option<Channel<MqttPublishInfo>> publications;
```
varunpuranik, line 25: Why are publications and forwardingLoop options?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: originally subscription changes were messages multiple consumers processed. Factoring that out into a separate class, maybe there is no more such message, so the loop could be broken. I need to check that

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
varunpuranik: If so, can we just get rid of the Discovery class and have a list of producers and list of consumers?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: because when I implemented the auth-endpoint with a similar pattern but using "NULL" instead of none until it is not started, I was told to use Option :)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: that would not solve the circular dependency I think. When TwinHandler is instantiated, MqttBrokerConnection must already be an instance (so it can be injected into TwinHandler). But when MqttBrokerConnection is instantiated, it needs the consumer instances (which TwinHander is), so coming from this way, TwinHandler must be created first. Something must be created first, and the other thing needs an additional method call to obtain the first one.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+            this.publications = Option.Some(Channel.CreateUnbounded<MqttPublishInfo>(
+                                    new UnboundedChannelOptions
+                                    {
+                                        SingleReader = true,
+                                        SingleWriter = true
+                                    }));
+
+            this.forwardingLoop = Option.Some(this.StartForwardingLoop());
+
+            // if ConnectAsync is supposed to manage starting it with broker down,
```
varunpuranik, line 72: Yes, seems like we should reuse the logic from Trigger reconnect

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+
+                lock (this.guard)
+                {
+                    this.mqttClient = Option.None<MqttClient>();
+                }
+
+                throw new Exception("Failed to start mqtt-bridge connector");
+            }
+
+            client.ConnectionClosed += this.TriggerReconnect;
```
varunpuranik, line 91: How will the connection recover if it goes down before we sub to this event?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+
+                    client = this.mqttClient.Expect(() => new Exception("No mqtt-bridge connector instance found to use"));
+                }
+
+                // don't trigger it to ourselves while we are trying
+                client.ConnectionClosed -= this.TriggerReconnect;
+
+                var isConnected = false;
+
+                while (!isConnected)
```
varunpuranik, line 217: Will this spin if the broker goes down? Do we need backoff retry?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+            });
+        }
+
+        Task StartForwardingLoop()
+        {
+            var loopTask = Task.Run(
+                                async () =>
+                                {
+                                    Events.ForwardingLoopStarted();
+                                    while (await this.publications.Expect(ChannelIsBroken).Reader.WaitToReadAsync())
```
varunpuranik, line 245: I would avoid making publications an option and init it in the ctor.. 
but if it is an option, then take it into a local variable to avoid doing multiple Expect calls..

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+            // if ConnectAsync is supposed to manage starting it with broker down,
+            // put a loop here to keep trying - see 'TriggerReconnect' below
+            var isConnected = await TryConnectAsync(client, this.components.Consumers, this.systemComponentIdProvider.EdgeHubBridgeId);
+
+            if (!isConnected)
+            {
+                client.MqttMsgPublished -= this.ConfirmPublished;
+                client.MqttMsgPublishReceived -= this.ForwardPublish;
+
+                await this.StopForwardingLoopAsync();
```
varunpuranik, line 81: I feel we can simplify this logic quite a bit by making this class non-resumable. The protocol head can just create a new connector if it wants to restart it.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
varunpuranik: Yes, no nulls pls, but here I think we can init these 2 in the ctor

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
varunpuranik: Hmm, but that seems to stem from the fact that we use the same class to implement IMessageProducer and IMessageConsumer.
Its a bit philosophical, but the SetConnector on the IMessageProducer doesn't seem necessary.. in fact feels like we don't need the IMessageProducer interface at all. 
But its okay for now, let's think about it later.

vipeller: The two interfaces can be merged, but not all producers are consumers and not all consumers are producers. Because of that, with a merged interface, we need to give "not implemented" implementations (either doing nothing or throwing an error). That is a typical sign of breaking the "interface segregation principle" from the solid principles - that is why am a bit against merging them. (I was crying deleting ISubscriber, but that at least was well aligned with IMessageConsumer) 

but even if it is merged, in the case of classes like TwinManager, I don't think that is solves the circular problem: the TwinHandler still needs an instance of the MqttBrokerConnector (because it uses it for sending messages), and the MqttBrokerConnector needs an instance of the TwinManager (because it wants to give incoming messages to it for processing) - so the circular reference is still there, making SetConnector a necessity.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: That can be done (by turning the class "not-restartable") - but for me it seems strange for a class that has a method to kick-in its function (the ConnectAsync()), especially the forwarding loop, which is sort of heavy for a constructor.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/TwinHandler.cs
```diff
@@ -0,0 +1,354 @@
...
+    using Microsoft.Extensions.Logging;
+
+    public class TwinHandler : ITwinHandler, IMessageConsumer, IMessageProducer
+    {
+        const string TwinGetDevice = "$edgehub/+/twin/get/#";
+        const string TwinGetModule = "$edgehub/+/+/twin/get/#";
+        const string TwinUpdateDevice = "$edgehub/+/twin/reported/#";
+        const string TwinUpdateModule = "$edgehub/+/+/twin/reported/#";
+
+        const string TwinGetPublishPattern = @"^\$edgehub/(?<id1>[^/\+\#]+)(/(?<id2>[^/\+\#]+))?/twin/get/\?\$rid=(?<rid>.+)";
```
varunpuranik, line 22: Nit: can we call it deviceId and moduleId (instead of id1 and id2). Same elsewhere.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/TelemetryHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+            var identity = id2.Success
+                                ? this.identityProvider.Create(id1.Value, id2.Value)
+                                : this.identityProvider.Create(id1.Value);
+
+            var maybeProxy = await this.connectionRegistry.GetDeviceListenerAsync(identity);
+            var proxy = default(IDeviceListener);
+
+            try
+            {
+                proxy = maybeProxy.Expect(() => new Exception($"No device listener found for {identity.Id}"));
```
varunpuranik, line 67: Nit: can we call this deviceListener (proxy confused me :))

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/TwinHandler.cs
```diff
@@ -0,0 +1,354 @@
...
+        const string DesiredUpdateModule = "$edgehub/{0}/{1}/twin/desired/?$version={2}";
+
+        static readonly string[] subscriptions = new[] { TwinGetDevice, TwinGetModule, TwinUpdateDevice, TwinUpdateModule };
+
+        static readonly SubscriptionPattern[] subscriptionPatterns = new SubscriptionPattern[]
+                                                                         {
+                                                                            new SubscriptionPattern(TwinSubscriptionForResultsPattern, DeviceSubscription.TwinResponse),
+                                                                            new SubscriptionPattern(TwinSubscriptionForPatchPattern, DeviceSubscription.DesiredPropertyUpdates)
+                                                                         };
+        readonly Channel<ProcessingInfo> notifications;
```
varunpuranik, line 41: It is interesting that we need the Channel pattern for the twins (which is not expected to be high throughput fast messaging communication). How slow is the SendGetTwin and UpdateReportedProperties?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Channels;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using uPLibrary.Networking.M2Mqtt;
+    using uPLibrary.Networking.M2Mqtt.Messages;
+
+    public class MqttBrokerConnector : IMqttBrokerConnector
```
varunpuranik, line 15: Lot of logic in this class.. we should add tests for it as well.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -0,0 +1,423 @@
...
+                        throw new Exception("MQTT server did not grant QoS-1 for every requested subscription");
+                    }
+
+                    void ConfirmSubscribe(object sender, MqttMsgSubscribedEventArgs e)
+                    {
+                        if (Interlocked.Decrement(ref expectedAckCount) == 0)
+                        {
+                            Volatile.Write(ref allQosGranted, e.GrantedQoSLevels.All(qos => qos == 1));
+
+                            acksArrived.Release();
```
vadim-kovalyov, line 375: What if we never receive an ack? Will this stuck? We can try to use CancellationTokenSource with timeout instead of SemaphoreSlim.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ConnectionHandler.cs
```diff
@@ -0,0 +1,297 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Hub.Core;
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Device;
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Identity;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Azure.Devices.Edge.Util.Concurrency;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    public class ConnectionHandler : IConnectionRegistry, IMessageConsumer
```
vadim-kovalyov, line 18: I wish we can add some comments on every major components that describes it's purpose. It would be much easier for the future generation to understand what's going on :)

vipeller: I can add (also there is a separate doc describing the whole working - bit outdated, but I can transform it into a text or md with the updates

PR #3216, opened 2020-07-13
[PnP] Add ModelIdStore for EdgeHub to store modelId per device identity

dylanbronson:
This PR adds the ModelIdStore for EdgeHub. This store will store modelIds per device identity.

Every time an identity (module or device) connects to edgeHub with a modelId, we will store the modelId in this store.
Currently, we only implement this change for MQTT.

We will then get the modelId when connecting to upstream and use it to create the device or module client with the modelId if it exists. This will register the identity as PnP in the PnP service.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ClientProvider.cs
```diff
@@ -82,5 +82,29 @@ public IClient Create(IIdentity identity, ITokenProvider tokenProvider, ITranspo
...
         }
+
+        public IClient Create(IIdentity identity, ITokenProvider tokenProvider, ITransportSettings[] transportSettings, string modelId)
+        {
+            Preconditions.CheckNotNull(identity, nameof(identity));
+            Preconditions.CheckNotNull(transportSettings, nameof(transportSettings));
+            Preconditions.CheckNotNull(tokenProvider, nameof(tokenProvider));
+            var options = new ClientOptions
+            {
+                ModelId = modelId,
```
richma-ms, line 93: >modelId [](start = 26, length = 7)

need to check null/empty?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -78,6 +81,8 @@ public async Task<Try<ICloudConnection>> Connect(IClientCredentials clientCreden
                 Events.CreatingCloudConnectionUsingClientCredentials(clientCredentials);
                 var cloudListener = new CloudListener(this.edgeHub.Expect(() => new InvalidOperationException("EdgeHub reference should not be null")), clientCredentials.Identity.Id);
                 string productInfo = await this.productInfoStore.GetEdgeProductInfo(clientCredentials.Identity.Id);
+                string modelId = await this.modelIdStore.GetModelId(clientCredentials.Identity.Id);
+                Option<string> modelIdOption = !string.IsNullOrEmpty(modelId) ? Option.Some(modelId) : Option.None<string>();
```
richma-ms, line 85: >!string.IsNullOrEmpty(modelId) ? Option.Some(modelId) : Option.None<string>(); [](start = 47, length = 78)

Option.Maybe(modelId)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ModelIdStore.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            Preconditions.CheckNonWhiteSpace(id, nameof(id));
+            return !string.IsNullOrWhiteSpace(modelId) ? this.modelIdEntityStore.Put(id, modelId) : Task.CompletedTask;
+        }
+
+        public async Task<string> GetModelId(string id)
+        {
+            Preconditions.CheckNonWhiteSpace(id, nameof(id));
+            Option<string> modelId = await this.modelIdEntityStore.Get(id);
+            return modelId.GetOrElse(string.Empty);
```
richma-ms, line 27: >return modelId.GetOrElse(string.Empty); [](start = 12, length = 39)

seems like you're returning potentially a null-string here, only to have the callers convert it back to an Option<string> again.  Why not just return an Option<string> from this interface?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
richma-ms: see other comment in ModelIdStore, it's prob better to just have it return an Option<string> instead.

---
In reply to: [454164249](https://github.com/Azure/iotedge/pull/3216#discussion_r454164249) [](ancestors = 454164249)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ModelIdStore.cs
dylanbronson: Fair point! Changed

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/DeviceIdentityProvider.cs
```diff
@@ -48,6 +51,7 @@ public async Task<IDeviceIdentity> GetAsync(string clientId, string username, st
                 Preconditions.CheckNonWhiteSpace(clientId, nameof(clientId));
 
                 (string deviceId, string moduleId, string deviceClientType, Option<string> modelId) = ParseUserName(username);
+                modelId.ForEach(async m => await this.modelIdStore.SetModelId(deviceId, m));
```
dylanbronson, line 54: Interesting bug I found: If I put this line below any of the credentials stuff below (aka where productinfo setting is happening currently - line 94), I saw an exception being thrown, and then we weren't able to connect with the modelId. 
Putting it up here fixes the issue

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ClientProvider.cs
```diff
@@ -82,5 +82,30 @@ public IClient Create(IIdentity identity, ITokenProvider tokenProvider, ITranspo
                     throw new InvalidOperationException($"Invalid client identity type {identity.GetType()}");
             }
         }
+
+        public IClient Create(IIdentity identity, ITokenProvider tokenProvider, ITransportSettings[] transportSettings, string modelId)
```
varunpuranik, line 86: Seems like modelId can be an option on the above method?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnection.cs
```diff
@@ -161,12 +168,14 @@ enum EventIds
...
             }
 
-            public static void AttemptingConnectionWithTransport(ITransportSettings[] transportSettings, IIdentity identity)
+            public static void AttemptingConnectionWithTransport(ITransportSettings[] transportSettings, IIdentity identity, Option<string> modelId)
             {
                 string transportType = transportSettings.Length == 1
                     ? TransportName(transportSettings[0].GetTransportType())
                     : transportSettings.Select(t => TransportName(t.GetTransportType())).Join("/");
-                Log.LogInformation((int)EventIds.AttemptingTransport, $"Attempting to connect to IoT Hub for client {identity.Id} via {transportType}...");
+                string message = $"Attempting to connect to IoT Hub for client {identity.Id} via {transportType}";
```
varunpuranik, line 176: Pls use the Events pattern we use everywhere

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -46,7 +47,8 @@ public class CloudConnectionProvider : ICloudConnectionProvider
             TimeSpan operationTimeout,
             bool useServerHeartbeat,
             Option<IWebProxy> proxy,
-            IProductInfoStore productInfoStore)
+            IProductInfoStore productInfoStore,
+            IModelIdStore modelIdStore)
```
varunpuranik, line 51: Wonder if these should be combined into a metadataStore. Check with @ancaantochi if it makes sense.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/DeviceIdentityProvider.cs
varunpuranik: That is where we should do it. What was the exception being thrown?

dylanbronson: Microsoft.Azure.Devices.Edge.Hub.CloudProxy.DeviceScopeApiException: Message: Error getting device scope result from IoTHub, HttpStatusCode: Unauthorized, Content: {"Message":"ErrorCode:IotHubUnauthorizedAccess;Unauthorized","ExceptionMessage":"Tracking ID:[redacted]"}


I have made sure to set the parent of the device to be edgeHub, so I'm not sure why I'm getting this

PR #3218, opened 2020-07-14
[mqtt] Load server certificate on start in edgehub mode

dmolokanov:
* Refactor mqttd bootstrapping logic for cleaner edgehub/generic broker separation
* Add workload API client that supports UDS/HTTP as protocols
* Refactor config to require explicit TCP/TLS specification instead of an array of transports


and-rewsmith:
A few questions. Some of these have been answered in the internal chat, but I think we should also discuss here for clarity.

1. The generated workload CA lasts for how long? When this expires is a systemctl restart needed? Does this happen automatically?
2.  The edgehub server cert lasts for how long? When this expires an edgehub restart is needed? Does this happen automatically?
3. Why does the broker need to support both http and uds? Wouldn't supporting just one be simpler?

mqtt/mqtt-broker/src/configuration/mod.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
and-rewsmith, line 1: Why are we allowing dead code here?

mqtt/mqtt-broker/src/auth.rs
```diff
@@ -1,42 +0,0 @@
...
-    };
-
-    pub(super) fn authenticator() -> impl Authenticator<Error = Box<dyn StdError>> {
-        authenticate_fn_ok(|_| Some(AuthId::Anonymous))
-    }
-
-    pub(super) fn authorizer() -> impl Authorizer {
-        authorize_fn_ok(|_| Authorization::Allowed)
-    }
-}
```
and-rewsmith, line 42: Why did we delete this file?

mqtt/mqtt-broker/config/with_tls.json
```diff
@@ -1,15 +1,11 @@
...
-                "address": "0.0.0.0:1883"
-            } 
+    "transports": {
+        "tcp": {
+            "address": "0.0.0.0:1883"
         },
-        { 
-            "tls": { 
-                "address": "0.0.0.0:8883", 
-                "certificate": "identity.pfx"
```
and-rewsmith, line 11: Can the broker still use a custom certificate?

mqtt/mqtt-broker/src/transport.rs
```diff
@@ -79,10 +41,17 @@ impl Transport {
         Ok(Transport::Tcp(tcp))
     }
 
-    async fn new_tls<A>(addr: A, identity: Identity) -> Result<Self, InitializeBrokerError>
+    pub async fn new_tls<A>(addr: A, cert_path: PathBuf) -> Result<Self, InitializeBrokerError>
     where
         A: ToSocketAddrs + Display,
     {
+        info!("Loading identity from {}", cert_path.display());
```
varunpuranik, line 48: This is not identity cert, right?

dmolokanov: Actually it is. native_tls works only with pkcs12 identity certs

mqtt/mqttd/src/lib.rs
```diff
@@ -1,28 +1 @@
...
-        Ok(())
-    }
-}
-
-impl From<Error> for Terminate {
-    fn from(error: Error) -> Self {
-        Terminate { error }
-    }
-}
+pub mod broker;
```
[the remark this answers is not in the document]
dmolokanov: this is the way in rust to define a module in rust. module `name = rust file name`. 
https://doc.rust-lang.org/book/ch07-05-separating-modules-into-different-files.html

mqtt/mqtt-broker/src/configuration/mod.rs
dmolokanov: oops, good catch!

mqtt/mqtt-broker/src/auth.rs
dmolokanov: I move it to another location

mqtt/mqtt-broker/config/with_tls.json
dmolokanov: yes. the idea is the broker in edgehub mode downloads cert from edgelet and puts it in the filesystem. later when transport starts it reads cert from the file and uses it to accept TLS connections

dmolokanov:
> A few questions. Some of these have been answered in the internal chat, but I think we should also discuss here for clarity.
> 
>     1. The generated workload CA lasts for how long? When this expires is a systemctl restart needed? Does this happen automatically?
good question, I need to do some digging to answer this.
> 
>     2. The edgehub server cert lasts for how long? When this expires an edgehub restart is needed? Does this happen automatically?
the workload API is designed to accept expiration date so the module can request a cert for the time interval it wants. edgehub requests for 90 days so broker does. 

> 
>     3. Why does the broker need to support both HTTP and UDS? Wouldn't supporting just one be simpler?
edgelet API works via both UDS and HTTP. so users can choose whatever protocol they want. moreover, HTTP is the only protocol that works on macOS now.

and-rewsmith:
@dmolokanov 

> so users can choose whatever protocol they want

Doesn't this mean the broker has the option to implement one protocol of its choice?

dmolokanov:
> @dmolokanov
> 
> > so users can choose whatever protocol they want
> 
> Doesn't this mean the broker has the option to implement one protocol of its choice?

@and-rewsmith Maybe, but I think the edgelet client should support everything eventually.

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+pub const WORKLOAD_URI: &str = "IOTEDGE_WORKLOADURI";
+pub const EDGE_DEVICE_HOST_NAME: &str = "EDGEDEVICEHOSTNAME";
+pub const MODULE_ID: &str = "IOTEDGE_MODULEID";
+pub const MODULE_GENERATION_ID: &str = "IOTEDGE_MODULEGENERATIONID";
+
+pub const CERTIFICATE_VALIDITY_DAYS: i64 = 90;
+
+async fn dowload_server_certificate(path: impl AsRef<Path>) -> Result<()> {
```
vadim-kovalyov, line 59: nit: dow**n**load

mqtt/mqtt-edgehub/src/tls.rs
```diff
@@ -0,0 +1,125 @@
+use openssl::{pkcs12::Pkcs12, pkey::PKey, stack::Stack, x509::X509};
+
+pub struct Identity(Vec<u8>);
```
vadim-kovalyov, line 3: nit: I feel that it can be confused with iot hub identity cert. Should we call it ServerIdentity or ServerCertificate? or at least add some comments to clarify

mqtt/mqtt-broker/src/transport.rs
```diff
@@ -79,10 +41,17 @@ impl Transport {
         Ok(Transport::Tcp(tcp))
     }
 
-    async fn new_tls<A>(addr: A, identity: Identity) -> Result<Self, InitializeBrokerError>
+    pub async fn new_tls<A>(addr: A, cert_path: PathBuf) -> Result<Self, InitializeBrokerError>
```
vadim-kovalyov, line 44: nit: What is the benefit of writing the cert to disk? Can we pass `Vec<u8>` instead of `PathBuf`? It can help eliminate storing the cert as a file in edgehub container.

mqtt/mqtt-edgehub/src/edgelet/workload/client.rs
```diff
@@ -0,0 +1,136 @@
...
+            .map_err(ApiError::ConstructRequest)?;
+
+        let res = self
+            .client
+            .request(req)
+            .await
+            .map_err(ApiError::ExecuteRequest)?;
+
+        if res.status() != StatusCode::OK {
+            return Err(ApiError::UnsuccessfulResponse(res.status()).into());
```
vadim-kovalyov, line 47: nit: can we get more than just a status? does workload api provides error description in the response body?

mqtt/mqtt-edgehub/src/edgelet/connect.rs
```diff
@@ -0,0 +1,136 @@
...
+        buf: &mut B,
+    ) -> Poll<std::io::Result<usize>> {
+        match self.get_mut() {
+            Self::Unix(stream) => Pin::new(stream).poll_read_buf(cx, buf),
+            Self::Http(stream) => Pin::new(stream).poll_read_buf(cx, buf),
+        }
+    }
+
+    fn poll_read(
+        self: std::pin::Pin<&mut Self>,
```
vadim-kovalyov, line 79:
```suggestion
        self: Pin<&mut Self>,
```

mqtt/mqtt-edgehub/src/edgelet/connect.rs
```diff
@@ -0,0 +1,136 @@
...
+use tokio::io::{AsyncRead, AsyncWrite};
+use tower_service::Service;
+
+#[derive(Debug, Clone)]
+pub enum Connector {
+    Unix(UnixConnector),
+    Http(HttpConnector),
+}
+
+impl Service<Uri> for Connector {
```
vadim-kovalyov, line 22: nit: it would be nice to have some comments for the future generation :)

mqtt/mqtt-edgehub/src/tls.rs
dmolokanov: it lives in tls module, still confusing?

mqtt/mqtt-broker/src/transport.rs
dmolokanov: I thought about that, but in generic broker world it is hard to describe such a setting. It's not like we download cert from some location, we make very specific HTTP POST request to edgelet. I'm open to any suggestions :)

mqtt/mqtt-edgehub/src/edgelet/workload/client.rs
dmolokanov: I think a can compose a string and put in the error value

mqtt/mqtt-broker/src/transport.rs
vadim-kovalyov: >but in generic broker world it is hard to describe such a setting.

I'm not sure I understand that. 

What I suggested is just to change to  `fn new_tls<A>(addr: A, cert: Vec<u8>) -> Result<Self, InitializeBrokerError>`, and `fn dowload_server_certificate() -> Result<Vec<u8>>`

Then in `edgehub.rs`
```rust
    if let Some(tls) = config.transports().tls() {
        let cert = download_server_certificate().await?;
        server.tls(tls.addr(), cert, authenticator.clone())?;
    }
```
And in `generic.rs`
```rust
    if let Some(tls) = config.transports().tls() {
        let authenticator = authenticate_fn_ok(|_| Some(AuthId::Anonymous));
        let cert = fs::read(tls.cert_path())
            .map_err(|e| InitializeBrokerError::LoadIdentity(cert_path.to_path_buf(), e))?;
        server.tls(tls.addr(), cert, authenticator)?;
    }
```
Pros - you don't have to store edgehub server cert on disk.
does that make sense?

dmolokanov: this part is clear. what to do with settings? now tls section always require a path to identity cert.
for generic broker it is should stay the same : address + cert_path.
for edgehub we can omit that. but it means we need a different setting object for tls.

PR #3226, opened 2020-07-15
Content Trust via Notary

ggjjj:
Feature Implementation contains 
1. Notary Configuration Setup
2. Notary lookup for digest 
3. Create modules via digest
4. Edge Agent reconciliation for images pulled by digest
5. Readme for getting started on Notary Content trust


edgelet/edgelet-docker/src/settings.rs
```diff
@@ -43,6 +45,12 @@ impl MobyRuntime {
     pub fn network(&self) -> &MobyNetwork {
         &self.network
     }
+
+    pub fn content_trust_root_ca_registry(
```
arsing, line 49:
```rust
    pub fn content_trust_root_ca_registry(
        &self,
    ) -> Option<&std::collections::HashMap<String, PathBuf>> {
        self.content_trust_root_ca_registry.as_ref()
    }
```

edgelet/edgelet-docker/test/linux/bad_settings_content_trust.yaml
```diff
@@ -0,0 +1,38 @@
...
+# For instance, when using the fd:// scheme for systemd
+listen:
+  workload_uri: "http://0.0.0.0:8081"
+  management_uri: "http://0.0.0.0:8080"
+homedir: "/tmp"
+moby_runtime:
+  uri: "http://localhost:2375"
+  network: "azure-iot-edge"
+  content_trust_root_ca_registry : "/path/to/root_registry1.ca"
+
```
arsing, line 38: Line ending seems to be busted. Should be a single newline.

edgelet/edgelet-docker/test/linux/sample_settings_content_trust.yaml
```diff
@@ -0,0 +1,43 @@
...
+moby_runtime:
+  uri: "http://localhost:2375"
+  network: "azure-iot-edge"
+  content_trust_root_ca_registry :
+    registryserverurl1 : "/path/to/root_registry1.ca"
+    registryserverurl2 : "/path/to/root_registry2.ca"
+    "" : "/path/to/root_registry3.ca"
+    registryserverurl4 : "/path/to/root_registry4.ca"
+    registryserverurl4 : "/path/to/root_registryreplaced.ca"
+    registryserverurl5 : ""
```
arsing, line 43: Missing newline at end of file.

edgelet/edgelet-docker/test/windows/sample_settings_content_trust.yaml
```diff
@@ -0,0 +1,43 @@
...
+moby_runtime:
+  uri: "npipe://./pipe/iotedge_moby_engine"
+  network: "azure-iot-edge"
+  content_trust_root_ca_registry :
+    registryserverurl1 : "/path/to/root_registry1.ca"
+    registryserverurl2 : "/path/to/root_registry2.ca"
+    "" : "/path/to/root_registry3.ca"
+    registryserverurl4 : "/path/to/root_registry4.ca"
+    registryserverurl4 : "/path/to/root_registryreplaced.ca"
+    registryserverurl5 : ""
```
arsing, line 43: Missing newline at end of file.

arsing:
I assume you're going to update this PR with the actual implementation later?

edgelet/edgelet-docker/src/settings.rs
```diff
@@ -33,6 +34,7 @@ pub struct MobyRuntime {
     #[serde(with = "url_serde")]
     uri: Url,
     network: MobyNetwork,
+    content_trust_root_ca_registry: Option<std::collections::HashMap<String, PathBuf>>,
```
arsing, line 37: Add an import for `std::collections::HashMap` (to the existing `std::collections` import on line 3) instead of using the fully-qualified name.

edgelet/edgelet-docker/src/settings.rs
```diff
@@ -33,6 +34,7 @@ pub struct MobyRuntime {
     #[serde(with = "url_serde")]
     uri: Url,
     network: MobyNetwork,
+    content_trust_root_ca_registry: Option<HashMap<String, PathBuf>>,
```
CindyXing, line 37: Just to confirm, the string is the GUN, right?

ggjjj: Its the registry name URL. ex. myregistry.azurecr.io
GUN would be myregistry.azurecr.io/myimage 

we get the image name and its tag via deployment manifest - we would use that for notary lookup

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -1093,7 +1111,7 @@ mod tests {
                 "management_uri": "unix:///var/run/iotedge/mgmt.sock",
                 "workload_uri": "unix:///var/run/iotedge/workload.sock"
             },
-            "homedir": "/var/lib/iotedge",
```
CindyXing, line 1089: What's the reason or this change? Would it create any impact to existing features?

CindyXing: Also, what if tempdir setting is empty?

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -0,0 +1,90 @@
...
+  let notary_dir = homedir.join("notary");
+
+  // Create trust directory
+  let mut trust_dir = notary_dir;
+  // hash hostname  and then santize and concat
+  // concat - + () 
+  trust_dir.push(&hostname);
+  info!("Trust directory is {}", trust_dir.display());
+  if let Err(err) = fs::remove_dir_all(&trust_dir){
+    if err.kind() != std::io::ErrorKind::NotFound {
```
CindyXing: So, on a clean machine, the notary direction will not be found. In that case, it is expected, right?

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -0,0 +1,90 @@
+use std::path::{Path,PathBuf};
+use log::{info, debug};
+use serde_json::json;
+use std::fs;
+use std::fs::File;
+use crate::{Error,ErrorKind};
+use failure::ResultExt;
+
+pub fn notary_init(homedir : &Path, hostname : String, path : PathBuf) -> Result<(), Error> {
```
CindyXing: Do we always initialize notary no matter customers specify the registry or not?

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -206,8 +208,21 @@ impl MakeModuleRuntime for DockerModuleRuntime {
         //      https://github.com/rust-lang/rust-clippy/issues/3730
         #[allow(clippy::result_map_unwrap_or_else)]
         let created = init_client(settings.moby_runtime().uri())
-            .map(|client| {
+            .and_then(|client| {
+                let home_dir = settings.homedir();
```
CindyXing, line 207: what if home_dir is empty?

edgelet/edgelet-docker/src/notary.rs
ggjjj: No only if it is configured in the config.yaml. The configuration does mandate mentioning the registry hostname and the associated root CA

edgelet/edgelet-docker/src/notary.rs
ggjjj: Yes in a clean machine - we won't have the path and it is expected. Thats why the check is to make sure the error kind is not "Not found" category.

edgelet/edgelet-docker/src/runtime.rs
ggjjj: The reason for this change is for the need to store notary trust collection in ~/home_dir/notary. So need the information of home directory to build notary directory on top of it. It won't affect the existing change. Before it was a fixed directory "/var/lib/iotedge" and now it is auto-generated for each test. Temp_Dir cannot be empty as we are using new() function to get a temp_dir

edgelet/edgelet-docker/src/runtime.rs
ggjjj: home_dir is configured via config.yaml and it is default var/lib/iotedge. I was searching for existing tests to see if there is a check for home_dir to be empty. but there is not one today. I will add the checks

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -1,90 +1,132 @@
...
-use crate::{Error,ErrorKind};
-use failure::ResultExt;
+use std::path::{Path, PathBuf};
 
-pub fn notary_init(homedir : &Path, hostname : String, path : PathBuf) -> Result<(), Error> {
-  // Validate inputs
-  if hostname.is_empty() {
-    return Err(ErrorKind::InitializeNotary("hostname is empty".to_owned()).into());
-  }
+pub fn notary_init(homedir: &Path, hostname: &str, path: &PathBuf) -> Result<PathBuf, Error> {
```
CindyXing, line 9: home_dir, host_name for the naming. 

Also the "host_name" is not very meaningful in this context. I think you meant the registry url, right?

edgelet/contrib/config/linux/config.yaml
```diff
@@ -336,15 +336,15 @@ homedir: "/var/lib/iotedge"
...
 # "/path/to/root_registry1.ca" - path to the root CA for registry 1
 #
-# registryserverurl2 - URL of registry 2
+# registryserverhostname2 - Hostname of registry 2
 # "/path/to/root_registry2.ca" - path to the root CA for registry 2
 #
 # content_trust_root_ca_registry :
-#   registryserverurl1 : "/path/to/root_registry1.ca"
-#   registryserverurl2 : "/path/to/root_registry2.ca"
+#   registryserverhostname1 : "/path/to/root_registry1.ca"
```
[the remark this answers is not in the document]
ggjjj: How about registryservername ? The idea is to make it easier for customer to just specify the server name. ex. myregistry.azurecr.io. We build up the URL by attaching https to the server name. ex. https://myregistry.azurecr.io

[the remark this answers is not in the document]
CindyXing: The myregistry.azurecr.io is a url; https is the protocol supported, right? Behind myregistry.azurecr.io, there can be different servers where customers wouldn't know the name.

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -95,30 +96,106 @@ impl std::fmt::Debug for DockerModuleRuntime {
...
+                                )
+                            }
+                        })
+                        .map(move |digest| (format!("{}@{}", gun, digest), true)),
+                )
+            } else {
+                info!("{} is not enabled for notary content trust", hostname);
+                future::Either::B(futures::future::ok((image_with_tag, false)))
+            }
+        } else {
```
CindyXing, line 171: I am curious what could result in this path?

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -95,30 +96,106 @@ impl std::fmt::Debug for DockerModuleRuntime {
...
+                            ))
+                        })?;
+                        base64::encode_config(&json, base64::URL_SAFE)
+                    }
+                    None => String::new(),
+                };
+                Ok((image, is_content_trust_enabled, creds))
+            })
+            .and_then(move |(image, is_content_trust_enabled, creds)| {
+                if is_content_trust_enabled {
```
CindyXing, line 193: These logging can be in the upper section. Am I right?

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -94,6 +97,39 @@ pub fn notary_init(homedir: &Path, hostname: &str, path: &PathBuf) -> Result<Pat
...
+    Command::new("notary")
+        .env("NOTARY_AUTH", notary_auth)
+        .arg("lookup")
+        .args(&[image_gun, tag])
+        .args(&["-c", config_path])
+        .output_async()
+        .then(|notary_output| {
+            drop(lock);
+            let notary_output = notary_output.with_context(|_| {
+                ErrorKind::LaunchNotary("could not spawn notary process".to_owned())
```
CindyXing, line 116: Can we open a tracking item to improve iotcheck for this kind of issue: docker content trust is enabled; however notary client is not setup/running correctly?

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -94,6 +97,39 @@ pub fn notary_init(homedir: &Path, hostname: &str, path: &PathBuf) -> Result<Pat
...
+            })?;
+            let notary_output_string =
+                String::from_utf8(notary_output.stdout).with_context(|_| {
+                    ErrorKind::LaunchNotary("could not retrieve notary output as string".to_owned())
+                })?;
+            let split_array: Vec<&str> = notary_output_string.split_whitespace().collect();
+            if split_array.len() < 2 {
+                return Err(ErrorKind::LaunchNotary(
+                    "notary digest split array is empty".to_owned(),
```
CindyXing, line 125: From customers' perspective, what can be the reason for this to be hit?

edgelet/edgelet-docker/src/runtime.rs
ggjjj: Else flow in line 171 results when there is no no auth creds for a registry server. it could be a public server like mcr or docker hub then image gets pulled by tag

edgelet/edgelet-docker/src/runtime.rs
ggjjj: Ya but the intention was to enable logging right before image creation

edgelet/edgelet-docker/Cargo.toml
```diff
@@ -19,6 +19,7 @@ serde_derive = "1.0"
 serde_json = "1.0"
 sysinfo = "0.14.10"
 tokio = "0.1.11"
+tokio-process = "0.2"
```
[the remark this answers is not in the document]
ggjjj: AFAIK https://docs.rs/crate/tokio-process/0.2.0 is different from tokio. Line 21 still retains the 0.1 version

edgelet/edgelet-docker/src/notary.rs
ggjjj: Good point. Will do

edgelet/edgelet-docker/src/notary.rs
ggjjj: The output from notary lookup results in a format like this : **tag digest bytes** example : **v1 sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65 528**. this is the resultant array. The goal is get this split array. Very rarely the customer can hit this error with this error being empty. Any other major issues will be hit before line 125.

edgelet/edgelet-docker/src/notary.rs
ggjjj: I have renamed it as registry_server_hostname for more clarification

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerEnvironment.cs
```diff
@@ -74,7 +75,19 @@ public async Task<ModuleSet> GetModulesAsync(CancellationToken token)
                     ? this.restartManager.ComputeModuleStatusFromRestartPolicy(moduleRuntimeInfo.ModuleStatus, dockerModule.RestartPolicy, moduleState.RestartCount, lastExitTime)
                     : moduleRuntimeInfo.ModuleStatus;
 
-                string image = !string.IsNullOrWhiteSpace(dockerRuntimeInfo.Config.Image) ? dockerRuntimeInfo.Config.Image : dockerModule.Config.Image;
+                // Check to see if the originalImage label is set. This label is set for content trust feature where the image is pulled by digest.
+                // OrignalImageStr contains the image with tag and it is used to handle the mismatch of image with digest during reconcilation.
```
massand, line 79: nit: typo
```suggestion
                // orgImageLabelStr contains the image with tag and it is used to handle the mismatch of image with digest during reconcilation.
```

edgelet/contrib/config/linux/config.yaml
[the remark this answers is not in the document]
massand: I believe the common term for it is just "registry". No url, no server, no hostname is needed.

edgelet/edgelet-docker/Cargo.toml
```diff
@@ -41,9 +42,7 @@ json-patch = "0.2.5"
...
 time = "0.1"
 tempdir = "0.3.7"
+tempfile = "3"
 typed-headers = "0.1"
 
-edgelet-test-utils = { path = "../edgelet-test-utils" }
-
-[target.'cfg(unix)'.dev_dependencies]
-tempfile = "3"
+edgelet-test-utils = { path = "../edgelet-test-utils" }
```
massand, line 48: Missing newline in EOF

edgelet/edgelet-docker/test/linux/sample_settings_content_trust.yaml
```diff
@@ -0,0 +1,43 @@
...
+# For instance, when using the fd:// scheme for systemd
+listen:
+  workload_uri: "http://0.0.0.0:8081"
+  management_uri: "http://0.0.0.0:8080"
+homedir: "/tmp"
+moby_runtime:
+  uri: "http://localhost:2375"
+  network: "azure-iot-edge"
+  content_trust_root_ca_registry :
+    registryserverhostname1 : "/path/to/root_registry1.ca"
```
massand, line 38: I would just rename these to registry1, registry2, etc.

edgelet/edgelet-http-mgmt/src/server/module/create.rs
```diff
@@ -6,6 +6,7 @@ use futures::{Future, Stream};
 use hyper::header::{CONTENT_LENGTH, CONTENT_TYPE};
 use hyper::{Body, Request, Response, StatusCode};
 use log::debug;
+// use log::{debug, info};
```
massand, line 9: Remove

edgelet/edgelet-docker/src/settings.rs
massand, line 37: Can we shorten this name? In fact, your tests had a short name for this :)  
```suggestion
    content_trusted_certs: Option<HashMap<String, PathBuf>>,
```

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -0,0 +1,183 @@
...
+                )
+                .into());
+            }
+            Ok((split_array[1].to_string(), lock))
```
massand, line 139: Can we document the notary output here? If we had a deserializer implemented for the output, this will be easier to add a test for too. @arsing - thoughts?

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -91,30 +97,110 @@ impl std::fmt::Debug for DockerModuleRuntime {
...
-            || Ok("".to_string()),
-            |a| {
-                let json = serde_json::to_string(a).with_context(|_| {
-                    ErrorKind::RegistryOperation(RegistryOperation::PullImage(image.clone()))
-                })?;
-                Ok(base64::encode_config(&json, base64::URL_SAFE))
-            },
-        );
+        let image_with_tag = config.image().to_string();
+        // check if the serveraddress exists & check if it exists in notary_registries
```
massand, line 119: nit on style: newline between preceding line and comment

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -91,30 +97,110 @@ impl std::fmt::Debug for DockerModuleRuntime {
...
+            Some(a) => (a.serveraddress(), a.username(), a.password()),
+            None => (None, None, None),
+        };
+        let hostname = registry_hostname.unwrap_or_default();
+        let username = registry_username.unwrap_or_default();
+        let password = registry_password.unwrap_or_default();
+
+        let image = if !hostname.is_empty() && !username.is_empty() && !password.is_empty() {
+            let config_path = self.notary_registries.get(hostname);
+            if let Some(c) = config_path {
```
massand, line 131: You don't need to move c into a temp variable. Just rename c to config_path_buf
```suggestion
            if let Some(config_path_buf) = config_path {
```

massand (changes requested):

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerEnvironment.cs
```diff
@@ -74,7 +75,19 @@ public async Task<ModuleSet> GetModulesAsync(CancellationToken token)
                     ? this.restartManager.ComputeModuleStatusFromRestartPolicy(moduleRuntimeInfo.ModuleStatus, dockerModule.RestartPolicy, moduleState.RestartCount, lastExitTime)
                     : moduleRuntimeInfo.ModuleStatus;
 
-                string image = !string.IsNullOrWhiteSpace(dockerRuntimeInfo.Config.Image) ? dockerRuntimeInfo.Config.Image : dockerModule.Config.Image;
+                // Check to see if the originalImage label is set. This label is set for content trust feature where the image is pulled by digest.
```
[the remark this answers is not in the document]
CindyXing: I have similar question... Does this mean customer will have to specify this label? I assume the label value needs to be the digest value?

doc/NotaryContentTrust.md
```diff
@@ -0,0 +1,109 @@
...
+
+Once Notary client is installed, install the [Docker client](https://docs.docker.com/get-docker/). Docker client has to be aware of the TUF trust collection so do not skip installing Docker client. 
+
+#### Notary Configuration
+Notary client directory exists in `~/.notary`. A configuration file has be created in `~/.notary/config.json` with the following contents
+```json
+{
+    "trust_dir" : "~/.docker/trust",
+    "remote_server": {
+     "url": "https://exampleregistry.azurecr.io"
```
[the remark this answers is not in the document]
ggjjj: Yes thats right

doc/NotaryContentTrust.md
```diff
@@ -0,0 +1,109 @@
...
+
+Note at the end of the above steps the following files are generated:
+
+- root_image.key is the private key used for initializing TUF trust collection 
+- root_image.csr is the Certificate Signing Request file 
+- root_image.crt is the public cert used for initializing TUF trust collection
+
+Important Note:
+
+-   The Comman Name (CN) in contain the registry name in which it is being pushed to along with the image name.
```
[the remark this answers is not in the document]
ggjjj: My bad. Statement is meant to say this "The Common Name (CN) of the root certificate for each container image should container the GUN. I will make it very clear. Thanks for catching it!

doc/NotaryContentTrust.md
```diff
@@ -0,0 +1,109 @@
+# Notary Content Trust in IoT Edge
+Content Trust is achieved in IoT edge with [Notary](https://github.com/theupdateframework/notary) which is based on The Update Framework (TUF). Notary aims to make the internet more secure by making it easy for people to publish and verify content. The goal of content trust in IoT edge is to secure the docker container images of the Edge modules with its trust pinned to the Edge device via a root Certificate Authority(CA). The IoT Edge device only deploys signed images and rejects any other images that are tampered or modified in the transfer over the internet. Feature is supported in Unix only. 
+
+## Sign and Publish Container Images in Azure Container Registry
+The container images can be signed and published in Azure Container Registry which has an in built Notary support for content trust. 
+
+To publish signed images, following tools must be installed in a publisher's environment which is different from the Edge device. 
+1. OpenSSL - to generate root CA and root certs for each container
```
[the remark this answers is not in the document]
ggjjj: root CA for each container registry and root certs for each container

doc/NotaryContentTrust.md
```diff
@@ -0,0 +1,109 @@
+# Notary Content Trust in IoT Edge
+Content Trust is achieved in IoT edge with [Notary](https://github.com/theupdateframework/notary) which is based on The Update Framework (TUF). Notary aims to make the internet more secure by making it easy for people to publish and verify content. The goal of content trust in IoT edge is to secure the docker container images of the Edge modules with its trust pinned to the Edge device via a root Certificate Authority(CA). The IoT Edge device only deploys signed images and rejects any other images that are tampered or modified in the transfer over the internet. Feature is supported in Unix only.
```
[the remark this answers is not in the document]
ggjjj: Yes the statement makes it look like we always turn content trust. What I meant to say is if turned on "The IoT Edge device only deploys signed images and rejects any other images that are tampered or modified in the transfer over the internet." I will make that change

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerEnvironment.cs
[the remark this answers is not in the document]
ggjjj: The label is set in the iotedged while creating the image with digest. The runtime is aware of the image with digest but edge agent is expecting an image with tag. Thats why the logic in edge agent is added - when a label is set then the image is the value of the label i.e image with tag. The reconciliation goes on smooth by this logic

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerEnvironment.cs
```diff
@@ -74,7 +75,19 @@ public async Task<ModuleSet> GetModulesAsync(CancellationToken token)
                     ? this.restartManager.ComputeModuleStatusFromRestartPolicy(moduleRuntimeInfo.ModuleStatus, dockerModule.RestartPolicy, moduleState.RestartCount, lastExitTime)
                     : moduleRuntimeInfo.ModuleStatus;
 
-                string image = !string.IsNullOrWhiteSpace(dockerRuntimeInfo.Config.Image) ? dockerRuntimeInfo.Config.Image : dockerModule.Config.Image;
+                // Check to see if the originalImage label is set by the iotedege daemon. This label is set for content trust feature where the image is pulled by digest.
+                // OrignalImageLabelStr contains the image with tag and it is used to handle the mismatch of image with digest during reconcilation.
+                var labels = dockerRuntimeInfo.Config.CreateOptions?.Labels ?? new Dictionary<string, string>();
+                labels.TryGetValue(Core.Constants.Labels.OriginalImage, out string orignalImageLabelStr);
```
massand, line 81:
```suggestion
                labels.TryGetValue(Core.Constants.Labels.OriginalImage, out string originalImageLabelStr);
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerEnvironment.cs
```diff
@@ -74,7 +75,19 @@ public async Task<ModuleSet> GetModulesAsync(CancellationToken token)
                     ? this.restartManager.ComputeModuleStatusFromRestartPolicy(moduleRuntimeInfo.ModuleStatus, dockerModule.RestartPolicy, moduleState.RestartCount, lastExitTime)
                     : moduleRuntimeInfo.ModuleStatus;
 
-                string image = !string.IsNullOrWhiteSpace(dockerRuntimeInfo.Config.Image) ? dockerRuntimeInfo.Config.Image : dockerModule.Config.Image;
+                // Check to see if the originalImage label is set by the iotedege daemon. This label is set for content trust feature where the image is pulled by digest.
+                // OrignalImageLabelStr contains the image with tag and it is used to handle the mismatch of image with digest during reconcilation.
```
massand, line 79:
```suggestion
                // OriginalImageLabelStr contains the image with tag and it is used to handle the mismatch of image with digest during reconciliation.
```

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerEnvironment.cs
```diff
@@ -74,7 +75,19 @@ public async Task<ModuleSet> GetModulesAsync(CancellationToken token)
...
+                var labels = dockerRuntimeInfo.Config.CreateOptions?.Labels ?? new Dictionary<string, string>();
+                labels.TryGetValue(Core.Constants.Labels.OriginalImage, out string orignalImageLabelStr);
+                if (!string.IsNullOrWhiteSpace(orignalImageLabelStr))
```
massand, line 82: Are you planning to add tests DockerEnvironmentTest for cases when DockerRuntimeInfo contains the image label set in content trust case?

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    #[cfg(unix)]
+    #[test]
+    fn check_for_root_ca_file_does_not_exist() {
+        let registry_server_hostname = r"myregistry.azurecr.io";
+        let home_dir = NamedTempFile::new().unwrap();
+        let root_ca_file = PathBuf::from("filedoesnotexist.crt");
+        let result = notary::notary_init(home_dir.path(), &registry_server_hostname, &root_ca_file);
+        let err = result.unwrap_err();
+        let _display_msg = format!("root ca at path {} does not exist", root_ca_file.display());
```
massand, line 179:
```suggestion
        let display_msg = format!("root ca at path {} does not exist", root_ca_file.display());
```

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -0,0 +1,185 @@
...
+        assert!(matches!(
+            err.kind(),
+            ErrorKind::InitializeNotary(_display_msg)
```
massand, line 182:
```suggestion
            ErrorKind::InitializeNotary(display_msg)
```

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -0,0 +1,185 @@
...
+                ErrorKind::LaunchNotary("could not spawn notary process".to_owned())
+            })?;
+            let notary_output_string =
+                String::from_utf8(notary_output.stdout).with_context(|_| {
+                    ErrorKind::LaunchNotary("could not retrieve notary output as string".to_owned())
+                })?;
+            let split_array: Vec<&str> = notary_output_string.split_whitespace().collect();
+            if split_array.len() < 2 {
+                return Err(ErrorKind::LaunchNotary(
+                    "notary digest split array is empty".to_owned(),
```
massand, line 135: Capture the `notary_output_string` and log it using `format!` macro. Diagnostics will become easier with this

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -91,30 +97,112 @@ impl std::fmt::Debug for DockerModuleRuntime {
...
+                    config_path_buf.clone().into_os_string().into_string();
+                let config_path_str;
+                match config_path_string_result {
+                    Ok(c) => config_path_str = c,
+                    Err(_) => config_path_str = "".to_string(),
+                }
+                let mut image_with_tag_parts = image_with_tag.split(':');
+                let gun = image_with_tag_parts
+                    .next()
+                    .expect("split always returns atleast one element")
```
massand, line 145: This is a panic, which isn't desirable. We should map_err and return Err, since `image_with_tag` is user defined

edgelet/edgelet-docker/src/notary.rs
ggjjj: The variable is unused so it has to be with "_". Using with assert doesn't count and clippy throws error

edgelet/edgelet-docker/src/runtime.rs
ggjjj: In this scenario, its never will hit this error and its fine to do the panic part.

edgelet/edgelet-docker/src/notary.rs
arsing: >Using with assert doesn't count and clippy throws error

You're not actually using it with assert, which is why clippy is correctly telling you that it's unused.

edgelet/edgelet-docker/src/notary.rs
arsing, line 182: This is not doing what you think it's doing.

This is equivalent to writing

```rust
assert!(match err.kind() {
    ErrorKind::InitializeNotary(_display_msg) => true,
    _ => false,
});
```

which means `_display_msg` is just a pattern that is being ignored.

You want `ErrorKind::InitializeNotary(m) where m == display_msg` to actually assert that the string in the enum variant is `display_msg`

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -91,30 +97,112 @@ impl std::fmt::Debug for DockerModuleRuntime {
...
+        let registry_auth = config.auth();
+        let (registry_hostname, registry_username, registry_password) = match registry_auth {
+            Some(a) => (a.serveraddress(), a.username(), a.password()),
+            None => (None, None, None),
+        };
+        let hostname = registry_hostname.unwrap_or_default();
+        let username = registry_username.unwrap_or_default();
+        let password = registry_password.unwrap_or_default();
+
+        let image = if !hostname.is_empty() && !username.is_empty() && !password.is_empty() {
```
arsing, line 131: This means content trust can't be used with public registries that don't have username and password.

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -327,46 +432,138 @@ impl ModuleRuntime for DockerModuleRuntime {
...
+                let notary_auth = format!("{}:{}", username, password);
+                let notary_auth = base64::encode(&notary_auth);
+                let lock = self.notary_lock.clone();
+                let mutex = MutexFuture(lock);
+                future::Either::A(
+                        mutex
+                            .and_then({
+                                let gun = gun.clone();
+                                move |lock| {
+                                    notary::notary_lookup(
```
arsing, line 479: The point of making `notary_lock` hold a `BTreeMap` and populate that map in `image_pull` was to skip calling `notary_lookup` if the map already contains an entry.

arsing (changes requested):

edgelet/edgelet-docker/src/runtime.rs
ggjjj: yes they have to explicitly sign (to have publisher's identity) and publish in their private registries

arsing: Signing and publishing require auth. We're talking about pulling images here, which doesn't need to be gated by auth.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerEnvironment.cs
ggjjj: Added them.

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -0,0 +1,190 @@
+use crate::{Error, ErrorKind};
```
arsing, line 1: Order of imports goes like:

1. `std::*`
1. Third-party crates
1. Our crates.
1. `crate::*`

with one empty line separating each category.

edgelet/edgelet-docker/src/notary.rs
```diff
@@ -0,0 +1,190 @@
...
+                )
+                .into());
+            }
+
+            // Notary Server output on lookup is of the format [tag, digest, bytes]
+            Ok((split_array[1].to_string(), lock))
+        })
+}
+
+#[cfg(test)]
```
arsing, line 150: Why are these tests Linux-only?

edgelet/edgelet-docker/src/notary.rs
arsing, line 1: Copyright header is missing.

arsing (changes requested):

edgelet/edgelet-docker/src/notary.rs
ggjjj: Targeting only Linux for now

arsing: The code isn't Linux-only.

ggjjj: Yea since we are supporting only Linux, I made it all for Linux so its easier to focus on dotnet checkins for linux platforms alone.

arsing: https://github.com/Azure/iotedge/blob/627acd2e7efa6b137c2586ca48631f5308f8ae41/edgelet/edgelet-docker/src/notary.rs#L12-L14

The code isn't Linux-only. Why are the tests Linux-only?

arsing (approved):

arsing:
I've signed off for the Rust part. Someone else needs to review the C# part.

massand (approved):

CindyXing (approved):

doc/NotaryContentTrust.md
```diff
@@ -1,21 +1,21 @@
...
+Content Trust is achieved in IoT edge with [Notary](https://github.com/theupdateframework/notary) which is based on The Update Framework (TUF). Notary aims to make the internet more secure by making it easy for people to publish and verify content. The goal of Content Trust in IoT edge is to secure the docker container images of the Edge modules with its trust pinned to the Edge device via a root Certificate Authority(CA). When Content Trust is enabled, IoT Edge device only deploys signed images and rejects any other images that are tampered or modified in the transfer over the internet. Feature is supported in Linux OS only for now.
```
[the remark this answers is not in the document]
ggjjj: Sure I will add in next PR as I already set the merge label.

PR #3228, opened 2020-07-15
[mqtt] Remove reqwest dependency in favor of hyper

dmolokanov:
Avoid multiple HTTP clients in the workspace. Remove reqwestdependency in favor of hyper.

### DIFF
--- a/mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
+++ b/mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
@@ -4,7 +4,9 @@ use std::{
 };
 
 use async_trait::async_trait;
-use reqwest::Client;
+use bytes::buf::BufExt;
+use http::StatusCode;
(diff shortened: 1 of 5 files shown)


mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
```diff
@@ -28,30 +30,33 @@ impl EdgeHubAuthenticator {
...
-            .json(&req)
-            .send()
+            .request(req)
             .await
             .map_err(AuthenticateError::SendRequest)?;
 
-        if response.status() != reqwest::StatusCode::OK {
-            return Err(AuthenticateError::ProcessResponse);
+        if http_res.status() != StatusCode::OK {
+            return Err(AuthenticateError::UnsuccessfullResponse(http_res.status()));
```
vadim-kovalyov, line 46: nit: same question here, is it possible to get some more info from the body? @vipeller , does your authN api provides useful info in the body in case of an error?

PR #3232, opened 2020-07-16
[k8s public preview] Have EdgeDeploymentOperator report status to Agent

darobs:
Instead of allowing Agent to crash on Edge Deployment failure, the
failure that was reported to edge Deployment status will get reported
back in the EdgeAgent's twin. This more closely follows how single
device Edge Agent operates.


darobs:
Reopening this to get it into beta11.  The EdgeDeploymentStatusCommandTest now confirms that the status message is *not* lost as @CindyXing 's request on #3315

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/edgedeployment/EdgeDeploymentOperatorTest.cs
```diff
@@ -216,8 +215,7 @@ public async void StatusIsFailedWithSpecialMessageWhenHttpExceptionIsThrown()
                 client,
                 controller);
 
-            HttpOperationException ex = await Assert.ThrowsAsync<HttpOperationException>(() => edgeOperator.EdgeDeploymentOnEventHandlerAsync(WatchEventType.Added, edgeDefinition));
-            Assert.Equal(controllerException, ex);
+            await edgeOperator.EdgeDeploymentOnEventHandlerAsync(WatchEventType.Added, edgeDefinition);
             Assert.True(reportedStatus.HasValue);
             Assert.Equal(expectedStatus, reportedStatus.OrDefault());
```
CindyXing, line 220: Given there is failure deploymentstatus, can we run the deploymentstatuscommand and verify? Same for the other one.

CindyXing (changes requested):

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/edgedeployment/EdgeDeploymentOperatorTest.cs
darobs: That's more of an integration test, but I can add one.

CindyXing: It's just we are eating exception. Or we can change the test case title.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentStatusCommand.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            // If the EdgeDeployment operator fails, report this to the plan executor
+            // as an exception.
+            this.activeStatus.ForEach(status =>
+                {
+                    if (status.State != EdgeDeploymentStatusType.Success)
+                    {
+                        throw new ConfigOperationFailureException(status.Message);
+                    }
+                });
```
zasherif, line 35: This sounds like reducing all failures into one scenario and all what we expect the operator to do it to report the failure. I understand that the ReportDeploymentFailure method only reports the failure without an attempt to recover from it. Will it be helpful that the exception class characterize a failure type to propagate useful information about the failure? For example the Exception can have the following types:

1. TemporaryError; signals that the exception handler can retry the deployment
2. PermanentError; signals that the exception handler must permanently fail the deployment

That's also means that the plan executor should have a mechanism to retry (note: I did not read that part of the code yet).

zasherif (commented):
lgtm, defer decisions to others.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentStatusCommand.cs
darobs: > This sounds like reducing all failures into one scenario and all what we expect the operator to do it to report the failure

That is approximately how the single-device edge handles problems from command execution. Exceptions are caught and this is reduced to success or failure. However, C# has a notion of [fatal errors](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.common.exceptionextensions.isfatal?view=azure-dotnet), and I should probably allow agent to crash if I get a fatal error. 

Overall, though, the idea of temporary errors vs. permanent may be useful - We have generally left it up to the plan runner to determine how often to retry commands, usually by progressively backing off, but usually the failures do land in one of these two categories.

zasherif (approved):
lgtm

PR #3235, opened 2020-07-16
EdgeHub Build Fixes

and-rewsmith:
This PR contains the following changes:
- Since the broker now dynamically retrieves the cert and stores it in the location specified in the config, we need to add this path to the config. We also need to change the config to fit the new format from Denis's recent workload PR.
- The build pipeline injects a build argument (EXE_DIR=.) when building the edgehub image, which overwrites the intended path and bloats the image size. We should not use this build argument in the Dockerfile.
- There was a misconfiguration on arm for the path of the executables. I forgot to remove the "./edge-hub" inside the arm dockerfiles while debugging something unrelated


mqtt/mqtt-broker/config/production.json
```diff
@@ -1,14 +1,11 @@
...
+        "tcp": {
+            "address": "0.0.0.0:1883"
         },
-        {
-            "tls": {
-                "address": "0.0.0.0:8883"
-            }
+        "tls": {
+            "address": "0.0.0.0:8883",
+            "certificate": "broker-production-cert.pfx"
```
and-rewsmith, line 8: We might want another name here.

edge-hub/docker/linux/arm64v8/Dockerfile
```diff
@@ -5,11 +5,9 @@ ADD ./watchdog/target/aarch64-unknown-linux-gnu/release/watchdog /usr/local/bin/
 ADD ./mqttd/target/aarch64-unknown-linux-gnu/release/mqttd /usr/local/bin/mqttd
 ADD ./mqttd/production.json /app/mqttd/production.json
 
-ARG EXE_DIR=./Microsoft.Azure.Devices.Edge.Hub.Service
-
 WORKDIR /app
 
-COPY $EXE_DIR/ ./edge-hub
```
and-rewsmith, line 12: Leftover from unrelated debugging. We should remove.

edge-hub/watchdog/src/main.rs
```diff
@@ -36,7 +36,7 @@ fn main() -> Result<()> {
     let edgehub_handle = run(
         "Edge Hub",
         "dotnet",
-        vec!["/app/Microsoft.Azure.Devices.Edge.Hub.Service/Microsoft.Azure.Devices.Edge.Hub.Service.dll".to_string()],
```
and-rewsmith, line 39: This was needed before because the build pipeline was injecting a faulty build argument, which had the side effect of copying the exe containing folder itself ```Microsoft.Azure.Devices.Edge.Hub.Service```. We no longer need this.

mqtt/mqtt-broker/config/production.json
dmolokanov: actually you need just remove this line at all.

mqtt/mqtt-broker/config/production.json
```diff
@@ -1,14 +1,11 @@
 {
```
dmolokanov, line 1: Let's move this file to `mqtt/contrib/edgehub/broker.json`
The folder `mqtt/contrib/edgehub/` might have more config files potentially (mqtt bridge) for instance.

mqtt/contrib/edgehub/broker.json
```diff
@@ -0,0 +1,11 @@
+{
+    "transports": {
+        "tcp": {
+            "address": "0.0.0.0:1883"
+        },
+        "tls": {
+            "address": "0.0.0.0:8883",
+            "certificate": "broker-production-cert.pfx"
```
and-rewsmith, line 8: Recommendations for new name?

PR #3237, opened 2020-07-16
Don't stage deployments

damonbarry:
Given recent improvements to the end-to-end test framework (#2635, #2753), the logic that splits deployments into multiple stages (#2248) is no longer needed. This change removes that logic.

In verifying the changes, I also ran into occasional failures in the `ValidateMetrics` test because the stringized JSON labels that were recently added to Edge Agent containers failed to compare. The expected and actual twin reported properties always had the same JSON properties, but sometimes in different orders:

```
systemModules.edgeAgent.settings.createOptions: {"Labels":{"net.azure-devices.edge.create-options":"{}","net.azure-devices.edge.env":"{\"RuntimeLogLevel\":{\"value\":\"debug\"},\"PerformanceMetricsUpdateFrequency\":{\"value\":\"00:00:20\"}}","net.azure-devices.edge.owner":"Microsoft.Azure.Devices.Edge.Agent"}
systemModules.edgeAgent.settings.createOptions: {"Labels":{"net.azure-devices.edge.create-options":"{}","net.azure-devices.edge.env":"{\"PerformanceMetricsUpdateFrequency\":{\"value\":\"00:00:20\"},\"RuntimeLogLevel\":{\"value\":\"debug\"}}","net.azure-devices.edge.owner":"Microsoft.Azure.Devices.Edge.Agent"}
```

The logic that compares these properties is trying to confirm that the configuration has been deployed, and the agent's labels aren't related to the deployment so I just removed them from the comparison.


test/Microsoft.Azure.Devices.Edge.Test.Common/DaemonConfiguration.cs
```diff
@@ -19,7 +19,7 @@ public DaemonConfiguration(string configYamlFile, Option<string> agentImage, Opt
             string contents = File.ReadAllText(this.configYamlFile);
             this.config = new YamlDocument(contents);
             this.UpdateAgentImage(
-                agentImage.GetOrElse("mcr.microsoft.com/microsoft/azureiotedge-agent:1.0"),
```
dylanbronson, line 22: Do we accidentally have an image here, or was this just not working if agent image wasn't supplied?

damonbarry: I don't think this ever worked. It is only used in a very specific situation: running the tests manually, without specifying "edgeAgentBootstrapImage" in context.json.

PR #3238, opened 2020-07-16
[mqtt] Upgrade for rust 1.45

dmolokanov:
upgrade `rust-toolchain` back to latest stable
fix clippy


mqtt/mqtt-broker/tests/common.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
and-rewsmith, line 1: why dead code now?

dmolokanov: Some methods are not used in PacketStream. clippy shows warnings

PR #3239, opened 2020-07-16
make the broker send the current subscription list when a client connects

vipeller:
The broker and edgehub needs a common state of the subscriptions of a client. in case of persistent sessions, when a client comes back, it already has a (previously established) subscription state. To avoid state-mismatches, it is best to store this state in one place, and the best place for that is in the broker. That means that when a client comes back, EdgeHub will not know the already established subscriptions. In order to let it know, the broker should send it when a client connects.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
@@ -412,9 +412,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
                 lock (this.deviceProxyLock)
                 {
                     Option<DeviceConnection> currentValue = this.DeviceConnection;
-                    ConcurrentDictionary<DeviceSubscription, bool> subscriptions = this.DeviceConnection.Map(d => d.Subscriptions)
-                        .GetOrElse(new ConcurrentDictionary<DeviceSubscription, bool>());
-                    this.DeviceConnection = Option.Some(new DeviceConnection(deviceProxy, subscriptions));
+                    this.DeviceConnection = Option.Some(new DeviceConnection(deviceProxy, new ConcurrentDictionary<DeviceSubscription, bool>()));
                     return currentValue;
                 }
             }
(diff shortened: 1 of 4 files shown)


mqtt/mqtt-broker/src/broker.rs
```diff
@@ -302,7 +302,11 @@ where
                     session.send(event)?;
                 }
 
+                let subscriptions =
```
dmolokanov, line 305: nit: move this line with variable initialization closer to usage, please. Now it is 1 line farther than it should be

dmolokanov: borrow checker rules prevent to move this line closer to publish method

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
```diff
@@ -412,9 +412,7 @@ public Option<DeviceConnection> AddDeviceConnection(IDeviceProxy deviceProxy)
                 lock (this.deviceProxyLock)
                 {
                     Option<DeviceConnection> currentValue = this.DeviceConnection;
-                    ConcurrentDictionary<DeviceSubscription, bool> subscriptions = this.DeviceConnection.Map(d => d.Subscriptions)
-                        .GetOrElse(new ConcurrentDictionary<DeviceSubscription, bool>());
-                    this.DeviceConnection = Option.Some(new DeviceConnection(deviceProxy, subscriptions));
+                    this.DeviceConnection = Option.Some(new DeviceConnection(deviceProxy, new ConcurrentDictionary<DeviceSubscription, bool>()));
```
varunpuranik, line 415: tests?

mqtt/mqtt-broker/tests/translation.rs
```diff
@@ -221,6 +221,9 @@ async fn test_twin_with_client_id(client_id: &str) {
         .with_client_id(ClientId::IdWithCleanSession(client_id.into()))
         .build();
 
+    // wait till device_1 gets connected before edge_hub starts listening to changes
+    device_1.connections().recv().await;
```
vadim-kovalyov, line 225: Since you are already here, can you fix all the tests please?

PR #3240, opened 2020-07-16
[PnP] PlugAndPlay first E2E test

dylanbronson:
Adding first End-to-End test for Plug and Play. The idea of the test is simple.
1. Use LeafDevice to send a message with a client that is created with ModelId
2. Wait for events to arrive in EventHub
3. Get Digital Twin from IoT Hub
4. Pass if ModelID in Digital Twin is expected. Fail if it is not or is not present.

In order to run this test, I needed to use Central US EUAP (preview) region for an IoTHub. I added a new IoTHub, fixtures and PREVIEW_IOT_HUB/PREVIEW_EVENT_HUB environment variables accordingly to the new e2e test framework. This seemed to be the easiest way to do this, and we just need to switch it over once the preview is GA (aka in the same region as testiothub114)


test/README.md
```diff
@@ -57,7 +57,9 @@ The tests also expect to find several _secret_ parameters. While these can techn
 |------|----------|-------------|
 | `[E2E_]DPS_GROUP_KEY` | * | The symmetric key of the DPS [enrollment group](https://docs.microsoft.com/en-us/azure/iot-dps/concepts-service#enrollment-group) to use. Required when running any DPS tests, ignored otherwise. |
 | `[E2E_]IOT_HUB_CONNECTION_STRING` | ✔ | Hub-scoped IoT Hub connection string that can be used to get/add/remove devices, deploy edge configurations, and get/update module twins. |
+| `[E2E_]PREVIEW_IOT_HUB_CONNECTION_STRING` | * | Hub-scoped IoT Hub connection string in the Central EUAP region for preview testing - required only for PlugAndPlay tests. |
```
damonbarry, line 60: This variable doesn't have to point to a hub in Central EUAP; that's just how you're using it in practice. The requirement is that you specify a hub that has PnP preview enabled on it.

test/README.md
```diff
@@ -57,7 +57,9 @@ The tests also expect to find several _secret_ parameters. While these can techn
...
 | `[E2E_]EVENT_HUB_ENDPOINT` | ✔ | Connection string used to connect to the Event Hub-compatible endpoint of your IoT Hub, to listen for D2C events sent by devices or modules. |
+| `[E2E_]PREVIEW_EVENT_HUB_ENDPOINT` | * | Event hub in the Central EUAP region, for preview testing - required only for PlugAndPlay tests. |
```
damonbarry, line 62: See previous comment about Central EUAP...

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeModule.cs
```diff
@@ -121,6 +121,37 @@ async Task WaitForStatusAsync()
             this.deviceId,
             this.iotHub.EntityPath);
 
+        public Task<string> WaitForEventsReceivedFromDeviceAsync(
```
damonbarry, line 124: This method really shouldn't be a member of `EdgeModule` since it is a copy of `EdgeModule.WaitForEventsReceivedAsync`, but with all the module stuff stripped out.

You have a whole bunch of one-off logic directly in your test (or at least in that test's file), which is understandable since you're doing lots of manual steps that will go away when the feature exits preview. I'd put this method in there too, or if you want to be clean, create a `PnpDevice` class that is similar to `LeafDevice`, and put it in there.

test/Microsoft.Azure.Devices.Edge.Test.Common/IotHub.cs
```diff
@@ -65,6 +65,12 @@ public IotHub(string iotHubConnectionString, string eventHubEndpoint, Option<Uri
         public string Hostname =>
             IotHubConnectionStringBuilder.Create(this.iotHubConnectionString).HostName;
 
+        public string HubName =>
```
damonbarry, line 68: Is `HubName` being used somewhere?

test/Microsoft.Azure.Devices.Edge.Test/helpers/Context.cs
```diff
@@ -102,9 +104,11 @@ IEnumerable<Registry> GetAndValidateRegistries()
             this.LoadGenImage = Option.Maybe(Get("loadGenImage"));
             this.RelayerImage = Option.Maybe(Get("relayerImage"));
             this.NetworkControllerImage = Option.Maybe(Get("networkControllerImage"));
+            this.PlugAndPlayIdentityImage = Option.Maybe(Get("plugAndPlayIdentityImage"));
             this.EdgeAgentBootstrapImage = Option.Maybe(Get("edgeAgentBootstrapImage"));
             this.TestTimeout = TimeSpan.FromMinutes(context.GetValue("testTimeoutMinutes", 5));
             this.Verbose = context.GetValue<bool>("verbose");
+            this.IoTHubAccessToken = Option.Maybe(Get("HUB_ACCESS_TOKEN"));
```
damonbarry, line 111: Is `IoTHubAccessToken` begin used anywhere? It isn't documented in the README and I don't see any references...

test/Microsoft.Azure.Devices.Edge.Test/helpers/ManualProvisioningFixturePreview.cs
```diff
@@ -0,0 +1,70 @@
...
+    // and even if it were, it doesn't run the teardown method when
+    // a test times out. We need teardown to run, to remove the
+    // device registration from IoT Hub and stop the daemon. So
+    // we have our own timeout mechanism.
+    public class ManualProvisioningFixturePreview : BaseFixture
+    {
+        protected readonly IotHub iotHub;
+        protected IEdgeDaemon daemon;
+
+        public ManualProvisioningFixturePreview()
```
damonbarry, line 22: The only difference I see between this and `ManualProvisioningFixture` is that you pass different arguments into the `IotHub` constructor. And I don't see any differences between `SasManualProvisioningFixturePreview` and `SasManualProvisioningFixture`.

Seems like you could save a lot of duplicate code by adding two parameters to the fixtures' constructors, with the non-preview values as defaults. Then in your test class, add a constructor that calls `base(previewConnectionString, previewEventHubEndpoint)`.

test/modules/PlugAndPlayDevice/PlugAndPlayIdentity.csproj
```diff
@@ -0,0 +1,47 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
damonbarry, line 1: Can you also exclude your module from code coverage by adding it to `CodeCoverage.runsettings`? I think the following will work, somewhere in [this section](https://github.com/Azure/iotedge/blob/1017802/CodeCoverage.runsettings#L57):

```
<ModulePath>.*plugandplaydevice\..*</ModulePath>
```

test/Microsoft.Azure.Devices.Edge.Test.Common/LeafDevice.cs
```diff
@@ -41,8 +42,11 @@ public class LeafDevice
             bool useSecondaryCertificate,
             CertificateAuthority ca,
             IotHub iotHub,
-            CancellationToken token)
+            CancellationToken token,
+            Option<string> modelId)
```
damonbarry, line 46: Are you able to do `Option<string> modelId = Option.None<string>())` here, so that existing tests that use LeafDevice don't have to change?

test/Microsoft.Azure.Devices.Edge.Test/helpers/ManualProvisioningFixture.cs
```diff
@@ -19,12 +19,14 @@ public class ManualProvisioningFixture : BaseFixture
         protected readonly IotHub iotHub;
         protected IEdgeDaemon daemon;
 
+        public ManualProvisioningFixture(string connectionString, string eventHubEndpoint)
```
damonbarry, line 22: These two constructors can be one constructor like this:
```csharp
public ManualProvisioningFixture(
    string connectionString = Context.Current.ConnectionString,
    string eventHubEndpoint = Context.Current.EventHubEndpoint)
{
    this.iotHub = new IotHub(connectionString, eventHubEndpoint, Context.Current.Proxy);
}
```

test/Microsoft.Azure.Devices.Edge.Test/helpers/SasManualProvisioningFixture.cs
```diff
@@ -13,6 +11,16 @@ public class SasManualProvisioningFixture : ManualProvisioningFixture
     {
         protected EdgeRuntime runtime;
 
+        public SasManualProvisioningFixture()
```
damonbarry, line 14: I think this empty constructor is unnecessary

test/Microsoft.Azure.Devices.Edge.Test.Common/LeafDevice.cs
dylanbronson: Unfortunately, no - default params need to be compile-time constant.
I could make it string modelId = null as default then do a null check, but we usually prefer to use Option in our codebase. That being said it is in the test side, so it could be a time to break that rule.

damonbarry: Oh yeah...no leave it like this. No big deal.

test/Microsoft.Azure.Devices.Edge.Test/helpers/ManualProvisioningFixture.cs
dylanbronson: Same problem as above - Using that method, the compiler complains because Context.Current.* aren't compile-time constant

test/Microsoft.Azure.Devices.Edge.Test/helpers/SasManualProvisioningFixture.cs
dylanbronson: When I remove this, compiler complains for other tests that were using SasManualProvisioningFixture without specifying the parameters.

Because I added a constructor with parameters in the ManualProvisioningFixture, C# won't create a default, paramater-less constructor automatically anymore. So I either need to keep this or add constructors with parameters in all the classes that use it. (for example adding a constructor to Module.cs with connection string and event hub endpoint string)

(that is my best guess at what's happening)

damonbarry: Well just ignore me today. I don't know what I'm talking about...

PR #3243, opened 2020-07-17
[mqtt] Make server certificate optional in config

dmolokanov:
In `edgehub` mode broker always tries to download server certificate from edgelet which is troublesome during development.

Make a path to a server certificate optional for both modes. When the `certificate` is missing in the config
* in `edgehub` mode it skips downloading server certificate from edgelet and tries to load it from a file
* in `generic` mode it returns an error

### DIFF
--- a/mqtt/mqtt-edgehub/src/tls.rs
+++ b/mqtt/mqtt-edgehub/src/tls.rs
@@ -1,8 +1,15 @@
 use chrono::{DateTime, NaiveDateTime, ParseResult, Utc};
+use native_tls::Identity;
 use openssl::{asn1::Asn1TimeRef, pkcs12::Pkcs12, pkey::PKey, stack::Stack, x509::X509};
+use std::{
+    convert::TryFrom,
+    fs,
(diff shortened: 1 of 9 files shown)


mqtt/mqtt-edgehub/src/tls.rs
```diff
@@ -1,16 +1,26 @@
...
+#[derive(Debug)]
 pub struct ServerCertificate {
     der: Vec<u8>,
     not_before: DateTime<Utc>,
     not_after: DateTime<Utc>,
 }
 
 impl ServerCertificate {
-    pub fn try_from<S, K>(certificate: S, private_key: K) -> Result<Self, ServerCertificateError>
+    pub fn from_pem_pair<S, K>(
```
vadim-kovalyov, line 20: Will this method be used anywhere except tests?

dmolokanov: no, it will be used when server cert downloaded from edgelet.

Issue #2066, opened 2019-12-04
IoT Edge Runtime Response: Error 406


lucasFigueiredoEE:
## Expected Behavior
IoT Edge Runtime Response should be OK and an IoT Edge Custom Module should be running.

## Current Behavior
IoT Edge Runtime Response: 406 -- The device is offline or not sending status reports.
IoT Edge Custom Module: Not reported by device.

## Steps to Reproduce
Provide a detailed set of steps to reproduce the bug.
1. Install IoT Edge in a computer and Deploy it with an IoT Hub connection string
2. Add the module OPC Publisher in the Azure Portal
3. Run iotedge list
4. Check IoT Edge Runtime Response for your device

## Context (Environment)
Windows 10 Pro. Build 17763.

### Output of `iotedge check`

<details>
<summary>Click here</summary>

```

 iotedge check
Configuration checks
--------------------
√ config.yaml is well-formed - OK
√ config.yaml has well-formed connection string - OK
√ container engine is installed and functional - OK
√ Windows host version is supported - OK
√ config.yaml has correct hostname - OK
√ config.yaml has correct URIs for daemon mgmt endpoint - OK
√ latest security daemon - OK
‼ host time is close to real time - Warning
    Could not query NTP server
√ container time is close to host time - OK
‼ DNS server - Warning
    Container engine is not configured with DNS server setting, which may impact connectivity to IoT Hub.
    Please see https://aka.ms/iotedge-prod-checklist-dns for best practices.
    You can ignore this warning if you are setting DNS server per module in the Edge deployment.
‼ production readiness: certificates - Warning
    Device is using self-signed, automatically generated certs.
    Please see https://aka.ms/iotedge-prod-checklist-certs for best practices.
√ production readiness: certificates expiry - OK
√ production readiness: container engine - OK
‼ production readiness: logs policy - Warning
    Container engine is not configured to rotate module logs which may cause it run out of disk space.
    Please see https://aka.ms/iotedge-prod-checklist-logs for best practices.
    You can ignore this warning if you are setting log policy per module in the Edge deployment.

Connectivity checks
-------------------
√ host can connect to and perform TLS handshake with IoT Hub AMQP port - OK
√ host can connect to and perform TLS handshake with IoT Hub HTTPS / WebSockets port - OK
√ host can connect to and perform TLS handshake with IoT Hub MQTT port - OK
√ container on the IoT Edge module network can connect to IoT Hub AMQP port - OK
[...]

PR #3244, opened 2020-07-17
Agent reported state as "406" when modules are in backoff.

darobs:
To reproduce

1.    create a deployment which has a module which fails to launch, like an incorrect image name, or some weird mount Docker doesn't like.
2.    Look at the reported status of the edge runtime - If you are lucky, you will see the real error.
3.    Let the runtime run for long enough that the "OrderedRetryPlanRunner" puts the module on backoff.
4.    Look at the reported status of the edge runtime - You won't see the real error message from the failure - you will see a "406 -- The device is offline or not sending status reports"

The root cause is that every reconcile starts with status=Unknown(406), and when a module is in backoff, the status doesn't update.

So it goes:
reconcile loop 1: module command fails, error is reported
reconcile loop 2: module is in backoff now, Unknown is reported
...
reconcile loop n: module command is attempted again, error is reported
reconcile loop n+1: module is in backoff again, Unknown is reported.

User is more likely to see "406 -- The device is offline or not sending status reports" which is not cool.
This has been seen before, one example: https://github.com/Azure/iotedge/issues/2066

The fix here is to keep track of status in Agent and report current status until it changes (either via a new error

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Agent.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Agent.cs
@@ -33,6 +33,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Core
         readonly IAvailabilityMetric availabilityMetric;
         IEnvironment environment;
         DeploymentConfigInfo currentConfig;
+        DeploymentStatus status;
 
         public Agent(
             IConfigSource configSource,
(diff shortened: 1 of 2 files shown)


damonbarry:
@darobs There are several `ReconcileAsync` tests in `AgentTests.cs` which mock the reporter and verify the passed-in status. Do these still pass, and do we need to modify or add to them?

darobs:
> @darobs There are several `ReconcileAsync` tests in `edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Core.Test/AgentTests.cs` which mock the reporter and verify the passed-in status. Do these still pass, and do we need to modify or add to them?

There were no unit tests where `ExecuteAsync` returns false.  I added two.

PR #3246, opened 2020-07-18
[mqtt] Fix error when parsing unix path

dmolokanov:
Fixes error when parsing `unix://workload.sock` path.
Adds `Content-Type: application/json` for authenticator HTTP call

### DIFF
--- a/mqtt/mqtt-edgehub/src/edgelet/mod.rs
+++ b/mqtt/mqtt-edgehub/src/edgelet/mod.rs
@@ -2,36 +2,30 @@ mod connect;
 mod workload;
 
 pub use connect::Connector;
-pub use workload::*;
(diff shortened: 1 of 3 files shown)


mqtt/mqtt-edgehub/src/edgelet/mod.rs
```diff
@@ -2,36 +2,30 @@ mod connect;
 mod workload;
 
 pub use connect::Connector;
-pub use workload::*;
+pub use workload::{CertificateResponse, ServerCertificateRequest, WorkloadClient};
```
and-rewsmith, line 5: Where are these used?

dmolokanov: they are exposed (re-exported) outside of the edgelet module

PR #3247, opened 2020-07-18
[k8s] Map docker "IpcMode" to k8s "HostIPC"

darobs:
Add IpcMode to HostConfig, if IpcMode is "host", set the podspec "hostIpc" to true

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/Constants.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/Constants.cs
@@ -82,5 +82,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
         public const string RunAsNonRootKey = "RunAsNonRoot";
 
         public const string UnknownImage = "unknown";
+
+        public const string HostIPC = "host";
     }
 }
--- a/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesDeploymentMapperTest.cs
+++ b/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesDeploymentMapperTest.cs
@@ -40,6 +40,16 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test
             }
         };
 
+        static readonly HostConfig HostIpcModeHostConfig = new HostConfig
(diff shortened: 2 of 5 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/Constants.cs
```diff
@@ -82,5 +82,7 @@ public static class EdgeDeployment
         public const string RunAsNonRootKey = "RunAsNonRoot";
 
         public const string UnknownImage = "unknown";
+
+        public const string HostIPC = "host";
     }
```
ggjjj, line 87: Do we set **host** as the default ipc mode?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesDeploymentMapperTest.cs
```diff
@@ -40,6 +40,16 @@ public class KubernetesDeploymentMapperTest
...
 
+        static readonly HostConfig HostIpcModeHostConfig = new HostConfig
+        {
+            IpcMode = "host"
+        };
+
+        static readonly HostConfig PrivateIpcModeHostConfig = new HostConfig
+        {
+            IpcMode = "private"
+        };
```
ggjjj, line 51: does **shareable** IPC mode make sense for k8s scenario?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/Constants.cs
davebob3: Default depends on the Docker daemon, but the typical default is "private" if it's not set.  I.e. each docker container gets its own shared memory.  K8s has a boolean, so it either "hostIPC" or not. I assumed the false case defaults to "private", but did not confirm.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesDeploymentMapperTest.cs
darobs: For unit tests, I really only needed IpcMode not set, set to host, and not set to host, since those are the only cases relevant to k8s API settings.

PR #3248, opened 2020-07-20
GetDeviceAndModuleOnBehalfOf in EdgeHub

richma-ms:
This change adds logic to use the new GetDeviceAndModuleOnBehalfOf API when calling upstream, and exposes the same REST endpoint for downstream Edge devices.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
         const int RetryCount = 2;
 
         const string GetDevicesAndModulesInTargetScopeUriFormat = "/devices/{0}/modules/{1}/devicesAndModulesInTargetDeviceScope?api-version={2}";
+        const string GetDeviceAndModuleOnBehalfOfUriFormat = "/devices/{0}/modules/{1}/getDeviceAndModuleOnBehalfOf?api-version={2}";
 
         const string NestedApiVersion = "2020-06-30-preview";
 
@@ -36,7 +37,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
         readonly Option<string> continuationToken;
         readonly int batchSize;
         readonly ITokenProvider edgeHubTokenProvider;
-        readonly IServiceIdentityHierarchy serviceIdentityTree;
+        readonly IServiceIdentityHierarchy serviceIdentityHierarchy;
         readonly Option<IWebProxy> proxy;
 
         public string TargetEdgeDeviceId { get; }
@@ -75,27 +76,35 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             this.continuationToken = Preconditions.CheckNotNull(continuationToken);
             this.batchSize = Preconditions.CheckRange(batchSize, 0, 1000, nameof(batchSize));
             this.edgeHubTokenProvider = Preconditions.CheckNotNull(edgeHubTokenProvider, nameof(edgeHubTokenProvider));
-            this.serviceIdentityTree = Preconditions.CheckNotNull(serviceIdentityTree, nameof(serviceIdentityTree));
+            this.serviceIdentityHierarchy = Preconditions.CheckNotNull(serviceIdentityTree, nameof(serviceIdentityTree));
             this.proxy = Preconditions.CheckNotNull(proxy, nameof(proxy));
             this.retryStrategy = retryStrategy ?? TransientRetryStrategy;
         }
 
         public Task<ScopeResult> GetIdentitiesInScopeAsync() =>
-            this.GetIdentitiesInTargetScopeWithRetry(this.GetServiceUri(), Option.None<string>());
+            this.GetIdentitiesInTargetScopeWithRetry(this.GetNestedScopeServiceUri(), Option.None<string>());
 
         public Task<ScopeResult> GetNextAsync(string continuationToken) =>
-            this.GetIdentitiesInTargetScopeWithRetry(this.GetServiceUri(), Option.Some(continuationToken));
+            this.GetIdentitiesInTargetScopeWithRetry(this.GetNestedScopeServiceUri(), Option.Some(continuationToken));
 
-        public Task<ScopeResult> GetIdentityAsync(string _, string __)
+        public Task<ScopeResult> GetIdentityAsync(string targetDeviceId, string targetModuleId)
         {
-            // TODO: 7026875: Refreshing scopes on Connect
-            return Task.FromResult(new ScopeResult(Enumerable.Empty<Device>(), Enumerable.Empty<Module>(), string.Empty));
+            Preconditions.CheckNonWhiteSpace(targetDeviceId, nameof(targetDeviceId));
+            return this.GetIdentityOnBehalfOfWithRetry(this.GetIdentityOnBehalfOfServiceUri(), targetDeviceId, Option.Maybe(targetModuleId));
         }
 
-        internal Uri GetServiceUri()
+        internal Uri GetNestedScopeServiceUri()
         {
             // The URI is always always in the context of the actor device
             string relativeUri = GetDevicesAndModulesInTargetScopeUriFormat.FormatInvariant(this.actorEdgeDeviceId, this.moduleId, NestedApiVersion);
+            var uri = new Uri(this.iotHubBaseHttpUri, relativeUri);
+            return uri;
+        }
+
+        internal Uri GetIdentityOnBehalfOfServiceUri()
+        {
+            // The URI is always always in the context of the actor device
+            string relativeUri = GetDeviceAndModuleOnBehalfOfUriFormat.FormatInvariant(this.actorEdgeDeviceId, this.moduleId, NestedApiVersion);
             var uri = new Uri(this.iotHubBaseHttpUri, relativeUri);
             return uri;
         }
@@ -105,6 +114,22 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             var transientRetryPolicy = new RetryPolicy(TransientErrorDetectionStrategy, retryStrategy);
             transientRetryPolicy.Retrying += (_, args) => onRetry(args);
             return transientRetryPolicy.ExecuteAsync(func);
+        }
+
+        async Task<ScopeResult> GetIdentityOnBehalfOfWithRetry(Uri uri, string deviceId, Option<string> moduleId)
+        {
+            try
+            {
+                return await ExecuteWithRetry(
+                    () => this.GetIdentityOnBehalfOf(uri, deviceId, moduleId),
+                    Events.RetryingGetIdentities,
+                    this.retryStrategy);
+            }
+            catch (Exception e)
+            {
+                Events.ErrorInScopeResult(e);
+                throw;
+            }
         }
 
         async Task<ScopeResult> GetIdentitiesInTargetScopeWithRetry(Uri uri, Option<string> continuationToken)
@@ -126,6 +151,38 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             }
         }
 
+        async Task<ScopeResult> GetIdentityOnBehalfOf(Uri uri, string deviceId, Option<string> moduleId)
+        {
+            HttpClient client = this.proxy
+                .Map(p => new HttpClient(new HttpClientHandler { Proxy = p }, disposeHandler: true))
+                .GetOrElse(() => new HttpClient());
+            using (var msg = new HttpRequestMessage(HttpMethod.Post, uri))
+            {
+                // Get the auth-chain for the target device
+                Option<string> maybeAuthChain = await this.serviceIdentityHierarchy.GetEdgeAuthChain(deviceId);
+                string authChain = maybeAuthChain.Expect(() => new InvalidOperationException($"No valid authentication chain for {deviceId}"));
+
+                var payload = new IdentityOnBehalfOfRequest(deviceId, moduleId.OrDefault(), authChain);
+                string token = await this.edgeHubTokenProvider.GetTokenAsync(Option.None<TimeSpan>());
+                msg.Headers.Add(HttpRequestHeader.Authorization.ToString(), token);
+                msg.Headers.Add(Constants.ServiceApiIdHeaderKey, this.actorEdgeDeviceId + "/" + this.moduleId);
+                msg.Content = new StringContent(JsonConvert.SerializeObject(payload), Encoding.UTF8, "application/json");
+
+                HttpResponseMessage response = await client.SendAsync(msg);
+                string content = await response.Content.ReadAsStringAsync();
+                if (response.IsSuccessStatusCode)
+                {
+                    var scopeResult = JsonConvert.DeserializeObject<ScopeResult>(content);
+                    Events.GotValidResult();
+                    return scopeResult;
+                }
+                else
+                {
+                    throw new DeviceScopeApiException("Error getting device scope result from upstream", response.StatusCode, content);
+                }
+            }
+        }
+
         async Task<ScopeResult> GetIdentitiesInTargetScope(Uri uri, Option<string> continuationToken)
         {
             HttpClient client = this.proxy
@@ -134,7 +191,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             using (var msg = new HttpRequestMessage(HttpMethod.Post, uri))
             {
                 // Get the auth-chain for the target device
-                Option<string> maybeAuthChain = await this.serviceIdentityTree.GetAuthChain(this.TargetEdgeDeviceId);
+                Option<string> maybeAuthChain = await this.serviceIdentityHierarchy.GetAuthChain(this.TargetEdgeDeviceId);
                 string authChain = maybeAuthChain.Expect(() => new InvalidOperationException($"No valid authentication chain for {this.TargetEdgeDeviceId}"));
 
                 var payload = new NestedScopeRequest(this.batchSize, continuationToken.OrDefault(), authChain);
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
@@ -49,10 +49,10 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             Option<string> continuationToken,
             int batchSize,
             ITokenProvider edgeHubTokenProvider,
-            IServiceIdentityHierarchy serviceIdentityTree,
+            IServiceIdentityHierarchy serviceIdentityHierarchy,
             Option<IWebProxy> proxy,
             RetryStrategy retryStrategy = null)
-            : this(iotHubHostName, deviceId, deviceId, moduleId, continuationToken, batchSize, edgeHubTokenProvider, serviceIdentityTree, proxy, retryStrategy)
+            : this(iotHubHostName, deviceId, deviceId, moduleId, continuationToken, batchSize, edgeHubTokenProvider, serviceIdentityHierarchy, proxy, retryStrategy)
         {
         }
 
@@ -64,7 +64,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             Option<string> continuationToken,
             int batchSize,
             ITokenProvider edgeHubTokenProvider,
-            IServiceIdentityHierarchy serviceIdentityTree,
+            IServiceIdentityHierarchy serviceIdentityHierarchy,
             Option<IWebProxy> proxy,
             RetryStrategy retryStrategy = null)
         {
@@ -76,7 +76,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
             this.continuationToken = Preconditions.CheckNotNull(continuationToken);
             this.batchSize = Preconditions.CheckRange(batchSize, 0, 1000, nameof(batchSize));
             this.edgeHubTokenProvider = Preconditions.CheckNotNull(edgeHubTokenProvider, nameof(edgeHubTokenProvider));
-            this.serviceIdentityHierarchy = Preconditions.CheckNotNull(serviceIdentityTree, nameof(serviceIdentityTree));
+            this.serviceIdentityHierarchy = Preconditions.CheckNotNull(serviceIdentityHierarchy, nameof(serviceIdentityHierarchy));
             this.proxy = Preconditions.CheckNotNull(proxy, nameof(proxy));
             this.retryStrategy = retryStrategy ?? TransientRetryStrategy;
         }
@@ -95,7 +95,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
 
         internal Uri GetNestedScopeServiceUri()
         {
-            // The URI is always always in the context of the actor device
+            // The URI is always in the context of the actor device
             string relativeUri = GetDevicesAndModulesInTargetScopeUriFormat.FormatInvariant(this.actorEdgeDeviceId, this.moduleId, NestedApiVersion);
             var uri = new Uri(this.iotHubBaseHttpUri, relativeUri);
             return uri;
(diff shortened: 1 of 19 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -126,6 +151,38 @@ async Task<ScopeResult> GetIdentitiesInTargetScopeWithRetry(Uri uri, Option<stri
...
+                .Map(p => new HttpClient(new HttpClientHandler { Proxy = p }, disposeHandler: true))
+                .GetOrElse(() => new HttpClient());
+            using (var msg = new HttpRequestMessage(HttpMethod.Post, uri))
+            {
+                // Get the auth-chain for the target device
+                Option<string> maybeAuthChain = await this.serviceIdentityHierarchy.GetEdgeAuthChain(deviceId);
+                string authChain = maybeAuthChain.Expect(() => new InvalidOperationException($"No valid authentication chain for {deviceId}"));
+
+                var payload = new IdentityOnBehalfOfRequest(deviceId, moduleId.OrDefault(), authChain);
+                string token = await this.edgeHubTokenProvider.GetTokenAsync(Option.None<TimeSpan>());
```
huguesBouvier, line 166: Does this means it awaits forever? If we lose connection does that mean threads will accumulate here?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -91,33 +91,77 @@ public async Task<IList<string>> GetAllIds()
...
             using (await this.nodesLock.LockAsync())
             {
                 // Auth-chain for a child somewhere in the tree
-                if (this.nodes.TryGetValue(id, out ServiceIdentityTreeNode treeNode))
+                if (this.nodes.TryGetValue(targetId, out ServiceIdentityTreeNode treeNode))
                 {
+                    // Check every Edge device in the authchain for disabled devices
+                    string[] authChainIds = treeNode.AuthChain.Map(chain => chain.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries)).OrDefault();
+
+                    foreach (string chainId in authChainIds)
```
huguesBouvier, line 116: I suppose that this is what we discussed last time, but why not do that in the function validate auth chain? 
You have a loop very similar there.

richma-ms: because this validation requires that we have access to the cache, so that we can check the individual identities.  The authchain validation function is really more of a format check, and doesn't contain any logic specific to the metadata in each identity of the link.

---
In reply to: [457535659](https://github.com/Azure/iotedge/pull/3248#discussion_r457535659) [](ancestors = 457535659)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
richma-ms: this is not the timeout value, it's the expiration for the token.  Passing in None results in the default expiration being used.

---
In reply to: [457529858](https://github.com/Azure/iotedge/pull/3248#discussion_r457529858) [](ancestors = 457529858)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -75,38 +76,62 @@ public class NestedDeviceScopeApiClient : IDeviceScopeApiClient
             this.continuationToken = Preconditions.CheckNotNull(continuationToken);
             this.batchSize = Preconditions.CheckRange(batchSize, 0, 1000, nameof(batchSize));
             this.edgeHubTokenProvider = Preconditions.CheckNotNull(edgeHubTokenProvider, nameof(edgeHubTokenProvider));
-            this.serviceIdentityTree = Preconditions.CheckNotNull(serviceIdentityTree, nameof(serviceIdentityTree));
+            this.serviceIdentityHierarchy = Preconditions.CheckNotNull(serviceIdentityTree, nameof(serviceIdentityTree));
```
philipktlin, line 79:
```suggestion
            this.serviceIdentityHierarchy = Preconditions.CheckNotNull(serviceIdentityHierarchy, nameof(serviceIdentityTree));
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -75,38 +76,62 @@ public class NestedDeviceScopeApiClient : IDeviceScopeApiClient
...
         {
             // The URI is always always in the context of the actor device
             string relativeUri = GetDevicesAndModulesInTargetScopeUriFormat.FormatInvariant(this.actorEdgeDeviceId, this.moduleId, NestedApiVersion);
             var uri = new Uri(this.iotHubBaseHttpUri, relativeUri);
             return uri;
         }
 
+        internal Uri GetIdentityOnBehalfOfServiceUri()
+        {
+            // The URI is always always in the context of the actor device
```
philipktlin, line 106:
```suggestion
            // The URI is always in the context of the actor device
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -75,38 +76,62 @@ public class NestedDeviceScopeApiClient : IDeviceScopeApiClient
...
-            // TODO: 7026875: Refreshing scopes on Connect
-            return Task.FromResult(new ScopeResult(Enumerable.Empty<Device>(), Enumerable.Empty<Module>(), string.Empty));
+            Preconditions.CheckNonWhiteSpace(targetDeviceId, nameof(targetDeviceId));
+            return this.GetIdentityOnBehalfOfWithRetry(this.GetIdentityOnBehalfOfServiceUri(), targetDeviceId, Option.Maybe(targetModuleId));
         }
 
-        internal Uri GetServiceUri()
+        internal Uri GetNestedScopeServiceUri()
         {
             // The URI is always always in the context of the actor device
```
philipktlin, line 98:
```suggestion
            // The URI is always in the context of the actor device
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -39,7 +39,18 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId)
             Option<ScopeResult> scopeResult = Option.None<ScopeResult>();
             try
             {
-                ScopeResult res = await this.securityScopesApiClientProvider.CreateDeviceScopeClient().GetIdentityAsync(deviceId, null);
+                IDeviceScopeApiClient client;
+
+                if (this.nestedEdgeEnabled)
+                {
+                    client = this.securityScopesApiClientProvider.CreateNestedDeviceScopeClient(Option.None<string>());
```
philipktlin, line 46: just curious, I can't find any usage of this method with non-empty value.  In what scenario we will pass continuation link to this method?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core
+{
+    using System;
+
+    public class AuthChainHelpers
```
philipktlin, line 6:
```suggestion
    public static class AuthChainHelpers
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
```diff
@@ -138,7 +138,7 @@ async Task<(bool isAuthenticated, bool serviceIdentityFound)> AuthenticateWithAu
             Option<string> authChain = await this.deviceScopeIdentitiesCache.GetAuthChain(credentials.Identity.Id);
```
philipktlin, line 138: I found that argument syncServiceIdentity is not used.  Do we need it?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -103,49 +103,9 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string id)
             return await this.GetServiceIdentityInternal(id);
         }
 
-        public async Task<Option<string>> GetAuthChain(string targetId)
-        {
-            Option<string> authChain = await this.serviceIdentityHierarchy.GetAuthChain(targetId);
+        public Task<Option<string>> GetAuthChain(string targetId) => this.serviceIdentityHierarchy.GetAuthChain(targetId);
 
-            // Check every Edge device in the authchain for disabled devices
```
philipktlin, line 110: We don't need to check disabled devices here, where does this logic go?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -91,33 +91,80 @@ public async Task<IList<string>> GetAllIds()
             }
         }
```
philipktlin, line 93: Should we rename this class to ServiceIdentityHierarchy to match IServiceIdentityHierarchy?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -91,33 +91,80 @@ public async Task<IList<string>> GetAllIds()
...
             }
-            else if (id == this.actorDeviceId + "/$edgeHub")
+            else if (targetId == this.actorDeviceId + "/" + Constants.EdgeHubModuleId)
             {
-                // TODO: richma - make this better
-                return Option.Some(this.actorDeviceId + "/$edgeHub" + ";" + this.actorDeviceId);
+                return Option.Some(this.actorDeviceId + "/" + Constants.EdgeHubModuleId + ";" + this.actorDeviceId);
             }
 
             using (await this.nodesLock.LockAsync())
```
philipktlin, line 108: it is able to avoid running these logic for non nested-edge scenario.

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/DeviceScopeControllerTest.cs
```diff
@@ -76,27 +76,131 @@ public async Task NestedDeviceScopeRoundtripTest()
...
+            string generationId = "generation1";
+            string primaryKey = "t3LtII3CppvtVqycKp9bo043vCEgWbGBJAzXZNmoBXo=";
+            string secondaryKey = "kT4ac4PpH5UY0vA1JpLQWOu2yG6qKoqwvzee3j1Z3bA=";
+            var authentication = new ServiceAuthentication(new SymmetricKeyAuthentication(primaryKey, secondaryKey));
+            var resultDeviceIdentity = new ServiceIdentity(childEdgeId, null, deviceScope, new List<string>() { parentScope }, generationId, Enumerable.Empty<string>(), authentication, ServiceIdentityStatus.Enabled);
+            var resultModuleIdentity = new ServiceIdentity(childEdgeId, moduleId, null, new List<string>() { deviceScope }, generationId, Enumerable.Empty<string>(), authentication, ServiceIdentityStatus.Enabled);
+            var resultIdentities = new List<ServiceIdentity>() { resultDeviceIdentity, resultModuleIdentity };
+            var authChainMapping = new Dictionary<string, string>();
+            string targetId = childEdgeId + "/" + moduleId;
+            authChainMapping.Add(targetId, targetId + ";edge2;edge1;edgeroot");
```
philipktlin, line 111:
```suggestion
            authChainMapping.Add(targetId, $"{targetId};{childEdgeId};{parentEdgeId};edgeroot");
```

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Http.Test/DeviceScopeControllerTest.cs
```diff
@@ -76,27 +76,131 @@ public async Task NestedDeviceScopeRoundtripTest()
...
+
+            var responseActualBytes = GetResponseBodyBytes(controller);
+            var responseActualJson = Encoding.UTF8.GetString(responseActualBytes);
+
+            Assert.Equal((int)HttpStatusCode.OK, controller.HttpContext.Response.StatusCode);
+            Assert.Equal(responseExpectedJson, responseActualJson);
+        }
+
+        [Fact]
+        public void ValidateChainTest()
```
philipktlin, line 171: I think it is better to move this test to another file "AuthChainHelpersTest.cs"

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/ServiceIdentityTreeTest.cs
```diff
@@ -55,7 +55,7 @@ internal static ServiceIdentity CreateServiceIdentity(string deviceId, string mo
                 "1234",
                 capabilities,
                 new ServiceAuthentication(ServiceAuthenticationType.None),
-                ServiceIdentityStatus.Enabled);
+                isEnabled ? ServiceIdentityStatus.Enabled : ServiceIdentityStatus.Disabled);
```
philipktlin, line 58: ServiceIdentityStatus only contains Enabled and Disabled enum values.  Do you know why we use enum type for ServiceIdentityStatus, but not boolean?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -32,102 +33,167 @@ public async Task GetDevicesAndModulesInTargetDeviceScope([FromRoute] string act
...
+        {
+            this.Response.StatusCode = (int)result.Status;
+
+            if (result.Status == HttpStatusCode.OK)
+            {
+                var resultJsonContent = JsonConvert.SerializeObject(result);
+                var resultUtf8Bytes = Encoding.UTF8.GetBytes(resultJsonContent);
+
+                this.Response.ContentLength = resultUtf8Bytes.Length;
+                this.Response.ContentType = SupportedContentType;
```
philipktlin, line 191:
```suggestion
                this.Response.ContentType = ApplicationJsonContentType;
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -32,102 +33,167 @@ public async Task GetDevicesAndModulesInTargetDeviceScope([FromRoute] string act
...
             result.Status = HttpStatusCode.OK;
 
             return result;
         }
 
-        async Task SendResponse(EdgeHubScopeResult result)
+        async Task<EdgeHubScopeResult> HandleGetDeviceAndModuleOnBehalfOf(string actorDeviceId, string actorModuleId, IdentityOnBehalfOfRequest request)
         {
-            this.Response.StatusCode = (int)result.Status;
+            Events.ReceivedIdentityOnBehalfOfRequest(actorDeviceId, request);
```
philipktlin, line 90: do we need to pass actorModuleId as well?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -32,102 +33,167 @@ public async Task GetDevicesAndModulesInTargetDeviceScope([FromRoute] string act
...
+                {
+                    targetDeviceId = ids[0];
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        async Task<bool> AuthorizeActor(IEdgeHub edgeHub, string actorDeviceId, string actorModuleId, string targetId)
```
philipktlin, line 155:
```suggestion
        async Task<bool> AuthorizeActorAsync(IEdgeHub edgeHub, string actorDeviceId, string actorModuleId, string targetId)
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -32,102 +33,167 @@ public async Task GetDevicesAndModulesInTargetDeviceScope([FromRoute] string act
...
+            if (!targetAuthChain.HasValue)
             {
                 return false;
             }
 
-            // The target device ID is the first device ID in the chain, which could be a module ID
-            // of the format "deviceId/moduleId"
-            var ids = actorAuthChainIds[0].Split('/', StringSplitOptions.RemoveEmptyEntries).ToList();
-            targetDeviceId = ids[0];
+            // Validate the the target auth-chain
```
philipktlin, line 172:
```suggestion
            // Validate the target auth-chain
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -32,102 +33,167 @@ public async Task GetDevicesAndModulesInTargetDeviceScope([FromRoute] string act
...
+            if (!AuthChainHelpers.ValidateAuthChain(actorDeviceId, targetId, targetAuthChain.Expect(() => new InvalidOperationException())))
```
philipktlin, line 173: line 167 already validate targetAuthChain has value.  Use OrDefault or refactor to use ForEach.
```suggestion
            if (!AuthChainHelpers.ValidateAuthChain(actorDeviceId, targetId, targetAuthChain.OrDefault()))
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -32,102 +33,167 @@ public async Task GetDevicesAndModulesInTargetDeviceScope([FromRoute] string act
...
+            // If the target is a module, we also need to
+            // include the parent device as well to match
+            // IoT Hub API behavior
+            if (isModule)
+            {
+                Option<ServiceIdentity> device = await edgeHub.GetIdentityAsync(request.TargetDeviceId);
+                device.ForEach(i => identityList.Add(i));
             }
+
+            Events.SendingScopeResult(identityList);
```
philipktlin, line 126: do we need to log scope result with inputs e.g. actor device id, actor module id, and IdentityOnBehalfOfRequest?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -32,102 +33,167 @@ public async Task GetDevicesAndModulesInTargetDeviceScope([FromRoute] string act
...
+            // Construct the result from the identities
             Events.SendingScopeResult(identities);
             result = MakeResultFromIdentities(identities);
             result.Status = HttpStatusCode.OK;
 
             return result;
         }
 
-        async Task SendResponse(EdgeHubScopeResult result)
+        async Task<EdgeHubScopeResult> HandleGetDeviceAndModuleOnBehalfOf(string actorDeviceId, string actorModuleId, IdentityOnBehalfOfRequest request)
```
philipktlin, line 88:
```suggestion
        async Task<EdgeHubScopeResult> HandleGetDeviceAndModuleOnBehalfOfAsync(string actorDeviceId, string actorModuleId, IdentityOnBehalfOfRequest request)
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeAuthenticator.cs
richma-ms: that's a bug, it should actually get passed down to AuthenticateWithServiceIdentity.  Good catch =)

---
In reply to: [458489677](https://github.com/Azure/iotedge/pull/3248#discussion_r458489677) [](ancestors = 458489677)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
richma-ms: originally it was IServiceIdentityTree, but we wanted to separate the interface definition from the implementation detail.

For example, the hierarchy can be backed by something other than a tree, as in the case of multiple-parents in the future, it'll be graph.  But interface contracts won't really need to change.

---
In reply to: [458492036](https://github.com/Azure/iotedge/pull/3248#discussion_r458492036) [](ancestors = 458492036)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
richma-ms: there won't be a config differentiation between nested and non-nested when we GA.  Legacy non-nested configuration is just nested with 1 level.  So there isn't really a way to diverge the behavior at runtime.  Doing that would add more code complexity anyways, it's arguably not worthwhile from a maintanence PoV.

---
In reply to: [458495127](https://github.com/Azure/iotedge/pull/3248#discussion_r458495127) [](ancestors = 458495127)

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/ServiceIdentityTreeTest.cs
richma-ms: I don't know the history for sure, but if I have to guess I'd imagine it's because we mirror Device class from the SDK, which also uses an enum instead of a boolean.

---
In reply to: [458523533](https://github.com/Azure/iotedge/pull/3248#discussion_r458523533) [](ancestors = 458523533)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
richma-ms: hmm....I think this was an oversight from an earlier refactor.  A previous implementation of the nested logic in ServiceProxy only had one creation method for the HTTP client, so it took a continuation token.  Then later the nested portion was moved to CreateOnBehalfOf, so nobody calls this with a continuation token anymore.  I can remove it.

---
In reply to: [458480574](https://github.com/Azure/iotedge/pull/3248#discussion_r458480574) [](ancestors = 458480574)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
richma-ms: where is ApplicationJsonContentType defined?

---
In reply to: [458524456](https://github.com/Azure/iotedge/pull/3248#discussion_r458524456) [](ancestors = 458524456)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
richma-ms: actor is probably extraneous, considering that we're logging it at the beginning of this method, but the target id can be helpful, I'll add it.

---
In reply to: [458528369](https://github.com/Azure/iotedge/pull/3248#discussion_r458528369) [](ancestors = 458528369)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
richma-ms: imo this is more defensive.  If the check on 167 gets changed/removed later, this will immediately throw an exception.  Whereas using OrDefault could potentially lead to a silent failure.

---
In reply to: [458526939](https://github.com/Azure/iotedge/pull/3248#discussion_r458526939) [](ancestors = 458526939)

PR #3249, opened 2020-07-20
Broker workload client fixes

and-rewsmith:
Two fixes:
- Expect 201 response
- Fix env var name casing


and-rewsmith:
I initially was going to treat any 201 response the same, but now think that we should keep it consistent with the workload client in edgehub:
https://github.com/Azure/iotedge/blob/master/edge-util/src/Microsoft.Azure.Devices.Edge.Util/edged/version_2018_06_28/generatedCode/HttpWorkloadClient.cs#L608

PR #3251, opened 2020-07-20
EdgeHub: Hardcode broker state directory to /tmp/mqttd

and-rewsmith:
The edgehubuser inside the edgehub container only has permissions to write to /tmp. The dotnet application creates  /tmp/edgehub folder and writes to it. The broker should follow this pattern and create a storage directory /tmp/mqttd and write to it.

### DIFF
--- a/mqtt/mqttd/src/broker/mod.rs
+++ b/mqtt/mqttd/src/broker/mod.rs
@@ -2,7 +2,8 @@ mod bootstrap;
 mod shutdown;
 mod snapshot;
 
-use std::env;
+use std::fs;
+use std::path::PathBuf;
 
 use anyhow::Result;
 use futures_util::pin_mut;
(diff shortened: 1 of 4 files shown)


and-rewsmith:
When the build fix PRs go in we need to fix the pfx write location to /tmp/mqttd

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -18,10 +18,15 @@ use mqtt_broker::{
 };
 use mqtt_broker_core::auth::Authorizer;
 
+const MQTTD_STATE_DIR: &str = "/tmp/mqttd";
+
 pub async fn run(config: BrokerConfig) -> Result<()> {
     info!("loading state...");
-    let state_dir = env::current_dir().expect("can't get cwd").join("state");
-    let mut persistor = FilePersistor::new(state_dir, VersionedFileFormat::default());
+    fs::create_dir(MQTTD_STATE_DIR).expect("can't create mqttd state dir");
```
vadim-kovalyov, line 25:
```suggestion
    fs::create_dir(MQTTD_STATE_DIR).expect(format!("can't create mqttd state dir {}", MQTTD_STATE_DIR));
```

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -18,10 +18,15 @@ use mqtt_broker::{
 };
 use mqtt_broker_core::auth::Authorizer;
 
+const MQTTD_STATE_DIR: &str = "/tmp/mqttd";
```
dmolokanov, line 21: unfortunately, it's not THAT simple :( also it will break perf tests.
it should be handled by the config section which was under refactoring now.

and-rewsmith:
@dmolokanov I saw a comment earlier that you didn't think this change was this simple. Can you confirm? I don't see the comment anymore. If so I will close and re-evaluate approach for public preview.

PR #3252, opened 2020-07-20
MQTT Broker Client Disconnection

and-rewsmith:


mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -0,0 +1,240 @@
...
+    #[error("failed sending broker signal to disconnect client")]
+    SignalError(),
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::command::parse_client_id;
+    use crate::command::DisconnectClientError;
+
+    #[tokio::test]
```
vadim-kovalyov, line 184: nit: does not have to be tokio test, no async in it.

and-rewsmith:
I opened #3393

PR #3253, opened 2020-07-20
[k8s public preview] Map docker "IpcMode" to k8s "HostIPC" (#3247)

darobs:
Add IpcMode to HostConfig, if IpcMode is "host", set the podspec "hostIpc" to true


darobs:
All failed checkin tests are edge-hub tests, which are not a part of the k8s public preview release.

PR #3254, opened 2020-07-20
Merge master into iiot

and-rewsmith:
Dylan handled the PnP conflicts so I am taking his conflict-resolved branch and opening this PR.

Dylan's explanation:
> Regarding pnp conflicts: Most of it was just putting things in the right place and adding modelId and modelId store where they needed to fill constructors/etc. Since some of the modelId changes were already in, it wasn't a huge problem to merge the rest of the conflicts.
The ClientProvider logic had to be changed slightly to accommodate the GatewayHostName Option and the modelId Option.


mqtt/mqtt-broker/src/configuration.rs
```diff
@@ -0,0 +1,322 @@
+use std::convert::From;
```
and-rewsmith, line 1: Why does this diff show? I'm guessing it was from a change ported to master (i.e. rust stable / clippy fixes). But then why isn't this also in iiot now?

and-rewsmith: Alright so it seems this was later moved so we will have to delete it. Doing so will then fix the checkin gate.

and-rewsmith:
I merged with master in commit 7527a71 despite what the message says.

and-rewsmith:
I was able to resolve the merge conflict because the change didn't relate to pnp. These commits conflicted:
https://github.com/and-rewsmith/iotedge/commit/cb1836eebccb36c4795a21ca0af6ee84806c01f2
https://github.com/Azure/iotedge/commit/c8a78ee97c8f301f886f8e38831dbcbbf24dc2a6

It was a simple fix because the intention of master's change (including transport settings in the DeviceClient.Create()) was already included in iiot.

PR #3256, opened 2020-07-21
Add edgeHub metrics for process time, retry and ttl cleanup

ancaantochi:
- Updated documentation for metrics.
- Added metric for offline
- Enable processing time
- fix input route is not set for edgehub_message_send_duration_seconds metric
- add metric for operation retry
- add metric for dropped messages because TTL expired or no route
- add metric for client connection failure
- add metric for unack message because storage failure


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/cloud/RetryingCloudProxy.cs
```diff
@@ -84,6 +85,7 @@ async Task<T> ExecuteOperation<T>(Func<ICloudProxy, Task<T>> func, string operat
                         throw;
                     }
 
+                    OperationMetrics.Instance.LogRetryOperation(1, this.id, operation);
```
varunpuranik, line 88: Are these the only 2 places where we retry?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceConnectionMetrics.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            this.authCounter = Metrics.Instance.CreateCounter(
+                "client_connect_failed",
+                "Client connection failure",
+                new List<string> { "id", "reason", "protocol", MetricsConstants.MsTelemetry });
+        }
+
+        public static DeviceConnectionMetrics Instance { get; } = new DeviceConnectionMetrics();
+
+        public void LogAuthenticationFailure(long metricValue, string id, string reason, string protocol)
```
varunpuranik, line 21: If we are only logging auth failure, can we centralize it in Authenticator?

ancaantochi:
> Changes look good. My only slight concern is that we are now adding a new pattern for adding Metrics from what we have. Wonder if it is better to unify into our current approach (with possible slight duplication)

I was thinking it would be good to refactor all metrics for edgeHub, but that would make a big change. I can make my changes back as the rest of the metrics. For routing it already had a metric interface, I just made small change for what we needed.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/cloud/RetryingCloudProxy.cs
ancaantochi: there is retry in getting scopes which I didn't and subscription processor. Do you think it would make sense to add those?

PR #3263, opened 2020-07-21
[mqtt] Refactor settings reading for broker

dmolokanov:
There are 2 modes broker can potentially work the generic broker and edge-hub protocol head. In edge-hub mode, the broker requires additional configuration that's not applicable for generic broker: custom authentication and system transport protocol. 

As the result, there is an additional crate in the workspace `mqtt-generic` which contains settings for generic mqtt broker now.


mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -32,38 +53,38 @@ pub async fn broker(
 }
 
 pub async fn start_server<Z, F>(
-    config: &BrokerConfig,
+    config: &Settings,
```
vadim-kovalyov, line 56: nit:
```suggestion
    config: Settings,
```

mqtt/mqttd/src/broker/bootstrap/generic.rs
```diff
@@ -26,7 +44,7 @@ pub async fn broker(
 }
 
 pub async fn start_server<Z, F>(
-    config: &BrokerConfig,
+    config: &Settings,
```
vadim-kovalyov, line 47: nit:
```suggestion
    config: Settings,
```

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -32,38 +53,38 @@ pub async fn broker(
...
-                    ServerCertificate::from_file(path)
-                        .with_context(|| ServerCertificateLoadError::File(path.to_path_buf()))?
-                }
-                None => {
-                    info!("downloading identity from edgelet");
-                    download_server_certificate()
-                        .await
-                        .with_context(|| ServerCertificateLoadError::Edgelet)?
-                }
+            let identity = if let Some(path) = tls.cert_path() {
```
vadim-kovalyov, line 79: q: so, are we ok to override the cert via config? or should we always download from edgelet no matter what?

dmolokanov: by default (when not overwritten in config) it downloads cert from edgelet. If the customer really wants something different, they can provide their own cert. Also in dev mode it somewhat convenient to debug without edgelet running

mqtt/mqtt-generic/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "mqtt-generic"
+version = "0.1.0"
+authors = ["Azure IoT Edge Devs"]
```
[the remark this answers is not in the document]
dmolokanov: I'm planning to get rid of `mqtt-broker-core` and move the code back to `mqtt-broker` soon. I think we did separation wrong previously. It should be
```
[mqtt-edgehub] --> [mqtt-broker] <-- [mqtt-generic]
```
instead of 
```
[mqtt-edgehub]  +--> [mqtt-broker-core] <-- [mqtt-generic]
        ^               ^ 
        |               |
[mqtt-broker] ----------+
```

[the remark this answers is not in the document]
vadim-kovalyov: :+1: great, I like it!

PR #3274, opened 2020-07-23
Rename reboot order from `priority` to be `startupOrder`

yophilav:
Rename reboot order from `priority` to be `startupOrder`


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/IModule.cs
```diff
@@ -112,8 +112,8 @@ public interface IModule : IEquatable<IModule>
         [JsonProperty(PropertyName = "imagePullPolicy")]
         ImagePullPolicy ImagePullPolicy { get; }
 
-        [JsonProperty(PropertyName = "priority")]
-        uint Priority { get; }
+        [JsonProperty(PropertyName = "startupOrder")]
+        uint StartupOrder { get; }
```
varunpuranik, line 116: I thought we discussed that we only need to change the public facing name of this. The internal name can stay priority (to minimize the changes?)

yophilav: That's the original intent but it ripples through all the various classes and their tests.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/IModule.cs
```diff
@@ -112,8 +112,8 @@ public interface IModule : IEquatable<IModule>
         [JsonProperty(PropertyName = "imagePullPolicy")]
         ImagePullPolicy ImagePullPolicy { get; }
 
-        [JsonProperty(PropertyName = "priority")]
-        uint Priority { get; }
+        [JsonProperty(PropertyName = "startupOrder")]
```
ancaantochi, line 115: it looks like the change is not backward compatible, what happens with customers using priority?

richma-ms: I don't think this went out for 1.0.9, so there shouldn't be anyone other than internal partners using this yet.

yophilav: https://github.com/Azure/azure-iotedge/releases/
Looks like the feature is first introduced in release 1.0.10-rc1

PR #3275, opened 2020-07-23
Allow features in rust checkin build script

and-rewsmith:
Allow for features


scripts/linux/generic-rust/build.sh
```diff
@@ -101,16 +106,17 @@ if [ "${REDUCED_LINKER}" = '1' ]; then
...
-
 [profile.test]
 codegen-units = 1
 incremental = false
 EOF
 fi
 
 cd "${PROJECT_ROOT}"
-$CARGO build ${PACKAGES_FORMATTED} ${CARGO_ARGS} --target "$TARGET"
+BUILD_COMMAND="$CARGO build ${PACKAGES_FORMATTED} $(echo ${CARGO_ARGS}) --target \"$TARGET\""
```
dmolokanov:
```suggestion
BUILD_COMMAND="$CARGO build ${PACKAGES_FORMATTED} ${CARGO_ARGS} --target \"$TARGET\""
```
this is also works

and-rewsmith: I had a misconception and thought this fixed what your error was originally. Thanks for cleaning it up. I would have never realized.

PR #3276, opened 2020-07-23
[mqtt] Switch to openssl from native_tls 

dmolokanov:
`native_tls` doesn't support settings to request client certificates. Refactor to switch to openssl directly for starting TLS server.


mqtt/mqtt-broker/src/connection.rs
```diff
@@ -154,6 +155,9 @@ where
 
                 if let Some(certificate) = certificate {
                     context.with_certificate(certificate);
+                    if let Some(chain) = cert_chain{
```
vadim-kovalyov, line 158: nit: formatting

PR #3278, opened 2020-07-23
[PnP] Consolidating productInfoStore and modelIdStore into one MetadataStore

dylanbronson:
Currently, we have a ProductInfoStore and a ModelIdStore that do similar things. They store some information about identities that are connecting. This PR deletes both of those stores and consolidates them into one MetadataStore. 

We have to be careful while doing this merge, because ProductInfoStore is in production already under the name "ProductInfo"
If we switch over without taking that into account, customers could lose some data (productInfo is used for metrics, so no messages would be lost, but some of the metrics may be incorrect). 
With this in mind, the MetadataStore has logic for seeing if the productInfo has already been stored. If productInfo already exists, it will turn it into ConnectionMetadata. It does this by assuming that if it gets a string out of the database, that it is old productInfo. If it gets a ConnectionMetadata object, it knows that it is newer and has already been switched over.

We do something very similar in UpdatableEncryptedStore.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionMetadata.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core
+{
+    public class ConnectionMetadata
+    {
+        public string ModelId { get; set; }
+
+        public string ProductInfo { get; set; }
```
richma-ms, line 8: don't these need to be JsonProperties?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/MetadataStore.cs
```diff
@@ -0,0 +1,115 @@
...
+                        return JsonConvert.DeserializeObject<ConnectionMetadata>(v).ProductInfo;
+                    }
+                    catch (JsonSerializationException)
+                    {
+                        // If deserialization fails, assume the string is an old productInfo.
+                        // We must do this only for migration purposes, since this store used to just be a productInfoStore.
+                        ConnectionMetadata metadata = new ConnectionMetadata() { ProductInfo = v };
+                        await this.metadataEntityStore.Put(id, JsonConvert.SerializeObject(metadata));
+                        return v;
+                    }
```
richma-ms, line 82: Seems like this can be refactored into a helper along the line of GetOrMigrateProductInfo()

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionMetadata.cs
dylanbronson: I guess not, because the tests pass, but it can't hurt to add them

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/MetadataStore.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        public async Task<Option<string>> GetModelId(string id)
+        {
+            Option<string> value = await this.metadataEntityStore.Get(id);
+            return value.Match(
+                v =>
+                {
+                    try
+                    {
```
ancaantochi, line 42: here you could also call getormigrate method

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/MetadataStore.cs
```diff
@@ -0,0 +1,109 @@
...
+                async v =>
+                {
+                    return (await this.GetOrMigrateConnectionMetadata(id, v)).ProductInfo;
+                },
+                () => Task.FromResult(string.Empty));
+        }
+
+        async Task<ConnectionMetadata> GetOrMigrateConnectionMetadata(string id, string entityValue)
+        {
+            try
```
ancaantochi, line 67: nit: try catch could be in a different method  TryDeserialize which returns true/false. I think it's more clear than having migration code in the catch function

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/linkhandlers/LinkHandler.cs
```diff
@@ -116,10 +113,12 @@ protected async Task<bool> Authenticate()
...
             if (authenticated)
             {
-                await this.ClientVersion
+                ConnectionMetadata connectionMetadata = new ConnectionMetadata();
+                this.ClientVersion
                     .Filter(c => !string.IsNullOrWhiteSpace(c))
-                    .ForEachAsync(c => this.productInfoStore.SetProductInfo(this.Identity.Id, c));
-                await this.ModelId.ForEachAsync(m => this.modelIdStore.SetModelId(this.Identity.Id, m));
+                    .ForEach(c => connectionMetadata.ProductInfo = c);
+                connectionMetadata.ModelId = this.ModelId;
```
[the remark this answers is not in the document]
varunpuranik: +1

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionMetadata.cs
varunpuranik: You can remove JsonProperty.. JsonProperty should only be needed if the serialized name is different.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/IMetadataStore.cs
```diff
@@ -0,0 +1,23 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public interface IMetadataStore
+    {
+        Task SetModelId(string id, string modelId);
+
+        Task<Option<string>> GetModelId(string id);
+
+        Task SetProductInfo(string id, string productInfo);
```
varunpuranik, line 13: Do we still need these methods? Why not just GetMetadata/SetMetadata?

dylanbronson: I believe we still need SetProductInfo/SetModelId since we can encounter a situation where we have just product info or just modelId. But GetModelId and GetProductInfo could just become GetMetadata

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionMetadata.cs
```diff
@@ -0,0 +1,58 @@
...
+        [JsonConstructor]
+        public ConnectionMetadata(string productInfo, string modelId, string edgeProductInfo)
+        {
+            Preconditions.CheckNonWhiteSpace(edgeProductInfo, nameof(edgeProductInfo));
+            this.ModelId = Option.Maybe(modelId);
+            this.ProductInfo = productInfo;
+            this.EdgeProductInfo = edgeProductInfo;
+        }
+
+        public ConnectionMetadata(string productInfo, Option<string> modelId, string edgeProductInfo)
```
ancaantochi, line 19: all other constructors could call this one

PR #3281, opened 2020-07-24
Fix uri decoding in token authentication

richma-ms:
the actor device and module from the token audience wasn't getting double URI decoded, this can cause problems when matching against the cached identities.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
```diff
@@ -61,13 +61,13 @@ protected override (Option<string> deviceId, Option<string> moduleId) GetActorId
             if (parts.Length == 3)
             {
                 hostName = parts[0];
-                deviceId = parts[2];
+                deviceId = WebUtility.UrlDecode(parts[2]);
             }
             else if (parts.Length == 5)
             {
                 hostName = parts[0];
                 deviceId = parts[2];
```
philipktlin, line 69: how about this deviceId?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
```diff
@@ -61,13 +61,13 @@ protected override (Option<string> deviceId, Option<string> moduleId) GetActorId
             if (parts.Length == 3)
             {
                 hostName = parts[0];
-                deviceId = parts[2];
+                deviceId = WebUtility.UrlDecode(parts[2]);
```
philipktlin, line 64: nit: should it be better to use regex to get device Id and module Id?

richma-ms: hmm...probably, but this is how it worked before so I didn't bother changing it.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
richma-ms: yes

PR #3282, opened 2020-07-24
Workaround for refreshing newly added devices

richma-ms:
The new GetDeviceAndModuleOnBehalfOf requires the root Edge to know who's the parent Edge of the target identity, so that the root can act OnBehalfOf of that Edge.  But in the scenario where a new device has been added, the root Edge doesn't have the device in its hierarchy yet.  So it has no way to know which Edge this new device is supposed to belong to.

This change implements a work-around where instead of pulling down a single identity, we just refresh the entire cache to sync down any and all changes from upstream.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -160,6 +160,15 @@ async Task<ScopeResult> GetIdentityOnBehalfOf(Uri uri, string deviceId, Option<s
             {
                 // Get the auth-chain for the target device
                 Option<string> maybeAuthChain = await this.serviceIdentityHierarchy.GetEdgeAuthChain(deviceId);
+
+                if (!maybeAuthChain.HasValue)
+                {
+                    // We might not have the target identity yet (e.g. it just got added).
+                    // In this case, we can still try to make the call as ourself, in case
+                    // the missing identity is one of our immediate children.
+                    maybeAuthChain = await this.serviceIdentityHierarchy.GetAuthChain(this.actorEdgeDeviceId);
```
varunpuranik, line 169: Who will make the call if the target is one of our immediate children? The EdgeHub itself? Do we will use this codepath for that flow?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
```diff
@@ -31,12 +31,12 @@ public class DeviceScopeTokenAuthenticator : DeviceScopeAuthenticator<ITokenCred
             this.edgeHubHostName = Preconditions.CheckNotNull(edgeHubHostName, nameof(edgeHubHostName));
         }
 
-        protected override (Option<string> deviceId, Option<string> moduleId) GetActorId(ITokenCredentials credentials)
-        {
-            // Parse the deviceId and moduleId out of the token's audience
-            SharedAccessSignature sharedAccessSignature = SharedAccessSignature.Parse(this.iothubHostName, credentials.Token);
-            (string hostName, string deviceId, Option<string> moduleId) = this.ParseAudience(sharedAccessSignature.Audience);
-            return (Option.Some(deviceId), moduleId);
+        protected override (Option<string> deviceId, Option<string> moduleId) GetActorId(ITokenCredentials credentials)
```
varunpuranik, line 34: what is the diff here? I cannot tell.

varunpuranik: Actually this entire file.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -68,9 +69,15 @@ public sealed class DeviceScopeIdentitiesCache : IDeviceScopeIdentitiesCache
         public void InitiateCacheRefresh()
         {
             Events.ReceivedRequestToRefreshCache();
+            this.refreshCacheCompleteSignal.Reset();
             this.refreshCacheSignal.Set();
         }
 
+        public async Task WaitForCacheRefresh(TimeSpan timeout)
```
varunpuranik, line 76:
```suggestion
        public Task WaitForCacheRefresh(TimeSpan timeout) => this.refreshCacheCompleteSignal.WaitAsync(timeout);
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
richma-ms: yea the whitespacing got messed up again, been trying to fix it but git hasn't been cooperating

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
richma-ms: the authenticator makes the call in that scenario, if we fail to find the auth target, we'll attempt to refresh the target identity and try again.  This block handles that by mirroring our legacy capability (GetDeviceAndModuleInDeviceScope).

PR #3290, opened 2020-07-24
WIP: Custom perf test cases for private preview

vadim-kovalyov:
A set of tests to run before private preview for a longer duration.

### DIFF
--- a/builds/misc/templates/mqtt-perf-run.yaml
+++ b/builds/misc/templates/mqtt-perf-run.yaml
@@ -4,262 +4,126 @@ jobs:
   # Fan-in Test cases (many publishers/few subscribers)
   ################################################################################	
   
-  ### QoS 0
-  
   - template: mqtt-perf-test-case.yaml
     parameters: 
       parentJob: setup_completed
       id: 1
-      publishers: 100
-      subscribers: 2
-      topics: 1
-      payload: 32
+      publishers: 10
+      subscribers: 10
+      topics: 10
+      payload: 128
       qos: 0
-  
+
   - template: mqtt-perf-test-case.yaml
     parameters: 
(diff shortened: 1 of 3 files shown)


builds/misc/templates/mqtt-perf-run.yaml
```diff
@@ -4,262 +4,126 @@ jobs:
...
-      publishers: 2
-      subscribers: 10
+      parentJob: run_completed_10
+      id: 11
+      publishers: 10
+      subscribers: 1
       topics: 1
-      payload: 131072 # 128kb
-      qos: 1
+      payload: 1048576
```
varunpuranik, line 118: 1Mb messags with 10 publishers and 1 sub? Will this run fine? Won't it eventually choke?

builds/misc/templates/mqtt-perf-run.yaml
```diff
@@ -4,262 +4,126 @@ jobs:
...
-  
-  - template: mqtt-perf-test-case.yaml
-    parameters: 
-      parentJob: run_completed_252
-      id: 253
-      publishers: 2
-      subscribers: 10
-      topics: 10
-      payload: 131072 # 128kb
+      parentJob: run_completed_11
```
varunpuranik, line 123: What is the diff between this and the one above?

varunpuranik: Oh qos0 vs qos1.. got it

builds/misc/templates/mqtt-perf-run.yaml
vadim-kovalyov: We'll see :)

vadim-kovalyov:
No longer needed.

PR #3293, opened 2020-07-26
Refresh identities cache if edgeAgent/edgeHub has empty auth-type

richma-ms:
The system modules can be initially created with no auth-type, and have the SAS keys get added later.  In this case, we must refresh the identity cache so get the latest credentials for the edgeHub/edgeAgent, otherwise we won't be able to authenticate incoming downstream Edge connections.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
         public const string InternalOriginInterface = "internal";
         public const string DownstreamOriginInterface = "downstream";
 
+        public const string EdgeAgentModuleId = "$edgeAgent";
         public const string EdgeHubModuleId = "$edgeHub";
         public const string IotEdgeIdentityCapability = "iotEdge";
         public const string ServiceIdentityRefreshMethodName = "RefreshDeviceScopeIdentityCache";
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
@@ -16,7 +16,6 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
         public const string InternalOriginInterface = "internal";
         public const string DownstreamOriginInterface = "downstream";
 
-        public const string EdgeAgentModuleId = "$edgeAgent";
         public const string EdgeHubModuleId = "$edgeHub";
         public const string IotEdgeIdentityCapability = "iotEdge";
         public const string ServiceIdentityRefreshMethodName = "RefreshDeviceScopeIdentityCache";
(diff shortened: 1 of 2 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -155,6 +155,33 @@ async Task<EdgeHubScopeResult> HandleGetDeviceAndModuleOnBehalfOfAsync(string ac
...
         }
 
+        bool NeedToRefreshIdentity(Option<ServiceIdentity> identityOption)
+        {
+            // Default refresh to true if we don't have the identity yet.
+            bool needToRefresh = true;
+
+            identityOption.ForEach(id =>
+            {
+                if (id.ModuleId.Contains(Constants.EdgeAgentModuleId) ||
```
varunpuranik, line 165: Do we need to special case EH/EA? We can prob do the same for any identity that is trying to connect, but has Auth type None, right?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
```diff
@@ -16,6 +16,7 @@ public class Constants
         public const string InternalOriginInterface = "internal";
         public const string DownstreamOriginInterface = "downstream";
 
+        public const string EdgeAgentModuleId = "$edgeAgent";
```
ancaantochi, line 19: is this constant used?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -155,6 +155,23 @@ async Task<EdgeHubScopeResult> HandleGetDeviceAndModuleOnBehalfOfAsync(string ac
             return result;
         }
 
+        bool NeedToRefreshIdentity(Option<ServiceIdentity> identityOption)
```
ancaantochi, line 158: nit: IsRefreshIdentityNeeded

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -155,6 +155,23 @@ async Task<EdgeHubScopeResult> HandleGetDeviceAndModuleOnBehalfOfAsync(string ac
...
+            {
+                // Identities can initially be created with no auth, and
+                // have their auth type updated later. In this case we
+                // must refresh the identity or we won't be able to auth
+                // incoming OnBehalfOf connections for those identities.
+                needToRefresh = id.Authentication.Type == ServiceAuthenticationType.None;
+            });
+
+            return needToRefresh;
+        }
```
varunpuranik, line 173:
```suggestion
        // Identities can initially be created with no auth, and
        // have their auth type updated later. In this case we
        // must refresh the identity or we won't be able to auth
        // incoming OnBehalfOf connections for those identities.
        bool IsRefreshIdentityNeeded(Option<ServiceIdentity> identityOption) => 
            !identityOption.Filter(i => i.Authentication.Type != ServiceAuthenticationType.None).HasValue;        
```

PR #3299, opened 2020-07-27
Stop existing modules when iotedged starts

damonbarry:
Normally iotedged will stop all modules when it shuts down. But if it crashed, modules will continue to run. On Linux systems where iotedged is responsible for creating/binding the socket (e.g., CentOS 7.5, which uses systemd but does not support systemd socket activation), modules will be left holding stale file descriptors for the workload and management APIs and calls on these APIs will begin to fail. It is expected that modules will be resilient to these sorts of failures, but in reality sometimes they aren't.

This change updates iotedged to stop any existing modules when it is starting. The stopped modules will be started again naturally once iotedged (and Edge Agent) are running again.


arsing:
>On some platforms (e.g. CentOS 7.5)

It would be worth making it clear (in both the commit message and the code comment) that this situation happens specifically when the socket is created by iotedged rather than using systemd socket activation. With systemd socket activation the socket file is reused when iotedged restarts, but without socket activation iotedged has to bind it itself, which means it has to unlink it first, which means it gets disconnected from the modules.

damonbarry:
> It would be worth making it clear (in both the commit message and the code comment) that this situation happens specifically when...

Updated, thanks!

edgelet/iotedged/src/lib.rs
```diff
@@ -328,6 +329,18 @@ where
                     crypto.clone(),
                 )?;
 
+                // Normally iotedged will stop all modules when it shuts down. But if it crashed,
+                // then modules will continue to run and they'll be holding stale file descriptors
+                // for the workload and management APIs. On some platforms (e.g. CentOS 7.5), calls
+                // to these APIs will begin to fail. Resilient modules should be able to deal with
+                // this, but we'll restart all modules to ensure a clean start.
+                const STOP_TIME: Duration = Duration::from_secs(30);
```
CindyXing, line 339: So, in case any module failed to be stopped within the 30 seconds, we'll fail edgelet being started?

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -834,10 +834,11 @@ impl ModuleRuntime for DockerModuleRuntime {
                     c.name(),
                     wait_before_kill,
                 )
-                .or_else(|err| match (&err).kind() {
-                    ErrorKind::NotFound(_) | ErrorKind::NotModified => Ok(()),
-                    _ => Err(err),
-                })
+                .or_else(|err| match Fail::find_root_cause(&err).downcast_ref::<ErrorKind>() {
+                        Some(ErrorKind::NotFound(_)) | Some(ErrorKind::NotModified) => Ok(()),
```
CindyXing, line 839: Would there be any other error code? Sometimes docker container can hang that the runtime is not able to stop it; or docker itself hangs.

edgelet/iotedged/src/lib.rs
damonbarry: Correct. 30 seconds is an arbitrary value, so we can adjust that based on feedback. But if we don't fail iotedged in this case, then there are likely to be downstream problems that are more difficult to diagnose, because the errors observed (errors in this zombie module that can't be stopped) will be farther away from the source of the problem.

edgelet/edgelet-docker/src/runtime.rs
damonbarry: After looking through the errors returned from stop, I think these are the two we want to ignore. Other errors signal something more sinister, and I think we want to fail in those cases.

PR #3301, opened 2020-07-28
[mqtt] Log message which was dropped due to queue limits

dmolokanov:
Adds a log message which was dropped due to queue limits.

### DIFF
--- a/mqtt/mqtt-broker/src/session/state/mod.rs
+++ b/mqtt/mqtt-broker/src/session/state/mod.rs
@@ -8,7 +8,7 @@ use set::SmallIndexSet;
 
 use std::{cmp, collections::HashMap};
 
-use tracing::debug;
+use tracing::{debug, warn};
 
 use mqtt3::proto;
 
@@ -123,7 +123,9 @@ impl SessionState {
 
     pub fn queue_publish(&mut self, publication: proto::Publication) -> Result<(), Error> {
         if let Some(publication) = self.filter(publication) {
-            self.waiting_to_be_sent.enqueue(publication);
+            if let Some(limit) = self.waiting_to_be_sent.enqueue(publication) {
+                warn!("{}. drop publication {:?}", limit, limit.publication())
+            }
         }
         Ok(())
     }
--- a/mqtt/mqtt-broker/src/session/state/mod.rs
+++ b/mqtt/mqtt-broker/src/session/state/mod.rs
@@ -8,7 +8,7 @@ use set::SmallIndexSet;
 
 use std::{cmp, collections::HashMap};
 
-use tracing::{debug, warn};
+use tracing::{debug, info};
 
 use mqtt3::proto;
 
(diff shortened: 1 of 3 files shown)


mqtt/mqtt-broker/src/session/state/mod.rs
```diff
@@ -123,7 +123,9 @@ impl SessionState {
 
     pub fn queue_publish(&mut self, publication: proto::Publication) -> Result<(), Error> {
         if let Some(publication) = self.filter(publication) {
-            self.waiting_to_be_sent.enqueue(publication);
+            if let Some(limit) = self.waiting_to_be_sent.enqueue(publication) {
+                warn!("{}. drop publication {:?}", limit, limit.publication())
```
and-rewsmith, line 127: How will publication display in the logs? In my temporary hack I am just showing the topic.

mqtt/mqtt3/examples/publisher.rs
```diff
@@ -117,7 +117,7 @@ fn main() {
     let publish_handle = client
         .publish_handle()
         .expect("couldn't get publish handle");
-    runtime_handle.clone().spawn(async move {
+    runtime.spawn(async move {
```
and-rewsmith, line 120: Why this change?

mqtt/mqtt-broker/src/session/state/mod.rs
dmolokanov: `proto::Publication` implements `Debug` but not `Display`. that's why it doesn't work with `{}` but works with `{:?}`.
https://doc.rust-lang.org/std/fmt/

vadim-kovalyov: Good question, it will dump the payload, do we really want it?

mqtt/mqtt3/examples/publisher.rs
dmolokanov: Clippy was upset with unnecessary cloning.

mqtt/mqtt-broker/src/session/state/mod.rs
dmolokanov: Any better suggestions?

dmolokanov: make info showing only a topic and debug with full dump

mqtt/mqtt-broker/src/session/state/mod.rs
```diff
@@ -124,7 +124,9 @@ impl SessionState {
     pub fn queue_publish(&mut self, publication: proto::Publication) -> Result<(), Error> {
         if let Some(publication) = self.filter(publication) {
             if let Some(limit) = self.waiting_to_be_sent.enqueue(publication) {
-                warn!("{}. drop publication {:?}", limit, limit.publication())
+                let dropped = limit.publication();
+                info!("{}. drop publication {}", limit, dropped.topic_name);
```
and-rewsmith, line 128: nit: "dropped" might be better for consistency

PR #3307, opened 2020-07-28
Ordered start for authentication and mqtt bridge

vipeller:
When edgeHub starts up, protocol heads start up parallel. In case of Authentication and mqtt bridge it could happen (and usually happened) that authentication came alive sooner. When a client connected after authentication came alive, but mqtt bridge had not, then the mqtt broker allowed the client to connect. When this client sent a message immediately, mqtt bridge was not subscribed to the necessary topics to handle the messages toward iothub. These messages got lost.
This change starts mqtt bride earlier and starts authentication only when everything is up. It does the reverse order on shutdown.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
@@ -115,7 +115,19 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
                             c.MqttMsgPublishReceived -= this.ForwardPublish;
                             c.ConnectionClosed -= this.TriggerReconnect;
 
-                            c.Disconnect();
+                            if (c.IsConnected)
+                            {
+                                try
+                                {
+                                    c.Disconnect();
+                                }
+                                catch
+                                {
+                                    // swallowing: when the container is shutting down, it is possible that the broker disconnected.
+                                    // in those case an internal socket will be deleted and this Disconnect() call ends up in a
+                                    // Disposed() exception.
+                                }
+                            }
 
                             await this.StopForwardingLoopAsync();
 
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
@@ -3,6 +3,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Service
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Security.Authentication;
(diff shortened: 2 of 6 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -115,7 +115,19 @@ public async Task DisconnectAsync()
                             c.MqttMsgPublishReceived -= this.ForwardPublish;
                             c.ConnectionClosed -= this.TriggerReconnect;
 
-                            c.Disconnect();
+                            if (c.IsConnected)
```
ancaantochi, line 118: Can this check be moved to Disconnect method?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -163,14 +164,29 @@ static async Task<EdgeHubProtocolHead> GetEdgeHubProtocolHeadAsync(ILogger logge
...
             }
 
-            if (configuration.GetValue("mqttBrokerSettings:enabled", true))
+            switch (orderedProtocolHeads.Count)
             {
-                protocolHeads.Add(container.Resolve<MqttBrokerProtocolHead>());
+                case 0:
+                    break;
+
+                case 1:
```
ancaantochi, line 183: this can be just an if orderedProtocolHeads.Count >= 1 add to the list, it should not matter if the orderedprotocolheads only contain one item because it will just work to start it

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: you mean the c.IsConnected check? This class maintains its own state (connected/disconnected), and when we get here, this state is "connected". What may happen is that the broker is already down, so the underlying socket is already disconnected (even if this class is still in "connected" state). This check is on the mqtt library I use, that may already noticed that  it does not have connection (and down in the code there is a TryReconnect() method, that may be running and it tries to re-connect). The object disposed exception came when the c.Disconnect() call was made in this state, that is why the check before the call. Also, there might be a race condition, that the c.IsConnected() still believes that the library is connected, but by the time the c.Disconnect() call gets there to close the socket, the socket is cleaned up (because the broker dropped the connection)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
vipeller: Yes, I thought the same, that the OrderedProtocolHead works 1 (or even zero) items in it - but then I thought to use it only when it actually makes something, otherwise it falls back to the original implementation.

PR #3311, opened 2020-07-29
Fix DeviceScope API AuthN

richma-ms:
Authentication for the REST endpoints are split over the AuthenticationMiddleWare and DeviceScopeController.  The middleware performs AuthN on the actor presented by "x-ms-edge-moduleId" header, whereas the controller performs AuthZ on the actor present in the URI.  Since we don't go back and do AuthN on the actor in the URI, it's possible to use any identity in a hierarchy to masquerade as any EdgeHub.

Since the middleware already performed AuthN on the actor from the header, the simplest fix is to verify that actor from URI is the same identity.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -202,6 +203,26 @@ async Task<bool> AuthorizeActorAsync(IDeviceScopeIdentitiesCache identitiesCache
                 return false;
             }
 
+            if (!this.Request.Headers.TryGetValue(Constants.ServiceApiIdHeaderKey, out StringValues clientIds) || clientIds.Count == 0)
```
varunpuranik, line 206: Should we log why the authorization is failing? Else it might be hard to tell..

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -291,9 +324,39 @@ public static void UnauthorizedActor(string actorDeviceId, string actorModuleId,
...
 
-            public static void InvalidAuthchain(string authChain)
+            public static void InvalidRequestAuthchain(string authChain)
+            {
+                Log.LogError((int)EventIds.InvalidRequestAuthchain, $"Invalid auth chain: {authChain}");
+            }
+
+            public static void AuthFail_BadActor(string actorDeviceId, string actorModuleId)
+            {
+                Log.LogError((int)EventIds.AuthFail_BadActor, $"Only EdgeHub is allowed to act OnBehalfOf another identity: {actorDeviceId}/{actorModuleId}");
```
varunpuranik, line 334: Nit: suggest reword - 
{actorDeviceId}/{actorModuleId} is not authorized to connect on behalf of {targetDeviceId}

varunpuranik, line 334:
```suggestion
                Log.LogError((int)EventIds.AuthFail_BadActor, $"{actorDeviceId}/{actorModuleId} is not authorized to connect on behalf of {targetDeviceId}.");
```

PR #3313, opened 2020-07-29
Add more EA metrics

lfitchett:
Adds the following metrics:

EdgeAgent metrics:
number of deployments (agent twin updates)
Time taken to apply deployment
number of successful IoT Hub syncs
number of unsuccessful IoT Hub syncs
number of times agent direct methods were invoked, per method.
number of "retries"
edgeagent version (hint: constant)

Device meta-data:
OS SKU/ distro 
H/W arch
IoT Edge daemon runtime version
Experimental features


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/SystemInfo.cs
```diff
@@ -6,12 +6,20 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Core
...
             this.OperatingSystemType = operatingSystemType;
             this.Architecture = architecture;
             this.Version = version;
+            this.ServerVersion = serverVersion;
+            this.KernelVersion = kernelVersion;
+            this.OperatingSystem = operatingSystem;
+            this.NumCpus = numCpus;
+        }
+
+        public SystemInfo(string operatingSystemType, string architecture, string version)
```
damonbarry, line 20: Do we need the older version of the constructor (with fewer params)?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleManagementHttpClient.cs
```diff
@@ -83,11 +83,17 @@ internal static ModuleManagementHttpClientVersioned GetVersionedModuleManagement
...
+            {
+                return new Version_2020_07_07.ModuleManagementHttpClient(managementUri);
+            }
+
             return new Version_2018_06_28.ModuleManagementHttpClient(managementUri);
         }
 
         static ApiVersion GetSupportedVersion(string serverSupportedApiVersion, string clientSupportedApiVersion)
         {
+            Console.WriteLine($"\n\n\n{serverSupportedApiVersion}\n{clientSupportedApiVersion}\n\n\n");
```
damonbarry, line 96: Is this dev-time logging you meant to remove, or do you want to check this in?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/SystemInfo.cs
lfitchett: There's some k8s code that uses it, so I figured the easiest thing would be to leave it, since the new fields are blank for k8s anyway

PR #3315, opened 2020-07-29
[k8s] Have EdgeDeploymentOperator report status to Agent.

darobs:
Instead of allowing Agent to crash on Edge Deployment failure, the
failure that was reported to edge Deployment status will get reported
back in the EdgeAgent's twin.  This more closely follows how single
device Edge Agent operates.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentStatusCommand.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentStatusCommand.cs
@@ -1,0 +1,47 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Agent.Core;
+    using Microsoft.Azure.Devices.Edge.Agent.Core.ConfigSources;
+    using Microsoft.Azure.Devices.Edge.Agent.Kubernetes;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public class EdgeDeploymentStatusCommand : ICommand
+    {
+        readonly Option<EdgeDeploymentStatus> activeStatus;
+
+        public string Id => "EdgeDeploymentStatusCommand";
+
+        public EdgeDeploymentStatusCommand(
+            Option<EdgeDeploymentStatus> activeStatus)
+        {
+            this.activeStatus = activeStatus;
+        }
+
+        public Task ExecuteAsync(CancellationToken token)
+        {
+            // If the EdgeDeployment operator fails, report this to the plan executor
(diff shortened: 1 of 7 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentStatusCommand.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public Task ExecuteAsync(CancellationToken token)
+        {
+            // If the EdgeDeployment operator fails, report this to the plan executor
+            // as an exception.
+            this.activeStatus.ForEach(status =>
+                {
+                    if (status.State != EdgeDeploymentStatusType.Success)
+                    {
+                        throw new ConfigOperationFailureException(status.Message);
```
CindyXing, line 33: Besides failure message, what can be the inner exception or callstack? It would be useful for troubleshooting.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/EdgeDeploymentStatusCommandTest.cs
```diff
@@ -0,0 +1,55 @@
...
+            await successStatus.ExecuteAsync(CancellationToken.None);
+        }
+
+        [Fact]
+        public async void ExecuteAsyncThrowsWithFailureStatus()
+        {
+            EdgeDeploymentStatus failure = EdgeDeploymentStatus.Failure("failure message");
+            EdgeDeploymentStatusCommand failStatus = new EdgeDeploymentStatusCommand(Option.Some(failure));
+
+            await Assert.ThrowsAsync<ConfigOperationFailureException>(() => failStatus.ExecuteAsync(CancellationToken.None));
```
CindyXing, line 38: How about the failure message? It should be ConfigOperationFailureException(failure.message), right?
While we eat the exception, are we able to collect the failure message and the inner exception?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentStatusCommand.cs
darobs: I chose not to report the full call stack, since this will eventually put into the twin, which has size limits.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/EdgeDeploymentStatusCommandTest.cs
CindyXing: Can you confirm if the failure message will be eaten here or not?

CindyXing: ping...

darobs:
Closing because 1st edge on k8s release will be on 1.1 branch.

PR #3317, opened 2020-07-30
Update prioritized routes documentation

richma-ms:
Making a few minor changes in the priority route doc to address feedback from CSE.

Also added the new metrics dimensions for priority to the metrics doc.


doc/Route_priority_and_TTL.md
```diff
@@ -22,32 +22,48 @@ Since message priority and TTL are associated with routes, the route section of
...
-          "FallbackPriority": <uint>
-          }
-        },</b>
-      "route2": <string>
-    },
-    "storeAndForwardConfiguration": {
-      "timeToLiveSecs": 7200
+"routes": {
+    "type": "object",
+    "patternProperties": {
```
dylanbronson, line 35: what is patternProperties?

richma-ms: It's just a way to specify multiple types of properties, each able to have multiple instances.  In this case, routes can be either legacy string property or the new JSON object format.

https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties

PR #3319, opened 2020-07-30
[mqtt] Add retry policy to authenticate client

dmolokanov:
We have 2 services starting at the same time, and the broker has a dependency on EdgeHub authentication endpoint. To race conditions between services during the startup retry policies implemented.
The authenticator will continue attempts to authenticate client for a 1-minute interval with exponential backoff.


mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
```diff
@@ -69,7 +72,17 @@ impl Authenticator for EdgeHubAuthenticator {
...
         context: AuthenticationContext,
     ) -> Result<Option<AuthId>, Self::Error> {
-        let auth_id = self.authenticate(context).await?;
+        let auth_id = (|| async {
+            info!("authenticate client");
+            let auth_id = self.authenticate(&context).await?;
+            Ok(auth_id)
+        })
+        .retry(ExponentialBackoff {
+            max_elapsed_time: Some(Duration::from_secs(60)),
```
varunpuranik, line 81: So it will give up after 60s? What does the exponential backoff look like (as in how many times will it retry during this time?)

varunpuranik: Maybe add a comment on why we are doing this and what the behavior will be like?

mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
```diff
@@ -69,7 +72,17 @@ impl Authenticator for EdgeHubAuthenticator {
...
+            ..ExponentialBackoff::default()
+        })
+        .await?;
```
and-rewsmith, line 84: For reference:
https://docs.rs/backoff/0.2.1/backoff/exponential/struct.ExponentialBackoff.html

mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
```diff
@@ -69,7 +72,17 @@ impl Authenticator for EdgeHubAuthenticator {
...
     ) -> Result<Option<AuthId>, Self::Error> {
-        let auth_id = self.authenticate(context).await?;
+        let auth_id = (|| async {
+            info!("authenticate client");
+            let auth_id = self.authenticate(&context).await?;
+            Ok(auth_id)
+        })
+        .retry(ExponentialBackoff {
+            max_elapsed_time: Some(Duration::from_secs(60)),
+            ..ExponentialBackoff::default()
```
and-rewsmith, line 82: dot dot syntax means ignore other fields?

mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
and-rewsmith, line 81: How many times would it end up retrying in 60 seconds with exponential backoff?

mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
vadim-kovalyov: yeah, similar to spread operator in TS/JS or other languages. https://cheats.rs/#data-structures

and-rewsmith: Here are the defaults used:
https://docs.rs/backoff/0.2.1/backoff/default/index.html

Here is a table showing it will retry 10 times in one minute with the default settings:
https://docs.rs/backoff/0.2.1/backoff/index.html

mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
and-rewsmith: It seems like it will retry 10 times in 60 seconds with the default settings. There is also some randomization, but that is by design so I think we should keep it.

PR #3321, opened 2020-07-31
Update EdgeAgent deployment schema validation to 1.1.0

richma-ms:
Bumping EdgeAgent schema version validation to 1.1.0 to reflect the new changes we have for module startup order.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -141,11 +142,47 @@ public async Task SendEventAsync(Message message)
...
-                Events.MismatchedMinorVersions(schemaVersion, ExpectedSchemaVersion);
+                throw new InvalidSchemaVersionException($"The desired properties schema version {schemaVersion} is not compatible with the expected version {ExpectedSchemaVersion}");
+            }
+
+            // Validate minor versions
+            if (actualSchemaVersion.Minor == 0)
+            {
+                // 1.0
+                //
+                // Module startup order is not supported
```
yophilav, line 166: Do we support PriorityQueue in 1.0 schema?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/EdgeAgentConnectionTest.cs
```diff
@@ -1614,29 +1615,55 @@ public async Task GetDeploymentConfigInfoAsync_CreateNewModuleClientWhenGetTwinT
...
-        [InlineData("1.2", null)]
-        [InlineData("1.3", null)]
-        [InlineData("1", typeof(InvalidSchemaVersionException))]
-        [InlineData("", typeof(InvalidSchemaVersionException))]
-        [InlineData(null, typeof(InvalidSchemaVersionException))]
-        [InlineData("0.1", typeof(InvalidSchemaVersionException))]
-        [InlineData("2.0", typeof(InvalidSchemaVersionException))]
-        [InlineData("2.1", typeof(InvalidSchemaVersionException))]
-        public void SchemaVersionCheckTest(string schemaVersion, Type expectedException)
+        [MemberData(nameof(GetDeploymentForSchemas))]
```
yophilav, line 1618: Thank you for the clean up.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -18,7 +19,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub
 
     public class EdgeAgentConnection : IEdgeAgentConnection
     {
-        internal static readonly Version ExpectedSchemaVersion = new Version("1.0");
+        internal static readonly Version ExpectedSchemaVersion = new Version("1.1.0");
```
yophilav, line 22: Per https://github.com/SchemaStore/schemastore/pull/1116/files#diff-25894c152d2d2b3bedcf8f3151fd5193, it looks like the schemaVersion should be `1.1` not `1.1.0`?

varunpuranik: What is the expected version of EdgeHub?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -141,11 +142,47 @@ public async Task SendEventAsync(Message message)
...
+                // 1.0
+                //
+                // Module startup order is not supported
+                foreach (KeyValuePair<string, IModule> kvp in config.Modules)
+                {
+                    IModule moduleConfig = kvp.Value;
+
+                    if (moduleConfig.StartupOrder != Constants.DefaultPriority)
+                    {
+                        throw new InvalidSchemaVersionException($"Module startup order is not supported in schema {actualSchemaVersion}.");
```
varunpuranik, line 173: Why do we need to throw here? Maybe log and ignore? Also, this doesn't seem to be a sustainable way to handle adding new properties to the schema. I feel like we need to come up with a generic way to handle this.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
richma-ms: it's 1.1.0 for both EH and EA in code, but it makes no difference for the validation logic, as we only care about major+minor, patch is ignored.

PR #3323, opened 2020-07-31
Adding modelId to IoTHub device authentication flow

dylanbronson:
When a device does not have edgeHub set as the parent, edgeHub tries to connect to IoTHub for the device as a means of authentication. This allows the user to still connect a device through edgeHub even if the user does not want to (or just forgets) set the parent of the device as edgeHub. 

We don't want to set the modelId of the auth until AFTER we know the device has been authenticated. Otherwise, someone could connect with a malicious or incorrect device, and edgeHub would save the modelId as the new modelId prematurely. Currently, if a user doesn't set edgeHub as the parent of a device, the modelId won't be stored for a device.

This PR passes the modelId to authentication through IoT Hub so that we can connect with the modelId when authenticating with IoT Hub. 

It works quite nicely for Mqtt. But is a bit awkward for Amqp, because we need to pass the modelId directly into the AuthenticateAsync call. It requires a change in all IAmqpAuthenticators.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -88,6 +88,7 @@ public async Task<bool> AuthenticateAsync(string id)
                         {
                             if (!credentialsInfo.IsAuthenticated)
                             {
+                                credentialsInfo.ClientCredentials.ModelId = modelId;
```
dylanbronson, line 91: This part is quite awkward. I would much rather put the modelId in the credentials map initially, but I couldn't think of a clean way of how to do it. 
This also makes modelId need the "set" property everywhere, which is also awkward. 
Open to ideas on it.

richma-ms: I'm going to have the same problem with passing through the authchain on the CBS link.  I was going to change this so that the cached dictionary stores the token instead of a credentials object, and then construct the credentials here on the fly.

You can do that refactor here as part of this work if you have the bandwidth.  Looks like it'll make you change simpler over all and save me work when I get to this.  Otherwise I can come back and clean it up later this month.

---
In reply to: [463754436](https://github.com/Azure/iotedge/pull/3323#discussion_r463754436) [](ancestors = 463754436)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Http/middleware/AuthenticationMiddleware.cs
```diff
@@ -84,7 +84,7 @@ internal async Task<(bool, string)> AuthenticateRequest(HttpContext context)
             if (clientCertificate != null)
             {
                 IList<X509Certificate2> certChain = context.GetClientCertificateChain();
-                clientCredentials = this.identityFactory.GetWithX509Cert(deviceId, moduleId, string.Empty, clientCertificate, certChain);
+                clientCredentials = this.identityFactory.GetWithX509Cert(deviceId, moduleId, string.Empty, clientCertificate, certChain, Option.None<string>());
```
richma-ms, line 87: >clientCredentials = this.identityFactory.GetWithX509Cert(deviceId, moduleId, string.Empty, clientCertificate, certChain, Option.None<string>()); [](start = 16, length = 144)

FYI I deleted this file in a pending PR to refactor the HTTP auth, one of us will need to deal with the merge conflict when this payload hits iiot branch.

PR #3328, opened 2020-07-31
[mqtt] Add get trust bundle call for workload client

dmolokanov:
Move `edgelet` module to its own separate crate. It will be easier to share it across multiple projects.
Add get trust bundle call for workload client


mqtt/edgelet/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "edgelet"
```
vadim-kovalyov, line 2: should we call it "edgelet-client" instead?

dmolokanov: yeah, it makes sense. it's hard to misinterpret `edgelet` workspace with `edgelet` client though.

PR #3337, opened 2020-08-01
[mqtt] Make transport enabled/disabled from ENV

dmolokanov:
Make it possible to disable transport protocols enabled by default.
`LISTENER__TCP__ENABLED = "false"` to disable MQTT over TCP (port: 1883)
`LISTENER__TLS__ENABLED = "false"` to disable MQTT over TCP (port: 8883)


mqtt/mqtt-broker/src/settings/mod.rs
```diff
@@ -153,3 +153,74 @@ pub struct SessionPersistenceConfig {
     time_interval: Duration,
     unsaved_message_count: u32,
 }
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct Enableble<T> {
```
varunpuranik, line 158: typo? Enable<T>?

dmolokanov: Not exactly a typo. It is some type, that can be enabled/disabled. But, probably Enable also a good option.

mqtt/mqtt-broker/src/settings/mod.rs
```diff
@@ -153,3 +153,132 @@ pub struct SessionPersistenceConfig {
     time_interval: Duration,
     unsaved_message_count: u32,
 }
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct Enable<T> {
```
vadim-kovalyov, line 158: nit: comments

PR #3339, opened 2020-08-01
[mqtt] Fix flaky edgehub authentication test

dmolokanov:
Fix flaky Edge Hub authentication test


mqtt/mqtt-edgehub/src/auth/authentication/edgehub.rs
```diff
@@ -393,6 +393,10 @@ ov2gTgQyaRE8rbX4SSPZghE5km7p6FAIjm/uqU9kGMUk3A==
     async fn it_retries_when_edgehub_unavailable_for_some_time() {
         let (tx, rx) = oneshot::channel();
 
+        // create mock http server with unused endpoint just to prevent
+        // other tests to make endpoint which could override the current one
+        let _ = mock("POST", "/unused").create();
```
and-rewsmith, line 398: Why does this behavior happen?

dmolokanov: each instance of mocked server shares a mutex. It makes all tests that use mockito to run sequentially. But in this particular test, we used to delay mock creation to emulate server being "not available". It leads to race conditions when HTTP clients from different tests try to access the same endpoint.

The fix is I made: force mockito to acquire mutex **before** the client makes the first call.

PR #3303, opened 2020-07-28
Test Modules: Fix module-base-full base images to specify moduleuser in dockerfile

and-rewsmith:
Duplicate PR as made to master:
https://github.com/Azure/iotedge/pull/3302

PR #3349, opened 2020-08-03
Test Modules: Fix module-base-full base images to specify moduleuser …

and-rewsmith:
…in dockerfile (#3303)

There were conflicts with this cherry pick due to azureiotedge vs edgebuilds, but it was easy resolution since base images are duplicated.


and-rewsmith:
Let's discuss the versioning offline tomorrow

and-rewsmith:
To my memory arm64 doesn't need module-base-full because the arm64 bionic base image we are using doesn't need the iproute tooling. Previously the pattern was:
- use module-base-full for arm32 network controller (because it needs iproute)
- use module-base for arm64 network controller (because it doesn't need iproute)


This pattern was broken in a base image update to 1.0.10

I think that Bear had a good point that the tests run fine where both network controller base images are module-base-full. Another good alternative he thought of was installing iproute on the module-base because it was small. Therefore I think the best plan is this:
- Restore 1.0.10 to its original state where we use module-base for arm64 network controller
- Make one commit to fix this how we think is best. Then cherry pick it everywhere.

PR #3356, opened 2020-08-04
Support bundle upload

lfitchett:
This allows the support bundle to be uploaded in a similar matter to the built in logs upload. Currently the support bundle is held in memory while it is streamed to the cloud, but is architectured such that it can be stored and streamed from a file on disk in the future. 

### C# Changes
* Added Support Bundle Direct Method
* Renamed LogsUploader to RequestUploader

### Rust Changes
* Split Support Bundle and Logs Collection to new library that is shared by iotedge tool and the management endpoint. 
* Refactored Support Bundle to use generic Read instead of File
* added endpoint to get support bundle

Note: The 2020-07-07 mgmt api version is brand new, and these additions are planned.


edgelet/support-bundle/src/support_bundle.rs
```diff
@@ -0,0 +1,785 @@
...
+use crate::logs::pull_logs;
+
+pub fn make_bundle<M>(
+    output_location: OutputLocation,
+    log_options: LogOptions,
+    include_ms_only: bool,
+    verbose: bool,
+    iothub_hostname: Option<String>,
+    runtime: M,
+) -> Box<dyn Future<Item = (Box<dyn Read + Send>, u64), Error = Error> + Send>
```
arsing, line 28: Do you need to make a trait object here? `-> impl Future<Item = (impl Read + Send, u64), Error = Error>` should work too, and then you wouldn't need to `Box` the `Future` nor the `Read` either.

edgelet/support-bundle/src/support_bundle.rs
```diff
@@ -0,0 +1,785 @@
...
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum OutputLocation {
+    File(OsString),
+    Memory,
+}
+
+// #[cfg(test)]
```
arsing, line 513: Why is this commented out?

lfitchett: forgot to uncomment, fixed

edgelet/support-bundle/src/support_bundle.rs
lfitchett: It has to be boxed since there are 2 match arms that return different futures

arsing: Ah yes, and two different `Read` impls as well.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/logs/ILogsUploader.cs
```diff
@@ -2,12 +2,15 @@
 namespace Microsoft.Azure.Devices.Edge.Agent.Core.Logs
 {
     using System;
+    using System.IO;
     using System.Threading.Tasks;
 
-    public interface ILogsUploader
+    public interface IRequestsUploader
```
varunpuranik, line 8: Why change this name?

varunpuranik: Change the file name too?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/SupportBundleRequest.cs
```diff
@@ -0,0 +1,37 @@
...
+        public string SchemaVersion { get; }
+
+        [JsonIgnore]
+        public string SasUrl { get; }
+
+        [JsonProperty("since")]
+        public Option<string> Since { get; }
+
+        [JsonProperty("iothubHostname")]
+        public Option<string> IothubHostname { get; }
```
varunpuranik, line 32: How is this used?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/SupportBundleRequestHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+    using System.IO;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Agent.Core.Logs;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public class SupportBundleRequestHandler : RequestHandlerBase<SupportBundleRequest, TaskStatusResponse>
+    {
+        public delegate Task<Stream> GetSupportBundle(Option<string> since, Option<string> iothubHostname, Option<bool> edgeRuntimeOnly, CancellationToken token);
```
varunpuranik, line 17: Instead of a delegate seems like an Interface might be cleaner?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/logs/ILogsUploader.cs
lfitchett: Since it also uploads the support bundle, I made the name more generic. I will rename the file, missed that

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/SupportBundleRequest.cs
lfitchett: iotedge check has it as an option, so the support bundle does too

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/SupportBundleRequestHandler.cs
lfitchett: I'm not sure which interface I would use, but a delegate seemed like an clean way to pass a single function in. I've used this pattern before in the code base.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleManagementHttpClient.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Edgelet
     using Microsoft.Azure.Devices.Edge.Util;
     using Microsoft.Azure.Devices.Edge.Util.Edged;
 
-    public class ModuleManagementHttpClient : IModuleManager, IIdentityManager, IDeviceManager
+    public class ModuleManagementHttpClient : IModuleManager, IIdentityManager, IDeviceManager, IDisposable
```
damonbarry, line 16: Is IDisposable needed?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/SupportBundleRequest.cs
varunpuranik: What happens if a user specifies a different IoTHub? 
I think this parameter definitely seems odd - I would say remove it and use the IOTHub that EdgeAgent is using. 
@veyalla ?

veyalla: Using IoT Hub that edge agent is using would actually be an improvement over running support-bundle locally because currently we cannot run any connectivity tests when using DPS because we can't get the iot hub name from config yaml. 

However, it would cause inconsistent behavior for the same command when run locally vs remotely which is weird. There was some thoughts from @arsing around getting the iot hub name from a cached provisioning result when run locally on a DPS provisioned device. If that can be implemented we could have a consistent experience.

arsing: `provisioning_backup.json` has the hub name, so `iotedge check` / `support-bundle` could read it from there.

arsing: (If you do make that change, you'll probably want to keep the `--iothub-hostname` parameter to `check` / `support-bundle` so that passing it is not an error, such as when running old EA with new host package.)

veyalla: IMO, we should make the change Arnav is suggesting, it will improve troubleshooting experience for customers running check / support-bundle locally and also provide consistency when running it locally or remotely. Thoughts @CindyXing ?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Edgelet/ModuleManagementHttpClient.cs
lfitchett: nope, removed

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/SupportBundleRequest.cs
varunpuranik: Yes, wherever we can infer the IoThub, we should I can't think of a scenario where support bundle or iotedge check would need to talk to a different IoTHub.

lfitchett: So for this pr, I'll remove the iothubHostname param, and just fill it in from ea's known hostname

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/ExperimentalFeatures.cs
```diff
@@ -7,12 +7,13 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Service
...
     public class ExperimentalFeatures
     {
-        ExperimentalFeatures(bool enabled, bool disableCloudSubscriptions, bool enableUploadLogs, bool enableGetLogs)
+        ExperimentalFeatures(bool enabled, bool disableCloudSubscriptions, bool enableUploadLogs, bool enableGetLogs, bool enableUploadSupportBundle)
         {
             this.Enabled = enabled;
             this.DisableCloudSubscriptions = disableCloudSubscriptions;
             this.EnableUploadLogs = enableUploadLogs;
             this.EnableGetLogs = enableGetLogs;
+            this.EnableUploadSupportBundle = enableUploadSupportBundle;
```
CindyXing, line 16: So, in 1.0.10, we'll keep this and uploadlogs, getlogs as experimental features?

lfitchett: the plan is to move at least get and upload logs out of experimental for 1.10, this will probably stay in experimental

PR #3360, opened 2020-08-04
Fix startupOrder deserialization

yophilav:
Fix startupOrder deserialization


philipktlin:
How about these classes - DockerDesiredModule, EdgeHubDockerModule?

philipktlin:
We don't have any tests for TypeSpecificSerDe<DeploymentConfig> for deserialization.  Would you please add some tests?

yophilav:
> How about these classes - DockerDesiredModule, EdgeHubDockerModule?

The two classes were updated in the previous PR for the name change. The classes that are updated here are the one that were missed from my change there.

> We don't have any tests for TypeSpecificSerDe for deserialization. Would you please add some tests?

Definitely but since this PR is a RC2 blocker, I want to merge this PR in first and then make sure to introduce the desterilize test for the structure before 1.0.10 goes out (along with other bug fixes)

philipktlin:
The previous PR (https://github.com/Azure/iotedge/commit/eed9c064ae0ad5342a2196bd42feff2d8b7e9cdc) doesn't touch these 2 files (DockerDesiredModule and EdgeHubDockerModule).  Please rename priority to startupOrder in these classes.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Core.Test/TestRuntimeModule.cs
```diff
@@ -22,10 +22,10 @@ public class TestRuntimeModule : TestModule, IRuntimeModule<TestConfig>
             DateTime lastRestartTimeUtc,
             ModuleStatus runtimeStatus,
             ImagePullPolicy imagePullPolicy = ImagePullPolicy.OnCreate,
-            uint priority = Constants.DefaultPriority,
+            uint startupOrder = Constants.DefaultPriority,
```
philipktlin, line 25: better to rename Constants.DefaultPriority.

yophilav: That's done in 1.0.10 I believe.

philipktlin: No, it still shows DefaultPriority in 1.0.10 branch, https://github.com/Azure/iotedge/blob/70661641799d16a940b191ee1c7faa24842fd4be/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Core.Test/TestRuntimeModule.cs#L25.

PR #3365, opened 2020-08-05
[PnP] Fixing bug that broke integ tests

dylanbronson:
The last pnp PR broke our integ tests because of two subtle bugs.

First is that the metadataStore needs to be mocked in two more places.
2nd is that I had a more stringent check in MetadataStore that broke some of the integ tests, since the integ tests use empty string as edgeProductInfo.

As I don't want to cause any breaking changes, I changed the metadataStore check to be the same as it was before (CheckNotNull instead of CheckNonWhitespace)

### DIFF
--- a/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionMetadata.cs
+++ b/edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionMetadata.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
 
         public ConnectionMetadata(string productInfo, Option<string> modelId, string edgeProductInfo)
         {
-            Preconditions.CheckNonWhiteSpace(edgeProductInfo, nameof(edgeProductInfo));
+            Preconditions.CheckNotNull(edgeProductInfo, nameof(edgeProductInfo));
             this.ModelId = modelId;
             this.ProductInfo = productInfo;
             this.EdgeProductInfo = this.BuildEdgeProductInfo(productInfo, edgeProductInfo);
--- a/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/EdgeHubConnectionTest.cs
+++ b/edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/EdgeHubConnectionTest.cs
@@ -63,7 +63,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.E2E.Test
             var signatureProvider = new SharedAccessKeySignatureProvider(sasKey);
             var credentialsCache = Mock.Of<ICredentialsCache>();
             var metadataStore = new Mock<IMetadataStore>();
-            metadataStore.Setup(m => m.GetMetadata(It.IsAny<string>())).ReturnsAsync(new ConnectionMetadata("dummyValue")); var cloudConnectionProvider = new CloudConnectionProvider(
+            metadataStore.Setup(m => m.GetMetadata(It.IsAny<string>())).ReturnsAsync(new ConnectionMetadata("dummyValue"));
+            var cloudConnectionProvider = new CloudConnectionProvider(
                 messageConverterProvider,
                 1,
                 new ClientProvider(),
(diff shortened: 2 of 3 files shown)


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionMetadata.cs
```diff
@@ -18,7 +18,7 @@ public ConnectionMetadata(string productInfo, string modelId, string edgeProduct
 
         public ConnectionMetadata(string productInfo, Option<string> modelId, string edgeProductInfo)
         {
-            Preconditions.CheckNonWhiteSpace(edgeProductInfo, nameof(edgeProductInfo));
```
and-rewsmith, line 21: You said you didn't want to cause breaking changes, right? Why can't we change the tests to send a edgeProductInfo?

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/EdgeHubConnectionTest.cs
```diff
@@ -62,7 +62,8 @@ public async Task TestEdgeHubConnection()
             string sasKey = ConnectionStringHelper.GetSharedAccessKey(deviceConnStr);
             var signatureProvider = new SharedAccessKeySignatureProvider(sasKey);
             var credentialsCache = Mock.Of<ICredentialsCache>();
-            var metadataStore = Mock.Of<IMetadataStore>();
+            var metadataStore = new Mock<IMetadataStore>();
+            metadataStore.Setup(m => m.GetMetadata(It.IsAny<string>())).ReturnsAsync(new ConnectionMetadata("dummyValue"));
```
and-rewsmith, line 66: Why is this change needed?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionMetadata.cs
and-rewsmith: Are you concerned about breaking changes from 1.0.10 RC to 1.0.10?

bronzeson: The breaking change I'm concerned about is that the old ProductInfo allows Empty string/whitespace as edge product info, but the new MetadataStore does not. This change relaxes it back to how the old ProductInfoStore had it.
So it's breaking change between 1.0.10 RC-1 -> 1.0.10 RC-2 and above

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/EdgeHubConnectionTest.cs
bronzeson: The metadata store object needs to be mocked because there is a GetMetadata in the new flow.
This GetMetadata is safe and will always return a ConnectionMetadata object, but if we don't mock it, the GetMetadata call fails and the test breaks

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionMetadata.cs
dylanbronson: The breaking change I'm concerned about is that the old ProductInfo allows Empty string/whitespace as edge product info, but the new MetadataStore does not. This change relaxes it back to how the old ProductInfoStore had it.
So it's breaking change between 1.0.10 RC-1 -> 1.0.10 RC-2 and above

edge-hub/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/EdgeHubConnectionTest.cs
dylanbronson: The metadata store object needs to be mocked because there is a MetadataStore.GetMetadata in the new flow.
This GetMetadata is safe and will always return a ConnectionMetadata object, but if we don't mock it, the GetMetadata call fails and the test breaks

PR #3369, opened 2020-08-06
EdgeHub-Bridge code review feedbacks from earlier

vipeller:
also removed the processing loops from certain handlers as it led to packet reordering (e.g. connection was handled later that the subscription, so the connection was not found for the subscriptions)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -91,6 +94,13 @@ public async Task ConnectAsync(string serverAddress, int port)
 
             client.ConnectionClosed += this.TriggerReconnect;
```
varunpuranik, line 95: What happens if you do this before the first connect ([line 62](https://github.com/Azure/iotedge/pull/3369/files#diff-cfa187e87717a80d693ad56d57a9c703L62))? Does it try to connect right away?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/MqttBridgeComponentDiscovery.cs
```diff
@@ -10,22 +10,18 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
...
     {
-        readonly ILogger logger;
-
         public IReadOnlyCollection<IMessageProducer> Producers { get; private set; }
         public IReadOnlyCollection<IMessageConsumer> Consumers { get; private set; }
 
         public static Type[] CandidateInterfaces { get; } = new[] { typeof(IMessageProducer), typeof(IMessageConsumer) };
 
-        public MqttBridgeComponentDiscovery(ILogger logger)
+        private MqttBridgeComponentDiscovery(IReadOnlyCollection<IMessageProducer> producers, IReadOnlyCollection<IMessageConsumer> consumers)
```
varunpuranik, line 18: Nit: remove private qualifier

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -26,6 +27,8 @@ public class MqttBrokerConnector : IMqttBrokerConnector
         Option<Task> forwardingLoop;
         Option<MqttClient> mqttClient;
 
+        int isRetrying = 0;
```
ancaantochi, line 30: Should this actually be a bool? You can use AtomicBoolean from utils

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -210,30 +220,18 @@ void TriggerReconnect(object sender, EventArgs e)
...
-
-                    client = this.mqttClient.Expect(() => new Exception("No mqtt-bridge connector instance found to use"));
+                    return;
                 }
 
-                // don't trigger it to ourselves while we are trying
-                client.ConnectionClosed -= this.TriggerReconnect;
-
-                var isConnected = false;
+                var client = default(MqttClient);
```
ancaantochi, line 229: no need to initialize with default value if you use MqttClient instead of vat

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -369,9 +404,19 @@ static async Task SubscribeAsync(MqttClient client, IReadOnlyCollection<IMessage
...
-
-                    client.MqttMsgSubscribed -= ConfirmSubscribe;
+                    try
+                    {
+                        await acksArrived.WaitAsync(TimeSpan.FromSeconds(SubAckTimeoutSecs));
+                    }
+                    catch (Exception ex)
+                    {
+                        Events.TimeoutReceivingSubAcks(ex);
+                        throw new Exception("MQTT server did not acknowledged subscriptions", ex);
```
ancaantochi, line 414: This should be a custom exception

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -369,9 +404,19 @@ static async Task SubscribeAsync(MqttClient client, IReadOnlyCollection<IMessage
                             Enumerable.Range(1, subscriber.Subscriptions.Count).Select(i => (byte)1).ToArray());
                     }
 
-                    await acksArrived.WaitAsync();
-
-                    client.MqttMsgSubscribed -= ConfirmSubscribe;
+                    try
```
ancaantochi, line 407: For some reason I can’t add comments to code that was not changed in this PR: allqosgranted should be AtomicBoolean and expectedAckCount could use AtomicLong

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/DirectMethodHandler.cs
```diff
@@ -90,24 +86,24 @@ async Task<bool> HandleMethodResponse(Match match, MqttPublishInfo publishInfo)
                                 ? this.identityProvider.Create(id1.Value, id2.Value)
                                 : this.identityProvider.Create(id1.Value);
 
-            var maybeProxy = await this.connectionRegistry.GetDeviceListenerAsync(identity);
-            var proxy = default(IDeviceListener);
+            var maybeListener = await this.connectionRegistry.GetDeviceListenerAsync(identity);
+            var listener = default(IDeviceListener);
```
ancaantochi, line 90: I would use the actual type instead of var and default to initialiaze

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -82,16 +82,16 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
                                 ? this.identityProvider.Create(id1.Value, id2.Value)
                                 : this.identityProvider.Create(id1.Value);
 
-            var proxy = default(IDeviceListener);
-            var proxyMaybe = await this.connectionRegistry.GetDeviceListenerAsync(identity);
+            var listener = default(IDeviceListener);
+            var maybeListener = await this.connectionRegistry.GetDeviceListenerAsync(identity);
 
-            if (!proxyMaybe.HasValue)
+            if (!maybeListener.HasValue)
```
ancaantochi, line 88: Use GetOrElse

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/TelemetryHandler.cs
```diff
@@ -59,16 +55,16 @@ async Task<bool> HandleTelemetry(Match match, MqttPublishInfo publishInfo)
...
 
-            var maybeProxy = await this.connectionRegistry.GetDeviceListenerAsync(identity);
-            var proxy = default(IDeviceListener);
+            var maybeListener = await this.connectionRegistry.GetDeviceListenerAsync(identity);
+            var listener = default(IDeviceListener);
 
             try
             {
-                proxy = maybeProxy.Expect(() => new Exception($"No device listener found for {identity.Id}"));
+                listener = maybeListener.Expect(() => new Exception($"No device listener found for {identity.Id}"));
```
ancaantochi, line 63: It’s the exception from expect caught right away by the try catch? It seems that you can just log and return false directly

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: How would that work if I want to return?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/TelemetryHandler.cs
vipeller: I need to extract the value from an Option, and because we don't have unwrap(), I need to call Expect(). The other way would be use something like GetOrDefault() and then check for null

PR #3370, opened 2020-08-07
Refactor REST endpoints authentication

richma-ms:
The AuthenticationMiddlerware was originally written when only TwinsController existed, and contains authentication logic that's specific to the invoke method REST endpoint.  Now that we have DeviceScope APIs as well, the middleware implementation no longer applies.

This change moves the logic that's common to all controller (authN) into a shared component (via dependency injection), and split the remaining API-specific logic (authZ) into their respective controller implementations.

Also updated existing tests / added new tests to cover all the changes.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/IEdgeHub.cs
```diff
@@ -17,6 +17,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
     /// </summary>
     public interface IEdgeHub : IDisposable
     {
+        string GetEdgeDeviceId();
```
varunpuranik, line 20: This is a const injected into the container.. why do we need this?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/IHttpRequestAuthenticator.cs
```diff
@@ -0,0 +1,25 @@
...
+            this.ErrorMsg = Preconditions.CheckNotNull(errorMsg);
+        }
+
+        public bool Authenticated { get; }
+        public string ErrorMsg { get; }
+    }
+
+    public interface IHttpRequestAuthenticator
+    {
+        Task<HttpAuthResult> AuthenticateRequest(string deviceId, Option<string> moduleId, HttpContext context);
```
varunpuranik, line 23: You can return a Try<bool> instead

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/HttpRequestAuthenticator.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            this.authenticator = Preconditions.CheckNotNull(authenticator, nameof(authenticator));
+            this.identityFactory = Preconditions.CheckNotNull(identityFactory, nameof(identityFactory));
+            this.iotHubName = Preconditions.CheckNonWhiteSpace(iotHubName, nameof(iotHubName));
+        }
+
+        public async Task<HttpAuthResult> AuthenticateRequest(string deviceId, Option<string> moduleId, HttpContext context)
+        {
+            Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId));
+            Preconditions.CheckNotNull(moduleId);
```
varunpuranik, line 38: Nit: no need, Option is a struct

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/HttpRequestAuthenticator.cs
```diff
@@ -0,0 +1,127 @@
...
+            Preconditions.CheckNotNull(context);
```
varunpuranik, line 39:
```suggestion
            Preconditions.CheckNotNull(context, nameof(context));
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/TwinsController.cs
```diff
@@ -82,18 +97,47 @@ internal static JRaw GetRawJson(byte[] bytes)
...
-                MethodResult methodResult = GetMethodResult(directMethodResponse);
-
-                return methodResult;
+            if (this.TryGetActorId(out string actorDeviceId, out string actorModuleId))
+            {
+                string actorId = $"{actorDeviceId}/{actorModuleId}";
+
+                if (actorDeviceId == currentEdgeDeviceId)
+                {
+                    IHttpRequestAuthenticator authenticator = await this.authenticatorGetter;
```
varunpuranik, line 112: We should move authentication to a Filter or to the OnActionExecuting method

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -35,8 +37,26 @@ public async Task GetDevicesAndModulesInTargetDeviceScopeAsync([FromRoute] strin
             actorDeviceId = WebUtility.UrlDecode(Preconditions.CheckNonWhiteSpace(actorDeviceId, nameof(actorDeviceId)));
             actorModuleId = WebUtility.UrlDecode(Preconditions.CheckNonWhiteSpace(actorModuleId, nameof(actorModuleId)));
 
-            EdgeHubScopeResult result = await this.HandleDevicesAndModulesInTargetDeviceScopeAsync(actorDeviceId, actorModuleId, request);
-            await this.SendResponse(result);
+            IHttpRequestAuthenticator authenticator = await this.authenticatorGetter;
```
varunpuranik, line 40: Same as the other - we should use a filter or move to OnActionExecuting

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/EdgeHubScopeResult.cs
```diff
@@ -11,6 +11,9 @@ public class EdgeHubScopeResult
         [JsonIgnore]
         public HttpStatusCode Status { get; set; }
 
+        [JsonIgnore]
+        public string ErrorMsg { get; set; }
```
philipktlin, line 15: why use short form? should it be ErrorMessage?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -248,20 +265,22 @@ async Task<bool> AuthorizeActorAsync(IDeviceScopeIdentitiesCache identitiesCache
...
+            this.Response.ContentType = SupportedContentType;
 
-                this.Response.ContentLength = resultUtf8Bytes.Length;
-                this.Response.ContentType = SupportedContentType;
+            await this.Response.Body.WriteAsync(resultUtf8Bytes, 0, resultUtf8Bytes.Length);
+        }
 
-                await this.Response.Body.WriteAsync(resultUtf8Bytes, 0, resultUtf8Bytes.Length);
-            }
+        async Task SendErrorResponse(HttpStatusCode status, string errorMsg)
```
philipktlin, line 280: Should this be consistent to return error in json format?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/IHttpRequestAuthenticator.cs
philipktlin, line 23: If it is an async method, please add Async suffix.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/IEdgeHub.cs
richma-ms: Previously this was used in the middleware, and we leveraged the appbuilder to pass in custom params.  But now it's used in the TwinsController, and I don't see a good way to pass the param from the DependencyManager into the controller instance.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/IHttpRequestAuthenticator.cs
richma-ms: I was trying to avoid using exceptions to wrap the error message, seems like exceptions should be reserved for when something goes wrong as opposed to expected auth errors.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/TwinsController.cs
richma-ms: Talked offline, turns out the auth filter approach doesn't work well, and the action filter approach is pretty awkward for passing in all the necessary params.  Leaving this as is for now.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
richma-ms: Split EdgeHubScopeResult into success/error types, and refactored the result construction to match.  Now that the response is always a EdgeHubScopeResult we can have errors in Json format.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/EdgeHubScopeResult.cs
```diff
@@ -6,18 +6,55 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
...
+
+        public EdgeHubScopeResultSuccess(IList<EdgeHubScopeDevice> devices, IList<EdgeHubScopeModule> modules)
+            : base(HttpStatusCode.OK)
+        {
+            this.Devices = devices;
+            this.Modules = modules;
+        }
+
+        [JsonIgnore]
+        public string ErrorMessage { get; set; }
```
philipktlin, line 37: don't need this property for success result, right?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/IHttpRequestAuthenticator.cs
```diff
@@ -0,0 +1,24 @@
...
+            this.ErrorMessage = Preconditions.CheckNotNull(errorMsg);
+        }
+
+        public bool Authenticated { get; }
+        public string ErrorMessage { get; }
+    }
+
+    public interface IHttpRequestAuthenticator
+    {
+        Task<HttpAuthResult> AuthenticateRequestAsync(string deviceId, Option<string> moduleId, HttpContext context);
```
philipktlin, line 22: nit: AuthenticateAsync(...)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/EdgeHubScopeResult.cs
richma-ms: oops forgot to remove

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/DependencyManager.cs
```diff
@@ -205,7 +205,7 @@ public void Register(ContainerBuilder builder)
                     true,
                     false));
 
-            builder.RegisterModule(new HttpModule());
+            builder.RegisterModule(new HttpModule("Edge1"));
```
philipktlin, line 208: what is the change used for?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/EdgeHubScopeResult.cs
```diff
@@ -6,18 +6,52 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
...
+        public EdgeHubScopeResultSuccess(IList<EdgeHubScopeDevice> devices, IList<EdgeHubScopeModule> modules)
+            : base(HttpStatusCode.OK)
+        {
+            this.Devices = devices;
+            this.Modules = modules;
+        }
 
         [JsonProperty(PropertyName = "devices", Required = Required.AllowNull)]
-        public IEnumerable<EdgeHubScopeDevice> Devices { get; set; }
+        public IList<EdgeHubScopeDevice> Devices { get; }
```
philipktlin, line 37: Why need to use IList?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/EdgeHubScopeResult.cs
```diff
@@ -6,18 +6,52 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
...
+
+    public class EdgeHubScopeResultError : EdgeHubScopeResult
+    {
+        public EdgeHubScopeResultError(HttpStatusCode status, string errorMessage)
+            : base(status)
+        {
+            this.ErrorMessage = errorMessage;
+        }
+
+        [JsonProperty("ErrorMessage")]
```
philipktlin, line 54: should it be small case "e", errorMessage?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -248,40 +233,34 @@ async Task<bool> AuthorizeActorAsync(IDeviceScopeIdentitiesCache identitiesCache
             return true;
         }
 
-        async Task SendResponse(EdgeHubScopeResult result)
+        async Task SendResponse(HttpStatusCode status, string responseJson)
```
philipktlin, line 236:
```suggestion
        async Task SendResponseAsync(HttpStatusCode status, string responseJson)
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/IHttpRequestAuthenticator.cs
```diff
@@ -0,0 +1,24 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public struct HttpAuthResult
+    {
+        public HttpAuthResult(bool authenticated, string errorMsg)
+        {
+            this.Authenticated = authenticated;
+            this.ErrorMessage = Preconditions.CheckNotNull(errorMsg);
```
philipktlin, line 13: CheckNotNull or CheckNonWhitespace?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/IHttpRequestAuthenticator.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Http.Controllers
+{
+    using System.Threading.Tasks;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.Azure.Devices.Edge.Util;
+
+    public struct HttpAuthResult
+    {
+        public HttpAuthResult(bool authenticated, string errorMsg)
```
philipktlin, line 10:
```suggestion
        public HttpAuthResult(bool authenticated, Option<string> errorMessage)
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/TwinsController.cs
```diff
@@ -82,18 +97,47 @@ internal static JRaw GetRawJson(byte[] bytes)
...
             }
+            else
+            {
+                methodResult = new MethodErrorResult(HttpStatusCode.BadRequest, $"Invalid header value for {Constants.ServiceApiIdHeaderKey}");
+            }
+
+            return methodResult;
         }
 
         async Task SendResponse(MethodResult methodResult)
```
philipktlin, line 143: It seems this method can be shared across DeviceScopeController and TwinController.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/EdgeHubScopeResult.cs
richma-ms: The caller needs to be able to add elements to the list (see DeviceScopeController:MakeResultFromIdentities)

PR #3374, opened 2020-08-07
MQTT Perf Tests Broker Build: Use same build procedure as mqttd for edgehub container

and-rewsmith:
The arm32 build is currently broken for perf testing mqttd. We can fix it by using the same build method we use for the edgehub container. This new build is better anyway because we have control over dependencies (cmake, openssl, rust toolchain).


builds/misc/images.yaml
```diff
@@ -26,7 +26,9 @@ jobs:
         name: build
         displayName: Build ($(Build.Configuration)) dotnet artifacts
 
-      - template: templates/build-broker-watchdog.yaml
```
and-rewsmith, line 29: This used to be one template but I split it up so that we can re-use it. We shouldn't reuse it though for perf tests until we start testing arm64.

builds/misc/templates/consolidate-edgehub-artifacts.yaml
```diff
@@ -2,24 +2,24 @@
...
+  mqtt.config.path: 'mqtt/contrib/edgehub/broker.json'
   watchdog.artifacts.path: 'edge-hub/watchdog/target'
   all.artifacts.path: 'publish/edge-hub'
 
 steps:
   - bash: |
       # Setup build context
       mkdir -p $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}
-      mkdir -p $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/mqttd
+      mkdir -p $(Build.BinariesDirectory)/${{ parameters['all.artifacts.path'] }}/mqtt
```
and-rewsmith, line 14: This directory needed to be changed for consistency. In the perf test build this consolidate script doesn't run. This leads to inconsistencies in the dockerfiles across container build (uses this consolidate template) and perf build (doesn't use).

This inconsistency can be fixed by naming the artifact containing folder "mqtt" vs "mqttd". Therefore in the dockerfiles for all builds we can follow the pattern:
```
ADD ./mqtt/target/[ARCH]/release/mqttd /usr/local/bin/mqttd
```

and-rewsmith: So this isn't totally true anymore since I decided to keep the build context small for the mqtt build. So instead in that dockerfile it is:
```
ADD ./target/[ARCH]/release/mqttd /usr/local/bin/mqttd
```

I am keeping this change because I still think it is cleaner.

and-rewsmith:
Builds to confirm it works:

images:
https://msazure.visualstudio.com/One/_build/results?buildId=33628432&view=results

images-release:
https://msazure.visualstudio.com/One/_build/results?buildId=33628429&view=results

images-mqtt:
https://dev.azure.com/msazure/One/_build/results?buildId=33628425&view=results

and-rewsmith:
The mqtt build started failing due to my most recent commit to shrink the build context. I still don't understand why. I am converting to draft and will reopen when I fix.

builds/misc/templates/consolidate-edgehub-artifacts.yaml
and-rewsmith: The build context shrink doesn't work for some reason. After investigation this is caused because post-build, the artifacts folder is missing in the build context. Right now I am thinking this is an azure pipelines problem. I will go back to the larger build context solution that works.

and-rewsmith: Wow I think I figured it out! There is a dockerignore in the mqtt folder which is why the target/ dir disappears when this is the build context. I will remove it.

mqtt/.dockerignore
```diff
@@ -4,7 +4,6 @@
 .gitignore
 .vscode
 state
-target
```
and-rewsmith, line 7: We should confirm removing this has no side effects. I'm assuming it won't?

dmolokanov: `target` contains a huge amount of binary artifacts that will be copied into the builder image context. But we don't do that anymore, right?

builds/misc/templates/consolidate-edgehub-artifacts.yaml
vadim-kovalyov: Hm, I'm not sure I understand the problem. I feels to me that removing .dockerignore is not the best thing to do. Maybe we should just copy executable into another folder?

mqtt/.dockerignore
vadim-kovalyov: Is it possible to copy build output (executable) at some point in some other folder? I would rather want to keep "target" in the .dockerignore

builds/misc/templates/consolidate-edgehub-artifacts.yaml
and-rewsmith: Yes that was what I was thinking too after Denis made his comment. For the edgehub build, we have consolidate-edgehub-artifacts template that does similar copying. I will follow the pattern and fix it.

and-rewsmith: Although the consolidate-edgehub-artifacts suffers from the same issue. I will add a task for fixing this under the PBI to decrease the edgehub container build time.

and-rewsmith:
I added a template to copy only the desired binaries from the target folder. Since the edgehub build copies the whole target folder into the build context, I will switch to using the same template for the edgehub build in a separate PR.

builds/misc/templates/move-rust-artifacts.yaml
```diff
@@ -0,0 +1,41 @@
...
+#
+# [destination.folder.path]/
+#  - [target.amd64]/release
+#     - [artifact.name]
+#  - [target.arm32v7]/release
+#     - [artifact.name]
+#  - [target.arm64v8]/release
+#     - [artifact.name]
+parameters:
+  target.arm32v7: 'armv7-unknown-linux-gnueabihf'
```
vadim-kovalyov, line 13: nit: can params be reduced to 
- target: 'armv7-unknown-linux-gnueabihf' | ....
- arch: 'arm32v7' | 'arm64v8' | 'amd64'

mqtt/.dockerignore
vadim-kovalyov: This should be reverted, right?

builds/misc/templates/move-rust-artifacts.yaml
and-rewsmith: I wanted to avoid having to specify the target when calling the template. If you still think this way is better then I can change it.

PR #3379, opened 2020-08-10
Add until to log and support bundle upload

lfitchett:
This adds support for the until parameter recently added to docker logs.


edgelet/support-bundle/src/support_bundle.rs
```diff
@@ -247,33 +258,49 @@ where
...
-            .arg("-Command")
-            .arg(&format!(r"Get-WinEvent -ea SilentlyContinue -FilterHashtable @{{ProviderName='iotedged';LogName='application';StartTime='{}'}} |
-                            Select TimeCreated, Message |
-                            Sort-Object @{{Expression='TimeCreated';Descending=$false}} |
-                            Format-List", since_time.to_rfc3339()))
-            .output();
+        let command = {
+            let until = until_time
+                .map(|until| format!(";EndTime='{}'", until))
+                .unwrap_or_else(|| "".to_owned());
```
arsing, line 285: `.unwrap_or_else(String::new)`

edgelet/support-bundle/src/support_bundle.rs
```diff
@@ -157,14 +157,25 @@ where
...
             };
+            let until_string = if let Some(until) = until_local {
+                format!(" until {}", until)
+            } else {
+                "".to_string()
+            };
             println!(
-                "Writing all logs {}since {} (local time {})",
-                max_lines, since_time, since_local
+                "Writing all logs since {} (local time {}){} with {} max_lines",
```
arsing, line 177: This is going to look strange if `max_lines` is `""`, isn't it?

edgelet/docker-rs/src/apis/container_api.rs
```diff
@@ -817,21 +818,27 @@ where
...
+        let mut serializer = ::url::form_urlencoded::Serializer::new(String::new());
+        serializer
             .append_pair("follow", &follow.to_string())
             .append_pair("stdout", &stdout.to_string())
             .append_pair("stderr", &stderr.to_string())
             .append_pair("since", &since.to_string())
             .append_pair("timestamps", &timestamps.to_string())
-            .append_pair("tail", &tail.to_string())
-            .finish();
+            .append_pair("tail", &tail.to_string());
```
arsing, line 836: I know this is old code, but `.append_pair("tail", tail);`

edgelet/support-bundle/src/support_bundle.rs
```diff
@@ -157,14 +157,25 @@ where
...
+            });
             let max_lines = if let LogTail::Num(tail) = self.log_options.tail() {
                 format!("(maximum {} lines) ", tail)
             } else {
                 "".to_owned()
             };
+            let until_string = if let Some(until) = until_local {
+                format!(" until {}", until)
+            } else {
+                "".to_string()
```
arsing, line 174: `"".to_owned()` to be consistent, though personally I would've used `.unwrap_or_else(String::new)` in both places, just like the suggestion below.

edgelet/support-bundle/src/support_bundle.rs
lfitchett: you're right, I forgot I already formatted it

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/RuntimeInfoProvider.cs
```diff
@@ -89,6 +89,8 @@ public Task<Stream> GetModuleLogs(string module, bool follow, Option<int> tail,
             };
             tail.ForEach(t => containerLogsParameters.Tail = t.ToString());
             since.ForEach(t => containerLogsParameters.Since = t.ToString());
+
+// TODO: UnTIL
```
damonbarry, line 93: Remove?

lfitchett: yep

lfitchett:
> I don't see any new tests for this functionality in the C# code, just updates to existing tests. Do we need new tests?

Those tests are all testing the parsing code, they never actually call out to docker. Since all this does is pass the param to docker, there isn't much to add a test for.

PR #3390, opened 2020-08-11
Fix Subscription Processing Workaround

dylanbronson:
Removing a workaround that we added for 1.0.9 that fixed an issue we were having that when an MQTT identity with only subscriptions (no telemetry) reauthenticated, the identity wouldn't reconnect to the cloud. (PR of previous problem: https://github.com/Azure/iotedge/pull/3201)

The new way of doing this is to listen to when a device gets connected to EdgeHub and process subscriptions when it does.
To prevent a re-triggering of subscription processing when EdgeHub connects to the cloud on behalf of the device (this situation occurs during the device connecting to edgeHub flow), we use a ConcurrentDictionary to keep track of if subscription processing has already been kicked off for a DeviceConnecting. We check this in the other methods. 
This effectively ignores the Cloud Connection event trigger.


edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessor.cs
```diff
@@ -44,11 +45,34 @@ public SubscriptionProcessor(IConnectionManager connectionManager, IInvokeMethod
             Preconditions.CheckNotNull(deviceConnectivityManager, nameof(deviceConnectivityManager));
             this.invokeMethodHandler = Preconditions.CheckNotNull(invokeMethodHandler, nameof(invokeMethodHandler));
             this.pendingSubscriptions = new ConcurrentDictionary<string, ConcurrentQueue<(DeviceSubscription, bool)>>();
+            this.subscriptionsBeingProcessed = new ConcurrentDictionary<string, bool>();
```
varunpuranik, line 48: Nit: Initialize inline

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessor.cs
```diff
@@ -44,11 +45,34 @@ public SubscriptionProcessor(IConnectionManager connectionManager, IInvokeMethod
...
+                Events.ProcessingSubscriptionsOnDeviceConnectedToEdgeHub(identity);
+                await this.ProcessExistingSubscriptions(identity.Id);
+            }
+            catch (Exception e)
+            {
+                Events.ErrorProcessingSubscriptions(e, identity);
+            }
+            finally
+            {
+                this.ConnectionManager.CloudConnectionEstablished += this.ClientCloudConnectionEstablished;
```
varunpuranik, line 71: Why is this needed?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessor.cs
```diff
@@ -44,11 +45,34 @@ public SubscriptionProcessor(IConnectionManager connectionManager, IInvokeMethod
...
             this.invokeMethodHandler = Preconditions.CheckNotNull(invokeMethodHandler, nameof(invokeMethodHandler));
             this.pendingSubscriptions = new ConcurrentDictionary<string, ConcurrentQueue<(DeviceSubscription, bool)>>();
+            this.subscriptionsBeingProcessed = new ConcurrentDictionary<string, bool>();
             this.processSubscriptionsBlock = new ActionBlock<string>(this.ProcessPendingSubscriptions);
+            connectionManager.DeviceConnected += this.DeviceConnected;
             deviceConnectivityManager.DeviceConnected += this.CloudConnectivityEstablished;
             connectionManager.CloudConnectionEstablished += this.ClientCloudConnectionEstablished;
         }
 
+        async void DeviceConnected(object sender, IIdentity identity)
```
varunpuranik, line 55: Instead of this, can we use the same callback for both DeviceConnected and CloudConnectionEstablished and if an identity is being processed, then skip processing it again?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessor.cs
```diff
@@ -142,11 +144,29 @@ async Task ProcessSubscriptionByIdentity(IIdentity identity)
...
+        void ClientConnectionToEdgeHubEstablished(object sender, IIdentity identity)
+        {
+            this.ClientConnectionEstablished(identity, false);
+        }
+
+        async void ClientConnectionEstablished(IIdentity identity, bool connectedToCloud)
+        {
+            if (connectedToCloud)
+            {
+                Events.ClientConnectedToCloudProcessingSubscriptions(identity);
```
varunpuranik, line 161: Nit: Move the events to the parent and remove this method.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessor.cs
```diff
@@ -142,11 +144,29 @@ async Task ProcessSubscriptionByIdentity(IIdentity identity)
...
+        }
+
+        void ClientConnectionToEdgeHubEstablished(object sender, IIdentity identity)
+        {
+            this.ClientConnectionEstablished(identity, false);
+        }
+
+        async void ClientConnectionEstablished(IIdentity identity, bool connectedToCloud)
+        {
+            if (connectedToCloud)
```
ancaantochi, line 159: nit: connectedToCloud could be passed to the Events method so the if can be inside it

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/SubscriptionProcessor.cs
```diff
@@ -157,16 +177,27 @@ async void ClientCloudConnectionEstablished(object sender, IIdentity identity)
...
                     {
-                        await this.ProcessSubscriptionWithRetry(id, cloudProxy, subscription.Key, subscription.Value);
-                    }
-                });
+                        foreach (KeyValuePair<DeviceSubscription, bool> subscription in s)
+                        {
+                            await this.ProcessSubscriptionWithRetry(id, cloudProxy, subscription.Key, subscription.Value);
+                        }
+                    });
+                this.subscriptionsBeingProcessed.TryRemove(id, out _);
```
ancaantochi, line 199: should have try finally and remove in finally to make sure it gets removed after it was processed

varunpuranik: +1

PR #3391, opened 2020-08-11
Proxy Support for `iotedge check`

lfitchett:
iotedge check will now use a proxy server if specified in config.yaml

This also rewrites the diagnostics module in c# since it would be difficult to get openssl in the container for rust to use.


edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -0,0 +1,97 @@
...
+                ProxyClientFactory factory = new ProxyClientFactory();
+                IProxyClient proxyClient = factory.CreateProxyClient(ProxyType.Http, proxyAddress, proxyPort);
+
+                // Setup timeouts
+                proxyClient.ReceiveTimeout = (int)TimeSpan.FromSeconds(60).TotalMilliseconds;
+                proxyClient.SendTimeout = (int)TimeSpan.FromSeconds(60).TotalMilliseconds;
+
+                // Get TcpClient to futher work
+                var client = proxyClient.CreateConnection(hostname, int.Parse(port));
+                client.GetStream();
```
arsing, line 87: Need to do a TLS handshake too.

edgelet/iotedge/src/check/checks/container_connect_iothub.rs
```diff
@@ -107,16 +110,27 @@ impl ContainerConnectIotHub {
...
             &port,
         ]);
-        self.diagnostics_image_name = Some(check.diagnostics_image_name.clone());
+
+        let proxy = settings
+            .agent()
+            .env()
+            .get("https_proxy")
+            .map(std::string::String::as_str);
+        self.proxy = proxy.map(std::borrow::ToOwned::to_owned);
```
arsing, line 130: `ToOwned` is already in prelude, so you don't need to use its full path.

edgelet/iotedge/src/check/checks/host_connect_dps_endpoint.rs
```diff
@@ -102,3 +128,37 @@ pub fn resolve_and_tls_handshake(
...
+    tls_hostname: String,
+    port: Option<u16>,
+    proxy: String,
+) -> impl Future<Item = (), Error = Error> {
+    futures::future::ok(())
+        .and_then({
+            move |_| -> Result<_, Error> {
+                let proxy_uri = Uri::from_str(&proxy)
+                    .with_context(|_| format!("Could not make proxi uri from {}", proxy))?;
+                let proxy_obj = Proxy::new(Intercept::All, proxy_uri);
```
arsing, line 142: Need to handle username and password in `proxy_uri` by extracting them and calling `proxy_obj.set_authorization`. See `hyperwrap::uri_to_proxy` for an example.

edgelet/iotedge/src/check/checks/host_connect_dps_endpoint.rs
```diff
@@ -102,3 +128,37 @@ pub fn resolve_and_tls_handshake(
...
+                let http_connector = HttpConnector::new(1);
+                let proxy_connector = ProxyConnector::from_proxy(http_connector, proxy_obj)
```
arsing, line 144: For TLS proxies `HttpConnector` won't be sufficient. You'll need an `HttpsConnector`.

You don't need to check the scheme. Just always use an `HttpsConnector`. It can handle `http://` too.

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
lfitchett: client.GetStream(); sends a Connect request

arsing: It creates a TCP connection to the IoT Hub through the proxy. I'm talking about the TLS handshake with the IoT Hub over that TCP connection.

lfitchett: all the old module did was call `TcpStream::connect_timeout`?

lfitchett:
I have an auto-formatter for yaml files that made it look like there were a large number of changes to the build yaml. I reverted those so now the only changes are relevant.

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
arsing: Yes, because it didn't have openssl. Now that you have a TLS stack, you might as well use it.

edgelet/iotedge/src/check/checks/host_connect_dps_endpoint.rs
lfitchett: Done. Had to add hyper_tls to the project

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
lfitchett: I think it already might? I can write to the stream, so presumably its done the handshake too

arsing: All you've given the `proxyClient` is the hostname and port you want it to proxy to. It doesn't know anything about what protocol you're going to use on top of the connection, so it can't possibly know that you want to run TLS over this connection, so it can't possibly have done a TLS handshake for you.

edge-modules/iotedge-diagnostics-dotnet/src/GetSocket.cs
```diff
@@ -0,0 +1,43 @@
...
+            byte[] bytesReceived = new byte[256];
+            server = server.Replace("unix://", string.Empty);
+
+            // Create a socket connection with the specified server and port.
+            using (Socket socket = new Socket(AddressFamily.Unix, SocketType.Stream, ProtocolType.IP))
+            {
+                socket.Connect(new UnixDomainSocketEndPoint(server));
+
+                if (socket == null)
+                    return "Connection failed";
```
damonbarry, line 26: `socket` will set itself to `null` if `Connect` fails? That seems weird...

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
arsing: As discussed, this may not be possible to do in C#, and is likely to be redundant with the host's TLS check anyway, so it's fine to not do it.

builds/misc/images-release.yaml
```diff
@@ -313,18 +313,12 @@ jobs:
...
-      - bash: |
-            scripts/linux/buildEdgelet.sh -i azureiotedge-diagnostics -n microsoft -P iotedge-diagnostics -c $(configuration) --target-arch armv7l &&
-            scripts/linux/buildImage.sh -r $(registry.address) -u $(registry.user) -p $(registry.password) -v $(version) -i azureiotedge-diagnostics -n $(namespace) -P azureiotedge-diagnostics --target-arch armv7l
-        displayName: 'Build Image - azureiotedge-diagnostics - ARM32'
-      - bash: |
-            scripts/linux/buildEdgelet.sh -i azureiotedge-diagnostics -n microsoft -P iotedge-diagnostics -c $(configuration) --target-arch aarch64 &&
-            scripts/linux/buildImage.sh -r $(registry.address) -u $(registry.user) -p $(registry.password) -v $(version) -i azureiotedge-diagnostics -n $(namespace) -P azureiotedge-diagnostics --target-arch aarch64
-        displayName: 'Build Image - azureiotedge-diagnostics - ARM64'
+       - template: templates/image-linux.yaml
+        parameters:
```
arsing, line 317: Indentation is wrong here.

edge-modules/iotedge-diagnostics-dotnet/src/GetSocket.cs
lfitchett: ah, nope, that's leftover from before. Deleted. Connect just throws an exception

builds/misc/images-release.yaml
```diff
@@ -355,11 +349,6 @@ jobs:
       - powershell: "scripts/windows/build/Publish-Branch.ps1 -Configuration:\"$(configuration)\" -PublishTests:$False -UpdateVersion"
         displayName: "Build ($(configuration))"
         name: build
-      - task: PowerShell@2
-        displayName: "Build iotedge-diagnostics"
-        inputs:
-          filePath: ./edgelet/build/windows/build-diagnostics.ps1
```
and-rewsmith, line 361: if we aren't going to use this script anymore I think we can delete it

lfitchett: yep. I deleted the linux one, missed windows

builds/misc/images-release.yaml
```diff
@@ -526,11 +515,11 @@ jobs:
               ]
           signConfigType: inlineSignParams
       - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@1
-        displayName: "azureiotedge-diagnostics Code Sign"
-        inputs: 
+        displayName: "dotnet-azureiotedge-diagnostics Code Sign"
```
philipktlin, line 518: Why do you change the step name?

lfitchett: I guess I didn't have to change the name, I was just going through and changing everything and got the name as well

builds/misc/images-release.yaml
```diff
@@ -313,18 +313,12 @@ jobs:
...
-        displayName: 'Build Image - azureiotedge-diagnostics - ARM32'
-      - bash: |
-            scripts/linux/buildEdgelet.sh -i azureiotedge-diagnostics -n microsoft -P iotedge-diagnostics -c $(configuration) --target-arch aarch64 &&
-            scripts/linux/buildImage.sh -r $(registry.address) -u $(registry.user) -p $(registry.password) -v $(version) -i azureiotedge-diagnostics -n $(namespace) -P azureiotedge-diagnostics --target-arch aarch64
-        displayName: 'Build Image - azureiotedge-diagnostics - ARM64'
+      - template: templates/image-linux.yaml
+        parameters:
+          imageName: azureiotedge-diagnostics
+          name: "Diagnostics Module"
+          project: IotedgeDiagnosticsDotnet
```
philipktlin, line 320: should the project name be Microsoft.Azure.Devices.Edge.Diagnostics?

lfitchett: Its in the same place as SimulatedTemperatureSensor, and I named it the same way

edge-modules/iotedge-diagnostics-dotnet/docker/linux/amd64/Dockerfile
```diff
@@ -0,0 +1,17 @@
+ARG base_tag=3.1.4-alpine3.11
+FROM mcr.microsoft.com/dotnet/core/runtime:${base_tag}
+
+ARG EXE_DIR=.
+
+ENV MODULE_NAME "DiagnosticsModule.dll"
```
philipktlin, line 6: why "DiagnosticsModule.dll" is different from project name IotedgeDiagnosticsDotnet?

edge-modules/iotedge-diagnostics-dotnet/docker/linux/arm64v8/base/Dockerfile
```diff
@@ -0,0 +1,6 @@
+ARG base_tag=3.1.5-bionic-arm64v8
```
philipktlin, line 1: you don't need to this docker file as we only need 1 copy of docker file for module-base docker image.

edge-modules/iotedge-diagnostics-dotnet/docker/linux/arm32v7/base/Dockerfile
```diff
@@ -0,0 +1,6 @@
+ARG base_tag=3.1.5-bionic-arm32v7
```
philipktlin, line 1: you don't need to this docker file as we only need 1 copy of docker file for module-base docker image.

edge-modules/iotedge-diagnostics-dotnet/docker/windows/arm32v7/base/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Base image for arm32v7 dotnet-runtime container
```
philipktlin, line 1: you don't need to this docker file as we only need 1 copy of docker file for module-base docker image.

edge-modules/iotedge-diagnostics-dotnet/src/GetSocket.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+namespace Diagnostics
```
philipktlin, line 3: I think this should be similar to Edge Agent and Edge Hub projects, it should start with Microsoft.Azure.Devices.Edge.

edge-modules/iotedge-diagnostics-dotnet/src/GetSocket.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    using System;
+    using System.IO;
+    using System.Net;
+    using System.Net.Sockets;
+    using System.Text;
+
+    public class GetSocket
```
philipktlin, line 11: Normally use noun for class name in C#

philipktlin: Do we need to be public? can it be internal?

edge-modules/iotedge-diagnostics-dotnet/src/GetSocket.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        public static string GetSocketResponse(string server, string endpoint)
+        {
+            string request = $"GET {endpoint} HTTP/1.1\r\nHost: {server}\r\nConnection: Close\r\n\r\n";
+            byte[] bytesSent = Encoding.ASCII.GetBytes(request);
+            byte[] bytesReceived = new byte[256];
+            server = server.Replace("unix://", string.Empty);
```
philipktlin, line 18: check server is not null or empty.

edge-modules/iotedge-diagnostics-dotnet/src/GetSocket.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            // Create a socket connection with the specified server and port.
+            using (Socket socket = new Socket(AddressFamily.Unix, SocketType.Stream, ProtocolType.IP))
+            {
+                socket.Connect(new UnixDomainSocketEndPoint(server));
```
philipktlin, line 23: Do you need to catch exception here?

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Client;
+    using Microsoft.Azure.Devices.Client.Transport.Mqtt;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Azure.Devices.Edge.Util.Concurrency;
+    using Microsoft.Azure.Devices.Edge.Util.TransientFaultHandling;
+    using Microsoft.Azure.Devices.Shared;
+    using Microsoft.Extensions.Configuration;
+    using Newtonsoft.Json;
+    using ProxyLib.Proxy;
```
philipktlin, line 21: Did you check the license for this library?

edge-modules/iotedge-diagnostics-dotnet/docker/linux/amd64/Dockerfile
lfitchett: I guess it should be IotedgeDiagnosticsDotnet, I didn't notice since its never run normally. I should probably change it to an echo that says as much

edge-modules/iotedge-diagnostics-dotnet/docker/linux/arm64v8/base/Dockerfile
lfitchett: its modeled after temp sensor, which does that

edge-modules/iotedge-diagnostics-dotnet/src/GetSocket.cs
lfitchett: If it is, it throws an exception, which is the intended behavior

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+                    break;
+                case "local-time":
+                    Console.WriteLine(DateTime.Now.ToUnixTimestamp());
+                    break;
+                default:
+                    throw new Exception("Invalid args");
+            }
+        }
+
+        static async Task EdgeAgent(string managementUri)
```
philipktlin, line 60: Normally starts with verb for method name in C# and uses Async suffix for async method.

edge-modules/iotedge-diagnostics-dotnet/src/GetSocket.cs
lfitchett: nope, thats the point, it should throw exception if it can't connect. Its just testing if it can connect

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
lfitchett: MIT, I probably need to add it to the sources list though

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+                return 1;
+            }
+
+            return 0;
+        }
+
+        static async Task MainAsync(string[] args)
+        {
+            var config = new ConfigurationBuilder().AddCommandLine(args).Build();
+            switch (args[0])
```
philipktlin, line 44: do we need to check args[0] exists to avoid index out of range exception?

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                case "edge-agent":
+                    await EdgeAgent(config["management-uri"]);
+                    break;
+                case "iothub":
+                    await Iothub(config["hostname"], config["port"], config["proxy"]);
+                    break;
+                case "local-time":
+                    Console.WriteLine(DateTime.Now.ToUnixTimestamp());
```
philipktlin, line 53: Just want to confirm if ToUnixTimestamp() is the right format to use.

edge-modules/iotedge-diagnostics-dotnet/src/GetSocket.cs
lfitchett: Its not in that namespace though. I don;t think this matters that much

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
philipktlin: And do we need to validate each argument value not null or empty?

lfitchett: it'll just error, which is fine. This is called from the iotedge check script, which will always pass a param. I don;t expect customers to be running this manually

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
lfitchett: yep, that's what iotedge check expects

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+
+            if (!modules.StartsWith("HTTP/1.1 200 OK"))
+            {
+                throw new Exception($"Got bad response: {modules}");
+            }
+        }
+
+        static async Task Iothub(string hostname, string port, string proxy)
```
philipktlin, line 83: should we use ushort for port argument?

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+            ProxyClientFactory factory = new ProxyClientFactory();
+            if (proxyUri.UserInfo == string.Empty)
+            {
+                return factory.CreateProxyClient(ProxyType.Http, proxyUri.Host, proxyUri.Port);
+            }
+            else
+            {
+                if (proxyUri.UserInfo.Contains(':'))
+                {
+                    var userPass = proxyUri.UserInfo.Split(':');
```
philipktlin, line 118: nit: name variable as credential

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
philipktlin: Please add comments for this detail here.

lfitchett:
Talked with philip and we agreed that since I don't expect the diagnostics module to be changed much (and its so short) there isn't a large upkeep cost to maintaining it if it doesn't follow the strict standards of the main repo. Most style comments will not be addressed

edge-modules/iotedge-diagnostics-dotnet/docker/linux/arm64v8/base/Dockerfile
philipktlin: tempSensor is the only module containing azureiotedge-module-base image docker file

lfitchett:
Arnav and Damon already approved, there haven't been any changes to anything except the build pipeline since then

PR #3393, opened 2020-08-12
MQTT Broker Client Disconnection

and-rewsmith:
We can use an mqtt3 client to listen for commands on certain topics then send signals to the broker. We can implement many scenarios with this pattern, but this PR only contains client disconnection.


mqtt/mqttd/src/broker/mod.rs
```diff
@@ -12,10 +12,13 @@ use tokio::{
...
 
+use mqtt3::ShutdownError;
 use mqtt_broker::{
-    BrokerHandle, FilePersistor, Message, Persist, ShutdownHandle, Snapshotter,
-    StateSnapshotHandle, SystemEvent, VersionedFileFormat,
+    BrokerHandle, FilePersistor, Message, Persist, ShutdownHandle as SnapshotShutdownHandle,
+    Snapshotter, StateSnapshotHandle, SystemEvent, VersionedFileFormat,
 };
+use mqtt_edgehub::command::{CommandHandler, ShutdownHandle as CommandShutdownHandle};
+use mqtt_edgehub::settings::ListenerConfig;
```
and-rewsmith, line 21: This will break the generic build so need to fix

and-rewsmith: Can I move all the command handler logic to bootstrap/edgehub.rs?

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -0,0 +1,249 @@
...
+            }
+
+            info!("succeeded sending broker signal to disconnect client");
+        }
+
+        Ok(())
+    }
+}
+
+async fn wait_for_suback(client: &mut Client<BrokerConnection>) -> Result<(), InitializationError> {
```
and-rewsmith, line 125: I want this to throw an initialization error if this suback wait "times out" of some arbitrary threshold. Any suggestions for how to do it? 

```while client.next()``` body waits until the next event is received so you can't break out of the loop. I was thinking of spawning a new tokio thread and then shutting it down when it timed out. Is this a good idea?

**Comment by and-rewsmith** [comment]
Moving to draft until we decide if we want to take sidecar in this PR

File: mqtt/mqtt-edgehub/src/command.rs (hunk 2 above)
Comment (and-rewsmith) [reply mqtt/mqtt-edgehub/src/command.rs]: talked with Denis and we decided that it is fine if the suback wait doesn't have a timeout, since the client will shutdown if the broker fails.

File: mqtt/mqtt-edgehub/tests/command.rs [hunk 3]
```diff
@@ -0,0 +1,74 @@
...
+    command_handler_shutdown_handle
+        .shutdown()
+        .await
+        .expect("failed to stop command handler client");
+    join_handle
+        .await
+        .expect("failed to shutdown command handler");
+
+    edgehub_client.shutdown().await;
+}
```
and-rewsmith, line 61: I tried to add tests where the command handler threw the initialization error, but there is only one case where this happens (failed subscribe). I couldn't think of a way to induce this failure. It isn't as simple as starting the command handler before the broker because the mqtt3 client will add the subscription to a queue of update and return success.

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -0,0 +1,248 @@
...
+use tokio::net::TcpStream;
+use tracing::{error, info, warn};
+
+use mqtt3::{
+    proto, Client, Event, IoSource, ShutdownError, SubscriptionUpdateEvent, UpdateSubscriptionError,
+};
+use mqtt_broker::{BrokerHandle, Error, Message, SystemEvent};
+
+const TOPIC_FILTER: &str = "$edgehub/+/disconnect";
+const CLIENT_EXTRACTION_REGEX: &str = r"\$edgehub/(.*)/disconnect";
```
dmolokanov, line 15: client_id is a bit more complicated https://github.com/Azure/iotedge/blob/iiot/mqtt/mqtt-edgehub/src/topic/translation.rs#L66

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -0,0 +1,248 @@
...
+    let captures = REGEX
+        .captures(topic_name.as_ref())
+        .ok_or_else(DisconnectClientError::RegexFailure)?;
+
+    let value = captures
+        .get(1)
+        .ok_or_else(DisconnectClientError::RegexFailure)?;
+
+    let client_id = value.as_str();
+    match client_id {
```
dmolokanov, line 162: redundant allocations
```rust 
    if value.empty() {
         Err(DisconnectClientError::NoClientId())
  } else {
        Ok(value)
    }
```

mqtt/mqtt-broker/src/error.rs
```diff
@@ -65,6 +65,9 @@ pub enum Error {
 
     #[error("An error occurred when constructing state change: {0}")]
     StateChange(#[from] serde_json::Error),
+
+    #[error("Failed to shutdown custom mqtt client: {0}")]
+    ShutdownClient(#[from] mqtt3::ShutdownError),
```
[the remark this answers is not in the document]
and-rewsmith: We use it if the shutting down the client fails.

mqtt/mqtt-edgehub/src/command.rs
and-rewsmith: ```value``` is a match type not a str type, so there is no method empty()

mqtt/mqtt-edgehub/src/command.rs
and-rewsmith: By more complicated, what do you mean? I see that it can be of "device/module" or "device" form but both will be captured by the existing regex?

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -0,0 +1,248 @@
...
+pub struct CommandHandler {
+    broker_handle: BrokerHandle,
+    client: Client<BrokerConnection>,
+}
+
+impl CommandHandler {
+    pub fn new(
+        broker_handle: BrokerHandle,
+        address: String,
+        device_id: &str,
```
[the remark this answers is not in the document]
and-rewsmith: clippy made me do this because it doesn't get consumed in the function

mqtt/mqtt-edgehub/tests/command.rs
and-rewsmith: I pushed a test that did this expecting the command handler to throw an error indicating it didn't receive the suback. It's hanging though in the ```subscribe()``` function when calling ```client.try_next()```. I haven't figured out why yet.

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -13,10 +13,17 @@ use tokio::{
 use tracing::{info, warn};
 
 use mqtt_broker::{
-    BrokerHandle, FilePersistor, Message, Persist, ShutdownHandle, Snapshotter,
-    StateSnapshotHandle, SystemEvent, VersionedFileFormat,
+    BrokerHandle, FilePersistor, Message, Persist, ShutdownHandle as SnapshotShutdownHandle,
+    Snapshotter, StateSnapshotHandle, SystemEvent, VersionedFileFormat,
 };
 
+#[cfg(feature = "edgehub")]
```
and-rewsmith, line 20: Note: These cfg do not currently work to support compiling with the generic flag. There will be a second PR to support sidecars that will fix this.

mqtt/mqtt-edgehub/tests/command.rs
```diff
@@ -0,0 +1,106 @@
...
+    let (_, join_handle) = start_command_handler(broker_handle, TEST_SERVER_ADDRESS.to_string())
+        .await
+        .expect("could not start command handler");
+
+    let command_handler_status = join_handle
+        .await
+        .expect("failed to shutdown command handler");
+
+    assert_matches!(command_handler_status, Err(_));
+}
```
and-rewsmith, line 87: This test currently fails because it hangs on ```subscribe()``` when calling client.try_next(). I haven't figured out why yet. I was assuming that if the broker denies auth then the client will get disconnected and the stream will get a ```None```? This doesn't appear to be the case though due to the hang. Any ideas?

mqtt/mqtt-edgehub/src/command.rs
vadim-kovalyov: I think this should be enough. In topic translation we need to extract deviceId and moduleId from the clientId, here we just need to know the string to drop the connection.

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -126,6 +163,44 @@ async fn server_certificate_renewal(renew_at: DateTime<Utc>) {
...
+        tx.await?;
+
+        command_handler_shutdown_handle.shutdown().await?;
+        command_handler_join_handle.await??;
+        info!("command handler shutdown.");
+        Ok::<(), anyhow::Error>(())
+    });
+
+    (SidecarShutdownHandle(termination_handle), event_loop)
+}
```
and-rewsmith, line 186: A few comments on this structure:
- This function is called by start_server() and is not publicly accessible. I chose to put it only in the edgehub bootstrap because currently we have no generic-feature-only sidecars.
- I have this nested shutdown handle structure because if someone wants to add a new sidecar, the won't have to add more code to the start_server() method.

mqtt/mqtt-edgehub/tests/command.rs
and-rewsmith: I think I know why. The client in the first passing test is just a stream which returns none when disconnected. Mqtt3 will try to reconnect so therefore it hangs.

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -0,0 +1,266 @@
...
+
+        if !self.subscribe(subscribe_topics).await? {
+            return Err(CommandHandlerError::MissingSubacks(
+                subscribe_topics.concat(),
+            ));
+        }
+
+        while let Some(event) = self
+            .client
+            .try_next()
```
vadim-kovalyov, line 88: nit: just curious why try_next? in what circumstances poll'ing the client can panic?

mqtt/mqtt-edgehub/tests/command.rs
```diff
@@ -0,0 +1,81 @@
...
+    )
+    .await;
+    test_client.next().await; // skip connack
+
+    let mut edgehub_client = TestClientBuilder::new(server_handle.address())
+        .with_client_id(ClientId::IdWithCleanSession("$edgehub".into()))
+        .build();
+
+    let topic = "$edgehub/test-client/disconnect";
+    edgehub_client.publish_qos1(topic, "qos 1", false).await;
```
vadim-kovalyov, line 48: hm, this test made me think... why do we put client id into the topic and use regex to extract it? can't we just put client id in the payload of the publication to "$edgehub/disconnect"?

and-rewsmith: Vadim and I talked offline and decided that this was a good change, because then we don't need regex logic / tests. It was originally done this way for consistency. I think it is best to merge the working version now and fix later.

mqtt/mqtt-edgehub/src/command.rs
and-rewsmith: I'm not sure why it could panic, but thought this was safer

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -0,0 +1,266 @@
...
+            .shutdown_handle()
+            .map_or(Err(ShutdownError::ClientDoesNotExist), |shutdown_handle| {
+                Ok(ShutdownHandle(shutdown_handle))
+            })
+    }
+
+    pub async fn run(mut self) -> Result<(), CommandHandlerError> {
+        let subscribe_topics = &[TOPIC_FILTER.to_string()];
+
+        if !self.subscribe(subscribe_topics).await? {
```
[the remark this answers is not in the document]
and-rewsmith: I think that we need it the way it is now because of the case on line 133

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -6,24 +6,50 @@ use std::{
...
+    command::{CommandHandler, CommandHandlerError, ShutdownHandle as CommandShutdownHandle},
     connection::MakeEdgeHubPacketProcessor,
     settings::Settings,
 };
 
+pub struct SidecarShutdownHandle(Sender<()>);
+
+impl SidecarShutdownHandle {
+    pub fn shutdown(self) -> Result<(), SidecarError> {
+        match self.0.send(()) {
```
dmolokanov, line 38: `self.0.send(()).map_err(SidecarError::SidecarShutdown)`
?

PR #3395, opened 2020-08-12
Mqtt Bridge Connector tests

vipeller:


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -387,7 +387,8 @@ static async Task<bool> TryConnectAsync(MqttClient client, IReadOnlyCollection<I
 
         static async Task SubscribeAsync(MqttClient client, IReadOnlyCollection<IMessageConsumer> subscribers)
         {
-            var expectedAckCount = new AtomicLong(subscribers.Count);
+            var subscribersWithSubscriptions = subscribers.Where(s => s.Subscriptions != null && s.Subscriptions.Count > 0).ToArray();
```
varunpuranik, line 390: Can Subscriptions be null for IMessageConsumer?

varunpuranik: If so, can we make it an Option.

vipeller: No, I ran into a null because of the mocked objects, so then I also added the check to make it more robust

PR #3397, opened 2020-08-12
[mqtt] Update mqtt dependecies and fix tracing error

dmolokanov:
Update mqtt dependecies and fix tracing error


mqtt/mqtt-broker/src/connection/mod.rs
```diff
@@ -158,7 +158,7 @@ where
                     Ok(Some(auth_id)) => Auth::Identity(auth_id),
                     Ok(None) => Auth::Unknown,
                     Err(e) => {
-                        warn!(message = "error authenticating client", error =% *e);
+                        warn!(message = "error authenticating client", error =% &*e);
```
and-rewsmith, line 161: So they changed the behavior in a minor version update?

dmolokanov: They did some refactoring in the internals, but it affected this particular case. When I make a smaller example I faced compiler error even without tracing. I wonder how it even worked before :)

PR #3400, opened 2020-08-12
[mqtt] Fix flaky tests when re-write env

dmolokanov:
There are 3 flaky tests in the `mqtt-edgehub/src/settings.rs` file. All 3 tests change ENV and rely on some values set as ENV. The problem occurs when those tests run in parallel. Because they change some behavior of the code via ENV, those tests are highly unreliable.

1. All tests which touch ENV for settings are run sequentially within one group. The rest of the tests runs the same way as before.
2. When the test with ENV changes is finished, previous value is reset


mqtt/mqtt-broker-tests-util/src/packet.rs
```diff
@@ -0,0 +1,109 @@
+use std::{
```
vadim-kovalyov, line 1: nit: let's name this module "packet_stream". I think just "packet" can be confusing.

dmolokanov:
> LGTM, but I'm curious if 2. from the description is necessary. What are you trying to fix with restoring values back to old ones?

If we don't unset some ENV changed in `test1`, they still can affect `test2`.
test1: assume TCP and TLS enabled by default. set TCP_DISABLED=true,  TCP is disabled, TLS is enabled
test2: assume TCP and TLS enabled by default. set TLS_DISABLED=false, both TCP and TLS still enabled.

As always, it is a good practice to clean up changes made in the test, right?

PR #3402, opened 2020-08-13
Clarify the comment in agent Dockerfile.

darobs:
Change the comment in Dockerfile to give the technical reason for the addition.


edge-agent/docker/linux/amd64/Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG EXE_DIR=.
 RUN apk update && \
     apk add --no-cache snappy
 
-# create a user to allow agent to optionally run as non-root
+# Create a user to allow installing the trustbundle under a user context
```
varunpuranik, line 12: What do you mean by installing trust bundle under a user context? Where does the trustbundle get installed?

darobs: We set the X509store location to `CurrentUser` - which creates the store in the current user's directory. If agent is running as root, it uses `/root` as the directory, this creates a user for installing. For Edge on K8s, we can run Agent as non-root, so we create this user for a place to put the X509 store.

edge-agent/docker/linux/arm64v8/base/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apt-get update && \
     ln -s /lib/aarch64-linux-gnu/libdl.so.2 /usr/lib/aarch64-linux-gnu/libdl.so && \
     ln -s /lib/aarch64-linux-gnu/libc.so.6 /usr/lib/aarch64-linux-gnu/libc.so.6
 
-# create a user to allow agent to optionally run as non-root
+# Create a user to allow installing the trustbundle under a user context
```
CindyXing, line 16: This was needed for K8s. I thought we planned to clarify it.

CindyXing: Given we'll be using the non-root user for single device, this comment without k8s specified is ok now.

darobs:
Since we've committed to allow agent to run as non-root in 1.0.10, this comment update is not necessary,

PR #3404, opened 2020-08-13
Fix edgeHub restart by ignoring version for system modules

yophilav:
Symptoms:
edgeHub is always restarted by edgeAgent after 20s of deployment if `version` field is defined a non-empty in a deployment file.

Root-Cause:
edgeAgent includes `version` check as part of module-update-check, in which, edgeAgent goes on and never persist the parameter in the deployment. This causes the `current` and `desire` deployment to always be difference; hence, when the reconciliation process happens, the edgeHub is always marked as updated and get restarted.

Fix: 
This PR fixes the issue by excluding the `version` checking for edgeHub (edgeAgent has been ignoring it).


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerModule.cs
```diff
@@ -109,6 +109,11 @@ public virtual bool IsOnlyModuleStatusChanged(IModule other)
                 EnvDictionaryComparer.Equals(this.Env, other.Env);
```
philipktlin, line 109: Call new method "IsEnvDictionaryEqual" here.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeHubDockerModule.cs
```diff
@@ -36,5 +36,28 @@ public class EdgeHubDockerModule : DockerModule, IEdgeHubModule
...
+            }
+
+            // EdgeHub Equal() is similar to base class except, we ignore the version since we don't ever use it.
+            return string.Equals(this.Name, other.Name) &&
+                string.Equals(this.Type, other.Type) &&
+                this.DesiredStatus == other.DesiredStatus &&
+                this.Config.Equals(other.Config) &&
+                this.RestartPolicy == other.RestartPolicy &&
+                this.ImagePullPolicy == other.ImagePullPolicy &&
+                this.StartupOrder == other.StartupOrder &&
```
philipktlin, line 59: where is the priority?

philipktlin:
And I found that IsOnlyModuleStatusChanged method in DockerModule; should it be overrided for EdgeAgentDockerModule and EdgeHubDockerModule?  Please trace the usage of IsOnlyModuleStatusChanged method.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeHubDockerModule.cs
```diff
@@ -36,5 +36,28 @@ public class EdgeHubDockerModule : DockerModule, IEdgeHubModule
             DefaultValueHandling = DefaultValueHandling.Populate)]
         [DefaultValue(RestartPolicy.Always)]
         public override RestartPolicy RestartPolicy { get; }
+
+        public override bool Equals(IModule<DockerConfig> other)
```
philipktlin, line 40: To share Equals logic across base (DockerModule) and derived classes (EdgeAgentDockerModule, EdgeHubDockerModule), you may need to extract some basic fields in another method e.g. BasicEquals or CoreEquals.

varunpuranik: Why doens't base Equals work here? It doesnt' compare versions either.

yophilav: @philipktlin 
The edgeAgent already has its own equal implementation that has been ignoring the versioning. 

@varunpuranik 
The base's Equal() check the `this.Version` as part of the equal. We don't need the version checking here for edgeHub.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerModule.cs
yophilav: Done

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/EdgeHubDockerModule.cs
```diff
@@ -36,5 +36,28 @@ public class EdgeHubDockerModule : DockerModule, IEdgeHubModule
...
+                return false;
+            }
+
+            if (ReferenceEquals(this, other))
+            {
+                return true;
+            }
+
+            // EdgeHub Equal() is similar to base class except, we ignore the version since we don't ever use it.
+            return string.Equals(this.Name, other.Name) &&
```
philipktlin, line 53: extract these lines (common fields) as a protected method e.g. "BaseEquals" and can be used in both EdgeHubDockerModule and DockerModule.  Nit: also suggest to rename "IsEnvDictionaryEqual" to "AreEnvironmentVariablesEqual".

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerModule.cs
```diff
@@ -84,32 +84,31 @@ public virtual bool Equals(IModule<DockerConfig> other)
...
         public virtual bool IsOnlyModuleStatusChanged(IModule other)
         {
             return other is DockerModule dockerModule &&
-                string.Equals(this.Name, other.Name) &&
+                this.BaseEquals(other) &&
                 string.Equals(this.Version, other.Version) &&
+                this.Config.Equals(dockerModule.Config);
+        }
+
+        protected bool BaseEquals(IModule other)
```
varunpuranik, line 100: This seems a bit odd. Why do we need this method?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerModule.cs
```diff
@@ -84,25 +84,24 @@ public virtual bool Equals(IModule<DockerConfig> other)
...
+        {
+            return string.Equals(this.Name, other.Name) &&
                 string.Equals(this.Type, other.Type) &&
                 this.DesiredStatus != other.DesiredStatus &&
```
philipktlin, line 104: DesiredStatus should not be in BaseEquals method.  It is different between Equals and IsOnlyModuleStatusChanged methods.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerModule.cs
philipktlin: @varunpuranik this can be reused in EdgeHubDockerModule.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerModule.cs
```diff
@@ -84,25 +84,24 @@ public virtual bool Equals(IModule<DockerConfig> other)
...
+                this.Config.Equals(other.Config);
         }
 
         public virtual bool IsOnlyModuleStatusChanged(IModule other)
         {
             return other is DockerModule dockerModule &&
-                string.Equals(this.Name, other.Name) &&
+                this.BaseEquals(other) &&
                 string.Equals(this.Version, other.Version) &&
+                this.Config.Equals(dockerModule.Config);
```
philipktlin, line 97: Why this.Config is not in BaseEquals method?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerModule.cs
philipktlin: I think the same approach (extracting base hash code method) should use for GetHashCode as well.

philipktlin: We should refactor and extract abstract base class for EdgeAgentDockerModule, EdgeHubDockerModule and DockerModule.  The base class should implements Equals (equivalent to what BaseEquals method) and GetHashCode method.

yophilav:
Per discussion with Varun, one would prefer to have the edgeHub & other modules to handle the `version` in the same manner.

yophilav:
Per discussion with Varun, one would prefer to have the edgeHub & other modules to handle the `version` in the same manner. Hence, we are using https://github.com/Azure/iotedge/pull/3405 as a solution

PR #3406, opened 2020-08-13
Broker refactoring to allow for sidecar 

and-rewsmith:
This is stacked on top of #3393


vadim-kovalyov:
Do we still need this?

PR #3407, opened 2020-08-14
Make broker persistence configurable

and-rewsmith:


mqtt/mqtt-generic/src/settings.rs
```diff
@@ -204,13 +208,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn it_refuses_persistence_with_no_file_path() {
-        let settings = Settings::from_file(Path::new("test/config_no_file_path.json"));
-
-        assert_matches!(settings, Err(_));
-    }
-
```
and-rewsmith, line 213: Now that we have defaults this test doesn't make sense

mqtt/mqtt-broker/src/settings/mod.rs
```diff
@@ -145,13 +145,35 @@ impl Default for RetainedMessagesConfig {
...
+            file_path,
+            time_interval,
+        }
+    }
+
+    pub fn file_path(&self) -> String {
+        self.file_path.clone()
+    }
+
+    pub fn time_interval(&self) -> Duration {
```
varunpuranik, line 168: What does this time_interval signify?

mqtt/mqtt-broker/src/settings/mod.rs
```diff
@@ -145,13 +145,35 @@ impl Default for RetainedMessagesConfig {
...
-// TODO: apply settings
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 pub struct SessionPersistenceConfig {
     file_path: String,
     #[serde(with = "humantime_serde")]
     time_interval: Duration,
-    unsaved_message_count: u32,
+}
+
+// TODO: &self?
```
and-rewsmith, line 155: remove

mqtt/mqtt-broker/src/settings/mod.rs
and-rewsmith: Interval at which the persisting happens. I can rename to make it more clear.

varunpuranik: So the broker will now persist the session on shutdown and periodically at an interval specified by this argument? 
And previously it was doing so every 5 mins?

and-rewsmith: The broker was persisting on shutdown before, the only thing that changes here is the path and the interval specified by the arg. Previously it was every 5 mins, yes.

mqtt/mqtt-broker/src/settings/mod.rs
```diff
@@ -145,13 +145,34 @@ impl Default for RetainedMessagesConfig {
     }
 }
 
-// TODO: apply settings
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 pub struct SessionPersistenceConfig {
     file_path: String,
```
dmolokanov, line 150: lets make it `PathBuf` instead of `String`

dmolokanov, line 150:
```suggestion
    dir_path: String,
```
or
```suggestion
    path: String,
```
?

mqtt/mqtt-broker/src/settings/mod.rs
```diff
@@ -145,13 +145,34 @@ impl Default for RetainedMessagesConfig {
...
+    pub fn file_path(&self) -> String {
+        self.file_path.clone()
+    }
+
+    pub fn time_interval(&self) -> Duration {
+        self.time_interval
+    }
+}
+
+impl Default for SessionPersistenceConfig {
```
dmolokanov, line 172: When doing so we make defaults scattered across multiple places. Let's leave it in default.json only? If you see more defaults for the settings, please remove it as well.

mqtt/mqtt-generic/src/settings.rs
dmolokanov: Agree. I believe we don't need `test/config_no_file_path.json` file anymore

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -2,18 +2,19 @@ mod bootstrap;
 mod shutdown;
 mod snapshot;
 
-use std::{env, path::Path};
+use std::path::Path;
 
 use anyhow::{Context, Result};
 use futures_util::pin_mut;
+use std::fs;
```
dmolokanov, line 9: Could you please merge import with std:: above?

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -52,9 +54,22 @@ where
     Ok(())
 }
 
+async fn init_broker_state(
```
dmolokanov, line 57: This method added more code than it was before. Do we really need it at all?

mqtt/mqtt-broker/src/settings/mod.rs
and-rewsmith: Aren't these defaults needed for this?
https://github.com/Azure/iotedge/blob/iiot/mqtt/mqtt-broker/src/broker.rs#L944

I think we need to have BrokerConfig default to reading the appropriate config.

mqtt/mqtt-broker/src/settings/mod.rs
vadim-kovalyov: Also, at some point in time we need to talk about our persistence strategy for edge hub. Right now I'm not sure that 5m interval is useful strategy, and I would rather disable it if possible. I would rather let the broker loose the state on crash/hard shutdown, then risk the inconsistent session state.

and-rewsmith: I'm confused. Are you disagreeing with the 5m interval only? If we disable this then the broker still maintains state on shutdown right?

varunpuranik: Yes, I agree with Vadim. Making it configurable is okay for now, but we need to revisit it before we release.

mqtt/mqtt-edgehub/src/settings.rs
```diff
@@ -266,11 +269,13 @@ mod tests {
     where
         F: FnOnce() -> Result<Settings, ConfigError>,
     {
-        let _tcp_address = env::set_var("LISTENER__TCP__ADDRESS", "10.0.0.1:1883");
-        let _tls_address = env::set_var("LISTENER__TLS__ADDRESS", "10.0.0.1:8883");
-        let _tls_certificate = env::set_var("LISTENER__TLS__CERTIFICATE", "/tmp/edgehub/cert.pem");
-        let _tls_private_key = env::set_var("LISTENER__TLS__PRIVATE_KEY", "/tmp/edgehub/pkey.pem");
-        let _auth_base_url = env::set_var("AUTH__BASE_URL", "/auth/");
+        env::set_var("LISTENER__TCP__ENABLED", "true");
+        env::set_var("LISTENER__TLS__ENABLED", "true");
```
and-rewsmith, line 273: This is a fix for the flaky tests

mqtt/mqtt-edgehub/src/settings.rs
```diff
@@ -266,11 +269,13 @@ mod tests {
...
+        env::set_var("LISTENER__TCP__ADDRESS", "10.0.0.1:1883");
+        env::set_var("LISTENER__TLS__ADDRESS", "10.0.0.1:8883");
+        env::set_var("LISTENER__TLS__CERTIFICATE", "/tmp/edgehub/cert.pem");
+        env::set_var("LISTENER__TLS__PRIVATE_KEY", "/tmp/edgehub/pkey.pem");
+        env::set_var("AUTH__BASE_URL", "/auth/");
```
and-rewsmith, line 278: Removing the let declarations is a fix for clippy

PR #3426, opened 2020-08-19
fix for certificate decoding on Linux

vipeller:
On Linux the X509Certificate2 class is not as flexible as on Windows and cannot figure out the certificate encoding (e.g. PEM, base64, DER). The broker sends messages PEM encoded, so modified the code to trim the PEM header and do the base64 decoding manually, giving a DER encoded certificate to X509Certificate2


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentController.cs
```diff
@@ -170,7 +173,7 @@ static X509Certificate2 DecodeCertificate(string encodedCertificate)
         {
             try
             {
-                var certificateContent = Encoding.UTF8.GetBytes(encodedCertificate);
+                var certificateContent = Convert.FromBase64String(TrimPemFrame(encodedCertificate));
```
varunpuranik, line 176: Can you pls add a comment on why you need to do this?

vipeller:
turned out this is not necessary

PR #3432, opened 2020-08-19
Check DPS provisioning gets hubname

lfitchett:


edgelet/iotedge/src/check/mod.rs
```diff
@@ -856,18 +861,56 @@ mod tests {
     #[test]
     fn settings_connection_string_dps_err() {
```
arsing, line 862: It's not an `_err` test any more. Maybe add a new `_err` test for when the backup isn't there?

edgelet/iotedge/src/check/checks/well_formed_connection_string.rs
```diff
@@ -49,10 +49,18 @@ impl WellFormedConnectionString {
...
-                            To run connectivity checks in this configuration please specify the backing IoT Hub name using --iothub-hostname switch if you have that information.\n\
-                            If no hostname is provided, all hub connectivity tests will be skipped.";
-            return Ok(CheckResult::Warning(Context::new(warning).into()));
+            let provisioning_file_path = settings
+                .homedir()
+                .join("cache")
+                .join("provisioning_backup.json");
+
+            let provision_result = provisioning::restore(provisioning_file_path).map_err(|e| {
+                let reason = format!("Couldn't restore provisioning file: {}", e);
```
arsing, line 58: Might be useful to include the message of the original warning still, but change it to also mention that we tried to use the provisioning backup but that also failed. You are still honoring the `--iothub-hostname` parameter after all, so you might as well still let people know it exists in the message. But do keep it as an error instead of a warning.

Check with Venkat for the exact verbiage.

lfitchett: ok, applied venkat's feedback

edgelet/iotedge/src/check/checks/well_formed_connection_string.rs
```diff
@@ -49,10 +49,21 @@ impl WellFormedConnectionString {
...
+            let provisioning_file_path = settings
+                .homedir()
+                .join("cache")
+                .join("provisioning_backup.json");
+
+            let provision_result = provisioning::restore(provisioning_file_path).map_err(|e| {
+                let reason = format!("Could not retrieve iothub_hostname from provisioning file.\n\
+                Please specify the backing IoT Hub name using --iothub-hostname switch if you have that information.\n\
+                If no hostname is provided, all hub connectivity tests will be skipped.\n\
+                Provisioning file error: {}", e);
```
arsing, line 61: Remove this last line. You're already using this message as a context of the original error, so it'll already get printed. There's no reason to print it twice.

edgelet/iotedge/src/check/mod.rs
```diff
@@ -854,20 +859,57 @@ mod tests {
...
+        std::fs::create_dir(tmp_dir.path().join("cache")).unwrap();
+
+        // replace homedir with temp directory
+        {
+            let mut new_config = File::create(&config_file).unwrap();
+            for line in BufReader::new(File::open(config_file_source).unwrap()).lines() {
+                if let Ok(line) = line {
+                    if line.contains("homedir") {
+                        let new_line = format!(
+                            "homedir: \"{}\"",
```
arsing, line 889: `r#"homedir: "{}""#,`

edgelet/iotedge/src/check/mod.rs
```diff
@@ -887,7 +929,50 @@ mod tests {
         }
 
         match WellFormedConnectionString::default().execute(&mut check, &mut runtime) {
-            CheckResult::Warning(err) => assert!(err.to_string().contains("Device not configured with manual provisioning, in this configuration 'iotedge check' is not able to discover the device's backing IoT Hub.")),
+            CheckResult::Ok => {
+                assert_eq!(check.iothub_hostname, Some(hub_name.to_owned()));
+            }
+            check_result => panic!("parsing {} returned {:?}", filename, check_result),
```
arsing, line 935: You may want to change this message to something else, so that it's clear from the panic log whether it was this check or the previous one that failed.

CindyXing (approved):

arsing (dismissed):

arsing (approved):

PR #3434, opened 2020-08-20
Add ServiceIdentitiesUpdated event to DeviceScopeIdentitiesCache

dylanbronson:
Adding ServiceIdentitiesUpdated event to DeviceScopeIdentitiesCache. This event will be used by MessageProducer to get all identities from the current nested edge hierarchy to pass to the mqtt broker for IoTHub topic authorization.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
@@ -318,8 +318,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
                 NotInScope,
                 AddInScope,
                 RefreshingServiceIdentity,
-                GettingServiceIdentity,
-                GettingAllIds
+                GettingServiceIdentity
             }
 
             public static void Created() =>
@@ -371,11 +370,6 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
 
             internal static void InitializingRefreshTask(TimeSpan refreshRate) =>
                 Log.LogDebug((int)EventIds.InitializingRefreshTask, $"Initializing device scope identities cache refresh task to run every {refreshRate.TotalMinutes} minutes.");
-
-            internal static void GettingAllIds()
-            {
-                Log.LogDebug((int)EventIds.GettingAllIds, "Getting all IDs within hierarchy.");
-            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -75,7 +77,7 @@ public void InitiateCacheRefresh()
 
         public Task WaitForCacheRefresh(TimeSpan timeout) => this.refreshCacheCompleteSignal.WaitAsync(timeout);
 
-        public async Task RefreshServiceIdentity(string id)
+        public async Task RefreshServiceIdentity(string id, bool invokeServiceIdentitiesUpdated = true)
```
dylanbronson, line 80: Using this bool so that we don't have to update every time if we're calling it from RefreshServiceIdentities, but so that we still will invoke the event if we're not calling from that method. I'm also open to changing this to be a separate method that is called.

richma-ms: when would we want this event to not fire?  Seems like having such a mechanism would lead to potential inconsistencies down the line.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -51,6 +51,8 @@ public sealed class DeviceScopeIdentitiesCache : IDeviceScopeIdentitiesCache
 
         public event EventHandler<ServiceIdentity> ServiceIdentityUpdated;
 
+        public event EventHandler<IList<string>> ServiceIdentitiesUpdated;
```
richma-ms, line 54: maybe augment one of the existing tests for this class to check this callback?  I think there's at least one test that validates the other two events already.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
dylanbronson: RefreshServiceIdentities calls this method within a loop of all the identities. And we fire the event after RefreshServiceIdentities completes. 

It seems wasteful to fire the event every time RefreshServiceIdentity is called within RefreshServiceIdentities

vadim-kovalyov: I would rather prefer not to expose such flag to public if it is only internal details. We can create a private method that both `RefreshServiceIdentity` and `RefreshServiceIdentities` can call.

PR #3435, opened 2020-08-20
Add `always_reprovision_on_startup` setting to DPS  provisioning configuration

arsing:
Before this change, when using any of the DPS provisioning methods,
iotedged would always reprovision with DPS on startup. It would maintain
a provisioning backup, but this backup was only used if the provisioning
with DPS failed for any reason.

If the new setting in this change is set to false, iotedged will first attempt
to restore the backup, and only reprovision with DPS if that fails.

The primary reason this was done was due to a quirk of DPS-TPM provisioning.
`iotedged`, or rather `utpm` as used by `libiothsm-std`, reserves a single slot
on the TPM to store keys. However DPS-TPM provisioning involves two keys that
must be loaded in the TPM - the initial challenge key and the final
device identity key. Thus the DPS-TPM provisioning method must store
the challenge key in the same slot that occupied
the previous device identity key. However, if the second part of
DPS-TPM provisioning fails and `iotedged` never receives
the new device identity key, the previous device identity key is
irrevocably lost even if the rest of the information can be restored from
the provisioning backup.

This creates a problem when modules request their SAS tokens to be signed
so they can talk to Edge Hub, `iotedged` will derive their module identity keys
using the device identity key that is actually the challenge key.
However Edge Hub will validate the SAS tokens against the module identity keys
that it received from IoT Hub, which were derived using the previous
device identity key. So Edge Hub will reject the modules' SAS tokens.

With this change, and if the setting has been set to false, `iotedged` won't
attempt the DPS-TPM provisioning at all, so the "previous" device identity key
in the TPM will remain.

One last thing to be handled carefully is the reprovisioning trigger.
If the device identity was deleted from IoT Hub, then before this change
the reprovision on startup would've caught it. With this change, the only way
it can be detected is when Edge Agent fails to connect to IoT Hub and thus
triggers a reprovision using the management API. However triggering the
reprovision using the management API is itself opt-in using
the `dynamic_reprovisioning` config flag. Since it doesn't make sense to have
a device that doesn't reprovision with the cloud but also doesn't reprovision
itself if it needs to, this change makes it so that setting
`always_reprovision_on_startup` to false implicitly sets
`dynamic_reprovisioning` to true.

Note that the `always_reprovision_on_startup` seting defaults to true
to retain the original behavior of always reprovisioning on startup,
for backward-compatiblity. However it is not expected that there would be
any problems from setting it to false.

Also note that this change makes `BackupProvisioning` generic over
the underlying provisioner (as it was always intended to be) rather than being
an internal detail of the three DPS provisioning methods. This means it could
be used for other provisioning methods like external in the future.
In this case, the `always_reprovision_on_startup` config would be moved
from being DPS-specific to applying to all provisioning methods. Luckily,
this can be done without any backward-incompatible changes to the config,
because the provisioning method parameters are already flattened, so that
the `always_reprovision_on_startup` flag is already at the top-level of
the provisioning settings.


edgelet/edgelet-core/src/settings.rs
```diff
@@ -411,7 +418,15 @@ impl Provisioning {
     }
 
     pub fn dynamic_reprovisioning(&self) -> bool {
-        self.dynamic_reprovisioning
+        // If using DPS provisioning and its always_reprovision_on_startup is false,
```
darobs, line 421: Should there be a test for this?

darobs (dismissed):
I don't know if that additional test is required. Otherwise, it looks good.

edgelet/iotedged/src/lib.rs
```diff
@@ -1533,25 +1529,22 @@ where
...
-                }
-                Err(Either::A((err, _y))) => Err(err).into_future(),
-                Err(Either::B((err, _y))) => {
-                    debug!("The mgmt shutdown and reprovision signal failed.");
-                    Err(err.unwrap()).into_future()
-                }
+        |res| match res {
+            Ok(Either::A((_edge_rt_ok, _mgmt_stop_and_reprovision_signaled_future))) => {
+                info!("Edge runtime will stop because of the shutdown signal.");
+                future::ok((StartApiReturnStatus::Shutdown, false))
```
massand, line 1535: The return value for this gigantic function seem like they could be represented as a tri-state enum than a (enum, bool) combination, where the bool really matters to just one enum.

edgelet/provisioning/src/provisioning.rs
```diff
@@ -853,7 +866,16 @@ where
     }
 
     fn reprovision(&self) -> Box<dyn Future<Item = (), Error = Error> + Send> {
-        panic!("A reprovisioning operation is not expected for `BackupProvisioning`")
+        // We've been told to reprovision, so whatever provisioning info we have in the backup should be discarded
```
CindyXing, line 869: To be conservative, how about we still keep the backup? 

One possibility is that customers switch from true to false. Or some offline scenario that we want things continue working?

CindyXing (approved):

ancaantochi (approved):

edgelet/edgelet-core/src/settings.rs
arsing: Talked it over with Anca and we decided it's better to not override it at all, just in case there's a bug that triggers spurious reprovisioning.

edgelet/provisioning/src/provisioning.rs
arsing: As discussed in yesterday's scrum, deleting the backup is the signal that we need to reprovision on next startup.

edgelet/edgelet-core/src/settings.rs
```diff
@@ -411,7 +418,15 @@ impl Provisioning {
...
+        // then dynamic reprovisioning is the only way we can know when the device identity has been deleted in IoT Hub.
+        //
+        // So force dynamic_reprovisioning to be true even if the user didn't set it.
+
+        match &self.provisioning {
```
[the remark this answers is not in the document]
arsing: As we discussed, I've removed the override entirely so that the user can keep it turned off in case there's a bug that triggers spurious reprovisioning.

edgelet/contrib/config/linux/config.yaml
```diff
@@ -73,6 +73,10 @@
 #                       provision with DPS, and only fall back to the provisioning backup
 #                       from a previous successful provisioning if the attempt fails.
 #
+#                       If setting this to false, it's recommended to also set
```
CindyXing, line 76: Do we have a doc bug around  always_provision_on_startup and the recommendation around dynamic provisioning? 
If not, please open one and we should work with Kelly to update together with the release.

CindyXing (approved):

edgelet/contrib/config/linux/config.yaml
arsing: I'll check with Varun and Anca. I'm not sure if we have docs for `dynamic_reprovisioning` either.

PR #3436, opened 2020-08-20
Identities refresh negative caching

richma-ms:
This change adds a limit to how often we can trigger refresh in DeviceScopeIdentitiesCache, both for individual identities refresh and whole-cache refresh.  This is so that we don't end up hammering IoTHub now that we expose a REST endpoint for anyone to call into and trigger refreshes.

The default delay is set to 2min.

Now that we have this protection, I've also updated GetDeviceAndModuleOnBehalfOf to always call upstream first, so that we can always sync down the latest when someone wants to fetch a single identity.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
@@ -2,15 +2,21 @@
 namespace Microsoft.Azure.Devices.Edge.Hub.Core
 {
     using System;
-    using System.Collections.Generic;
-    using System.Linq;
+    using Microsoft.Azure.Devices.Edge.Util;
 
     public static class AuthChainHelpers
     {
-        public static string[] GetAuthChainIds(string authChain) => authChain.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+        public static string[] GetAuthChainIds(string authChain)
+        {
+            Preconditions.CheckNonWhiteSpace(authChain, nameof(authChain));
+            return authChain.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+        }
 
         public static bool ValidateAuthChain(string actorDeviceId, string targetId, string authChain)
         {
+            Preconditions.CheckNonWhiteSpace(actorDeviceId, nameof(actorDeviceId));
+            Preconditions.CheckNonWhiteSpace(targetId, nameof(targetId));
+            Preconditions.CheckNonWhiteSpace(actorDeviceId, nameof(actorDeviceId));
             string[] authChainIds = GetAuthChainIds(authChain);
 
             // Should have at least 1 element in the chain
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
@@ -2,12 +2,16 @@
 namespace Microsoft.Azure.Devices.Edge.Hub.Core
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
 
     public static class AuthChainHelpers
     {
+        public static string[] GetAuthChainIds(string authChain) => authChain.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+
         public static bool ValidateAuthChain(string actorDeviceId, string targetId, string authChain)
         {
-            var authChainIds = authChain.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+            string[] authChainIds = GetAuthChainIds(authChain);
 
             // Should have at least 1 element in the chain
             if (authChainIds.Length < 1)
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
@@ -6,6 +6,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
+    using Antlr4.Runtime.Sharpen;
     using Microsoft.Azure.Devices.Edge.Hub.Core.Identity.Service;
     using Microsoft.Azure.Devices.Edge.Storage;
     using Microsoft.Azure.Devices.Edge.Util;
@@ -22,8 +23,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
         readonly Timer refreshCacheTimer;
         readonly TimeSpan periodicRefreshRate;
         readonly TimeSpan refreshDelay;
-        readonly AsyncAutoResetEvent refreshCacheSignal = new AsyncAutoResetEvent();
-        readonly AsyncManualResetEvent refreshCacheCompleteSignal = new AsyncManualResetEvent();
+        readonly AsyncManualResetEvent refreshCacheSignal = new AsyncManualResetEvent(false);
+        readonly AsyncManualResetEvent refreshCacheCompleteSignal = new AsyncManualResetEvent(false);
         readonly object refreshCacheLock = new object();
 
         Task refreshCacheTask;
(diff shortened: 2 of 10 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -20,25 +20,32 @@ public sealed class DeviceScopeIdentitiesCache : IDeviceScopeIdentitiesCache
...
+            TimeSpan refreshDelay)
         {
             this.serviceIdentityHierarchy = serviceIdentityHierarchy;
             this.serviceProxy = serviceProxy;
             this.encryptedStore = encryptedStorage;
-            this.refreshRate = refreshRate;
-            this.refreshCacheTimer = new Timer(this.RefreshCache, null, TimeSpan.Zero, refreshRate);
+            this.periodicRefreshRate = periodicRefreshRate;
+            this.refreshDelay = refreshDelay;
+            this.refreshCacheTimer = new Timer(this.RefreshCache, null, TimeSpan.Zero, this.periodicRefreshRate);
```
philipktlin, line 46: nit: you may use PeriodicTask instead of Timer.

richma-ms: this is how it was before so I'll leave it.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/appsettings_hub.json
```diff
@@ -82,6 +82,7 @@
   },
   "AuthenticationMode": "CloudAndScope",
   "DeviceScopeCacheRefreshRateSecs": 3600,
+  "DeviceScopeCacheRefreshDelaySecs": 120,
```
philipktlin, line 85: nit: should we start to do it right to use "00:00:00" format?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -57,22 +64,38 @@ public sealed class DeviceScopeIdentitiesCache : IDeviceScopeIdentitiesCache
...
+                this.refreshCacheCompleteSignal.Reset();
+                this.refreshCacheSignal.Set();
+
+                // Update the cache refresh timestamp
+                this.cacheLastRefreshTime = DateTime.UtcNow;
+            }
+            else
+            {
+                Events.SkipRefreshCache(this.cacheLastRefreshTime, this.refreshDelay);
+                this.refreshCacheCompleteSignal.Set();
```
philipktlin, line 97: Current we return refreshCacheCompleteSignal.Set() immediately, this may cause problems.  For example, there are 2 requests to call to initiateCacheRefresh.  The 1st request successfully goes into line 88 and set refresthCacheSignal, then RefreshCache method will call serviceProxy to get identities.  Then 2nd request comes in and since it is still within refresh delay (2 mins), it will call refreshCacheCompleteSignal.Set() in line97 which will release both 1st and 2nd request WaitForCacheRefresh calls.  I think we want both 1st and 2nd requests should wait for 1st request identities result comes back, right?

should we just remove line 97 since this.refreshCacheCompleteSignal.Set() should be called once 1st request is finished?

philipktlin: Do you have test to verify this case?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -97,11 +120,23 @@ async Task RefreshServiceIdentityInternal(string id, bool invokeServiceIdentitie
...
+                    Events.RefreshingServiceIdentity(id);
+                    Option<ServiceIdentity> serviceIdentity = await this.GetServiceIdentityFromService(id);
+                    await serviceIdentity
+                        .Map(s => this.HandleNewServiceIdentity(s))
+                        .GetOrElse(() => this.HandleNoServiceIdentity(id));
+
+                    // Update the timestamp for this identity so we
+                    // don't repeatedly refresh the same identity
+                    // in rapid succession.
+                    this.identitiesLastRefreshTime[id] = DateTime.UtcNow;
```
philipktlin, line 134: Do we need to update this.identitiesLastRefreshTime in RefreshCache method?

philipktlin:
Would you tell me where we add delegate function to ServiceIdentitiesUpdated event handler?  thanks.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -159,13 +204,31 @@ internal Task<Option<ServiceIdentity>> GetServiceIdentityFromService(string id)
             return cache;
         }
 
+        async Task<bool> ShouldRefreshIdentity(string id)
+        {
+            bool hasRefreshed = this.identitiesLastRefreshTime.TryGetValue(id, out DateTime lastRefreshTime);
```
philipktlin, line 209: Do we also need to consider this.cacheLastRefreshTime?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -57,22 +67,45 @@ public sealed class DeviceScopeIdentitiesCache : IDeviceScopeIdentitiesCache
...
+            {
+                Events.SkipRefreshCache(this.cacheLastRefreshTime, this.refreshDelay);
+
+                if (!this.refreshCacheSignal.IsSet)
+                {
+                    // The background thread for refresh the cache is idle,
+                    // in this case we need to signal completion right away
+                    // or anyone calling WaitForCacheRefresh() will be stuck
+                    this.refreshCacheCompleteSignal.Set();
+                }
```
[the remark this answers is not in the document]
philipktlin: First I would like to confirm the requirement.  Do we want to control the identities refresh not run too often (refreshDelay).  If so, current implementation doesn't fulfill it.

Suggestion: define a TaskCompletionSource class field to store for the latest refresh.  Use 1 AutoResetEvent to limit single thread to update it.  Any time TaskCompletionSource should be able to return the latest refresh task.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
```diff
@@ -2,12 +2,16 @@
 namespace Microsoft.Azure.Devices.Edge.Hub.Core
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
 
     public static class AuthChainHelpers
     {
+        public static string[] GetAuthChainIds(string authChain) => authChain.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
```
varunpuranik, line 10: Nit: Null check?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -57,22 +67,45 @@ public sealed class DeviceScopeIdentitiesCache : IDeviceScopeIdentitiesCache
...
+            {
+                this.refreshCacheCompleteSignal.Reset();
+                this.refreshCacheSignal.Set();
+
+                // Update the cache refresh timestamp
+                this.cacheLastRefreshTime = DateTime.UtcNow;
+            }
+            else
+            {
+                Events.SkipRefreshCache(this.cacheLastRefreshTime, this.refreshDelay);
```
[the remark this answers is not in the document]
richma-ms: don't all log entries already contain datetime stamp?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
richma-ms: no, we shouldn't wait even in that case.  On-demand refresh should always be treated with a higher priority that cache refresh, because it's invoked when a new device connects, but we couldn't find the device's identity in the cache.  We can't predict how long into the delay timer we'll be if we start conflating the two scenarios, what if it's 1m59s since we last refreshed the cache?  What if the user changed the delay so that it's significantly longer?

IMO it's better to always keep individual vs cache refresh separate, both in terms of predictability and simplicity.

PR #3440, opened 2020-08-22
Update mqtt bridge to use common structs for upstream and remotes

ancaantochi:
Settings updates to use a single struct for connection and merge env variables for upstream bridge with file settings
Changed Bridge to be used for both nested and remote.


mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,27 +1,28 @@
...
+impl Bridge {
+    pub fn new(connection_settings: ConnectionSettings) -> Self {
+        Bridge {
+            connection_settings,
+        }
     }
 
     pub async fn start(&self) {
-        info!("Starting nested bridge...{:?}", self.settings);
+        info!("Starting nested bridge...{:?}", self.connection_settings);
```
varunpuranik, line 18:
```suggestion
        info!("Starting bridge...{:?}", self.connection_settings);
```

mqtt/mqtt-bridge/config/default.json
```diff
@@ -1,4 +1,13 @@
...
+    "upstream":
+    {
+        "keep_alive": "1m",
+        "clean_session": true,
+        "subscriptions": [],
+        "forwards": []
+    },
+    "remotes": [],
+    "messages": {
+        "max_count": 100
```
varunpuranik, line 11: This is max messages in memory? As in read from disk at a time? Why is this needed as a setting? (I see it in the doc, just curious).

dmolokanov: iirc it is maximum inflight messages settings. it should not be that big number. Broker has 16 now

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,27 +1,28 @@
...
+impl Bridge {
+    pub fn new(connection_settings: ConnectionSettings) -> Self {
+        Bridge {
+            connection_settings,
+        }
     }
 
     pub async fn start(&self) {
-        info!("Starting nested bridge...{:?}", self.settings);
+        info!("Starting bridge...{:?}", self.connection_settings);
```
vadim-kovalyov, line 18: `self.connection_settings` may have creds in it, maybe we shouldn't dump it into the logs.

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -43,21 +40,140 @@ impl Settings {
...
+            upstream: UpstreamSettings,
+
+            remotes: Vec<ConnectionSettings>,
+
+            messages: MessagesSettings,
+        }
+
+        let value: Inner = serde::Deserialize::deserialize(deserializer)?;
+
+        let upstream_connection_settings = match value.nested_bridge.gateway_hostname.clone() {
```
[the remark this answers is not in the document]
ancaantochi: As we discussed offline I did the deconstruct, I also had to call clone for gateway_hostname before map because otherwise I would get this error:   
```
81 | /             nested_bridge
82 | |                 .gateway_hostname
   | |_________________________________- value moved here
83 |                   //.clone()
84 |                   .map(|gateway_hostname| ConnectionSettings {
   |                        ^^^^^^^^^^^^^^^^^^ value used here after partial move
...
89 |                       credentials: Credentials::Provider(nested_bridge),
   |                                                          ------------- use occurs due to use in closure
   |
```

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -43,21 +40,146 @@ impl Settings {
...
+            upstream: UpstreamSettings,
+
+            remotes: Vec<ConnectionSettings>,
+
+            messages: MessagesSettings,
+        }
+
+        let Inner {
+            nested_bridge,
+            upstream,
```
and-rewsmith, line 75: Should these be separate if they are getting combined into one settings object?

and-rewsmith: After looking at this more I agree. They are separate at first because it is easier to input, then get combined for ease of use in the application.

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -1,23 +1,20 @@
 #![allow(dead_code)] // TODO remove when ready
```
and-rewsmith, line 1: Do we still need this?

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -170,15 +294,59 @@ mod tests {
...
+        assert_eq!(settings.messages.max_count(), 10);
+    }
+
+    #[test]
+    #[serial(env_settings)]
+    fn from_default_sets_messages_settings() {
+        let settings = Settings::new().unwrap();
+
+        assert_eq!(settings.messages.max_count(), 100);
     }
```
and-rewsmith, line 350: I noticed that we have tests in the broker verifying the defaults. I don't think this is super useful, but others might disagree and recommend adding here.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,27 +1,28 @@
...
+impl Bridge {
+    pub fn new(connection_settings: ConnectionSettings) -> Self {
+        Bridge {
+            connection_settings,
+        }
     }
 
     pub async fn start(&self) {
-        info!("Starting nested bridge...{:?}", self.settings);
+        info!("Starting bridge...{:?}", self.connection_settings.name());
```
and-rewsmith, line 18: I know there are concerns about leaking credentials, but what we can do is have a to_string() method that only returns the non-private fields.

mqtt/mqtt-bridge/src/settings.rs
ancaantochi: yes, if I remove it complains about functions which are not used (like remotes, messages, etc.)

mqtt/mqtt-bridge/src/settings.rs
ancaantochi: I think they are useful to make sure that defaults file is loaded

mqtt/mqtt-bridge/src/settings.rs
ancaantochi: this is used because the settings from env have a different format, but then they are combined in same struct

mqtt/mqtt-bridge/config/default.json
ancaantochi: from the spec this is number of messages per topic bridge stores in memory which I understand as how many are read from disk at a time, not inflight messages. It should be renamed to make it clear what it refers to.

mqtt/mqtt-bridge/src/settings.rs
dmolokanov: Yes, it is a nice idea to check defaults, because of several sources of settings: env, default config file, impl Default, etc

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -15,7 +15,7 @@ impl Bridge {
     }
 
     pub async fn start(&self) {
-        info!("Starting bridge...{:?}", self.connection_settings);
+        info!("Starting bridge...{:?}", self.connection_settings.name());
```
dmolokanov, line 18: nit: Could you please check formatting for all the code you added? It may not be too convenient to dump objects internal structure in the logs.
```suggestion
        info!("Starting bridge...{}", self.connection_settings.name());
```
This will use `Display` implementation to write data instead of `Debug` https://doc.rust-lang.org/std/fmt/

mqtt/mqtt-bridge/config/default.json
ancaantochi: I removed max_count for now since it was not clear what was needed for and will add it as needed.

dmolokanov (approved):
LGTM 🚀

PR #3441, opened 2020-08-24
Add "RocksDB_MaxOpenFiles" to options we can pass to Agent and Hub.

darobs:
Expose the MaxOpenFiles setting in RocksDb to the user via the `RocksDB_MaxOpenFiles` environment variable.

### DIFF
--- a/edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbOptionsProvider.cs
+++ b/edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbOptionsProvider.cs
@@ -44,19 +44,25 @@ namespace Microsoft.Azure.Devices.Edge.Storage.RocksDb
         // Once write-ahead logs exceed this size, we will start forcing the flush of
         // column families whose memtables are backed by the oldest live WAL file
         const ulong DefaultMaxTotalWalSize = 512 * 1024 * 1024;
+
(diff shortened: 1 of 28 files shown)


edge-util/src/Microsoft.Azure.Devices.Edge.Storage.RocksDb/RocksDbOptionsProvider.cs
```diff
@@ -44,19 +44,25 @@ public class RocksDbOptionsProvider : IRocksDbOptionsProvider
         // Once write-ahead logs exceed this size, we will start forcing the flush of
         // column families whose memtables are backed by the oldest live WAL file
         const ulong DefaultMaxTotalWalSize = 512 * 1024 * 1024;
+
+        // Default is to allow RocksDb to keep open as much files as it can, but
+        // to keep RocksDb from over-using system resources, this may be set.
+        const int DefaultMaxOpenFiles = -1;
```
varunpuranik, line 50: Should we change the default to something like 5000, so that customers don't run into it by default?

darobs: Agreed - [I used 5000 because this was the previous default.](https://github.com/facebook/rocksdb/blob/master/DEFAULT_OPTIONS_HISTORY.md#480-522016)

PR #3449, opened 2020-08-24
[mqtt] Cleanup client disconnect and add logging

and-rewsmith:
There are some cleanup improvements we can make to client disconnect. Also I added a few lines of logging.

### DIFF
--- a/mqtt/mqtt-edgehub/src/command.rs
+++ b/mqtt/mqtt-edgehub/src/command.rs
@@ -80,7 +80,7 @@ impl CommandHandler {
         let subscribe_topics = &[TOPIC_FILTER.to_string()];
         if !self.subscribe(subscribe_topics).await? {
             return Err(CommandHandlerError::MissingSubacks(
-                subscribe_topics.to_vec().join(", "),
+                subscribe_topics.join(", "),
             ));
         }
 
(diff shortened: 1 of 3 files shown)


mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -75,11 +75,12 @@ impl CommandHandler {
...
 
     pub async fn run(mut self) -> Result<(), CommandHandlerError> {
-        let subscribe_topics = &[TOPIC_FILTER.to_string()];
+        debug!("starting command handler");
 
+        let subscribe_topics = &[TOPIC_FILTER.to_string()];
         if !self.subscribe(subscribe_topics).await? {
             return Err(CommandHandlerError::MissingSubacks(
-                subscribe_topics.concat(),
+                subscribe_topics.to_vec().join(", "),
```
dmolokanov, line 83:
```suggestion
                subscribe_topics.join(", "),
```
nit: I believe you don't need `to_vec()`

dmolokanov: But from a design point of view, I'd rather pass array itself. Later on we may want to iterate through a list of topics for some reason. In the log it also be nice JSON like array
```suggestion
                return Err(CommandHandlerError::MissingSubacks(*subscribe_topics);
```

and-rewsmith: So if we have ```*subscribe_topics``` then it is ```[std::string::String; 1]```. How do we get the String out safely? We can use ```get()``` but will need to handle the error here which creates nested errors.

dmolokanov: Yeah, that's true...

PR #3451, opened 2020-08-25
[mqtt] Parse disconnecting client from message body

and-rewsmith:
We were previously parsing client id from topic. This was originally chosen for consistency with other operations, but here are the problems:
- client id can contain slashes, so you cannot subscribe to a multi-level match
- we have error-prone logic around parsing client id from topic


mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -135,7 +132,8 @@ impl CommandHandler {
 
     async fn handle_event(&mut self, event: Event) -> Result<(), HandleDisconnectError> {
         if let Event::Publication(publication) = event {
-            let client_id = parse_client_id(&publication.topic_name)?;
+            let client_id = str::from_utf8(&publication.payload)
+                .map_err(HandleDisconnectError::ParseClientId)?;
```
and-rewsmith, line 136: EdgeHub will only disconnect one client at a time

dmolokanov: For the sake of minimizing communication, should we be ready to process a list of client ids?

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -75,11 +72,11 @@ impl CommandHandler {
     }
 
     pub async fn run(mut self) -> Result<(), CommandHandlerError> {
-        let subscribe_topics = &[TOPIC_FILTER.to_string()];
+        let subscribe_topic = &[DISCONNECT_TOPIC.to_string()];
 
-        if !self.subscribe(subscribe_topics).await? {
+        if !self.subscribe(subscribe_topic).await? {
```
vadim-kovalyov, line 77: nit: why not return the error directly from subscribe method? `if` here is not really necessary.

mqtt/mqtt-edgehub/src/command.rs
and-rewsmith: I was actually thinking of doing this but seemed like a premature optimization and also adds unnecessary parse logic (which is one of the things I didn't like about the regex). I thought it was a good idea to not do it now, but if you disagree I will add it.

mqtt/mqtt-edgehub/src/command.rs
and-rewsmith: Oh yes I can return my own error. I will change.

mqtt/mqtt-edgehub/src/command.rs
dmolokanov: Wait, what parsing you are talking about? As far as I can see, it's not a JSON, but it should be

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -125,17 +118,18 @@ impl CommandHandler {
...
 
                 if subacks.is_empty() {
-                    return Ok(true);
+                    return Ok(());
                 }
             }
         }
 
-        Ok(false)
+        Err(CommandHandlerError::MissingSubacks(topics.join(", ")))
```
dmolokanov, line 126: shouldn't it be a list of missing topics? 
```suggestion
        Err(CommandHandlerError::MissingSubacks(subacks.join(", ")))
```

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -97,7 +90,7 @@ impl CommandHandler {
         Ok(())
     }
 
-    async fn subscribe(&mut self, topics: &[String]) -> Result<bool, CommandHandlerError> {
+    async fn subscribe(&mut self, topics: &[String]) -> Result<(), CommandHandlerError> {
         let subscriptions = topics.iter().map(|topic| proto::SubscribeTo {
             topic_filter: topic.to_string(),
             qos: proto::QoS::AtLeastOnce,
```
dmolokanov, line 96:
```rust
line 105
- let mut subacks: HashSet<String> = HashSet::from_iter(topics.to_vec());
+ let mut subacks = topics.to_vec();
```
Probably we don't even need a HashSet due to a small number of items in the list (<100)

dmolokanov, line 96:
```rust
line 113
- if let Event::SubscriptionUpdates(subscriptions) = event_or_err {
+ if let Event::SubscriptionUpdates(subscriptions) = event {
```
it's not an error anymore

mqtt/mqtt-edgehub/src/command.rs
dmolokanov: if subacks will become a Vec, lets pass it inside an error

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -125,17 +118,18 @@ impl CommandHandler {
...
         }
 
-        Ok(false)
+        Err(CommandHandlerError::MissingSubacks(topics.join(", ")))
     }
 
     async fn handle_event(&mut self, event: Event) -> Result<(), HandleDisconnectError> {
         if let Event::Publication(publication) = event {
-            let client_id = parse_client_id(&publication.topic_name)?;
+            let client_id = str::from_utf8(&publication.payload)
```
dmolokanov, line 131: I think we should parse client_id (or a list of ids) from a json

mqtt/mqtt-edgehub/src/command.rs
and-rewsmith: oops. yes i agree

mqtt/mqtt-edgehub/src/command.rs
and-rewsmith: I went with single id

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -90,14 +84,14 @@ impl CommandHandler {
             .map_err(CommandHandlerError::PollClientFailure)?
         {
             if let Err(e) = self.handle_event(event).await {
-                warn!(message = "failed to disconnect client", error = %e);
+                warn!(message = "error processing command handler event", error = %e);
```
and-rewsmith, line 87: I made this more generic so if we want to add logic to the command handler we can

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -109,33 +103,37 @@ impl CommandHandler {
...
             .try_next()
             .await
             .map_err(CommandHandlerError::PollClientFailure)?
         {
-            if let Event::SubscriptionUpdates(subscriptions) = event_or_err {
+            if let Event::SubscriptionUpdates(subscriptions) = event {
                 for subscription in subscriptions {
                     if let SubscriptionUpdateEvent::Subscribe(sub) = subscription {
-                        subacks.remove(&sub.topic_filter);
+                        subacks.retain(|topic| *topic != sub.topic_filter);
```
and-rewsmith, line 117: There might be a better way to do this. It will keep all topics that don't match the incoming suback. So, it will remove all instances of the topic from the vector.

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -123,13 +124,16 @@ impl CommandHandler {
...
-        Err(CommandHandlerError::MissingSubacks(topics.join(", ")))
+        Err(CommandHandlerError::MissingSubacks(subacks.join(", ")))
     }
 
     async fn handle_event(&mut self, event: Event) -> Result<(), HandleDisconnectError> {
         if let Event::Publication(publication) = event {
-            let client_id = str::from_utf8(&publication.payload)
-                .map_err(HandleDisconnectError::ParseClientId)?;
+            let payload = publication.payload.to_vec();
+            let payload = String::from_utf8(payload)
```
dmolokanov, line 133: You can skip a couple of allocations and deserialize to ClientId directly.
```rust
   async fn handle_event(&mut self, event: Event) -> Result<(), HandleDisconnectError> {
        if let Event::Publication(publication) = event {
            let client_id: ClientId = serde_json::from_slice(&publication.payload)
                .map_err(HandleDisconnectError::ParseClientId)?;

            info!("received disconnection request for client {}", client_id);

            if let Err(e) = self
                .broker_handle
                .send(Message::System(SystemEvent::ForceClientDisconnect(
                    client_id.clone(),
                )))
                .await
            {
                return Err(HandleDisconnectError::SignalError(e));
            }

            info!(
                "succeeded sending broker signal to disconnect client{}",
                client_id
            );
        }

        Ok(())
    }
```

mqtt/mqtt-edgehub/tests/command.rs
```diff
@@ -50,7 +50,7 @@ async fn disconnect_client() {
 
     let topic = "$edgehub/disconnect";
     edgehub_client
-        .publish_qos1(topic, "test-client", false)
+        .publish_qos1(topic, "\"test-client\"", false)
```
dmolokanov, line 53: why? also you can use raw string literals https://doc.rust-lang.org/stable/rust-by-example/std/str.html#literals-and-escapes

and-rewsmith: It needs to be valid json. I can change it to a literal.

mqtt/mqtt-edgehub/src/command.rs
dmolokanov: Oops, yeah, I missed that. Ordering is not important here, so we can use [swap_remove](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap_remove)

PR #3456, opened 2020-08-25
Update ARM32 & ARM64 base images to 3.1.7-bionic

yophilav:
Update ARM32 & ARM64 base images to 3.1.7-bionic to fix security vulnerability relates to glibc.2.27-3ubuntu1.2 (https://launchpad.net/ubuntu/+source/glibc/2.27-3ubuntu1.2).


edge-agent/docker/linux/arm32v7/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.6.0-linux-arm32v7
+ARG base_tag=1.0.6.2-linux-arm32v7
```
philipktlin, line 1: who/which branch uses 1.0.6.1? that's why we use 1.0.6.2 here, right?

yophilav: 1.0.6.1 was used for TRC, TestAnalyzer, NetworkController, & TwinTester. Those modules were added with non-root user recently.

philipktlin: Please discuss with Varun for long term schema versioning.

PR #3458, opened 2020-08-25
Update agent schema version in E2E test

richma-ms:
We've rev'ed the default schema version to 1.1, so the test data need to be updated to match.


edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/EdgeAgentConnectionTest.cs
```diff
@@ -1787,7 +1787,7 @@ static async Task UpdateAgentDesiredProperties(RegistryManager rm, string device
         {
             var dp = new
             {
-                schemaVersion = "1.0",
+                schemaVersion = "1.1.0",
```
yophilav, line 1790: Update the PR description please.

PR #3460, opened 2020-08-26
Expose unix domain sockets as parameters

SlavyMihov:
Mariner (Linux) uses different sockets than Ubuntu.
This change accommodates for the difference by exposing the unix domain sockets as optional parameters which can be customized.

### DIFF
--- a/scripts/linux/runE2ETest.sh
+++ b/scripts/linux/runE2ETest.sh
@@ -278,6 +278,10 @@ function process_args() {
     print_highlighted_message 'Process arguments'
     saveNextArg=0
     BYPASS_EDGE_INSTALLATION=""
+    CONNECT_MANAGEMENT_URI=""
+    CONNECT_WORKLOAD_URI=""
+    LISTEN_MANAGEMENT_URI=""
+    LISTEN_WORKLOAD_URI=""
     for arg in "$@"
     do
         if [ $saveNextArg -eq 1 ]; then
@@ -415,6 +419,18 @@ function process_args() {
         elif [ $saveNextArg -eq 44 ]; then
             RUNTIME_LOG_LEVEL="$arg"
             saveNextArg=0
+        elif [ $saveNextArg -eq 45 ]; then
+            CONNECT_MANAGEMENT_URI="$arg"
+            saveNextArg=0
+        elif [ $saveNextArg -eq 46 ]; then
+            CONNECT_WORKLOAD_URI="$arg"
+            saveNextArg=0
+        elif [ $saveNextArg -eq 47 ]; then
+            LISTEN_MANAGEMENT_URI="$arg"
+            saveNextArg=0
+        elif [ $saveNextArg -eq 48 ]; then
+            LISTEN_WORKLOAD_URI="$arg"
+            saveNextArg=0
         else
             case "$arg" in
                 '-h' | '--help' ) usage;;
@@ -462,6 +478,10 @@ function process_args() {
                 '-testInfo' ) saveNextArg=42;;
                 '-testStartDelay' ) saveNextArg=43;;
                 '-runtimeLogLevel' ) saveNextArg=44;;
+                '-connectManagementUri' ) saveNextArg=45;;
+                '-connectWorkloadUri' ) saveNextArg=46;;
+                '-listenManagementUri' ) saveNextArg=47;;
+                '-listenWorkloadUri' ) saveNextArg=48;;
                 '-bypassEdgeInstallation' ) BYPASS_EDGE_INSTALLATION="--bypass-edge-installation";;
                 '-cleanAll' ) CLEAN_ALL=1;;
                 * ) usage;;
@@ -564,6 +584,10 @@ function run_directmethod_test()
         -t "$ARTIFACT_IMAGE_BUILD_NUMBER-linux-$image_architecture_label" \
         --initialize-with-agent-artifact "$INITIALIZE_WITH_AGENT_ARTIFACT" \
         --verify-data-from-module "DirectMethodSender" \
+        --use-connect-management-uri="$CONNECT_MANAGEMENT_URI" \
+        --use-connect-workload-uri="$CONNECT_WORKLOAD_URI" \
+        --use-listen-management-uri="$LISTEN_MANAGEMENT_URI" \
+        --use-listen-workload-uri="$LISTEN_WORKLOAD_URI" \
         $BYPASS_EDGE_INSTALLATION \
         -l "$deployment_working_file" && ret=$? || ret=$?
 
@@ -694,6 +718,10 @@ function run_dps_provisioning_test() {
         -n "$(hostname)" \
         -tw "$E2E_TEST_DIR/artifacts/core-linux/e2e_test_files/twin_test_tempSensor.json" \
         --optimize_for_performance="$optimize_for_performance" \
+        --use-connect-management-uri="$CONNECT_MANAGEMENT_URI" \
+        --use-connect-workload-uri="$CONNECT_WORKLOAD_URI" \
+        --use-listen-management-uri="$LISTEN_MANAGEMENT_URI" \
+        --use-listen-workload-uri="$LISTEN_WORKLOAD_URI" \
         $BYPASS_EDGE_INSTALLATION \
         $dps_command_flags \
         -t "$ARTIFACT_IMAGE_BUILD_NUMBER-linux-$image_architecture_label" && ret=$? || ret=$? \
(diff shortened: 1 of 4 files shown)


smoke/IotEdgeQuickstart/Program.cs
```diff
@@ -84,6 +84,18 @@ class Program
         [Option("-h|--use-http=<hostname>", Description = "Modules talk to iotedged via tcp instead of unix domain socket")]
         public (bool useHttp, string hostname) UseHttp { get; } = (false, string.Empty);
 
+        [Option("--use-connect-management-uri=<connect_management_uri>", Description = "Modules talk to a custom connect management socket (default is unix:///var/run/iotedge/mgmt.sock)")]
```
philipktlin, line 87: If this parameter is not provided, it will use the value in config.yaml.  Don't put unix:// as default here, what if we change that value in config.yaml in the future.

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
```diff
@@ -65,20 +65,57 @@ static string GetIpAddress()
...
     }
 
+    public class UriSocks
+    {
+        const int ManagementPort = 15580;
+        const int WorkloadPort = 15581;
+
+        public UriSocks()
+        {
+            this.ConnectManagement = "unix:///var/run/iotedge/mgmt.sock";
```
philipktlin, line 75: Same as previous comment.  Don't hard-code this value here, you can get the value from config.yaml and override if it is provided explicitly.

smoke/IotEdgeQuickstart/Program.cs
SlavyMihov: This is just a description to the input parameters. I can remove the default socket values from the parameters descriptions if such clarification is unnecessary.

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
SlavyMihov: The default socket values used to be hardcoded in this file, to begin with.
My understanding is that _IoTEdgeQuickStart_ is responsible for creating the _config.yaml_ (and populating the correct socket values). The _config.yaml_ may not exist until _IoTEdgeQuickStart_ is run.

philipktlin: IoTEdgeQuickStart is used to install iotedge daemon and will create config.yaml.  After install iotedged, config.yaml will be created with default values.  In configure method https://github.com/Azure/iotedge/blob/6e7321fcb08d0aabf618a756eb58cad57758616c/smoke/IotEdgeQuickstart/details/IotedgedLinux.cs#L188, you can see that it only touches config parameters if they are provided.  Therefore for these URIs, if they are not provided, no action is needed in this method.

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
```diff
@@ -65,20 +65,57 @@ static string GetIpAddress()
         }
     }
 
+    public class UriSocks
```
philipktlin, line 68: Move to its own file.

philipktlin:
Thanks for your clarification, which those management and workload uri are hard-coded in IoTEdgedLinux.cs in IoTEdgeQuickstart project.

scripts/linux/runE2ETest.sh
```diff
@@ -278,6 +278,10 @@ function process_args() {
     print_highlighted_message 'Process arguments'
     saveNextArg=0
     BYPASS_EDGE_INSTALLATION=""
+    CONNECT_MANAGEMENT_URI="unix:///var/run/iotedge/mgmt.sock"
```
philipktlin, line 281: Just leave empty string for default value in this script.

smoke/IotEdgeQuickstart/Program.cs
```diff
@@ -84,6 +84,18 @@ class Program
...
+        public (bool useConnMan, string connect_management_uri) UseConnMan { get; } = (false, string.Empty);
```
philipktlin, line 88: why do you need to return (bool, string)? return string should be good enough, right?

scripts/linux/runE2ETest.sh
SlavyMihov: Makes sense.

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
```diff
@@ -65,20 +65,57 @@ static string GetIpAddress()
...
+    {
+        const int ManagementPort = 15580;
+        const int WorkloadPort = 15581;
+
+        public UriSocks()
```
philipktlin, line 73: Combine this constructor with the other one.  Set to default value if it doesn't provide,

smoke/IotEdgeQuickstart/Program.cs
```diff
@@ -209,7 +221,15 @@ async Task<int> OnExecuteAsync()
                                 ? Option.Some(string.IsNullOrEmpty(hostname) ? new HttpUris() : new HttpUris(hostname))
                                 : Option.None<HttpUris>();
 
-                            bootstrapper = new IotedgedLinux(this.BootstrapperArchivePath, credentials, uris, proxy, upstreamProtocolOption, !this.BypassEdgeInstallation);
+                            (bool useConnMan, string connect_management_uri) = this.UseConnMan;
+                            (bool useConnWrkld, string connect_workload_uri) = this.UseConnWrkld;
+                            (bool useLisMan, string listen_management_uri) = this.UseLisMan;
+                            (bool useLisWrkld, string listen_workload_uri) = this.UseLisWrkld;
+                            Option<UriSocks> socks = useConnMan || useConnWrkld || useLisMan || useLisWrkld
```
philipktlin, line 228: Once address UriSocks constructor comment, you should always call to the same constructor and just use UriSocks instead of Option<UriSocks>.

smoke/IotEdgeQuickstart/Program.cs
SlavyMihov: Good suggestion. Simplified it.

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
SlavyMihov: Makes sense.

smoke/IotEdgeQuickstart/Program.cs
SlavyMihov: Fixed.

scripts/linux/runE2ETest.sh
```diff
@@ -1097,6 +1149,10 @@ function usage() {
     echo ' -testInfo                                      Contains comma delimiter test information, e.g. build number and id, source branches of build, edgelet and images.' 
     echo ' -testStartDelay                                Tests start after delay for applicable modules'
     echo ' -runtimeLogLevel                               Value of RuntimeLogLevel envivronment variable for EdgeAgent in Long Haul and Stress tests [Default: debug] (EdgeHub RuntimeLogLevel is set implicitly set to be the same with edgeAgent)'
+    echo ' -connectManagementUri                          Customize connect management socket)'
```
philipktlin, line 1152: remove bracket?
```suggestion
    echo ' -connectManagementUri                          Customize connect management socket'
```

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
philipktlin: use internal

SlavyMihov: Fixed.

scripts/linux/runE2ETest.sh
SlavyMihov: Fixed.

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
```diff
@@ -65,20 +65,44 @@ static string GetIpAddress()
         }
     }
 
+    internal class UriSocks
```
philipktlin, line 68: create a new file for this class.

SlavyMihov: Done.

smoke/IotEdgeQuickstart/details/UriSocks.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+namespace IotEdgeQuickstart.Details
+{
+    internal class UriSocks
+    {
+        const int ManagementPort = 15580;
```
philipktlin, line 6: where is this used?

philipktlin (dismissed):
LGTM

smoke/IotEdgeQuickstart/details/UriSocks.cs
SlavyMihov: You're right. Removing some artifacts from the past.

PR #3461, opened 2020-08-26
Update to use Device.Client 1.28.1-NestedEdge

richma-ms:
Update to use Device.Client 1.28.1-NestedEdge


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/Microsoft.Azure.Devices.Edge.Agent.Diagnostics.csproj
```diff
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.27.3-NestedEdge" />
+    <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.28.1-NestedEdge" />
```
varunpuranik, line 12: Nit: why 1.28.1-NE instead of 1.28.1-NE?

PR #3462, opened 2020-08-26
Prepare E2E pipeline for ARM64 job

gordonwang0:


builds/e2e/templates/e2e-setup.yaml
```diff
@@ -122,7 +122,7 @@ steps:
       $context['installerPath'] = Convert-Path '$(Build.SourcesDirectory)/scripts/windows/setup'
     }
 
-    if ('$(arch)' -eq 'arm32v7')
+    if ('$(arch)' -eq 'arm32v7' -Or '$(arch)' -eq 'arm64v8')
```
damonbarry, line 125: Is the term `arm64v8` more appropriate/correct than `aarch64`?

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
```diff
@@ -35,6 +35,7 @@ public class EdgeDaemon : IEdgeDaemon
             switch (os)
             {
                 case "Ubuntu":
+                case "NXP-LSDK":
```
damonbarry, line 38: Since we want to check in generic support for ARM64, let's remove this board-specific reference for now.

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -103,6 +103,11 @@ public async Task TempFilter()
         // Test Temperature Filter Function: https://docs.microsoft.com/en-us/azure/iot-edge/tutorial-deploy-function
         public async Task TempFilterFunc()
         {
+            if (OsPlatform.IsArm() && OsPlatform.Is64Bit())
+            {
+                Assert.Ignore("TempFilterFun is disabled for arm64 because azureiotedge-functions-filter does not exist for arm64");
```
damonbarry, line 108: Missing the 'c' on the end of "TempFilterFunc"

builds/e2e/templates/e2e-setup.yaml
gordonwang0: I believe the two terms are equivalent. Seems like they're used interchangeably.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
gordonwang0: Removed

test/Microsoft.Azure.Devices.Edge.Test/Module.cs
gordonwang0: Fixed

builds/e2e/e2e.yaml
```diff
@@ -42,9 +42,31 @@ jobs:
     - template: templates/e2e-clear-docker-cached-images.yaml
     - template: templates/e2e-run.yaml
 
-################################################################################	
+################################################################################
```
damonbarry, line 45: Since we don't have a 64-bit arm agent ready (yet) to run tests, let's remove this job from `e2e.yaml`.

gordonwang0: Removed

damonbarry:
@gordonwang0 Before merging can you update the PR title to reflect the final changes? We're preparing the end-to-end pipeline for arm64, but not enabling it yet...

PR #3463, opened 2020-08-26
Fixed potentialy zombie task in TestClient.

vadim-kovalyov:
In current code, even when we return from the async block in TestClient, it looks like that `event_loop` task remains running.

This is an attempt to fix it.

### DIFF
--- a/mqtt/mqtt-broker-tests-util/src/client.rs
+++ b/mqtt/mqtt-broker-tests-util/src/client.rs
@@ -1,13 +1,11 @@
+#![allow(clippy::mut_mut)]
 use std::time::Duration;
 
 use bytes::Bytes;
-use futures_util::{future::select, pin_mut, StreamExt};
+use futures::{select, stream::StreamExt};
 use tokio::{
     net::ToSocketAddrs,
-    sync::{
-        mpsc::{self, UnboundedReceiver},
-        oneshot::{self, Sender},
-    },
+    sync::mpsc::{self, Sender, UnboundedReceiver},
     task::JoinHandle,
 };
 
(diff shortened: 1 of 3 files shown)


mqtt/mqtt-broker-tests-util/src/client.rs
```diff
@@ -1,13 +1,11 @@
+#![allow(clippy::mut_mut)]
 use std::time::Duration;
 
 use bytes::Bytes;
-use futures_util::{future::select, pin_mut, StreamExt};
+use futures::{select, stream::StreamExt};
```
dmolokanov, line 5: why? I believe it is re-import anyways

mqtt/mqtt-broker-tests-util/src/client.rs
```diff
@@ -240,27 +239,40 @@ where
...
-                    };
-                }
-            };
-            pin_mut!(event_loop);
-            select(event_loop, tx).await;
+            let mut termination_receiver = termination_receiver.fuse();
+            let mut client = client.fuse();
+
+            loop {
+                select! {
```
dmolokanov, line 249: Yeah, every time I try to use `select` macro it doesn't look nice and makes a lot of nesting

mqtt/mqtt-broker-tests-util/src/client.rs
```diff
@@ -240,27 +239,40 @@ where
         let (sub_sender, sub_receiver) = mpsc::unbounded_channel();
         let (conn_sender, conn_receiver) = mpsc::unbounded_channel();
 
-        let (termination_handle, tx) = oneshot::channel::<()>();
+        let (termination_handle, termination_receiver) = mpsc::channel::<()>(1);
```
dmolokanov, line 242: why not oneshot?

vadim-kovalyov: oneshot is not a stream, and select needs a fused stream.... this is a negligible downside of this approach.

vadim-kovalyov: As discussed offline, it looks like I could use FutureExt and fuse oneshot future

PR #3467, opened 2020-08-28
Plumb originating Edge device ID to the root Edge on single identity refresh

richma-ms:
Refreshing an individual identity in a nested hierarchy can run into problems if the target identity is newly added, and the root Edge doesn't know who's the direct parent of the target, because the root then wouldn't know who it's acting OnBehalfOf.

This change introduce a custom Edge header on the HTTP request for GetDeviceAndModuleOnBehalfOf to pass along the device ID of the original Edge that made the request, so the the root can act OnBehalfOf that Edge to get the target identity.

There's still the corner-case where the leaf device can skip its parent, and connect directly to the grandparent, in which case the "original" Edge will be wrong, so we still have the full cache refresh fallback to handle it.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -116,12 +118,12 @@ static Task<T> ExecuteWithRetry<T>(Func<Task<T>> func, Action<RetryingEventArgs>
             return transientRetryPolicy.ExecuteAsync(func);
         }
 
-        async Task<ScopeResult> GetIdentityOnBehalfOfWithRetry(Uri uri, string deviceId, Option<string> moduleId)
+        async Task<ScopeResult> GetIdentityOnBehalfOfWithRetry(Uri uri, string deviceId, Option<string> moduleId, string onBehalfOfDevice)
```
philipktlin, line 121: Nit:This method is only called from GetIdentityOnBehalfOfAsync.  Suggest to combine.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -87,10 +87,12 @@ public class NestedDeviceScopeApiClient : IDeviceScopeApiClient
         public Task<ScopeResult> GetNextAsync(string continuationToken) =>
             this.GetIdentitiesInTargetScopeWithRetry(this.GetNestedScopeServiceUri(), Option.Some(continuationToken));
 
-        public Task<ScopeResult> GetIdentityAsync(string targetDeviceId, string targetModuleId)
+        public Task<ScopeResult> GetIdentityAsync(string deviceId, string moduleId) => throw new NotImplementedException("Use GetIdentityOnBehalfOfAsync() instead");
+
+        public Task<ScopeResult> GetIdentityOnBehalfOfAsync(string targetDeviceId, Option<string> targetModuleId, string onBehalfOfDevice)
```
philipktlin, line 92: check onBehalfOfDevice not null or empty?  And onBehalfOfDevice is the auth chain, right? will it contain edgeHub as part of the value? if so, using "Device" wording may be confusing.

philipktlin: onBehalfOfDevice is the auth chain, right? will it contain edgeHub as part of the value? if so, using "Device" wording may be confusing.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
```diff
@@ -151,25 +153,17 @@ async Task<ScopeResult> GetIdentitiesInTargetScopeWithRetry(Uri uri, Option<stri
             }
         }
 
-        async Task<ScopeResult> GetIdentityOnBehalfOf(Uri uri, string deviceId, Option<string> moduleId)
+        async Task<ScopeResult> GetIdentityOnBehalfOf(Uri uri, string deviceId, Option<string> moduleId, string onBehalfOfDevice)
```
philipktlin, line 156: add Async suffix for async method; you can add "Internal" to existing name before appending "Async".

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -34,23 +33,25 @@ public IServiceIdentitiesIterator GetServiceIdentitiesIterator()
             }
         }
 
-        public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId)
+        public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId, string onBehalfOfDevice)
```
philipktlin, line 36: check not null or empty for onBehalfOfDevice

philipktlin: Async suffix

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -152,12 +159,12 @@ await serviceIdentity
...
                 else
                 {
-                    Events.SkipRefreshServiceIdentity(id, this.identitiesLastRefreshTime[id], this.refreshDelay);
+                    Events.SkipRefreshServiceIdentity(refreshTarget, this.identitiesLastRefreshTime[refreshTarget], this.refreshDelay);
                 }
             }
             catch (Exception e)
             {
-                Events.ErrorRefreshingCache(e, id);
+                Events.ErrorRefreshingCache(e, refreshTarget);
```
philipktlin, line 167: log onBehalfOfDevice and invokeServiceIdentitiesUpdated?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -191,17 +198,17 @@ public void Dispose()
             this.refreshCacheTask?.Dispose();
         }
 
-        internal Task<Option<ServiceIdentity>> GetServiceIdentityFromService(string id)
+        internal Task<Option<ServiceIdentity>> GetServiceIdentityFromService(string targetId, string onBehalfOfDevice)
```
philipktlin, line 201: AFAIK we would like to stop all bad inputs in public methods.  Please make sure all callers to this method provides a valid value for targetId and onBehalfOfDevice, e.g. targetId cannot be null or empty.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -115,12 +115,21 @@ async Task<EdgeHubScopeResult> HandleGetDeviceAndModuleOnBehalfOfAsync(string ac
                 targetId += "/" + request.TargetModuleId;
             }
 
+            IEdgeHub edgeHub = await this.edgeHubGetter;
+
+            // Check if the actor has provided the originating EdgeHub ID,
+            // if not, then we must by definition by the origin.
```
philipktlin, line 121:
```suggestion
            // if not, then we must by definition be the origin.
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
richma-ms: no, the onBehalfOfDevice is the ID of the first Edge that received the leaf connection, so it's just a DeviceId.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/NestedDeviceScopeApiClient.cs
richma-ms: yea it's a little weird, but it seems to be a pattern from the legacy device scope client implementation, so I opted to just keep it as is for consistency.

---
In reply to: [479387562](https://github.com/Azure/iotedge/pull/3467#discussion_r479387562) [](ancestors = 479387562)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
richma-ms: None of the others have the async suffix (they prob should), but updating it means updating the interface and all the related consumers...going to call this one out of scope for this PR =/

---
In reply to: [479393278](https://github.com/Azure/iotedge/pull/3467#discussion_r479393278) [](ancestors = 479393278)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
richma-ms: added log for onBehalfOfDevice, doesn't really make sense to log invokeServiceIdentitiesUpdated.

---
In reply to: [479403827](https://github.com/Azure/iotedge/pull/3467#discussion_r479403827) [](ancestors = 479403827)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -94,24 +98,31 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId)
...
+                    res = await client.GetIdentityOnBehalfOfAsync(deviceId, Option.Some(moduleId), onBehalfOfDevice);
                 }
                 else
                 {
                     client = this.securityScopesApiClientProvider.CreateDeviceScopeClient();
+                    res = await client.GetIdentityAsync(deviceId, moduleId);
                 }
 
-                ScopeResult res = await client.GetIdentityAsync(deviceId, moduleId);
                 scopeResult = Option.Maybe(res);
```
philipktlin, line 126: Even you can remove "ScopeResult res" variable and directly assign to scopeResult variable.
```suggestion
                scopeResult = Option.Some(res);
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
```diff
@@ -26,6 +26,7 @@ public class Constants
         public const string SecurityMessageIoTHubInterfaceId = "urn:azureiot:Security:SecurityAgent:1";
 
         public const string ServiceApiIdHeaderKey = "x-ms-edge-moduleId";
+        public const string OriginEdgeHeaderKey = "x-ms-edge-origin";
```
philipktlin, line 29: is this used only for nested scenario? and is this always edge device id?  if so, we should rename it to make it obvious.

richma-ms: any suggestions for naming?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/IServiceIdentityHierarchy.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
 
     public interface IServiceIdentityHierarchy
     {
+        string GetActorDeviceId();
```
philipktlin, line 11: I think this is not a method related to this interface.  I think we should pass edgeDeviceId to DeviceScopeIdentitiesCache directly instead.

richma-ms: The hierarchy is the source of truth for caching and getting all things identity related, i.e. all IDs, properties, and relationships between them.  The actor is just a special identity that's the root of the hierarchy.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
philipktlin: if it is used for nested scenario only, should it be "x-ms-edge-nested-origin". @varunpuranik what do you think?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/IServiceIdentityHierarchy.cs
philipktlin: Say we have childEdge, parentEdge and rootEdge.  When requests come from childEdge to parentEdge.  The authChain returns for childEdge should be "childEdge;parentEdge;rootEdge", not "childEdge;parentEdge", do i understand correctly?

huguesBouvier (approved):

PR #3469, opened 2020-08-28
[mqtt] Improve sidecar startup and command handler error handling

and-rewsmith:
There are three motivations for this PR:

1 -  The sidecar startup logic is not structured in a way where it can appropriately handle errors from within the sidecar threads. We can change the structure so that we can init the components, fail if we can't init them, then trust them to run without having a non-recoverable failure.
2 - Refactor the command handler such that it has non-recoverable logic in a grouping of init steps that can percolate the error all the way up. When the command handler is run it can log recoverable failures. Doing it this way eliminates the problem where the command handler can fail and we never know.
~3 - The mqtt3 client that the command handler uses should never disconnect, however if it does the command handler will stop and not get restarted. I changed the shutdown handle logic so that the client will attempt to reconnect unless the command handler is explicitly told to shutdown.~


mqtt/mqtt-edgehub/tests/command.rs
```diff
@@ -28,13 +27,13 @@ async fn disconnect_client() {
...
 
-    let (mut command_handler_shutdown_handle, join_handle) =
+    let (command_handler_shutdown_handle, join_handle) =
         start_command_handler(broker_handle, TEST_SERVER_ADDRESS.to_string())
             .await
             .expect("could not start command handler");
 
     // TODO: This wait is necessary because edgehub can send disconnect before command handler subscribes to disconnect topic
-    //       We can remove once we have a proper approach for starting command handler before edgehub sends disconnects
+    //       We will soon have a proper approach for starting command handler before edgehub sends disconnects
```
and-rewsmith, line 36: Once we have this proper startup logic, we still can't remove this wait unless we make other changes to the test. I'm not sure how this is going to look.

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -58,81 +78,49 @@ impl CommandHandler {
...
-                Ok(ShutdownHandle(shutdown_handle))
-            })
+        Ok(ShutdownHandle(
+            self.termination_handle.clone(),
+            self.client.shutdown_handle()?,
+        ))
     }
 
-    pub async fn run(mut self) -> Result<(), CommandHandlerError> {
+    pub async fn run(mut self) -> Result<(), Error> {
```
and-rewsmith, line 101: This function will poll the client, attempting to handle the events. If the client is disconnected, it will reconnect unless the command handler has been signaled to shutdown. This is an improvement because this structure allows for the client to reconnect if something weird happens inside the edgehub container.

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -16,6 +16,8 @@ impl BridgeController {
     }
 
     pub async fn start(&mut self) -> Result<()> {
+        info!("starting bridge");
```
and-rewsmith, line 19: I feel it is a better pattern to have the "starting xxx" logging inside the actual start method

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -165,50 +165,35 @@ async fn server_certificate_renewal(renew_at: DateTime<Utc>) {
...
     system_address: String,
-) -> (SidecarShutdownHandle, JoinHandle<Result<()>>) {
-    let (termination_handle, tx) = channel::<()>();
+) -> Result<(SidecarShutdownHandle, JoinHandle<Result<()>>)> {
+    let (sidecar_termination_handle, sidecar_termination_receiver) = channel::<()>();
+
+    let device_id = env::var(DEVICE_ID_ENV)?;
+    let command_handler =
+        CommandHandler::new(broker_handle, system_address, device_id.as_str()).await?;
+    let command_handler_shutdown_handle = command_handler.shutdown_handle()?;
```
and-rewsmith, line 174: These are fail-able init steps that will percolate errors all the way up. Once the components are initialized we can trust that they will handle errors in the event loop below.

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -165,50 +165,35 @@ async fn server_certificate_renewal(renew_at: DateTime<Utc>) {
...
+
+    let mut bridge_controller = BridgeController::new();
```
and-rewsmith, line 176: These are fail-able init steps that will percolate errors all the way up. Once the components are initialized we can trust that they will handle errors in the event loop below.

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -165,50 +165,35 @@ async fn server_certificate_renewal(renew_at: DateTime<Utc>) {
...
-        info!("starting command handler...");
-        let (mut command_handler_shutdown_handle, command_handler_join_handle) =
-            start_command_handler(broker_handle, system_address).await?;
+    let device_id = env::var(DEVICE_ID_ENV)?;
+    let command_handler =
+        CommandHandler::new(broker_handle, system_address, device_id.as_str()).await?;
+    let command_handler_shutdown_handle = command_handler.shutdown_handle()?;
 
-        start_bridge().await?;
+    let mut bridge_controller = BridgeController::new();
```
and-rewsmith, line 176: These are fail-able init steps that will percolate errors all the way up. Once the components are initialized we can trust that they will handle errors in the event loop below.

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -58,81 +78,48 @@ impl CommandHandler {
...
+                        error!(message = "error processing command handler event", error = %e);
                     }
                 }
+                Ok(None) => {
+                    error!("command handler client disconnected");
 
-                if subacks.is_empty() {
-                    debug!("command handler successfully subscribed to disconnect topic");
-                    return Ok(());
+                    if let Ok(()) = self.termination_receiver.try_recv() {
```
vadim-kovalyov, line 114: I'm not sure if this will work. I believe that client never returns `None`. What you need to do is to use `futures::select!` macro to select b/w `self.termination_receiver.try_recv()` and `self.client.try_next()`. 

Take a look at recent fix that I've made as an example - https://github.com/Azure/iotedge/blob/0cffc25ac206d23c96d594c11cb68062f58a75e7/mqtt/mqtt-broker-tests-util/src/client.rs#L251-L279

mqtt/mqtt-edgehub/tests/command.rs
vadim-kovalyov: I thought that by the time `start_command_handler` completes, the CH already subscribed? so there is no reason for `defay_for` anymore.

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -165,50 +165,35 @@ async fn server_certificate_renewal(renew_at: DateTime<Utc>) {
...
-    let command_handler = CommandHandler::new(broker_handle, system_address, device_id.as_str());
-    let shutdown_handle = command_handler.shutdown_handle()?;
-
-    let join_handle = tokio::spawn(command_handler.run());
+        command_handler_shutdown_handle.shutdown().await?;
+        command_handler_join_handle.await?;
 
-    Ok((shutdown_handle, join_handle))
-}
+        bridge.await?;
```
vadim-kovalyov, line 188: As I understand, bridge shutdown is not yet implemented?

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -58,81 +78,48 @@ impl CommandHandler {
             Duration::from_secs(60),
         );
 
-        CommandHandler {
+        let subscribe_topics = &[DISCONNECT_TOPIC.to_string()];
+        subscribe(&mut client, subscribe_topics).await?;
+
+        let (termination_handle, termination_receiver) = mpsc::channel(5);
```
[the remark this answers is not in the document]
and-rewsmith: I was also going to ask about this but forgot to make the comment. I saw everywhere else used 5

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -12,12 +16,22 @@ use mqtt_broker::{BrokerHandle, ClientId, Error, Message, SystemEvent};
 
 const DISCONNECT_TOPIC: &str = "$edgehub/disconnect";
 
-#[derive(Debug)]
-pub struct ShutdownHandle(mqtt3::ShutdownHandle);
+pub struct ShutdownHandle(Sender<()>, mqtt3::ShutdownHandle);
```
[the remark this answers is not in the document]
and-rewsmith: I tried making them named by expanding the struct as such:
```
struct foo {
  field: type,
  field2: type2,
}
```

This didn't work because I was using the parenthesis, but I will try again just to make sure. Lmk if I am missing something obvious.

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
and-rewsmith: Yep. When it gets implemented then we can add the shutdown logic like we do here.

mqtt/mqtt-edgehub/tests/command.rs
and-rewsmith: Oh you are correct.

mqtt/mqtt-edgehub/src/command.rs
and-rewsmith: The client returns none because the client is shutdown when the command handler shutdown() method is called, right? Also, the shutdown works in the test so I think the client has to be returning None.

mqtt/mqtt-edgehub/src/command.rs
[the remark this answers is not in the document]
and-rewsmith: Nevermind I tried again and it works

mqtt/mqtt-edgehub/src/command.rs
vadim-kovalyov: Yeah, I think you're right, it will return `None` in case of shutdown. But then what's the point of `command_handler_shutdown` channel? If client is being shut down we can just return out of the func and CommandHandler is done.

and-rewsmith: Vadim and I discussed offline and he pointed that client is non-recoverable after it returns None, which is only possible via client.shutdown. Therefore we can revert this change to command handler shutdown.

PR #3471, opened 2020-08-30
Allow edgeAgent to run as non-root in Docker.

darobs:
This allows Agent to run as non-root in docker.

We have a default user (edgeagentuser, default UID 1000) available if the user wants to start the container as non-root.  This will be created when the edgeAgent image is built.  However, if the user does not want to use UID 1000, they can start the image as root, set `EDGEAGENTUSER_ID` to the value that they want, and the container's script will create the user in the container and setup the correct permissions for this user to run edgeAgent. 

Some notes:

### UID and certificate stores
Edge agent wants to install a trust bundle into a store. The dotnet core code gives us 2 location options for `X509Store`, `LocalMachine` or `CurrentUser`. The former needs write access to system directories and the latter needs a known user's home directory. The agent code uses the `CurrentUser` location. In order for the Agent to run as non-root, the current user's UID needs to have a writable home directory associated with it.  I chose to create UID 1000 when doing this for Edge on K8s, and I chose to continue to use this as the default for running as non-root on docker.

Regardless of the UID chosen for the container, this **must match** the UID ownership of the management socket file to allow edgeAgent access to the management socket.  If the user runs the agent as non-root through Docker settings, the management socket must be set.

### Setting UID on management socket.

Example using UID 1000.

**Ubuntu/debian 9 systems**
```bash
sudo systemctl edit iotedge.mgmt.socket
# add the following lines
```
> [Socket]
> SocketUser=1000
```bash
#save the override file.
sudo systemctl daemon-reload
sudo systemctl restart iotedge.mgmt.socket
```

**On other systems**
```bash
sudo chown 1000:iotedge /var/lib/iotedge/mgmt.sock
```


edgelet/contrib/centos/iotedge.spec
```diff
@@ -121,7 +122,7 @@ echo "==========================================================================
 %{_unitdir}/%{name}.service
 
 # sockets
-%attr(660, %{iotedge_user}, %{iotedge_group}) %{iotedge_home}/mgmt.sock
+%attr(660, %{agentuser}, %{iotedge_group}) %{iotedge_home}/mgmt.sock
```
massand, line 125: We are replacing the iotedge user here, which will break current functionality for customers. Instead add an ACL for agent user for mgmt sock.

edgelet/contrib/debian8/postinst
```diff
@@ -28,7 +29,7 @@ case "$1" in
             # Do this for debian systems which don't support FDNAMES
             touch /var/lib/iotedge/mgmt.sock
             touch /var/lib/iotedge/workload.sock
-            chown ${USER}:${GROUP} /var/lib/iotedge/mgmt.sock
+            chown ${AGENTUSER}:${GROUP} /var/lib/iotedge/mgmt.sock
```
massand, line 32: This opens up edgeAgent to access all APIs. Is there a plan to limit the APIs in iotedged?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
```diff
@@ -0,0 +1,99 @@
...
+echo "$(date --utc +"%Y-%m-%d %H:%M:%S %:z") Starting Edge Agent"
+
+TARGET_UID="${EDGEAGENTUSER_ID:-1000}"
+cuid=$(id -u)
+
+if [ $cuid -eq 0 ]
+then
+  # which user creation tool to use.
+  usercreate=useradd
+  if cat /etc/os-release | grep -i alpine > /dev/null
```
arsing, line 35: Test with `! command -v useradd` instead of `/etc/os-release`

```sh
# command -v useradd; echo $?

127
```

edge-agent/docker/linux/amd64/Dockerfile
```diff
@@ -9,8 +9,9 @@ ARG EXE_DIR=.
 RUN apk update && \
     apk add --no-cache snappy
 
-# create a user to allow agent to optionally run as non-root
-ARG EDGEAGENTUSER_ID=1000	
+# create a default user to allow agent to run as non-root
+ARG EDGEAGENTUSER_ID
+ENV EDGEAGENTUSER_ID ${EDGEAGENTUSER_ID:-1000}
 RUN adduser -Ds /bin/sh -u ${EDGEAGENTUSER_ID} edgeagentuser
```
arsing, line 15: Is there a point in still doing this since the entrypoint will do it anyway?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
```diff
@@ -0,0 +1,99 @@
...
+    else
+      adduser -Ds /bin/sh -u "${TARGET_UID}" "agent${TARGET_UID}"
+    fi
+  fi
+
+  username=$(getent passwd "${TARGET_UID}" | awk -F ':' '{ print $1; }')
+
+  # If "StorageFolder" env variable exists, use as basepath, else use /tmp
+  # same for backuppath
+  agentstorage=$(env | grep -m 1 -i StorageFolder | awk -F '=' '{ print $2;}')
```
arsing, line 57:
```diff
-  agentstorage=$(env | grep -m 1 -i StorageFolder | awk -F '=' '{ print $2;}')
+  agentstorage=$(env | grep -m 1 -i StorageFolder | awk -F '=' '{ print $2; }')
```

Same for BackupFolder.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
```diff
@@ -0,0 +1,99 @@
...
+  agentbackup=$(env | grep -m 1 -i BackupFolder | awk -F '=' '{ print $2;}')
+  storagepath=${agentstorage:-/tmp}/edgeAgent
+  backuppath=${agentbackup:-/tmp}/edgeAgent_backup
+  # If basepath/edgeAgent exists, make sure all files are owned by TARGET_UID
+  if [ -d $storagepath ]
+  then
+    storageuid=$(stat -c "%u" "$storagepath")
+    if [ ${TARGET_UID} -ne ${storageuid} ]
```
arsing, line 65: Is there a point in testing first? Just `chown` it anyway?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
```diff
@@ -0,0 +1,99 @@
...
+  # Strip "unix://" prefix, and if that is a file that exists, change the ownership.
+  mgmt=${IOTEDGE_MANAGEMENTURI#unix:\/\/}
+  if [ -e "$mgmt" ]
+  then
+    echo "$(date --utc +"%Y-%m-%d %H:%M:%S %:z") Changing ownership of management socket: ${mgmt}"
+    chown -f "${TARGET_UID}" "$mgmt"
+  fi
+
+  # Ensure backup.json is writeable.
+  touch backup.json
```
arsing, line 92: Might be better to only `chown` it after checking if the file exists, rather than always `touch`ing it into existence as an empty file? I see that `DeploymentFileBackup.ReadFromBackupAsync` would would throw an exception for a malformed file whereas it would silently return `Empty` if the file didn't exist at all. The exception is handled fine, so it's not a problem, but it would be nice to not have it throw at all?

arsing (changes requested):

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
```diff
@@ -0,0 +1,93 @@
...
+  then
+    # Alpine only has "adduser"
+    usercreate=adduser
+  fi
+
+  # Create the agent user id if it does not exist
+  if ! getent passwd "${TARGET_UID}" >/dev/null
+  then
+    echo "Creating UID ${TARGET_UID} as agent${TARGET_UID}"
+    if [ $usercreate = useradd ]
```
CindyXing, line 45: Nit pick, why not just check the alpine here and choose the command accordingly?

edge-agent/docker/linux/amd64/Dockerfile
CindyXing: I have the same comment given the script setup the environment variable & default value as well.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
```diff
@@ -0,0 +1,93 @@
...
+    then
+      useradd -ms /bin/bash -u "${TARGET_UID}" "agent${TARGET_UID}"
+    else
+      adduser -Ds /bin/sh -u "${TARGET_UID}" "agent${TARGET_UID}"
```
CindyXing, line 49: What if user creation failed?

CindyXing (commented):
The success of the script execution is very important for launching edgeagent. I'd suggest we look through and add loggging in case of failure.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
darobs: The main reason why I `touch`ed it is because it resides in the current working directory, "/app" which is not writeable by the new user, and so it can't create the `backup.json` file in the code. I should probably just move the `backup.json` into the storageFolder.

arsing: >The main reason why I `touch`ed it is because it resides in the current working directory, "/app" which is not writeable by the new user, and so it can't create the `backup.json` file in the code.

I see. It does seem brittle in that it relies on EA writing the file by opening the existing one instead of deleting and recreating it. It *does* do that, but it could just as well not have.

You could `chown` `/app` to be writable instead of this, but...

>I should probably just move the backup.json into the storageFolder.

... yes, writing to the "installation" directory did seem like a strange decision, so this the better approach.

edge-agent/docker/linux/amd64/Dockerfile
darobs: This provides the user a UID they can set to run as non-root at start.  This is also needed for Edge on K8s, as some clusters we support have a policy to start running as non-root.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
darobs: I was trying to be more efficient, if there were a lot of files in the directory, but it's probably not necessary.

edge-agent/docker/linux/amd64/Dockerfile
arsing: Ah yes, forgot about that.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
darobs: I took @arsing 's suggestion (use `command -v`), because it's more elegant than what I did, and is the only place where the commands differ between Alpine and Bionic (our two agent bases).

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Service/agentStart.sh
darobs: The script returns and the container crashes.  It *shouldn't* be an issue, but the logs should be as clear as an sh script allows,
You'll see 
```
<datetime> Creating UID 2434 as agent2434
<error from useradd/adduser>
```

arsing (approved):

PR #3473, opened 2020-09-01
Expanding the Prometheus metric parser

daweim0:
This PR expands the prometheus metric parser to allow scientific notation and escaped quotes inside label values. It also removes the last o from "prometheous" in a few places.

Edits and suggestions are welcome
Best, David Michelman


arsing:
Merged in https://github.com/Azure/iotedge/pull/3580

PR #3474, opened 2020-09-01
Refactor Command handler and add handle_authorized_identities event

dylanbronson:
Refactoring CommandHandler to take multiple commands. Also, adding a new Command, handle_authorized_identities, that will intake authorized identities as a payload from edgeHub, convert it to ServiceIdentity objects and send that to the broker as a SystemEvent. 

Currently, the broker just has a dummy implementation for this SystemEvent. 

We do this because edgeHub does the authorization for all identities, and the broker just receives it from edgeHub whenever the edgeHub cache is updated.


mqtt/mqtt-broker/src/lib.rs
```diff
@@ -161,6 +161,13 @@ impl ConnReq {
         (self.peer_addr, self.connect, self.handle)
     }
 }
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ServiceIdentity {
```
vadim-kovalyov, line 165: as discussed offline, we would need to come up with more generic approach for Authorization updates. One option that I played with is in this commit (based from your branch) - https://github.com/vadim-kovalyov/iotedge/commit/15ed66fa223a7c6e9cd1e8c74009c8b2563f1f76

Feel free to just take it and patch your branch.

mqtt/mqtt-edgehub/tests/command.rs
```diff
@@ -1,17 +1,21 @@
...
 use mqtt_broker::{auth::AllowAll, BrokerBuilder, BrokerHandle};
 use mqtt_broker_tests_util::{
     client::TestClientBuilder,
     packet_stream::PacketStream,
     server::{start_server, DummyAuthenticator},
 };
-use mqtt_edgehub::command::{CommandHandler, ShutdownHandle};
+use mqtt_edgehub::command::{handle_authorized_identities, handle_disconnect, Command};
 
+use mqtt_edgehub::command_handler::{CommandHandler, ShutdownHandle};
```
vadim-kovalyov, line 15: nit: no need for new line

mqtt/mqtt-edgehub/tests/command.rs
```diff
@@ -62,12 +67,32 @@ async fn disconnect_client() {
     edgehub_client.shutdown().await;
 }
 
+fn init_commands() -> HashMap<String, Command> {
```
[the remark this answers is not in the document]
vadim-kovalyov: I think once you implement a trait (as @dmolokanov suggested) and have a separate public struct for every command, you can just instantiate a command.

```
fn init_commands() -> HashMap<String, Box<dyn Command>> {
    let mut commands = HashMap::new();
    commands.insert(Box::new(DisconnectCommand));
    commands.insert(Box::new(AuthorizationUpdateCommand));
    commands
}
```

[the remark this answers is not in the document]
vadim-kovalyov: But, on the other side, you can just instantiate a command inline in the test, b/c you don't need to have AuthorizationUpdateCommand in disconnect_client test.

vadim-kovalyov (commented):
So far this looks great, you are on the right track. Let's just implement a Command trait as @dmolokanov suggested and see how it looks.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -1,9 +1,12 @@
+use crate::command::ServiceIdentity;
```
vadim-kovalyov, line 1: nit: rearrange

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -1,200 +1,116 @@
...
-use tokio::{net::TcpStream, stream::StreamExt};
-use tracing::{debug, error, info};
+use tracing::info;
 
-use mqtt3::{
-    proto, Client, Event, IoSource, ShutdownError, SubscriptionUpdateEvent, UpdateSubscriptionError,
-};
+use mqtt3::ReceivedPublication;
 use mqtt_broker::{BrokerHandle, ClientId, Error, Message, SystemEvent};
+use std::collections::HashMap;
```
vadim-kovalyov, line 7: nit: rearrange

mqtt/mqtt-edgehub/tests/command.rs
```diff
@@ -1,17 +1,20 @@
...
     packet_stream::PacketStream,
     server::{start_server, DummyAuthenticator},
 };
-use mqtt_edgehub::command::{CommandHandler, ShutdownHandle};
-
+use mqtt_edgehub::{
+    command::{init_commands, Command},
+    command_handler::{CommandHandler, ShutdownHandle},
+};
+const DISCONNECT_TOPIC: &str = "$edgehub/disconnect";
```
vadim-kovalyov, line 17: nit: newline

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -1,200 +1,116 @@
...
-            if subacks.is_empty() {
-                debug!("command handler successfully subscribed to disconnect topic");
-                return Ok(());
-            }
+    fn handle(
+        &self,
+        broker_handle: &mut BrokerHandle,
+        publication: &ReceivedPublication,
+    ) -> Result<(), HandleEventError> {
+        let array: ServiceIdentity = serde_json::from_slice(&publication.payload)
```
vadim-kovalyov, line 68: Make sure that `ServiceIdentity` struct matches the actual json payload from edge hub.

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -1,200 +1,116 @@
...
 }
 
 #[derive(Debug, thiserror::Error)]
-enum HandleDisconnectError {
+pub enum HandleEventError {
     #[error("failed to parse client id from message payload")]
     ParseClientId(#[from] SerdeError),
 
+    #[error("failed while sending authorized identities to broker")]
+    SendAuthorizedIdentitiesToBroker(Error),
```
and-rewsmith, line 112: super nit: these error messages aren't consistently worded

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -1,200 +1,116 @@
...
+        DISCONNECT_TOPIC.to_string(),
+        Box::new(Disconnect {
+            topic: DISCONNECT_TOPIC.to_string(),
+        }),
+    );
+    commands.insert(
+        AUTHORIZED_IDENTITIES_TOPIC.to_string(),
+        Box::new(AuthorizedIdentities {
+            topic: AUTHORIZED_IDENTITIES_TOPIC.to_string(),
+        }),
```
and-rewsmith, line 93: This needs to be in a Box because commands is a hashmap where values are implementors of trait Command, which does not have a determined size?

mqtt/mqtt-broker/src/lib.rs
```diff
@@ -239,7 +240,7 @@ pub enum SystemEvent {
     Shutdown,
     StateSnapshot(StateSnapshotHandle),
     ForceClientDisconnect(ClientId),
-    // ConfigUpdate,
+    AuthorizationUpdate(Box<dyn Any + Send + Sync>),
```
and-rewsmith, line 243: Why does it need to be a box of send and sync?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -172,6 +175,14 @@ impl Authorizer for EdgeHubAuthorizer {
...
         Ok(auth)
     }
+
+    fn update(&self, update: Box<dyn Any>) -> Result<(), Self::Error> {
+        let a = update.as_ref();
+        if let Some(identities) = a.downcast_ref::<ServiceIdentity>() {
+            debug!("{:?}", identities);
+        }
+        Ok(())
+    }
```
and-rewsmith, line 185: is this is a dummy impl just for now?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -172,6 +175,14 @@ impl Authorizer for EdgeHubAuthorizer {
 
         Ok(auth)
     }
+
+    fn update(&self, update: Box<dyn Any>) -> Result<(), Self::Error> {
+        let a = update.as_ref();
```
and-rewsmith, line 180: Why ```a```?

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -1,200 +1,116 @@
...
     #[error("failed to parse client id from message payload")]
     ParseClientId(#[from] SerdeError),
 
+    #[error("failed while sending authorized identities to broker")]
+    SendAuthorizedIdentitiesToBroker(Error),
+
     #[error("failed sending broker signal to disconnect client")]
-    SignalError(#[from] Error),
+    DisconnectSignal(Error),
 }
```
and-rewsmith, line 116: So it is better to have this HandleEventError enum for all command handler errors, or should we have a separate Error enum for every command handler?

and-rewsmith (commented):
LGTM mostly with some questions

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -1,200 +1,116 @@
...
-                if let SubscriptionUpdateEvent::Subscribe(sub) = subscription {
-                    subacks.remove(&sub.topic_filter);
-                }
-            }
-
-            if subacks.is_empty() {
-                debug!("command handler successfully subscribed to disconnect topic");
-                return Ok(());
-            }
+    fn handle(
```
and-rewsmith, line 63: Should we add some logging?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
dylanbronson: yep. Next PR will fill it in

mqtt/mqtt-edgehub/src/command.rs
dylanbronson: What's the standard here? Just alphabetical?

mqtt/mqtt-edgehub/src/command.rs
dylanbronson: That is correct. Talked with Denis offline. It's ok for now since these Commands won't be a hot path. If they need to be in the future, we can change it then

mqtt/mqtt-edgehub/src/command.rs
dylanbronson: We need to have a common error that the error handlers return, since the Command Trait defines an error to return

mqtt/mqtt-broker/src/lib.rs
dylanbronson: Send and Sync need to be implemented so this box can be shared across threads, which it is as a SystemEvent

mqtt/mqtt-edgehub/src/command.rs
vadim-kovalyov: @dylanrobson, we follow this approach:
- std
- 3rd party crates (tokio, etc)
-1st party crates (mqtt_broker, etc)
-local (crate::)
-mod declaration (if any)

a new line should separate each group.

mqtt/mqtt-edgehub/src/command.rs
```diff
@@ -1,200 +1,178 @@
...
-#[derive(Debug, thiserror::Error)]
-enum HandleDisconnectError {
-    #[error("failed to parse client id from message payload")]
-    ParseClientId(#[from] SerdeError),
+#[cfg(test)]
+mod tests {
+    use crate::command::ServiceIdentity;
+
+    #[test]
+    fn deserialize_broker_service_identity() {
```
vadim-kovalyov, line 139: to be honest I'm not sure if these tests have value. You are basically testing serde_json, which is 3rd party crate and already well-tested. What you could do, is to add integration tests (similar to what we have for disconnect command and topic translation). The test case could be:
- Create a broker
- Send an update to a auth command with one or few service identities
- Connect client with this identity (should be successful)
- Send an update where indentity is no longer there
- Expected: Client should be disconnected.

You can add this test at the end, when you implement system event handler in the broker

mqtt/mqtt-edgehub/src/command_handler.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+
+async fn subscribe(
+    client: &mut mqtt3::Client<BrokerConnection>,
+    topics: Vec<String>,
+) -> Result<(), CommandHandlerError> {
+    let subscriptions = topics.iter().map(|topic| proto::SubscribeTo {
+        topic_filter: topic.to_string(),
+        qos: proto::QoS::AtLeastOnce,
```
vadim-kovalyov, line 128: nit: remove FQN

mqtt/mqtt-edgehub/src/command.rs
dylanbronson: Yes makes sense. At least they were helpful to prove to myself that the serde works as expected :)

vadim-kovalyov (approved):
Looks great!

PR #3475, opened 2020-09-01
Policy engine core

vadim-kovalyov:
**Context:**
There are many use cases in the IoT Platform (IoT Edge and IoT Hub) that require some sort of policy engine to evaluate policies against certain input and state of the system. The main example is authorization for MQTT Broker.  

**This PR:**
Adds generic json-based authorization policy engine. Policy uses simple Identity->Operation->Resource approach to define authorization rules. Below is an example of a policy definition consisting of two rules.

```
{
            "schemaVersion": "2020-10-30",
            "statements": [
                {
                    "effect": "deny",
                    "identities": [
                        "actor_a"
                    ],
                    "operations": [
                        "write"
                    ],
                    "resources": [
                        "resource_1"
                    ]
                },
                {
                    "effect": "allow",
                    "identities": [
                        "actor_b"
                    ],
                    "operations": [
                        "read"
                    ],
                    "resources": [
                        "resource_1"
                    ]
                }
            ]
        }
```

**Open questions:**
- Exact location of the policy core code.

**Next steps:**
- Integration w/ MQTT Broker. Broker specific substitutor, validator, resource marcher implementations. (separate PRs)


mqtt/policy/src/core/builder.rs
```diff
@@ -0,0 +1,706 @@
...
+    lazy_static! {
+        static ref VAR_PATTERN: Regex =
+            Regex::new(r#"\{\{[^\{\}]+\}\}"#).expect("failed to create a Regex from pattern");
+    }
+    VAR_PATTERN.is_match(value)
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct PolicyDefinition20201030 {
```
dylanbronson, line 249: How often do we see this naming scheme changing? 
Are Statement/Effect/PolicyDefinition tied together?

mqtt/policy/src/core/builder.rs
```diff
@@ -0,0 +1,706 @@
...
+    fn into(self) -> EffectOrd {
+        match self.effect {
+            Effect20201030::Allow => EffectOrd::new(CoreEffect::Allow, self.order),
+            Effect20201030::Deny => EffectOrd::new(CoreEffect::Deny, self.order),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
dylanbronson, line 284: Why not put the tests in a separate file?

vadim-kovalyov: it is rust practice to keep unit tests in the same file, as a submodule.

You can add tests in a separate file - but those will be integration tests for the crate.

mqtt/policy/src/core/builder.rs
vadim-kovalyov: >How often do we see this naming scheme updating?

I'm not sure how often that would be, but we need to think about versioning.

>Are Statement/Effect/PolicyDefinition tied together?

yes, in current policy definition schema.

mqtt/policy/src/core/mod.rs
```diff
@@ -0,0 +1,576 @@
...
+    }
+}
+
+impl From<BTreeMap<String, EffectOrd>> for Resources {
+    fn from(map: BTreeMap<String, EffectOrd>) -> Self {
+        Resources(map)
+    }
+}
+
+/// Represents a request that needs to be `evaluat`ed by `Policy` engine.
```
dylanbronson, line 224: nit: tick mark

mqtt/policy/src/errors.rs
```diff
@@ -0,0 +1,21 @@
+use thiserror::Error;
+
+/// A specialized `Result` type for mini-redis operations.
```
dmolokanov, line 3: _mini-redis_  :thinking:

mqtt/policy/Cargo.lock
```diff
@@ -0,0 +1,165 @@
+# This file is automatically @generated by Cargo.
```
dmolokanov, line 1: I think it should be removed from the workspace

dmolokanov (approved):
Great work! I like the way code is structure. Super neat!

mqtt/policy/src/errors.rs
vadim-kovalyov: Oops.. ^-^

dmolokanov (dismissed):
:shipit:

dmolokanov (approved):
🚀

PR #3477, opened 2020-09-01
[mqtt] In-Memory Persistence

and-rewsmith:
## Context
This PR adds in-memory persistence for the bridge. This implementation does not address ttl or priority. The way iothub telemetry will be sent to the bridge means that we might not need to handle priority / ttl explicitly in the bridge. In the spec we referred to this component as a queue, however I have avoided that name because we thought it added confusion. What we are adding in this PR isn't a queue as it supports removing elements by key.

Each bridge will own two persistent collections (upstream and downstream).  A bridge entity will interact with each collection from both ends. It will insert incoming publications into the collection, and read messages out to be published to their destinations.

## Main Sections
Here are the main components in this PR. I have added code comments above each one for further clarity.

**Persist Interface**: Defines operations for both the in-memory (contained in this PR) and disk persistence (will be added later)

**InMemoryPersist Struct**: In-memory implementation of the ```Persist``` trait. This does not handle priority or ttl.

**Loader Struct**: Responsible for loading elements from the ```InMemoryPersist```. It implements the stream trait to simplify the logic in the bridge.

**WakingMap Struct**: Non-exposed implementation struct that serves as the state for the ```InMemoryPersist```. An instance of this struct will be shared between the ```InMemoryPersist``` and the ```Loader```. When streams are waiting for new elements, it can wake them up whenever a new element is inserted into the collection.


mqtt/mqtt-bridge/Cargo.toml
```diff
@@ -5,14 +5,26 @@ authors = ["Azure IoT Edge Devs"]
 edition = "2018"
 
 [dependencies]
+async-std = "1.6"
```
vadim-kovalyov, line 8: where is `async-std` used?

mqtt/mqtt-bridge/src/persist/memory/loader.rs
```diff
@@ -0,0 +1,431 @@
...
+
+use crate::persist::{memory::waking_map::WakingMap, Key};
+
+// Message loader used to extract elements from bridge queues
+// This component is responsible for message extraction from the queue
+// It works by grabbing a snapshot of the most important messages from the queue
+// Then, will return these elements in order
+// When the batch is exhausted it will grab a new batch
+// If no elements are removed from the queue, the second batch will be identical to first
+pub struct InMemoryMessageLoader {
```
vadim-kovalyov, line 15: thanks for the code docs, they really help. Small nit: use /// for [code docs](https://doc.rust-lang.org/stable/rust-by-example/meta/doc.html).

mqtt/mqtt-bridge/src/persist/memory/loader.rs
```diff
@@ -0,0 +1,431 @@
+use std::{pin::Pin, sync::Arc, task::Context, task::Poll, vec::IntoIter};
+
+use futures_util::stream::Stream;
+use mqtt3::proto::Publication;
+use parking_lot::{Mutex, MutexGuard};
```
vadim-kovalyov, line 5: q: what's the benefit of `parking_lot`? if they are better than std, should we use them everywhere? /cc @dmolokanov

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn new() -> Self;
+
+    async fn insert(
+        &mut self,
+        priority: u32,
+        ttl: Duration,
+        message: Publication,
+    ) -> Result<Key, QueueError>;
+
+    async fn remove(&mut self, key: Key) -> Result<bool, QueueError>;
```
vadim-kovalyov, line 25: nit suggestion on the interface. Why not follow what std collections usually do for remove methods
```
 async fn remove(&mut self, key: Key) -> Option<Publication>;
 ```

File: mqtt/mqtt-bridge/src/persist/mod.rs [hunk 5]
```diff
@@ -0,0 +1,113 @@
...
+mod memory;
+
+// Queue used in bridge.
+#[async_trait]
+trait Persist<'a> {
+    type Loader: Stream;
+
+    fn new() -> Self;
+
+    async fn insert(
```
vadim-kovalyov, line 18: nit: `push` instead `insert`?

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+            priority: 1,
+            offset: 0,
+            ttl: Duration::from_secs(5),
+        };
+        assert_eq!(key1 > key2, false);
+        assert_eq!(key1 < key2, false);
+        assert!(key1 != key2);
+    }
+
+    // TODO REVIEW: Does this guarantee that btreemap is not ordered by ttl? Why this weird behavior with equals?
```
vadim-kovalyov, line 95: It is indeed weird behaviour here, when all key1>key2, key1<key2 and key1==key2 are all false. But it makes sense from the implementation of PartialOrd and PartialEq for Key. So, I believe, in BTreeMap the keys will be ordered by offset only (desired behaviour), but at the same time, keys with the same offset but different pri or ttl are not the same keys. It would be good to verify that with a test

vadim-kovalyov (approved):
LGTM, some minor suggestions on the Persist trait.

mqtt/mqtt-bridge/src/persist/memory/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+        let mut state_lock = self.state.lock();
+        state_lock
+            .remove_in_flight(&key)
+            .map_err(|_| PersistError::Removal())?;
+        Ok(true)
+    }
+
+    async fn loader(&'a mut self, batch_size: usize) -> InMemoryMessageLoader {
+        InMemoryMessageLoader::new(Arc::clone(&self.state), batch_size).await
+    }
```
and-rewsmith, line 54: Discussed with Anca and we think that it is best not to allow multiple separate instances of loader. These will share the same underlying state and will break the waking functionality. I am going to make loader a struct field that get's initialized on InMemoryPersist::new()

mqtt/mqtt-bridge/src/persist/memory/loader.rs
and-rewsmith: Parking lot has fairness enforcement. Denis recommended it over the mutex from std

mqtt/mqtt-bridge/Cargo.toml
and-rewsmith: I accidentally added. Now removed

and-rewsmith:
I saw an issue with our loader design considering the persist remove method, in that it is primarily used for removing in-flight messages once a puback is received. 
 
We made this loader get a snapshot and return in batches. We intentionally made it possible for the loader to return a duplicate message if this message is not removed from the persist. If the batch size is small enough and the puback takes long enough to come back, it won't get removed from the persist. Then we could get the same message from the loader and try to send this duplicate message again.
 
The logic behind the current design was that if the message didn't get handled properly then it would be returned again from the loader. It seems like we will need to not return duplicates to avoid this duplicate message send scenario.

Here is my plan.
	
**In memory**
- Have in-memory queue and an in-memory in-flight storage
- Elements will be removed from the queue and added to this in-flight storage when they are returned by the loader
- Now with this in-flight collection, we might want to do something. Log?
	
	
**Disk**
- Have a BTreeMap and an in-memory in-flight collection. We don't want to persist twice to disk.
- Elements not removed from BTreeMap. They are added to an in-flight collection when they are returned by the loader.
- When the next batch is retrieved the in-flight messages will be ignored and it will fill a full batch of non-duplicate messages
- In case of a crash, the in-flight messages shouldn't get an ack back, so it is logical that we shouldn't care that this in-memory collection of in-flight messages is lost. 
- When the bridge comes back up it will send all previously in-flight messages again.
- Now with this in-flight collection, we might want to do something. Log?
 
This solves the duplicate problem which is good. But the problem is for the in-memory case, if we don't add the in-flight messages back to the main queue, then they are essentially removed when read by loader. I am thinking now that this is fine since:
- the caller of the loader should keep retrying to send extracted messages
- in case of a crash, everything is lost anyway so we don't care about lost in-flight messages 

Let me know if you have any concerns with my approach for in-memory and disk implementations. I have already converted the implementation in this PR to the in-memory implementation outlined above.

mqtt/mqtt-bridge/src/persist/mod.rs
and-rewsmith: I talked to Varun and at this point we aren't thinking of removing priority / ttl in the bridge. Denis and I decided to remove it from the interface, key struct, etc.

mqtt/mqtt-bridge/src/persist/memory/loader.rs
```diff
@@ -0,0 +1,337 @@
...
+        let poll_stream = async move {
+            let maybe_extracted = loader.next().await;
+            if let Some(extracted) = maybe_extracted {
+                assert_eq!((key_copy, pub_copy), extracted);
+            }
+        };
+
+        // start the function and make sure it starts polling the stream before next step
+        let poll_stream_handle = tokio::spawn(poll_stream);
+        time::delay_for(Duration::from_secs(2)).await;
```
and-rewsmith, line 329: I don't think we can use notify here because we would need to notify in the stream implementation

mqtt/mqtt-bridge/src/persist/memory/waking_map.rs
```diff
@@ -0,0 +1,225 @@
...
+        self.waker = Some(waker.clone());
+    }
+
+    // TODO REVIEW: exposed just for testing.
+    // We don't use the in-flight now, but we need to find a way to use it.
+    // Otherwise just have the queue.
+    #[allow(dead_code)]
+    fn in_flight(&mut self) -> &HashMap<Key, Publication> {
+        &self.in_flight
+    }
```
and-rewsmith, line 74: The in-flight messages for the in-memory persistence implementation don't have a clear purpose. 

For the disk case (not this PR) it makes sense. For disk case, messages cannot be removed from the main collection right away because in case of crash we want them to be resent. Thus a separate in-flight collection is needed to prevent duplicates from being returned from the loader. When an element is removed form the in-flight it can be removed from the main persistence.

For the in-mem case everything is lost in the case of a crash anyway, so the in-flight collection doesn't have the same purpose. Removing from the queue immediately on retrieval by the loader allows for a simple queue to be used. The messages can be removed from the main queue directly when retrieved by the loader, rather than linger around in the main queue until being removed from in-flight. I think having this collection is still necessary for logging, but we need to find an appropriate place to do that. 

If we cannot think of a place to log in-flight, then we can remove the in-flight collection from this in-memory implementation.

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+pub struct Key {
+    offset: u32,
+}
+
+// TODO REVIEW: these generic error types don't wrap impl specific types
+#[derive(Debug, Error)]
+pub enum PersistError {
+    #[error("Failed to remove messages from persistence")]
+    Removal(),
+}
```
and-rewsmith, line 34: What is the best practice here for handling impl specific errors?

and-rewsmith: I still have this question, but it isn't relevant anymore. In handling another comment, I converted the remove return type to option. Now we don't need this.

mqtt/mqtt-bridge/src/persist/memory/mod.rs
and-rewsmith: I now make loader inside the new() and wrap in Arc<Mutex<>>

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+/// Persistence used in bridge.
+/// Elements are added, then can be removed once retrieved by the loader
+/// If one attempts to remove added elements before reading via the loader remove returns None
+#[async_trait]
+trait Persist<'a> {
+    type Loader: Stream;
+
+    async fn new(batch_size: usize) -> Self;
+
+    async fn push(&mut self, message: Publication) -> Result<Key, Box<dyn Error>>;
```
and-rewsmith, line 20: Is the correct way to return a generic error type?

mqtt/mqtt-bridge/src/persist/memory/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+#[async_trait]
+impl<'a> Persist<'a> for InMemoryPersist {
+    type Loader = InMemoryMessageLoader;
+
+    async fn new(batch_size: usize) -> Self {
+        let state = WakingMap::new();
+        let state = Arc::new(Mutex::new(state));
+
+        let loader = InMemoryMessageLoader::new(Arc::clone(&state), batch_size).await;
+        let loader = Arc::new(Mutex::new(loader));
```
ancaantochi, line 31: is Mutex needed here because next is mutating the loader? It looks weird that it locks to iterate, but I don't know what is the solution.

mqtt/mqtt-bridge/src/persist/memory/waking_map.rs
```diff
@@ -0,0 +1,232 @@
...
+            if let Some(pair) = removed {
+                output.push(pair);
+            } else {
+                error!("failed retrieving message from persistence");
+                continue;
+            }
+        }
+
+        for element in output.iter() {
+            self.in_flight.insert(element.0.clone(), element.1.clone());
```
ancaantochi, line 52: can the insert be after line 44 and remove this loop?

mqtt/mqtt-bridge/src/persist/mod.rs
ancaantochi: I think it's adding a new type: type Error; 
then return Result<Key, Self::Error>

mqtt/mqtt-bridge/src/persist/memory/mod.rs
and-rewsmith: I had the same concern and was going to ask Denis. @dmolokanov Thoughts?

and-rewsmith:
I opened a new PR into master: #3520

PR #3478, opened 2020-09-02
Update ARM32 base image to 2.1.21-bionic-arm32v7

yophilav:
Update ARM32 base image to 2.1.21-bionic-arm32v7 to fix security vulnerability relates to glibc.2.27-3ubuntu1.2 (https://launchpad.net/ubuntu/+source/glibc/2.27-3ubuntu1.2).

### DIFF
--- a/edge-modules/TestAnalyzer/docker/linux/arm64v8/Dockerfile
+++ b/edge-modules/TestAnalyzer/docker/linux/arm64v8/Dockerfile
@@ -1,5 +1,5 @@
 ARG base_tag=1.0.4-linux-arm64v8
-FROM edgebuilds.azurecr.io/microsoft/azureiotedge-module-base-rocksdb:${base_tag}
+FROM azureiotedge/azureiotedge-module-base-rocksdb:${base_tag}
 
 ARG EXE_DIR=.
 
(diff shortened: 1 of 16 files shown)


edge-modules/TestAnalyzer/docker/linux/arm64v8/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG base_tag=1.0.4-linux-arm64v8
```
philipktlin, line 1: update base tag?

edge-modules/TwinTester/docker/linux/arm64v8/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG base_tag=1.0.4-linux-arm64v8
```
philipktlin, line 1: update base tag?

philipktlin:
number of updated files is less than what we did before in master or 1.0.10 branch because 1.0.9 branch doesn't contain those test projects, is it right?

edge-modules/TestAnalyzer/docker/linux/arm64v8/Dockerfile
yophilav: Nope. We don't publish ARM64 since it's using Dotnet3.0.

edge-modules/TwinTester/docker/linux/arm64v8/Dockerfile
yophilav: Again, nope we stop supporting ARM64.

philipktlin: We need to change as the old base image is deleted.

PR #3479, opened 2020-09-02
Adding unit tests for BrokerServiceIdentity

dylanbronson:
Adding serde tests and a compare unit tests for BrokerServiceIdentity.


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/BrokerServiceIdentityTest.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        [Fact]
+        public void SerializeTest()
+        {
+            string id = "id";
+            string authChain = "authChain";
+            BrokerServiceIdentity brokerServiceIdentity = new BrokerServiceIdentity(id, Option.Some(authChain));
+            var s = JsonConvert.SerializeObject(brokerServiceIdentity);
+            Assert.Equal($"{{\"Identity\":\"{id}\",\"AuthChain\":\"{authChain}\"}}", s);
+        }
```
and-rewsmith, line 17: Assuming we don't need to test serialization without an auth chain, because it shouldn't error if this passes

and-rewsmith (approved):

PR #3486, opened 2020-09-02
Increase long haul log size for edgeAgent and edgeHub

yophilav:
Increase long haul log size for edgeAgent and edgeHub so that they can handle log up to 24 hr in case of `debug` logging

### DIFF
--- a/e2e_deployment_files/long_haul_deployment.template.json
+++ b/e2e_deployment_files/long_haul_deployment.template.json
@@ -22,7 +22,7 @@
             "type": "docker",
             "settings": {
               "image": "<Container_Registry>/microsoft/azureiotedge-agent:<Build.BuildNumber>-linux-<Architecture>",
-              "createOptions": "{\"HostConfig\":{\"LogConfig\":{\"Type\":\"json-file\",\"Config\":{\"max-size\":\"1230m\",\"max-file\":\"1\"}}}}"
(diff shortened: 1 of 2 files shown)


e2e_deployment_files/long_haul_deployment.template.json
```diff
@@ -22,7 +22,7 @@
             "type": "docker",
             "settings": {
               "image": "<Container_Registry>/microsoft/azureiotedge-agent:<Build.BuildNumber>-linux-<Architecture>",
-              "createOptions": "{\"HostConfig\":{\"LogConfig\":{\"Type\":\"json-file\",\"Config\":{\"max-size\":\"4m\",\"max-file\":\"7\"}}}}"
+              "createOptions": "{\"HostConfig\":{\"LogConfig\":{\"Type\":\"json-file\",\"Config\":{\"max-size\":\"1230m\",\"max-file\":\"1\"}}}}"
```
philipktlin, line 25: First max-file default value is 1.  If log is full, it will remove the old file.  The max-size is the maximum file size of each log file, right? if so, I think at least half the max-size and make max-file to 2.  And how do you come up 1230M?

yophilav: the 1230M threshold comes from the observation of the log size for LH given release.1.0.10-rc2. 
Ok, I can make the log size half, but 2 files.

philipktlin (approved):
Do you plan to do it for master as well?

PR #3495, opened 2020-09-03
Remove module user from diagnostics dockerfile

lfitchett:
This is a regression from the iotedge check proxy changes. Previously it ran as root since it accesses the mgmt uri. This change sets it as root again


arsing (approved):

CindyXing (approved):

lfitchett:
Just tested it, it is able to access mgmt api

PR #3499, opened 2020-09-03
Update schema version for other failing tests

richma-ms:
Looks like I missed a couple the first time around.

### DIFF
--- a/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/EdgeAgentConnectionTest.cs
+++ b/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/EdgeAgentConnectionTest.cs
@@ -1934,7 +1934,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test
         {
             var desiredProperties = new
             {
-                schemaVersion = "1.0",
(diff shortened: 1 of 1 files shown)


edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/EdgeAgentConnectionTest.cs
```diff
@@ -2001,7 +2001,7 @@ static object GetEdgeAgentBaseConfiguration()
         {
             var desiredProperties = new
             {
-                schemaVersion = "1.0",
+                schemaVersion = "1.1.0",
```
yophilav, line 2004: Do we need to concern for the following tests as well? 
- https://github.com/Azure/iotedge/blob/master/edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/reporters/IoTHubReporterTest.cs#L1433
-

richma-ms: Those tests seem to be just verifying that the roundtrip payload comes back as expected, and not doing any actual schema verification.

yophilav (approved):

PR #3502, opened 2020-09-03
Removed iotedge-diagnostics package (fix bad master merge)

vadim-kovalyov:
Looks like we had a bad merge from master to iiot. This is to fix it and bring iotedged checkin pipeline back to green.


dmolokanov (approved):
Looks good. Other than changing upstream->iothub. To support the nested edge we use "upstream" term, but those changes seem the opposite. @lfitchett ?

PR #3505, opened 2020-09-04
Environment variable in image name

huguesBouvier:
Edge Agent can now dereference env var in image name.
For example:
$upstream:8000/azureiotedge-hub:1.0.10-private-preview
will be evaluated as
parenthostname:8000/azureiotedge-hub:1.0.10-private-preview


edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/DockerConfigTest.cs
```diff
@@ -176,7 +199,7 @@ public void TestSerializationDigest()
 
             Assert.Equal(expected, json);
         }
-
+        /*
```
philipktlin, line 202: why comment?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/DockerConfigTest.cs
```diff
@@ -36,6 +36,29 @@ public class DockerConfigTest
         static readonly DockerConfig Config16 = new DockerConfig("image1:42", @"{""Env"": [""k1=v1"", ""k2=v2"", ""k3=v3""], ""HostConfig"": {""PortBindings"": {""43/udp"": [{""HostPort"": ""43""}], ""42/tcp"": [{""HostPort"": ""42""}]}}}", Option.Some("4562124545"));
         static readonly DockerConfig ConfigUnknown = new DockerConfig("unknown");
         static readonly DockerConfig ConfigUnknownExpected = new DockerConfig("unknown:latest");
+        internal class MockEnvironment: DockerConfig.IEnvironmentWrapper
```
philipktlin, line 39: General comment: please build solution with release config in order to find out any code style errors.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -15,10 +15,24 @@ public class DockerConfig : IEquatable<DockerConfig>
     {
         // This is not the actual docker image regex, but a less strict version.
         const string ImageRegexPattern = @"^(?<repo>([^/]*/)*)(?<image>[^/:]+)(?<tag>:[^/:]+)?$";
+        const string ImageUpstreamRegexPattern = @"^\$(?<host>.*?)(?<path>:.*)";
```
philipktlin, line 18: I don't have context here, would you please share the design spec?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/DockerConfigTest.cs
huguesBouvier: mistake... fixed!

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
huguesBouvier: Ok, I'll add a comment too.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +100,29 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
...
         internal static string ValidateAndGetImage(string image)
+        {
+            return ValidateAndGetImage(image, new EnvironmentWrapper());
+        }
+
+        internal static string ValidateAndGetImage(string image, IEnvironmentWrapper env)
         {
             image = Preconditions.CheckNonWhiteSpace(image, nameof(image)).Trim();
+
+            Match match_host = ImageUpstreamRegex.Match(image);
```
philipktlin, line 111: I think you used rust code style in C#, e.g match_host should be matchHost.  Normally we don't use _ character in variable name.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +100,29 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
         }
 
         internal static string ValidateAndGetImage(string image)
+        {
+            return ValidateAndGetImage(image, new EnvironmentWrapper());
+        }
+
+        internal static string ValidateAndGetImage(string image, IEnvironmentWrapper env)
```
philipktlin, line 107: check null for IEnvironmentWrapper

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +100,29 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
...
         {
             image = Preconditions.CheckNonWhiteSpace(image, nameof(image)).Trim();
+
+            Match match_host = ImageUpstreamRegex.Match(image);
+            if ((match_host.Success)
+                && (match_host.Groups["host"]?.Length > 0)
```
philipktlin, line 113: you don't need to check length if use "+" in regex instead of "*".

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +100,29 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
...
+                && (match_host.Groups["path"]?.Length > 0))
+            {
+                String host_address = env.GetVariable(match_host.Groups["host"].Value);
+
+                if (host_address != null)
```
philipktlin, line 118: if it is null, you don't change image variable and let it continue to original code path.  shouldn't we return fail immediately?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
philipktlin, line 18: I need to understand more here.  It seems what host you are going to match is same as repo.  is it possible to combine regex into 1?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
philipktlin: you need to check null or white space

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -15,10 +15,24 @@ public class DockerConfig : IEquatable<DockerConfig>
     {
         // This is not the actual docker image regex, but a less strict version.
         const string ImageRegexPattern = @"^(?<repo>([^/]*/)*)(?<image>[^/:]+)(?<tag>:[^/:]+)?$";
```
philipktlin, line 17: sidetrack: why we need to split into 3 parts to validate the image address.  It is better to split into 2 parts - first part is full URL and 2nd part is image tag.  Then first part can be validated from Uri.IsWellFormedUriString.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
philipktlin: I think you need to handle the case if image address don't have any port mentioned, e.g. $IOTEDGEGATEWAY_HOSTNAME/comp/ea:tag1.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
huguesBouvier: I agree. Repo is not used, however I am in favor of not touching since any change introduces a risk.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
huguesBouvier: Yes, I think I can combine them. I'll look into it. I was not so much in favor because it is a different logic.
By design a port has to be specified. If a port is missing we dont do anything.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
huguesBouvier: I'll add a check

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
huguesBouvier: After giving it some thought I think it is less confusing to separate them. The regex is hard to read has it is already. It is less error prone to keep the regex simple and focused on one task.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Docker.Test/DockerConfigTest.cs
huguesBouvier: Thank I didn't know. I fixed it.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
huguesBouvier: I realized :). Thanks

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
huguesBouvier: Fixed!

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -15,10 +15,27 @@ public class DockerConfig : IEquatable<DockerConfig>
...
+        // Match for an environment variable in the official POSIX format $ + "letter" + "letter|digit"
+        // Then the port has to be specified with ":"
+        const string ImageUpstreamRegexPattern = @"^\$(?<host>[a-zA-Z_][a-zA-Z0-9_]*?)(?<path>:.*)";
 
         static readonly Regex ImageRegex = new Regex(ImageRegexPattern);
+        static readonly Regex ImageUpstreamRegex = new Regex(ImageUpstreamRegexPattern);
         readonly CreateContainerParameters createOptions;
 
+        internal interface IEnvironmentWrapper
```
philipktlin, line 26: nit: move to its own file for IEnvrionmentWrapper and EnvironmentWrapper.  Or at least move out from DockerConfig class.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +103,39 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
...
         internal static string ValidateAndGetImage(string image)
+        {
+            return ValidateAndGetImage(image, new EnvironmentWrapper());
+        }
+
+        internal static string ValidateAndGetImage(string image, IEnvironmentWrapper env)
         {
             image = Preconditions.CheckNonWhiteSpace(image, nameof(image)).Trim();
+
+            if (image.Substring(0, 1) == "$")
```
philipktlin, line 114:
```suggestion
            if (image[0] == '$')
```

philipktlin: check image is not null or empty.

philipktlin: As agreed during discussion, we are going to support $upstream only.  You should check if image starts with $upstream and replace it.

huguesBouvier: 1. Fixed
2. That is what I did. I separate, this way I can catch when an image is not correctly formated, Iike typo.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -17,25 +17,13 @@ public class DockerConfig : IEquatable<DockerConfig>
         const string ImageRegexPattern = @"^(?<repo>([^/]*/)*)(?<image>[^/:]+)(?<tag>:[^/:]+)?$";
         // Match for an environment variable in the official POSIX format $ + "letter" + "letter|digit"
         // Then the port has to be specified with ":"
-        const string ImageUpstreamRegexPattern = @"^\$(?<host>[a-zA-Z_][a-zA-Z0-9_]*?)(?<path>:.*)";
+        const string ImageUpstreamRegexPattern = @"^\$upstream(?<path>:[1-9].*)";
+        const string Gateway_hostname = "IOTEDGE_GATEWAYHOSTNAME";
```
philipktlin, line 21: use Constants.GatewayHostnameVariableName

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -111,22 +99,23 @@ internal static string ValidateAndGetImage(string image, IEnvironmentWrapper env
         {
             image = Preconditions.CheckNonWhiteSpace(image, nameof(image)).Trim();
 
-            if (image.Substring(0, 1) == "$")
+            if ((image.Length > 0) && (image[0] == '$'))
```
philipktlin, line 102: you don't need to check image.Length > 0 as Preconditions.CheckNonWhiteSpace already does.

philipktlin: I think you can just use image.StartsWith for "$upstream" with string comparison to ignore case.  https://docs.microsoft.com/en-us/dotnet/api/system.string.startswith?view=netcore-3.1#System_String_StartsWith_System_String_System_StringComparison_

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -111,22 +99,23 @@ internal static string ValidateAndGetImage(string image, IEnvironmentWrapper env
...
-                    string hostAddress = env.GetVariable(enVarName);
+                    string hostAddress = env.GetVariable(Gateway_hostname);
 
                     if (hostAddress != null)
                     {
                         image = hostAddress + matchHost.Groups["path"].Value;
                     }
+                    else
+                    {
+                        throw new ArgumentException($"Could not find environment variable: {Gateway_hostname}");
```
philipktlin, line 117: we don't have any guide to throw what exception when a system parameter is missing.  But Argument exception should not be used as its definition is "The exception that is thrown when one of the arguments provided to a method is not valid.".  And this is not a value passing to this method.  I would suggest to use InvalidOperationException, please confirm with Varun.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
philipktlin: I think it is better to support case-insensitive for $upstream.

huguesBouvier: I agree for the first two. Not the last part. I'll rather be pedantic.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -10,13 +10,18 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker
     using Newtonsoft.Json.Linq;
     using static System.FormattableString;
 
+
```
philipktlin, line 13: remove blank line.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -10,13 +10,18 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Docker
...
     [JsonConverter(typeof(DockerConfigJsonConverter))]
     public class DockerConfig : IEquatable<DockerConfig>
     {
         // This is not the actual docker image regex, but a less strict version.
         const string ImageRegexPattern = @"^(?<repo>([^/]*/)*)(?<image>[^/:]+)(?<tag>:[^/:]+)?$";
+        // Match for an environment variable in the official POSIX format $ + "letter" + "letter|digit"
+        // Then the port has to be specified with ":"
+        const string ImageUpstreamRegexPattern = @"^\$upstream(?<path>:[1-9].*)";
```
philipktlin, line 21: check if it's prefix is "$upstream" and replace with environment variable. Should support $upstream for any case.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +91,40 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
...
+                        image = hostAddress + matchHost.Groups["path"].Value;
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException($"Could not find environment variable: {Core.Constants.GatewayHostnameVariableName}");
+                    }
+                }
+                else
+                {
+                    throw new ArgumentException($"Image {image} is not in the right format. If your intention is to use an environment variable, check the port is specified.");
```
philipktlin, line 122: This should not be argument exception, which means invalid input from caller.  This should be more appropriate to throw InvalidOperationException.  Please confirm with @varunpuranik

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
philipktlin: should be easier to use simple string.startsWith.

huguesBouvier: I did it as explained above, I ll rather keep using image[0]

philipktlin: would you tell me the benefit using regex over startsWith method call in this case?

huguesBouvier: Does 2 things:
1. checks for a port
2. Can generate exception if user made typos in $upstream

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -15,8 +15,10 @@ public class DockerConfig : IEquatable<DockerConfig>
     {
         // This is not the actual docker image regex, but a less strict version.
         const string ImageRegexPattern = @"^(?<repo>([^/]*/)*)(?<image>[^/:]+)(?<tag>:[^/:]+)?$";
-
+        // Check if it's prefix is "$upstream" and replace with environment variable. Should support $upstream for any case.
```
philipktlin, line 18: as discussed, you want to enforce only $upstream will be supported.  Please update the comment here.

PR #3507, opened 2020-09-04
Allow release image pipeline to use a custom CR

yophilav:
Allow release image pipeline to use a custom CR

### DIFF
--- a/builds/misc/images-release.yaml
+++ b/builds/misc/images-release.yaml
@@ -13,8 +13,6 @@ jobs:
     variables:
       build.registry.address: edgebuilds.azurecr.io
       build.registry.user: EdgeBuilds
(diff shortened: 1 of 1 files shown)


builds/misc/images-release.yaml
```diff
@@ -13,8 +13,6 @@ jobs:
     variables:
       build.registry.address: edgebuilds.azurecr.io
       build.registry.user: EdgeBuilds
-      registry.address: edgerelease.azurecr.io
```
philipktlin, line 16: Please move these default values in pipeline for all these 4 variables.

yophilav: done

philipktlin (approved):
Thanks.  Please make sure these variables are defined in the pipeline with default values.

PR #3510, opened 2020-09-04
Davilu/component order

davilu:
Added logic to send notification to MQTT broker while EHC `AuthAgentProtocolHead` boot up;
Abstracted `AbstractNotificationHandler` for any component to send notification to MQTT broker

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/Microsoft.Azure.Devices.Edge.Agent.IoTHub.csproj
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/Microsoft.Azure.Devices.Edge.Agent.IoTHub.csproj
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.3-NestedEdge" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.4-NestedEdge" />
     <PackageReference Include="Microsoft.Azure.Devices.Client" Version="1.28.1-NestedEdge" />
   </ItemGroup>
 
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/Microsoft.Azure.Devices.Edge.Hub.Amqp.csproj
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/Microsoft.Azure.Devices.Edge.Hub.Amqp.csproj
@@ -9,7 +9,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Amqp" Version="2.4.4" />
-    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.3-NestedEdge" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.4-NestedEdge" />
   </ItemGroup>
 
   <ItemGroup>
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.csproj
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/Microsoft.Azure.Devices.Edge.Hub.CloudProxy.csproj
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.3-NestedEdge" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.4-NestedEdge" />
     <PackageReference Include="Stateless" Version="4.0.0" />
   </ItemGroup>
 
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/Microsoft.Azure.Devices.Edge.Hub.Http.csproj
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/Microsoft.Azure.Devices.Edge.Hub.Http.csproj
@@ -13,7 +13,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.3-NestedEdge" />
+    <PackageReference Include="Microsoft.Azure.Devices" Version="1.22.4-NestedEdge" />
   </ItemGroup>
 
   <ItemGroup>
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -88,8 +88,10 @@ where
                         }
                         SystemEvent::AuthorizationUpdate(update) => {
                             if let Err(e) = self.authorizer.update(update) {
-                                warn!(message = "an error occurred updating authorization info", error = %e);
+                                error!(message = "an error occurred while updating authorization info", error = %e);
+                                break;
                             }
+                            self.reevaluate_subscriptions();
                         }
                     }
                 }
@@ -98,6 +100,61 @@ where
 
         info!("broker is shutdown.");
         Ok(self.snapshot())
+    }
+
+    fn prepare_activities(client_id: &ClientId, session: &Session) -> Vec<(ClientId, Activity)> {
+        let mut activities = Vec::new();
+        for sub in session
+            .subscriptions()
+            .into_iter()
+            .flat_map(HashMap::values)
+        {
+            if let Ok(client_info) = session.client_info() {
+                let sub_topic_filter = sub.filter().to_string();
+                let operation = Operation::new_subscribe(proto::SubscribeTo {
+                    topic_filter: sub_topic_filter.clone(),
+                    qos: *sub.max_qos(),
+                });
+                activities.push((
+                    client_id.clone(),
+                    Activity::new(client_id.clone(), client_info.clone(), operation),
+                ));
+            }
+        }
+        activities
+    }
+
+    fn reevaluate_subscriptions(&mut self) {
+        let disconnecting: Vec<ClientId> = self
+            .sessions
+            .iter()
+            .flat_map(|(client_id, session)| Self::prepare_activities(client_id, session))
+            .filter_map(
+                |(client_id, activity)| match self.authorizer.authorize(activity) {
+                    Ok(Authorization::Allowed) => None,
+                    Ok(Authorization::Forbidden(reason)) => {
+                        debug!(
+                            "client {} not allowed to subscribe to topic. {}",
+                            client_id, reason
+                        );
+                        Some(client_id)
+                    }
+                    Err(e) => {
+                        warn!(message="error authorizing client subscription: {}", error = %e);
+                        Some(client_id)
+                    }
+                },
+            )
+            .collect();
+
+        for client_id in disconnecting {
+            if let Err(x) = self.process_drop_connection(&client_id) {
+                warn!(
+                    "error dropping connection for client {}. Reason {}",
+                    client_id, x
+                );
+            }
+        }
     }
 
     fn snapshot(&self) -> BrokerSnapshot {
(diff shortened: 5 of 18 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
```diff
@@ -121,6 +132,33 @@ public void Dispose()
...
+
+            return Task.FromResult(messages);
+        }
+
+        public override Task<IEnumerable<Message>> ConvertNotificationToMessagesAsync(bool notification)
+        {
+            IEnumerable<Message> messages = new[] { new Message(Topic, Payload) };
+            return Task.FromResult(messages);
+        }
+
```
vipeller, line 161: I think it is too much logic here in the protocol head class just to be able to handle both situations when the connector is connected or not connected. I feel that this should not be the concern of this class. Can it be implemented the way that this (and the ScopeIdentities) would not care about the connected state - they just call a 'SendMessageWheneverYouCanAsync()' method with the baked message (topic and binary content, just like connector.SendAsync(), and then the Connector would contain the logic that once it is connected, it sends the messages out? In that case we don't need these three methods here and in the ScopeIdentities, and both AuthAgentProtocolHead and ScopeIdentities would have a very narrow code to implement the notification. Also, we could get rid of the base class. 
The connector class could store these messages in a MqttPublishInfo it already uses (so we could get rid of the Message class introduced here), and once it gets connected, it fires out the collected messages.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
```diff
@@ -2,27 +2,34 @@
...
+        static readonly string Payload = "\"AuthAgentProtocolHead started.\"";
+
         readonly IAuthenticator authenticator;
         readonly IUsernameParser usernameParser;
         readonly IClientCredentialsFactory clientCredentialsFactory;
         readonly ISystemComponentIdProvider systemComponentIdProvider;
         readonly AuthAgentProtocolHeadConfig config;
         readonly object guard = new object();
 
+        bool _started;
```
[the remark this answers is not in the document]
davilu: I'm not quite understand your point. AFAIK, `StopAsync` will be called only when we bring down EHC. But if EHC is down, connection to MQTT broker is down. Even in future we allow restart, since `_started` is `false`, no message will be send until `StartAsync` was called again. Only gap is since `IMqttBrokerConnector` has no API to tell disconnection, if it connects and disconnect and after `StartAsync`, it will still try to send message through `IMqttBrokerConnector`.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
davilu: The 3 API is very clear: prepare what message need to be send. 

- `ConvertStoredNotificationsToMessagesAsync` is based on if `sendNotificationWhenConnected`, return the notification message.
- `ConvertNotificationToMessagesAsync`is return the notification message.
- `StoreNotificationAsync` is ask to store state while notification message can't be sent(connection to broker is not established).

varunpuranik:
@davilu - nit: can you pls add description on the changes included in this PR? 
Also, pls make sure to target it to master instead of iiot.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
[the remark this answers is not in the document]
davilu: Modified to always send out notification while connected if `AuthAgentProtocolHead` is started

davilu:
> @davilu - nit: can you pls add description on the changes included in this PR?
> Also, pls make sure to target it to master instead of iiot.

Modified target branch and added descriptions.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/NotificationHandler.cs
```diff
@@ -0,0 +1,162 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Hub.Core;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+
+    public class NotificationHandler<T>
```
varunpuranik, line 13: What is this for?

davilu: A tool class for any notification need to be sent to mqtt broker

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -342,6 +358,11 @@ async Task<Option<ServiceIdentity>> GetServiceIdentityInternal(string id)
 
         async Task HandleNoServiceIdentity(string id)
         {
+            if (this.lifeCycleSignal.IsCancellationRequested)
+            {
+                return;
+            }
```
richma-ms, line 364: seems like this should be consistent with HandleNewServiceIdentity, either we check cancellation in both, or neither.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -206,9 +213,11 @@ public void Dispose()
         {
             this.encryptedStore?.Dispose();
             this.refreshCacheTimer?.Dispose();
-            this.refreshCacheTask?.Dispose();
+            this.lifeCycleSignal.Cancel();
```
richma-ms, line 216: >this.lifeCycleSignal.Cancel(); [](start = 12, length = 30)

hmm...this just signals cancel but doesn't wait.  What happens when background task gets disposed now?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ScopeIdentitiesHandler.cs
```diff
@@ -20,97 +13,41 @@ public class ScopeIdentitiesHandler : IMessageProducer
...
-        async Task OnConnect()
+        async Task<IEnumerable<Message>> RetrieveDeviceScopeIdentitiesCacheAsync()
         {
-            using (await this.cacheUpdate.LockAsync())
-            {
-                var deviceScopeIdentitiesCache = await this.deviceScopeIdentitiesCacheGetter;
-                deviceScopeIdentitiesCache.ServiceIdentitiesUpdated +=
-                    async (sender, serviceIdentities) => await this.ServiceIdentitiesUpdated(serviceIdentities);
+            var deviceScopeIdentitiesCache = await this.deviceScopeIdentitiesCacheGetter;
+            deviceScopeIdentitiesCache.ServiceIdentitiesUpdated += async (sender, serviceIdentities) => await this.notificationHandler.NotifyAsync(serviceIdentities);
```
richma-ms, line 29: >ServiceIdentitiesUpdated [](start = 39, length = 24)

ServiceIdentityRemoved should also be hooked up (I think this was missing before as well)

richma-ms:
>             await Task.Delay(TimeSpan.FromSeconds(refreshDelaySec));

Can you add the +1 sec timer fix here as well? (it's basically the same test logic)

---
Refers to: edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/DeviceScopeIdentitiesCacheTest.cs:804 in 10fd0bd. [](commit_id = 10fd0bd006c2537fa9070856a5186a28a2e65ab5, deletion_comment = False)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
```diff
@@ -133,7 +139,10 @@ enum EventIds
                 Closing,
                 Closed,
                 ClosedWhenNotRunning,
-                StartedWhenAlreadyRunning
+                StartedWhenAlreadyRunning,
+                NotifyingBroker,
+                NotifiedBroker,
+                NotifyBrokerFailed
```
richma-ms, line 145: these don't seem used anymore

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
```diff
@@ -66,15 +68,15 @@ public async Task StartAsync()
...
+            this.startSignal.Set();
             Events.Started();
         }
 
         public async Task CloseAsync(CancellationToken token)
         {
             Events.Closing();
-
             Option<IWebHost> hostToStop;
+            this.startSignal.Reset();
```
richma-ms, line 79: >this.startSignal.Reset(); [](start = 12, length = 25)

this should probably come after the call to StopAsync, otherwise there's a window of inconsistency.  Not an issue right now but could lead to bugs in the future.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
davilu: The background task will finish by itself. It could wait `refreshCacheTask` finish here if would like to guarantee the task quit. But block `Dispose` is not a ggod idea.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
davilu: Good catch. Missed `HandleNewServiceIdentity`

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
davilu: Will do. Nice catch.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
davilu: Will remove them.

richma-ms (dismissed):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/NotificationHandler.cs
```diff
@@ -0,0 +1,152 @@
...
+                }
+            }
+            catch (Exception ex)
+            {
+                Events<T>.SendingNotificationError(message, ex);
+            }
+        }
+    }
+
+    public class Message
```
dylanbronson, line 111: Should we rename this BrokerMessage? Message is already overloaded in EdgeHub

davilu: done

PR #3515, opened 2020-09-06
ScopeIdentitiesHandler refactor

varunpuranik:
IDeviceScopeIdentitiesCache is registered as a Task, so ScopeIdentitiesHandler cannot get the instance in its ctor. 
So refactored the class to set up the callback in OnConnect instead. So the logic now is - 
- OnConnect - Get a lock, subscribe to Identities changes, Get latest identities and send them to broker
- On identities update - Get a lock, get the changes and send them to broker.


dylanbronson (approved):
LGTM

PR #3517, opened 2020-09-08
Update MQTT pipeline to run on master only

vadim-kovalyov:
After we merged IIOT back to master, we now can move some MQTT related pipelines to run from master only.


builds/misc/images-mqtt.yaml
```diff
@@ -2,8 +2,7 @@ trigger:
   batch: true
   branches:
     include:
-      - iiot
-      - preview/iiot
```
vadim-kovalyov, line 6: I think we don't need to run mqtt images from preview or release. Images only used for perf tests atm.

dmolokanov (approved):
LGTM

and-rewsmith:
I think we are missing removal for images and e2e.

and-rewsmith (changes requested):
LGTM once we remove the triggers for images and e2e

vadim-kovalyov:
> LGTM once we remove the triggers for images and e2e

@and-rewsmith In this PR I'm changing mqtt-specific pipelines only. I think we wanted to keep e2e and images as a safeguard in case iotedged changes break integration with edge-hub/mqtt broker. Or are you talking about something else?

and-rewsmith:
@vadim-kovalyov Ok that sounds good to me.

and-rewsmith (approved):

PR #3520, opened 2020-09-09
[mqtt] In-Memory Persistence for bridge

and-rewsmith:
## Context
This PR adds in-memory persistence for the bridge. This implementation does not address ttl or priority. The way iothub telemetry will be sent to the bridge means that we might not need to handle priority / ttl explicitly in the bridge. In the spec we referred to this component as a queue, however I have avoided that name because we thought it added confusion. What we are adding in this PR isn't a queue as it supports removing elements by key.

Each bridge will own two persistent collections (upstream and downstream).  A bridge entity will interact with each collection from both ends. It will insert incoming publications into the collection, and read messages out to be published to their destinations.

## Main Sections
Here are the main components in this PR. I have added code comments above each one for further clarity.

**Persist Interface**: Defines operations for both the in-memory (contained in this PR) and disk persistence (will be added later)

**InMemoryPersist Struct**: In-memory implementation of the ```Persist``` trait. This does not handle priority or ttl.

**Loader Struct**: Responsible for loading elements from the ```InMemoryPersist```. It implements the stream trait to simplify the logic in the bridge.

**WakingMap Struct**: Non-exposed implementation struct that serves as the state for the ```InMemoryPersist```. An instance of this struct will be shared between the ```InMemoryPersist``` and the ```Loader```. When streams are waiting for new elements, it can wake them up whenever a new element is inserted into the collection.


mqtt/mqtt-bridge/src/persist/memory/loader.rs
```diff
@@ -0,0 +1,328 @@
...
+        let poll_stream = async move {
+            let maybe_extracted = loader.next().await;
+            if let Some(extracted) = maybe_extracted {
+                assert_eq!((key_copy, pub_copy), extracted);
+            }
+        };
+
+        // start the function and make sure it starts polling the stream before next step
+        let poll_stream_handle = tokio::spawn(poll_stream);
+        time::delay_for(Duration::from_secs(2)).await;
```
and-rewsmith, line 320: I don't think we can use notify here because we would need to notify in the stream implementation

mqtt/mqtt-bridge/src/persist/memory/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+        debug!(
+            "removing publication with offset {} from in-flight collection",
+            self.offset
+        );
+
+        let mut state_lock = self.state.lock();
+        state_lock.remove_in_flight(&key)
+    }
+
+    async fn loader(&'a mut self) -> Arc<Mutex<InMemoryMessageLoader>> {
```
and-rewsmith, line 70: It looks weird that it locks to iterate, but I don't know what is the solution to avoid mutex here. @dmolokanov Thoughts?

mqtt/mqtt-bridge/src/persist/memory/waking_map.rs
```diff
@@ -0,0 +1,229 @@
...
+use tracing::error;
+
+use crate::persist::Key;
+
+/// Responsible for waking waiting streams when new elements are added
+/// Exposes a get method for retrieving a count of elements starting from queue head
+/// Once elements are retrieved they are added to the in flight collection
+pub struct WakingMap {
+    queue: VecDeque<(Key, Publication)>,
+    in_flight: HashMap<Key, Publication>,
```
and-rewsmith, line 13: The in-flight messages for the in-memory persistence implementation don't have a clear purpose.

For the disk case (not this PR) it makes sense. For disk case, messages cannot be removed from the main collection right away because in case of crash we want them to be resent. Thus a separate in-flight collection is needed to prevent duplicates from being returned from the loader. When an element is removed form the in-flight it can be removed from the main persistence.

For the in-mem case everything is lost in the case of a crash anyway, so the in-flight collection doesn't have the same purpose. Removing from the queue immediately on retrieval by the loader allows for a simple queue to be used. The messages can be removed from the main queue directly when retrieved by the loader, rather than linger around in the main queue until being removed from in-flight. I think having this in-flight collection is still necessary for logging, but we need to find an appropriate place to do that.

If we cannot think of a place to log in-flight, then we can remove the in-flight collection from this in-memory implementation.

mqtt/mqtt-bridge/src/persist/memory/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+struct InMemoryPersist {
+    state: Arc<Mutex<WakingMap>>,
+    offset: u32,
+    loader: Arc<Mutex<InMemoryMessageLoader>>,
+}
+
+#[async_trait]
+impl<'a> Persist<'a> for InMemoryPersist {
+    type Loader = InMemoryMessageLoader;
+    type Error = Error;
```
and-rewsmith, line 27: This error type isn't used, but I had to define some error type to satisfy the trait. Is this fine?

mqtt/mqtt-bridge/src/persist/memory/waking_map.rs
and-rewsmith: I now disagree with this. In-flight serves the purpose of only allowing elements to be removed from the collection which have been extracted by the loader component.

vadim-kovalyov (approved):
Great quality stuff! :rocket:

PR #3523, opened 2020-09-09
[mqtt] Mitigate bad startup order

and-rewsmith:
I made a mistake overlooking that hot tasks do not work the C# way. This leads us to a state where the broker hangs and doesn't open its ports. The tests don't catch this because our test-util startup is not implemented the same way as our actual startup. 

This PR serves as a mitigation of the issue. I have a bug open on me to fix the underlying issue which will involve changing the startup logic.

### DIFF
--- a/mqtt/mqttd/src/broker/bootstrap/edgehub.rs
+++ b/mqtt/mqttd/src/broker/bootstrap/edgehub.rs
@@ -139,6 +139,7 @@ where
     // Start serving new connections
     let state = server.serve(shutdown).await?;
 
+    // Shutdown the sidecars
     sidecar_shutdown_handle.shutdown()?;
     sidecar_join_handle.await??;
 
--- a/mqtt/mqttd/src/broker/bootstrap/edgehub.rs
+++ b/mqtt/mqttd/src/broker/bootstrap/edgehub.rs
@@ -164,7 +164,7 @@ async fn server_certificate_renewal(renew_at: DateTime<Utc>) {
 
 // TODO: We need to elevate failable init steps out of the async block running in it's own thread.
 //       This is because we need to have defined setup steps, that once complete, the sidecars can run without failing.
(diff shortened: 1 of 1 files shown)


mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -162,22 +161,31 @@ async fn server_certificate_renewal(renew_at: DateTime<Utc>) {
...
-    let device_id = env::var(DEVICE_ID_ENV)?;
-    let command_handler =
-        CommandHandler::new(broker_handle, system_address, device_id.as_str()).await?;
-    let command_handler_shutdown_handle = command_handler.shutdown_handle()?;
-
     let mut bridge_controller = BridgeController::new();
+    let bridge = bridge_controller.start();
+    bridge.await?;
+
+    let sidecars = tokio::spawn(async move {
```
and-rewsmith, line 182: I don't think event loop was a good name so I renamed to sidecars

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -189,15 +197,11 @@ async fn start_sidecars(
         if let Err(e) = command_handler_join_handle.await {
             error!(message = "failed waiting for command handler shutdown", error = %e);
         }
-        if let Err(e) = bridge.await {
-            error!(message = "failed waiting for bridge shutdown", error = %e);
-        }
```
and-rewsmith, line 194: This was not correct either so I moved it. Since it is not fail-able and doesn't depend on the broker I put it in the start_sidecars() method. It doesn't actually do anything yet though.

When it is implemented it will be structured the same way as the command handler. Ideally with fail-able init steps, then in a separate thread the sidecars will run and get shutdown.

vadim-kovalyov (approved):

vadim-kovalyov (approved):

PR #3528, opened 2020-09-09
Environment variable in image name

huguesBouvier:
Edge Agent can now dereference env var in image name.
For example:
$upstream:8000/azureiotedge-hub:1.0.10-private-preview
will be evaluated as
parenthostname:8000/azureiotedge-hub:1.0.10-private-preview


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +88,40 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
...
+                        image = hostAddress + matchHost.Groups["path"].Value;
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException($"Could not find environment variable: {Core.Constants.GatewayHostnameVariableName}");
+                    }
+                }
+                else
+                {
+                    throw new InvalidOperationException($"Image {image} is not in the right format. If your intention is to use an environment variable, check the port is specified.");
```
philipktlin, line 119: use ArgumentException.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +88,40 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
...
+        internal static string ValidateAndGetImage(string image, IEnvironmentWrapper env)
         {
             image = Preconditions.CheckNonWhiteSpace(image, nameof(image)).Trim();
+
+            if (image[0] == '$')
+            {
+                Match matchHost = ImageUpstreamRegex.Match(image);
+                if (matchHost.Success
+                    && (matchHost.Groups["path"]?.Length > 0)
+                    && (env != null))
```
richma-ms, line 104: >&& (env != null) [](start = 20, length = 16)

seems unnecessary for internal methods

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +88,40 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
...
         internal static string ValidateAndGetImage(string image)
+        {
+            return ValidateAndGetImage(image, new EnvironmentWrapper());
+        }
+
+        internal static string ValidateAndGetImage(string image, IEnvironmentWrapper env)
         {
             image = Preconditions.CheckNonWhiteSpace(image, nameof(image)).Trim();
+
+            if (image[0] == '$')
```
richma-ms, line 99: >if (image[0] == '$') [](start = 12, length = 20)

isn't this check already part of your regex expression?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -204,4 +237,17 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
     }
+
+    internal interface IEnvironmentWrapper
+    {
+        string GetVariable(string variableName);
+    }
+
+    internal class EnvironmentWrapper : IEnvironmentWrapper
+    {
+        public string GetVariable(string variableName)
```
richma-ms, line 248: >string [](start = 15, length = 6)

for correctness this should prob return an Option<string>, then the caller can either use one of Option methods to get around having to null-check

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
```diff
@@ -86,9 +88,40 @@ internal static CreateContainerParameters GetCreateOptions(string createOptions)
...
+            if (image[0] == '$')
+            {
+                Match matchHost = ImageUpstreamRegex.Match(image);
+                if (matchHost.Success
+                    && (matchHost.Groups["path"]?.Length > 0)
+                    && (env != null))
+                {
+                    string hostAddress = env.GetVariable(Core.Constants.GatewayHostnameVariableName);
+
+                    if (hostAddress != null)
```
richma-ms, line 108: >if (hostAddress != null) [](start = 20, length = 24)

see other comment below, this should be an Option instead of a null-check

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
huguesBouvier: I kinda agree but it was part of a previous comment: 

philipktlin 6 days ago • 
 Member
check null for env argument

It is defensive coding, I understand it too.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Docker/DockerConfig.cs
huguesBouvier: It allows flagging if user wanted to put $upstream but instead made a typo: $upsteam

huguesBouvier: if the user put $ then we assume they want to use an $upstream. We then check for sanity and create exception if wrong.

richma-ms (approved):

huguesBouvier:
test

PR #3529, opened 2020-09-09
Remove 500 tail on GetModuleLogs

lfitchett:
GetModuleLogs returns an error message on too large payload instead of using tail 500


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/ModuleLogsRequestHandler.cs
```diff
@@ -13,7 +13,9 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Core.Requests
 
     public class ModuleLogsRequestHandler : RequestHandlerBase<ModuleLogsRequest, IEnumerable<ModuleLogsResponse>>
     {
-        const int MaxTailValue = 500;
+        // Max size is 128 KB, leave 1KB buffer
+        const int MaxPayloadSize = 127000;
```
varunpuranik, line 17: Tests?

varunpuranik (dismissed):

CindyXing (dismissed):

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/ModuleLogsRequestHandler.cs
```diff
@@ -58,25 +61,26 @@ protected override async Task<Option<IEnumerable<ModuleLogsResponse>>> HandleReq
 
                     if (l.logOptions.ContentEncoding == LogsContentEncoding.Gzip)
                     {
-                        if (moduleLogs.Length > MaxPayloadSize)
-                        {
-                            return new ModuleLogsResponse(l.id, Events.LargePayload(moduleLogs.Length, l.logOptions));
-                        }
-
+                        messageSize += moduleLogs.Length;
```
damonbarry, line 64: There are potentially multiple module log requests being processed simultaneously here...is there a possibility that two threads will try to update `messageSize` at the same time and corrupt its value?

damonbarry: If so, we could sum up message sizes after `Task.WhenAll` below, rather than using a sync primitive...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/ModuleLogsRequestHandler.cs
```diff
@@ -108,10 +112,10 @@ public static void MismatchedMinorVersions(string payloadSchemaVersion, Version
...
+            public static string LargePayload(int size, IEnumerable<ModuleLogOptions> options)
             {
                 // TODO: make/get aka link for documentation
-                string message = $"The payload is too large for a direct method. {Name} supports up to {MaxPayloadSize} bytes of logs. The current request returned {size} bytes.\nTry reducing the size of the logs by setting the 'tail', 'since' and 'from' fields in log options filter. For more information, see [[aka.ms/TODO:LinkToDocumentation]].\nCurrent options settings are:\n{Newtonsoft.Json.JsonConvert.SerializeObject(options, Newtonsoft.Json.Formatting.Indented)}";
+                string message = $"The payload is too large for a direct method. {Name} supports up to {MaxPayloadSize} bytes of logs. The current request returned {size} bytes.\nTry reducing the size of the logs by setting the 'tail', 'since' and 'from' fields in log options filter. For more information, see https://aka.ms/iotedge-log-pull \nCurrent options settings are:\n{Newtonsoft.Json.JsonConvert.SerializeObject(options, Newtonsoft.Json.Formatting.Indented)}";
```
damonbarry, line 118: Is there _any_ possibility of this message exceeding the max payload size? 😃

lfitchett: extremely unlikely. The options sent have to be under the limit, so the options would have to be within a hundred chars of the max size. I'm not concerned.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/ModuleLogsRequestHandler.cs
lfitchett: I thought about using Interlocked.Add, but then I read that int read/write was atomic. Doing another quick on the internet, it looks like I misinterpreted that and read-modify write isn't atomic. Ill replace it.

damonbarry (approved):

PR #3532, opened 2020-09-10
Registry API change

philipktlin:
Remove registry proxy middleware and implement registry controller for existing and on-behalf-of registry API.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/IRegistryApiClient.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Http
+{
+    using System.Net.Http;
+    using System.Threading.Tasks;
+
+    public interface IRegistryApiClient
+    {
+        public Task<HttpResponseMessage> PutModuleOnBehalfOfAsync(string actorEdgeDeviceId, CreateOrUpdateModuleOnBehalfOfData requestData);
```
varunpuranik, line 9: Why do these need to be declared public?

varunpuranik, line 9: Nit: Suggest removing the "OnBehalfOfAsync" suffix..

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryApiClient.cs
```diff
@@ -0,0 +1,132 @@
...
+    using System.Text;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    public class RegistryApiClient : IRegistryApiClient
+    {
+        internal const string MediaType = "application/json";
+        const string PutModuleOnBehalfOfUriTemplate = "devices/{0}/modules/$edgeHub/putModuleOnBehalfOf?api-version={1}";
```
varunpuranik, line 17: Nit: make the moduleId a parameter as well, and sub it with the EdgeHub module Id.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryApiClient.cs
```diff
@@ -0,0 +1,132 @@
...
+        public RegistryApiClient(string upstreamHostname, Option<IWebProxy> proxy, ITokenProvider tokenProvider)
+        {
+            this.UpstreamHostName = Preconditions.CheckNonWhiteSpace(upstreamHostname, nameof(upstreamHostname));
+            this.proxy = proxy;
+            this.tokenProvider = Preconditions.CheckNotNull(tokenProvider, nameof(tokenProvider));
+
+            this.baseUri = new UriBuilder(Uri.UriSchemeHttps, upstreamHostname).Uri;
+        }
+
+        internal string UpstreamHostName { get; }
```
varunpuranik, line 36: Where else is this used?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryApiClient.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        public async Task<HttpResponseMessage> PutModuleOnBehalfOfAsync(string actorEdgeDeviceId, CreateOrUpdateModuleOnBehalfOfData requestData)
+        {
+            var requesteUri = new Uri(this.baseUri, string.Format(CultureInfo.InvariantCulture, PutModuleOnBehalfOfUriTemplate, actorEdgeDeviceId, ApiVersion));
```
varunpuranik, line 40: Do we need some kind of retry logic here, if the call fails because of a transient error?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryApiClient.cs
```diff
@@ -0,0 +1,132 @@
...
+                string token = await this.tokenProvider.GetTokenAsync(Option.None<TimeSpan>());
+                request.Headers.Add(HttpRequestHeader.Authorization.ToString(), token);
+
+                if (!string.IsNullOrEmpty(etag))
+                {
+                    request.Headers.Add("If-Match", etag);
+                }
+
+                var jsonPayload = JsonConvert.SerializeObject(payload);
+                LogDebug($"Uri={method} {request.RequestUri}");
```
varunpuranik, line 83: Events pattern, pls. And add details to the logs.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryApiClient.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        static readonly ILogger Log = Logger.Factory.CreateLogger<RegistryApiClient>();
+
+        public static void LogDebug(string message)
+        {
+            Log.LogDebug(9999, message);
+        }
+    }
+
+    public class CreateOrUpdateModuleOnBehalfOfData
```
varunpuranik, line 100: - Does this class have to be public? If not make it private and nested
- Pls make this immutable

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryApiClient.cs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        [JsonProperty(PropertyName = "authChain", Required = Required.Always)]
+        public string AuthChain { get; set; }
+
+        [JsonProperty(PropertyName = "module", Required = Required.Always)]
+        public Module Module { get; set; }
+    }
+
```
varunpuranik, line 108: Why do we need Required.Always?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,747 @@
...
+                Events.Authenticated(nameof(this.CreateOrUpdateModuleAsync), deviceId);
+                IEdgeHub edgeHub = await this.edgeHubGetter;
+                string edgeDeviceId = edgeHub.GetEdgeDeviceId();
+                string targetId = $"{deviceId}/{moduleId}";
+                IDeviceScopeIdentitiesCache identitiesCache = edgeHub.GetDeviceScopeIdentitiesCache();
+
+                (bool authorized, string authChain) = await this.AuthorizeActorAsync(identitiesCache, edgeDeviceId, Constants.EdgeHubModuleId, targetId);
+                if (!authorized)
+                {
+                    await this.SendResponseAsync(HttpStatusCode.Unauthorized);
```
varunpuranik, line 81: Log?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
```diff
@@ -100,7 +100,7 @@ internal bool ValidateAudienceIds(string audienceDeviceId, Option<string> audien
                 if (identity is IDeviceIdentity deviceIdentity && deviceIdentity.DeviceId != audienceDeviceId)
                 {
                     Events.IdMismatch(audienceId, identity, deviceIdentity.DeviceId);
-                    return false;
+                    return string.Equals(deviceIdentity.DeviceId, audienceDeviceId, StringComparison.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
philipktlin: just for testing, but confirmed that this is a bug and I need to discuss with Richard for solution.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
```diff
@@ -49,5 +50,54 @@ public static bool ValidateAuthChain(string actorDeviceId, string targetId, stri
...
+            // which could be a module ID of the format "deviceId/moduleId".
+            var actorAuthChainIds = authChain.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+
+            if (actorAuthChainIds.Length > 0)
+            {
+                var ids = actorAuthChainIds[0].Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+
+                if (ids.Length == 1 || ids.Length == 2)
+                {
+                    targetId = actorAuthChainIds[0];
```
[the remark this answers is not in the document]
philipktlin: No, it returns target id which can be "deviceid" or "deviceid/moduleid".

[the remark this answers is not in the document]
varunpuranik: Oh got it. Why do you need the ids Length check then?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,747 @@
...
+
+        [HttpPut]
+        [Route("devices/{actorDeviceId}/putModuleOnBehalfOf")]
+        public async Task CreateOrUpdateModuleOnBehalfOfAsync(
+            [FromRoute] string actorDeviceId,
+            [FromBody] CreateOrUpdateModuleOnBehalfOfData requestData)
+        {
+            try
+            {
+                Events.ReceivedOnBehalfOfRequest(nameof(this.CreateOrUpdateModuleOnBehalfOfAsync), actorDeviceId, Events.GetAdditionalInfo(requestData));
```
varunpuranik, line 106: Lot of duplicate code in these methods. Maybe we should refactor?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,747 @@
...
+            this.authenticatorGetter = Preconditions.CheckNotNull(authenticatorGetter, nameof(authenticatorGetter));
+            this.edgeHubGetter = Preconditions.CheckNotNull(edgeHubGetter, nameof(edgeHubGetter));
+        }
+
+        [HttpPut]
+        [Route("devices/{deviceId}/modules/{moduleId}")]
+        public async Task CreateOrUpdateModuleAsync(
+            [FromRoute] string deviceId,
+            [FromRoute] string moduleId,
+            [FromBody] Module module)
```
varunpuranik, line 38: Where is Module defined?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,747 @@
...
+                try
+                {
+                    deviceId = WebUtility.UrlDecode(Preconditions.CheckNonWhiteSpace(deviceId, nameof(deviceId)));
+                    moduleId = WebUtility.UrlDecode(Preconditions.CheckNonWhiteSpace(moduleId, nameof(moduleId)));
+                    Preconditions.CheckNotNull(module, nameof(module));
+
+                    if (!string.Equals(deviceId, module.DeviceId) || !string.Equals(moduleId, module.Id))
+                    {
+                        throw new ApplicationException("Device Id or module Id doesn't match between request URI and body.");
+                    }
```
varunpuranik, line 53: Interesting - why are the deviceId and moduleId specified twice in the request? Do we know?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/IRegistryApiClient.cs
philipktlin: What do you think to rename "IRegistryApiClient" to "IRegistryOnBehalfOfApiClient"?

philipktlin: will clean it up.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,747 @@
...
+                var resultUtf8Bytes = Encoding.UTF8.GetBytes(jsonContent);
+
+                this.Response.ContentLength = resultUtf8Bytes.Length;
+                this.Response.ContentType = ContentType;
+
+                await this.Response.Body.WriteAsync(resultUtf8Bytes, 0, resultUtf8Bytes.Length);
+            }
+        }
+
+        string FormatErrorResponseMessage(Exception ex)
```
varunpuranik, line 615: Nit: These kind of single line methods are not very useful. I would suggest removing them. Or maybe keep just the last FormatJsonMessage

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,747 @@
...
+        {
+            return this.FormatErrorResponseMessage(ex.ToString());
+        }
+
+        string FormatErrorResponseMessage(string errorMessage)
+        {
+            return this.FormatJsonMessage("errorMessage", errorMessage);
+        }
+
+        string FormatJsonMessage(string key, string value)
```
varunpuranik, line 625: pls make static

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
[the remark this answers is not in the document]
philipktlin: To ensure it is a valid target id.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryApiClient.cs
philipktlin: Yes, as this class requires to be public.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryApiClient.cs
philipktlin: This is used for create or update scenario, therefore this should be provided.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: in Microsoft.Azure.Devices namespace.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: Already has log inside this.AuthorizeActorAsync.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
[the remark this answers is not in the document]
philipktlin: removed.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryApiClient.cs
philipktlin: Keep it simple for now as RegistryProxyMiddleware don't have any retry logic.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: as mentioned, this is existing registry api method.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/IRegistryApiClient.cs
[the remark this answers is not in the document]
philipktlin: Create a new class "RegistryApiHttpResult"

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: Would you be specific what kind of refactoring do you mean? if using func delegate, I would like to keep it simple and don't change now.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: Cleaned up.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
```diff
@@ -49,5 +50,39 @@ public static bool ValidateAuthChain(string actorDeviceId, string targetId, stri
 
             return true;
         }
+
+        public static string SkipFirstIdentityFromAuthChain(string authChain)
```
varunpuranik, line 54: We need tests for all of these methods.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/IRegistryOnBehalfOfApiClient.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Http
+{
+    using System.Net.Http;
+    using System.Threading.Tasks;
+
+    public interface IRegistryOnBehalfOfApiClient
+    {
+        Task<RegistryApiHttpResult> PutModuleAsync(string actorEdgeDeviceId, CreateOrUpdateModuleOnBehalfOfData requestData, string ifMatchHeader);
```
varunpuranik, line 9: Nit: Does this have to be an Edge Device?

varunpuranik, line 9: Where is RegistryApiHttpResult defined?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryOnBehalfOfApiClient.cs
```diff
@@ -0,0 +1,167 @@
...
+
+    public class RegistryApiHttpResult
+    {
+        public RegistryApiHttpResult(HttpStatusCode statusCode, string jsonContent)
+        {
+            this.StatusCode = statusCode;
+            this.JsonContent = jsonContent ?? string.Empty;
+        }
+
+        public HttpStatusCode StatusCode { get; private set; }
```
varunpuranik, line 107: remove private set..

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryOnBehalfOfApiClient.cs
```diff
@@ -0,0 +1,167 @@
...
+                        }
+
+                        return new RegistryApiHttpResult(response.StatusCode, jsonContent);
+                    }
+                }
+            }
+        }
+    }
+
+    public class RegistryApiHttpResult
```
varunpuranik, line 99: Move public classes to their own file

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryOnBehalfOfApiClient.cs
```diff
@@ -0,0 +1,167 @@
...
+            this.StatusCode = statusCode;
+            this.JsonContent = jsonContent ?? string.Empty;
+        }
+
+        public HttpStatusCode StatusCode { get; private set; }
+
+        public string JsonContent { get; private set; }
+    }
+
+    public class CreateOrUpdateModuleOnBehalfOfData
```
varunpuranik, line 112: Do we need to define all these types? WHy not make them parameters on the call itself? Seems like its just a few parameters anyways

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
philipktlin: in RegistryOnBehalfOfApiClient.cs

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/IRegistryOnBehalfOfApiClient.cs
philipktlin: RegistryOnBehalfOfApiClient.cs

philipktlin: in RegistryOnBehalfOfApiClient.cs

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryOnBehalfOfApiClient.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        public async Task<RegistryApiHttpResult> DeleteModuleAsync(string actorEdgeDeviceId, DeleteModuleOnBehalfOfData requestData)
+        {
+            var requesteUri = new Uri(this.baseUri, string.Format(CultureInfo.InvariantCulture, this.deleteModuleOnBehalfOfUriTemplate, actorEdgeDeviceId, ApiVersion));
+            var response = await this.SendRequestAsync(requesteUri, HttpMethod.Post, requestData);
+            return response;
+        }
+
+        async Task<RegistryApiHttpResult> SendRequestAsync<T>(Uri uri, HttpMethod method, T payload, string ifMatchHeader = null)
```
varunpuranik, line 64: Nit: Can we use consistent naming for this and call it etag or ifMatchHeader everywehre?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/RegistryOnBehalfOfApiClient.cs
philipktlin: This is the payload and need serialization/deserialization.  Keep as-is for now.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,693 @@
...
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Events.BadRequest(nameof(this.CreateOrUpdateModuleAsync), ex.Message);
+                    await this.SendResponseAsync(HttpStatusCode.BadRequest, FormatErrorResponseMessage(ex.Message));
+                    return;
+                }
+
+                if (!await this.AuthenticateAsync(deviceId, Option.None<string>()))
```
varunpuranik, line 62: Why do we pass none here for moduleId?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,693 @@
...
+                    await this.SendResponseAsync(HttpStatusCode.Unauthorized);
+                    return;
+                }
+
+                IEdgeHub edgeHub = await this.edgeHubGetter;
+                string edgeDeviceId = edgeHub.GetEdgeDeviceId();
+                string targetId = $"{deviceId}/{moduleId}";
+                IDeviceScopeIdentitiesCache identitiesCache = edgeHub.GetDeviceScopeIdentitiesCache();
+
+                (bool authorized, string authChain) = await this.AuthorizeActorAsync(identitiesCache, edgeDeviceId, Constants.EdgeHubModuleId, targetId);
```
varunpuranik, line 200: Who are we authorizing here? And why? 
In this scenario, the child edge device is making a call to getModule, and the current edge device will make the on-behalf-of call, right? Are we validating if the device is our child? If so, won't Authenticate do that?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: As this is the endpoint, call by iotedged.  Therefore no module id for this codepath.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: yes, this will validate target device is current edge's child and help to return authChain (which is required to use for on-behalf-of api call).  For authenticate we are verifying child device sas token or certificate is valid.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
richma-ms: It shouldn't hurt to pass in the module ID though, it would in theory allow custom modules to call this API as well.

---
In reply to: [490643207](https://github.com/Azure/iotedge/pull/3532#discussion_r490643207) [](ancestors = 490643207)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,693 @@
...
```
richma-ms, line 73: >AuthorizeActorAsync [](start = 65, length = 19)

not necessary for existing API calls, but fine for now as I need to come back and remove these anyways.  All AuthZ checks will be handled by the DeviceScopeAuthenticator as of my pending change.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
richma-ms: Authenticate already handles that today for existing API calls.  But it's fine for for now as I need to come back and clean out all duplidated AuthZ implementations from both controllers (see other comment).

---
In reply to: [490649553](https://github.com/Azure/iotedge/pull/3532#discussion_r490649553) [](ancestors = 490649553)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,693 @@
...
+                if (!authorized)
+                {
+                    await this.SendResponseAsync(HttpStatusCode.Unauthorized);
+                    return;
+                }
+
+                Events.Authorized(nameof(this.GetModuleAsync), edgeDeviceId, Constants.EdgeHubModuleId, targetId, authChain);
+                var requestData = new GetModuleOnBehalfOfData($"{AuthChainHelpers.SkipFirstIdentityFromAuthChain(authChain)}", moduleId);
```
richma-ms, line 208: >SkipFirstIdentityFromAuthChain [](start = 82, length = 30)

Maybe rename to GetOriginEdgeAuthChain for clarity?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,693 @@
...
+            {
+                Events.Authenticated(deviceId, moduleId.GetOrElse(string.Empty));
+                return true;
+            }
+
+            Events.AuthenticateFail(deviceId, moduleId.GetOrElse(string.Empty));
+            return false;
+        }
+
+        async Task<(bool, string)> AuthorizeActorAsync(IDeviceScopeIdentitiesCache identitiesCache, string actorDeviceId, string actorModuleId, string targetId)
```
richma-ms, line 529: >actorModuleId [](start = 129, length = 13)

not needed since you're always calling with $edgeHub as a constant string anyways

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
varunpuranik: We should not assume who is making the call here. 
Also, if this is always called by a device, why does it have module in the route?

varunpuranik (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: What I understand is, an edge device owns its module and create/update module identity using their credential.  Therefore device Id should be considered only.  Module Id in the route is used for module identity creation.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,631 @@
...
+                }
+                catch (Exception ex)
+                {
+                    Events.BadRequest(nameof(this.CreateOrUpdateModuleOnBehalfOfAsync), ex.Message);
+                    await this.SendResponseAsync(HttpStatusCode.BadRequest, FormatErrorResponseMessage(ex.ToString()));
+                    return;
+                }
+
+                actorDeviceId = WebUtility.UrlDecode(actorDeviceId);
+                if (!AuthChainHelpers.TryGetTargetDeviceId(requestData.AuthChain, out string targetDeviceId))
```
richma-ms, line 268: >TryGetTargetDeviceId [](start = 38, length = 20)

Seems like the target identity won't ever be a module, both this method and CreateOrUpdateModuleAsync always forwards the call OnBehalfOf an Edge device.  So maybe it's better to explicit check the target ID and throw an error if it's a module.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -0,0 +1,631 @@
...
+                }
+                catch (Exception ex)
+                {
+                    Events.BadRequest(nameof(this.GetModuleOnBehalfOfAsync), ex.Message);
+                    await this.SendResponseAsync(HttpStatusCode.BadRequest, FormatErrorResponseMessage(ex.ToString()));
+                    return;
+                }
+
+                actorDeviceId = WebUtility.UrlDecode(actorDeviceId);
+                if (!AuthChainHelpers.TryGetTargetDeviceId(requestData.AuthChain, out string targetDeviceId))
```
richma-ms, line 339: >TryGetTargetDeviceId [](start = 38, length = 20)

same as above, this should always be an auth-chain for a device as opposed to module.

richma-ms (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: For registry api calls starting from iotedged, target id is always device id.  As far as I know, for on-behalf-of registry api call, authchain should not contain any module identity.  Please confirm if my understanding is correct.

philipktlin: Update TryGetTargetDeviceId to return target device Id only if 1st identity is a device identity.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
philipktlin: same reply as above.

philipktlin:
Tested with docker image version 202021.2.

richma-ms (approved):

PR #3534, opened 2020-09-10
Batch rqt upload

lfitchett:
This batches rqt metrics before uploading them if the resulting message would be more than 256 KB. Previously it simply dropped the message.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/MetricsWorker.cs
```diff
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics
 
     public class MetricsWorker : IDisposable
     {
-        public static RetryStrategy RetryStrategy = new ExponentialBackoff(20, TimeSpan.FromMinutes(5), TimeSpan.FromMinutes(30), TimeSpan.FromMinutes(1), false);
+        public static RetryStrategy RetryStrategy = new ExponentialBackoff(20, TimeSpan.FromMinutes(5), TimeSpan.FromHours(12), TimeSpan.FromMinutes(1), false);
```
lfitchett, line 25: I also realized a 30 minute retry is way too frequent for extended offline scenarios

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EdgeRuntimeDiagnosticsUpload.cs
```diff
@@ -14,6 +14,9 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Publisher
 
     public sealed class EdgeRuntimeDiagnosticsUpload : IMetricsPublisher
     {
+        // max message size is 256KB, keep a 1KB buffer for safty
+        const int MaxMessageSize = 255000;
```
damonbarry, line 18: Isn't the 256KB max really 256 * 1024 = 262,144 bytes? So that's nearly a 7KB buffer...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EdgeRuntimeDiagnosticsUpload.cs
```diff
@@ -25,26 +28,43 @@ public EdgeRuntimeDiagnosticsUpload(IEdgeAgentConnection edgeAgentConnection)
...
+            {
+                return Enumerable.Empty<Message>();
+            }
+
+            byte[] data = MetricsSerializer.MetricsToBytes(metrics).ToArray();
+
+            if (data.Length > MaxMessageSize)
+            {
+                var part1 = this.BatchAndBuildMessages(metrics.Slice(0, metrics.Count / 2));
+                var part2 = this.BatchAndBuildMessages(metrics.Slice(metrics.Count / 2, metrics.Count / 2 + metrics.Count % 2));
```
damonbarry, line 58: There's another overload of ArraySegment.Slice that only takes an index arg, and returns a slice from the index to the end of the parent slice. Seems like a better option than doing the math yourself for `part2`.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EdgeRuntimeDiagnosticsUpload.cs
```diff
@@ -25,26 +25,43 @@ public EdgeRuntimeDiagnosticsUpload(IEdgeAgentConnection edgeAgentConnection)
...
+            {
+                return Enumerable.Empty<Message>();
+            }
+
+            byte[] data = MetricsSerializer.MetricsToBytes(metrics).ToArray();
+
+            if (data.Length > 250000)
+            {
+                var part1 = this.BatchAndBuildMessages(metrics.Slice(0, data.Length / 2));
+                var part2 = this.BatchAndBuildMessages(metrics.Slice(data.Length / 2, data.Length / 2));
```
CindyXing: part2 would lose one segment if data.length is odd.

CindyXing: With recursion, this would lose more in between.

CindyXing (dismissed):

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/publisher/EdgeRuntimeDiagnosticsUpload.cs
lfitchett: You are looking at a very outdated version somehow. I think you are looking at the first commit only. There are tests to check both even and odd versions and that is not an issue in the current version.

damonbarry (approved):

PR #3536, opened 2020-09-10
Handle Close requests for clients and send disconnect signal to broker

vipeller:


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/DeviceProxyTest.cs
```diff
@@ -27,7 +28,7 @@ public async Task DirectMethodCallForwarded()
                 .Setup(h => h.CallDirectMethodAsync(It.Is<DirectMethodRequest>(m => m == request), It.Is<IIdentity>(i => i == identity)))
                 .Returns(Task.FromResult(new DirectMethodResponse("123", new byte[0], 200)));
 
-            var sut = new DeviceProxy(identity, twinHandler, m2mHandler, c2dHandler, directMethodHandler);
+            var sut = new DeviceProxy(identity, connectionHanlder, twinHandler, m2mHandler, c2dHandler, directMethodHandler);
```
and-rewsmith, line 31: typo

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IConnectionRegistry.cs
```diff
@@ -11,5 +11,7 @@ public interface IConnectionRegistry
     {
         Task<Option<IDeviceListener>> GetDeviceListenerAsync(IIdentity identity);
         Task<Option<IDeviceProxy>> GetDeviceProxyAsync(IIdentity identity);
+
```
and-rewsmith, line 14: nit: should we remove this whitespace?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ConnectionHandler.cs
```diff
@@ -94,6 +99,20 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
             return false;
         }
 
+        public async Task CloseConnectionAsync(IIdentity identity)
+        {
+            var identityFound = false;
+            using (await this.guard.LockAsync())
```
and-rewsmith, line 105: We need to lock this because something might get added to knownConnections?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ConnectionHandler.cs
```diff
@@ -94,6 +99,20 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+            {
+                identityFound = this.knownConnections.TryGetValue(identity, out var _);
+            }
+
+            if (identityFound)
+            {
+                await this.connector.SendAsync(TopicDisconnect, Encoding.UTF8.GetBytes($"\"{identity.Id}\""));
+            }
```
and-rewsmith, line 113: If the identity isn't found should we log error?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/ConnectionHandlerTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            DeviceProxy GetProxy(IIdentity identity)
+            {
+                return new DeviceProxy(
+                                identity,
+                                sut,
+                                Mock.Of<ITwinHandler>(),
+                                Mock.Of<IModuleToModuleMessageHandler>(),
+                                Mock.Of<ICloud2DeviceMessageHandler>(),
+                                Mock.Of<IDirectMethodHandler>());
+            }
```
and-rewsmith, line 66: Is defining the function this way (i.e. inside CloseSendsDisconnectSignal) better than having a separate function?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/DeviceProxyTest.cs
```diff
@@ -49,16 +51,17 @@ public async Task DesiredUpdateCallForwarded()
                 .Setup(h => h.SendDesiredPropertiesUpdate(It.Is<IMessage>(m => m == desired), It.Is<IIdentity>(i => i == identity)))
                 .Returns(Task.CompletedTask);
 
-            var sut = new DeviceProxy(identity, twinHandler, m2mHandler, c2dHandler, directMethodHandler);
+            var sut = new DeviceProxy(identity, connectionHanlder, twinHandler, m2mHandler, c2dHandler, directMethodHandler);
 
             await sut.OnDesiredPropertyUpdates(desired);
 
-            Mock.Get(directMethodHandler).VerifyAll();
+            Mock.Get(twinHandler).VerifyAll();
```
and-rewsmith, line 58: Why is this changed to twinHandler?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/DeviceProxyTest.cs
```diff
@@ -71,16 +74,17 @@ public async Task Cloud2DeviceMessagesForwarded()
                 .Setup(h => h.SendC2DMessageAsync(It.Is<IMessage>(m => m == message), It.Is<IIdentity>(i => i == identity)))
                 .Returns(Task.CompletedTask);
 
-            var sut = new DeviceProxy(identity, twinHandler, m2mHandler, c2dHandler, directMethodHandler);
+            var sut = new DeviceProxy(identity, connectionHanlder, twinHandler, m2mHandler, c2dHandler, directMethodHandler);
 
             await sut.SendC2DMessageAsync(message);
 
-            Mock.Get(directMethodHandler).VerifyAll();
```
and-rewsmith, line 78: Why this change?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/DeviceProxy.cs
```diff
@@ -41,15 +44,15 @@ public class DeviceProxy : IDeviceProxy
 
         public IIdentity Identity { get; }
 
-        public Task CloseAsync(Exception ex)
+        public async Task CloseAsync(Exception ex)
```
and-rewsmith, line 47: Do we do anything with this exception?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ConnectionHandler.cs
vadim-kovalyov: What race conditions we are trying to prevent with the lock? knownConnections is a ConcurrentDictionary already

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ConnectionHandler.cs
```diff
@@ -94,6 +99,20 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+        {
+            var identityFound = false;
+            using (await this.guard.LockAsync())
+            {
+                identityFound = this.knownConnections.TryGetValue(identity, out var _);
+            }
+
+            if (identityFound)
+            {
+                await this.connector.SendAsync(TopicDisconnect, Encoding.UTF8.GetBytes($"\"{identity.Id}\""));
```
vadim-kovalyov, line 112: do we need to remove from knownConnections?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/DeviceProxy.cs
vipeller: added to the log

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/IConnectionRegistry.cs
vipeller: I added because those two above belong to one group of operations and the one below to a different one

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ConnectionHandler.cs
vipeller: The story of the lock is that when the code updates the 'knowConnection', that needs a lock anyway (because it does multistep things. so a parallel call cannot mess it up) - so it would not be necessary using a concurrent dictionary (although I ended up using one, because it has a member normal dictionaries don't) And because of the lock on writes, I felt it is better to keep it symmetric, and lock the reads, too.   But yes, with the current ConcurrentDictionary the read operation would not be dangerous doing without a lock.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ConnectionHandler.cs
vipeller: I am not sure if that is an error. Can a device be gone by the time edgeHub realizes that something is wrong a calls Close()? I can add an 'Info'.

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/ConnectionHandlerTests.cs
vipeller: it can capture local variables (sut in this case), also in tests these helper functions many times are used only from a single method, so it is easier to see what serves what.

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/DeviceProxyTest.cs
vipeller: it was a copy-paste error from earlier and it did not call the "verifyall" on the set up mock object. Now it checks the right mock. Earlier the test passed only because nothing was setup on 'directMethodHandler', so verifyAll meant 'verify nothing' - which always passed.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ConnectionHandler.cs
vipeller: No, I figured that after this point it is the turn of the broker to do something, and when it make the client disconnect, the notification will come in a second round, and the identity will be removed that time

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/DeviceProxyTest.cs
vipeller: it was a copy-paste error from earlier and it did not call the "verifyall" on the set up mock object. Now it checks the right mock. Earlier the test passed only because nothing was setup on 'directMethodHandler', so verifyAll meant 'verify nothing' - which always passed.

vadim-kovalyov (approved):

varunpuranik (approved):

and-rewsmith (approved):

PR #3540, opened 2020-09-11
mqtt bridge connect to local broker and upstream to receive messages

ancaantochi:
- connects to local broker and saves messages to storage
- connect to upstream and saves messages to storage


mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
+
+        handler
+            .handle_event(Event::Publication(pub1))
+            .await
+            .unwrap();
 
-    async fn connect_to_local(&self) {
-        info!("connecting to local broker");
+        let key1 = Key { offset: 0 };
+        assert_eq!(None, handler.inner.remove(key1).await);
```
and-rewsmith, line 383: This is a hard test to write because of the behavior of the persistor. A remove will always fail if the message hasn't been extracted by the loader. Therefore the test isn't checking correctly right now.

If we want to verify that a message doesn't exist, we will have to prove that the loader can't get it. It's hard because the loader is a stream.

and-rewsmith: It is maybe best to use a more roundabout way of solving this. For example, make a second received publication with a valid topic match and verify the loader retrieves this message first. The test will fail if there is a topic match for the first publication (intended non-topic match) because in this scenario the loader would get this message first.

and-rewsmith (commented):
I still need to look at some files, but here are my intial comments.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
+            self.connection_settings.clean_session(),
+            MessageHandler::new(topic_filters, BATCH_SIZE),
+            self.connection_settings.credentials(),
+        );
+
+        let subscriptions: Vec<String> = self.subs_map.keys().map(|key| key.into()).collect();
+        debug!("subscribe to remote {:?}", subscriptions);
+
+        client.subscribe(subscriptions).await?;
+        let _events_task = tokio::spawn(client.handle_events());
```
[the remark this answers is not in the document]
ancaantochi: yes, I will add a TODO here, it should be handled when shutdown is handled

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
+
+impl<'a, T> MessageHandler<'a, T>
+where
+    T: Persist<'a>,
+{
+    pub fn new(topics: Vec<TopicMapper>, batch_size: usize) -> Self {
+        Self {
+            topics,
+            inner: T::new(batch_size),
+            phantom: PhantomData,
```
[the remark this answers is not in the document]
ancaantochi: it is needed because the lifetime 'a is not used here, it's to make compiler happy.

mqtt/mqtt-bridge/tests/config.json
```diff
@@ -4,11 +4,15 @@
...
+    "iothubhostname": "iothub",
     "upstream": {
         "subscriptions": [
             {
                 "pattern": "temp/#",
-                "remote": "floor/kitchen/"
+                "remote": "floor/kitchen"
+            },
+            {
+                "pattern": "pattern/#"
```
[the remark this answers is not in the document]
ancaantochi: yes, just forwards on the same topic

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
-        info!("connecting to remote broker");
+    fn transform(&self, topic_name: &str) -> Option<String> {
+        for mapper in &self.topics {
+            let pattern = mapper
+                .topic_settings
+                .local()
+                .map_or(Some(topic_name.to_owned()), |local_prefix| {
+                    if topic_name.to_owned().starts_with(local_prefix) {
+                        let rs: String = topic_name
+                            .strip_prefix(local_prefix)
```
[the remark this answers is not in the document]
ancaantochi: because it should not be part of the forwarded topic

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
+
+            if pattern.is_some() {
+                return pattern;
+            }
+        }
+        None
     }
+}
+
+// TODO: implement for generic T where T: Persist
```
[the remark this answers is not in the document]
ancaantochi: yes, it will need to integrate with the changes from your PR, let's sync about this

and-rewsmith:
If you think there is too much friction merging this, we can have a separate PR for just the client/sas token stuff

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
+            });
+        }
+
+        let mut client = MqttClient::new(
+            self.system_address.as_str(),
+            *self.connection_settings.keep_alive(),
+            self.connection_settings.clean_session(),
+            MessageHandler::new(topic_filters, BATCH_SIZE),
+            self.connection_settings.credentials(),
+        );
```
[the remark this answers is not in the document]
ancaantochi: no, removed

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,223 @@
...
+                }
+            };
+
+            let io = TcpStream::connect(address).await;
+            io.map(|io| (io, password))
+        })
+    }
+}
+
+pub struct MqttClient<T>
```
and-rewsmith, line 74: nit: it might be confusing to have Client and MqttClient types. Also, it is just a wrapper. We could maybe indicate that in the name? I don't think it's too bad if we leave it as-is though.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        error!("failed to subscribe to topics");
+        Err(ClientConnectError::MissingSubacks(
+            subacks
+                .into_iter()
+                .map(std::string::ToString::to_string)
+                .collect(),
+        ))
+    }
+}
```
and-rewsmith, line 203: Since this logic is being duplicated, we should have a plan to consolidate if we are not going to do it now.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,223 @@
...
+                    Ok(x) => Some(x),
+                    Err(e) => {
+                        warn!("Failed to get token {}", e);
+                        None
+                    }
+                }
+            };
+
+            let io = TcpStream::connect(address).await;
+            io.map(|io| (io, password))
```
and-rewsmith, line 69: Do you know how this password from this return type gets used?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,223 @@
...
+pub trait EventHandler {
+    async fn handle_event(&mut self, event: Event) -> Result<()>;
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ClientConnectError {
+    #[error("failed to receive expected subacks for command topics: {0:?}")]
+    MissingSubacks(Vec<String>),
+
+    #[error("failed to subscribe command handler to command topic")]
```
and-rewsmith, line 215: I think we should rename since it isn't the command handler

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -17,13 +17,12 @@ impl BridgeController {
         Self::default()
     }
 
-    pub async fn start(&mut self) -> Result<()> {
+    pub async fn start(&mut self, system_address: String, device_id: &str) -> Result<()> {
```
and-rewsmith, line 20: At some point we will need to return a shutdown handle and integrate properly into the startup logic

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -17,13 +17,12 @@ impl BridgeController {
...
-
         let settings = Settings::new()?;
         if let Some(upstream) = settings.upstream() {
-            let nested_bridge = Bridge::new(upstream.clone());
-            nested_bridge.start().await;
+            let nested_bridge = Bridge::new(system_address, device_id.into(), upstream.clone());
+            nested_bridge.start().await?;
 
             self.bridges
                 .insert(upstream.name().to_string(), nested_bridge);
```
and-rewsmith, line 28: Assuming remote bridges are coming in a later PR.

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -17,13 +17,12 @@ impl BridgeController {
         Self::default()
     }
 
-    pub async fn start(&mut self) -> Result<()> {
+    pub async fn start(&mut self, system_address: String, device_id: &str) -> Result<()> {
         info!("starting bridge");
-
         let settings = Settings::new()?;
         if let Some(upstream) = settings.upstream() {
```
and-rewsmith, line 23: If there are no upstream settings then we log:

> "No nested bridge found"

I think we should say something like:

> "No upstream settings detected. Not starting bridge.

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -180,75 +177,71 @@ impl AuthenticationSettings {
...
 
 #[derive(Debug, Clone, Default, PartialEq, Deserialize)]
-pub struct Forward {
+pub struct Topic {
     pattern: String,
 
-    remote: String,
+    local: Option<String>,
+
+    remote: Option<String>,
```
and-rewsmith, line 231: Anca and I discussed offline and we think it should be:

```
"pattern": "temp/#",
"forward-prefix": "floor/kitchen/",
"subscribe-prefix": "local"
```

and-rewsmith (commented):
Comments for the rest of the files

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,387 @@
-use tracing::info;
+use std::{collections::HashMap, marker::PhantomData, str::FromStr, time::Duration};
 
-use crate::settings::ConnectionSettings;
+use anyhow::Result;
```
dmolokanov, line 3: let's not use anyhow in lib crates for error handling. use thiserror instead

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,387 @@
...
-    pub fn new(connection_settings: ConnectionSettings) -> Self {
+    pub fn new(
+        system_address: String,
+        device_id: String,
+        connection_settings: ConnectionSettings,
+    ) -> Self {
+        let forwards_map: HashMap<String, Topic> = connection_settings
+            .forwards()
+            .iter()
+            .map(move |sub| Bridge::format_key_value(sub))
```
dmolokanov, line 33: I don't think `move` in necessary here

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,387 @@
...
+use crate::settings::{ConnectionSettings, Credentials, Topic};
+
+const BATCH_SIZE: usize = 10;
 
 /// Bridge implementation that connects to local broker and remote broker and handles messages flow
 pub struct Bridge {
+    system_address: String,
+    device_id: String,
     connection_settings: ConnectionSettings,
+    forwards_map: HashMap<String, Topic>,
```
dmolokanov, line 20: _map is a bit confusing suffix. can we just remove it at all?

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,387 @@
...
+            .map(move |sub| Bridge::format_key_value(sub))
+            .collect();
+
+        let subs_map: HashMap<String, Topic> = connection_settings
+            .subscriptions()
+            .iter()
+            .map(|sub| Bridge::format_key_value(sub))
+            .collect();
+
         Bridge {
```
dmolokanov, line 42: If you use type name inside the type definition, you can use `Self` instead of type name `Bridge` like
`Self { ... }` or `Self::foo(bar)`

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,387 @@
...
 }
 
 impl Bridge {
-    pub fn new(connection_settings: ConnectionSettings) -> Self {
+    pub fn new(
+        system_address: String,
+        device_id: String,
+        connection_settings: ConnectionSettings,
+    ) -> Self {
+        let forwards_map: HashMap<String, Topic> = connection_settings
```
dmolokanov, line 30: collect combinator wants to know a type of resulting collection, not items. so we can omit that like
```suggestion
        let forwards_map: HashMap<_> = connection_settings
```
or even 
```suggestion
        let forwards_map = connection_settings
```
because you construct Bridge and HashMap required there

mqtt/mqtt-bridge/src/bridge.rs
ancaantochi: I though anyhow error work with thiserror mapping, I am confused by the examples.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
+    fn transform(&self, topic_name: &str) -> Option<String> {
+        for mapper in &self.topics {
+            let pattern = mapper
+                .topic_settings
+                .local()
+                .map_or(Some(topic_name.to_owned()), |local_prefix| {
+                    if topic_name.to_owned().starts_with(local_prefix) {
+                        let rs: String = topic_name
+                            .strip_prefix(local_prefix)
+                            .unwrap_or(&topic_name)
```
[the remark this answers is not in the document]
ancaantochi: it is, not actually possible because it checks it starts with prefix

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,223 @@
...
+use tracing::{debug, error, warn};
+
+use mqtt3::{
+    proto, Client, Event, IoSource, ShutdownError, SubscriptionUpdateEvent, UpdateSubscriptionError,
+};
+
+use crate::settings::Credentials;
+use crate::token_source::{SasTokenSource, TokenSource};
+
+const DEFAULT_TOKEN_DURATION_SECS: Duration = Duration::from_secs(60 * 60);
```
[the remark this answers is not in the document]
ancaantochi: no

mqtt/mqtt-bridge/src/client.rs
ancaantochi: just that it is used in the connect packet.

mqtt/mqtt-bridge/src/bridge.rs
ancaantochi: the case when there is a message is if the test runs in parallel with other tests or how would there be another message in the loader. I think it should try to get the message and timeout after a while, but I couldn't make that work.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,223 @@
...
+
+impl ShutdownHandle {
+    pub async fn shutdown(&mut self) -> Result<(), ClientConnectError> {
+        self.0
+            .shutdown()
+            .await
+            .map_err(ClientConnectError::ShutdownClient)?;
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
ancaantochi: This would be used so the client is not exposed

mqtt/mqtt-bridge/src/client.rs
ancaantochi: yes, the plan is to consolidate at a later time

mqtt/mqtt-bridge/src/controller.rs
ancaantochi: agree, shutdown is not handled yet by bridge

mqtt/mqtt-bridge/tests/config.json
```diff
@@ -4,11 +4,15 @@
     "moduleid": "mymodule",
     "modulegenerationid": "321",
```
and-rewsmith, line 5: What casing are/should we use for these fields.

mqtt/mqtt-bridge/tests/config.json
```diff
@@ -4,11 +4,15 @@
     "moduleid": "mymodule",
```
and-rewsmith, line 4: From Denis: All these fields are only relevant to upstream, so maybe we should move to the upstream section.

and-rewsmith:
More comments

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
+}
+
+#[derive(Clone)]
+struct TopicMapper {
+    topic_settings: Topic,
+    topic_filter: TopicFilter,
+}
+
+#[derive(Clone)]
+struct MessageHandler<'a, T>
```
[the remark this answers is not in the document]
ancaantochi: added comments to the transform function

mqtt/mqtt-bridge/src/controller.rs
ancaantochi: correct

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
+                    }
+                });
+
+            if pattern.is_some() {
+                return pattern;
+            }
+        }
+        None
     }
+}
```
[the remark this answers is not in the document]
ancaantochi: I added some comments. As discussed, the config could be changed to make it easier to understand using in/out and prefix, but will keep it like this until schema is finalized

mqtt/mqtt-bridge/tests/config.json
ancaantochi: they are actually retrieved from env variables so they are at top level, here are just for configuring tests

mqtt/mqtt-bridge/tests/config.json
ancaantochi: it's a string

mqtt/mqtt-bridge/src/settings.rs
ancaantochi: will leave it local/remote for now, until we finalize the config schema

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,385 @@
...
+                    } else {
+                        pattern
+                    }
+                });
+
+            if pattern.is_some() {
+                return pattern;
+            }
+        }
+        None
```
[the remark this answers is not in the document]
ancaantochi: right, it should not happen. it's not threated as an error, it just logs that's why I didn't return Result

mqtt/mqtt-bridge/src/bridge.rs
ancaantochi: made it work with timer

mqtt/mqtt-bridge/src/persist/memory/loader.rs
```diff
@@ -18,7 +18,7 @@ pub struct InMemoryMessageLoader {
 }
 
 impl InMemoryMessageLoader {
-    pub async fn new(state: Arc<Mutex<WakingMap>>, batch_size: usize) -> Self {
+    pub fn new(state: &Arc<Mutex<WakingMap>>, batch_size: usize) -> Self {
```
dmolokanov, line 21: I think you can still consume `state` and omit cloning later

mqtt/mqtt-bridge/src/persist/memory/mod.rs
```diff
@@ -25,11 +25,11 @@ impl<'a> Persist<'a> for InMemoryPersist {
     type Loader = InMemoryMessageLoader;
```
dmolokanov, line 25: why lifetime `'a` here?

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -5,17 +5,17 @@ use futures_util::stream::Stream;
...
 
-mod memory;
+pub mod memory;
 
 /// Persistence used in bridge.
 /// Elements are added, then can be removed once retrieved by the loader
 /// If one attempts to remove added elements before reading via the loader remove returns None
 #[async_trait]
-trait Persist<'a> {
+pub trait Persist<'a> {
```
dmolokanov, line 14: I don't see lifetime `'a` used anywhere. Should we remove that?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+    type Error = std::io::Error;
+    type Future = BoxFuture<'static, Result<(TcpStream, Option<String>), std::io::Error>>;
+
+    fn connect(&mut self) -> Self::Future {
+        let address = self.address.clone();
+        let token_source = self.token_source.as_ref().cloned();
+
+        Box::pin(async move {
+            let expiry = Utc::now()
+                + chrono::Duration::from_std(DEFAULT_TOKEN_DURATION_SECS).unwrap_or_else(|_| {
```
dmolokanov, line 56: Can we make this a const or a static `chrono::Duration` instead of `std::time::Duration` defined on top of the module?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+                    chrono::Duration::minutes(DEFAULT_TOKEN_DURATION_MINS.try_into().unwrap())
+                });
```
dmolokanov, line 58: unlike c# everything in Rust is an expression, so you can assign result of if-block to variable:
```rust
            let password = if let Some(ts) = token_source {
                match ts.get(&expiry).await {
                    Ok(x) => Some(x),
                    Err(e) => {
                        error!("Failed to get token for connection {} {}", address, e);
                        None
                    }
                }
            } else {
                None
            };
```

dmolokanov: Also there are some combinators to make conversion Result -> Option if you need. https://doc.rust-lang.org/std/result/enum.Result.html#method.ok

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+                creds.client_id().into(),
+                Some(SasTokenSource::new(connection_credentials.clone())),
+            ),
+            Credentials::Anonymous(client_id) => (client_id.into(), None),
+        };
+
+        let client = if clean_session {
+            Client::new(
+                Some(client_id),
+                None,
```
dmolokanov, line 107: I think it wants some `username`

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub async fn shutdown(&mut self) -> Result<(), ClientConnectError> {
+        self.0
+            .shutdown()
+            .await
+            .map_err(ClientConnectError::ShutdownClient)?;
+        Ok(())
+    }
+}
+
+pub struct TcpConnection<T>
```
dmolokanov, line 34: lets make a `new` method for convenience? rust_analyzer can generate it for you

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,473 @@
...
+            keep_alive,
+            clean_session,
+            MessageHandler::new(topic_filters, BATCH_SIZE),
+            credentials,
+        );
+
+        let subscriptions: Vec<String> = topics.keys().map(|key| key.into()).collect();
+        debug!("subscribe to remote {:?}", subscriptions);
+
+        client
```
dmolokanov, line 133: lets add TryFrom for convenience
```rust

impl TryFrom<Topic> for TopicMapper {
    type Error = BridgeError;

    fn try_from(topic: Topic) -> Result<Self, Self::Error> {
        let topic_filter = topic
            .pattern()
            .parse()
            .map_err(BridgeError::TopicFilterParseError)?;

        Ok(Self {
            topic_settings: topic,
            topic_filter,
        })
    }
}
```

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,473 @@
...
+        &self,
+        topics: &HashMap<String, Topic>,
+        address: &str,
+        keep_alive: Duration,
+        clean_session: bool,
+        credentials: &Credentials,
+    ) -> Result<(), BridgeError> {
+        let mut topic_filters = Vec::new();
+        for val in topics.values() {
+            topic_filters.push(TopicMapper {
```
dmolokanov, line 115:
```rust 
let topic_filters = topics
            .values()
            .map(|topic| TopicMapper::from_topic(topic))
            .collect::<Result<Vec<_>, _>>()?;
```
in a more declarative fashion

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub fn shutdown_handle(&self) -> Result<ShutdownHandle, ShutdownError> {
+        self.client
+            .shutdown_handle()
+            .map_or(Err(ShutdownError::ClientDoesNotExist), |shutdown_handle| {
+                Ok(ShutdownHandle(shutdown_handle))
+            })
+    }
+
+    pub async fn handle_events(mut self) -> Result<(), ClientConnectError> {
+        while let Some(event) = self.client.try_next().await.unwrap_or_else(|e| {
```
dmolokanov, line 145: I thought Err here means that we should stop polling stream because of IO error. Could you double check that we can continue work with a client in this case?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+                TcpConnection {
+                    address: address.into(),
+                    token_source,
+                },
+                DEFAULT_MAX_RECONNECT,
+                keep_alive,
+            )
+        };
+
+        MqttClient {
```
dmolokanov, line 130: Self instead of MqttClient

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+            .map_err(ClientConnectError::PollClientFailure)?
+        {
+            if let Event::SubscriptionUpdates(subscriptions) = event {
+                for subscription in subscriptions {
+                    if let SubscriptionUpdateEvent::Subscribe(sub) = subscription {
+                        subacks.remove(&sub.topic_filter);
+                    }
+                }
+
+                if subacks.is_empty() {
```
dmolokanov, line 190: It should be OK to receive NotAllowed from a broker, right? I think we need to log error and continue, instead of erroring out client

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+                .into_iter()
+                .map(std::string::ToString::to_string)
+                .collect(),
+        ))
+    }
+}
+
+#[async_trait]
+pub trait EventHandler {
+    type Error: std::fmt::Display;
```
dmolokanov, line 209: type Error: Display;

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        Ok(())
+    }
+
+    pub async fn subscribe(&mut self, topics: Vec<String>) -> Result<(), ClientConnectError> {
+        debug!("subscribing to topics");
+        let subscriptions = topics.iter().map(|topic| proto::SubscribeTo {
+            topic_filter: topic.to_string(),
+            qos: proto::QoS::AtLeastOnce,
```
dmolokanov, line 161: shouldn't we take QoS from outside? while it is QoS1 for iothub topics, it can be QoS0 for everything else

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -131,11 +127,11 @@ impl ConnectionSettings {
...
 
-    pub fn subscriptions(&self) -> &Vec<Subscription> {
+    pub fn subscriptions(&self) -> &Vec<Topic> {
         &self.subscriptions
     }
 
-    pub fn forwards(&self) -> &Vec<Forward> {
+    pub fn forwards(&self) -> &Vec<Topic> {
         &self.forwards
     }
```
dmolokanov, line 137:
```rust
pub fn keep_alive(&self) -> Duration
```
we can remove `&` symbol here and get rid of `*` wherever we use this method. It works because `Duration` implements `Copy` trait which is fast cloning.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,473 @@
...
+    async fn connect_to_remote(&self) -> Result<(), BridgeError> {
+        info!(
+            "connecting to remote broker {}",
+            self.connection_settings.address()
+        );
+
+        self.connect(
+            &self.subscriptions,
+            self.connection_settings.address(),
+            *self.connection_settings.keep_alive(),
```
dmolokanov, line 77: we can do `self.connection_settings.keep_alive()` here

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,473 @@
...
+    }
+
+    async fn connect_to_remote(&self) -> Result<(), BridgeError> {
+        info!(
+            "connecting to remote broker {}",
+            self.connection_settings.address()
+        );
+
+        self.connect(
+            &self.subscriptions,
```
dmolokanov, line 75: you borrow subsctiptions here but inside connect method you clone everything else. I wonder if it worthwhile to clone hashmap here and just deconstruct it inside connect method. It'll make code easier I believe

mqtt/mqtt-bridge/src/bridge.rs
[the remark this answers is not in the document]
dmolokanov: unlike C# in Rust generic name is usually not T everywhere, but the first letter of trait that generic implements:
`struct MessageHandler<'a, P>` P - Persist

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,473 @@
...
+struct TopicMapper {
+    topic_settings: Topic,
+    topic_filter: TopicFilter,
+}
+
+/// Handle events from client and saves them with the forward topic
+#[derive(Clone)]
+struct MessageHandler<'a, T>
+where
+    T: Persist<'a>,
```
dmolokanov, line 155: I believe we don't need lifetime and phantom altogether

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,473 @@
...
+                .topic_settings
+                .local()
+                // maps if local does not have a value it uses the topic that was received,
+                // else it checks that the received topic starts with local prefix and removes the local prefix
+                .map_or(Some(topic_name.to_owned()), |local_prefix| {
+                    let prefix = format!("{}/", local_prefix);
+                    if topic_name.to_owned().starts_with(&prefix) {
+                        let rs: String = topic_name
+                            .strip_prefix(&prefix)
+                            .unwrap_or(&topic_name)
```
dmolokanov, line 186: I think you don't need to check starts_with to remove prefix. strip_prefix will do the same for you

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,473 @@
...
+    }
+}
+
+#[derive(Clone)]
+struct TopicMapper {
+    topic_settings: Topic,
+    topic_filter: TopicFilter,
+}
+
+/// Handle events from client and saves them with the forward topic
```
dmolokanov, line 151: I removed some redundand allocations with to_owned
```rust
mapper
                .topic_settings
                .local()
                // maps if local does not have a value it uses the topic that was received,
                // else it checks that the received topic starts with local prefix and removes the local prefix
                .map_or(Some(topic_name), |local_prefix| {
                    let prefix = format!("{}/", local_prefix);
                    topic_name.strip_prefix(&prefix)
                })
                // match topic without local prefix with the topic filter pattern
                .filter(|topic| mapper.topic_filter.matches(topic))
                .map(|topic| {
                    if let Some(remote_prefix) = mapper.topic_settings.remote() {
                        format!("{}/{}", remote_prefix, topic)
                    } else {
                        topic.to_string()
                    }
                });
```

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,31 +1,473 @@
...
+#[derive(Clone)]
```
dmolokanov, line 152: I removed some redundand allocations with to_owned
```rust
        self.topic_mappers
            .iter()
            .filter_map(|mapper| {
                mapper
                    .topic_settings
                    .local()
                    // maps if local does not have a value it uses the topic that was received,
                    // else it checks that the received topic starts with local prefix and removes the local prefix
                    .map_or(Some(topic_name), |local_prefix| {
                        let prefix = format!("{}/", local_prefix);
                        topic_name.strip_prefix(&prefix)
                    })
                    // match topic without local prefix with the topic filter pattern
                    .filter(|topic| mapper.topic_filter.matches(topic))
                    .map(|topic| {
                        if let Some(remote_prefix) = mapper.topic_settings.remote() {
                            format!("{}/{}", remote_prefix, topic)
                        } else {
                            topic.to_string()
                        }
                    })
            })
            .next()
```

mqtt/mqtt-bridge/src/persist/memory/loader.rs
ancaantochi: I made these changes to the loader to be able to integrate with persistence, but Andrew's PR completely changes the API so all changes in persist folder are throw-away

mqtt/mqtt-bridge/src/persist/mod.rs
ancaantochi: same as above, so will just leave it as is for now

mqtt/mqtt-bridge/src/bridge.rs
ancaantochi: with the change that Andrew has in PR for persistence this will me removed.

mqtt/mqtt-bridge/src/client.rs
ancaantochi: this is the same as for command handler, only client_id is specified

mqtt/mqtt-bridge/src/bridge.rs
ancaantochi: thanks, looks pretty neat!

mqtt/mqtt-bridge/src/client.rs
ancaantochi: I changed it to have the minutes as const and create it form that. I doesn't make sense to create it from string and unwrap. Let me know if there is a better way, as const it didn't compile

mqtt/mqtt-bridge/src/client.rs
ancaantochi: the way the mqtt client is implemented it would not send the error to the caller so actually it would keep looping if the sub is not allowed. So a put a temporary fix, but will need to modify the client to send an event if the sub is not allowed.

mqtt/mqtt-bridge/src/client.rs
ancaantochi: I added a default qos for now. When we talked offline I said that I'll add it to topic, but it's easier to have it as const and change it when we have a final way for settings

mqtt/mqtt-bridge/src/client.rs
ancaantochi: you're right, client can't be reused so it needs to be recreated. As discussed offline, I will have a different PR to handle this case.

mqtt/mqtt-bridge/src/client.rs
dmolokanov: Ok, I see a complication here...
`std::time::Duration` has `const fn from_secs()` function but `chrono::Duration` does not have that...
So in order to do what I suggested you need to import lazy_static crate and make
```rust
lazy_static! {
    static ref DEFAULT_TOKEN_DURATION: chrono::Duration = chrono::Duration::minutes(60);
}
```
or just leave it with `_MINS` integer.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -169,48 +188,35 @@ where
     }
 
     fn transform(&self, topic_name: &str) -> Option<String> {
-        for mapper in &self.topic_mappers {
-            let forward_topic = mapper
+        self.topic_mappers.iter().find_map(|mapper| {
```
dmolokanov, line 191: Cool! I didn't know about https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -52,19 +65,18 @@ where
...
-                password = match ts.get(&expiry).await {
-                    Ok(x) => Some(x),
-                    Err(e) => {
-                        warn!("Failed to get token {}", e);
-                        None
-                    }
-                }
+            let expiry = Utc::now() + chrono::Duration::minutes(DEFAULT_TOKEN_DURATION_MINS);
+
+            let password: Option<String> = if let Some(ts) = token_source {
```
dmolokanov, line 70: nit
```suggestion
            let password = if let Some(ts) = token_source {
```

mqtt/mqtt-bridge/src/token_source.rs
```diff
@@ -14,7 +13,7 @@ define_encode_set! {
 
 #[async_trait]
 pub trait TokenSource {
```
dmolokanov, line 15: Usually, it is better to make an associated type for Error. and then the implementor defines exact error type.

dmolokanov (approved):
Neat. Thanks!

mqtt/mqtt-bridge/src/client.rs
dmolokanov: Yes, but also in addition to what you said, it is not an error, when the bridging client cannot subscribe to a broker, right? lets say it wants to sub to both "foo" and "bar" topics. Even if the broker allows only "foo" and not "bar" the bridge should not stop with an error. I should log an error and continue to work.

bridge behavior in this case is different from what CommandHandler has. CommandHandler errors out when it cannot subscribe to **all** required topics.

PR #3541, opened 2020-09-11
Policy engine enhancements

vadim-kovalyov:
As part of the work to integrate policy engine into MQTT broker, I'm doing some enhancements:
- Added Context as an associated type to the Request. That context will have info needed for proper work of Substituter.
   - In case of MQTT broker, `type Context = auth::Activity`, and `MqttSubstituter` will get access to `Activity` struct to extract identity and other related info.
- Changed Validator to accept the whole statement to make validation more flexible.

This is the first PR in a series of PRs that yet to come. Just wanted to push it in smaller chunks.

### DIFF
--- a/mqtt/policy/src/core/mod.rs
+++ b/mqtt/policy/src/core/mod.rs
@@ -348,6 +348,15 @@ impl From<EffectOrd> for Decision {
     }
 }
 
+impl From<&Statement> for EffectOrd {
+    fn from(statement: &Statement) -> Self {
+        match statement.effect() {
+            builder::Effect::Allow => EffectOrd::new(Effect::Allow, statement.order()),
+            builder::Effect::Deny => EffectOrd::new(Effect::Deny, statement.order()),
+        }
+    }
+}
+
 #[cfg(test)]
 pub(crate) mod tests {
     use super::*;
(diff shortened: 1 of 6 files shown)


dmolokanov (approved):
Overall looks good. What was the reasoning behind adding/removing schema versioning?

vadim-kovalyov:
> Overall looks good. What was the reasoning behind adding/removing schema versioning?

I thought that I can keep `PolicyDefinition` private and use it for versioning. But it looks like there is no good way to have comprehensive validation without exposing whole `PolicyDefinition` to the plugins. Also, it seems like if policy definition struct needs to be changed in the future versions, it would be breaking changes for validators as well. I don't see a better option so far.

mqtt/policy/src/core/mod.rs
```diff
@@ -348,6 +348,15 @@ impl From<EffectOrd> for Decision {
     }
 }
 
+impl From<&Statement> for EffectOrd {
```
dmolokanov, line 351: > Used to do value-to-value conversions while **_consuming_** the input value.
The client of the `Statement` should decide where to clone data before consumption. https://doc.rust-lang.org/std/convert/trait.From.html

dmolokanov (approved):

PR #3542, opened 2020-09-12
[mqtt] Refactor command handler

dmolokanov:
The code related to a command handler used to be in 2 separate modules. There are some changes to consolidate code for command in the one parent module. Each command handler is in its own module and `Command` trait is changed a bit to support custom Error type per command.
Removed unnecessary allocations (usually with strings).
`CommandHandler` is initialized in the bootstrap code.


vadim-kovalyov (approved):
:rocket:

mqtt/mqtt-edgehub/src/command/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+
+pub trait Command {
+    type Error;
+
+    fn topic(&self) -> &str;
+
+    fn handle(&mut self, publication: &ReceivedPublication) -> Result<(), Self::Error>;
+}
+
+pub struct DynCommand<C> {
```
vadim-kovalyov, line 21:
```
/// Wrapper to take any command and wrap error type as a trait object.
///
/// We need this to collect all commands in a `HashMap<dyn Command<Error = Box<dyn StdError + Send>>`.
/// Otherwise there is no good way to bring different commands to the same trait object type with associated trait object error type.
```

mqtt/mqtt-edgehub/src/command/handler.rs
```diff
@@ -70,14 +73,17 @@ impl CommandHandler {
...
-            broker_handle,
+        CommandHandler {
             client,
-            commands,
-        })
+            commands: HashMap::new(),
+        }
+    }
+
+    // TODO refactor and move it inside the [`run`] method
```
dylanbronson, line 82: What does the refactor entail? Why can't it be moved in as is?

dmolokanov: As per discussion with @and-rewsmith. That's what we agreed to do next to avoid sidecars startup order.

vadim-kovalyov (approved):
:+1:

dylanbronson (approved):
Command/handler refactor lgtm

mqtt/mqtt-broker/src/auth/authentication.rs
```diff
@@ -125,7 +125,8 @@ where
 impl<F, E> Authenticator for F
 where
     F: Fn(AuthenticationContext) -> Result<Option<AuthId>, E> + Sync,
-    E: Deref<Target = dyn StdError> + 'static,
+    // E: Deref<Target = dyn StdError> + 'static,
```
and-rewsmith, line 128: Should we remove this comment?

mqtt/mqtt-edgehub/src/command/handler.rs
```diff
@@ -70,14 +73,17 @@ impl CommandHandler {
...
+    pub async fn init(&mut self) -> Result<(), CommandHandlerError> {
```
and-rewsmith, line 83: Why not name this new()?

mqtt/mqtt-edgehub/tests/command.rs
```diff
@@ -1,19 +1,16 @@
-use std::collections::HashMap;
+use std::error::Error as StdError;
```
and-rewsmith, line 1: Do we have tests for Authorized Identities commands?

and-rewsmith (approved):
LGTM with some minor comments / questions. Thanks for doing this Denis :+1: .

mqtt/mqtt-broker/src/auth/authentication.rs
dmolokanov: yes. good catch!

mqtt/mqtt-edgehub/src/command/handler.rs
dmolokanov: `new` - create an instance of the handler, but then it takes commands one by one. only after all commands added we should ask mqtt client to subscribe to all topics.

mqtt/mqtt-edgehub/tests/command.rs
dmolokanov: I don't see any. @dylanbronson may know better

PR #3543, opened 2020-09-12
Report provisioning type in metrics

damonbarry:
Iotedged's management API exposes `/systeminfo`, which returns information about the host, such as iotedged version, OS, architecture, and # CPUs. We also want it to return information about how the device is provisioned, to make it available on Edge agent's Prometheus endpoint (as part of the`host_information` tag in the `edgeAgent_metadata` gauge). This change adds `provisioning` to the `/systeminfo` response, and plumbs it through to the metric. It's an object that looks like:

```
{
  "type": string,
  "dynamicReprovisioning": boolean,  // default: false
  "alwaysReprovisionOnStartup": boolean  // optional, default: true
}
```

The `type` property can have one of the following values:

```
manual.device_connection_string
manual.x509
dps.tpm
dps.symmetric_key
dps.x509
external
```

The `alwaysReprovisionOnStartup` property is optional because it's (currently) only used for DPS provisioning.


edgelet/edgelet-kube/src/runtime.rs
```diff
@@ -237,6 +237,7 @@ where
                             os_type: "Kubernetes".to_string(),
                             architecture: serde_json::to_string(&architectures).unwrap(),
                             version: edgelet_core::version_with_source_version(),
+                            provisioning_type: not_supported.clone(),
```
damonbarry, line 240: I need to find out about the expected behavior on k8s.

CindyXing: We support the same as single node. See https://github.com/Azure/iotedge/blob/master/kubernetes/charts/edge-kubernetes/values.yaml

edgelet/edgelet-core/src/settings.rs
```diff
@@ -424,6 +424,23 @@ pub enum ProvisioningType {
     External(External),
 }
 
+impl ProvisioningType {
+    pub fn name(&self) -> &str {
+        match self {
+            ProvisioningType::Manual(manual) => match manual.authentication_method() {
```
CindyXing, line 430: Can we add the dynamic_provisioning and always_reprovision_on_startup (see https://github.com/Azure/iotedge/pull/3435)?

CindyXing: This should give us a pretty good idea on how customers are provision edge devices.

CindyXing (changes requested):

edgelet/edgelet-kube/src/runtime.rs
damonbarry: Because our k8s support is on a different release timeline and therefore not part of the 1.0.10 release, I'd like to add k8s support as a separate PR.

edgelet/edgelet-core/src/settings.rs
damonbarry: Done.

damonbarry:
I rebased to include @arsing's recent change (#3435) that adds the `always_reprovision_on_startup` setting to config.yaml. I also updated my PR to log this setting in our metrics.

edgelet/edgelet-kube/src/runtime.rs
CindyXing: Sounds good. K8s is not needed for now. We should confirm with Venkat.

damonbarry: @veyalla?

CindyXing (dismissed):

edgelet/edgelet-core/src/module.rs
```diff
@@ -366,12 +366,33 @@ pub struct SystemInfo {
...
     pub server_version: String,
     pub kernel_version: String,
     pub operating_system: String,
     pub cpus: i32,
 }
 
+#[derive(Clone, Debug, Default, Serialize)]
+pub struct ProvisioningInfo {
+    /// IoT Edge provisioning type, examples: manual.device_connection_string, dps.x509
+    pub r#type: String,
```
lfitchett, line 379: why is this r#type instead of type?

lfitchett (approved):
LGTM

edgelet/edgelet-core/src/module.rs
damonbarry: Because `type` is a reserved word.

arsing: (The alternative is `#[serde(rename = "type")] pub type_: String`, but since we have edition 2018 we might as well take advantage of it.)

edgelet/edgelet-kube/src/runtime.rs
veyalla: Agree with Cindy, k8s not needed for now.

arsing (dismissed):
Signed off for the Rust part.

lfitchett (approved):

lfitchett (approved):

CindyXing (approved):

PR #3549, opened 2020-09-14
[mqtt] RocksDB Persistence

and-rewsmith:
Adding support for rocksDB persistence. We previously only had in-memory persistence.

Previously, the plan was to have a persistor interface for in-memory and disk persistence implementations. I refactored the module structure to move the interface from the persistor to the waking_state. This way, we can better re-use code in that we only need one persistor/loader implementation.

So now it is:
```
persist/
--loader
--persistor
--waking_state/
------waking_map (in-memory)
------waking_store (rocksdb)
```


mqtt/.gitignore
```diff
@@ -1 +1,2 @@
 state*.dat
+unit-tests/
```
and-rewsmith, line 2: This is where the rocksdb unit tests dump their state. It isn't getting cleaned up after the test. Should we have a step to explicitly remove this folder after the tests run?

mqtt/mqtt-bridge/src/lib.rs
```diff
@@ -8,7 +8,8 @@
     clippy::use_self,
     clippy::match_same_arms,
     clippy::must_use_candidate,
-    clippy::missing_errors_doc
+    clippy::missing_errors_doc,
+    dead_code
```
and-rewsmith, line 12: Since the persistor/loader/waking_state isn't used yet, is this ```dead_code``` necessary?

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -1,36 +1,28 @@
...
-
-    async fn push(&mut self, message: Publication) -> Result<Key, Self::Error>;
-
-    async fn remove(&mut self, key: Key) -> Option<Publication>;
-
-    async fn loader(&'a mut self) -> Arc<Mutex<Self::Loader>>;
-}
+pub mod loader;
+pub mod persistor;
+pub mod waking_state;
```
and-rewsmith, line 8: These should all be pub right?

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -1,36 +1,28 @@
...
 }
 
+#[derive(Debug, Error)]
+pub enum PersistError {
+    #[error("Failed to serialize on database insert")]
+    Serialization(#[from] Box<ErrorKind>),
+
+    #[error("Failed to serialize on database insert")]
+    Insertion(#[from] Error),
+}
```
and-rewsmith, line 24: I talked to Vadim and we decided to have this error type live here, then be implemented in the waking_state implementations. This approach (compared to having a WakingStateError and a PersistError) means we don't have to duplicate error types. It comes at the cost of using this PersistError type in the StreamWakeableState implementations.

mqtt/mqtt-bridge/src/persist/mod.rs
and-rewsmith: For context, the way I structured the loader / persistor is that they take a StreamWakeableState implementation for their new() method.

and-rewsmith:
@dmolokanov You missed this last PR discussion, but is there any way to avoid mutex on the loader? This PR is a good opportunity to address this if there is a workaround.

mqtt/mqtt-bridge/src/lib.rs
and-rewsmith: I saw this pattern in Anca's PR, but I can add this allow to the top of specific files. Any objections?

mqtt/mqtt-bridge/src/persist/mod.rs
and-rewsmith: Anca and I discussed and we thought of a better approach. If we construct the waking_state implementation from inside the persistor, then we don't need to expose it (since it is just an implementation detail leaking out).

I will add a setting for use persistence. Based on this setting I will construct the appropriate waking_state. I am working on a separate PR for handling all persistence settings, so I am fine merging it as is.

and-rewsmith: Actually, I disagree with myself. I will change it now.

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -4,32 +4,34 @@ use futures_util::stream::Stream;
...
+pub struct MessageLoader<S: StreamWakeableState> {
+    state: Arc<Mutex<S>>,
     batch: IntoIter<(Key, Publication)>,
     batch_size: usize,
 }
 
-impl InMemoryMessageLoader {
-    pub async fn new(state: Arc<Mutex<WakingMap>>, batch_size: usize) -> Self {
+impl<S: StreamWakeableState> MessageLoader<S> {
+    pub async fn new(state: Arc<Mutex<S>>, batch_size: usize) -> Self {
```
and-rewsmith, line 23: remove async

mqtt/mqtt-bridge/src/lib.rs
dmolokanov: lets move it in modules that contain dead code instead? please put a TODO to not to forget remove it later

mqtt/mqtt-bridge/src/persist/persistor.rs
```diff
@@ -1,46 +1,34 @@
...
 
-#[async_trait]
-impl<'a> Persist<'a> for InMemoryPersist {
-    type Loader = InMemoryMessageLoader;
-    type Error = Error;
-
-    async fn new(batch_size: usize) -> Self {
-        let state = WakingMap::new();
+impl<S: StreamWakeableState> Persistor<S> {
+    async fn new(state: S, batch_size: usize) -> Self {
```
dmolokanov, line 18: It feels like you don't need any methods to be `async` right?

mqtt/mqtt-bridge/src/persist/persistor.rs
```diff
@@ -1,46 +1,34 @@
...
         let state = Arc::new(Mutex::new(state));
-
-        let loader = InMemoryMessageLoader::new(Arc::clone(&state), batch_size).await;
+        let loader = MessageLoader::new(Arc::clone(&state), batch_size).await;
         let loader = Arc::new(Mutex::new(loader));
 
         let offset = 0;
-        InMemoryPersist {
+        Persistor {
```
dmolokanov, line 24:
```suggestion
Self
```

mqtt/mqtt-bridge/src/persist/persistor.rs
```diff
@@ -66,7 +54,7 @@ impl<'a> Persist<'a> for InMemoryPersist {
         state_lock.remove_in_flight(&key)
     }
 
-    async fn loader(&'a mut self) -> Arc<Mutex<InMemoryMessageLoader>> {
+    async fn loader(&mut self) -> Arc<Mutex<MessageLoader<S>>> {
         Arc::clone(&self.loader)
```
dmolokanov, line 58: Shouldn't we create loader in this methods instead of constructor?

mqtt/mqtt-bridge/src/persist/mod.rs
dmolokanov: you can re-import as `pub` only those types as required, instead of making all modules public

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -52,7 +54,7 @@ impl Stream for InMemoryMessageLoader {
 }
 
 fn get_elements(
```
dmolokanov, line 56: `get_elements()` -> `load_batch()` ?

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -52,7 +54,7 @@ impl Stream for InMemoryMessageLoader {
...
-    state: &mut MutexGuard<'_, WakingMap>,
+    state: &mut MutexGuard<'_, impl StreamWakeableState>,
     batch_size: usize,
 ) -> IntoIter<(Key, Publication)> {
     let batch: Vec<_> = state
```
dmolokanov, line 60: you can do tuple deconstruction in closures as well `.map(|(key, publication)| (key.clone(), publication.clone()))`

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -4,32 +4,34 @@ use futures_util::stream::Stream;
...
+    state: Arc<Mutex<S>>,
     batch: IntoIter<(Key, Publication)>,
     batch_size: usize,
 }
 
-impl InMemoryMessageLoader {
-    pub async fn new(state: Arc<Mutex<WakingMap>>, batch_size: usize) -> Self {
+impl<S: StreamWakeableState> MessageLoader<S> {
+    pub async fn new(state: Arc<Mutex<S>>, batch_size: usize) -> Self {
         let batch: IntoIter<(Key, Publication)> = Vec::new().into_iter();
```
dmolokanov, line 24: this IntoIter looks a bit weird. what you could do is to use VecDeque. and instead of calling `self.batch.next()` you could do `self.batch.pop_back()`

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
```diff
@@ -0,0 +1,67 @@
...
+        self.queue.push_back((key, value));
+
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
+        }
+
+        Ok(())
+    }
+
+    fn get(&mut self, count: usize) -> Vec<(Key, Publication)> {
```
dmolokanov, line 42: get usually means get *one* item from the collection. `fn batch()` sounds better

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
```diff
@@ -0,0 +1,67 @@
...
+        let count = min(count, self.queue.len());
+        let mut output = vec![];
+        for _ in 0..count {
+            let removed = self.queue.pop_front();
+
+            if let Some(pair) = removed {
+                output.push((pair.0.clone(), pair.1.clone()));
+                self.in_flight.insert(pair.0, pair.1);
+            } else {
+                error!("failed retrieving message from persistence");
```
dmolokanov, line 52: if we got a None here, we have an error in the thought process. In this case it is safer to fail fast.

dmolokanov: it is safer to use combinators or split_off function?
```rust
let count = min(count, self.queue.len());
let output = self.queue;
self.queue = output.split_off(count); // split_off leaves [0..count) in the output, but we want the opposite
output
```

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
```diff
@@ -0,0 +1,97 @@
...
+/// Only when elements are removed from the in-flight collection they will be removed from the store.
+pub struct WakingStore {
+    db: DB,
+    in_flight: HashMap<Key, Publication>,
+    waker: Option<Waker>,
+}
+
+impl WakingStore {
+    pub fn new(db: DB) -> Self {
+        let in_flight = HashMap::new();
```
dmolokanov, line 24: why not inline these two?

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
```diff
@@ -0,0 +1,97 @@
...
+            db,
+            in_flight,
+            waker,
+        }
+    }
+}
+
+impl StreamWakeableState for WakingStore {
+    fn insert(&mut self, key: Key, value: Publication) -> Result<(), PersistError> {
+        let key_bytes = serialize(&key).map_err(PersistError::Serialization)?;
```
dmolokanov, line 37: bincode::serialize

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use crate::persist::{waking_state::StreamWakeableState, Key, PersistError};
+
+/// Responsible for waking waiting streams when new elements are added.
+///
+/// Exposes a get method for retrieving a count of elements in order of insertion.
+/// When elements are retrieved they are added to the in flight collection, but kept in the original db store.
+/// Only when elements are removed from the in-flight collection they will be removed from the store.
+pub struct WakingStore {
+    db: DB,
```
dmolokanov, line 17: initial idea was we have only on DB per all bridges. For each bridge there will be 2 collections of publication in and out. with this approach you have right now for each bridge there will be a separate DB which seems like overkill.

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
+        }
+
+        Ok(())
+    }
+
+    /// Get count elements of store, exluding those that are already in in-flight
+    fn get(&mut self, count: usize) -> Vec<(Key, Publication)> {
```
dmolokanov, line 51: I think it should return Result and propagade error up the callstack

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
```diff
@@ -0,0 +1,97 @@
...
+        let iter = self.db.iterator(IteratorMode::Start);
+        let mut output = vec![];
+
+        for (iterations, extracted) in iter.enumerate() {
```
dmolokanov, line 55: `iter.take(count).map(...).collect()` instead of loops?

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
dmolokanov: something lile this
```rust
fn get(&mut self, count: usize) -> Result<Vec<(Key, Publication)>, PersistError> {
        let iter = self.db.iterator(IteratorMode::Start);

        iter.map(|(key, publication)| {
            bincode::deserialize(&key)
                .map_err(PersistError::Serialization)
                .and_then(|key| {
                    bincode::deserialize(&publication)
                        .map_err(PersistError::Serialization)
                        .map(|publication| (key, publication))
                })
        })
        .collect()
    }
        ```

File: mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs [hunk 19]
```diff
@@ -0,0 +1,97 @@
+use std::collections::hash_map::Entry;
+use std::{boxed::Box, collections::HashMap, task::Waker};
+
+use bincode::{deserialize, serialize, ErrorKind};
+use mqtt3::proto::Publication;
```
dmolokanov, line 5: https://github.com/birktj/rocksbin-db this crate seems small and neat. what do you think?

dmolokanov (changes requested):
Good work! I have some additional thoughts

mqtt/mqtt-bridge/src/persist/persistor.rs
and-rewsmith: We discussed this in the last PR. The problem with this is that it allow multiple loaders to exist, which the waking logic is not set up to deal with.

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
and-rewsmith: I agree that we should share the db. So we need to wrap in a mutex?

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
and-rewsmith: Regarding the first approach, it won't work as is because the creation of ```let output``` cannot move queue out of self. Also, we will need to return Vec to fit the interface so I think we will have to convert from VecDeque to Vec which might be expensive?

I chose to go with the second suggestion.

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
and-rewsmith: If we go with this approach then we can't deserialize anything if there is a failure. I think that you are correct though in that if this is a logic error it shouldn't matter. We should percolate up and blow up.

and-rewsmith: The issue though is that this batch function is used in the loader. The loader implements stream, and this batch() func is used in the poll_next() impl for stream. 

Since poll_next doesn't return result, it cannot get percolated all the way up. Should we still return a Result here and log an error in the stream implementation? I don't see an easy way we can blow up without a bigger change to the structure.

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::persist::{waking_state::StreamWakeableState, Key, PersistError};
+
+/// Responsible for waking waiting streams when new elements are added.
+///
+/// Exposes a get method for retrieving a count of elements in order of insertion.
+/// When elements are retrieved they are moved to the in flight collection.
+pub struct WakingMap {
+    queue: VecDeque<(Key, Publication)>,
+    in_flight: HashMap<Key, Publication>,
+    waker: Option<Waker>,
```
ancaantochi, line 14: I think it shouldn't be possible to create a WakingMap without Waker. Why is it an option?

and-rewsmith: A waker is set when a stream needs to be waked. When the waking map is created, then at that point there won't be any waiting streams, and thus no waker.

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
and-rewsmith: You might have more developed opinions here, but my intuition is that using unpopular crates is bad from a security perspective.

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -1,36 +1,33 @@
...
     offset: u32,
 }
 
+#[derive(Debug, Error)]
+pub enum PersistError {
+    #[error("Failed to serialize on database insert")]
+    Serialization(#[source] Box<ErrorKind>),
+
+    #[error("Failed to deserialize database entry")]
+    Deserialization(#[source] Box<ErrorKind>),
```
and-rewsmith, line 25: From looking at the docs, I don't understand the difference between from and source.

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
and-rewsmith: Since DB is already Send / Sync I will use an Arc.

and-rewsmith: Although now that we are reusing the same DB struct, there are other things I need to change. For example:
- how do we handle offset
- how do we only get/insert elements for the store in question

and-rewsmith: We can have a column family for each store based on a generated uuid. Offset will be specific to each store as described in the spec.

mqtt/mqtt-bridge/src/persist/waking_state/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+                break;
+            }
+        }
+
+        Ok(output)
+    }
+
+    fn remove_in_flight(&mut self, key: &Key) -> Option<Publication> {
+        let key_bytes = bincode::serialize(&key).unwrap();
+        self.db.delete(key_bytes).unwrap();
```
and-rewsmith, line 139: I forgot to remove this unwrap

mqtt/mqtt-bridge/src/persist/waking_state/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+/// Responsible for waking waiting streams when new elements are added.
+///
+/// Exposes a get method for retrieving a count of elements in order of insertion.
+#[async_trait]
+pub trait StreamWakeableState {
+    fn insert(&mut self, key: Key, value: Publication) -> Result<(), PersistError>;
+
+    fn batch(&mut self, count: usize) -> Result<Vec<(Key, Publication)>, PersistError>;
+
+    fn remove_in_flight(&mut self, key: &Key) -> Result<Publication, PersistError>;
```
and-rewsmith, line 25: I converted this interface to return result in all cases. Previously we aligned with standard collection types and returned Option. I feel like being consistent with error is the best choice because we should never attempt to remove anything that doesn't exist.

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use crate::persist::{waking_state::StreamWakeableState, Key, PersistError};
+
+/// When elements are retrieved they are added to the in flight collection, but kept in the original db store.
+/// Only when elements are removed from the in-flight collection they will be removed from the store.
+pub struct WakingStore {
+    db: DB,
+    in_flight: HashMap<Key, Publication>,
+    waker: Option<Waker>,
+    column_family: String,
```
and-rewsmith, line 18: I tried to store the ColumnFamily struct but ran into issues. You get it back as a reference from the rocksdb api and there are difficulties moving it around. Adding lifetimes isn't enough as WakingStore needs to be Send/Sync so we would need to wrap in a mutex.

I decided it was better to store the information necessary to access the column family, rather than the &ColumnFamily itself.

mqtt/.gitignore
dmolokanov: yeah, probably it is a good idea. take a look at the https://docs.rs/tempfile/3.1.0/tempfile/

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
dmolokanov: good question. if we have an error on this step it is either someone forged the database or we the code didn't expect database schema change. in this case, we should fail fast and loud

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
dmolokanov: this is correct. but in this case, the library is so small and looks like does exactly what we need. it is tempting to take it :)
Alternatively, I believe we should make similar abstraction to support serde for rocksdb

mqtt/mqtt-bridge/src/persist/mod.rs
dmolokanov: `source` only generates code to return marked value from `Error::source` method
`from` in addition to what `source` does it generates conversion `impl From<SomeError> for Error` as well

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -1,48 +1,69 @@
...
             batch_size,
         }
     }
+
+    fn next_batch(&mut self) -> Result<VecDeque<(Key, Publication)>, PersistError> {
+        let mut state_lock = self.state.lock();
+        let batch: VecDeque<_> = state_lock
+            .batch(self.batch_size)?
+            .iter()
+            .map(|(key, publication)| (key.clone(), publication.clone()))
```
dmolokanov, line 45: it feels like we have unneccessary cloning here. Lets make `StreamWakeableState::batch` to return VecDeque it use it without cloning?
`iter::collect` can collect to `VecDeque` as well, so minimal code changes

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl WakingMap {
+    pub fn new() -> Self {
+        let queue: VecDeque<(Key, Publication)> = VecDeque::new();
```
dmolokanov, line 18: nit:
```suggestion
        WakingMap {
            queue: VecDeque::new(),
            in_flight: HashMap::new(),
            waker: None,
        }
```

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
```diff
@@ -0,0 +1,57 @@
...
+            waker.wake();
+        }
+
+        Ok(())
+    }
+
+    fn batch(&mut self, count: usize) -> Result<Vec<(Key, Publication)>, PersistError> {
+        let count = min(count, self.queue.len());
+        let output: Vec<_> = self.queue.drain(..count).collect();
+        self.in_flight.extend(output.clone().into_iter());
```
dmolokanov, line 43: this looks suspicious. why all items from the next batch is automatically going to inflight?

dmolokanov: usually inflilght queue is 16, but batch size can be 100

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
```diff
@@ -0,0 +1,57 @@
...
+
+        Ok(output)
+    }
+
+    fn remove_in_flight(&mut self, key: &Key) -> Result<Publication, PersistError> {
+        self.in_flight
+            .remove(key)
+            .ok_or(PersistError::RemovalForMissing())
```
dmolokanov, line 51: nit
```suggestion
            .ok_or(PersistError::RemovalForMissing)
```

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -1,34 +1,43 @@
...
+    #[error("Failed to create rocksdb column family")]
+    CreateColumnFamily(#[source] Error),
 
-    async fn new(batch_size: usize) -> Self;
+    #[error("Failed to deserialize database entry")]
+    Deserialization(#[source] Box<ErrorKind>),
 
-    async fn push(&mut self, message: Publication) -> Result<Key, Self::Error>;
+    #[error("Failed to get rocksdb column family")]
+    GetColumnFamily(),
```
dmolokanov, line 28: nit 
```suggestion
    GetColumnFamily,
```

mqtt/mqtt-bridge/src/persist/persistor.rs
```diff
@@ -1,46 +1,53 @@
...
-    loader: Arc<Mutex<InMemoryMessageLoader>>,
+    loader: Arc<Mutex<MessageLoader<S>>>,
 }
 
-#[async_trait]
-impl<'a> Persist<'a> for InMemoryPersist {
-    type Loader = InMemoryMessageLoader;
-    type Error = Error;
+impl Persistor<WakingMap> {
+    fn new_memory(batch_size: usize) -> Persistor<WakingMap> {
```
dmolokanov, line 24:
```suggestion
    pub fn new_memory(batch_size: usize) -> Persistor<WakingMap> {
```
?

mqtt/mqtt-bridge/src/persist/persistor.rs
```diff
@@ -1,46 +1,53 @@
...
+        Self::new(WakingMap::new(), batch_size)
+    }
+}
 
-    async fn new(batch_size: usize) -> Self {
-        let state = WakingMap::new();
-        let state = Arc::new(Mutex::new(state));
+impl Persistor<WakingStore> {
+    fn new_disk(db: DB, batch_size: usize) -> Result<Persistor<WakingStore>, PersistError> {
```
dmolokanov, line 30:
```suggestion
    pub fn new_disk(db: DB, batch_size: usize) -> Result<Persistor<WakingStore>, PersistError> {
```
?

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
```diff
@@ -0,0 +1,57 @@
+#![allow(dead_code)] // TODO remove when ready
```
dmolokanov, line 1: mqtt_bridge::persist::waking_state::waking_map seems a bit too much of waking. lets rename
`waking_map` -> `memory`
`waking_store` -> `rocksdb`
?

dmolokanov: AFAIU Bridge will work with `Persistor<S>`, right? Let's rename `Persistor` to `Store` or `PublicationStore` or something else? Persistor has a bit more low-level meaning for me at least

dmolokanov (commented):
Great work! Some naming suggestions

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
```diff
@@ -0,0 +1,109 @@
...
+        })
+    }
+}
+
+impl StreamWakeableState for WakingStore {
+    fn insert(&mut self, key: Key, value: Publication) -> Result<(), PersistError> {
+        let key_bytes = bincode::serialize(&key).map_err(PersistError::Serialization)?;
+        let publication_bytes = bincode::serialize(&value).map_err(PersistError::Serialization)?;
+
+        let column_family = self
```
dmolokanov, line 41: getting a column family handle is the cheap operation?

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
```diff
@@ -0,0 +1,109 @@
...
+pub struct WakingStore {
+    db: DB,
+    in_flight: HashMap<Key, Publication>,
+    waker: Option<Waker>,
+    column_family: String,
+}
+
+impl WakingStore {
+    pub fn new(mut db: DB) -> Result<Self, PersistError> {
+        let column_family = Uuid::new_v4().to_string();
```
dmolokanov, line 23: I think `new` should take the column family name as an argument

mqtt/mqtt-bridge/Cargo.toml
```diff
@@ -7,24 +7,28 @@ edition = "2018"
...
 parking_lot = "0.11.0"
+rocksdb = "0.15.0"
 mqtt3 = { path = "../mqtt3", features = ["serde1"] }
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_derive = "1.0"
 serial_test = "0.4"
 thiserror = "1.0"
 tracing = "0.1"
 tokio = { version = "0.2", features = ["sync"] }
+uuid = { version = "0.8", features = ["v4"] }
```
dmolokanov, line 25: why?

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
dmolokanov: lets extract getting a ColumnFamily into private method?

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
and-rewsmith: Yes good point. They should only go in flight once actually returned by the loader. So by this:
> usually inflilght queue is 16, but batch size can be 100

Do you mean that we should fail a loader.next() if we have 16 already in flight?

mqtt/mqtt-bridge/src/persist/persistor.rs
and-rewsmith: I need to make all of them pub

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
and-rewsmith: I assumed it is. Here is the documentation:
https://docs.rs/rocksdb/0.15.0/rocksdb/struct.DB.html#method.cf_handle

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
and-rewsmith: What is the problem with generating a new uuid inside new?

mqtt/mqtt-bridge/Cargo.toml
and-rewsmith: generating new column families for rocksdb databases

mqtt/mqtt-bridge/src/persist/waking_state/memory.rs
```diff
@@ -0,0 +1,58 @@
...
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
+        }
+
+        Ok(())
+    }
+
+    fn batch(&mut self, count: usize) -> Result<VecDeque<(Key, Publication)>, PersistError> {
+        let count = min(count, self.queue.len());
+        let output: VecDeque<_> = self.queue.drain(..count).collect();
```
and-rewsmith, line 43: I just looked at the docs for drain and see that it can panic. Although it should never do this because we guarantee count is at most the length of the queue, it makes me scared. Should we change this approach?

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
and-rewsmith: Ok that makes sense. Should we panic? Also, I remember there was a macro for failing with a message? I looked around and can't find it.

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
and-rewsmith: I am thinking it is better to write a simple wrapper ourselves. If we could trust the package to not have bugs or update rocksdb (which looks to be using an old version) then I would use it. Also, the ```unwarp()``` spelling in the README makes me quite suspect

mqtt/mqtt-bridge/src/persist/waking_state/rocksdb.rs
```diff
@@ -0,0 +1,110 @@
...
+        Ok(())
+    }
+
+    /// Get count elements of store, exluding those that are already in in-flight
+    fn batch(&mut self, count: usize) -> Result<VecDeque<(Key, Publication)>, PersistError> {
+        let column_family = self.column_family()?;
+        let iter = self.db.iterator_cf(column_family, IteratorMode::Start);
+
+        let mut output = VecDeque::new();
+        for (iterations, extracted) in iter.enumerate() {
```
and-rewsmith, line 69: clippy suggested to refactor, but introduced a bug. Iterations should only increment when we find a non-in-flight and add to the output VecDeque

and-rewsmith: I am refactoring this now to have the serialization logic in a DB-wrapping abstraction. I will fix it then.

mqtt/mqtt-bridge/src/persist/waking_state/waking_map.rs
and-rewsmith: I think changing this will have negative perf implications. Previously, the could avoid many locks of the underlying state because it could process off of its current batch (which was already marked in-flight). If we change to add to in-flight only once loaded, then the loader has to lock the shared state every time to see if in_flight has enough remaining capacity. So we can either:

**Option A**
Keep the current approach. This will have better perf, but has less stringent enforcement of what can be removed from the underlying state. For example, it is possible to remove an element once it is in the loader's next batch, but not returned by a ```loader.next()``` call. 

With this approach, batch size and in-flight size are the same.

**Option B**
Switch to remove in-flight only once returned from the loader. This will require locking the shared state to check the in-flight size on every ```loader.next()``` call.

With this approach, batch size and in-flight size can be configured differently. However, batch size becomes less meaningful because it won't save locks. Previously, we only had to lock when we got a new batch. Now we have to lock every time to check the in-flight size.

mqtt/mqtt-bridge/src/persist/waking_state/memory.rs
vadim-kovalyov: Any alternatives? for me this is fine as long as we guarantee the count is at most the length of the queue

mqtt/mqtt-bridge/src/persist/waking_state/memory.rs
```diff
@@ -0,0 +1,56 @@
...
+        self.in_flight.extend(output.clone().into_iter());
```
vadim-kovalyov, line 42: nit: this is a hot path. Would it be better to iterate over Drain and insert into both output and in_flight instead of extra cloning of VecDeque?

mqtt/mqtt-bridge/src/persist/waking_state/rocksdb.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        let column_family = self.column_family()?;
+        self.db
+            .put_cf(column_family, key_bytes, publication_bytes)
+            .map_err(PersistError::Insertion)?;
+
+        Ok(())
+    }
+
+    pub fn iter_except(
```
vadim-kovalyov, line 97: nit: no need to be pub

mqtt/mqtt-bridge/src/persist/waking_state/rocksdb.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Get count elements of store, exluding those that are already in in-flight
+    fn batch(&mut self, count: usize) -> Result<VecDeque<(Key, Publication)>, PersistError> {
+        let iter = self
+            .db
+            .iter_except(count, &self.in_flight.keys().collect())?;
```
vadim-kovalyov, line 48: I would just pass HashMap and use `contains_key`, to save some cpu cycles on creating a separate HashSet.
```suggestion
            .iter_except(count, &self.in_flight)?;
```

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -1,34 +1,46 @@
...
+    memory::WakingMemoryStore, rocksdb::WakingRocksDBStore, StreamWakeableState,
+};
 
-use async_trait::async_trait;
-use futures_util::stream::Stream;
-use mqtt3::proto::Publication;
-use parking_lot::Mutex;
+/// Keys used in persistence.
+/// Ordered by offset
+#[derive(Hash, Eq, Ord, PartialOrd, PartialEq, Clone, Debug, Deserialize, Serialize)]
```
vadim-kovalyov, line 17: Key can be made `Copy`

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -1,34 +1,46 @@
...
-/// Keys used in persistence.
-/// Ordered by offset
-#[derive(Hash, Eq, Ord, PartialOrd, PartialEq, Clone, Debug)]
-pub struct Key {
-    offset: u32,
+    #[error("Failed to serialize on database insert")]
+    Serialization(#[source] Box<ErrorKind>),
 }
 
 #[cfg(test)]
```
vadim-kovalyov, line 46: we simplified the Key for now, probably these tests are no longer needed.

vadim-kovalyov (commented):
Looks good to me! had some nit questions.

mqtt/mqtt-bridge/src/persist/waking_state/memory.rs
and-rewsmith: You are saying this is better because rather than iterating over drain twice like we are doing here, we only have to iterate once? If there is a sneaky optimization with the clone() function, this new approach might be less performant. I'm not sure so I'll do whatever you think it is most likely best.

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
dmolokanov: do you ask about https://doc.rust-lang.org/std/result/enum.Result.html#method.expect ?

and-rewsmith: No I remember seeing something different. Is it ok to use expect here?

mqtt/mqtt-bridge/src/persist/waking_state/memory.rs
vadim-kovalyov: we can keep it as is for now, and maybe play with it when we have bench tests.

dmolokanov: I also not sure what @vadim-kovalyov mentioned because we still need a copy in both places: current batch to iterate and inflight messages to avoid dups.

vadim-kovalyov: It's probably better describe with the code :)
```rust
    fn batch(&mut self, count: usize) -> Result<VecDeque<(Key, Publication)>, PersistError> {
        let count = min(count, self.queue.len());
        let mut output = VecDeque::with_capacity(count);
        for (key, publication) in self.queue.drain(..count) {
            self.loaded.insert(key, publication.clone());
            output.push_back((key, publication));
        }

        Ok(output)
    }
```

vadim-kovalyov: but also, why do we need to keep the publications in `self.loaded`? isn't just keys enough?

and-rewsmith: It is there to facilitate returning the publication when ```remove() -> Publication``` is called. It is only absolutely necessary for the in-memory case. We could simplify it to just keys for the disk case though.

Thinking more about how we are going to use the persistor, I think the ```remove()``` method can return void. I'm going to make this change. This means we will only need the keys like you are suggesting.

mqtt/mqtt-bridge/src/persist/waking_state/waking_store.rs
and-rewsmith: I am using expect

mqtt/mqtt-bridge/src/persist/publication_store.rs
```diff
@@ -1,46 +1,56 @@
-use std::{io::Error, sync::Arc};
+#![allow(dead_code)] // TODO remove when ready
```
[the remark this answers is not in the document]
and-rewsmith: Yes because we aren't using the rocksdb persistor yet.

vadim-kovalyov (dismissed):
Overall looks good to me! Great quality and I really like great test and comments coverage.

dmolokanov (approved):
LGTM!

PR #3551, opened 2020-09-15
[mqtt] Re-evaluate subscriptions on authorization update

dylanbronson:
This PR fills in the update method for the Authorizer and re-evaluates all subscriptions on any update.

When edgeHub sends an Authorized_Identities command to the broker, Authorization's update method will be called. The Authorized_Identities command contains a list of Service_Identities that have been authorized by edgeHub. The identities are cached in the Authorizer. They simply replace the Authorizer's current cache. 

Then, we must re-evaluate the current subscriptions, to see if the subscriptions are still valid (We omit connections here because we assume that connections will disconnect on a disconnect signal that originated on the edgeHub side). If they are valid, do nothing, but if a subscription is invalid, drop that client's connection


mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,44 @@ where
         Ok(self.snapshot())
     }
 
+    fn reevaluate_subscriptions(&mut self) -> Result<(), Error> {
+        let mut client_ids_to_remove: Vec<ClientId> = Vec::new();
+        for session in &self.sessions {
```
vadim-kovalyov, line 109: you can deconstruct the item right away to avoid things like this `session.1...`
```suggestion
       for (client_id, session) in &self.sessions {
```

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,44 @@ where
...
+            for sub in session.1.subscriptions().clone().unwrap().values() {
```
vadim-kovalyov, line 110: as a rule, we must not use `.unwrap()` in prod code. You need to filter our Disconnecting sessions. You can probably use `filter` combinator for it, or match. Whatever feels better for you.

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,44 @@ where
...
+                let client_info = session.1.client_info()?.clone();
+                let operation = Operation::new_subscribe(proto::SubscribeTo {
+                    topic_filter: sub.filter().clone().to_string(),
+                    qos: *sub.max_qos(),
+                });
+                let activity = Activity::new(session.0.clone(), client_info, operation);
+                match self.authorizer.authorize(activity) {
+                    Ok(Authorization::Allowed) => {
+                        debug!("client {} successfully reauthorized", session.0);
```
vadim-kovalyov, line 119: nit: I would use the same log messages we use in normal flow. I would just add another log entry at the beginning of the method.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -195,6 +233,24 @@ pub struct ServiceIdentity {
...
+impl ServiceIdentity {
+    pub fn identity(&self) -> String {
+        self.identity.clone()
+    }
+    pub fn auth_chain(&self) -> Option<String> {
+        self.auth_chain.clone()
+    }
+}
+
+impl Clone for ServiceIdentity {
```
vadim-kovalyov, line 245: you can just add `Clone` to derive on line 227.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -119,6 +121,38 @@ impl EdgeHubAuthorizer {
...
+
+    fn check_authorized_cache(&self, client_id: &ClientId, topic: &str) -> bool {
+        let on_behalf_of_id: &str = match topic.split('/').collect::<Vec<&str>>().get(1) {
+            Some(id) => {
+                id
+            }
+            None => {
+                return false
+            }
+        };
```
vadim-kovalyov, line 133: I would extract this into a separate func `get_on_behalf_of_id()`. That logic might change later.

vadim-kovalyov (commented):
I think you are on a right track.

mqtt/mqtt-broker/src/broker.rs
dylanbronson: There is different behavior here, though.
If a client originally tries to subscribe and is forbidden, they can move on and try other subscriptions.
But if a client reauthorization fails for a subscription, the whole client must be disconnected. Since it's different behavior, I think different logs would be helpful

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -88,7 +88,11 @@ where
                         }
                         SystemEvent::AuthorizationUpdate(update) => {
                             if let Err(e) = self.authorizer.update(update) {
-                                warn!(message = "an error occurred updating authorization info", error = %e);
+                                warn!(message = "an error occurred while updating authorization info", error = %e);
```
vadim-kovalyov, line 91: nit: if we are going to shutdown the broker, let's treat this log entry as `error!`

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,52 @@ where
         Ok(self.snapshot())
     }
 
+    fn reevaluate_subscriptions(&mut self) -> Result<(), Error> {
+        let mut client_ids_to_remove: Vec<ClientId> = Vec::new();
+        for (client_id, session) in &self.sessions {
+            for sub in session
+                .subscriptions()
+                .into_iter()
+                .flat_map(std::collections::HashMap::values)
```
vadim-kovalyov, line 113: nit: 
```suggestion
                .flat_map(HashMap::values)
```

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,52 @@ where
...
+            {
+                let client_info = session.client_info()?.clone();
```
vadim-kovalyov, line 115: this can error out for offline sessions, but at the same time we need to have client_info to perform authorization... @dmolokanov, do you think we can save "last used" client info in offline session?

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,52 @@ where
...
+                let sub_topic_filter = sub.filter().clone().to_string();
```
vadim-kovalyov, line 116:
```suggestion
                let sub_topic_filter = sub.filter().to_string();
```

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -119,6 +122,64 @@ impl EdgeHubAuthorizer {
...
+                None => None,
+            },
+            None => match device_id {
+                Some(id) => Some((*id).to_string()),
+                None => None,
+            },
+        }
+    }
+
+    fn check_authorized_cache(&self, client_id: &ClientId, topic: &str) -> bool {
```
vadim-kovalyov, line 155: I would restructure it like that. I believe it would be easier to follow the logic:
```rust
    fn check_authorized_cache(&self, client_id: &ClientId, topic: &str) -> bool {
        if let Some(on_behalf_of_id) = EdgeHubAuthorizer::get_on_behalf_of_id(topic) {
            if client_id.as_str() == on_behalf_of_id {
                self.service_identities_cache
                    .contains_key(client_id.as_str())
            } else {
                match self.service_identities_cache.get(&on_behalf_of_id) {
                    Some(service_identity) => {
                        match service_identity.auth_chain() {
                            Some(auth_chain) => {
                                // if auth_chain for on_behalf_of_id contains client_id, then on_behalf_of_id is a descendant of client_id
                                auth_chain.contains(client_id.as_str())
                            }
                            None => false,
                        }
                    }
                    None => false,
                }
            }
        } else {
            // If there is no on_behalf_of_id, we are dealing with a legacy topic
            // The client_id must still be in the identities cache
            self.service_identities_cache
                .contains_key(client_id.as_str())
        }
    }
```

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -11,6 +11,7 @@ use mqtt_broker::{
 #[derive(Debug, Default)]
 pub struct EdgeHubAuthorizer {
     iothub_allowed_topics: RefCell<HashMap<ClientId, Vec<String>>>,
+    service_identities_cache: HashMap<String, ServiceIdentity>,
```
vadim-kovalyov, line 14: @dmolokanov can you please review this file to make sure we are not breaking anything.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -176,17 +253,24 @@ impl Authorizer for EdgeHubAuthorizer {
...
         let identities = update.as_ref();
-        if let Some(service_identities) = identities.downcast_ref::<ServiceIdentity>() {
-            debug!("{:?}", service_identities);
-            // TODO: fill in update method
+        if let Some(service_identities) = identities.downcast_ref::<Vec<ServiceIdentity>>() {
+            debug!(
+                "service identities update received. Service identities: {:?}",
+                service_identities
+            );
+            let mut service_identity_map: HashMap<String, ServiceIdentity> = HashMap::new();
```
dmolokanov, line 263: nit: 
```suggestion
  self.service_identities_cache = service_identities.iter().map(|id|(id.identity(), id.clone()).collect();
```

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -119,6 +122,64 @@ impl EdgeHubAuthorizer {
...
+            },
+            None => match device_id {
+                Some(id) => Some((*id).to_string()),
+                None => None,
+            },
+        }
+    }
+
+    fn check_authorized_cache(&self, client_id: &ClientId, topic: &str) -> bool {
+        let on_behalf_of_id = if let Some(id) = EdgeHubAuthorizer::get_on_behalf_of_id(topic) {
```
dmolokanov, line 156: the code here is a bit difficult to read. I guess it is because of too many nesting levels. try to flatten it somehow

dmolokanov (changes requested):
Overall looks good. Deeply nested code hurts readability and is error-prone. Try to avoid multiple levels of nesting. https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

mqtt/mqtt-broker/src/broker.rs
dmolokanov: Yeah, it was not the requirement previously, we probably need this info stored

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,52 @@ where
         Ok(self.snapshot())
     }
 
+    fn reevaluate_subscriptions(&mut self) -> Result<(), Error> {
+        let mut client_ids_to_remove: Vec<ClientId> = Vec::new();
+        for (client_id, session) in &self.sessions {
+            for sub in session
```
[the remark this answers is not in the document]
dylanbronson: I think this falls under the category of re-evaluating all connections on auth updates. @vadim-kovalyov and I have discussed this and landed on not re-evaluating connections on auth updates. Instead, we will rely on edgeHub to disconnect the client. Then the client will send disconnect signal to broker.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
dylanbronson: Restructured this part based on Vadim's suggestion

mqtt/mqtt-broker/src/broker.rs
dylanbronson: I've dealt with this for now by filtering out the offline case. Easiest way was to make a new method in Session. We can remove this when we do the refactor to store ths last used client info in offline session

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,52 @@ where
...
+                .subscriptions()
+                .into_iter()
+                .flat_map(std::collections::HashMap::values)
+            {
+                let client_info = session.client_info()?.clone();
+                let sub_topic_filter = sub.filter().clone().to_string();
+                let operation = Operation::new_subscribe(proto::SubscribeTo {
+                    topic_filter: sub_topic_filter.clone(),
+                    qos: *sub.max_qos(),
```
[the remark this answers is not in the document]
dylanbronson: Compiler complains if I just use sub.max_qos() as it's not expecting a reference
Clippy complains if I use clone
Copy doesn't work

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,64 @@ where
...
+            }
+        }
+        activities
+    }
+
+    fn reevaluate_subscriptions(&mut self) -> Result<(), Error> {
+        let disconnecting: Vec<ClientId> = self
+            .sessions
+            .iter()
+            .flat_map(|(client_id, session)| Self::prepare_activities(client_id, session))
```
vadim-kovalyov, line 134: nit: you could return tuple `(client_id, Self::prepare_activities(client_id, session))` to avoid cloning the activity. not super important, we can fix it later when we fix the offline session.

vadim-kovalyov (dismissed):
:rocket:

mqtt/mqtt-broker/src/broker.rs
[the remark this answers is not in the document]
dmolokanov: Oh, yes. We discussed this. We probably need to document this behavior somewhere

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,64 @@ where
...
+            .filter_map(
+                |activity: Activity| match self.authorizer.authorize(activity.clone()) {
```
dmolokanov, line 136: This .clone() is the reason you added `#derive(Clone)` in a bunch of places, right?

mqtt/mqtt-broker/src/session/mod.rs
```diff
@@ -74,6 +74,15 @@ impl Session {
         }
     }
 
+    pub fn client_info_omit_offline_sessions(&self) -> Option<&ClientInfo> {
```
dmolokanov, line 77: there is a method `Session::client_info` already. It didn't work for you in some cases?

dmolokanov: If it is Result and not Option, you can easily convert one to another: `session.client_info().ok()`

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -119,6 +122,52 @@ impl EdgeHubAuthorizer {
             .iter()
             .any(|allowed_topic| topic.starts_with(allowed_topic))
     }
+
+    fn get_on_behalf_of_id(topic: &str) -> Option<String> {
```
dmolokanov, line 126:
```suggestion
    fn get_on_behalf_of_id(topic: &str) -> Option<ClientId> {
```
?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -119,6 +122,52 @@ impl EdgeHubAuthorizer {
...
+
+        match (device_id, module_id) {
+            (Some(did), Some(mid)) => Some(format!("{}/{}", did, mid)),
+            (Some(did), None) => Some((*did).to_string()),
+            _ => None,
+        }
+    }
+
+    fn check_authorized_cache(&self, client_id: &ClientId, topic: &str) -> bool {
+        if let Some(on_behalf_of_id) = EdgeHubAuthorizer::get_on_behalf_of_id(topic) {
```
dmolokanov, line 146:
```rust
match EdgeHubAuthorizer::get_on_behalf_of_id(topic) {
  Some(on_behalf_of_id) if client_id == on_behalf_of_id => self.service_identities_cache.contains_key(client_id),
  Some(on_behalf_of_id) => {
    // if auth_chain for on_behalf_of_id contains client_id, then on_behalf_of_id is a descendant of client_id
    self.service_identities_cache.get(&on_behalf_of_id)
      .and_then(|service_identity| service_identity.auth_chain())
      .unwrap_or(false, auth_chain.contains(client_id))
  },
  None => {
    // If there is no on_behalf_of_id, we are dealing with a legacy topic
    // The client_id must still be in the identities cache
    self.service_identities_cache.contains_key(client_id)
  }
}
```

also, we could make `auth_chain: HashMap<ClientId>`

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -195,6 +267,15 @@ pub struct ServiceIdentity {
     auth_chain: Option<String>,
 }
 
+impl ServiceIdentity {
+    pub fn identity(&self) -> String {
```
dmolokanov, line 271: Can we avoid cloning in both methods?

dmolokanov (commented):
If we can avoid much cloning of activities in this PR. And remove `#[derive(Clone)]` as well

mqtt/mqtt-broker/src/session/mod.rs
dylanbronson: I added this in order to explicitly filter out offline sessions until we deal with them correctly

mqtt/mqtt-broker/src/broker.rs
dylanbronson: That's correct

mqtt/mqtt-broker/src/session/mod.rs
dylanbronson: Actually it's unnecessary. I removed it

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,62 @@ where
...
+                    client_id.clone(),
+                    Activity::new(client_id.clone(), client_info.clone(), operation),
+                ));
+            }
+        }
+        activities
+    }
+
+    fn reevaluate_subscriptions(&mut self) -> Result<(), Error> {
+        let disconnecting: Vec<ClientId> = self
```
dmolokanov, line 130: nit
```suggestion
        let disconnecting = self
```

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,62 @@ where
...
+                        );
+                        Some(client_id)
+                    }
+                    Err(e) => {
+                        warn!(message="error authorizing client subscription: {}", error = %e);
+                        Some(client_id)
+                    }
+                },
+            )
+            .collect();
```
dmolokanov, line 150: actually you don't need to call collect here. just iterate on the filter_map result

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -100,6 +104,62 @@ where
...
+                activities.push((
+                    client_id.clone(),
+                    Activity::new(client_id.clone(), client_info.clone(), operation),
+                ));
+            }
+        }
+        activities
+    }
+
+    fn reevaluate_subscriptions(&mut self) -> Result<(), Error> {
```
dmolokanov, line 129: I don't see you returning Err from this method
```suggestion
    fn reevaluate_subscriptions(&mut self) {
```

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -176,17 +234,24 @@ impl Authorizer for EdgeHubAuthorizer {
...
-        if let Some(service_identities) = identities.downcast_ref::<ServiceIdentity>() {
-            debug!("{:?}", service_identities);
-            // TODO: fill in update method
+        if let Some(service_identities) = identities.downcast_ref::<Vec<ServiceIdentity>>() {
+            debug!(
+                "service identities update received. Service identities: {:?}",
+                service_identities
+            );
+
+            self.service_identities_cache = service_identities
```
dmolokanov, line 245: I believe we can consume update here and save some clones
```suggestion
            self.service_identities_cache = service_identities
                 .into_iter()
                .map(|id| (d.identity().into(), id))
                .collect();
```

dmolokanov (commented):
Some more suggestions

dmolokanov (approved):
Great work! :shipit:

PR #3552, opened 2020-09-15
Refactor AMQP auth to use client auth-chain

richma-ms:
This change refactors the AMQP OnBehalfOf auth to fully use the client auth-chain.

Previously we made OnBehalfOf auth work by getting the actor from the token and the auth target from the audience property of the UpdateToken message on the CBS link.  But it turns out this isn't compatible with older SDKs, where the audience property wasn't set correctly.

So now we get both pieces of information from the client provided auth-chain, which gives us an explicit way to distinguish between "normal" auth and OnBehalfOf auth.  This conveniently removes any dependency we had on the token payload, and will help enable OnBehalfOf cert-based auth in the future.

MQTT path is unmodified, because it does not have a concept of OnBehalfOf AuthN, and AuthZ is performed by the broker.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
@@ -80,37 +80,49 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Amqp
             try
             {
                 // See if we received direct credentials for the target identity
-                ClientToken credentialsInfo;
-                bool hasCredentials = this.clientTokens.TryGetValue(id, out credentialsInfo);
+                bool hasToken = this.clientTokens.TryGetValue(id, out ClientToken clientToken);
 
                 // Otherwise, this could be a child Edge acting OnBehalfOf of the target,
                 // in which case we would have the actor's credentials instead
-                if (!hasCredentials)
+                if (!hasToken)
                 {
                     Option<string> actorDeviceId = AuthChainHelpers.GetActorDeviceId(authChain);
-                    hasCredentials = actorDeviceId.Map(actor => this.clientTokens.TryGetValue(actor + $"/{CoreConstants.EdgeHubModuleId}", out credentialsInfo)).GetOrElse(false);
+                    hasToken = actorDeviceId.Map(actor => this.clientTokens.TryGetValue($"{actor}/{CoreConstants.EdgeHubModuleId}", out clientToken)).GetOrElse(false);
                 }
 
-                if (hasCredentials)
+                if (hasToken)
                 {
-                    if (!authenticatedClients.ContainsKey(id))
+                    // Lock the auth-state check and the auth call to avoid
+                    // redundant calls into the authenticator
+                    using (await this.identitySyncLock.LockAsync())
                     {
-                        IClientCredentials clientCredentials = this.clientCredentialsFactory.GetWithSasToken(
-                            credentialsInfo.DeviceId,
-                            credentialsInfo.ModuleId.OrDefault(),
-                            string.Empty,
-                            credentialsInfo.Token,
-                            true,
-                            modelId,
-                            authChain);
-
-                        if (await this.authenticator.AuthenticateAsync(clientCredentials))
+                        if (this.authenticatedClients.ContainsKey(id))
                         {
-                            // Authentication success, add an entry for the authenticated
-                            // client identity, the value in the dictionary doesn't matter
-                            // as we're effectively using it as a thread-safe HashSet
-                            authenticatedClients[id] = new byte();
+                            // We've previously authenticated this client already
                             return true;
+                        }
+                        else
+                        {
+                            IClientCredentials clientCredentials = this.clientCredentialsFactory.GetWithSasToken(
+                                clientToken.DeviceId,
+                                clientToken.ModuleId.OrDefault(),
+                                string.Empty,
+                                clientToken.Token,
+                                true,
+                                modelId,
+                                authChain);
+
+                            if (await this.authenticator.AuthenticateAsync(clientCredentials))
+                            {
+                                // Authentication success, add an entry for the authenticated
+                                // client identity, the value in the dictionary doesn't matter
+                                // as we're effectively using it as a thread-safe HashSet
+                                this.authenticatedClients[id] = default(byte);
+
+                                // TODO: After auth success, we should be adding the client
+                                // token to the credentials cache
+                                return true;
+                            }
                         }
                     }
                 }
@@ -196,7 +208,6 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Amqp
             throw new InvalidOperationException($"Matching template not found for audience {audienceUri}");
         }
 
-        // Note: This method updates this.clientCredentialsMap, and should be invoked only within this.identitySyncLock
         internal async Task<(AmqpResponseStatusCode, string)> UpdateCbsToken(AmqpMessage message)
         {
             ClientToken clientToken;
@@ -214,26 +225,26 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Amqp
             // Insert/update the cached tokens with the new value
             this.clientTokens[clientToken.Id] = clientToken;
 
-            if (this.clientTokens.TryGetValue(clientToken.Id, out ClientToken credentialsInfo))
+            // TODO: This check is incorrect for OnBehalfOf auth, as the
+            // token might be from the actor EdgeHub, but a leaf identity
+            // gets authenticated instead
+            if (this.authenticatedClients.ContainsKey(clientToken.Id))
             {
-                if (this.authenticatedClients.ContainsKey(clientToken.Id))
-                {
-                    IClientCredentials clientCredentials = this.clientCredentialsFactory.GetWithSasToken(
-                        credentialsInfo.DeviceId,
-                        credentialsInfo.ModuleId.OrDefault(),
-                        string.Empty,
-                        credentialsInfo.Token,
-                        true,
-                        Option.None<string>(),
-                        Option.None<string>());
+                IClientCredentials clientCredentials = this.clientCredentialsFactory.GetWithSasToken(
+                    clientToken.DeviceId,
+                    clientToken.ModuleId.OrDefault(),
+                    string.Empty,
+                    clientToken.Token,
+                    true,
+                    Option.None<string>(),
+                    Option.None<string>());
 
-                    await this.credentialsCache.Add(clientCredentials);
-                    Events.CbsTokenUpdated(credentialsInfo.Id);
-                }
-                else
-                {
-                    Events.CbsTokenNotUpdated(credentialsInfo.Id);
-                }
+                await this.credentialsCache.Add(clientCredentials);
+                Events.CbsTokenUpdated(clientToken.Id);
+            }
+            else
+            {
+                Events.CbsTokenNotUpdated(clientToken.Id);
             }
 
             return (AmqpResponseStatusCode.OK, AmqpResponseStatusCode.OK.ToString());
@@ -277,18 +288,15 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Amqp
 
         async Task HandleTokenUpdate(AmqpMessage message)
         {
-            using (await this.identitySyncLock.LockAsync())
+            try
             {
-                try
-                {
-                    (AmqpResponseStatusCode statusCode, string description) = await this.UpdateCbsToken(message);
-                    await this.SendResponseAsync(message, statusCode, description);
-                }
-                catch (Exception e)
-                {
-                    await this.SendResponseAsync(message, AmqpResponseStatusCode.InternalServerError, e.Message);
-                    Events.ErrorUpdatingToken(e);
-                }
+                (AmqpResponseStatusCode statusCode, string description) = await this.UpdateCbsToken(message);
+                await this.SendResponseAsync(message, statusCode, description);
+            }
+            catch (Exception e)
+            {
+                await this.SendResponseAsync(message, AmqpResponseStatusCode.InternalServerError, e.Message);
+                Events.ErrorUpdatingToken(e);
             }
         }
 
@@ -312,7 +320,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Amqp
             {
                 this.DeviceId = deviceId;
                 this.ModuleId = moduleId;
-                this.Token = new AtomicReference<string>(token);
+                this.Token = token;
             }
 
             public string Id => this.ModuleId.Map(m => $"{this.DeviceId}/{m}").GetOrElse(this.DeviceId);
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
@@ -17,6 +17,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Amqp
     using Microsoft.Azure.Devices.Edge.Util;
     using Microsoft.Azure.Devices.Edge.Util.Concurrency;
     using Microsoft.Extensions.Logging;
+    using CoreConstants = Microsoft.Azure.Devices.Edge.Hub.Core.Constants;
 
     /// <summary>
     /// This class is used to get tokens from the Client on the CBS link. It generates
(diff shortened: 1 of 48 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -73,24 +74,42 @@ public void RegisterLink(IAmqpLink link)
             Events.LinkRegistered(link);
         }
 
-        public async Task<bool> AuthenticateAsync(string id, Option<string> modelId)
+        public async Task<bool> AuthenticateAsync(string id, Option<string> modelId, Option<string> authChain)
```
dylanbronson, line 77: Probably fine for now, but should we consider passing in some sort object that contains modelId, authChain, and whatever else we may add in the future?

richma-ms: Hmm....model ID and authchain aren't logically related in any way though.  I guess maybe if the param-list gets really long in the future?  For now there's only 3 so I think it's okay.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -73,24 +74,42 @@ public void RegisterLink(IAmqpLink link)
...
+                // See if we received direct credentials for the target identity
+                CredentialsInfo credentialsInfo;
+                bool hasCredentials = this.clientCredentialsMap.TryGetValue(id, out credentialsInfo);
+
+                // Otherwise, this could be a child Edge acting OnBehalfOf of the target,
+                // in which case we would have the actor's credentials instead
+                if (!hasCredentials)
+                {
+                    Option<string> actorDeviceId = AuthChainHelpers.GetActorDeviceId(authChain);
+                    hasCredentials = actorDeviceId.Map(actor => this.clientCredentialsMap.TryGetValue(actor + $"/{CoreConstants.EdgeHubModuleId}", out credentialsInfo)).GetOrElse(false);
```
dylanbronson, line 90: So we will store "on behalf of" clientCredentials as "device_z/$edgeHub" if the child edge is acting on behalf of device_z?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -73,24 +74,42 @@ public void RegisterLink(IAmqpLink link)
...
+
+                if (hasCredentials)
                 {
                     // Not -ve caching isAuthenticated here.
                     // If incorrect credentials are sent, then it authenticates every time and fails
                     // If correct credentials are sent later, then the authentication will succeed.
                     if (!credentialsInfo.IsAuthenticated)
                     {
-                        using (await credentialsInfo.AsyncLock.LockAsync())
+                        if (!credentialsInfo.IsAuthenticated)
```
dylanbronson, line 100: unnecessary?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/linkhandlers/LinkHandler.cs
```diff
@@ -41,10 +41,12 @@ public abstract class LinkHandler : ILinkHandler
...
 
             string modelId = null;
-            if (this.Link.Settings?.Properties?.TryGetValue(IotHubAmqpProperty.ModelId, out modelId) ?? false)
-            {
-                this.ModelId = Option.Maybe(modelId);
-            }
+            this.Link.Settings?.Properties?.TryGetValue(IotHubAmqpProperty.ModelId, out modelId);
+            this.ModelId = Option.Maybe(modelId);
+
+            string authChain = null;
```
dylanbronson, line 47: nit: can you combine this line and the next?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
```diff
@@ -12,11 +13,46 @@ public static string[] GetAuthChainIds(string authChain)
...
+            if (!authChain.HasValue)
+            {
+                return Option.None<string>();
+            }
+
+            string authChainString = Preconditions.CheckNonWhiteSpace(authChain.OrDefault(), nameof(authChain));
+            string[] authChainIds = GetAuthChainIds(authChainString);
+
+            // The auth target is always the first element of the auth-chain
+            return authChainIds.FirstOption(id => true);
```
dylanbronson, line 27: auth chain goes like

edgeA;edgeB;device_z 
right? So the auth target here is edgeA?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Amqp.Test/CbsNodeTest.cs
```diff
@@ -181,7 +178,42 @@ public async Task UpdateCbsTokenTest()
...
+            {
+                Value = TokenHelper.CreateSasToken("edgehubtest1.azure-devices.net/devices/edge1/modules/$edgeHub")
+            };
+            AmqpMessage validAmqpMessage = AmqpMessage.Create(amqpValue);
+            validAmqpMessage.ApplicationProperties.Map[CbsConstants.PutToken.Type] = "azure-devices.net:sastoken";
+            validAmqpMessage.ApplicationProperties.Map[CbsConstants.PutToken.Audience] = "edgehubtest1.azure-devices.net/devices/edge1/modules/$edgeHub";
+            validAmqpMessage.ApplicationProperties.Map[CbsConstants.Operation] = CbsConstants.PutToken.OperationValue;
+            Option<string> authChain = Option.Some("device1;edge1");
+
+            var actorEdgeHubIdentity = Mock.Of<IIdentity>(i => i.Id == "edge1/$edegHub");
```
dylanbronson, line 213: nit: spelling

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
richma-ms: yep

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: No, it'd be stored as "edge1/$edgeHub", where edge1 is the parent of device_z.  This basically reverts the audience logic to what we had before, where it gets pulled from the client token.  And since the token is for the actor edgeHub, that's how it gets stored.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms, line 100: >if (!credentialsInfo.IsAuthenticated) [](start = 24, length = 37)

this is actually problematic, we're checking the authenticated status of the actor instead of the auth target.  So if the actor opens multiple links for different devices, only the first device will go through auth, and all others will end up skipping.

richma-ms: yea good catch, my eyes just glossed over this like multiple times

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: actually this leads to a bug when the authentication status is cached, I mentioned it in another comment and have pushed out a new iteration with the fix.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/linkhandlers/LinkHandler.cs
richma-ms: aren't they already combined?  or are you talking about combining both model ID and auth-chain sections?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -73,33 +75,47 @@ public void RegisterLink(IAmqpLink link)
...
 
-        public async Task<bool> AuthenticateAsync(string id, Option<string> modelId)
+        public async Task<bool> AuthenticateAsync(string id, Option<string> modelId, Option<string> authChain)
         {
             try
             {
-                if (this.clientCredentialsMap.TryGetValue(id, out CredentialsInfo credentialsInfo))
+                // See if we received direct credentials for the target identity
+                ClientToken credentialsInfo;
+                bool hasCredentials = this.clientTokens.TryGetValue(id, out credentialsInfo);
```
varunpuranik, line 84:
```suggestion
                bool hasCredentials = this.clientTokens.TryGetValue(id, out ClientToken credentialsInfo);
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -73,33 +75,47 @@ public void RegisterLink(IAmqpLink link)
...
+                // Otherwise, this could be a child Edge acting OnBehalfOf of the target,
+                // in which case we would have the actor's credentials instead
+                if (!hasCredentials)
                 {
-                    // Not -ve caching isAuthenticated here.
-                    // If incorrect credentials are sent, then it authenticates every time and fails
-                    // If correct credentials are sent later, then the authentication will succeed.
-                    if (!credentialsInfo.IsAuthenticated)
+                    Option<string> actorDeviceId = AuthChainHelpers.GetActorDeviceId(authChain);
+                    hasCredentials = actorDeviceId.Map(actor => this.clientTokens.TryGetValue(actor + $"/{CoreConstants.EdgeHubModuleId}", out credentialsInfo)).GetOrElse(false);
```
varunpuranik, line 91:
```suggestion
                    hasCredentials = actorDeviceId.Map(actor => this.clientTokens.TryGetValue($"{actor}/{CoreConstants.EdgeHubModuleId}", out credentialsInfo)).GetOrElse(false);
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -73,33 +75,47 @@ public void RegisterLink(IAmqpLink link)
...
+                }
+
+                if (hasCredentials)
+                {
+                    if (!authenticatedClients.ContainsKey(id))
```
varunpuranik, line 96: We should return true if authenticatedClients has the id, right?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
varunpuranik, line 84: Nit: suggest calling this "hasToken"

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -73,33 +75,47 @@ public void RegisterLink(IAmqpLink link)
...
         }
 
-        public async Task<bool> AuthenticateAsync(string id, Option<string> modelId)
+        public async Task<bool> AuthenticateAsync(string id, Option<string> modelId, Option<string> authChain)
         {
             try
             {
-                if (this.clientCredentialsMap.TryGetValue(id, out CredentialsInfo credentialsInfo))
+                // See if we received direct credentials for the target identity
+                ClientToken credentialsInfo;
```
varunpuranik, line 83: Nit: clientToken?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
varunpuranik: We should also pass in the productInfo in here. 
Check with @ancaantochi on how to get that?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -73,33 +75,47 @@ public void RegisterLink(IAmqpLink link)
...
+                            authChain);
+
+                        if (await this.authenticator.AuthenticateAsync(clientCredentials))
                         {
-                            if (!credentialsInfo.IsAuthenticated)
-                            {
-                                credentialsInfo.ClientCredentials.ModelId = modelId;
-                                credentialsInfo.IsAuthenticated = await this.authenticator.AuthenticateAsync(credentialsInfo.ClientCredentials);
-                            }
+                            // Authentication success, add an entry for the authenticated
```
varunpuranik, line 109: How do we catch the token expiring?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
varunpuranik, line 96: Do we need to lock here? I am not sure if the ConcurrentDictionary suffices..

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -185,48 +199,51 @@ internal static (string deviceId, string moduleId) ParseIds(string audience)
...
 
-            if (!this.clientCredentialsMap.TryGetValue(clientCredentials.Identity.Id, out CredentialsInfo credentialsInfo))
-            {
-                this.clientCredentialsMap[clientCredentials.Identity.Id] = new CredentialsInfo(clientCredentials);
-            }
-            else
+            // Insert/update the cached tokens with the new value
+            this.clientTokens[clientToken.Id] = clientToken;
+
+            if (this.clientTokens.TryGetValue(clientToken.Id, out ClientToken credentialsInfo))
```
varunpuranik, line 217: ? You are adding the token and then immediately checking if the token exists?
Here we should just always add the token.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -289,23 +306,22 @@ Task SendResponseAsync(AmqpMessage requestMessage, AmqpResponseStatusCode status
             }
         }
 
-        /// <summary>
-        /// This type is deliberately mutable because of the use case.
-        /// </summary>
-        class CredentialsInfo
+        internal class ClientToken
```
varunpuranik, line 309: Internal for tests?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -289,23 +306,22 @@ Task SendResponseAsync(AmqpMessage requestMessage, AmqpResponseStatusCode status
...
         {
-            public CredentialsInfo(IClientCredentials clientCredentials)
+            public ClientToken(string deviceId, Option<string> moduleId, string token)
             {
-                this.ClientCredentials = clientCredentials;
-                this.IsAuthenticated = false;
-                this.AsyncLock = new AsyncLock();
+                this.DeviceId = deviceId;
+                this.ModuleId = moduleId;
+                this.Token = new AtomicReference<string>(token);
```
varunpuranik, line 315: Why is this an AtomicRef?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: Read/write on a concurrent collection itself is threadsafe, right?  Since we're never modifying the stored value, the keys should be safe to read from multiple threads.

---
In reply to: [489200704](https://github.com/Azure/iotedge/pull/3552#discussion_r489200704) [](ancestors = 489200704)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: token expiration wasn't handled by CbsNode before, iirc the re-authenticator should catch it and have the ConnectionManager terminate the client connection if the token is expired, which should ultimately cause the CbsLink to be dropped.

---
In reply to: [489200223](https://github.com/Azure/iotedge/pull/3552#discussion_r489200223) [](ancestors = 489200223)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: Forgot to remove the extra check when I split out the authentication state and the token.  The intent is to call into the credcache if we've already authenticated this token.

---
In reply to: [489204250](https://github.com/Azure/iotedge/pull/3552#discussion_r489204250) [](ancestors = 489204250)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: yea used by one of the unit tests

---
In reply to: [489204579](https://github.com/Azure/iotedge/pull/3552#discussion_r489204579) [](ancestors = 489204579)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: forgot to clean this up as well when I split up the struct =(

---
In reply to: [489205083](https://github.com/Azure/iotedge/pull/3552#discussion_r489205083) [](ancestors = 489205083)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: Talked offline, the concern is around re-entrancy of the method itself, added a lock.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: Talked offline, filed another PBI to come back and fix the cred-cache usage.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
```diff
@@ -37,7 +38,8 @@ class CbsNode : ICbsNode, IAmqpAuthenticator
         readonly IAuthenticator authenticator;
         readonly string iotHubHostName;
         readonly ICredentialsCache credentialsCache;
-        readonly ConcurrentDictionary<string, CredentialsInfo> clientCredentialsMap = new ConcurrentDictionary<string, CredentialsInfo>();
+        readonly ConcurrentDictionary<string, ClientToken> clientTokens = new ConcurrentDictionary<string, ClientToken>();
+        readonly ConcurrentDictionary<string, byte> authenticatedClients = new ConcurrentDictionary<string, byte>();
```
varunpuranik, line 42: Does this still need to be ConcurrentDictionary?

varunpuranik (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/CbsNode.cs
richma-ms: yea still needed for UpdateCbsToken which happens on a 3rd thread.

PR #3554, opened 2020-09-15
Remove `version` from GetHash for EdgeAgentDockerModule

yophilav:
Remove `version` from GetHash for EdgeAgentDockerModule to make the function GetHash() consistent with Equal(). They both are not considering `version` as a part of the redeployment condition.


yophilav:
I have tested the offline scenario with 
1. Set the version on both edgeAgent and edgeHub
2. Deploy
3. Disconnect the network
4. reconnect the network
5. check if either of the module is restarted.

philipktlin (approved):

PR #3559, opened 2020-09-17
Enable connectivity test for nested edge scenario

philipktlin:
Use parent hostname in connectivity test if provided

### DIFF
--- a/smoke/IotEdgeQuickstart/Program.cs
+++ b/smoke/IotEdgeQuickstart/Program.cs
@@ -59,6 +59,7 @@ Defaults:
   --runtime-log-level                   debug
   --clean_up_existing_device            false
   --proxy                               no proxy is used
+  --parent-hostname                     no parent hostname is used
 ")]
     [HelpOption]
     class Program
--- a/smoke/IotEdgeQuickstart/Program.cs
+++ b/smoke/IotEdgeQuickstart/Program.cs
@@ -174,6 +174,9 @@ Defaults:
         [Option("--device_identity_pk", Description = "Optional path to the device identity private key file. Used for either DPS or manual provisioning flows")]
         public string DeviceIdentityPk { get; } = string.Empty;
 
+        [Option("--parent-hostname", Description = "Optional input to specify parent hostname for nested edge scenario")]
+        public string ParentHostname { get; } = string.Empty;
+
         // ReSharper disable once UnusedMember.Local
         static int Main(string[] args) => CommandLineApplication.ExecuteAsync<Program>(args).Result;
 
@@ -287,6 +290,7 @@ Defaults:
                     tag,
                     this.DeviceId,
                     this.EdgeHostname,
+                    string.IsNullOrWhiteSpace(this.ParentHostname) ? Option.None<string>() : Option.Some(this.ParentHostname),
                     this.LeaveRunning,
                     this.NoVerify,
                     this.BypassEdgeInstallation,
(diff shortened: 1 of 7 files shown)


test/connectivity/scripts/connectivityTest.sh
```diff
@@ -343,6 +343,10 @@ function run_connectivity_test() {
     print_highlighted_message "Run connectivity test with -d '$device_id' started at $test_start_time"
 
     SECONDS=0
+    PARENT_HOSTNAME=$(printEnv E2E_PARENT_HOSTNAME)
```
yophilav, line 346: Is `printEnv` a valid bash command?

smoke/IotEdgeQuickstart/Program.cs
```diff
@@ -174,6 +174,9 @@ class Program
         [Option("--device_identity_pk", Description = "Optional path to the device identity private key file. Used for either DPS or manual provisioning flows")]
         public string DeviceIdentityPk { get; } = string.Empty;
 
+        [Option("--parent-hostname", Description = "Optional input to specify parent hostname for nested edge scenario")]
```
yophilav, line 177: Please add the default behavior description to line62.

smoke/IotEdgeQuickstart/Program.cs
```diff
@@ -287,6 +290,7 @@ async Task<int> OnExecuteAsync()
                     tag,
                     this.DeviceId,
                     this.EdgeHostname,
+                    string.IsNullOrWhiteSpace(this.ParentHostname) ? Option.None<string>() : Option.Some(this.ParentHostname),
```
yophilav, line 293: Do we need this `IsNullOrWhiteSpace()` here? Since on line178, we already have the default value set to empty. Do we need the type to be `Option<string>` instead of just plain `string`?

smoke/IotEdgeQuickstart/details/Details.cs
```diff
@@ -273,7 +277,7 @@ protected Task ConfigureBootstrapper()
                 agentImage = Option.Some<string>(this.EdgeAgentImage());
             }
 
-            return this.bootstrapper.Configure(method, agentImage, this.hostname, this.deviceCaCert, this.deviceCaPk, this.deviceCaCerts, this.runtimeLogLevel);
+            return this.bootstrapper.Configure(method, agentImage, this.hostname, this.parentHostname, this.deviceCaCert, this.deviceCaPk, this.deviceCaCerts, this.runtimeLogLevel);
```
yophilav, line 280: Did you forget to update `Configure(8)` in `iotedge\smoke\IotEdgeQuickstart\details\IotedgedWindows.cs`? It takes the argument but it doesn't do the `parentHostname.ForEach(v => doc.ReplaceOrAdd("parent_hostname", v));`

smoke/IotEdgeQuickstart/Program.cs
philipktlin: To use IsNullOrWhiteSpace(), it is defensive to protect if white spaces are passing in.  For Option<string>, all optional string argument should use it to show intent explicitly if the argument is not provided.

test/connectivity/scripts/connectivityTest.sh
philipktlin: I think so, but wrong case.  should be "printenv".

smoke/IotEdgeQuickstart/details/Details.cs
philipktlin: I added the comment in IotedgedWindows.cs to say that Windows is no longer supported in 1.1.0 or later.  Therefore I didn't make any change to use it.

yophilav (approved):

PR #3561, opened 2020-09-17
Put limits on rqt data upload

lfitchett:
This will not upload rqt data older than 1 week (default value, configurable with MetricMaxUploadAge).


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/storage/MetricsStorage.cs
```diff
@@ -27,23 +31,54 @@ public Task StoreMetricsAsync(IEnumerable<Metric> metrics)
...
+        async IAsyncEnumerable<Metric> GetAllMetricsInternal([EnumeratorCancellation] CancellationToken cancellationToken = default)
+        {
+            HashSet<long> maxAgeEntriesToRemove = new HashSet<long>();
+
+            foreach ((long offset, IEnumerable<Metric> metrics) in await this.dataStore.GetBatch(0, 1000, cancellationToken))
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                // This makes the assumption that metrics are stored in batches with matching timestamps.
+                // This is the case b/c the metrics are stored immediately after being scraped.
+                if (metrics.First().TimeGeneratedUtc < DateTime.UtcNow - this.maxMetricAge)
```
damonbarry, line 47: `UtcNow` will be reevaluated each time through the loop, so different batches will be evaluated against different max-age timestamps. Is that the right behavior or do we want to take a snapshot of `UtcNow` before we enter the loop?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/storage/MetricsStorage.cs
```diff
@@ -27,23 +31,54 @@ public Task StoreMetricsAsync(IEnumerable<Metric> metrics)
...
+                {
+                    // TODO: log removed
```
damonbarry, line 49: Can we add the logging in this PR? I imagine we'd count up the total # of removed entries and report it in a single log statement (at least # removed entries, although time range would be nice to have)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/storage/MetricsStorage.cs
```diff
@@ -27,23 +31,54 @@ public Task StoreMetricsAsync(IEnumerable<Metric> metrics)
...
+                }
+            }
+
+            cancellationToken.ThrowIfCancellationRequested();
+
+            // remove metrics past ttl
+            await this.RemoveOffsets(maxAgeEntriesToRemove);
+        }
+
+        public Task RemoveAllReturnedMetricsAsync()
```
damonbarry, line 68: Remove the `Async` suffix?

lfitchett: Its still a task, I thought we named it async if it returns a task. Thats all the caller cares about anyway.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/storage/MetricsStorage.cs
lfitchett: It should happen so fast it won't matter (we're comparing on a week timescale), but I'll go ahead and change it for consistency

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/storage/MetricsStorage.cs
lfitchett: added logging. I didn't bother logging at the end. The only time multiple scrapes will be deleted at once is if the device was powered off for an extended period of time.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/storage/MetricsStorage.cs
damonbarry: You're right. Not sure what I was thinking...

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/storage/MetricsStorage.cs
damonbarry: Looks good.

damonbarry (approved):

damonbarry (approved):

PR #3566, opened 2020-09-17
Nested Edge E2E

yophilav:
- Enabling the E2E test to be running on L3 of a nested edge setup. 
- Only the nested edge E2E test will be run if and only if the `agent.name` is specified in the pipeline. Otherwise, the E2E test is running the standard 3 environment setups.


builds/e2e/e2e.yaml
```diff
@@ -33,6 +33,9 @@ jobs:
       name: $(pool.name)
       demands: rpi3-e2e-tests
 
+    # If Agent name is specified, run only nested edge test
+    condition: eq(variables['agent.name'], '')
```
damonbarry, line 37: We don't need the conditions in `e2e.yaml` right? That was kind of the point of moving the nested edge job to its own pipeline...

PR #3567, opened 2020-09-18
Blob upload

huguesBouvier:
Adding blob upload for nested edge.

Tested with 3 level of nesting:
Upload blob at level 2 and level 1 (deepest).
Download container at level 3, 2 and 1.
Upload support bundle at level 1.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/INestedEdgeParentUrlParser.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/INestedEdgeParentUrlParser.cs
@@ -1,0 +1,9 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Agent.Core
+{
+    using Microsoft.Azure.Devices.Edge.Util;
+    public interface INestedEdgeParentUriParser
+    {
+        public Option<string> ParseURI(string uri);
+    }
+}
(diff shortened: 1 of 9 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/INestedEdgeParentUrlParser.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Agent.Core
+{
+    using Microsoft.Azure.Devices.Edge.Util;
+    public interface INestedEdgeParentUriParser
+    {
+        public Option<string> ParseURI(string uri);
```
richma-ms, line 7: >ParseURI [](start = 30, length = 8)

hmm maybe something like TranslateUpstreamUri?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Core.Test/NestedEdgeParentUrlParserTest.cs
```diff
@@ -0,0 +1,69 @@
...
+            // Check for entry not existing and add to dictionary
+            this.Map[variableName] = value;
+        }
+    }
+
+    [Unit]
+    public class NestedEdgeParentUriParserTest
+    {
+        [Theory]
+        [InlineData("$upstream:9000/comp/ea:tag1", "parentaddress:9000/comp/ea:tag1", null, "IOTEDGE_GATEWAYHOSTNAME", "parentaddress")]
```
richma-ms, line 37: >9000 [](start = 31, length = 4)

prob out of scope for this change, but I don't think we're tracking the change to enforce that port is omitted in parent_hostname, and that should probably be bundled with this feature area

richma-ms (approved):

huguesBouvier:
Thanks Richard. I am merging right now then. I am making another PR with your comments fixed.

PR #3568, opened 2020-09-18
Add edgeHub moduleId to RoutingEdgeHub to fix message source bug

dylanbronson:
Reported Property Updates (RPU) as telemetry messages have the wrong deviceId as their source.
This bug manifested itself when we set route like "FROM /* TO $upstream" and also set a route in iothub for twin change notifications. EdgeHub sends an RPU on behalf of a downstream device, but also sends the RPU as telemetry. This is by design, but the RPU as telemetry should be from $edgeHub, not from the device. 

This PR adds edge device's id as the DeviceConnectionId and the edgeHub module ID as the ModuleConnectionId to the telemetry message before sending it out. I tested using Azure IoT explorer, and the telemetry messages do show up from edge device, not from leaf device.

### DIFF
--- a/edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/routing/RoutingTest.cs
+++ b/edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/routing/RoutingTest.cs
@@ -370,12 +370,10 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Test.Routing
             };
 
             string edgeDeviceId = "edge";
-            ;
             var iotHub = new IoTHub();
             (IEdgeHub edgeHub, IConnectionManager connectionManager) = await SetupEdgeHub(routes, iotHub, edgeDeviceId);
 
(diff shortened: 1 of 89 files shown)


varunpuranik (commented):
Changes look good, can you add a test case for it?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/routing/RoutingTest.cs
```diff
@@ -504,6 +526,11 @@ class IoTHub
             public bool HasReceivedTwinChangeNotification() => this.ReceivedMessages.Any(
                 m =>
                     m.SystemProperties[SystemProperties.MessageType] == Constants.TwinChangeNotificationMessageType);
+
+            public bool HasReceivedMessageFromModule(string deviceId, string moduleId) => this.ReceivedMessages.Any(
```
varunpuranik, line 530: Instead of this, I would say update the HasReceivedTwinChangeNotification to check if it came from edgeHub.

varunpuranik (approved):

varunpuranik (approved):

PR #3569, opened 2020-09-18
EdgeAgent Direct Method Tests

lfitchett:
This adds tests for several of the edgeAgent direct methods.


test/Microsoft.Azure.Devices.Edge.Test/helpers/Context.cs
```diff
@@ -162,6 +163,8 @@ IEnumerable<Registry> GetAndValidateRegistries()
 
         public Option<string> TempSensorImage { get; }
 
+        public Option<string> NumberLoggerImage { get; }
```
damonbarry, line 166: This new parameter in context.json needs to be documented in `test/README.md`

test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
```diff
@@ -0,0 +1,150 @@
...
+            CancellationToken token = this.TestToken;
+
+            string numberLoggerImage = Context.Current.NumberLoggerImage.Expect(() => new InvalidOperationException("Missing Number Logger image"));
+            await this.runtime.DeployConfigurationAsync(
+                builder =>
+                {
+                    builder.AddModule(moduleName, numberLoggerImage)
+                        .WithEnvironment(new[] { ("Count", count.ToString()) });
+                }, token);
+            await Task.Delay(30000);
```
damonbarry, line 51: This might make things easier for you now, but it will cause headaches down the road. Are you sure you can't do a little more digging to see why this is necessary?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/requests/ModuleLogsRequestHandler.cs
```diff
@@ -50,23 +52,29 @@ protected override async Task<Option<IEnumerable<ModuleLogsResponse>>> HandleReq
...
                     Events.ReceivedModuleLogs(moduleLogs, l.id);
-                    return logOptions.ContentEncoding == LogsContentEncoding.Gzip
-                        ? new ModuleLogsResponse(l.id, moduleLogs)
-                        : new ModuleLogsResponse(l.id, moduleLogs.FromBytes());
+
+                    if (l.logOptions.ContentEncoding == LogsContentEncoding.Gzip)
+                    {
+                        if (moduleLogs.Length > MaxPayloadSize)
+                        {
+                            return new ModuleLogsResponse(l.id, Events.LargePayload(moduleLogs.Length, l.logOptions));
```
damonbarry, line 63: This non-trivial change to the behavior of the product code should probably be its own PR.

test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            TaskStatusResponse response = null;
+            for (int i = 0; i < 10; i++)
+            {
+                Assert.AreEqual((int)HttpStatusCode.OK, result.Status);
+                response = JsonConvert.DeserializeObject<TaskStatusResponse>(result.GetPayloadAsJson());
+
+                if (response.Status == BackgroundTaskRunStatus.NotStarted || response.Status == BackgroundTaskRunStatus.Running)
+                {
+                    break;
```
damonbarry, line 130: Why do you want to break out of the loop if the upload hasn't completed yet?

test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
```diff
@@ -0,0 +1,150 @@
...
+                        .WithEnvironment(new[] { ("Count", count.ToString()) });
+                }, token);
+            await Task.Delay(10000);
+
+            var request = new ModuleLogsUploadRequest("1.0", new List<LogRequestItem> { new LogRequestItem(moduleName, new ModuleLogFilter(Option.None<int>(), Option.None<string>(), Option.None<string>(), Option.None<int>(), Option.None<string>())) }, LogsContentEncoding.None, LogsContentType.Text, sasUrl);
+
+            CloudToDeviceMethodResult result = await this.iotHub.InvokeMethodAsync(this.runtime.DeviceId, ConfigModuleName.EdgeAgent, new CloudToDeviceMethod("UploadModuleLogs", TimeSpan.FromSeconds(300), TimeSpan.FromSeconds(300)).SetPayloadJson(JsonConvert.SerializeObject(request)), token);
+
+            TaskStatusResponse response = null;
+            for (int i = 0; i < 10; i++)
```
damonbarry, line 123: Is 10 a magic number? Why not loop until the test times out?

test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
```diff
@@ -0,0 +1,150 @@
...
+
+                await Task.Delay(5000);
+                var correlation = new TaskStatusRequest("1.0", response.CorrelationId);
+                result = await this.iotHub.InvokeMethodAsync(this.runtime.DeviceId, ConfigModuleName.EdgeAgent, new CloudToDeviceMethod("GetTaskStatus", TimeSpan.FromSeconds(300), TimeSpan.FromSeconds(300)).SetPayloadJson(JsonConvert.SerializeObject(correlation)), token);
+            }
+
+            Assert.AreEqual(BackgroundTaskRunStatus.Completed, response.Status, response.Message);
+
+            // BlobServiceClient blobServiceClient = new BlobServiceClient(sasUrl);
+            // string expected = string.Join('\n', Enumerable.Range(0, count));
```
damonbarry, line 141: Remove?

test/modules/NumberLogger/src/Program.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            string countString = Environment.GetEnvironmentVariable("Count");
+            int parsedCount = int.Parse(countString);
+
+            for (int i = 0; i < parsedCount; i++)
+            {
+                Console.WriteLine(i);
+            }
+
+            Task.Delay(-1).Wait();
```
damonbarry, line 19: This container won't respond when docker asks it to close, so docker will have to forcibly close it after some number of seconds. I think it would be better to use the helpers in Edge.Util and Edge.ModuleUtil

test/modules/NumberLogger/config/appsettings.json
```diff
@@ -0,0 +1,3 @@
+﻿{
+  "EdgeHubConnectionString": "<YourConnectionString>"
+}
```
damonbarry, line 3: This file isn't needed

test/modules/NumberLogger/NumberLogger.csproj
```diff
@@ -0,0 +1,44 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\edge-util\src\Microsoft.Azure.Devices.Edge.Util\Microsoft.Azure.Devices.Edge.Util.csproj" />
+    <ProjectReference Include="..\ModuleLib\Microsoft.Azure.Devices.Edge.ModuleUtil.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="config/appsettings.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
damonbarry, line 35: I don't think any of these package/project references are needed. Also appsettings.json...

test/Microsoft.Azure.Devices.Edge.Test/helpers/Context.cs
damonbarry, line 166: This new parameter should be documented in `tests/README.md`.

damonbarry (changes requested):

test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
lfitchett: yea, 10 is arbitrary, but I wanted it to fail with a specific error instead of just timeout so its easier to debug

CodeCoverage.runsettings
```diff
@@ -67,7 +67,7 @@ Included items must then not match any entries in the exclude list to remain inc
                 <ModulePath>.*leafdevice\..*</ModulePath>
                 <ModulePath>.*load-gen\..*</ModulePath>
                 <ModulePath>.*metricscollector\..*</ModulePath>
-                <ModulePath>.*metricsvalidator\..*</ModulePath>
```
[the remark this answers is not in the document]
lfitchett: yep, I read that as metricscollector and thought it was a duplicate

test/README.md
```diff
@@ -46,6 +46,7 @@ The end-to-end tests take several parameters, which they expect to find in a fil
 | `tempFilterFuncImage` | * | Azure temperature filter function to be used. Required when running the test 'TempFilterFunc', ignored otherwise.|
 | `tempFilterImage` | * | Docker image to pull/use for the temperature filter module. Required when running the test 'TempFilter', ignored otherwise.|
 | `tempSensorImage` || Docker image to pull/use for the temperature sensor module (see the test 'TempSensor'). If not given, `mcr.microsoft.com/azureiotedge-simulated-temperature-sensor:1.0` is used.|
+| `numberLoggerImage` || Docker image to pull/use for the direct method tests. Used to generate predictable logs. |
```
damonbarry, line 49: Nit: "Docker image to pull/use for the **Edge agent** direct method tests..." (since there are other direct method tests which don't use this module)

test/modules/NumberLogger/src/Program.cs
```diff
@@ -0,0 +1,33 @@
...
+            int parsedCount = int.Parse(countString);
+
+            for (int i = 0; i < parsedCount; i++)
+            {
+                Console.WriteLine(i);
+            }
+
+            try
+            {
+                Task.Delay(-1, cts.Token).Wait();
```
damonbarry, line 26: I think the `Task.Delay` should be replaced with `await cts.Token.WhenCanceled();`

test/modules/NumberLogger/src/Program.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+            catch
+            {
+            }
+        }
```
damonbarry, line 31: Before exiting the function, it looks like the common pattern if you're using the EdgeUtil stuff is:
```csharp
    completed.Set();
```

test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
lfitchett: fixed

damonbarry (approved):

PR #3571, opened 2020-09-18
Remove redundant AuthZ from DeviceScopeController

richma-ms:
OnBehalfOf AuthZ is now centrally handled by DeviceScopeAuthenticator, so there's no need for the individual controller implementations to handle it anymore.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/DeviceScopeController.cs
```diff
@@ -85,15 +99,9 @@ async Task<EdgeHubScopeResult> HandleDevicesAndModulesInTargetDeviceScopeAsync(s
                 return new EdgeHubScopeResultError(HttpStatusCode.BadRequest, Events.InvalidRequestAuthchain(request.AuthChain));
             }
 
-            // Check that the actor device is authorized to act OnBehalfOf the target
+            // Get the children of the target device and the target device itself;
             IEdgeHub edgeHub = await this.edgeHubGetter;
             IDeviceScopeIdentitiesCache identitiesCache = edgeHub.GetDeviceScopeIdentitiesCache();
-            if (!await this.AuthorizeActorAsync(identitiesCache, actorDeviceId, actorModuleId, targetDeviceId))
```
varunpuranik, line 91: So this is done implicitly ay the Authenticate call?

varunpuranik (approved):

PR #3574, opened 2020-09-20
Fix for wsl2

veyalla:
Built-in Moby engine is not used in LCOW configuration. However, the service remains active and consumes the DNS port (53) that caused WSL2 to fail (https://github.com/microsoft/WSL/issues/4929).

This PR makes two simple changes when running w/ LCOW:  (a) disables the built-in Moby engine  and (b) removes the iotedge service dependency. It also doesn't enable the "Containers" Windows feature which is only required for Windows containers.

### DIFF
--- a/scripts/windows/setup/IotEdgeSecurityDaemon.ps1
+++ b/scripts/windows/setup/IotEdgeSecurityDaemon.ps1
@@ -371,6 +371,7 @@ function Initialize-IoTEdge {
     Set-Hostname
     if ($ContainerOs -eq 'Linux') {
         Set-ListenConnectUriForLinuxContainers
+        Disable-IoTEdgeMoby
     }
     else {
         Set-CorrectProgramData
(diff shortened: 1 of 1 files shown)


veyalla:
#ready-to-merge

scripts/windows/setup/IotEdgeSecurityDaemon.ps1
```diff
@@ -1963,6 +1964,12 @@ function Set-ListenConnectUriForLinuxContainers {
     $env:IOTEDGE_HOST = "${listenAddress}:15580"
 }
 
+function Disable-IoTEdgeMoby {
+    Invoke-Native "sc config iotedge depend= """""
```
arsing, line 1968: Use `'` for strings without interpolation, which will also make it so you don't have to escape the `"`.

Same for the next line.

arsing (changes requested):

arsing (approved):

PR #3575, opened 2020-09-21
Disabling flaky RefreshCacheTest to unblock checkins

dylanbronson:
Disabling RefreshCacheTest while we figure out what is wrong with it. It is blocking our checkins.


varunpuranik (dismissed):

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/DeviceScopeIdentitiesCacheTest.cs
```diff
@@ -46,7 +46,7 @@ public async Task InitializeFromStoreTest()
             Assert.True(si2.Equals(receivedServiceIdentity2.OrDefault()));
         }
 
-        [Fact]
+        [Fact(Skip = "Consistently flakey test - bug logged")]
         public async Task RefreshCacheTest()
```
richma-ms, line 50: Can disable RefreshCacheNegativeCachingTest for now as well?  I've been seeing it being flakey in the pipeline from time to time.

richma-ms (approved):

PR #3576, opened 2020-09-21
Cherry pick twinNotificationChange source bug fix 

dylanbronson:
Cherry picking https://github.com/Azure/iotedge/pull/3568 to release/1.0.10 branch


varunpuranik (changes requested):
Thanks, we need to check if we want to take this for 1.0.10. So let's keep it open for now, I will confirm and then we can merge it.

varunpuranik (approved):

yophilav (approved):

PR #3582, opened 2020-09-22
Setup Nested Edge E2E Test pipline

yophilav:
Setup Nested Edge E2E Test pipline


builds/e2e/nested-e2e.yaml
```diff
@@ -0,0 +1,46 @@
...
+  pipelines:
+  - pipeline: images
+    source: 'Azure-IoT-Edge-Core Build Images'
+    branch: master
+    trigger:
+      branches:
+      - master
+  - pipeline: packages
+    source: 'Azure-IoT-Edge-Core Edgelet Packages'
+    branch: master
```
philipktlin, line 14: it seems this is optional, will it pick up the triggered branch?

builds/e2e/nested-e2e.yaml
```diff
@@ -0,0 +1,46 @@
...
+    trigger:
+      branches:
+      - master
```
philipktlin, line 17: It is more flexile if it is defined as pipeline variable with default set to master.

builds/e2e/nested-e2e.yaml
philipktlin: btw I think we want to run it explicitly.  Therefore we should use branch argument and get this from pipeline variable.

damonbarry: @philipktlin This is structured exactly the same as the new end-to-end test pipeline. It is automatically triggered whenever the 'Build Images' or 'Edgelet Packages' pipeline completes a run based on master. If you run the pipeline manually, it will automatically pick the most recent successful runs of 'Build Images' and 'Edgelet Pipelines' from master, but you can override the defaults with your own build IDs if desired. Does that match what you're thinking?

@yophilav I thought you wanted this pipeline to _only_ run manually. If that's the case, you'll want to remove the `trigger`s from each resource pipeline here.

philipktlin (approved):

builds/e2e/nested-e2e.yaml
```diff
@@ -0,0 +1,38 @@
+trigger: none
+pr: none
+
+resources:
+  pipelines:
+  - pipeline: images
+    source: 'Azure-IoT-Edge-Core Build Images'
```
damonbarry, line 7: You removed `trigger`, which for sure is good. You also removed `branch`, which means that unless the user explicitly gives build IDs for 'Build Images' and 'Edgelet Packages' builds, it will pick the most recent successful run _from any branch, including PRs_. I think you want to leave `branch: master` here as a default.

```
resources:
  pipelines:
  - pipeline: images
    source: 'Azure-IoT-Edge-Core Build Images'
    branch: master
  - pipeline: packages
    source: 'Azure-IoT-Edge-Core Edgelet Packages'
    branch: master
```

damonbarry (changes requested):
See my comment about the default branch...

damonbarry (approved):

PR #3586, opened 2020-09-22
Some enhancements to the validation errors

vadim-kovalyov:
- Added associated error type for `PolicyValidator`.
- Improved default validator

### DIFF
--- a/mqtt/policy/src/core/builder.rs
+++ b/mqtt/policy/src/core/builder.rs
@@ -121,10 +121,9 @@ where
     }
 
     fn validate(&self, definition: &PolicyDefinition) -> Result<()> {
-        if let Err(e) = self.validator.validate(definition) {
-            return Err(Error::Validation(e.into()));
-        }
-        Ok(())
+        self.validator
+            .validate(definition)
+            .map_err(|e| Error::Validation(e.into()))
     }
 }
 
--- a/mqtt/policy/src/core/builder.rs
+++ b/mqtt/policy/src/core/builder.rs
@@ -1,3 +1,5 @@
+use std::error::Error as StdError;
+
 use lazy_static::lazy_static;
 use regex::Regex;
 use serde::Deserialize;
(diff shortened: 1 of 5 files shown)


mqtt/policy/src/core/builder.rs
```diff
@@ -119,7 +121,10 @@ where
     }
 
     fn validate(&self, definition: &PolicyDefinition) -> Result<()> {
-        self.validator.validate(definition)
+        if let Err(e) = self.validator.validate(definition) {
```
dmolokanov, line 124:
```suggestion
        self.validator.validate(definition).map_err(|e|Error::Validation(e.into()))
```

mqtt/policy/src/core/builder.rs
```diff
@@ -687,17 +706,17 @@ mod tests {
...
     struct FailAllValidator;
 
     impl PolicyValidator for FailAllValidator {
-        fn validate(&self, _definition: &PolicyDefinition) -> Result<()> {
-            Err(Error::ValidationSummary(vec![Error::Validation(
-                "Unable to parse policy definition".to_string(),
-            )]))
+        type Error = ValidatorError;
+
+        fn validate(&self, _definition: &PolicyDefinition) -> StdResult<(), Self::Error> {
```
dmolokanov, line 718: will it complain if you put just Result<(), Self::Error> ?

mqtt/policy/src/validator.rs
```diff
@@ -1,19 +1,53 @@
...
+            .collect::<Vec<_>>();
+
+        if !errors.is_empty() {
+            return Err(ValidatorError::ValidationSummary(errors));
+        }
         Ok(())
     }
 }
+
+fn visit_statement(statement: &Statement) -> Vec<String> {
```
dmolokanov, line 38: this Vec<String> looks suspicious as a list of errors

dmolokanov (approved):
LGTM

mqtt/policy/src/core/builder.rs
vadim-kovalyov: Yeah, b/c local Result type is imported from super::*.

mqtt/policy/src/validator.rs
vadim-kovalyov: it what way? any suggestions? not sure what you mean.

dmolokanov: I mean there is a method that returns `Vec<String>` and the last thing I'd imagine looking at the signature, that it will return a list of errors as strings.

dmolokanov (approved):
:shipit:

PR #3593, opened 2020-09-23
Added mqtt-policy crate

vadim-kovalyov:
This is a second to last PR in a series of PRs to integrate Policy Engine into the Broker.

In this PR I've added a crate that contains MQTT specific implementation of Policy engine plugins.

### DIFF
--- a/mqtt/mqtt-broker/src/auth/authorization.rs
+++ b/mqtt/mqtt-broker/src/auth/authorization.rs
@@ -72,7 +72,7 @@ impl Authorizer for AllowAll {
 }
 
 /// Describes a client activity to authorized.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct Activity {
     client_id: ClientId,
     client_info: ClientInfo,
@@ -110,7 +110,7 @@ impl Activity {
 }
 
 /// Describes a client operation to be authorized.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub enum Operation {
     Connect(Connect),
     Publish(Publish),
@@ -135,7 +135,7 @@ impl Operation {
 }
 
 /// Represents a client attempt to connect to the broker.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct Connect {
     will: Option<Publication>,
 }
@@ -155,7 +155,7 @@ impl From<proto::Connect> for Connect {
 }
 
 /// Represents a publication description without payload to be used for authorization.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct Publication {
     topic_name: String,
     qos: proto::QoS,
@@ -187,7 +187,7 @@ impl From<proto::Publication> for Publication {
 }
 
 /// Represents a client attempt to publish a new message on a specified MQTT topic.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct Publish {
     publication: Publication,
 }
--- a/mqtt/mqtt-broker/src/auth/authorization.rs
+++ b/mqtt/mqtt-broker/src/auth/authorization.rs
@@ -12,7 +12,7 @@ pub trait Authorizer {
     /// Authorizes a MQTT client to perform some action.
     fn authorize(&self, activity: Activity) -> Result<Authorization, Self::Error>;
 
-    fn update(&self, _update: Box<dyn Any>) -> Result<(), Self::Error> {
+    fn update(&mut self, _update: Box<dyn Any>) -> Result<(), Self::Error> {
         Ok(())
     }
 }
@@ -72,6 +72,7 @@ impl Authorizer for AllowAll {
 }
 
 /// Describes a client activity to authorized.
+#[derive(Clone, Debug)]
 pub struct Activity {
     client_id: ClientId,
     client_info: ClientInfo,
@@ -91,6 +92,10 @@ impl Activity {
         }
     }
 
+    pub fn client_id(&self) -> &ClientId {
+        &self.client_id
+    }
+
     pub fn client_info(&self) -> &ClientInfo {
         &self.client_info
     }
@@ -105,6 +110,7 @@ impl Activity {
 }
 
 /// Describes a client operation to be authorized.
+#[derive(Clone, Debug)]
 pub enum Operation {
     Connect(Connect),
     Publish(Publish),
@@ -129,6 +135,7 @@ impl Operation {
 }
 
 /// Represents a client attempt to connect to the broker.
+#[derive(Clone, Debug)]
 pub struct Connect {
     will: Option<Publication>,
 }
@@ -148,6 +155,7 @@ impl From<proto::Connect> for Connect {
 }
 
 /// Represents a publication description without payload to be used for authorization.
+#[derive(Clone, Debug)]
 pub struct Publication {
     topic_name: String,
     qos: proto::QoS,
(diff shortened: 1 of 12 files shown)


mqtt/mqtt-policy/src/substituter.rs
```diff
@@ -0,0 +1,261 @@
...
+                return match variable {
+                    "{{mqtt:client_id}}" => replace(value, variable, context.client_id().as_str()),
+                    "{{iot:identity}}" => replace(
+                        value,
+                        variable,
+                        &context.client_info().auth_id().to_string(),
+                    ),
+                    "{{iot:device_id}}" => replace(value, variable, &extract_device_id(&context)),
+                    "{{iot:module_id}}" => replace(value, variable, &extract_module_id(&context)),
+                    "{{iot:this_device_id}}" => replace(value, variable, self.device_id()),
```
dylanbronson, line 39: What is this_devic_id vs. device_id?

vadim-kovalyov: @dylanbronson, it is something that is still being discussed, but the idea is that `this_device_id` will be replaced with current edge hub device id, and `device_id` is the id of the connected device.

mqtt/mqtt-policy/src/substituter.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    fn replace_variable(&self, value: &str, context: &Request<Activity>) -> String {
+        if let Some(context) = context.context() {
+            if let Some(variable) = extract_variable(value) {
+                return match variable {
+                    "{{mqtt:client_id}}" => replace(value, variable, context.client_id().as_str()),
+                    "{{iot:identity}}" => replace(
+                        value,
+                        variable,
+                        &context.client_info().auth_id().to_string(),
```
dmolokanov, line 35: AuthId::Anonymous.to_string() -> "*"
is it fine?

mqtt/mqtt-policy/src/substituter.rs
```diff
@@ -0,0 +1,261 @@
...
+        Ok(self.replace_variable(value, context))
+    }
+
+    fn visit_resource(&self, value: &str, context: &Request<Self::Context>) -> Result<String> {
+        Ok(self.replace_variable(value, context))
+    }
+}
+
+pub(super) fn extract_variable(value: &str) -> Option<&str> {
+    if let Some(start) = value.find("{{") {
```
dmolokanov, line 62: Regex in rust also a pretty fast :)

dmolokanov: Did you validate something like an incorrect variable name somewhere? eg `{{{}foo}}}`

mqtt/mqtt-policy/src/substituter.rs
```diff
@@ -0,0 +1,261 @@
...
+fn replace(value: &str, variable: &str, substitution: &str) -> String {
+    value.replace(variable, substitution)
+}
+
+fn extract_device_id(activity: &Activity) -> String {
+    let auth_id = activity.client_info().auth_id().to_string();
+    auth_id
+        .split('/')
+        .next()
+        .map(str::to_owned)
```
dmolokanov, line 87: Allocating a String that later  o use a &str seems like a bit of overhead

mqtt/mqtt-policy/src/validator.rs
```diff
@@ -0,0 +1,322 @@
...
+        }
+    }
+    if TopicFilter::from_str(value).is_err() {
+        return Err(Error::InvalidResource(value.into()));
+    }
+    Ok(())
+}
+
+fn is_connect_op(statement: &Statement) -> bool {
+    statement.operations().len() == 1 && statement.operations()[0] == "mqtt:connect"
```
dmolokanov, line 115:
```suggestion
    statement.operations().first(|op| op == "mqtt:connect").is_some()
```

mqtt/mqtt-broker/src/auth/authorization.rs
```diff
@@ -72,6 +72,7 @@ impl Authorizer for AllowAll {
 }
 
 /// Describes a client activity to authorized.
+#[derive(Clone, Debug)]
```
dmolokanov, line 75: Do you use these clones anywhere?

mqtt/mqtt-policy/src/validator.rs
```diff
@@ -0,0 +1,322 @@
...
+}
+
+lazy_static! {
+    static ref VALID_IDENTITY_VARIABLES: HashSet<String> = HashSet::from_iter(vec![
+        "{{iot:identity}}".into(),
```
dmolokanov, line 120: should you make it consts and use it in both places here and in the Substituer

mqtt/mqtt-policy/src/validator.rs
```diff
@@ -0,0 +1,322 @@
...
+    }
+    Ok(())
+}
+
+fn is_connect_op(statement: &Statement) -> bool {
+    statement.operations().len() == 1 && statement.operations()[0] == "mqtt:connect"
+}
+
+lazy_static! {
+    static ref VALID_IDENTITY_VARIABLES: HashSet<String> = HashSet::from_iter(vec![
```
dmolokanov, line 119: I wonder if using small array [&str] will be more performant than HashSet. because of compiler optimizations and data fetching

mqtt/mqtt-policy/src/substituter.rs
vadim-kovalyov: I wasn't sure and wanted to avoid that :)

mqtt/mqtt-policy/src/validator.rs
vadim-kovalyov: I specifically need to check that `.len() == 1`.

mqtt/mqtt-policy/src/substituter.rs
vadim-kovalyov: good point

mqtt/mqtt-policy/src/substituter.rs
vadim-kovalyov: Good question. It is fine for now, since we don't support Anonymous for now. I have a note to think about that later.

mqtt/mqtt-broker/src/auth/authorization.rs
vadim-kovalyov: you're right, probably I needed them at some point, but now I don't use them.

mqtt/mqtt-policy/src/substituter.rs
vadim-kovalyov: >Did you validate something like an incorrect variable name somewhere? eg {{{}foo}}}

I still hope to add proptests sometime soon:)

mqtt/mqtt-broker/src/auth/mod.rs
```diff
@@ -27,6 +27,15 @@ pub enum AuthId {
...
+        match self {
+            AuthId::Anonymous => "*",
+            AuthId::Identity(identity) => identity.as_str(),
+        }
+    }
+}
+
 impl Display for AuthId {
     fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
         match self {
```
dmolokanov, line 41: Could you please change this method as well?
```suggestion
        write!(f, self.as_str())
```

mqtt/mqtt-policy/src/validator.rs
dmolokanov: Cool. Could you please add a comment on why you need exactly 1 here? Someone could try to improve code...

vadim-kovalyov: someone like you? :D done!

dmolokanov: we'll never know

dmolokanov (approved):
Very clean code! Really enjoyed reviewing the PR! And now :shipit:

PR #3597, opened 2020-09-23
Mqtt bridge upstream connection

ancaantochi:
- Added TLS stream to connect to upstream and set certs from workload api
- Changed mqtt3 client to send subscription rejected
- Added setting for turn on/off bridge (it uses UpstreamProtocol for now, but there might be a different env variable to be used)

Note: Doesn't handle the client error and recreate yet.


mqtt/mqtt3/src/client/mod.rs
```diff
@@ -520,6 +520,7 @@ pub enum Event {
 pub enum SubscriptionUpdateEvent {
     Subscribe(crate::proto::SubscribeTo),
     Unsubscribe(String),
+    SubscriptionRejectedByServer(String),
```
and-rewsmith, line 523: What was mqtt3 doing before if it was rejected?

and-rewsmith (commented):
I looked very briefly. Will look again later.

mqtt/mqtt3/src/client/mod.rs
ancaantochi: reconnecting the client and retry to send the subscription, no event was sent to caller

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,84 +40,189 @@ impl ShutdownHandle {
...
-                provider_settings.device_id().into(),
+                format!(
+                    "{}/{}",
+                    provider_settings.device_id().to_owned(),
+                    provider_settings.module_id().to_owned()
+                ),
                 //TODO: handle properties that are sent by client in username (modelId, authchain)
                 Some(format!(
-                    "{}/{}/{}",
+                    "{}/{}/{}/?api-version={}",
```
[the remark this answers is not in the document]
ancaantochi: it's in the sdk when the token is validated which is used by authenticator in edge hub core

mqtt/mqtt-bridge/src/token_source.rs
```diff
@@ -29,50 +31,88 @@ impl SasTokenSource {
 
 #[async_trait]
 impl TokenSource for SasTokenSource {
-    async fn get(&self, expiry: &DateTime<Utc>) -> Result<String, TokenSourceError> {
-        let token: String = match &self.creds {
+    async fn get(&self, expiry: &DateTime<Utc>) -> Result<Option<String>, Error> {
```
and-rewsmith, line 34: Why would this return ```Option<String>```?

mqtt/mqtt-bridge/src/token_source.rs
```diff
@@ -29,50 +31,88 @@ impl SasTokenSource {
...
+pub struct TrustBundleSource {
+    creds: Credentials,
+}
+
+impl TrustBundleSource {
+    pub fn new(creds: Credentials) -> Self {
+        Self { creds }
+    }
+
+    pub async fn get_trust_bundle(&self) -> Result<Option<String>, Error> {
```
and-rewsmith, line 94: Similar question as above. Why ```Option<String>```? In what cases would the trusted CA be None? Isn't this an error?

mqtt/mqtt-edgehub/src/command/handler.rs
```diff
@@ -154,8 +154,14 @@ async fn subscribe(
...
-                if let SubscriptionUpdateEvent::Subscribe(sub) = subscription {
-                    subacks.remove(&sub.topic_filter);
+                match subscription {
+                    SubscriptionUpdateEvent::Subscribe(sub) => {
+                        subacks.remove(&sub.topic_filter);
+                    }
+                    SubscriptionUpdateEvent::SubscriptionRejectedByServer(sub) => {
+                        return Err(CommandHandlerError::RejectedByServer(sub));
+                    }
+                    SubscriptionUpdateEvent::Unsubscribe(_) => {}
```
and-rewsmith, line 164: Should we log something?

mqtt/mqtt3/src/client/subscriptions.rs
```diff
@@ -32,119 +32,117 @@ impl State {
...
-                                        ));
-                                    }
+                        packet_identifiers.discard(packet_identifier);
 
-                                    self.subscriptions.insert(topic_filter, expected_qos);
+                        // We can't put subscribe_to back into self.subscription_updates_waiting_to_be_acked within the below loop
+                        // since we would've partially consumed it.
+                        // Instead, if there's an error, we'll update self.subscriptions anyway with the expected QoS, and set the error to be returned here.
+                        // The error will reset the session and resend the subscription requests, including these that didn't match the expected QoS,
+                        // so pretending the subscription succeeded does no harm.
```
and-rewsmith, line 73: We need to update this TODO right?

and-rewsmith:
LGTM mostly with some minor nit comments.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,84 +40,189 @@ impl ShutdownHandle {
...
+                let token_source = connect.token_source.as_ref().cloned();
+                let host = port.map_or(address.clone(), |p| format!("{}:{}", address.clone(), p));
+
+                Box::pin(async move {
+                    let expiry =
+                        Utc::now() + chrono::Duration::minutes(DEFAULT_TOKEN_DURATION_MINS);
+
+                    let io = TcpStream::connect(&host);
+
+                    let token = if let Some(ref ts) = token_source {
```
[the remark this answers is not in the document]
ancaantochi: the returned types are different

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,84 +40,189 @@ impl ShutdownHandle {
...
-    type Io = TcpStream;
-    type Error = std::io::Error;
-    type Future = BoxFuture<'static, Result<(TcpStream, Option<String>), std::io::Error>>;
+impl IoSource for BridgeIoSource {
+    type Io = Pin<Box<dyn BridgeIo>>;
+    type Error = Error;
+    #[allow(clippy::type_complexity)]
+    type Future = BoxFuture<'static, Result<(Self::Io, Option<String>), Error>>;
 
     fn connect(&mut self) -> Self::Future {
```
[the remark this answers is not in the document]
ancaantochi: yes, the proxy module has similar code

mqtt/mqtt-bridge/src/token_source.rs
ancaantochi: if there is no workload api it doesn't get any certs, and it should be fine if upstream has a cert signed by a known CA

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,84 +40,189 @@ impl ShutdownHandle {
...
+                Box::pin(async move {
+                    let expiry =
+                        Utc::now() + chrono::Duration::minutes(DEFAULT_TOKEN_DURATION_MINS);
+
+                    let io = TcpStream::connect(&host);
+
+                    let token = if let Some(ref ts) = token_source {
+                        futures_util::future::Either::Left(ts.get(&expiry))
+                    } else {
+                        futures_util::future::Either::Right(futures_util::future::ok(None))
```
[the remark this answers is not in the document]
ancaantochi: When the creds are Anonymous

mqtt/mqtt-bridge/src/token_source.rs
ancaantochi: for anonymous credential type

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,84 +40,189 @@ impl ShutdownHandle {
...
-                    })
-                    .ok()
-            } else {
-                None
-            };
-
-            let io = TcpStream::connect(address).await;
-            io.map(|io| (io, password))
-        })
+        match self {
```
vadim-kovalyov, line 90: this code seems too long. Try to extract methods

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,84 +40,189 @@ impl ShutdownHandle {
...
+                        Error::new(
+                            ErrorKind::Other,
+                            format!("could not create TLS connector: {}", err),
+                        )
+                    })?;
+                    let connector: tokio_tls::TlsConnector = connector.into();
+
+                    let io = connector.connect(&address, stream).await;
+
+                    debug!("Tls connection {:?} for {:?}", io, address);
```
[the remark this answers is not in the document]
ancaantochi: For Error is displaying Err and the reason and for Success it displayed success and cert validation passed. I think it's useful for debug

mqtt/mqtt-bridge/src/client.rs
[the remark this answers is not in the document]
dmolokanov: yeah, it is pre async/await code style
```
let token = async {
    match token_source {
        Some(ts) => ts.get(&expiry).await,
        None => Ok(None),
    }
};
```

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,84 +40,189 @@ impl ShutdownHandle {
...
+
+                    let stream: Pin<Box<dyn BridgeIo>> = Box::pin(io);
+                    Ok((stream, password))
+                })
+            }
+            BridgeIoSource::Tls(connect) => {
+                let address = connect.address.clone();
+                let port = connect.port.clone();
+                let token_source = connect.token_source.as_ref().cloned();
+                let trust_bundle_source = connect.trust_bundle_source.clone();
```
dmolokanov, line 125: we could use async block instead of Either thingy

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -194,20 +318,36 @@ impl<T: EventHandler> MqttClient<T> {
...
-            //right now it can't detect if the the broker doesn't allow to subscribe to the specific topic, but it will send a connect event
-            if let Event::NewConnection { reset_session: _ } = event {
-                return Ok(());
+            if let Event::SubscriptionUpdates(subscriptions) = event {
+                for subscription in subscriptions {
+                    match subscription {
+                        SubscriptionUpdateEvent::Subscribe(sub) => {
+                            subacks.remove(&sub.topic_filter);
+                        }
+                        SubscriptionUpdateEvent::SubscriptionRejectedByServer(sub) => {
```
dmolokanov: it is already SubscriptionEvent, maybe just `RejectedByServer`?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -194,20 +318,36 @@ impl<T: EventHandler> MqttClient<T> {
...
+                            subacks.remove(&sub);
+                            error!("subscription rejected by server {}", sub);
```
dmolokanov: I think it is a error for subscribing to local broker, but just warning for remote, don't you agree?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -194,20 +318,36 @@ impl<T: EventHandler> MqttClient<T> {
...
-                .collect(),
-        ))
+        if subacks.is_empty() {
+            debug!("successfully subscribed to topics");
+        } else {
+            error!(
+                "failed to receive expected subacks for topics: {0:?}",
+                subacks
+                    .iter()
+                    .map(std::string::ToString::to_string)
```
dmolokanov, line 354: .map(ToString::to_string)

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -194,20 +318,36 @@ impl<T: EventHandler> MqttClient<T> {
...
+                    .collect::<String>()
```
dmolokanov, line 355: nit: 
.collect()

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -22,7 +24,7 @@ impl Settings {
         let mut config = Config::new();
 
         config.merge(File::from_str(DEFAULTS, FileFormat::Json))?;
-        config.merge(Environment::with_prefix(ENVIRONMENT_PREFIX))?;
+        config.merge(Environment::new())?;
```
dmolokanov, line 26: could you please get more details on this?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,84 +40,189 @@ impl ShutdownHandle {
...
-
-            let io = TcpStream::connect(address).await;
-            io.map(|io| (io, password))
-        })
+        match self {
+            BridgeIoSource::Tcp(connect) => {
+                let address = connect.address.clone();
+                let port = connect.port.clone();
+                let token_source = connect.token_source.as_ref().cloned();
+                let host = port.map_or(address.clone(), |p| format!("{}:{}", address.clone(), p));
```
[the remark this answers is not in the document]
ancaantochi: I got errors if not cloning: cannot move out of `connect_settings.token_source` which is behind a mutable reference

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -77,15 +79,29 @@ impl<'de> serde::Deserialize<'de> for Settings {
...
-            name: "upstream".into(),
-            address: nested_bridge.gateway_hostname.clone(),
-            subscriptions: upstream.subscriptions,
-            forwards: upstream.forwards,
-            credentials: Credentials::Provider(nested_bridge),
-            clean_session: upstream.clean_session,
-            keep_alive: upstream.keep_alive,
-        });
+        let upstream_connection_settings = nested_bridge
+            .filter(|nested_bridge| {
```
dmolokanov, line 82: nit: filtering looks a bit confusing.

mqtt/mqtt-bridge/src/token_source.rs
```diff
@@ -13,7 +15,7 @@ define_encode_set! {
 
 #[async_trait]
 pub trait TokenSource {
-    async fn get(&self, expiry: &DateTime<Utc>) -> Result<String, TokenSourceError>;
+    async fn get(&self, expiry: &DateTime<Utc>) -> Result<Option<String>, Error>;
```
dmolokanov, line 18: I don't think we need to use std::io::Error for our business code errors. It's better we use our own enums

dmolokanov (changes requested):
Nice work diving into mqtt client code!

mqtt/mqtt-bridge/src/token_source.rs
ancaantochi: I changed because it was needed for try_join and it's not exposed from the client

mqtt/mqtt-bridge/src/settings.rs
ancaantochi: it used to filter for env variables starting with IOTEDGE, but then I added a new env to enable/disable bridge which doesn't start with the prefix so changed this to read all env variables

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -110,9 +108,9 @@ impl Bridge {
         &self,
         mut topics: HashMap<String, Topic>,
         address: &str,
-        keep_alive: Duration,
-        clean_session: bool,
+        port: Option<String>,
```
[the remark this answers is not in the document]
ancaantochi: initially it was like that, but tls connection failed because it is using the address:port to connect to tls stream and cert failed, it needs only the address part when it opens the tls stream. Other option is no parse the address:port and extract the address

mqtt/mqtt-bridge/src/client.rs
ancaantochi: I get error: subacks.iter().map(ToString::to_string).collect()   cannot infer type for type parameter `B` declared on the associated function `collect`

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -1,23 +1,31 @@
...
-use std::{collections::HashSet, fmt::Display, time::Duration};
+use std::{
+    collections::HashSet, fmt::Display, io::Error, io::ErrorKind, pin::Pin, str, time::Duration,
+};
 
 use async_trait::async_trait;
 use chrono::Utc;
 use futures_util::future::BoxFuture;
-use tokio::{net::TcpStream, stream::StreamExt};
+use native_tls::{Certificate, TlsConnector};
```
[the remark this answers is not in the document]
ancaantochi: replaced with tokio_openssl

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,84 +40,189 @@ impl ShutdownHandle {
...
 impl<T: EventHandler> MqttClient<T> {
     pub fn new(
         address: &str,
+        port: Option<String>,
         keep_alive: Duration,
-        clean_session: bool,
+        _clean_session: bool,
         event_handler: T,
         connection_credentials: &Credentials,
+        secure: bool,
```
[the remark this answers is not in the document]
ancaantochi: there should be a bool at some point to configure if it's tls or not so if I add 2 ctors I will have to check the value and call tls() or tcp(), so it moves the decision up

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -123,6 +141,10 @@ impl ConnectionSettings {
         &self.address
     }
 
+    pub fn port(&self) -> &str {
```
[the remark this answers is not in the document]
ancaantochi: reverted to be included in the address

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -158,8 +280,10 @@ impl<T: EventHandler> MqttClient<T> {
     pub async fn handle_events(mut self) -> Result<(), ClientConnectError> {
         while let Some(event) = self.client.try_next().await.unwrap_or_else(|e| {
             error!(message = "failed to poll events", error=%e);
+            // TODO: handle the error and recreat the client/bridge
             None
         }) {
+            debug!("handle event {:?}", event);
```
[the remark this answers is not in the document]
ancaantochi: Ok with the Event type and then each property. Probably I bit to much to log every message,  it might be good for initial testing, then I can change and just log message received.

mqtt/mqtt3/src/client/subscriptions.rs
ancaantochi: added a comment where it returns the event, I think this part still makes sense for other cases

mqtt/mqtt-bridge/src/settings.rs
ancaantochi: I changed it a little

mqtt/mqtt-bridge/src/client.rs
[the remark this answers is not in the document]
dmolokanov: also, it can be mqtt over websockets, right?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,55 +40,157 @@ impl ShutdownHandle {
...
+        let token_source = connection_settings.token_source;
+
+        Box::pin(async move {
+            let expiry = Utc::now() + chrono::Duration::minutes(DEFAULT_TOKEN_DURATION_MINS);
+
+            let io = TcpStream::connect(&address);
+
+            let token_task = async {
+                match token_source {
+                    Some(ref ts) => ts.get(&expiry).await,
```
dmolokanov, line 109:
```suggestion
                    Some(ts) => ts.get(&expiry).await,
```

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -1,23 +1,31 @@
...
 
 use async_trait::async_trait;
 use chrono::Utc;
-use futures_util::future::BoxFuture;
-use tokio::{net::TcpStream, stream::StreamExt};
-use tracing::{debug, error};
+use futures_util::future::{self, try_join3, BoxFuture};
+use openssl::{ssl::SslConnector, ssl::SslMethod, x509::X509};
+use tokio::{io::AsyncRead, io::AsyncWrite, net::TcpStream, stream::StreamExt};
+use tokio_openssl::connect;
```
dmolokanov, line 11: we usually don't import functions from modules, and use with module name instead. it helps to avoid ambighuity: methods is in the current module or from somewhere else.

dmolokanov: same for try_join_3

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,55 +40,157 @@ impl ShutdownHandle {
...
-                        error!("Failed to get token for connection {} {}", address, e);
-                        e
-                    })
-                    .ok()
+                    .map_err(|err| {
+                        Error::new(ErrorKind::Other, format!("failed to connect: {}", err))
+                    })?;
+
+            let config = SslConnector::builder(SslMethod::tls())
+                .map(|mut builder| {
```
dmolokanov, line 155: it is a bit too many combinators I guess...

dmolokanov: SHouldn't we return an error instead of silently converting it to Option?

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -221,8 +211,8 @@ pub struct CredentialProviderSettings {
 }
 
 impl CredentialProviderSettings {
-    pub fn enable_upstream_bridge(&self) -> Option<bool> {
-        self.enable_upstream_bridge
+    pub fn enable_upstream_bridge(&self) -> Option<&str> {
+        self.enable_upstream_bridge.as_ref().map(AsRef::as_ref)
```
dmolokanov, line 215:
```suggestion
        self.enable_upstream_bridge.as_deref()
```

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -80,21 +79,18 @@ impl<'de> serde::Deserialize<'de> for Settings {
...
-                    |p| {
-                        debug!("upstream bridge setting {}", p);
-                        p
-                    },
-                )
+                nested_bridge
+                    .enable_upstream_bridge()
+                    .unwrap_or("false")
+                    .to_lowercase()
+                    == "true"
```
dmolokanov, line 86: looks weird.. does serde have problems converting to bool?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -89,108 +88,110 @@ impl IoSource for BridgeIoSource {
...
-                        )
-                    })?;
-                    let connector: tokio_tls::TlsConnector = connector.into();
+                    builder.build()
+                })
+                .and_then(|conn| conn.configure())
+                .ok();
 
-                    let io = connector.connect(&address, stream).await;
+            let hostname = address
```
dmolokanov, line 171: no cloning
```suggestion
            let hostname = address.split(':').next().unwrap_or(&address);
```

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,55 +39,154 @@ impl ShutdownHandle {
...
+    type Error = Error;
+    #[allow(clippy::type_complexity)]
+    type Future = BoxFuture<'static, Result<(Self::Io, Option<String>), Error>>;
 
     fn connect(&mut self) -> Self::Future {
-        let address = self.address.clone();
-        let token_source = self.token_source.as_ref().cloned();
+        match self {
+            BridgeIoSource::Tcp(connect_settings) => Self::get_tcp_source(connect_settings.clone()),
+            BridgeIoSource::Tls(connect_settings) => Self::get_tls_source(&connect_settings),
```
vadim-kovalyov, line 91: nit: I would make both method consistend - pass by value (clone()).

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,55 +39,154 @@ impl ShutdownHandle {
...
+                    }
+
+                    builder.build()
+                })
+                .and_then(|conn| conn.configure())
+                .ok();
+
+            let hostname = address.split(':').next().unwrap_or(&address);
+
+            let res = if let Some(c) = config {
```
vadim-kovalyov, line 172: nit: also `config.map(...)`?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -32,55 +39,154 @@ impl ShutdownHandle {
...
+                })
+                .ok()
             } else {
                 None
             };
 
-            let io = TcpStream::connect(address).await;
-            io.map(|io| (io, password))
+            res.map_or(
+                Err(Error::new(ErrorKind::Other, "could not connect")),
```
vadim-kovalyov, line 187: nit: you could do `map_err` on line 181 instead?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -89,27 +195,37 @@ pub struct MqttClient<T>
 where
```
vadim-kovalyov, line 195: Can you please put some comments on the purpose of this client? It's easy to get lost b/w all the clients that we already have, so let's keep it documented :)

mqtt/mqtt-bridge/src/client.rs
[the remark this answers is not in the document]
vadim-kovalyov: I vote for @dmolokanov suggestion. @ancaantochi, I understand your point that we would need to make the decision up in the stack, but I tend to avoid such bools as params and I believe having separate method will lead to a better design.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -158,8 +280,10 @@ impl<T: EventHandler> MqttClient<T> {
     pub async fn handle_events(mut self) -> Result<(), ClientConnectError> {
         while let Some(event) = self.client.try_next().await.unwrap_or_else(|e| {
             error!(message = "failed to poll events", error=%e);
+            // TODO: handle the error by recreting the connection
             None
         }) {
+            debug!("handle event {:?}", event);
```
vadim-kovalyov, line 286: Maybe just bubble the error up and log it there? or, alternatively, you can use `loop { match ... }`
```suggestion
while let Some(event) = self.client.try_next().await? {
```

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -194,20 +318,35 @@ impl<T: EventHandler> MqttClient<T> {
             .await
             .map_err(ClientConnectError::PollClientFailure)?
         {
-            //TODO: change the mqtt client to send an error back when the subscriotion fails instead of reconnecting and resending the sub
-            //right now it can't detect if the the broker doesn't allow to subscribe to the specific topic, but it will send a connect event
-            if let Event::NewConnection { reset_session: _ } = event {
-                return Ok(());
+            if let Event::SubscriptionUpdates(subscriptions) = event {
+                for subscription in subscriptions {
+                    match subscription {
```
vadim-kovalyov, line 323: Is it guaranteed that .try_next() will return the SubscriptionUpdates and not any other event for already running client? Also, I'm not sure if this loop will terminate properly.

mqtt/mqtt-bridge/src/client.rs
ancaantochi: it should not be different local broker than remote, both they can reject the sub if it is not authorized.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -89,108 +88,110 @@ impl IoSource for BridgeIoSource {
...
-                        let certs = X509::stack_from_pem(trust_bundle.as_bytes())
-                            .unwrap()
-                            .into_iter()
-                            .flat_map(|cert| {
-                                cert.to_der()
-                                    .ok()
-                                    .map(|der| Certificate::from_der(&der).ok())
+                        X509::stack_from_pem(trust_bundle.as_bytes())
+                            .map(|mut certs| {
+                                if let Some(ca) = certs.pop() {
```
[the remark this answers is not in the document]
ancaantochi: yes, changed to  while

mqtt/mqtt-bridge/src/client.rs
ancaantochi: It can be that it returns something else, but the subscription awaits so there should be no other events in the meantime. I think it can be consolidated when Andrew has the changes for the pump.

mqtt/mqtt-bridge/src/client.rs
ancaantochi: this can't be bubbled up because we don't want to end the thread, the missing part is handling this error properly and recreating the client. The issue will be solved by the pump implementation.

mqtt/mqtt-bridge/src/settings.rs
ancaantochi: it's not serde with the issue, it's environment from config, I looked at the code and it sets all env variables as string and that's why it can be deserialized properly. I wanted to write a workaround for that, but then acording to Varun doc there should be no new env variable added to enable bridge. I think it should be UpstreamProtocol which should be mqtt, but I will leave it like this for now until that's clarified

mqtt/mqtt-bridge/src/client.rs
ancaantochi: I couldn't use map because of the await, is there a way to use map with async?

mqtt/mqtt-bridge/src/client.rs
ancaantochi: It would in the end return the error because None would result in an error, but anyway I changed it to map_err here and map_err again at line 180. I though many combinators is the  dream.

mqtt/mqtt-bridge/src/client.rs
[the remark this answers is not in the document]
ancaantochi: > also, it can be mqtt over websockets, right?
Not implemented, I think it might be after public preview

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -273,6 +291,15 @@ impl<T: EventHandler> MqttClient<T> {
         }
     }
 
+    fn get_token_source(connection_credentials: &Credentials) -> Option<SasTokenSource> {
```
dmolokanov, line 294: nit
```suggestion
    fn token_source(connection_credentials: &Credentials) -> Option<SasTokenSource> {
```

mqtt/mqtt-bridge/src/token_source.rs
```diff
@@ -27,50 +27,55 @@ impl SasTokenSource {
     pub fn new(creds: Credentials) -> Self {
         SasTokenSource { creds }
     }
+
+    async fn get_token_from_workload(
```
dmolokanov, line 31: nit
```suggestion
    async fn generate_sas_token(
```

dmolokanov (dismissed):
Some naming nits. Otherwise looks good to me

mqtt/mqtt-bridge/src/client.rs
vadim-kovalyov: I believe you could just mark lambda as async.

dmolokanov: async closures are not in stable rust yet :(

dmolokanov (approved):
:shipit:

PR #3600, opened 2020-09-24
[mqtt] Sidecar startup order

and-rewsmith:
The sidecars currently do not start up separately from the broker, which is not how we want to structure the startup order. This PR fixes this and addresses this TODO:

> // TODO: We need to elevate failable init steps out of the async block running in it's own thread.	
//       This is because we need to have defined setup steps, that once complete, the sidecars can run without failing.	
//       Once this is done, complex startup order will be easier to implement.	
//       We can also stop the broker process if sidecars fail to init.	
//	
//       This change depends on restructuring the startup logic so that:	
//         1. sidecars and server run at same time	
//         2. if one fails the other shuts down


mqtt/mqttd/src/broker/mod.rs
```diff
@@ -5,45 +5,82 @@ mod snapshot;
...
+                handle.await?;
+            }
+
+            sidecar_shutdown.shutdown().await?;
+            broker_handle.send(Message::System(SystemEvent::Shutdown))?;
+            let state = server_join_handle.await??;
+
+            state
+        }
+    };
```
and-rewsmith, line 83: We are probably going to want to log these errors instead of blowing up, right? If you agree how would we handle these nested results cleanly? (i.e. ```let state = server_join_handle??;```)

mqtt/mqttd/src/broker/snapshot.rs
```diff
@@ -1,7 +1,59 @@
...
+    let start = Instant::now() + period;
+    let mut interval = tokio::time::interval_at(start, period);
+    loop {
+        interval.tick().await;
+        if let Err(e) = broker_handle.send(Message::System(SystemEvent::StateSnapshot(
+            snapshot_handle.clone(),
+        ))) {
+            warn!(message = "failed to tick the snapshotter", error=%e);
+        }
+    }
```
and-rewsmith, line 56: I moved all the snapshotter logic to this file.

PR #3605, opened 2020-09-24
[mqtt] Publication delivery confirmation topic for M2M messages

dmolokanov:
Mqtt broker doesn't support back-pressure when edgehub tries to send all messages for m2m communication. To tackle this problem broker will publish a message on a special topic `$edgehub/delivery`  when the receiver sends PUBACK. This way edgehub understands which packets were received by the client.


mqtt/mqtt-broker/src/connection/mod.rs
```diff
@@ -169,15 +169,14 @@ where
                 broker_handle.send(message)?;
 
                 let (outgoing, incoming) = codec.split();
+                let (outgoing_processor, incoming_processor) = make_processor.make(&client_id);
 
                 // prepare processing incoming packets
-                let incoming_processor = make_processor.make_incoming(&client_id);
                 let incoming_task =
-                    incoming_task(client_id.clone(), incoming, broker_handle.clone(), incoming_processor);
+                    incoming_task(client_id.clone(), incoming, broker_handle.clone(),incoming_processor);
```
vadim-kovalyov, line 176: fmt?

mqtt/mqtt-broker/src/lib.rs
```diff
@@ -241,6 +242,7 @@ pub enum SystemEvent {
     StateSnapshot(StateSnapshotHandle),
     ForceClientDisconnect(ClientId),
     AuthorizationUpdate(Box<dyn Any + Send + Sync>),
+    Publish(Publication),
```
vadim-kovalyov, line 245: I think we need to start documenting these variants. I would be good to understand how this is different from normal publish

mqtt/mqtt-edgehub/src/connection/delivery.rs
```diff
@@ -0,0 +1,137 @@
+use std::{collections::HashMap, sync::Arc};
+
+use async_trait::async_trait;
+use lazy_static::lazy_static;
+use mqtt_broker::{
```
vadim-kovalyov, line 5: nit: rearrange

mqtt/mqtt-edgehub/src/connection/delivery.rs
```diff
@@ -0,0 +1,137 @@
...
+
+pub struct PublicationDelivery<P> {
+    inner: P,
+    broker_handle: BrokerHandle,
+    waited_to_be_acked: Arc<Mutex<HashMap<PacketIdentifier, String>>>,
+}
+
+/// MQTT packet processor wrapper. It identifies `IoTHub` M2M outgoing publishes,
+/// saves PACKETID to send a confirmation packet back to sender on special topic
+/// "$edgehub/delivery".
```
vadim-kovalyov, line 22: I would also add "why" this is needed. That this is used as a back pressure mechanism to avoid M2M message loss.

mqtt/mqtt-edgehub/src/connection/delivery.rs
```diff
@@ -0,0 +1,137 @@
...
+        Packet::Publish(proto::Publish {
+            packet_identifier_dup_qos,
+            topic_name,
+            ..
+        }) => (packet_identifier_dup_qos, topic_name),
+        _ => return None,
+    };
+
+    let packet_identifier = match packet_identifier_dup_qos {
+        proto::PacketIdentifierDupQoS::AtLeastOnce(packet_identifier, _) => packet_identifier,
```
vadim-kovalyov, line 91: nit: qos2, no? You check for PubRel on line 44

mqtt/mqtt-edgehub/src/connection/delivery.rs
```diff
@@ -0,0 +1,137 @@
...
+                .insert(packet_identifier, topic_name);
+        }
+    }
+}
+
+fn match_m2m_publish(packet: &Packet) -> Option<(proto::PacketIdentifier, String)> {
+    const DEVICE_ID: &str = r"(?P<device_id>[^/]+)";
+    const MODULE_ID: &str = r"(?P<module_id>[^/]+)";
+    lazy_static! {
+        static ref M2M_PUBLISH_PATTERN: Regex = Regex::new(&format!(
```
vadim-kovalyov, line 74: maybe simplify regex a bit? you don't need named groups, you just need to check it if matches.

mqtt/mqtt-edgehub/tests/delivery.rs
```diff
@@ -0,0 +1,79 @@
...
+    server::{self, DummyAuthenticator},
+};
+use mqtt_edgehub::connection::MakeEdgeHubPacketProcessor;
+
+/// Scenario:
+/// create broker
+/// create command handler
+/// connect client
+/// publish message to disconnect client
+/// verify client disconnected
```
vadim-kovalyov, line 16: copy paste? :)

mqtt/mqtt-broker/src/connection/mod.rs
dmolokanov: seems like there is a bug in the formatter. all checks have passed with ✅

mqtt/mqtt-broker/src/lib.rs
dmolokanov: yeah, normal publish has a ClientId of the publisher. In this case, the broker publishes itself.

mqtt/mqtt-edgehub/src/connection/delivery.rs
dmolokanov: It'd be nice to have a plugin to formatter for custom formatting rules

mqtt/mqtt-edgehub/src/connection/delivery.rs
dmolokanov: good question, since iothub topics are qos1 only, we don't need that in both places

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -79,19 +79,22 @@ where
                                 info!("sent state to snapshotter.");
                             }
                         }
-                        SystemEvent::ForceClientDisconnect(client_id) => {
```
vadim-kovalyov, line 82: is this no longer needed?

vadim-kovalyov (approved):

mqtt/mqtt-broker/src/broker.rs
dmolokanov: Nope, we can re-use `ClientEvent::DropConnection` which has the same semantics

PR #3611, opened 2020-09-25
Nested edge: Fixing iotedge checks

huguesBouvier:
Fixing iotedge checks: IoT edge check will now use parent to performs connectivity checks.

By default, if user perform IoTedge check and a parent hostname is define, it will use the parents instead of edge hub to perform the tests


lfitchett (changes requested):
Looks good, one minor nit, the "host_connect_iothub" should probably be names host_connect_upstream and its description updated to reflect the change

huguesBouvier:
> Looks good, one minor nit, the "host_connect_iothub" should probably be names host_connect_upstream and its description updated to reflect the change

Good point!

lfitchett (dismissed):
LGTM

lfitchett (approved):

PR #3616, opened 2020-09-26
[k8s] Add all root CA certs to TLS connections

darobs:
There were a few places in the code where we were not reading the root CA data correctly. It is possible to have multiple certificates in the file/data, and we were only extracting the first one.

The three places where we need to consider the possibilities of multiple root CAs:
1. iotedged-proxy. iotedged mounts its trusted CAs for proxy to use. 
2. When reading the KubeConfig context's certificate_authority
3. When reading the `serviceaccount/ca.crt`


edgelet/edgelet-http-workload/tests/dns-san.rs
```diff
@@ -3,7 +3,7 @@
 // We disable this test on Windows in order to avoid having to add a dependency
 // on OpenSSL.
 #![cfg(not(windows))]
-#![deny(rust_2018_idioms, warnings)]
```
dmolokanov, line 6: do you think it is a good idea to disable this?

edgelet/iotedge-proxy/src/proxy/config.rs
```diff
@@ -51,10 +51,30 @@ pub fn get_config(settings: &ServiceSettings) -> Result<Config<ValueToken>, Erro
...
-        tls.add_root_certificate(cert);
+        if file.is_empty() {
+            return Err(Error::from(ErrorKind::Initialize(
+                InitializeErrorReason::ClientConfig,
+            )));
+        }
+
+        // Extract each certificate's string. The final string from the split will either be empty
+        // or a non-certificate entry, so it is dropped.
+        let delimiter = "-----END CERTIFICATE-----";
```
dmolokanov, line 62: this parsing looks a bit sketchy. should we use OpenSSL load the chain?

edgelet/iotedge-proxy/src/proxy/config.rs
```diff
@@ -51,10 +51,30 @@ pub fn get_config(settings: &ServiceSettings) -> Result<Config<ValueToken>, Erro
...
+        // replace with split_inclusive once stable.
+        let raw_certs: Vec<&str> = file.split(delimiter).collect();
+        let len = raw_certs.len();
+        if len == 1 {
```
dmolokanov, line 66: Could you please add a comment why it should contain 1+ certs in the file?

edgelet/kube-client/src/lib.rs
```diff
@@ -18,3 +18,109 @@ pub mod kube;
 pub use self::client::{Client, HttpClient};
 pub use self::config::{get_config, Config, TokenSource, ValueToken};
 pub use self::error::{Error, ErrorKind, RequestType};
+
+#[cfg(test)]
+#[allow(dead_code)]
```
dmolokanov, line 23: should we remove that?

edgelet/kube-client/src/lib.rs
```diff
@@ -18,3 +18,109 @@ pub mod kube;
...
+        common_name: Option<String>,
+    }
+
+    impl CertGenerator {
+        pub fn common_name(&mut self, name: String) -> &Self {
+            self.common_name = Some(name);
+            self
+        }
+
+        pub fn generate(&self) -> Result<Vec<u8>, CertGeneratorError> {
```
dmolokanov, line 78: Finally, we have cert generator! I may steal it for the mqtt broker 😉

dmolokanov (approved):
Some nits and a couple of questions

edgelet/edgelet-http-workload/tests/dns-san.rs
darobs: This was a build error fix, similar to #3465 - edge on k8s branch is still using latest Rust compiler, although this particular file is probably a mistake.

edgelet/kube-client/src/lib.rs
darobs: There is some dead code I wound up not using for my tests. I figured it would be more valuable to leave in the CertGenerator options in.

dmolokanov (approved):
:shipit:

PR #3620, opened 2020-09-28
Metrics - check message was not sent

ancaantochi:
Fix for case when ttl is small (example set to 60 secs) dropped messages because of ttl expired metric increment even though they are delivered to clients.


varunpuranik:
Nit: can you pls add description instead of the bug link, as those are not accessible to everyone.

varunpuranik (approved):

PR #3624, opened 2020-09-28
Add metric to verify if edge device is running in some virtualized environment.

pmzara:
A new metric - Virtualized - was added to indicate if the edge device is running in some virtualized environment. We're not collecting details about the virtualized environment like product name.
This metric will be collect only for Linux machines using the command systemd-detect-virt (https://www.freedesktop.org/software/systemd/man/systemd-detect-virt.html).
The new metric was added in the model SystemInfo.
To verify the new field check the edgeagent_metadata metric.


CindyXing (approved):
LGTM. Great to see this.

edgelet/edgelet-core/src/virtualization.rs
```diff
@@ -0,0 +1,20 @@
...
+use failure::ResultExt;
+use std::process::Command;
+
+pub fn is_virtualized_env() -> Result<Option<bool>, Error> {
+    if cfg!(target_os = "linux") {
+        let status = Command::new("systemd-detect-virt")
+            .status()
+            .context(ErrorKind::GetVirtualizationStatus)?;
+
+        match status.code() {
```
lfitchett, line 13: I would have used `status.code().and_then(|s| s == 0)`, but a match works perfectly well. You could also do an `if let`.

edgelet/edgelet-core/src/virtualization.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use crate::error::{Error, ErrorKind};
+use failure::ResultExt;
+use std::process::Command;
+
+pub fn is_virtualized_env() -> Result<Option<bool>, Error> {
+    if cfg!(target_os = "linux") {
```
lfitchett, line 8: I would do this check at compile time. Make 2 functions, one with #[cfg(unix)] and one with #[cfg(windows)]. The windows only returns Ok(None), and the linux function can remove the if.

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -726,6 +726,11 @@ impl ModuleRuntime for DockerModuleRuntime {
                             version: edgelet_core::version_with_source_version(),
                             provisioning,
                             cpus: system_info.NCPU().unwrap_or_default(),
+                            virtualized: match edgelet_core::is_virtualized_env() {
```
lfitchett, line 729: Generally it would be better to use an enum here, rather than Some<bool>. That way more options can be added in the future and more information can be sent if needed, since enums can contain other values. This id fine for now, but may need to change in the future.

lfitchett (changes requested):
A few minor comments

edgelet/edgelet-core/src/virtualization.rs
pmzara: I understood your point but we are not planning to run this function on Windows OS.

edgelet/edgelet-docker/src/runtime.rs
pmzara: Okay.

lfitchett (approved):

PR #3625, opened 2020-09-28
Using self-host Windows Agent for E2E test framework

yophilav:
When the end-to-end tests run on Windows hosted agents, all tests with modules that connect to edge hub fail. Tests that are only concerned with provisioning iotedged, or that use leaf devices to talk to edge hub, pass. The module-based tests all fail with:
```
Microsoft.Azure.Devices.Client.Exceptions.IotHubCommunicationException: Transient network error occurred, please retry.
 ---> System.Net.Sockets.SocketException (11001): No such host is known.
```
It has been difficult to figure out what's happening on the hosted agents since we can't debug them. We know the problem doesn't happen on custom Windows agents, so this change switches the pipeline to custom agents.

### DIFF
--- a/builds/e2e/templates/e2e-clear-docker-cached-images.yaml
+++ b/builds/e2e/templates/e2e-clear-docker-cached-images.yaml
@@ -19,16 +19,31 @@ steps:
         | where { $_.Name -match 'Image$' } `
         | foreach { $_.Value }
 
-    # Pull required images
-    $images | foreach { sudo --preserve-env docker pull $_ }
+    if ($IsWindows)
+    {
+      # set docker command prefix
+      $dockerCommand="docker -H npipe:////./pipe/iotedge_moby_engine"
+
+      # Pull required images
+      $images | foreach { Invoke-Expression ( $dockerCommand + " pull $_" ) }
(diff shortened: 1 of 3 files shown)


builds/e2e/templates/e2e-clear-docker-cached-images.yaml
```diff
@@ -19,16 +19,31 @@ steps:
...
+      $images | foreach { sudo --preserve-env docker pull $_ }
+    }
 
     # Remove old images
-    $remove = sudo docker images --format '{{.Repository}}:{{.Tag}}' `
+    $remove = Invoke-Expression ($dockerCommand + " images --format '{{.Repository}}:{{.Tag}}'") `
         | where { $images -notcontains $_ }
-    sudo docker rm -f $(docker ps -a -q)
-    $remove | foreach { sudo docker rmi $_ }
+    Invoke-Expression ($dockerCommand + " rm -f $(docker ps -a -q)")
```
damonbarry, line 42: The inner `docker` command also needs to be replaced with the variable.

damonbarry (approved):

PR #3627, opened 2020-09-28
Enable connectivity test in nested edge environment

philipktlin:
Create a new pipeline to run connectivity test in nested edge environment.


philipktlin:
tested and deployment successfully went though.

PR #3637, opened 2020-09-30
Update CredentialsCache to handle OnBehalfof credentials

richma-ms:
This change updates the CredentialsCache and PersistedCredentialsCache so that we maintain a correct mapping of OnBehalfOf token credentials.

For OnBehalfOf connections, everything we receive on the CbsNode points to the acting child EdgeHub as the owner of the credentials.  But adding just this to the cache would be incorrect, as the individual links are opened OnBehalfOf a descendent.  The auth-chain that we receive when these links open contains the actual auth target, so the cache needs remember map the auth-target to the token that we received earlier for the actor EdgeHub.  This way when the reauthenticator iterates over the existing OnBehalfOf connections, the credentials cache can surface up the actor's token for validation.

Additionally, when the actor EdgeHub token is updated, the credentials cache now needs to go update all the corresponding entries that the EdgeHub is acting OnBehalfOf.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/NestedCredentialCache.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/NestedCredentialCache.cs
@@ -1,0 +1,114 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Core
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Identity;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Azure.Devices.Edge.Util.Concurrency;
+
+    public class NestedCredentialsCache : ICredentialsCache
+    {
+        // Maps client identity -> credential used for auth
+        readonly IDictionary<string, IClientCredentials> credentials = new Dictionary<string, IClientCredentials>();
+
+        // Maps OnBehalfOf actor device -> children, e.g. edge1 -> [leaf1, edge1/$edgeAgent, ...]
+        readonly IDictionary<string, HashSet<string>> onBehalfOfRelations = new Dictionary<string, HashSet<string>>();
+
+        readonly ICredentialsCache underlyingCache;
+        readonly AsyncLock cacheLock = new AsyncLock();
+
+        public NestedCredentialsCache(ICredentialsCache underlyingCache)
+        {
+            this.underlyingCache = Preconditions.CheckNotNull(underlyingCache, nameof(underlyingCache));
+        }
+
+        public async Task Add(IClientCredentials clientCredentials)
+        {
+            using (await this.cacheLock.LockAsync())
+            {
+                // Update the in-memory relations
+                this.UpdateCachedCredentials(clientCredentials);
+
+                // Update the DB
+                await this.underlyingCache.Add(clientCredentials);
+            }
+        }
+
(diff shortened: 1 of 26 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/NestedCredentialCache.cs
```diff
@@ -0,0 +1,114 @@
...
+
+                return Option.Some(clientCredentials);
+            }
+        }
+
+        void UpdateCachedCredentials(IClientCredentials clientCredentials)
+        {
+            string targetId = AuthChainHelpers.GetAuthTarget(clientCredentials.AuthChain).GetOrElse(clientCredentials.Identity.Id);
+
+            if (clientCredentials.AuthChain.HasValue)
```
varunpuranik, line 59: better way to do this is clientCredentials.AuthChain.Map().OrElse(), or clientCredentials.AuthChain.Match()

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/NestedCredentialCache.cs
```diff
@@ -0,0 +1,114 @@
...
+                        this.credentials[client] = updatedCredentials;
+                    }
+                }
+            }
+
+            // Update the client credentials
+            this.credentials[targetId] = clientCredentials;
+        }
+
+        public static IClientCredentials GetCredentialsWithOnBehalfOfUpdates(IClientCredentials originalClientCreds, IClientCredentials updatedActorCreds)
```
varunpuranik, line 96: Make it private (or internal if used in tests)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/PersistedTokenCredentialsCache.cs
```diff
@@ -22,10 +23,13 @@ public async Task Add(IClientCredentials clientCredentials)
...
             if (clientCredentials is ITokenCredentials tokenCredentials)
             {
+                string targetId = AuthChainHelpers.GetAuthTarget(tokenCredentials.AuthChain).GetOrElse(tokenCredentials.Identity.Id);
+
                 try
                 {
-                    var tokenCredentialsData = new TokenCredentialsData(tokenCredentials.Token, tokenCredentials.IsUpdatable);
-                    await this.encryptedStore.Put(tokenCredentials.Identity.Id, tokenCredentialsData.ToJson());
+                    var tokenCredentialsData = new TokenCredentialsData(tokenCredentials.Token, tokenCredentials.IsUpdatable, tokenCredentials.AuthChain);
+                    string tokenCredentialsString = JsonConvert.SerializeObject(tokenCredentialsData);
```
varunpuranik, line 31: Why does toJson not work here?

varunpuranik (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/PersistedTokenCredentialsCache.cs
richma-ms: there's a definition ambiguity where ToJson() is declare in Newtonsoft.Json and Edge.Util.Json, this is just my work around for it.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/NestedCredentialCache.cs
richma-ms: both approach requires a return value in the lambdas, which doesn't really apply to either if/else block.  If I make them return some arbitrary type then it prob looks more awkward =(

PR #3640, opened 2020-10-01
Fix rule ordering bug

vadim-kovalyov:
Scenario:
- Have a policy with a custom resource matcher
- Have conflicting rules for resources that
  are different, but both will match according to
  custom resource matcher.
- Result: matches first resource alphabetically.
- Expected: match first resource by order of appearance in json.


dmolokanov (approved):
LGTM! except for one param that seems off

dmolokanov (approved):
:shipit:

PR #3641, opened 2020-10-01
[k8s] Update Helm charts to allow for correct proxy settings.

darobs:
Kubernetes on premises behind a proxy needs more configuration to work correctly.  Specifically, we need to provide the proxy for external connections, but we also need to provide a "NO_PROXY" setting as well.  This also revealed a problem in our Helm charts for edgeAgent, we were only setting environment variables for known settings (i.e. the user couldn't set "no_proxy" for themselves).  This PR also allows the user to be more flexible and define their own environment variables.

Changes:
- Create a "no_proxy" in iotedged data
- Add "no_proxy" to edgeAgent environment
- Rework _helpers.tpl to allow other environment vars to be set for edgeAgent along with the ones we recognize.


darobs:
@zasherif

ggjjj (approved):
:thumbsup:

kubernetes/charts/edge-kubernetes/templates/iotedged-deployment.yaml
```diff
@@ -32,7 +32,11 @@ spec:
           {{- if .Values.iotedged.data.httpsProxy }}
           - name: "https_proxy"
             value: {{ .Values.iotedged.data.httpsProxy | quote}}
-          {{end}}
+          {{- end}}
+          {{- if .Values.iotedged.data.noProxy }}
+          - name: "no_proxy"
+            value: {{ .Values.iotedged.data.noProxy | quote}}
+          {{- end}}
```
zasherif, line 39: Do we need to add HTTP_PROXY as well so we cover all proxy env variables?

zasherif (commented):
One comment on HTTP_PROXY. Ow, lgtm.

kubernetes/charts/edge-kubernetes/templates/iotedged-deployment.yaml
darobs: Not for edgeAgent or iotedged - they only make https requests.  

Reference: https://docs.microsoft.com/en-us/azure/iot-edge/how-to-configure-proxy-support#configure-the-iot-edge-agent

zasherif (approved):

kubernetes/charts/edge-kubernetes/values.yaml
```diff
@@ -167,6 +167,9 @@ edgeAgent:
     # false dy default.
     # runAsNonRoot: false
 
+    # If an environment variables is required and not defined above, they may be added as needed.
+    # ANOTHER_ENV_VAR: "another env value"
```
CindyXing, line 171: Does this mean we allow customer to define their own environment variable?

kubernetes/charts/edge-kubernetes/templates/_helpers.tpl
```diff
@@ -60,42 +60,39 @@ agent:
...
+    {{- else if eq $envkey "storageClassName"}}
+    StorageClassName: {{- if (eq "-" $envval) }} "" {{- else }} {{ $envval | quote }} {{- end }}
+    {{- else if eq $envkey "enableExperimentalFeatures" }}
+    ExperimentalFeatures__Enabled: {{ $envval | quote }}
+    {{- else if eq $envkey "enableK8sExtensions" }}
+    ExperimentalFeatures__EnableK8SExtensions: {{ $envval | quote }}
+    {{- else if eq $envkey "runAsNonRoot" }}
+    RunAsNonRoot: {{ $envval | quote }}
+    {{- else }}
+    {{ $envkey }}: {{$envval | quote }}
```
CindyXing, line 94: What could be the environment variable?

darobs: I gave the example of "CoolOffTimeUnitInSeconds", but another would be "MaxRestartCount". These are app settings not commonly set by the user.

If we had this behavior, we could have let the customer set "NO_PROXY" and would be easier than the current work-around.

kubernetes/charts/edge-kubernetes/values.yaml
darobs: Yes. We have added parameters for all known useful values, but there may be something we have missed, or some variable we hadn't considered the user wanting to set, like "CoolOffTimeUnitInSeconds" which can be modified, but rarely used.

kubernetes/charts/edge-kubernetes/templates/_helpers.tpl
darobs: To us the example, it would be set like:
```yaml
  env:
    # [...other values...]
    ANOTHER_ENV_VAR: "another env value"
```

CindyXing (approved):

PR #3650, opened 2020-10-02
New checks

huguesBouvier:
1. Added 1 check: When in nested configuration pool.ntp.org:123 might not be available and IoTedge will connect to a parent IoTedge and not to IoT Hub.
The time is the checked directly against the parent IoT edge. The check is performed by making an HTTPS and the reading the "date" header from the response.

2. Improved parent hostname check in container. Adding extrahosts from agent to diagnostic image when performing parent hostname resolution

3. Changed API-config template so API proxy can route all types of blob request (instead of just append blob type).

4. Fixed a bug in API proxy script to build the image

Test:
1. Checked without parent hostname and checked all test pass (no regression check)
2. Checked local time check by displaying the offset in milliseconds first.
3. Check add extrahost by hard coding "dummy" has the parent hostname in the code adding extahost=dummy:"parent ip adress" and checked it worked. Then changed the IP with a dummy IP and checked it failed.


edgelet/docker-rs/src/models/host_config.rs
```diff
@@ -184,8 +184,8 @@ pub struct HostConfig {
     // #[serde(rename = "DnsSearch", skip_serializing_if = "Option::is_none")]
     // dns_search: Option<Vec<String>>,
     // /// A list of hostnames/IP mappings to add to the container's `/etc/hosts` file. Specified in the form `[\"hostname:IP\"]`.
-    // #[serde(rename = "ExtraHosts", skip_serializing_if = "Option::is_none")]
-    // extra_hosts: Option<Vec<String>>,
+    #[serde(rename = "ExtraHosts", skip_serializing_if = "Option::is_none")]
```
varunpuranik, line 187: What scenario does this enable? I believe specifying extrahosts in createOptions was already working?

edgelet/iotedge/src/check/checks/container_resolve_parent_hostname.rs
```diff
@@ -49,27 +55,33 @@ impl ContainerResolveParentHostname {
...
+            .agent()
+            .config()
+            .create_options()
+            .host_config()
+            .and_then(docker::models::HostConfig::extra_hosts)
+            .iter()
+            .for_each(|extra_hosts| {
+                extra_hosts
+                    .iter()
+                    .for_each(|host| args.push(format!("--add-host={}", host)))
```
varunpuranik, line 70: I suppose this is why you need to uncomment extraHosts above?

edgelet/iotedge/src/check/checks/pull_agent_from_upstream.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[derive(Default, serde_derive::Serialize)]
+pub(crate) struct PullAgentFromUpstream {}
+
+impl Checker for PullAgentFromUpstream {
+    fn id(&self) -> &'static str {
+        "Pull-Agent-from-upstream"
+    }
+    fn description(&self) -> &'static str {
+        "config.yaml edge Agent can be pulled from upstream"
```
varunpuranik, line 13:
```suggestion
        "EdgeAgent module can be pulled from upstream"
```

varunpuranik:
Nit: can you please add a more descriptive title and description - what checks you are adding, and what is the approach.

edgelet/docker-rs/src/models/host_config.rs
huguesBouvier: The scenario is the following:
We read the extra host give to the agent
We run the diagnostic module with the "extrahost" parameter and we resolve the hostname.

To be able to read extra host we need to enable this.

edgelet/iotedge/src/check/checks/container_resolve_parent_hostname.rs
huguesBouvier: Correct :)

huguesBouvier:
> it: can you please add a more descriptive title and description - what checks you are addi

Yes, the code is completed. I still need to do some stuff:
1. better description
2. more complete testing
3. rebasing

lfitchett (approved):

PR #3651, opened 2020-10-02
[mqtt] Sidecar startup

and-rewsmith:
The sidecars currently do not start up separately from the broker, which is not how we want to structure the startup order. This PR fixes this and addresses this TODO:

> // TODO: We need to elevate failable init steps out of the async block running in it's own thread.	
//       This is because we need to have defined setup steps, that once complete, the sidecars can run without failing.	
//       Once this is done, complex startup order will be easier to implement.	
//       We can also stop the broker process if sidecars fail to init.	
//	
//       This change depends on restructuring the startup logic so that:	
//         1. sidecars and server run at same time	
//         2. if one fails the other shuts down


mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -16,24 +17,42 @@ impl BridgeController {
...
             info!("No upstream settings detected. Not starting bridge.")
         };
         Ok(())
     }
+
+    pub async fn run(self) -> () {
+        info!("starting bridge...");
+
+        // TODO REVIEW: Bridge will need to update subscriptions
+        //              assuming this can be solved by some sync entity we can pass into the bridge and then update
```
and-rewsmith, line 41: See comment

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -16,24 +17,42 @@ impl BridgeController {
...
+        let mut bridge_handles = vec![];
+        for (_, bridge) in self.bridges {
+            let bridge_handle = tokio::spawn(async move { bridge.start().await });
+            bridge_handles.push(bridge_handle);
+        }
+
+        // TODO REVIEW: Once we implement bridge shutdown we can switch to select all
+        //              If one bridge fails then we can shutdown all bridges
```
and-rewsmith, line 49: See comment

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -16,24 +17,42 @@ impl BridgeController {
...
+        let bridges_status = join_all(bridge_handles).await;
+        for status in bridges_status {
+            if let Err(e) = status {
+                // TODO REVIEW: how do we give context to this error message?
```
and-rewsmith, line 53: See comment

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -5,45 +5,107 @@ mod snapshot;
...
+    let shutdown_signal = shutdown::shutdown();
+
+    // start broker
+    let server_join_handle =
+        tokio::spawn(bootstrap::start_server(settings, broker, shutdown_signal));
+
+    // start sidecars
+    // TODO REVIEW: it is fine to blow up here
+    //              a failure here would indicate either bridge or command handler init failed
+    //              broker doesn't open ports until these are both initialized
```
and-rewsmith, line 47: See comment

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -5,45 +5,107 @@ mod snapshot;
...
+    let state;
+    if let Some((sidecars_shutdown, sidecar_join_handles)) =
+        bootstrap::start_sidecars(broker_handle.clone(), listener_settings).await?
+    {
+        // combine future for all sidecars
+        // wait on future for sidecars or broker
+        // if one of them exits then shut the other down
+        // TODO REVIEW: we are only blowing up if either:
+        //              1 - we can't stop the server and can't get the needed state
+        //              2 - we can't signal server or sidecars to shutdown
```
and-rewsmith, line 57: See comment

mqtt/mqtt-bridge/src/controller.rs
and-rewsmith: Each bridge should start in its own thread for perf reasons.

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -57,52 +119,6 @@ where
...
-    loop {
-        interval.tick().await;
-        if let Err(e) = broker_handle.send(Message::System(SystemEvent::StateSnapshot(
-            snapshot_handle.clone(),
-        ))) {
-            warn!(message = "failed to tick the snapshotter", error=%e);
-        }
-    }
-}
-
```
and-rewsmith, line 105: I moved this logic into broker/snapshot.rs

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -5,45 +5,107 @@ mod snapshot;
...
     fs::create_dir_all(state_dir.clone())?;
     let mut persistor = FilePersistor::new(state_dir, VersionedFileFormat::default());
     let state = persistor.load().await?;
     info!("state loaded.");
 
-    let broker = bootstrap::broker(config.broker(), state).await?;
+    let broker = bootstrap::broker(settings.broker(), state).await?;
+    let mut broker_handle = broker.handle();
 
-    info!("starting snapshotter...");
```
and-rewsmith, line 37: I moved this and similar logging inside the function impls.

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -5,45 +5,107 @@ mod snapshot;
...
+        //              2 - we can't signal server or sidecars to shutdown
+        let sidecars_fut = select_all(sidecar_join_handles);
+        state = match select(server_join_handle, sidecars_fut).await {
+            // server finished first
+            Either::Left((server_output, sidecar_join_handles)) => {
+                // extract state from finished server
+                let state = server_output??;
+
+                // shutdown sidecars
+                sidecars_shutdown.shutdown().await?;
```
dmolokanov, line 66: Just as an idea: will it improve the code if we group together code, that calls shutdown on each sidecar and waits until it finishes? I mean `fn start_sidecars` returns a struct that holds all joins and a shoutdown handles for each sidecar, and exposes `fn shutdown` method that calls shutdown for each sidecar and also joins all of them.

dmolokanov (changes requested):
I think we need to add some kind of tests for shutting down sidecars to verify our ideas. Is it possible use mocking to emulate completion/failure?

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -143,6 +123,7 @@ where
     };
 
     // Prepare shutdown signal which is either SYSTEM shutdown signal or cert renewal timout
+    pin_mut!(shutdown_signal);
```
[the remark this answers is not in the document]
and-rewsmith: This was the original error:
```
`shutdown_signal` does not live long enough
borrowed value does not live long enoughrustc(E0597)mod.rs(101, 1): `shutdown_signal` dropped here while still borrowed
mod.rs(53, 5): argument requires that `shutdown_signal` is borrowed for `'static`
stack_pin.rs(22, 60): Error originated from macro here
```

You suggested:
> [9/21 1:20 PM] Denis Molokanov
    Just move pining inside async block that is spawned

mqtt/mqttd/src/broker/bootstrap/mod.rs
```diff
@@ -1,11 +1,49 @@
...
+            if let Err(e) = handle.await {
+                error!(message = "failed waiting for sidecar shutdown", err = %e);
+            }
+        }
+    }
+
+    pub fn shutdown_handle(&self) -> SidecarShutdownHandle {
+        self.shutdown_handle.clone()
+    }
+}
```
and-rewsmith, line 49: I should write a simple test

mqtt/mqttd/src/broker/mod.rs
and-rewsmith: I made a sidecar manager struct to do this

mqtt/mqttd/src/broker/bootstrap/mod.rs
and-rewsmith: Although, since this struct is specifically tailored to specific sidecars that need to interact with a broker (i.e. bridge and command handler), simple unit tests won't work. I'm exploring integration tests

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -152,23 +134,69 @@ where
...
+            command_handler_shutdown,
+        }
+    }
+
+    pub async fn shutdown(self) -> Result<(), SidecarError> {
+        self.command_handler_shutdown
+            .shutdown()
+            .await
+            .map_err(SidecarError::CommandHandlerShutdown)
+    }
```
and-rewsmith, line 166: Forgot to mention, since bridge shutdown is not implemented, the bridge will never finish once signaled to shutdown. This results in the broker process being killed. Once bridge shutdown works then broker process can stop properly.

If we don't want this bad behavior in the short term, we can simply not return the bridge join handle from the below ```start_sidecars()``` method.

and-rewsmith:
Regarding testing, I think I have an idea. We can refactor the bootstrap start method. Currently it is this:
```
pub async fn run<P>(config_path: Option<P>) -> Result<()>
```
Instead:
```
pub async fn run<P>(broker: Broker) -> Result<(BrokerSnapshot)>
```

Now we can test this function using the test broker. 
- **Verify init failure:** Set auth policies so that sidecar init will fail. Verify expected Result from ```run()``` method.
- **Verify can stop components once running:** Store the shutdown handles for broker server, bridge, command handler. Verify expected result if we shutdown the server, broker, or command handler.

This approach would explore the important branches, but wouldn't cover some things (i.e. a shutdown or join handle fails).

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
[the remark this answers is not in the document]
dmolokanov: make sense. thanks

mqtt/mqttd/src/broker/bootstrap/mod.rs
```diff
@@ -1,11 +1,57 @@
...
 #[cfg(all(not(feature = "edgehub"), feature = "generic"))]
 mod generic;
 
 #[cfg(all(not(feature = "edgehub"), feature = "generic"))]
-pub use generic::{broker, config, start_server};
+pub use generic::{
+    broker, config, start_server, start_sidecars, SidecarError, SidecarShutdownHandle,
+};
+
+pub struct SidecarManager {
```
and-rewsmith, line 23: add comment

mqtt/mqttd/src/broker/bootstrap/mod.rs
```diff
@@ -1,11 +1,57 @@
+#![allow(dead_code)] // needed because we have no sidecars for the generic feature
```
and-rewsmith, line 1: move to structure that is dead for generic case

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -5,100 +5,120 @@ mod snapshot;
...
+
+                // shutdown sidecars
+                sidecar_shutdown_handle.shutdown().await?;
+
+                // wait for sidecars to finish
+                if let Err(e) = sidecars_fut.await {
+                    error!(message = "failed running sidecars", err = %e)
+                }
+
+                state
```
and-rewsmith, line 75: make this a result and percolate up if fail

dmolokanov (changes requested):

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -41,84 +47,82 @@ where
...
     {
-        // combine future for all sidecars
         // wait on future for sidecars or broker
         // if one of them exits then shut the other down
-        // TODO REVIEW: we are only blowing up if either:
-        //              1 - we can't stop the server and can't get the needed state
-        //              2 - we can't signal server or sidecars to shutdown
-        let sidecars_fut = select_all(sidecar_join_handles);
+        let sidecar_shutdown_handle = sidecar_manager.shutdown_handle();
+        let sidecars_fut = sidecar_manager.wait_for_shutdown();
```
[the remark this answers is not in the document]
and-rewsmith: We discussed offline how we also need to expose some future for sidecars that we can select on so that if any sidecar stops, we can shut the server down. I have added comments to the SidecarManager.

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -16,24 +18,45 @@ impl BridgeController {
...
+            if let Err(e) = status {
+                error!(message = "error while running bridge", err = %e);
+            }
+        }
+
+        // TODO: bridge controller will eventually listen for updates via the twin
+        //       until this is complete we need to wait here indefinitely
+        //       if we stop the bridge controller, our startup/shutdown logic will shut eveything down
+        let century_in_secs = 60 * 60 * 24 * 365 * 100;
+        time::delay_for(Duration::from_secs(century_in_secs)).await;
```
and-rewsmith, line 58: @dmolokanov Do you agree with this approach?

mqtt/mqttd/src/broker/mod.rs
```diff
@@ -52,55 +98,15 @@ where
...
-        if let Err(e) = broker_handle.send(Message::System(SystemEvent::StateSnapshot(
-            snapshot_handle.clone(),
-        ))) {
-            warn!(message = "failed to tick the snapshotter", error=%e);
-        }
-    }
+fn collapse_snapshot_result(
+    server_output: Result<Result<BrokerSnapshot>, JoinError>,
+) -> Result<BrokerSnapshot> {
+    server_output?
```
and-rewsmith, line 109: Any recommendations for how to do this more cleanly? I wanted to use a Result function, but couldn't find a good approach.

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
and-rewsmith: This happens only if we have an upstream bridge enabled

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -49,5 +50,14 @@ impl BridgeController {
                 error!(message = "error while running bridge", err = %e);
             }
         }
+
+        // TODO: bridge controller will eventually listen for updates via the twin
+        //       until this is complete we need to wait here indefinitely
+        //       if we stop the bridge controller, our startup/shutdown logic will shut eveything down
+        // loop necessary because of https://github.com/tokio-rs/tokio/issues/2408
+        let day_in_secs = 60 * 60 * 24;
```
dmolokanov, line 58: use pending instead
```suggestion
        futures_util::future::pending()
```

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -49,5 +50,14 @@ impl BridgeController {
                 error!(message = "error while running bridge", err = %e);
             }
         }
+
+        // TODO: bridge controller will eventually listen for updates via the twin
```
dmolokanov, line 54: I'm not sure why we need to hand controller indefinitely.  shouldn't it be stopped once all bridges finish their work?

mqtt/mqttd/src/broker/mod.rs
dmolokanov: lets discuss offline?

mqtt/mqtt-bridge/src/controller.rs
and-rewsmith: I was thinking that the bridge controller should always be listening to updates. It might need to start a remote bridge eventually so it should not stop.

and-rewsmith: Since shutdown and the twin update listening are both not implemented yet, I felt this was a good short term solution. Once shutdown is implemented it will be able to cancel the twin update listening and shut down.

dmolokanov (approved):
Good work, thanks!

PR #3652, opened 2020-10-03
api proxy doc updates

ebertrams:


edge-modules/api-proxy-module/readme.md
```diff
@@ -1,147 +1,215 @@
 # API proxy module for nested IoT Edge in layered network (ISA-95)
-When inside a layered network, IoT Edge doesn't have direct internet access. 
-The API proxy module provides a way to maintain all the services an IoT Edge provide - inside a layered network - without tunneling and by terminating the connection at each layer
 
-## Architecture
-We leverage a [nginx](http://nginx.org/) reverse proxy IoT Edge module to route messages through the layers. The diagram below illustrates this approach: 
+When inside a layered network, IoT Edge doesn't have direct internet access. The API proxy module provides a way to download container images and maintain services that require their own connectivity to the cloud like the Azure Blob Store module inside a layered network, without tunneling and by terminating connections at each layer.
```
varunpuranik, line 3: Nit: the download container images and the Blob Store example in the same sentence is a bit confusing imo. 
Suggest something like - 
The API proxy module provides a secure way for clients to communicate to multiple services over HTTPs without tunneling, and instead by terminating the connections at each layer.

ebertrams: intro updated

huguesBouvier (dismissed):

huguesBouvier (approved):

PR #3653, opened 2020-10-03
mqtt client send disconnect event

ancaantochi:
mqtt3 client to send an event when it disconnects from server.


edge-modules/api-proxy-module/rust-sdk/azure-iot-mqtt/src/device.rs
```diff
@@ -154,6 +154,8 @@ impl futures_core::Stream for Client {
 					if *reset_session {
 						match std::pin::Pin::new(&mut this.inner).poll_next(cx) {
 							std::task::Poll::Ready(Some(Ok(mqtt3::Event::NewConnection { .. }))) => (),
+							                            
+                            std::task::Poll::Ready(Some(Ok(mqtt3::Event::Disconnected(_)))) => (),
```
dmolokanov, line 158: seems like formatting issue

mqtt/mqtt3/src/client/mod.rs
```diff
@@ -810,6 +818,16 @@ impl Error {
             _ => false,
         }
     }
+
+    fn is_connection_error(&self) -> bool {
+        match self {
+            Error::DecodePacket(crate::proto::DecodeError::Io(_))
```
dmolokanov, line 824: interesting... could you please add comments on the reasoning why these types of errors considered connection error? encode/decode looks like not related to networking

vadim-kovalyov (dismissed):
Where this event will be used? I also assume that, even after Disconnect event, the client is still in a good state and will retry the connection? Also, will it fire Disconnect only failed connect attempt?

mqtt/mqtt3/src/client/mod.rs
ancaantochi: it's based on some testing and from the code, for example if there is no connection and trying to send it will return IO error with WriteZero. I added Arnav to the PR if he can confirm this is correct. Also PingTimer I am not so sure should be connection erro.

ancaantochi:
> Where this event will be used? 

I will have another PR where it is going to use Disconnect event in Bridge to send a message to the broker when is gets disconnected from upstream. 

>I also assume that, even after Disconnect event, the client is still in a good state and will retry the connection? 

If it sends disconnect the client can be reused, it will try to reconnect.

> Also, will it fire Disconnect only failed connect attempt?

I am not sure I understand the question. It fires Disconnect after it was in a connected state and if it fails to send an event or if ping fails.

edge-modules/api-proxy-module/rust-sdk/azure-iot-mqtt/src/device.rs
ancaantochi: fixed for what I added, but for the rest of the file I didn't push the changes after I run cargo fmt because it's causing a lot of file changes.

mqtt/mqtt3/src/client/mod.rs
```diff
@@ -810,6 +818,16 @@ impl Error {
...
+            | Error::EncodePacket(crate::proto::EncodeError::Io(_))
+            | Error::ServerClosedConnection
+            | Error::PingTimer(_) => true,
+            _ => false,
```
and-rewsmith, line 828: I don't think we have tests for all these cases now

and-rewsmith:
You said:
> It fires Disconnect after it was in a connected state and if it fails to send an event or if ping fails.

What if it never is able to connect initially? What is the expected behavior here?

ancaantochi:
> if it never is ab

edge-modules/api-proxy-module/rust-sdk/azure-iot-mqtt/src/module.rs
```diff
@@ -277,7 +279,9 @@ impl futures_core::Stream for Client {
 						},
 
 						// Don't expect any subscription updates at this point
-						std::task::Poll::Ready(Some(Ok(mqtt3::Event::SubscriptionUpdates(_)))) => unreachable!(),
+                        std::task::Poll::Ready(Some(Ok(mqtt3::Event::SubscriptionUpdates(_)))) => unreachable!(),
```
huguesBouvier, line 282: cargo fmt

huguesBouvier (approved):

edge-modules/api-proxy-module/rust-sdk/azure-iot-mqtt/src/module.rs
ancaantochi: it completely changes the files and I wanted to keep the changes I made easy to be seen.

mqtt/mqtt3/src/client/mod.rs
arsing: It's an annoying constraint of `tokio_util::codec` that the codec errors have to be `From<std::io::Error>` even though the codec code itself doesn't do any I/O, because tokio uses the same type to bubble up errors from writing the encoded bytes to the stream / reading the bytes to be decoded from the stream through the `Framed`.

I had it in the back of my mind to separate the codec from the network for this reason. Basically stop using `tokio_util::codec` and instead encode/decode manually. It means the client has to do the job of the `Framed` wrt handling partial reads and writes, but it at least allows it to differentiate between I/O and encoding errors.

mqtt/mqtt3/src/client/mod.rs
```diff
@@ -337,6 +337,12 @@ where
                                 }
 
                                 connect.reconnect();
+
+                                if err.is_connection_error() {
+                                    return std::task::Poll::Ready(Some(Ok(Event::Disconnected(
+                                        format!("Connection failed {}", err),
```
arsing, line 343: Why not return the `EncodeError` itself instead of a `String` ?

ancaantochi: added ConnectionError

mqtt/mqtt3/src/client/mod.rs
```diff
@@ -892,6 +909,39 @@ impl std::fmt::Display for UnexpectedSubUnsubAckReason {
...
+    }
+}
+
+impl PartialEq for ConnectionError {
+    fn eq(&self, other: &Self) -> bool {
+        self.to_string() == other.to_string()
+    }
+}
+
+impl Eq for ConnectionError {}
```
dmolokanov, line 933: Do we really need this one?

dmolokanov (approved):
LGTM

mqtt/mqtt3/src/client/mod.rs
ancaantochi: it's needed because Event implements Eq

mqtt/mqtt3/src/client/mod.rs
ancaantochi: DecodePacket is not covered in tests, didn't find a way to test it. PingTimer was removed as it seems it is not used.

dmolokanov (approved):
:shipit:

PR #3659, opened 2020-10-05
Default auth mode to Scope for nested Edge

richma-ms:
On startup, we detect whether the upstream is another Edge, and if it is we default to local authentication only for child device/modules.


huguesBouvier (approved):

richma-ms:
az restart

PR #3660, opened 2020-10-06
[mqtt] Add default IoTHub policy auth test for subscription re-evaluation

dylanbronson:
Adding integration test for the re-evaluation flow for Default IoTHub policies. The flow goes like this:
1. Start broker
2. Start CommandHandler with AuthorizedIdentities command
3. Start up edgeHub client
4. EdgeHub client sends AuthorizedIdentities to broker, one of which is named "test_device" client
5. Broker caches authorized identities
6. Connect test_device client 
7. Subscribe to some topic for test_device
8. Re-send the authorized identities from edgeHub as empty list
9. Broker should automatically re-evaluate all subscriptions and disonnect client
10. Assert client is disconnected

As a part of this change, I also moved test/Command.rs code to a common util file that test/command.rs and test/authorization.rs both use.


mqtt/mqtt-edgehub/tests/authorization.rs
```diff
@@ -0,0 +1,105 @@
...
+        )))
+        .build();
+    let broker_handle = broker.handle();
+
+    let server_handle = start_server(broker, DummyAuthenticator::anonymous());
+
+    // start command handler with AuthorizedIdentitiesCommand
+    let command = AuthorizedIdentitiesCommand::new(&broker_handle);
+    let (command_handler_shutdown_handle, join_handle) =
+        common::start_command_handler(common::TEST_SERVER_ADDRESS.to_string(), command)
```
vadim-kovalyov, line 41: I know you used disconnect test as an example, and I think there is a problem here and in the disconnect test.

if you look at `mqtt_broker_test_util::server::run` you'll find a code that increment a port number on every server start. This is done to avoid multiple tests running in parallel to compete for the same port number. Using hardcoded `TEST_SERVER_ADDRESS` will lead to unexpected behaviour.

Quick fix can be like this, since we don't care much about public/private transports in tests. @dmolokanov, let me know if you have better suggestions.
```suggestion
        common::start_command_handler(server_handle.address(), command)
```

mqtt/mqtt-edgehub/tests/common/mod.rs
```diff
@@ -0,0 +1,66 @@
+use std::error::Error as StdError;
+
+use tokio::task::JoinHandle;
+
+use mqtt3::ShutdownError;
+use mqtt_broker::auth::{Activity, Authorization, Authorizer};
+use mqtt_edgehub::command::{Command, CommandHandler, ShutdownHandle};
+
+pub const DISCONNECT_TOPIC: &str = "$edgehub/disconnect";
```
vadim-kovalyov, line 9: I would rather make constants public in prod code and use them in tests.

mqtt/mqtt-edgehub/tests/authorization.rs
```diff
@@ -0,0 +1,105 @@
...
+            .await
+            .expect("could not start command handler");
+
+    let mut edgehub_client = TestClientBuilder::new(server_handle.address())
+        .with_client_id(ClientId::IdWithCleanSession("$edgehub".into()))
+        .build();
+
+    let service_identity1 =
+        IdentityUpdate::new("device-1".into(), Some("device-1;$edgehub".into()));
+    let identities: Vec<IdentityUpdate> = vec![service_identity1];
```
vadim-kovalyov, line 51: nit:
```suggestion
    let identities = vec![service_identity1];
```

mqtt/mqtt-edgehub/tests/authorization.rs
```diff
@@ -0,0 +1,105 @@
...
+        None,
+        None,
+    )
+    .await;
+    device_client.next().await; // skip connack
+                                // client subscribes to topic
+    device_client.send_subscribe(s).await;
+    device_client.next().await; // skip suback
+
+    let identities2: Vec<IdentityUpdate> = vec![];
```
vadim-kovalyov, line 82: nit: it is ok to shadow previous vars.
```suggestion
    let identities: Vec<IdentityUpdate> = vec![];
```

mqtt/mqtt-edgehub/tests/authorization.rs
```diff
@@ -0,0 +1,105 @@
...
+    edgehub_client
+        .publish_qos1(
+            common::AUTHORIZED_IDENTITIES_TOPIC,
+            serde_json::to_string(&identities2).ok().unwrap(),
+            false,
+        )
+        .await;
+
+    // next() will return None only if the client is disconnected, so this
+    // asserts that the subscription has been re-evaluated and disonnected by broker.
```
vadim-kovalyov, line 94:
```suggestion
    // asserts that the subscription has been re-evaluated and disconnected by broker.
```

mqtt/mqtt-edgehub/tests/authorization.rs
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+    let mut device_client = PacketStream::connect(
+        ClientId::IdWithCleanSession("device-1".into()),
+        server_handle.address(),
+        None,
+        None,
+    )
+    .await;
+    device_client.next().await; // skip connack
```
vadim-kovalyov, line 77: For this scenario I wouldn't skip connack and suback, b/c it is important to validate that the client has connected and subscribed. In fact, if you add assertion you will see that it actually fails to connect. We need to look into this...

```
    // assert connack
    assert_matches!(device_client.next().await, Some(Packet::ConnAck(_)));

    device_client.send_subscribe(s).await; // client subscribes to topic

    // assert suback
    assert_matches!(device_client.next().await, Some(Packet::SubAck(_)));
```

vadim-kovalyov (changes requested):
It looks like client is unable to subscribe correctly. @dylanbronson, can you please take a look? see my comments for more details.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -293,6 +293,12 @@ impl IdentityUpdate {
     pub fn auth_chain(&self) -> Option<&str> {
         self.auth_chain.as_deref()
     }
+    pub fn new(identity: String, auth_chain: Option<String>) -> Self {
```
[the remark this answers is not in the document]
dmolokanov: nit: lets put it in the first position inside impl block?

mqtt/mqtt-edgehub/tests/authorization.rs
dmolokanov: looks good

mqtt/mqtt-edgehub/tests/common/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+pub const AUTHORIZED_IDENTITIES_TOPIC: &str = "$internal/identities";
+pub const LOCAL_BROKER_SUFFIX: &str = "$edgeHub/$broker";
+pub const TEST_SERVER_ADDRESS: &str = "localhost:5555";
```
dmolokanov, line 12: this should not be used as per @vadim-kovalyov explanation

mqtt/mqtt-edgehub/tests/authorization.rs
dylanbronson: I have a hard time letting go of my explicit types

mqtt/mqtt-edgehub/tests/authorization.rs
```diff
@@ -0,0 +1,114 @@
...
+    packet_stream::PacketStream,
+    server::{start_server, DummyAuthenticator},
+};
+use mqtt_edgehub::{
+    auth::EdgeHubAuthorizer, auth::IdentityUpdate, command::AuthorizedIdentitiesCommand,
+    command::AUTHORIZED_IDENTITIES_TOPIC,
+};
+
+#[macro_use]
+extern crate assert_matches;
```
vadim-kovalyov, line 19: this is an old syntax. it should work just like this:
```rust
use assert_matches::assert_matches;
```

mqtt/mqtt-edgehub/tests/authorization.rs
```diff
@@ -0,0 +1,114 @@
...
+            AUTHORIZED_IDENTITIES_TOPIC,
+            serde_json::to_string(&identities).ok().unwrap(),
+            false,
+        )
+        .await;
+
+    let s = Subscribe {
+        packet_identifier: PacketIdentifier::new(1).unwrap(),
+        subscribe_to: vec![SubscribeTo {
+            topic_filter: "$edgehub/device-1/inputs/telemetry/#".into(), // "devices/device-1/inputs/telemetry/#".into(),
```
vadim-kovalyov, line 71: nit: do you need that comment?

mqtt/mqtt-edgehub/tests/authorization.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    let service_identity1 =
+        IdentityUpdate::new("device-1".into(), Some("device-1;$edgehub".into()));
+    let identities = vec![service_identity1];
+
+    // EdgeHub sends authorized identities + auth chains to broker
+    edgehub_client
+        .publish_qos1(
+            AUTHORIZED_IDENTITIES_TOPIC,
+            serde_json::to_string(&identities).ok().unwrap(),
```
vadim-kovalyov, line 63: nit: here and below ok() is not needed, also its better to use .expect() w/ error message - allows to find a place in the code more quickly.
```suggestion
            serde_json::to_string(&identities).expect("unable to serialize identities"),
```

mqtt/mqtt-broker-tests-util/src/server.rs
```diff
@@ -101,6 +101,10 @@ impl DummyAuthenticator {
     pub fn anonymous() -> Self {
         Self(AuthId::Anonymous)
     }
+
+    pub fn id(id: &str) -> Self {
```
vadim-kovalyov, line 105: nit: better naming for constructor func `with_id()`, can be confused with id property getter.

mqtt/mqtt-edgehub/tests/command.rs
```diff
@@ -30,7 +26,7 @@ async fn disconnect_client() {
 
     let command = DisconnectCommand::new(&broker_handle);
     let (command_handler_shutdown_handle, join_handle) =
-        start_command_handler(TEST_SERVER_ADDRESS.to_string(), command)
+        common::start_command_handler(server_handle.address(), command)
```
vadim-kovalyov, line 29: :+1:

vadim-kovalyov (approved):
Looks great! just few nits.

vadim-kovalyov (dismissed):
🚀

mqtt/mqtt-edgehub/tests/common/mod.rs
```diff
@@ -43,6 +40,24 @@ where
...
     }
+
+    fn update(&mut self, update: Box<dyn Any>) -> Result<(), Self::Error> {
+        self.0.update(update)
+    }
+}
+
+pub struct BottomLevelDummyAuthorizer;
+
+impl Authorizer for BottomLevelDummyAuthorizer {
```
dmolokanov, line 51: noticed you didn't use `authorize_fn_ok` in-place. just curious you didn't like it?

dmolokanov, line 51: shouldn't this guy to implement `fn update` ?

dmolokanov (approved):
some nits

mqtt/mqtt-edgehub/tests/common/mod.rs
dylanbronson: For the DummyAuthorizer, I needed to take in another authorizer, which authorize_fn_ok doesn't support. It complicated the code quite a bit to add it, so this just seemed easier.

But for the BottomLevelDummyAuthorizer I just forgot about it. So I've used it in this new edition.

dylanbronson: It doesn't need to since it's never called, so it just takes the default, but doesn't matter anymore since I've deleted it.

mqtt/mqtt-edgehub/tests/authorization.rs
```diff
@@ -0,0 +1,119 @@
...
+/// connect authorized client and subscribe
+/// publish authorization update with client removed
+/// verify client has disconnected
+#[tokio::test]
+async fn disconnect_client_on_auth_update() {
+    // Start broker with DummyAuthorizer that allows everything from CommandHandler and $edgeHub,
+    // but otherwise passes authorization along to EdgeHubAuthorizer
+    let broker = BrokerBuilder::default()
+        .with_authorizer(DummyAuthorizer::new(EdgeHubAuthorizer::new(
+            authorize_fn_ok(|activity| {
```
vadim-kovalyov, line 35: nice! :+1:

vadim-kovalyov (approved):

mqtt/mqtt-edgehub/tests/common/mod.rs
dmolokanov: ah, I missed the default implementation

PR #3662, opened 2020-10-06
Fixing bug preventing connection to parent when a capital letter was in VM name

huguesBouvier:
Fixing bug preventing connection to parent when a capital letter was in VM name


lfitchett (approved):
Ok, since its limited to parent hostname only. I'd check with other people if you need to change the iothub hostname as well

PR #3665, opened 2020-10-06
Remove edgehub_queue_length metric

yophilav:
Currently `edgehub_queue_length` is providing a misleading metric value. 
An immediate "fix" is to remove the metric in release 1.0.10.


richma-ms (approved):

yophilav:
Per Shiproom discussion on 10/7/2020:
This looks like a metrics issue, not a regression in functionality. We will ship 1.0.10 with this issue since we do not expect customers to hit this in the main line scenario. To hit this the customer must set a low TTL which can result in dropped messages. Loosing messages is not something most customers want to have.

Closing this PR.

PR #3670, opened 2020-10-07
Fix quantiles for Duration and Timer to align with the quantiles for histogram

varunpuranik:
We want to use  0.1, 0.5, 0.9, 0.99 quantiles for all our metrics. This was updated for Histogram metric in this commit https://github.com/Azure/iotedge/commit/f85b8389f2c6366a97476c13b336afe7050a79f8 but was not updated for Duration/Timer. This PR takes care of that.


lfitchett (approved):
Ah, missed those

PR #3672, opened 2020-10-07
API proxy doc update

huguesBouvier:
- Streamlined API proxy default template: Removed NGINX_HAS_ROOT
- Completed doc


edge-modules/api-proxy-module/readme.md
```diff
@@ -205,8 +203,75 @@ Lastly, all the module image URIs in **any lower layer** should use the domain n
...
+
+>If you want to upload support bundle, currently, the only supported configuration is when the blob storage module is at the top level. To be able to upload a support bundle, a blob container needs to be created and as of now that is only possible when the blob storage module is at the top level.
+
+The configuration of the API Proxy module at the **top layer** for this scenario is the following
+
+- Environment variables:
+
+    | Name  | Value |
+    | ------------- |  ------------- |
+    |BLOB_UPLOAD_ROUTE_ADDRESS|module_blob:11002|
```
ebertrams, line 217: is module_blob the name of the blob module? If yes, it may be worth making that explicit. Also, by default the name of the Blob module will be "AzureBlobStorageonIoTEdge" when deployed form the marketplace, so i d recommend switching to this default.

richma-ms (approved):

PR #3683, opened 2020-10-08
Workload api can take hostname with capital letter

huguesBouvier:
Fixing this bug:
https://msazure.visualstudio.com/One/_workitems/edit/7993355

Also fixed a bug for nested configuration: Connectivity tests were not run against parent.

Test:
- Created nested network (2 levels) with only IP in "hostname" and "parent_hostname" field in config.yaml.
- ran iotedge check on both lowest level and root(root is not a nested device)


edgelet/edgelet-utils/src/lib.rs
```diff
@@ -90,7 +90,9 @@ pub fn append_dns_san_entries(sans: &str, names: &[&str]) -> String {
     let mut dns_sans = names
         .iter()
         .filter_map(|name| {
-            if name.trim().is_empty() {
+            if IpAddr::from_str(name).is_ok() {
```
varunpuranik, line 93: Test?

varunpuranik (dismissed):

edgelet/edgelet-utils/src/lib.rs
huguesBouvier: added

varunpuranik (approved):

edgelet/edgelet-utils/src/lib.rs
```diff
@@ -90,7 +87,9 @@ pub fn append_dns_san_entries(sans: &str, names: &[&str]) -> String {
     let mut dns_sans = names
```
ancaantochi, line 87: nit: this should be renamed as it can be dns and ips

edgelet/iotedge/src/check/checks/container_connect_upstream.rs
```diff
@@ -105,12 +105,17 @@ impl ContainerConnectUpstream {
             return Ok(CheckResult::Skipped);
         };
 
-        let upstream_hostname = if let Some(iothub_hostname) = &check.iothub_hostname {
+        let parent_hostname: String;
```
ancaantochi, line 108: is this needed, can it just return &upstream_hostname.to_string() ?

edgelet/edgelet-utils/src/lib.rs
huguesBouvier: Fixed

edgelet/iotedge/src/check/checks/container_connect_upstream.rs
huguesBouvier: I tried but I get: "creates a temporary which is freed while still in use"

ancaantochi (approved):

ancaantochi (approved):

PR #3685, opened 2020-10-08
Update policy definition from EdgeHub twin (1/2) - plumbing

vadim-kovalyov:
Authorization policy definition for MQTT broker would come from EdgeHub twin. In this PR I have added initial plumbing for authorization policy object into `EdgeHubConfig` and `EdgeHubDesiredProperties` (twin) and also created `PolicyUpdateHandler` that will listen for twin updates and will publish a system message to MQTT broker with new policy definition.

TODO:
- [x] Existing tests compiled
- [x] Existing tests are fixed
- [x] New tests added

This is 1st out of 2 PRs for this feature. Validation of policy definition will come in the second PR.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -37,7 +42,9 @@ public async Task Init(IConfigSource configProvider)
             Preconditions.CheckNotNull(configProvider, nameof(configProvider));
             try
             {
-                configProvider.SetConfigUpdatedCallback(this.HandleUpdateConfig);
+                configProvider.ConfigUpdates +=
```
vadim-kovalyov, line 45: the downside of Event vs Callback is that async event will be "fire and forget". Do you see any issues with that?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubConfigParser.cs
```diff
@@ -16,24 +19,44 @@ public static EdgeHubConfig GetEdgeHubConfig(EdgeHubDesiredProperties desiredPro
...
 
-            return new EdgeHubConfig(desiredProperties.SchemaVersion, new ReadOnlyDictionary<string, RouteConfig>(routes), desiredProperties.StoreAndForwardConfiguration);
+            return new ReadOnlyDictionary<string, RouteConfig>(routes);
+        }
+
+        private static void ValidateAuthorizationPolicy(AuthorizationConfiguration authorizations)
+        {
+            if (authorizations != null)
+            {
+                throw new NotImplementedException();
```
vadim-kovalyov, line 58: this will be fixed in the next PR

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
```diff
@@ -0,0 +1,99 @@
...
+        async Task ConfigUpdateHandler(EdgeHubConfig config)
+        {
+            PolicyUpdate update = ConfigToPolicyUpdate(config);
+            await this.PublishPolicyUpdate(update);
+        }
+
+        static PolicyUpdate ConfigToPolicyUpdate(EdgeHubConfig config)
+        {
+            // TODO
+            throw new NotImplementedException();
```
vadim-kovalyov, line 60: this will be fixed in the next PR

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
```diff
@@ -1,19 +1,29 @@
...
+    /// </summary>
+    public class EdgeHubDesiredProperties
+    {
         [JsonConstructor]
-        public EdgeHubDesiredProperties(string schemaVersion, IDictionary<string, RouteConfiguration> routes, StoreAndForwardConfiguration storeAndForwardConfiguration)
+        public EdgeHubDesiredProperties(
+            string schemaVersion,
+            IDictionary<string, RouteConfiguration> routes,
+            StoreAndForwardConfiguration storeAndForwardConfiguration,
+            AuthorizationConfiguration authorizations)
```
ancaantochi, line 20: This should be mqttBroker configuration which contains authorizations and bridges (bridges will added later)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
```diff
@@ -1,19 +1,29 @@
...
         {
             this.SchemaVersion = Preconditions.CheckNonWhiteSpace(schemaVersion, nameof(schemaVersion));
             this.Routes = Preconditions.CheckNotNull(routes, nameof(routes));
             this.StoreAndForwardConfiguration = Preconditions.CheckNotNull(storeAndForwardConfiguration, nameof(storeAndForwardConfiguration));
+            // can be null for previous versions.
+            this.Authorizations = authorizations;
```
ancaantochi, line 26: I think set it to empty list if it's empty so we don't deal with null

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/AuthorizationConfiguration.cs
```diff
@@ -0,0 +1,144 @@
...
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Domain object that represents Authorization configuration for Edge Hub Module (MQTT Broker).
+    ///
+    /// This object is being eventually constructed from the EdgeHub twin's desired properties.
+    /// See <see cref="EdgeHubDesiredProperties"/> for DTO.
+    /// </summary>
+    public class AuthorizationConfiguration : List<Statement>, IEquatable<AuthorizationConfiguration>
```
ancaantochi, line 16: I would add a list of statements instead of extending List<Statement>

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
```diff
@@ -28,6 +28,6 @@ public class Constants
         public const string ServiceApiIdHeaderKey = "x-ms-edge-moduleId";
         public const string OriginEdgeHeaderKey = "x-ms-edge-origin";
 
-        public static readonly Version ConfigSchemaVersion = new Version("1.1.0");
+        public static readonly Version ConfigSchemaVersion = new Version("1.2.0");
```
and-rewsmith, line 31: So this PR cannot be merged until the new twin schema is rolled out?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/IConfigSource.cs
```diff
@@ -5,12 +5,20 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Config
...
         Task<Option<EdgeHubConfig>> GetCachedConfig();
 
         Task<Option<EdgeHubConfig>> GetConfig();
 
-        void SetConfigUpdatedCallback(Func<EdgeHubConfig, Task> callback);
+        /// <summary>
+        /// Occurres when config is updated in the config source.
+        /// For example, by a twin update pushed from the cloud.
+        /// </summary>
+        event EventHandler<EdgeHubConfig> ConfigUpdated;
```
and-rewsmith, line 22: Why change this to an event?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/IMqttBrokerConnector.cs
```diff
@@ -11,6 +11,6 @@ public interface IMqttBrokerConnector
 
         Task<bool> SendAsync(string topic, byte[] payload);
 
-        public event EventHandler OnConnected;
```
and-rewsmith, line 14: This was removed because public is redundant in an interface, right?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        async Task ConfigUpdateHandler(EdgeHubConfig config)
+        {
+            PolicyUpdate update = ConfigToPolicyUpdate(config);
+            await this.PublishPolicyUpdate(update);
+        }
+
+        static PolicyUpdate ConfigToPolicyUpdate(EdgeHubConfig config)
+        {
+            // TODO: conversion here.
```
and-rewsmith, line 59: Convert from AuthorizationConfiguration to the PolicyUpdate?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/RoutingModule.cs
```diff
@@ -577,11 +580,22 @@ protected override void Load(ContainerBuilder builder)
...
+                                routeFactory);
                         }
                         else
                         {
-                            return new LocalConfigSource(routeFactory, this.routes, this.storeAndForwardConfiguration);
+                            return new LocalConfigSource(
+                                routeFactory,
+                                this.routes,
+                                this.storeAndForwardConfiguration,
+                                this.authorizationConfiguration);
```
and-rewsmith, line 598: The default behavior here is what? I looked and a new AuthorizationConfiguration is empty, so does this mean that everything is denied?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/RoutingModule.cs
```diff
@@ -577,11 +580,22 @@ protected override void Load(ContainerBuilder builder)
...
+                                this.versionInfo,
+                                twinManager,
+                                twinMessageConverter,
+                                twinCollectionMessageConverter,
+                                routeFactory);
                         }
                         else
                         {
-                            return new LocalConfigSource(routeFactory, this.routes, this.storeAndForwardConfiguration);
+                            return new LocalConfigSource(
```
and-rewsmith, line 594: Unrelated to this PR, why do we even have an option for not using twin config?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
vadim-kovalyov: Yes, this will be added in the next PR.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/IMqttBrokerConnector.cs
vadim-kovalyov: Right

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/IConfigSource.cs
vadim-kovalyov: B/c now I have multiple (two) subscribers: `ConfigUpdater` and `PolicyUpdateHandler`.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
vadim-kovalyov: Good question, I'm not sure. I thought that this is max supported version, and all previous version would still work. @varunpuranik, @ancaantochi, @richma-ms do you have any input on this?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
vadim-kovalyov: Good catch, fixing this now.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/AuthorizationConfiguration.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        public override bool Equals(object obj)
+            => this.Equals(obj as Statement);
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                int hash = 17;
+                hash = this.Identities.Aggregate(hash, (acc, item) => (acc * 31 + item.GetHashCode()));
```
philipktlin, line 91: What happens if the identities are the same, but with different order?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/AuthorizationConfiguration.cs
```diff
@@ -0,0 +1,144 @@
...
+                hash = this.Deny.Aggregate(hash, (acc, item) => (acc * 31 + item.GetHashCode()));
+                return hash;
+            }
+        }
+    }
+
+    public class Rule : IEquatable<Rule>
+    {
+        [JsonConstructor]
+        public Rule(IList<string> operations, IList<string> resources)
```
philipktlin, line 102: nit: IEnumerable can be used instead of IList, as we don't care about the count or the order.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/AuthorizationConfiguration.cs
philipktlin, line 16: Add tests

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/IConfigSource.cs
philipktlin: You can still manage multiple callback in the ConfigSource implementation.  For example, you can define same event inside the ConfigSource implementation and calling SetConfigUpdatedCallback (Maybe rename it to AddConfigUpdatedCallback) will add callback to the event.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -37,7 +42,9 @@ public async Task Init(IConfigSource configProvider)
             Preconditions.CheckNotNull(configProvider, nameof(configProvider));
             try
             {
-                configProvider.SetConfigUpdatedCallback(this.HandleUpdateConfig);
+                configProvider.ConfigUpdated +=
+                    async (sender, serviceIdentities) => await this.HandleUpdateConfig(serviceIdentities);
```
philipktlin, line 46: why serviceIdentities? please rename.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubConfig.cs
```diff
@@ -5,13 +5,24 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core.Config
...
+    /// Domain object that represents EdgeHub configuration.
+    ///
+    /// This object is being eventually constructed from the EdgeHub twin's desired properties.
+    /// See <see cref="EdgeHubDesiredProperties"/> for DTO.
+    /// </summary>
     public class EdgeHubConfig : IEquatable<EdgeHubConfig>
     {
-        public EdgeHubConfig(string schemaVersion, IReadOnlyDictionary<string, RouteConfig> routes, StoreAndForwardConfiguration storeAndForwardConfiguration)
+        public EdgeHubConfig(
+            string schemaVersion,
```
philipktlin, line 17: nit: should use System.Version type

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubConfigParser.cs
```diff
@@ -16,24 +19,45 @@ public static EdgeHubConfig GetEdgeHubConfig(EdgeHubDesiredProperties desiredPro
...
+            ValidateAuthorizationPolicy(desiredProperties.Authorizations);
+
+            return new EdgeHubConfig(
+                desiredProperties.SchemaVersion,
+                routes,
+                desiredProperties.StoreAndForwardConfiguration,
+                desiredProperties.Authorizations);
+        }
+
+        private static ReadOnlyDictionary<string, RouteConfig> ParseRoutes(EdgeHubDesiredProperties desiredProperties, RouteFactory routeFactory)
```
philipktlin, line 33: nit: remove private, which is default accessor.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/IConfigSource.cs
philipktlin: I think event approach is fine here.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/modules/RoutingModule.cs
vadim-kovalyov: Don't have answer to that ¯\_(ツ)_/¯

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/AuthorizationConfiguration.cs
vadim-kovalyov: Good point, it will lead to the identical hash codes but equality will return false. I think this is fine for our use case, unless you have better suggestions.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/AuthorizationConfiguration.cs
vadim-kovalyov: @philipktlin what test cases do you have in mind? i've added tests for deserialization and equality. There is nothing much else to test here for now.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
vadim-kovalyov: oops, copy/paste :)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubConfig.cs
vadim-kovalyov: since this is nit I'll skip that. Don't want to add more to this PR. Maybe in the next one.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/AuthorizationConfiguration.cs
vadim-kovalyov: Good point, but I'll need count later when I implement validation.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/AuthorizationConfiguration.cs
vadim-kovalyov: Extending a list allows me to define my own deep equality comparison logic without extra level of indirection (a class that wraps a list). I guess rust has a "bad" influence on me :D. I thought about that and would like to leave it as is, unless you have strong reasons not to do that.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
vadim-kovalyov: dues to difficulties with deserializing Option<T> I'll keep this as nullable but convert into Option<BrokerConfig> in EdgeHubConfigParser

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
```diff
@@ -55,17 +70,33 @@ internal void ValidateSchemaVersion()
                         throw new InvalidSchemaVersionException($"Route priority/TTL is not supported in schema {this.SchemaVersion}.");
                     }
                 }
-            }
+
+                // Authorization policies not allowed.
+                if (this.BrokerConfiguration != null)
+                {
+                    throw new InvalidSchemaVersionException($"Broker configuration is not supported in schema {this.SchemaVersion}.");
```
richma-ms, line 77: I'm removing strict schema validation as per an earlier design decision change, so you're probably going to see a conflict here as soon as you merge.
[https://github.com/Azure/iotedge/pull/3698](https://github.com/Azure/iotedge/pull/3698)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
```diff
@@ -55,17 +70,33 @@ internal void ValidateSchemaVersion()
...
             else if (version.Minor == 1)
             {
                 // 1.1.0
                 //
-                // Everything from 1.0 is supported
+                // Authorization policies not allowed.
+                if (this.BrokerConfiguration != null)
+                {
+                    throw new InvalidSchemaVersionException($"Broker configuration is not supported in schema {this.SchemaVersion}.");
+                }
```
richma-ms, line 88: this is not needed for now, see other comments.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubDesiredProperties.cs
```diff
@@ -1,19 +1,30 @@
...
+
+    /// <summary>
+    /// DTO that is used to deserialize EdgeHub Desired properties of the twin
+    /// into <see cref="EdgeHubConfig" /> by <see cref="TwinConfigSource" />.
+    /// </summary>
+    public class EdgeHubDesiredProperties
+    {
         [JsonConstructor]
-        public EdgeHubDesiredProperties(string schemaVersion, IDictionary<string, RouteConfiguration> routes, StoreAndForwardConfiguration storeAndForwardConfiguration)
+        public EdgeHubDesiredProperties(
```
richma-ms, line 17: >EdgeHubDesiredProperties [](start = 15, length = 24)

chatted offline about this, we're moving towards a model where ver 1.0/1.1/1.2 will have separate JSON classes for deserialization, so that each individual version only parse the properties that it knows about.  This is okay for now, but we'll need to make sure to come back and split out the newly added properties.

richma-ms (dismissed):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BrokerConfig.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                return true;
+            }
+
+            return this.Bridges.Equals(other.Bridges)
+                && this.Authorizations.Equals(other.Authorizations);
+        }
+
+        public override bool Equals(object obj)
+            => this.Equals(obj as AuthorizationConfig);
```
ancaantochi, line 47: should be: as BrokerConfig

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Responsible for listening for EdgeHub config updates (twin updates)
+    /// and pushing new authorization policy definition from EdgeHub config to the Mqtt Broker.
+    /// </summary>
+    public class PolicyUpdateHandler : IMessageProducer
+    {
+        // !Important: please keep in sync with mqtt-edgehub::command::POLICY_UPDATE_TOPIC
```
ancaantochi, line 19: we should have a test for this to make sure they are in sync

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        async Task ConfigUpdateHandler(EdgeHubConfig config)
+        {
+            PolicyUpdate update = ConfigToPolicyUpdate(config);
+            await this.PublishPolicyUpdate(update);
+        }
+
+        static PolicyUpdate ConfigToPolicyUpdate(EdgeHubConfig config)
+        {
+            Option<AuthorizationConfig> maybe_policy = config.BrokerConfiguration.AndThen(
```
ancaantochi, line 59: nit: rename to maybePolicy

richma-ms (dismissed):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
vadim-kovalyov: I believe there would be not much benefit of such test. I think comment is enough to notify future devs in case they want to change this constant.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -37,7 +42,9 @@ public async Task Init(IConfigSource configProvider)
             Preconditions.CheckNotNull(configProvider, nameof(configProvider));
             try
             {
-                configProvider.SetConfigUpdatedCallback(this.HandleUpdateConfig);
+                configProvider.ConfigUpdated +=
+                    async (sender, config) => await this.HandleUpdateConfig(config);
+
```
vipeller, line 47: I see some subscriptions but never unsubscribe to ConfigUpdated - do these things live till the end of the program together? (I am asking in the context of GC that it cannot clean up stuff when they referenced through event handlers)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/Constants.cs
varunpuranik: What do you mean by new twin schema is rolled out? 
It would be the other way round. Also, Vadim is right, this would be the max supported version
There is a version matching item on @richma-ms, but that may/may not come before we release this.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BrokerProperties.cs
```diff
@@ -0,0 +1,33 @@
...
+    using System.Linq;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Azure.Devices.Edge.Util.Json;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// DTO that is used to deserialize MQTT Broker Config from EdgeHub twin
+    /// into <see cref="BrokerConfig" />.
+    /// </summary>
+    public class BrokerProperties
```
varunpuranik, line 14: Why do we need this and BrokerConfig separately? The Option?

vadim-kovalyov: Yeah, I couldn't make it work with Option<T> so I split json model and domain model. I guess at some point we can fix `OptionConverter`, but this will be a separate task.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
vadim-kovalyov: Good question. `ConfigUpdater` and `PolicyUpdateHandler` never unsubscribe and should receive events throughout the lifetime of EH core. `ConfigUpdater` was already "linked" with TwinConfigSource by the latter referencing a callback from the former. So I don't think that my change adds more to it.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdate.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    using System;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// PolicyUpdate is a Data Transfer Object used for sending authorization policy
+    /// definition from EdgeHub core to Mqtt Broker.
+    /// </summary>
+    internal class PolicyUpdate
```
varunpuranik, line 12: Nit: move to PolicyUpdateHandler.cs as a nested class (since this is internal and used only in 1 class)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            this.connector = connector;
+            this.connector.OnConnected += async (sender, args) => await this.OnConnect();
+        }
+
+        async Task OnConnect()
+        {
+            var configSource = await this.configSource;
+            configSource.ConfigUpdated +=
+                async (sender, config) => await this.ConfigUpdateHandler(config);
```
varunpuranik, line 41: Wrap in try/catch

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+            return maybePolicy.Match(
+                policy => new PolicyUpdate(JsonConvert.SerializeObject(policy)),
+                () => GetEmptyPolicy());
+        }
+
+        static PolicyUpdate GetEmptyPolicy()
+        {
+            return new PolicyUpdate(@"
+            {
+                'schemaVersion': '2020-10-30',
```
varunpuranik, line 71: Does the policy section also have a version?

varunpuranik: Is it used when the section is not empty? Where is that?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Config;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Responsible for listening for EdgeHub config updates (twin updates)
+    /// and pushing new authorization policy definition from EdgeHub config to the Mqtt Broker.
+    /// </summary>
+    public class PolicyUpdateHandler : IMessageProducer
```
varunpuranik, line 17: Will you have a separate handler for bridge config? Why not combine the two?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
vadim-kovalyov: that's the question I wanted to discuss offline. Right now policy engine needs a version, but we can remove that if we don't really need it. We can decide and fix in the next PR

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
vadim-kovalyov: I thought it better to have them separate, especially on the broker side, since we update different parts of the broker. But I might combine that on EH Core side when I work on Bridge config updates.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -37,7 +42,9 @@ public async Task Init(IConfigSource configProvider)
             Preconditions.CheckNotNull(configProvider, nameof(configProvider));
             try
             {
-                configProvider.SetConfigUpdatedCallback(this.HandleUpdateConfig);
+                configProvider.ConfigUpdated +=
```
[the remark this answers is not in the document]
vadim-kovalyov: Not sure what exactly do you want to wrap?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/PolicyUpdateHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+        async Task ConfigUpdateHandler(EdgeHubConfig config)
+        {
+            PolicyUpdate update = ConfigToPolicyUpdate(config);
+            await this.PublishPolicyUpdate(update);
+        }
+
+        static PolicyUpdate ConfigToPolicyUpdate(EdgeHubConfig config)
+        {
+            Option<AuthorizationConfig> maybePolicy = config.BrokerConfiguration.AndThen(
+                config => config.Authorizations.Map(policy => policy));
```
varunpuranik, line 60:
```suggestion
            Option<AuthorizationConfig> maybePolicy = config.BrokerConfiguration.FlatMap(c => c.Authorizations);
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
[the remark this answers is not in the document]
vadim-kovalyov: Nvm, discussed offline, will fix.

varunpuranik (approved):

PR #3686, opened 2020-10-08
Fixing e2e tests

dylanbronson:
Our e2e pipeline is broken. 

When I started up edgeHub and connected a device there were 2 problems.

1. GetAuthChain threw an exception which crashed edgeHub. Instead we should return Option.None
2. PolicyAuthorizer throws exception on device trying to connect because policies aren't there yet. Instead, setting this to AllowAll for now


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityDictionary.cs
```diff
@@ -25,7 +25,7 @@ public ServiceIdentityDictionary(string actorDeviceId)
 
         public string GetActorDeviceId() => this.actorDeviceId;
 
-        public Task<Option<string>> GetAuthChain(string id) => throw new NotImplementedException("Nested Edge not enabled");
+        public Task<Option<string>> GetAuthChain(string id) => Task.FromResult(Option.None<string>());
 
         public Task<Option<string>> GetEdgeAuthChain(string id) => throw new NotImplementedException("Nested Edge not enabled");
```
and-rewsmith, line 30: Could a similar problem happen due to these related methods throwing?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityDictionary.cs
```diff
@@ -25,7 +25,7 @@ public ServiceIdentityDictionary(string actorDeviceId)
 
         public string GetActorDeviceId() => this.actorDeviceId;
 
-        public Task<Option<string>> GetAuthChain(string id) => throw new NotImplementedException("Nested Edge not enabled");
+        public Task<Option<string>> GetAuthChain(string id) => Task.FromResult(Option.None<string>());
```
yophilav, line 28: Per `\edge-hub\core\src\Microsoft.Azure.Devices.Edge.Hub.CloudProxy\ServiceProxy.cs` line 115 and line 47, why do they think the nested edge is enabled? That shouldn't be the case here for master branch E2E. 
@richma-ms Thoughts on this?

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -73,9 +70,7 @@ pub async fn broker(
     config: &BrokerConfig,
     state: Option<BrokerSnapshot>,
 ) -> Result<Broker<impl Authorizer>> {
-    let device_id = env::var(DEVICE_ID_ENV)?;
-
-    let authorizer = LocalAuthorizer::new(EdgeHubAuthorizer::new(PolicyAuthorizer::new(device_id)));
+    let authorizer = LocalAuthorizer::new(EdgeHubAuthorizer::new(AllowAll));
```
yophilav, line 73: If you want to enable all authorizer, could you please add a TODO comment and make sure we have a work item to track this. Otherwise, this can lead to a security vulnerability down the road if unattended.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityDictionary.cs
dylanbronson: In the case I'm fixing, they don't think nested edge is enabled - GetAuthChain is called by ScopeIdentitiesHandler when sending over scopes to broker, as a part of the normal, non-nested case. So GetAuthChain should not throw

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityDictionary.cs
dylanbronson: I don't think so since they're not called anywhere else in non-nested scenarios

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
dylanbronson: Vadim's going to reactivate it in his next pr that sends over policy to edgeHub. 
Added todo

and-rewsmith (approved):

PR #3687, opened 2020-10-08
[mqtt] RPC command handler (part 1)

dmolokanov:
The first part of the RPC command handler work.
This part contains command definition, deserialization from BSON, and the command handler itself.
Exact integration with Bridge infra will be is separate PR.


mqtt/mqtt-bridge/src/rpc.rs
```diff
@@ -0,0 +1,503 @@
...
+use lazy_static::lazy_static;
+use mqtt3::{
+    proto::Publication, proto::QoS, proto::SubscribeTo, Event, PublishError,
+    SubscriptionUpdateEvent,
+};
+use regex::Regex;
+use serde::{Deserialize, Serialize};
+use tracing::{error, warn};
+
+#[cfg(test)]
```
vadim-kovalyov, line 16: q: why do you need this here?

dmolokanov: for tests, I want to pass mocks into event handler code. for production code I want real implementation to be used

mqtt/mqtt-bridge/src/rpc.rs
```diff
@@ -0,0 +1,503 @@
...
+
+use crate::client::EventHandler;
+
+/// MQTT client event handler to react on RPC commands that EdgeHub sends
+/// to execute.
+///
+/// The main purpose of this handler is to establish a communication channel
+/// between EdgeHub and the upstream bridge.
+/// EdgeHub will use low level commands SUB, UNSUB, PUB. In turn the bridge
+/// sends corresposponding MQTT packet to upstream broker and waits for an ack
```
vadim-kovalyov, line 32:
```suggestion
/// sends corresponding MQTT packet to upstream broker and waits for an ack
```

mqtt/mqtt-bridge/src/rpc.rs
```diff
@@ -0,0 +1,503 @@
...
+    use matches::assert_matches;
+    use mqtt3::{
+        proto::QoS, Event, MockPublishHandle, MockUpdateSubscriptionHandle, ReceivedPublication,
+    };
+    use test_case::test_case;
+
+    use super::*;
+
+    #[test]
+    fn it_deserizes_from_bson() {
```
vadim-kovalyov, line 301:
```suggestion
    fn it_deserializes_from_bson() {
```

vadim-kovalyov: Also, what is the purpose of this test? it seems to me that you're just testing bson crate :)

vadim-kovalyov (dismissed):
LGTM! :rocket:

mqtt/mqtt-bridge/src/rpc.rs
dmolokanov: It verifies, that I assigned all serde features correct :) and yes, bson treats binary fields with some hiccups. see `serde_bytes` crate

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -183,29 +185,28 @@ impl BridgeIoSource {
 }
 
 /// This is a wrapper over mqtt3 client
-pub struct MqttClient<T>
+pub struct MqttClient<H>
```
ancaantochi, line 188: why are you changing the name for the generic type? I think T is standard value for generic types

mqtt/mqtt-bridge/src/rpc.rs
```diff
@@ -0,0 +1,504 @@
...
+///
+/// The main purpose of this handler is to establish a communication channel
+/// between EdgeHub and the upstream bridge.
+/// EdgeHub will use low level commands SUB, UNSUB, PUB. In turn the bridge
+/// sends corresponding MQTT packet to upstream broker and waits for an ack
+/// from the upstream. After ack is received it sends a special publish to
+/// downstream broker.
+pub struct RpcHandler {
+    upstream_subs: UpdateSubscriptionHandle,
+    upstream_pubs: PublishHandle,
```
ancaantochi, line 37: there might be a complication when the client needs to be recreated because of an error it can't recover from. It's not handled yet, but the plan is to recreate the pump. So maybe the rpc handle should send the event to the pump instead of having the public and sub handles.

mqtt/mqtt-bridge/src/rpc.rs
```diff
@@ -0,0 +1,504 @@
...
+        Ok(())
+    }
+
+    async fn publish_nack(&mut self, command_id: String, reason: String) -> Result<(), RpcError> {
+        let mut payload = Vec::new();
+        let doc = doc! { "reason": reason };
+        doc.to_writer(&mut payload)?;
+
+        let publication = Publication {
+            topic_name: format!("$edgehub/rpc/nack/{}", &command_id),
```
ancaantochi, line 177: extract to const

mqtt/mqtt-bridge/src/client.rs
dmolokanov: In C# - yes. In rust usually - the first letter of the trait

mqtt/mqtt-bridge/src/rpc.rs
dmolokanov: Shouldn't we re-create handler in this case?

vadim-kovalyov (approved):

mqtt/mqtt-bridge/src/rpc.rs
ancaantochi: If the message is passed to the pump then the handle doesn't need to be recreated.

mqtt/mqtt-bridge/src/client.rs
ancaantochi: According to Rust Programming Language book, T is the default.

ancaantochi (dismissed):

mqtt/mqtt-bridge/src/rpc.rs
dmolokanov: Discussed offline with Anca. There are more changes required. Will address them in following PRs

vadim-kovalyov (approved):

PR #3688, opened 2020-10-09
Log new identity services in support-bundle

damonbarry:
`iotedge support-bundle` already saves the logs for our existing host service dependencies, iotedged and dockerd. With this update, support-bundle will also save logs for the three new identity-related services that will be broken out from iotedged: certd, keyd, and identityd. The logic to pull the logs has been consolidated, and I updated a related test.

Note: I removed the Windows code block that pulls logs via `Get-WinEvent`.

### DIFF
--- a/edgelet/support-bundle/src/support_bundle.rs
+++ b/edgelet/support-bundle/src/support_bundle.rs
@@ -147,7 +147,7 @@ where
             .and_then(|s| s.write_system_log("aziot-certd", "aziot-certd"))
             .and_then(|s| s.write_system_log("aziot-identityd", "aziot-identityd"))
             .and_then(|s| s.write_system_log("iotedged", "iotedge"))
-            .and_then(Self::write_docker_log)
+            .and_then(|s| s.write_system_log("docker", "docker"))
             .and_then(Self::write_all_inspects)
             .and_then(Self::write_all_network_inspects)
     }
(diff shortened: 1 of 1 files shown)


edgelet/support-bundle/src/support_bundle.rs
```diff
@@ -143,7 +143,10 @@ where
         future::ok(self)
             .and_then(Self::write_check)
             .and_then(Self::write_module_logs)
-            .and_then(Self::write_edgelet_log)
+            .and_then(|s| s.write_system_log("aziot-keyd", "aziot-keyd"))
```
CindyXing, line 146: Before the MR refactory and integrating with iot-identity-service repo. done, I'd recommend we align this effort with MR refactory in the IIOT repo. 

Then we can link the two repo, verify and merge to master.

edgelet/support-bundle/src/support_bundle.rs
```diff
@@ -278,89 +281,19 @@ where
...
             if result.status.success() {
-                ("iotedged.txt", result.stdout)
+                (format!("{}.txt", name), result.stdout)
             } else {
-                ("iotedged_err.txt", result.stderr)
+                (format!("{}_err.txt", name), result.stderr)
             }
         } else {
             let err_message = command.err().unwrap().to_string();
             println!(
```
massand, line 292: I would recommend removing this println and the `err.txt` output for missing services. Since `'iotedge` is used for earlier versions, the `println!` will cause more support calls or Github issues.

edgelet/support-bundle/src/support_bundle.rs
```diff
@@ -143,8 +143,11 @@ where
         future::ok(self)
             .and_then(Self::write_check)
             .and_then(Self::write_module_logs)
-            .and_then(Self::write_edgelet_log)
-            .and_then(Self::write_docker_log)
+            .and_then(|s| s.write_system_log("aziot-keyd", "aziot-keyd"))
+            .and_then(|s| s.write_system_log("aziot-certd", "aziot-certd"))
+            .and_then(|s| s.write_system_log("aziot-identityd", "aziot-identityd"))
+            .and_then(|s| s.write_system_log("iotedged", "iotedge"))
+            .and_then(|s| s.write_system_log("docker", "docker"))
```
massand, line 150: In a separate method (something like `get_modules` where there is an `include_ms_only` flag), can we build a list of system services we want to query and use the iotedge version as a way to determine that list?

massand (changes requested):

edgelet/support-bundle/src/support_bundle.rs
damonbarry: When journalctl is called for a service that doesn't exist, it simple returns empty logs with status code 0. It does not return an error. So this logic is for real errors, which users will want to see.
```bash
$ journalctl -u blah
-- Logs begin at Sun 2020-08-30 18:22:36 PDT, end at Thu 2020-10-08 15:19:28 PDT. --
-- No entries --
$ echo $?
0
```

edgelet/support-bundle/src/support_bundle.rs
damonbarry: Maybe, but it sounds like more effort than it's worth. While it is possible to run one version of iotedge CLI against a different version of iotedged, they're currently packaged together so it's not very likely. And the code in this PR is backwards-compatible.

edgelet/support-bundle/src/support_bundle.rs
damonbarry: Rebased this PR on the `iiot` branch where module runtime refactoring is happening

massand (approved):
:shipit:

PR #3692, opened 2020-10-09
Remove deployment manifest minor version strict validation

richma-ms:
This removes manifest strict checking as per earlier design discussions.  Just removing the validation is not an ideal fix, as it allows people to enable 1.1 features while specifying a 1.0 version.  But this is the lowest possible risk change, later we'll need to version the JSON deserialization types so we only parse the elements relevant to a specific version, and ignore all the others.


yophilav (dismissed):
Note: Without schema version check, we are relying on tooling & logging to catch that a feature is not supported on a given schema version i.e. `StartupOrder` is not supported on 1.0 schema

PR #3695, opened 2020-10-09
Add windows images for direct method tests

lfitchett:
This adds windows images for the NumberLogger, a module used for the direct method e2e tests. It also adds a comment to the Upload direct method test, since it only works manually doe to needing a valid external blob store.


builds/misc/images.yaml
```diff
@@ -313,6 +313,13 @@ jobs:
           imageName: azureiotedge-metrics-validator
           project: MetricsValidator
 
+      # NumberLogger
```
yophilav, line 316: Please add the description to the PR comment.

test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
```diff
@@ -122,15 +122,17 @@ public async Task TestRestartModule()
...
         }
 
         [Test]
         public async Task TestUploadModuleLogs()
         {
+            Assert.Ignore("This test currently does not run automatically. To manually run, replace the sas url below with one pointing to a valid blob store");
+
             string moduleName = "NumberLogger";
             int count = 10;
-            string sasUrl = "https://lefitcheblobtest1.blob.core.windows.net/upload-test?sv=2019-02-02&st=2020-08-03T17%3A14%3A16Z&se=2020-11-04T18%3A14%3A00Z&sr=c&sp=racwdl&sig=phKgqaaxSJTcZzUcggE%2FnhDljs4%2BhvCg7IOKk8iWTcY%3D";
```
yophilav, line 133: You may want to regenerate your blob key after this.

yophilav (approved):

test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
lfitchett: its fine, that blobs going to be deleted anyway

PR #3700, opened 2020-10-09
[mqtt] Outgoing and incoming message pump for bridge

and-rewsmith:
Currently the bridge only supports connecting to a broker and ingesting messages into a store. We need to support outgoing and incoming message pumps to actually send messages.

This PR does that by refactoring the BridgeController and Bridge. It also adds a Pump component that is responsible for connecting to a broker and setting up the message ingress/egress.


mqtt/mqtt-bridge/config/default.json
```diff
@@ -2,8 +2,16 @@
     "upstream": {
         "keep_alive": "1m",
         "clean_session": false,
-        "subscriptions": [],
-        "forwards": []
+        "forwards": [
```
ancaantochi, line 5: this should be reverted

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,482 +1,204 @@
...
+        };
 
-            if let Some(f) = forward_publication {
-                debug!("Save message to store");
-                self.inner.push(f).map_err(BridgeError::Store)?;
-            } else {
-                warn!("No topic matched");
+        let local_pump = self.local_pump.run(local_shutdown_listener);
+        let remote_pump = self.remote_pump.run(remote_shutdown_listener);
+        pin_mut!(local_pump);
```
ancaantochi, line 152: nit: pin_mut!(local_pump, remote_pump);

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -1,47 +1,42 @@
...
-        system_address: String,
-        device_id: &str,
-    ) -> Result<(), BridgeError> {
-        info!("initializing bridge controller...");
-        let settings = Settings::new().map_err(BridgeError::LoadingSettings)?;
-
-        if let Some(upstream) = settings.upstream() {
-            let bridge = Bridge::new(system_address, device_id.into(), upstream.clone());
-            self.bridges.insert(upstream.name().to_string(), bridge);
+        let mut bridge_handles = vec![];
```
ancaantochi, line 15: hashmap will be useful for configuration from twin, at that point the bridge will be searched by name

mqtt/mqtt-bridge/config/default.json
and-rewsmith: Yes I am going to keep it here until I do a final test.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -300,30 +300,41 @@ impl<T: EventHandler> MqttClient<T> {
...
+            .client
+            .publish_handle()
+            .map_err(ClientError::PublishHandle)?;
+
+        Ok(publish_handle)
+    }
+
+    pub async fn handle_events(&mut self) {
+        // TODO REVIEW: Is client id ever not set
+        let client_id = self.client_id.clone().unwrap_or(String::new());
```
and-rewsmith, line 322: Client id is defined in this client wrapper as an option. Will it ever be set to None (i.e. clean session)?

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -1,47 +1,48 @@
...
+        if let Some(upstream_settings) = settings.upstream() {
+            let upstream_settings = upstream_settings.clone();
+            let upstream_bridge = async move {
+                let bridge =
+                    Bridge::new(system_address, device_id.into(), upstream_settings.clone()).await;
+
+                // TODO REVIEW: Better way to log?
+                match bridge {
```
and-rewsmith, line 23: There is a better way to log err with format params, but I didn't see it from the docs.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,170 @@
...
+        let persist = self.persist.clone();
+        let loader = self.loader.clone();
+        let mut client_shutdown = self.client_shutdown.clone();
+        let bridge_name = self.bridge_name.clone();
+        let pump_type = self.pump_type.clone();
+
+        // egress pump
+        let f1 = async move {
+            // TODO REVIEW: Is panicking fine?
+            let mut loader_borrow = loader.borrow_mut();
```
and-rewsmith, line 88: Is potential panic ok here? If panic then the thread will stop and the edgehub container will restart.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,170 @@
...
+                        debug!("egress pump extracted publication from store");
+
+                        if let Ok(Some((key, publication))) = loaded_element {
+                            // TODO REVIEW: should we be retrying?
+                            // if this failure is due to something that will keep failing it is probably safer to remove and never try again
+                            // otherwise we should retry
+                            debug!("publishing publication {:?} for egress pump", key);
+                            if let Err(e) = publish_handle.publish(publication).await {
+                                error!(message = "failed publishing publication for bridge pump", err = %e);
+                            }
```
and-rewsmith, line 118: If this publish fails, what should be our retry logic? I think it depends on why the publish fails. I think the mqtt3 client does some retries by itself. If true then that means this publish might just keep failing.

mqtt/mqtt-bridge/src/persist/publication_store.rs
```diff
@@ -25,44 +30,68 @@ impl PublicationStore<WakingMemoryStore> {
...
+        let inner = self.inner.borrow_mut();
+
         debug!("removing publication with key {:?}", key);
 
-        let mut state_lock = self.state.lock();
-        state_lock.remove(key)?;
+        let mut state_borrow = inner
+            .state
+            .try_borrow_mut()
+            .map_err(PersistError::BorrowSharedState)?;
```
and-rewsmith, line 79: We might want to panic and fail loud here. If there is an error then it is a significant logic error and we want to be able to easily see the effect of it.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,170 @@
...
+
+                            // TODO REVIEW: if removal fails, what do we do?
+                            if let Err(e) = persist.remove(key) {
+                                error!(message = "failed removing publication from store", err = %e);
+                            }
```
and-rewsmith, line 123: This removal should only fail if we pass a key which DNE. Or if we have a logic error and a borrow fails (whether we should have this failed-borrow case panic outright is a discussion in another comment).

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,482 +1,204 @@
...
+            self.connection_settings.name()
+        );
+        match select(local_pump, remote_pump).await {
+            Either::Left(_) => {
+                shutdown_handle.shutdown().await?;
+            }
+            Either::Right(_) => {
+                shutdown_handle.shutdown().await?;
             }
         }
```
and-rewsmith, line 171: For context, pump.run() cannot error. That means a shutdown signifies that shutdown has been requested. If we decide to handle client errors by recreating the pump, then we will need to alter this shutdown logic to fit this new behavior.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -343,28 +353,36 @@ impl<T: EventHandler> MqttClient<T> {
...
             .await
-            .map_err(ClientConnectError::PollClientFailure)?
+            .map_err(ClientError::PollClient)?
         {
             if let Event::SubscriptionUpdates(subscriptions) = event {
                 for subscription in subscriptions {
                     match subscription {
                         SubscriptionUpdateEvent::Subscribe(sub) => {
                             subacks.remove(&sub.topic_filter);
+                            debug!("successfully subscribed to topics");
```
and-rewsmith, line 372: give context to logs

mqtt/mqtt-bridge/src/message_handler.rs
```diff
@@ -0,0 +1,290 @@
...
+
+        handler.handle_event(Event::Publication(pub1)).unwrap();
+
+        let loader = handler.inner.loader();
+        let mut loader_borrow = loader.borrow_mut();
+
+        let mut interval = tokio::time::interval(std::time::Duration::from_secs(1));
+        futures_util::future::select(interval.next(), loader_borrow.next()).await;
+    }
+}
```
and-rewsmith, line 290: I just extracted the message handler and moved it into it's own module

mqtt/mqttd/src/main.rs
```diff
@@ -1,3 +1,4 @@
+#![type_length_limit = "1230974"]
```
and-rewsmith, line 1: Necessary to fix the build with these new changes. It doesn't seem to signify a problem, just that we now are exceeding a type substitution limit.
https://doc.rust-lang.org/reference/attributes/limits.html

mqtt/mqtt-bridge/src/controller.rs
and-rewsmith: The issue with this is that the bridge cannot be shared across threads without using locks. We can talk about the best approach for a workaround.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,482 +1,199 @@
...
+        )?;
+
+        local_pump.subscribe().await?;
+        remote_pump.subscribe().await?;
+
+        debug!("created bridge...{}", connection_settings.name());
+        Ok(Bridge {
+            local_pump,
+            remote_pump,
+            connection_settings,
```
and-rewsmith, line 127: We can make some helper functions to shrink this method

mqtt/mqtt-bridge/src/pump.rs
and-rewsmith: Actually, we should probably just exit from this async block so the pump shuts down and logs an error. That way we can handle however we want.

mqtt/mqtt-bridge/src/persist/publication_store.rs
and-rewsmith: Considering that in the pump, the best way to handle borrow is to not panic, but shut down the pump cleanly, I would feel weird being inconsistent and panicking only in this case.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+}
+
+pub struct PumpPair {
+    pub local_pump: Pump,
+    pub remote_pump: Pump,
+}
+
+impl PumpPair {
```
and-rewsmith, line 51: I am going to add doc comments for all these types.

mqtt/mqtt-bridge/src/bridge.rs
and-rewsmith: Actually I am wrong, pump.run() I think should error in two cases:
- The client needs to be recreated
- We fail a borrow of the shared state (might not be relevant if we switch to locking)

I think if a pump stops, the cleanest approach is to recreate it.

dmolokanov (dismissed):
Good work!

mqtt/mqtt-bridge/src/persist/publication_store.rs
```diff
@@ -28,44 +33,68 @@ where
...
 
-    pub fn loader(&mut self) -> Arc<Mutex<MessageLoader<S>>> {
-        Arc::clone(&self.loader)
+    pub fn loader(&self) -> Rc<RefCell<MessageLoader<S>>> {
+        let inner = self.inner.borrow_mut();
+        inner.loader.clone()
+    }
+}
+
+impl<S: StreamWakeableState> Clone for PublicationStore<S> {
```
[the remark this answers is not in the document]
and-rewsmith: That didn't work. I remember trying and it complained. I looked it up and the resources told me to implement clone myself. I can find the error if you want confirmation.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,326 @@
...
+            _ = shutdown => {
+                if let Err(e) = client_shutdown.shutdown().await {
+                    error!(message = "failed to shutdown incoming publication loop for bridge pump", err = %e);
+                }
+
+                if let Err(e) = loader_shutdown.send(()) {
+                    error!(message = "failed to shutdown publish loop for bridge pump");
+                }
+            },
+        }
```
and-rewsmith, line 321: I am getting this clippy error:
```
generally you want to avoid `&mut &mut _` if possible
```

Any idea why?

mqtt/mqtt-bridge/src/persist/publication_store.rs
```diff
@@ -28,44 +33,68 @@ where
...
 
-        let mut state_lock = self.state.lock();
-        state_lock.insert(key, message)?;
-        self.offset += 1;
+        let mut state_borrow = inner
+            .state
+            .try_borrow_mut()
+            .map_err(PersistError::BorrowSharedState)?;
+        state_borrow.insert(key, message)?;
+        drop(state_borrow);
```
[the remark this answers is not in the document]
and-rewsmith: Since there are no awaits we don't have to drop here. Will remove.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,491 +1,128 @@
...
-        } else {
-            MqttClient::tcp(
-                address,
-                self.connection_settings.keep_alive(),
-                self.connection_settings.clean_session(),
-                MessageHandler::new(persistor, topic_filters),
-                credentials,
-            )
-        };
+        pumps.local_pump.subscribe().await?;
```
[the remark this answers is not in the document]
and-rewsmith: I was thinking that when the subscription update change comes through, this can be changed then. I am not sure where is best to move it now. We could move it to run() now but it is going to be doing the same thing.

mqtt/mqtt-bridge/src/persist/publication_store.rs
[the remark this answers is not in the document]
and-rewsmith: Wait we actually need this to adjust the offset two lines down. I have renamed vars for clarity.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,283 @@
...
+        let persist = self.persist.clone();
+        let loader = self.loader.clone();
+        let mut client_shutdown = self.client_shutdown.clone();
+        let bridge_name = self.pump_context.bridge_name.clone();
+        let pump_type = self.pump_context.pump_type.clone();
+
+        // egress pump
+        let f1 = async move {
+            let mut loader_borrow = loader.borrow_mut();
+            let mut receive_fut = loader_shutdown_rx.into_stream();
```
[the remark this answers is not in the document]
and-rewsmith: The problem is that it cannot be awaited multiple times, so we need to convert to stream.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,283 @@
...
+                bridge_name, pump_type
+            );
+
+            loop {
+                let mut publish_handle = publish_handle.clone();
+                match select(receive_fut.next(), loader_borrow.try_next()).await {
+                    Either::Left((shutdown, _)) => {
+                        debug!("egress pump received shutdown signal");
+                        if let None = shutdown {
+                            error!(message = "unexpected behavior from shutdown signal while signalling bridge pump shutdown")
```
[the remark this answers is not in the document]
and-rewsmith: Great question. The return type of receive_fut.next() is ```Option<Stream::Item<{unknown}>>```. I was assuming it would return None if something happened to the underlying receiver, but I now realize that is a big assumption.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,283 @@
...
+                        }
+                    }
+                }
+            }
+
+            debug!("pumps for {} bridge stopped...", bridge_name);
+        };
+
+        // incoming pump
+        let bridge_name = self.pump_context.bridge_name.clone();
```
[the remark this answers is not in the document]
and-rewsmith: I thought the logs would be cleaner if I separated the fields out, but this is not maintainable. I am going to try to implement display and then change the logging.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,283 @@
...
+
+        let f1 = f1.fuse();
+        let f2 = f2.fuse();
+        let mut shutdown = shutdown.fuse();
+        pin_mut!(f1);
+        pin_mut!(f2);
+
+        select! {
+            _ = f1 => {
+                error!(message = "publish loop failed and exited for bridge pump");
```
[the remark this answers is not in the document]
and-rewsmith: I believe this case is that the pump is stopped before shutdown, in which case it is an error right? See the shutdown in the select below on 317.

mqtt/mqtt-bridge/src/pump.rs
and-rewsmith: Denis suggested that we can use borrow and then our tests can catch any bad panic cases if they arise.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -1,541 +1,126 @@
...
+        debug!("Bridge {} stopped...", self.connection_settings.name());
+        Ok(())
     }
 }
 
 /// Bridge error.
 #[derive(Debug, thiserror::Error)]
 pub enum BridgeError {
-    #[error("failed to save to store.")]
+    #[error("Failed to save to store.")]
```
ancaantochi, line 97: I see RpcError uses lowercase and I think from other error I saw it was using lowercase

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,283 @@
...
+                "{} bridge starting ingress publication processing for pump {:?}...",
+                bridge_name, pump_type
+            );
+            self.client.handle_events().await;
+        };
+
+        let f1 = f1.fuse();
+        let f2 = f2.fuse();
+        let mut shutdown = shutdown.fuse();
+        pin_mut!(f1);
```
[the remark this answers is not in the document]
and-rewsmith: I get a cannot be unpinned error if I don't fuse.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,283 @@
...
+    client::{ClientShutdownHandle, MqttClient},
+    message_handler::MessageHandler,
+    persist::{MessageLoader, PublicationStore, WakingMemoryStore},
+    settings::{ConnectionSettings, Credentials, TopicRule},
+};
+
+const BATCH_SIZE: usize = 10;
+
+/// Provides context used for logging in the bridge pumps and client implementations
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
and-rewsmith: I cleaned up all the logging and implemented Display

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,342 @@
...
+            bridge_name,
+        }
+    }
+}
+
+impl Display for PumpContext {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        write!(
+            f,
+            "{:?} pump on {} bridge",
```
and-rewsmith, line 88: Is it fine to use debug on this enum PumpType?

mqtt/mqtt-bridge/src/pump.rs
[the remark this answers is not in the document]
dmolokanov: if it finished on its own - yes, it is the error. but someone can request it to shut it down. in this case, it is a legit operation. In other words - if the result of operation Err - log the error, otherwise it is a clean shutdown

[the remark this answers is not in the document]
and-rewsmith: I totally agree with this, but isn't that the way it is currently implemented?

mqtt/mqtt-bridge/src/pump.rs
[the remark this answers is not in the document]
dmolokanov: weird, but makes sense

mqtt/mqtt-bridge/src/pump.rs
dmolokanov: nope, I see this too

dmolokanov: try to expand macros and see what code it generates. alternatively you can try `fn future::select()`

mqtt/mqtt-bridge/src/pump.rs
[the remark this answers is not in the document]
and-rewsmith: Sorry mistyped. If I have the fuse and I don't pin, I get a "cannot be unpinned" error

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -0,0 +1,347 @@
...
+        let mut shutdown = shutdown.fuse();
+        pin_mut!(egress_pump, ingress_pump);
+
+        select! {
+            _ = egress_pump => {
+                error!("{} failed egress publication loop and exited", pump_context);
+            },
+            _ = ingress_pump => {
+                error!("{} failed ingress publication loop and exited", pump_context);
+            },
```
and-rewsmith, line 332: We need to shutdown and exit in both of the above cases. Currently it won't work.

mqtt/mqtt-bridge/src/bridge.rs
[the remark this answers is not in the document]
dmolokanov: makes sense I guess

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -1,48 +1,44 @@
...
+                            error!(err = %e, "failed running {} bridge", upstream_settings.name());
+                        }
+                    }
+                    Err(e) => {
+                        error!(err = %e, "failed to create {} bridge", upstream_settings.name());
+                    }
+                };
+            };
+
+            let upstream_bridge_handle = task::spawn_local(upstream_bridge);
```
and-rewsmith, line 33: This needs to be select. It panics now.

ancaantochi (approved):

PR #3704, opened 2020-10-12
Fix iotedge support-bundle doesn't work if iotedged isn't running

pmzara:
Currently iotedge support-bundle will fail if iotedged isn't running. It should still collect host and docker logs.
For this fix, I'm catching the failed operation and return a empty list. This will not cause any problem in the execution of the functions and will produce the logs for all the others operations except for the operations that using the information produced by the runtime.


lfitchett (approved):
Looks good as long as you tested that it works correctly. Get arnav to look as well. Since I helped with the code, I don't really count as a second set of eyes.

PR #3711, opened 2020-10-13
[mqtt] Unblock external transports when Authorizer ready

dmolokanov:
Encapsulate the exact mechanism to notify protocols to start serving external clients with `BrokerReady` component.
`BrokerReady` consist of two parts: `BrokerReadyHandle` and `BrokerReadySignal`.
`BrokerReadyHandle` - a handle to send readiness events.
`BrokerReadySignal` - await for all required readiness events collected and then proceeds.

It is only one readiness event required so far: `EdgeHubAuthorizer` is updated with initial device scopes. However, it can be more.

### DIFF
--- a/mqtt/mqtt-broker/src/ready.rs
+++ b/mqtt/mqtt-broker/src/ready.rs
@@ -57,7 +57,7 @@ where
     }
 }
 
-/// A handle to await all readiness events collected buy the handle.
+/// A handle to await all readiness events collected by the handle.
 pub struct BrokerReadySignal<E>(Receiver<E>);
 
 impl<E> BrokerReadySignal<E>
--- a/mqtt/mqtt-broker/src/ready.rs
+++ b/mqtt/mqtt-broker/src/ready.rs
@@ -1,0 +1,180 @@
+use std::{collections::HashSet, fmt::Display, hash::Hash};
+
+use tokio::sync::broadcast::{self, Receiver, RecvError, Sender};
+use tracing::{debug, warn};
+
+/// `BrokerReady` component acts as a intermidiate entity to determine whether
+/// `Broker` is ready to serve external clients. It awaits on several events
+/// from components Broker depends on.
+///
+/// Notes: Authenticator and Authorizer both needs some data which not
+/// available when process started. They receive it using internal
+/// communication channels. Once data is loaded and components initialized
+/// they send corresponding events to the `BrokerReady` which in turn unblocks
+/// external transports.
+pub struct BrokerReady<E> {
+    event_send: Sender<E>,
+}
+
+impl<E> Default for BrokerReady<E> {
+    fn default() -> Self {
+        let (event_send, _) = broadcast::channel(5);
+        Self { event_send }
+    }
+}
+
+impl<E> BrokerReady<E> {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn handle(&self) -> BrokerReadyHandle<E> {
+        BrokerReadyHandle(self.event_send.clone())
+    }
+
+    pub fn signal(&self) -> BrokerReadySignal<E> {
+        BrokerReadySignal(self.event_send.subscribe())
+    }
+}
+
+/// A handle to send readiness signals to `BrokerReady`.
+pub struct BrokerReadyHandle<E>(Sender<E>);
+
+impl<E> BrokerReadyHandle<E>
+where
+    E: Clone + Display,
+{
+    pub fn send(&mut self, event: E) {
+        loop {
+            let event = event.clone();
+            debug!("sending broker ready event {}", event);
+
+            // attempt to send event until all recepients receive it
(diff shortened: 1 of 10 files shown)


mqtt/mqtt-broker/src/ready.rs
```diff
@@ -0,0 +1,180 @@
...
+
+            // attempt to send event until all recepients receive it
+            if matches!(self.0.send(event), Ok(recv) if recv == self.0.receiver_count()) {
+                break;
+            }
+        }
+    }
+}
+
+/// A handle to await all readiness events collected buy the handle.
```
vadim-kovalyov, line 60:
```suggestion
/// A handle to await all readiness events collected by the handle.
```

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -14,23 +14,32 @@ use mqtt_broker::{
 /// telemetry messages, etc...
 ///
 /// For non-iothub-specific primitives it delegates the request to an inner authorizer (`PolicyAuthorizer`).
-#[derive(Debug)]
```
vadim-kovalyov, line 17: why remove?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -14,23 +14,32 @@ use mqtt_broker::{
...
 where
     Z: Authorizer<Error = E>,
     E: StdError,
 {
-    pub fn new(authorizer: Z) -> Self {
+    pub fn new(authorizer: Z, broker_ready: BrokerReadyHandle) -> Self {
+        Self::create(authorizer, Some(broker_ready))
+    }
+
+    pub fn without_ready_handle(authorizer: Z) -> Self {
```
vadim-kovalyov, line 33: nit: vice versa? 
```
new(inner: Z) -> Self
with_ready_handle(inner: Z, ready_handle: BrokerReadyHandle) -> Self
```

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -266,6 +275,11 @@ where
                 self.inner.update(update)?;
             }
         };
+
+        if let Some(mut broker_ready) = self.broker_ready.take() {
```
vadim-kovalyov, line 279: I think you need to move this into match - Ok case. It is possible that PolicyUpdate comes first, and would cause false-positive ready signal.

vadim-kovalyov (changes requested):
LGTM, just one thing I believe we need to fix - put signal into match-ok case in EHAuthorizer.

mqtt/mqtt-broker/src/ready.rs
```diff
@@ -0,0 +1,180 @@
...
+/// communication channels. Once data is loaded and components initialized
+/// they send corresponding events to the `BrokerReady` which in turn unblocks
+/// external transports.
+pub struct BrokerReady<E> {
+    event_send: Sender<E>,
+}
+
+impl<E> Default for BrokerReady<E> {
+    fn default() -> Self {
+        let (event_send, _) = broadcast::channel(5);
```
ancaantochi, line 21: it should use a const for 5

mqtt/mqtt-broker/src/ready.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// A handle to send readiness signals to `BrokerReady`.
+pub struct BrokerReadyHandle<E>(Sender<E>);
+
+impl<E> BrokerReadyHandle<E>
+where
+    E: Clone + Display,
+{
+    pub fn send(&mut self, event: E) {
+        loop {
```
ancaantochi, line 48: should there be a wait time between sends?

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
dmolokanov: it is supposed to be used with ready_handle always. however, for tests, it adds some inconvenience.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
dmolokanov: oops, that's right. good catch!

mqtt/mqtt-broker/src/ready.rs
dmolokanov: I made it a second thought it probably doesn't need a loop

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
dmolokanov: I guess with intermediate changes I doesn't want to be Debug. and we don't use it anywhere anyways

vadim-kovalyov (approved):

PR #3713, opened 2020-10-13
Pass SSL client certificate through API proxy to EdgeHub via custom header

richma-ms:
This change makes nginx forward the client SSL cert to EdgeHub via a customer header, so that EdgeHub can parse it back into an X509Certificate2 when receiving the forwarded HTTP request and perform authentication.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/HttpRequestAuthenticator.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/HttpRequestAuthenticator.cs
@@ -4,7 +4,9 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Http.Controllers
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Net;
     using System.Security.Cryptography.X509Certificates;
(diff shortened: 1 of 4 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/HttpRequestAuthenticator.cs
```diff
@@ -40,6 +42,26 @@ public async Task<HttpAuthResult> AuthenticateAsync(string actorDeviceId, Option
             IClientCredentials clientCredentials;
             X509Certificate2 clientCertificate = await context.Connection.GetClientCertificateAsync();
 
+            if (clientCertificate == null)
+            {
+                // If the connection came through the API proxy, the client cert
+                // would have been forwarded in a custom header
+                if (context.Request.Headers.TryGetValue(Constants.ClientCertificateHeaderKey, out StringValues clientCertHeader) && clientCertHeader.Count > 0)
```
huguesBouvier, line 49: Shoudn't you do the clientCertHeader.Count > 0 before?

huguesBouvier (approved):

huguesBouvier (approved):

PR #3714, opened 2020-10-13
Make container registry config in end-to-end tests optional

damonbarry:
The service deployment pipeline (`builds/service/service-deployment.yaml`) runs a single end-to-end test ("TempSensor") in an Azure pipeline. The IoT Hub team uses it to validate new deployments of their service. Since the pipeline only runs a single, basic test, it's context.json file is very simple:
```json
{
  "logFile": "/path/to/testoutput.log"
}
```
The pipeline has been running against 1.0.9, but now that 1.0.10 is released I tried to upgrade it. Something has changed during this release, and the "registries" object in context.json is no longer optional, causing a thrown exception in the fixtures. This change restores the correct behavior.

### DIFF
--- a/test/Microsoft.Azure.Devices.Edge.Test/SetupFixture.cs
+++ b/test/Microsoft.Azure.Devices.Edge.Test/SetupFixture.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.Edge.Test
         {
             using var cts = new CancellationTokenSource(Context.Current.SetupTimeout);
             CancellationToken token = cts.Token;
-            Option<Registry> bootstrapRegistry = Option.Maybe(Context.Current.Registries.First());
+            Option<Registry> bootstrapRegistry = Option.Maybe(Context.Current.Registries.FirstOrDefault());
 
             this.daemon = await OsPlatform.Current.CreateEdgeDaemonAsync(
                 Context.Current.InstallerPath,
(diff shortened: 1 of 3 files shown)


test/Microsoft.Azure.Devices.Edge.Test/SetupFixture.cs
```diff
@@ -22,7 +22,7 @@ public async Task BeforeAllAsync()
         {
             using var cts = new CancellationTokenSource(Context.Current.SetupTimeout);
             CancellationToken token = cts.Token;
-            Option<Registry> bootstrapRegistry = Option.Maybe(Context.Current.Registries.First());
+            Option<Registry> bootstrapRegistry = Option.Maybe(Context.Current.Registries.FirstOrDefault());
```
yophilav, line 25: 1. Given that 1.0.9 is going to be replaced with 1.0.10, do you think it would be reasonable to update the `GetAndValidateRegistries()` in `Context.cs` ensuring that at least a register is provided? 
2. The `SasManualProvisioningFixture` and `X509ManualProvisioningFixture` also uses `Context.Current.Registries` which would likely to be running into an exception if the default registry is not provied.

damonbarry: 1. You don't need to provide any registry information in context.json. If you don't, the end-to-end tests will run using publicly-available mcr.microsoft.com images, which should work just fine. In fact, that's what the service deployment pipeline does; it uses publicly-released images.
2. True, those two fixtures also reference `Context.Current.Registries`, but they don't assume the IEnumerable has one or more items, so they don't run into this problem.

yophilav (approved):

PR #3721, opened 2020-10-15
Bridge update from twin 1/2

ancaantochi:
First part of bridge twin updates:

- Change settings to align with bridge twin settings 
- Add bridge update command

### DIFF
--- a/mqtt/mqtt-bridge/src/controller.rs
+++ b/mqtt/mqtt-bridge/src/controller.rs
@@ -1,15 +1,13 @@
 use std::collections::HashMap;
 
 use futures_util::future::{self, join_all};
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::sync::mpsc::{self, UnboundedSender};
 use tracing::{error, info};
 
+use crate::bridge::{Bridge, BridgeError};
 use crate::settings::Settings;
-use crate::{
-    bridge::{Bridge, BridgeError},
-    BridgeUpdate,
-};
 
 /// Controller that handles the settings and monitors changes, spawns new Bridges and monitors shutdown signal.
 pub struct BridgeController {
@@ -80,19 +78,23 @@ impl Default for BridgeController {
 
 #[derive(Clone, Debug)]
 pub struct BridgeControllerHandle {
-    sender: UnboundedSender<BridgeUpdate>,
+    sender: UnboundedSender<BridgeControllerUpdate>,
 }
 
 impl BridgeControllerHandle {
-    pub fn send(&mut self, message: BridgeUpdate) -> Result<(), Error> {
+    pub fn send(&mut self, message: BridgeControllerUpdate) -> Result<(), Error> {
         self.sender
             .send(message)
             .map_err(Error::SendControllerMessage)
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct BridgeControllerUpdate {
+    // TODO: add settings
+}
 #[derive(Debug, Error)]
 pub enum Error {
     #[error("An error occurred sending a message to the controller.")]
-    SendControllerMessage(#[source] tokio::sync::mpsc::error::SendError<BridgeUpdate>),
+    SendControllerMessage(#[source] tokio::sync::mpsc::error::SendError<BridgeControllerUpdate>),
 }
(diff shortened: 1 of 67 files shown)


mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -56,3 +69,33 @@ impl BridgeController {
...
+    pub fn send(&mut self, message: BridgeControllerUpdate) -> Result<(), Error> {
+        self.sender
+            .send(message)
+            .map_err(Error::SendControllerMessage)
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct BridgeControllerUpdate {
+    // TODO: add settings
```
vadim-kovalyov, line 94: are you going to address this in the next PR?

vadim-kovalyov (approved):

mqtt/mqtt-bridge/src/controller.rs
ancaantochi: yes

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -1,20 +1,33 @@
...
+use tokio::sync::mpsc::{self, UnboundedSender};
 use tracing::{error, info};
 
 use crate::bridge::{Bridge, BridgeError};
 use crate::settings::Settings;
 
 /// Controller that handles the settings and monitors changes, spawns new Bridges and monitors shutdown signal.
-#[derive(Default)]
 pub struct BridgeController {
     bridges: HashMap<String, Bridge>,
```
and-rewsmith, line 14: I think there is going to be an issue with this new structure where we don't store the hashmap of bridges. This was eliminated because we were going to use Rc types within the bridge internally, but switched back to locking because of some errors spawning threads. 

If we think we are going to keep the locks, we should have a PR to add back the hashmap of bridges, or you can add it back here.

ancaantochi: removed the hashmap as it's not needed

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -1,14 +1,30 @@
...
 /// Controller that handles the settings and monitors changes, spawns new Bridges and monitors shutdown signal.
-pub struct BridgeController {}
+pub struct BridgeController {
+    handle: BridgeControllerHandle,
+}
 
 impl BridgeController {
-    pub async fn run(system_address: String, device_id: String, settings: BridgeSettings) {
+    pub fn new() -> Self {
+        let (sender, _updates) = mpsc::unbounded_channel();
```
and-rewsmith, line 17: What is the structure for these update publications that edgehub will send?

mqtt/mqtt-bridge/tests/config.json
```diff
@@ -39,14 +41,14 @@
...
+                    "direction": "in",
+                    "topic": "temp/#",
+                    "outPrefix": "floor/kitchen"
+                },
                 {
-                    "pattern": "some",
-                    "remote": "remote"
+                    "direction": "out",
+                    "topic": "some",
+                    "outPrefix": "remote"
```
and-rewsmith, line 51: Why is this changing now?

mqtt/mqtt-edgehub/src/command/bridge_update.rs
```diff
@@ -0,0 +1,51 @@
...
+        info!("received bridge update from EdgeHub.");
+        let update: BridgeControllerUpdate =
+            serde_json::from_slice(&publication.payload).map_err(Error::ParseBridgeUpdate)?;
+
+        self.controller_handle
+            .send(update)
+            .map_err(Error::SendBridgeUpdate)?;
+        Ok(())
+    }
+}
```
and-rewsmith, line 42: What is the testing plan here? Integration tests or mocking? I think it is hard to test now since we don't expose functionality for reading these updates.

and-rewsmith (approved):
I have some questions but LGTM

mqtt/mqtt-bridge/tests/config.json
ancaantochi: to align with twin updates schema

mqtt/mqtt-bridge/src/controller.rs
ancaantochi: will be an array of Bridge updates each with a name and an array of topic_rules

mqtt/mqtt-edgehub/src/command/bridge_update.rs
ancaantochi: I think mocking or both. Will follow what authz does.

PR #3722, opened 2020-10-15
Proxy environment for end-to-end tests

damonbarry:
This change adds a new job to the end-to-end tests pipeline. The job runs all the end-to-end tests on a self-hosted Linux agent that can only access HTTP resources through a proxy server. We had a similar environment in the older E2E release pipeline; this adds it to the newer pipeline.

I also updated the documentation and ARM scripts for deploying the environment (including removing Windows resources). I moved them from builds/e2e (where they were mingled with all the pipeline definitions) into a new proxy/ subfolder.

In the new environment, the VMs that actually run the tests are the agents. Due (I think) to improved proxy support in Azure pipelines agents compared to two years ago, we no longer need to install the agent on the proxy server (which has unrestricted outbound internet access) and forward commands to the target VMs via SSH scripts. So this new topology is much simpler.

_Note_ that while proxy support in the agent is much improved, it's not perfect. For example, I had to disable the PublishTestResults task in the proxy pipeline because it takes ~15 minutes to complete, compared to ~30 seconds normally. This means you won't be able to find the proxy pipeline results under the "Tests" tab in Azure DevOps. Instead you'll just have to download the full log artifacts from the "Summary" tab.


builds/e2e/e2e.yaml
```diff
@@ -63,26 +63,50 @@ jobs:
...
+    - template: templates/e2e-clear-docker-cached-images.yaml
     - template: templates/e2e-run.yaml
 
+################################################################################
+  - job: linux_amd64_proxy
+################################################################################
+    displayName: Linux amd64 behind a proxy
+
+    pool:
+      name: $(pool.name)
```
yophilav, line 95: Do you have a walk-through document that goes over the steps to create/maintenance a proxy-test-agent?

builds/e2e/proxy/configure_proxy.sh
```diff
@@ -3,26 +3,18 @@
...
 
-home="$(eval echo ~$user)"
-mkdir -p "$home/.ssh"
-echo -e "$encoded_key" | base64 -d > "$home/.ssh/id_rsa"
-chown -R "$user:$user" "$home/.ssh"
-chmod 700 "$home/.ssh"
-chmod 600 "$home/.ssh/id_rsa"
+for i in `seq 3`
+do
+    sleep 5
```
yophilav, line 12: Why is `sleep 5` needed here?

builds/e2e/proxy/configure_runner.sh
```diff
@@ -0,0 +1,51 @@
...
+
+echo 'Installing PowerShell Core and .NET Core 3.1'
+
+apt-get update
+apt-get install -y curl git wget apt-transport-https
+wget -q 'https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb'
+dpkg -i packages-microsoft-prod.deb
+apt-get update
+add-apt-repository universe
+apt-get install -y powershell dotnet-sdk-3.1
```
yophilav, line 18: I thought these (Line 12-18) are one-time prerequisite for a VM to run E2E test. 
I don't think we need to install that every time the test is run?

builds/e2e/proxy/configure_proxy.sh
yophilav, line 12: Why is `sleep 5` needed here?

builds/e2e/e2e.yaml
yophilav, line 95: Do you have a walk-through document that goes over the steps to create/maintenance a proxy-test-agent?

builds/e2e/proxy/proxy.md
```diff
@@ -0,0 +1,84 @@
...
+
+With a secret URL and an IP address, you can SSH into a runner VM like this:
+
+```sh
+az keyvault secret show --id '<>' -o tsv --query value > ~/.ssh/id_rsa.runner
+chmod 600 ~/.ssh/id_rsa.runner
+ssh -i ~/.ssh/id_rsa.runner azureuser@<ip addr>
+```
+
+To install and configure Azure Pipelines agent, see [Self-hosted Linux Agents](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/v2-linux?view=azure-devops) and [Run a self-hosted agent behind a web proxy](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/proxy?view=azure-devops&tabs=unix).
```
yophilav, line 73: I think you can automate this part using the VSTS command line argument `VSTS_AGENT_INPUT_<ARGUMENT_NAME> `.  I will need to revisit this to automate nested edge E2E as well.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -59,30 +60,41 @@ public string[] GetInstallCommandsFromLocal(string path)
...
+                {
+                    // Based on instructions at:
+                    // https://github.com/MicrosoftDocs/azure-docs/blob/058084949656b7df518b64bfc5728402c730536a/articles/iot-edge/how-to-install-iot-edge-linux.md
+                    $"{curl} https://packages.microsoft.com/config/{this.os}/{this.version}/multiarch/prod.list > /etc/apt/sources.list.d/microsoft-prod.list",
+                    $"{curl} https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > /etc/apt/trusted.gpg.d/microsoft.gpg",
+                    $"{prefix}apt-get update",
+                    $"{prefix}apt-get install --yes iotedge"
+                },
+                // TODO: 10/14/2020 support yum behind a proxy
+                SupportedPackageExtension.Rpm => new[]
```
yophilav, line 85: What's the existing issue with yum behind proxy?

builds/e2e/proxy/configure_runner.sh
damonbarry: This script is part of the ARM (Azure Resource Manager) automated deployment for the proxy environment. This script is only run when the VMs are created. It is not part of the end-to-end tests. Everything in the `proxy/` folder is part of the deployment, not the tests.

builds/e2e/proxy/configure_proxy.sh
damonbarry: VM creation was intermittently failing, so I added the retry logic (including sleep) and the problem went away.

builds/e2e/e2e.yaml
damonbarry: See `builds/e2e/proxy/proxy.md` in this PR.

builds/e2e/proxy/proxy.md
damonbarry: Yes, you can do unattended install of the agent, but that's a project for another day. The updates here provide more/better automation than the previous version, but agent installation is still manual.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
damonbarry: None that I know of, but we don't run CentOs behind a proxy in our test pipeline so I didn't implement it.

damonbarry: I'll replace the TODO with a comment clarifying that we haven't implemented proxy support in the CentOs path.

damonbarry: Updated

yophilav (approved):
Thank you for the clarification!

PR #3726, opened 2020-10-15
IIOT feature flags

richma-ms:
- Consolidate nested edge and mqtt broker feature flags in "experimentalFeatures"
- Default both feature flags to false
- Skip all MQTT broker components in both watchdog and EH core if it's disabled
- Skip MQTT protocol head if MQTT broker is enabled

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
@@ -85,8 +85,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Service
             (bool isEnabled, bool usePersistentStorage, StoreAndForwardConfiguration config, string storagePath, bool useBackupAndRestore, Option<string> storageBackupPath, Option<ulong> storageMaxTotalWalSize, Option<int> storageMaxOpenFiles, Option<StorageLogLevel> storageLogLevel) storeAndForward =
                 this.GetStoreAndForwardConfiguration();
 
-            IConfiguration configuration = this.configuration.GetSection("experimentalFeatures");
-            ExperimentalFeatures experimentalFeatures = ExperimentalFeatures.Create(configuration, Logger.Factory.CreateLogger("EdgeHub"));
+            IConfiguration experimentalFeaturesConfig = this.configuration.GetSection(Constants.ConfigKey.ExperimentalFeatures);
+            ExperimentalFeatures experimentalFeatures = ExperimentalFeatures.Create(experimentalFeaturesConfig, Logger.Factory.CreateLogger("EdgeHub"));
 
             MetricsConfig metricsConfig = new MetricsConfig(this.configuration.GetSection("metrics:listener"));
 
@@ -96,10 +96,14 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Service
             this.RegisterAmqpModule(builder);
             builder.RegisterModule(new HttpModule(this.iotHubHostname));
 
-            var authConfig = this.configuration.GetSection("authAgentSettings");
-            builder.RegisterModule(new AuthModule(authConfig));
-            var mqttBrokerConfig = this.configuration.GetSection("mqttBrokerSettings");
-            builder.RegisterModule(new MqttBrokerModule(mqttBrokerConfig));
+            if (experimentalFeaturesConfig.GetValue(Constants.ConfigKey.MqttBrokerEnabled, false))
+            {
+                var authConfig = this.configuration.GetSection("authAgentSettings");
+                builder.RegisterModule(new AuthModule(authConfig));
+
+                var mqttBrokerConfig = this.configuration.GetSection("mqttBrokerSettings");
+                builder.RegisterModule(new MqttBrokerModule(mqttBrokerConfig));
+            }
         }
 
         internal static Option<UpstreamProtocol> GetUpstreamProtocol(IConfigurationRoot configuration) =>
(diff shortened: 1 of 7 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -95,11 +95,15 @@ public void Register(ContainerBuilder builder)
             this.RegisterMqttModule(builder, storeAndForward, optimizeForPerformance);
             this.RegisterAmqpModule(builder);
             builder.RegisterModule(new HttpModule(this.iotHubHostname));
+
+            if (experimentalFeaturesConfig.GetValue(Constants.ConfigKey.MqttBrokerEnabled, false))
```
varunpuranik, line 99: We should probably use the ExperimentalFeatures type? https://github.com/Azure/iotedge/blob/master/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/ExperimentalFeatures.cs

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/appsettings_hub.json
```diff
@@ -1,6 +1,10 @@
 ﻿{
+  "experimentalFeatures": {
+    "nestedEdgeEnabled": false,
+    "mqttBrokerEnabled": false
+  },
```
varunpuranik, line 5: Remove this.

varunpuranik: No need to add it to appsettings_hub.json

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/appsettings_hub.json
```diff
@@ -35,12 +38,14 @@
...
-    "enabled": true,
     "port": 7120,
     "baseUrl": "/authenticate/"
   },
   "IotHubConnectionPoolSize": 1,
-  "IotHubConnectionString": "",
+  "EdgeHubDevServerCertificateFile": "C:\\temp\\edgehub_certs\\iot-edge-device-ca-edgehubCA-full-chain.cert.pem",
+  "EdgeHubDevServerPrivateKeyFile": "C:\\temp\\edgehub_certs\\iot-edge-device-ca-edgehubCA.key.pem",
+  "EdgeHubDevTrustBundleFile": "C:\\temp\\edgehub_certs\\iot-edge-device-ca-edgehubCA.cert.pem",
+  "IotHubConnectionString": "HostName=iiot-nested-edge-hub.azure-devices.net;DeviceId=richmaEdge2;ModuleId=$edgeHub;SharedAccessKey=es6OvUmPP/yeky9TU0juGhPD7ZNsNVCl3ZnGrCKWkxA=",
```
varunpuranik, line 48: Revert, and rotate keys for this device.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/appsettings_hub.json
```diff
@@ -70,7 +75,7 @@
   "storeAndForward": {
     "timeToLiveSecs": "-1"
   },
-  "RuntimeLogLevel": "info",
+  "RuntimeLogLevel": "debug",
```
varunpuranik, line 78: Revert

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -123,7 +127,11 @@ void RegisterMqttModule(ContainerBuilder builder, (bool isEnabled, bool usePersi
             bool clientCertAuthEnabled = this.configuration.GetValue(Constants.ConfigKey.EdgeHubClientCertAuthEnabled, false);
 
             IConfiguration mqttSettingsConfiguration = this.configuration.GetSection("mqttSettings");
-            builder.RegisterModule(new MqttModule(mqttSettingsConfiguration, topics, this.serverCertificate, storeAndForward.isEnabled, clientCertAuthEnabled, optimizeForPerformance, this.sslProtocols));
+
+            if (mqttSettingsConfiguration.GetValue("enabled", false))
+            {
```
vipeller, line 132: in 'GetEdgeHubProtocolHeadAsync' the legacy mqtt depends on 2 flags (mqttSettingsConfiguration and experimental/brokerEnabled) while here the registration happens based only on mqttSettingsCongfiguration. Doesn't seem a problem as it cannot happen the legacy mqtt is not registered but then enabled, but is it worth to use the same logic here to make it consistent?

edge-hub/watchdog/src/main.rs
```diff
@@ -40,19 +44,25 @@ fn main() -> Result<()> {
...
-        Arc::clone(&should_shutdown),
-    ) {
-        Ok(handle) => Some(handle),
-        Err(e) => {
-            should_shutdown.store(true, Ordering::Relaxed);
-            error!("Could not start MQTT Broker process. {}", e);
-            None
-        }
-    };
+    let mut broker_handle = None;
```
varunpuranik, line 47: Something like this might be better. Check with @dmolokanov  
let broker_handle = match env::var("experimentalFeatures:mqttBrokerEnabled").ok().filter(|v| v =="true")
    {
        Some(_) => ...,
        None => None,
    };

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
richma-ms: yea good point

varunpuranik (approved):

PR #3728, opened 2020-10-16
M2M fix - property parsing problems

vipeller:
M2M messages could go through, but the parameters encoded in the topic were miss-parsed due to a missing /inputs/ tag


dmolokanov (approved):
👍

PR #3731, opened 2020-10-16
[TEST-ONLY] MQTT Broker only E2E

yophilav:
MQTT Broker only E2E


e2e_deployment_files/dm_module_to_module_deployment.json
```diff
@@ -40,8 +40,17 @@
...
+                "value": "true"
+              },
+              "experimentalFeatures:nestedEdgeEnabled": {
                 "value": "false"
+              },
+              "experimentalFeatures:mqttBrokerEnabled": {
+                "value": "true"
+              },
+              "RUST_LOG": {
+                "value": "true"
```
dmolokanov, line 53:
```suggestion
                "value": "debug"
```
I guess you wanted to put some log level here? In addition broker and bridge support the same ENV to configure log level as EdgeHub core `RuntimeLogLevel`

yophilav: yeah. this is not a real PR though. I'm using them for testing. I think something underwent incorrectly with an automerge

PR #3737, opened 2020-10-16
Fix metrics upload

lfitchett:
One of the changes to edgeAgent cause the SendEventBatchAsync function of the module sdk to stop working correctly. This pr removes the call to SendEventBatchAsync and replaces it with multiple calls to SendEventAsync. 

The main difference between the 2 is that SendEventBatchAsync is transactional, while iterative SendEventAsync is not. Because the metrics worker will do a full retry on upload failure, this means that if the metrics upload fails partway through a batch, metrics could be duplicated. I am not worried about this in practice, because:
* Metrics being batched should be fairly rare, only happening when a device is offline for an extended period of time (>3 days)
* In addition to batching being rare, the upload still needs to fail partway through a batch for data to be duplicated.
* Any duplicate data is easy to de-dupe since it has a unique instance guid and timestamp


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.IoTHub/EdgeAgentConnection.cs
```diff
@@ -128,7 +128,29 @@ public async Task UpdateReportedPropertiesAsync(TwinCollection patch)
             }
         }
 
-        public async Task SendEventBatchAsync(IEnumerable<Message> messages)
+        //// public async Task SendEventBatchAsync(IEnumerable<Message> messages)
```
damonbarry, line 131: Can you remove all the old code instead of commenting it out? It's in the commit history...

lfitchett: I'll remove this line, I'm going to leave the boilerplate in the connection handler since we did the same for device streams.

damonbarry: There's commented-out code across 6 source files...do those files comprise the connection handler?

damonbarry (approved):

PR #3739, opened 2020-10-16
Disabling all flaky tests in CI

dylanbronson:
Disabling all flaky tests so we can get CI to green. 

CI is basically useless if it's not green all the time. Let's get it to green so we can use it with more confidence.
I made bugs for all the tests that have been disabled. It's up to us to prioritize the fixes.


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/storage/MessageStoreTest.cs
```diff
@@ -229,7 +229,7 @@ public async Task CleanupTestCheckpointed(bool checkEntireQueueOnCleanup)
 
         [Theory]
         [InlineData(false)]
-        [InlineData(true)]
+        // [InlineData(true)] TODO: Investigate why this is failing re-enable
```
yophilav, line 232: Disable this test for both cases until the issue is fixed. I think you can use `Assert.Ignore("Disabling to unblock CI while we investigate") ` here

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.IoTHub.Test/EdgeAgentConnectionTest.cs
```diff
@@ -252,7 +252,7 @@ static IEdgeAgentConnection CreateEdgeAgentConnection(IotHubConnectionStringBuil
         }
 
         [Integration]
-        [Fact]
+        [Fact(Skip = "Disabling to unblock CI while we investigate")]
```
yophilav, line 255: By using the `[Fact(Skip = "")]` and `[Theory(Skip = "")]`, Are the skipping test case being count towards the overall `Skipped test cases ` generated by the dotnet report at the end?

dylanbronson: Yes, I believe so

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/storage/MessageStoreTest.cs
dylanbronson: Why both? If one works consistently, can we keep it to give us confidence?

yophilav: "False confidence". We know one of the scenario is failing, but we don't know why. How can we be sure the test is not giving use a false negative?

yophilav (dismissed):

yophilav (approved):

PR #3740, opened 2020-10-16
[mqtt] Bridge in flight management

and-rewsmith:
We need to manage the amount of in flight messages for the local/remote bridge clients. As discussed with Denis and Varun, I added a wrapper around the mqtt3 publish handle that controls the amount of in flight messages with a semaphore. This will allow us to more easily swap out the client if we need to while also not having to re-implement in-flight everywhere we need to publish.


mqtt/mqtt-bridge/src/client.rs
```diff
@@ -184,6 +186,36 @@ impl BridgeIoSource {
...
+    fn new(publish_handle: PublishHandle, permits: Arc<Semaphore>) -> Self {
+        Self {
+            publish_handle,
+            permits,
+        }
+    }
+
+    pub async fn publish(&self, publication: Publication) {
+        let mut publish_handle = self.publish_handle.clone();
+        let permits = self.permits.clone();
```
and-rewsmith, line 205: We can't pass once acquired because it contains a reference to the original semaphore with a lifetime 'a.

and-rewsmith:
I think we need some integration tests to verify this.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -267,10 +271,8 @@ impl Pump {
                         debug!("extracted publication from store");
 
                         if let Ok(Some((key, publication))) = loaded_element {
-                            debug!("publishing {:?}", key);
-                            if let Err(e) = publish_handle.publish(publication).await {
-                                error!(err = %e, "failed publish");
-                            }
+                            debug!("publishing publication {:?}", key);
+                            publish_handle.publish(publication).await;
```
and-rewsmith, line 287: We now log an error internally. Mqtt3 will retry itself right? If it fails this retry limit and errors then I feel the correct approach is to not have our own retry logic here, but just log the error.

mqtt/mqtt-bridge/src/pump.rs
```diff
@@ -265,7 +269,7 @@ impl Pump {
             info!("starting egress publication processing");
 
             loop {
-                let mut publish_handle = publish_handle.clone();
+                let publish_handle = publish_handle.clone();
                 match select(receive_fut.next(), loader.try_next()).await {
                     Either::Left((shutdown, _)) => {
```
and-rewsmith, line 274: In the shutdown case we will need a way to drain the in-flight messages. I know this structure will change with Denis incoming PR. I will look if I can do it in a way that will integrate nicely with his changes.

and-rewsmith: I think what we can do is create a FuturesUnordered collection in the egress loop that will keep polling for messages to be sent. Then in the egress loop we can select on three things: shutdown, loader.next(), futures unordered stream for sent messages for logging.

dmolokanov: What will happen if we call client_handle.shutdown()? Will that cause client to drain all in-flight messages? If yes, all we need to do is to wait while client is stopped.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -184,6 +186,36 @@ impl BridgeIoSource {
...
+
+        let fut = async move {
+            let permit = permits.acquire().await;
+            if let Err(e) = publish_handle.publish(publication).await {
+                error!(message = "failed to publish", err = %e);
+            }
+            drop(permit)
```
dmolokanov, line 212: I guess you don't need to drop that manually, because the scope ends right on the next line

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -184,6 +186,36 @@ impl BridgeIoSource {
...
+        };
+
+        tokio::spawn(fut);
```
dmolokanov, line 215: why do we need spawn here?

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -281,13 +319,16 @@ impl<H: EventHandler> MqttClient<H> {
             keep_alive,
         );
 
+        let in_flight = Arc::new(Semaphore::new(max_in_flight));
```
dmolokanov, line 322: I think instead of storing semaphore, it can store InFlighPublishHandle. Return self.publish_handle copy in the `fn publish_handle`()`

dmolokanov (changes requested):
I think we need to discuss the way we drain inflihgt messages.

mqtt/mqtt-bridge/src/client.rs
and-rewsmith: We need to spawn so that a separate task can wait for the permit and the original call to publish can complete. That way the caller can keep calling publish without having to manage all the futures. 

For the in-flight draining though, I think what we want to do is return a join handle from this future so we can confirm all the in-flights are complete at time of shutdown.

mqtt/mqtt-bridge/src/client.rs
and-rewsmith: Creating the client's publish_handle can error. It might be better to see this potential error every time we get the client wrapper's publish handle. 

I don't think it will matter though. Even when we recreate the pump it will recreate the client and only get the publish handle once at the beginning.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -184,6 +186,36 @@ impl BridgeIoSource {
...
+            permits,
+        }
+    }
+
+    pub async fn publish(&self, publication: Publication) {
+        let mut publish_handle = self.publish_handle.clone();
+        let permits = self.permits.clone();
+
+        let fut = async move {
+            let permit = permits.acquire().await;
```
and-rewsmith, line 208: I think this needs to be outside the async block. If it stays here then all calls to publish will spawn a new task waiting for a permit. Unfortunately, I think that this is hard because of the above-mentioned lifetime issue.

mqtt/mqtt-bridge/src/pump.rs
and-rewsmith: We discussed offline and decided since we are QOS1 then we can disregard draining all the in-flight messages. They might get sent twice but that is fine.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -458,16 +458,147 @@ impl ShutdownHandle {
...
+}
+
+mock! {
+    pub PublishHandle {}
+
+    #[async_trait]
+    pub trait InnerPublishHandle {
+        async fn publish(&mut self, publication: Publication) -> Result<(), PublishError>;
+    }
+}
```
and-rewsmith, line 535: I tried using automock, but there is weird behavior with the test where it thinks the automock is for InFlightPublishHandle. Having a more explicit mock fixes this error.

mqtt/mqttd/src/main.rs
```diff
@@ -1,4 +1,4 @@
-#![type_length_limit = "1230974"]
+#![type_length_limit = "2093492"]
```
and-rewsmith, line 1: Bumping this limit, but not sure when it becomes an issue

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -427,10 +427,10 @@ impl<H> MqttClientExt for MqttClient<H> {
 #[cfg(test)]
 /// Implements `MqttClientExt` for tests.
 impl<H> MqttClientExt for MqttClient<H> {
-    type PublishHandle = MockPublishHandle;
+    type PublishHandle = InFlightPublishHandle<MockallPublishHandleWrapper>;
 
-    fn publish_handle(&self) -> Result<Self::PublishHandle, ClientError> {
-        Ok(MockPublishHandle::new())
+    fn publish_handle(&self) -> Self::PublishHandle {
+        InFlightPublishHandle::new(MockallPublishHandleWrapper::new(), 100)
```
and-rewsmith, line 433: Since the tests that use this don't test for in flight, I am making the in-flight very large so it will never block. We could avoid hardcoding this if we store a semaphore and create the publish handle each time. However, Denis thought it was cleaner if we create the publish handle once when creating the client.

mqtt/mqtt-bridge/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![recursion_limit = "256"]
```
and-rewsmith, line 1: Guessing this is acceptable?

mqtt/mqtt-bridge/src/pump/egress.rs
```diff
@@ -60,27 +64,37 @@ where
...
+                        debug!("scheduling publish for publication {:?}", key);
+                        let publish_and_remove = async move {
+                            debug!("publishing publication {:?}", key);
+                            if let Err(e) = publish_fut.await {
+                                error!(err = %e, "failed sending publication {:?}", key);
+                            }
+
+                            if let Err(e) = store.remove(key) {
+                                error!(err = %e, "failed removing publication from store {:?}", key);
+                            }
```
and-rewsmith, line 91: We can move this store removal to the futures_unordered.next() select arm. Should I do it?

and-rewsmith:
Closing as this approach is dangerous. New PR #3797

PR #3741, opened 2020-10-16
[mqtt] Logging improvements

and-rewsmith:
While testing I noticed we had some logs that looked messy / inconsistent


mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -91,7 +91,7 @@ impl Bridge {
         pin_mut!(local_pump, remote_pump);
 
         debug!(
-            "Starting pumps for {} bridge...",
```
and-rewsmith, line 94: The pattern I have observed is that it should only be capital if more than one sentence.

ancaantochi (approved):

PR #3742, opened 2020-10-16
[mqtt] upstream bridge event handlers

dmolokanov:
`Pump` suppose to support several activities running in parallel: 
* Ingress - a reading publication from store and publishing using MQTT client
* Egress - a receiving publication from the MQTT client and put it in the opposite pump store
* Messages Processor - reacting on messages coming to the `Pump`

Pump messages can be different for different pumps (local/remote, upstream/east-west). 
`PumpMessageHandler` addressing this variety of options.

This PR includes `Connectivity` message functionality


mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+        loader: MessageLoader<WakingMemoryStore>,
+        store: PublicationStore<WakingMemoryStore>,
+    ) -> Result<Self, BridgeError> {
+        let client_shutdown = client.shutdown_handle()?;
+        let publish_handle = client
+            .publish_handle()
+            .map_err(BridgeError::PublishHandle)?;
+
+        let egress = Egress::new(publish_handle, loader, store);
+        let ingress = Ingress::new(client, client_shutdown);
```
ancaantochi, line 91: I think ingress client should have a list of EventHandlers and process the messages, one of the handlers will be the connectivity

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+}
+
+struct MessagesProcessor {
+    messages: mpsc::Receiver<PumpMessage>,
+    handle: Option<PumpHandle>,
+}
+
+impl MessagesProcessor {
+    fn new(messages: mpsc::Receiver<PumpMessage>, handle: PumpHandle) -> Self {
```
ancaantochi, line 166: messages processor needs a publish handle, at least for connectivity it needs to be able to send a message to the broker

mqtt/mqtt-bridge/src/pump/mod.rs
dmolokanov: It's going to be in MessagesProcessor

mqtt/mqtt-bridge/src/pump/mod.rs
dmolokanov: correct, I added the PumpMessageHandler trait. So you need to pass **another** PumpHandle in order this to work

mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -0,0 +1,161 @@
...
+        Self {
+            local: PumpBuilder::default(),
+            remote: PumpBuilder::default(),
+            store: Box::new(|| PublicationStore::new_memory(0)),
+        }
+    }
+}
+
+impl Builder {
+    pub fn with_local<F>(&mut self, mut apply: F) -> &mut Self
```
ancaantochi, line 44: is there a case when the pump doesn't have both local and remote? I tough the with_ is used when there are cases when you can construct with/without

mqtt/mqtt-bridge/src/pump/mod.rs
ancaantochi: Not clear why in message processor, I though that's used for messages between pumps. The eventhandler is to process messages from ingress pumo

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -67,12 +67,12 @@ where
     T: TokenSource + Clone + Send + Sync + 'static,
 {
     pub fn new(
-        address: String,
+        address: impl Into<String>,
```
ancaantochi, line 70: what is the advantage of this change?

mqtt/mqtt-bridge/src/pump/mod.rs
ancaantochi: nvm, I think I missed last change, I think it makes sense

mqtt/mqtt-bridge/src/pump/builder.rs
dmolokanov: no, there has to be both local and remote right now. there are some `.expect()` calls in the build method which will panic in runtime. not super rusty though...

mqtt/mqtt-bridge/src/client.rs
dmolokanov:
```rust
// if the function looks like this 
fn foo(bar: impl Into<String>) {..}

// it is convenient to use it the same way with both in production code
let bar: String = ..;
foo(bar);

// and in tests for example identically without .to_string()/.to_owned()/.into()
foo("hello");
```

mqtt/mqtt-bridge/src/pump/builder.rs
ancaantochi: Then it means it should create them in the constructor instead?

mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -0,0 +1,161 @@
...
+            messages,
+        )?;
+
+        let (subscriptions, topic_filters) = make_topics(&self.remote.rules)?;
+
+        let rpc = RemoteRpcHandler;
+        let messages = MessageHandler::new(egress_store, topic_filters);
+        let handler = RemoteUpstreamHandler::new(messages, rpc);
+
+        let config = self.local.client.take().expect("local client config");
```
ancaantochi, line 106: shouldn't this be self.remote.client ?

mqtt/mqtt-bridge/src/upstream/rpc/local.rs
```diff
@@ -0,0 +1,203 @@
...
+
+use crate::{
+    client::{EventHandler, Handled},
+    pump::{PumpHandle, PumpMessage},
+    upstream::{CommandId, RemoteUpstreamPumpEvent, RpcCommand},
+};
+
+use super::RpcError;
+
+/// An RPC handlers that reposible to connect to part of the bridge which
```
and-rewsmith, line 17: typo responsible

mqtt/mqtt-bridge/src/upstream/rpc/local.rs
```diff
@@ -0,0 +1,203 @@
...
+
+#[async_trait]
+impl EventHandler for LocalRpcHandler {
+    type Error = RpcError;
+
+    async fn handle(&mut self, event: &Event) -> Result<Handled, Self::Error> {
+        if let Event::Publication(publication) = event {
+            if let Some(command_id) = capture_command_id(&publication.topic_name) {
+                let doc = Document::from_reader(&mut publication.payload.clone().reader())?;
+                match bson::from_document(doc)? {
```
and-rewsmith, line 38: ```if let``` instead of match? Guessing there will be more though.

mqtt/mqtt-bridge/src/upstream/rpc/local.rs
```diff
@@ -0,0 +1,203 @@
...
+    use bson::{bson, spec::BinarySubtype};
+    use bytes::Bytes;
+    use matches::assert_matches;
+    use mqtt3::{proto::QoS, ReceivedPublication};
+    use test_case::test_case;
+
+    use super::*;
+
+    #[test]
+    fn it_deserializes_from_bson() {
```
and-rewsmith, line 87: If you feel like adding it, maybe a test for parsing a publication.

mqtt/mqtt-bridge/src/upstream/rpc/remote.rs
```diff
@@ -205,164 +209,56 @@ impl RpcHandler {
...
+        //         if handled == subscriptions.len() {
+        //             return Ok(Handled::Fully);
+        //         } else {
+        //             return Ok(Handled::Partially);
+        //         }
+        //     }
+        //     _ => {}
+        // }
+
+        // Ok(Handled::Skipped)
```
and-rewsmith, line 247: This was commented because it was moved to LocalRpcHandler? What is the plan here?

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -34,72 +42,94 @@ impl BridgeShutdownHandle {
...
+            .with_remote(|pump| {
+                pump.with_config(MqttClientConfig::new(
+                    settings.address(),
+                    settings.keep_alive(),
+                    settings.clean_session(),
+                    settings.credentials().clone(),
+                ))
+                .with_rules(settings.subscriptions());
+            })
+            .with_store(|| PublicationStore::new_memory(BATCH_SIZE))
```
and-rewsmith, line 83: We will need multiple stores.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -34,72 +42,94 @@ impl BridgeShutdownHandle {
...
+            .build()?;
+
+        // TODO move susbscriptions into run method
```
and-rewsmith, line 86: What is the intention of this change? Without some change to the bridge controller, this will have the same behavior.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -34,72 +42,94 @@ impl BridgeShutdownHandle {
...
+            _ = local_pump => {
+                // TODO shutdown remote pump
+                // shutdown_handle.shutdown().await?;
             }
-            Either::Right(_) => {
-                shutdown_handle.shutdown().await?;
+
+            _ = remote_pump => {
+                // TODO shutdown local pump
+                // shutdown_handle.shutdown().await?;
```
and-rewsmith, line 132: So we have the ingress / egress shutdown handles internally within the pump, but don't yet have a exposed unifying shutdown handle to call the internal handles?

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+
+#[derive(Debug, thiserror::Error)]
+#[error("unable to send command to pump")]
+pub struct PumpError;
+
+/// Pump used to connect to either local broker or remote brokers (including the upstream edge device)
+/// It contains an mqtt client that connects to a local/remote broker
+/// After connection there are two simultaneous processes:
+/// 1) persist incoming messages into an ingress store to be used by another pump
+/// 2) publish outgoing messages from an egress store to the local/remote broker
```
and-rewsmith, line 66: We should add that the pump will also process messages

and-rewsmith:
Let me know if you want me to make sure the demo scenario still works.

mqtt/mqtt-bridge/src/bridge.rs
and-rewsmith: Nevermind, I forgot the other store is inside the client.

mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        self.store = Box::new(store);
+        self
+    }
+
+    pub fn build(&mut self) -> PumpsResult {
+        let ingress_store = (self.store)();
+        let ingress_loader = ingress_store.loader();
+
+        let egress_store = (self.store)();
```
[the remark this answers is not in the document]
and-rewsmith: Nevermind, forgot the other store is inside the client.

mqtt/mqtt-bridge/src/upstream/events/local.rs
```diff
@@ -0,0 +1,26 @@
...
+    upstream::{CommandId, ConnectivityState},
+};
+
+#[derive(Debug, PartialEq)]
+pub enum LocalUpstreamPumpEvent {
+    ConnectivityUpdate(ConnectivityState),
+    RpcAck(CommandId),
+}
+
+pub struct LocalUpstreamPumpEventHandler;
```
and-rewsmith, line 14: Can we add doc comments to say what this does?

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+
+#[async_trait]
+pub trait PumpMessageHandler {
+    type Message;
+
+    async fn handle(&self, message: Self::Message);
+}
+
+struct MessagesProcessor<M>
```
and-rewsmith, line 170: Recommend adding doc comments

mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -0,0 +1,161 @@
...
+
+        let rpc = LocalRpcHandler::new(PumpHandle::new(remote_messages_send.clone()));
+        let messages = MessageHandler::new(ingress_store.clone(), topic_filters);
+        let handler = LocalUpstreamHandler::new(messages, rpc);
+
+        let config = self.local.client.take().expect("local client config");
+        let client = MqttClient::tls(config, handler);
+
+        let handler = LocalUpstreamPumpEventHandler;
+        let pump_handle = PumpHandle::new(local_messages_send.clone());
```
and-rewsmith, line 88: remote_messages_send?

mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -0,0 +1,161 @@
...
+        let messages = MessagesProcessor::new(handler, local_messages_recv, pump_handle);
```
and-rewsmith, line 89: I am confused on how this works. Each pump has a message processor and a client. They both have their own internal EventHandler implementations.

For the local pump, its Client EventHandlers will:
- persist messages
- wait for incoming publications and send a PumpMessage to the remote message handler

For the local pump, its MessageProcessor EventHandlers will:
- ?

For the remote pump, its Client EventHandlers will:
- only persist messages

For the remote pump, its MessageProcessor EventHandlers will:
- ?

mqtt/mqtt-bridge/src/pump/builder.rs
and-rewsmith, line 88: This confused me at first, but the reason it takes the local_messages_send here is for shutdown. I suppose that is fine, but it wasn't obvious to me.

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+
+        let mut shutdown = shutdown_recv.fuse();
+        let mut loader = loader.fuse();
+
+        loop {
+            select! {
+                _ = &mut shutdown => {
+                    info!(" received shutdown signal for egress messages");
+                    break;
+                }
```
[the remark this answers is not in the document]
and-rewsmith: I was thinking of using futures unordered collection to track all in flight messages so that they can be drained. I think we should discuss this approach.

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -248,30 +267,27 @@ pub struct TopicRule {
...
+        self.in_prefix.as_deref()
     }
 }
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 #[serde(tag = "direction")]
 pub enum Direction {
     #[serde(rename = "in")]
-    In,
+    In(TopicRule),
```
vadim-kovalyov, line 287: :+1:

mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -0,0 +1,161 @@
...
+        let (local_messages_send, local_messages_recv) = mpsc::channel(100);
+
+        let (subscriptions, topic_filters) = make_topics(&self.local.rules)?;
+
+        let rpc = LocalRpcHandler::new(PumpHandle::new(remote_messages_send.clone()));
+        let messages = MessageHandler::new(ingress_store.clone(), topic_filters);
+        let handler = LocalUpstreamHandler::new(messages, rpc);
+
+        let config = self.local.client.take().expect("local client config");
+        let client = MqttClient::tls(config, handler);
```
vadim-kovalyov, line 85: does local client have to be tls?

mqtt/mqtt-bridge/src/upstream/rpc/remote.rs
```diff
@@ -32,31 +32,31 @@ use crate::client::{EventHandler, Handled};
 /// sends corresponding MQTT packet to upstream broker and waits for an ack
 /// from the upstream. After ack is received it sends a special publish to
 /// downstream broker.
-pub struct RpcHandler {
-    upstream_subs: UpdateSubscriptionHandle,
-    upstream_pubs: PublishHandle,
+pub struct RemoteRpcHandler {
+    subsription: UpdateSubscriptionHandle,
```
vadim-kovalyov, line 36:
```suggestion
    subscription: UpdateSubscriptionHandle,
```

vadim-kovalyov (commented):
LGTM overall, although I'm not in context of the logic

mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    pub fn build(&mut self) -> PumpsResult {
+        let ingress_store = (self.store)();
+        let ingress_loader = ingress_store.loader();
+
+        let egress_store = (self.store)();
+        let egress_loader = egress_store.loader();
+
+        let (remote_messages_send, remote_messages_recv) = mpsc::channel(100);
```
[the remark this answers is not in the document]
dmolokanov: just a random number actually. how many events concurrent we anticipate?

mqtt/mqtt-bridge/src/pump/builder.rs
dmolokanov: I saw you fixed that in your PR

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    pub async fn send(&mut self, message: PumpMessage<M>) -> Result<(), PumpError> {
+        self.sender.send(message).await.map_err(|_| PumpError)
+    }
+}
+
+#[cfg(test)]
+pub fn channel<M>() -> (PumpHandle<M>, mpsc::Receiver<PumpMessage<M>>) {
+    let (tx, rx) = tokio::sync::mpsc::channel(10);
```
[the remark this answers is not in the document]
dmolokanov: some random number for tests

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+                shutdown_ingress.shutdown().await;
+                shutdown_egress.shutdown().await;
+            }
+        };
+
+        info!("stopped pump");
+    }
+}
+
+struct MessagesProcessorShutdownHandle<M>(Option<PumpHandle<M>>);
```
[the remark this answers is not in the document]
dmolokanov: Once you use it, you don't need it. oneshot-like approach

mqtt/mqtt-bridge/src/pump/mod.rs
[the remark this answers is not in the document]
dmolokanov: Yeah, lets discuss

mqtt/mqtt-bridge/src/upstream/connectivity.rs
```diff
@@ -37,11 +55,13 @@ impl EventHandler for ConnectivityHandler {
...
                         self.state = ConnectivityState::Disconnected;
-                        self.sender
-                            .send(PumpMessage::ConnectivityUpdate(
-                                ConnectivityState::Disconnected,
-                            ))
-                            .await?;
+
+                        let event = LocalUpstreamPumpEvent::ConnectivityUpdate(
+                            ConnectivityState::Disconnected,
+                        );
```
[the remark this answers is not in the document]
dmolokanov: 1. remote client detects disconnect happened
2. remote upstream handler sends a ConnectivityUpdate event to a local pump
3. using publish_handle for a local client local pump sends retain MQTT message to a special topic
4. EHC receives it and does corresponding logic

mqtt/mqtt-bridge/src/upstream/mod.rs
```diff
@@ -0,0 +1,90 @@
+//! Module contains code related to upstream bridge.
```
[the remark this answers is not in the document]
dmolokanov: I hope it helps to define what code suppose to live in the module. Hover on some module name in vscode and it will give you some info

mqtt/mqtt-bridge/src/pump/builder.rs
dmolokanov: @ancaantochi / @and-rewsmith may have more context on this

mqtt/mqtt-bridge/src/bridge.rs
dmolokanov: the pump subscriptions can be changed with the twin update. so it has to be changed to react to changes. we discussed with @ancaantochi how we can achieve this with changes in her PR

mqtt/mqtt-bridge/src/bridge.rs
dmolokanov: we have `PumpMessage::Shutdown` event which I anticipated to be used to stop pumps. work to implement shutdown procedure and make tests is to be done.

mqtt/mqtt-bridge/src/pump/builder.rs
dmolokanov: Agree. Any ideas to improve that?

mqtt/mqtt-bridge/src/pump/builder.rs
dmolokanov: > I am confused on how this works. Each pump has a message processor and a client. They both have their own internal EventHandler implementations.
> 
> For the local pump, its Client EventHandlers will:
> 
>     * persist messages
> 
>     * wait for incoming publications and send a PumpMessage to the remote message handler
> 
> 
> For the local pump, its MessageProcessor EventHandlers will:
> 
* Receive `ConnectivityUpdate` event and publish corresponding MQTT message to EHC
* Receive RPC Ack event and publish corresponding MQTT message to EHC
* Receive RPC Nack (negative ack) event and publish corresponding MQTT message to EHC
> 
> 
> For the remote pump, its Client EventHandlers will:
> 
>     * only persist messages
> 
> 
> For the remote pump, its MessageProcessor EventHandlers will:
> 
* Receive `RpcCommand` and make corresponding action against remote broker (pub, sub, unsub)
> 
> 
> What will these MessageProcessors do? I think doc comments would be helpful.

mqtt/mqtt-bridge/src/upstream/rpc/remote.rs
dmolokanov: yeah, this module is to be implemented. part of the logic moved to LocalRpcHandler and the rest will be in RemoteRpcHandler

mqtt/mqtt-bridge/src/pump/builder.rs
and-rewsmith: local client should be tcp

mqtt/mqtt-bridge/src/pump/builder.rs
and-rewsmith: Wait so are we going to fix this in this PR? The pump won't work otherwise.

mqtt/mqtt-bridge/src/upstream/rpc/local.rs
dmolokanov: iirc clippy was a bit opionated on the structure

mqtt/mqtt-bridge/src/upstream/rpc/local.rs
dmolokanov: I believe `it_handles_rpc_commands` tests is all about parsing publication and making a pump message

mqtt/mqtt-bridge/src/pump/builder.rs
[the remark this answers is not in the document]
ancaantochi: not sure, I think we can decide after we do some testing if it needs to be adjusted

ancaantochi (dismissed):
LGTM, just needs the change of local client to use tcp.

mqtt/mqtt-bridge/src/pump/builder.rs
dmolokanov: we merged @ancaantochi changes into this PR

mqtt/mqtt-bridge/src/pump/builder.rs
and-rewsmith: With the current design I don't see an easy way. Maybe elevate the shutdown handle.

and-rewsmith (approved):

ancaantochi (approved):

PR #3743, opened 2020-10-16
[k8s] Discovered some problems with createOptions parsing.

darobs:
Two problems were discovered today:

1. cmd, entrypoint, workingdir should be looked up in a case-insensitive manner.
The code extracting these settings was strictly using the capitalization in Docker spec. (`Cmd`, `EntryPoint`, `WorkingDir`).

2. environment variables in createOptions should allow "=" inside the value.
Environment variables in Docker are allowed to have "=" in them, like this:
```sh
$ docker run -it --rm --env "A===" ubuntu:18.04 
root@d3303984b9fc:/# echo $A
==
root@d3303984b9fc:/# 
```
The code extracting environment variables is not doing this correctly.


dmolokanov (approved):
LGTM

PR #3745, opened 2020-10-17
arm build

huguesBouvier:
Using custom script to compile instead of relying on external source.

Test:
https://dev.azure.com/msazure/One/_build/results?buildId=35864771&view=results


scripts/linux/buildAPIProxy.sh
```diff
@@ -151,11 +151,11 @@ build_project()
 {
     # build project with cross
     if [[ "$ARCH" == "amd64" ]]; then
-        execute scripts/linux/cross-platform-rust-build.sh --os ubuntu18.04 --arch "amd64" --build-path edge-modules/api-proxy-module
+        execute scripts/linux/cross-platform-rust-build.sh --os alpine1.19.3 --arch "amd64" --build-path edge-modules/api-proxy-module
```
and-rewsmith, line 154: Just curious, why do we need execute?

scripts/linux/cross-platform-rust-build.sh
```diff
@@ -206,39 +296,29 @@ case "$PACKAGE_OS.$PACKAGE_ARCH" in
...
-                MAKE_FLAGS="'CARGOFLAGS=$CARGOFLAGS --target aarch64-unknown-linux-gnu'"
-                MAKE_FLAGS="$MAKE_FLAGS 'TARGET=target/aarch64-unknown-linux-gnu/release'"
-                MAKE_FLAGS="$MAKE_FLAGS 'STRIP_COMMAND=/usr/aarch64-linux-gnu/bin/strip'"
-                ;;
-        esac
+MAKE_COMMAND="make release $MAKE_FLAGS"
 
-        MAKE_COMMAND="make release $MAKE_FLAGS"
-        ;;
-esac
```
and-rewsmith, line 241: Moving everything into one block might make things messy going forward, but I suppose it is fine for now.

scripts/linux/cross-platform-rust-build.sh
```diff
@@ -253,12 +333,7 @@ docker run --rm \
         cat /etc/os-release &&
 
         $SETUP_COMMAND
-
-        cd /project/$BUILD_PATH && 
-        echo 'Installing rustup' &&
-        curl -sSLf https://sh.rustup.rs | sh -s -- -y &&
-        . ~/.cargo/env &&
```
and-rewsmith, line 260: Do we not need this? I think this will install the latest version of rust so it was needed even if the build agents have rust installed.

scripts/linux/cross-platform-rust-build.sh
```diff
@@ -70,34 +70,20 @@ case "$PACKAGE_OS" in
...
-esac
-
-if [ -n "$RUST_TARGET" ]; then
-    RUST_TARGET_COMMAND="rustup target add $RUST_TARGET &&"
-fi
-
-
 case "$PACKAGE_OS.$PACKAGE_ARCH" in
-    ubuntu18.04.amd64)
+    ubuntu18.04.amd64 | alpine1.19.3.amd64)
```
and-rewsmith, line 85: Since we aren't building for ubuntu18.04, I think we should change it to alpine in our build scripts.

and-rewsmith (commented):
If we are fine with eliminating rust install and we change all the amd64 build script calls to alpine, this LGTM

scripts/linux/buildAPIProxy.sh
huguesBouvier: We need to build the API proxy executable. API proxy consists of both nginx proxy and a rust executable (that monitors twins, get certs from worload API etc...) . Here we build the rust executable. The target is alpine, hence the change.

scripts/linux/cross-platform-rust-build.sh
huguesBouvier: I moved it inside each image.  The script to install for arm is slighty different and needs curl -sSLf https://sh.rustup.rs to be installed BEFORE $SETUP_COMMAND runs.
However I cannot really put it before (line 254, because curl has not been installed yet. The only solution was to move it inside $SETUP_COMMAND.
You can still see it line: 230 or 138 for example.

scripts/linux/cross-platform-rust-build.sh
huguesBouvier: Possibly yes. For now we only have alpine, but it would not be too hard to break it down in smaller files.

scripts/linux/buildAPIProxy.sh
and-rewsmith: Oh sorry I was asking about the execute keyword. It doesn't matter though.

scripts/linux/cross-platform-rust-build.sh
huguesBouvier: I agree but this close to the release I don't feel confident changing to 100 different places :X

scripts/linux/cross-platform-rust-build.sh
and-rewsmith: I am confused because nothing changed with our arm builds and wasn't it working?

and-rewsmith (approved):

PR #3750, opened 2020-10-19
Policy ready event + fine tuning of log messages.

vadim-kovalyov:
- Added `PolicyReady` event to indicate that authorization policy engine is ready
- A bugfix for deserializing policy update.
- Log messages fine tuning

### DIFF
--- a/mqtt/mqtt-broker/src/ready.rs
+++ b/mqtt/mqtt-broker/src/ready.rs
@@ -3,7 +3,7 @@ use std::{collections::HashSet, fmt::Display, hash::Hash};
 use tokio::sync::broadcast::{self, Receiver, RecvError, Sender};
 use tracing::{debug, error, warn};
 
-/// `BrokerReady` component acts as a intermidiate entity to determine whether
+/// `BrokerReady` component acts as a intermediate entity to determine whether
 /// `Broker` is ready to serve external clients. It awaits on several events
 /// from components Broker depends on.
 ///
@@ -45,7 +45,7 @@ where
     E: Clone + Display,
 {
     pub fn send(&mut self, event: E) {
-        debug!("sending broker ready event {}", event);
+        debug!("sending broker ready event: {}", event);
         if self.0.send(event).is_err() {
             error!("no active receiver found")
         }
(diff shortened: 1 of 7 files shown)


mqtt/mqtt-broker/src/ready.rs
```diff
@@ -95,7 +95,10 @@ pub trait AwaitingEvents {
 /// Broker readiness events.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum BrokerReadyEvent {
+    /// Signals that `EdgeHubAuthorizer` is ready.
     AuthorizerReady,
+    /// Signals that `PolicyAuthorizer` is ready.
+    PolicyReady,
```
dmolokanov, line 101: What is the difference between `PolicyReady` and `AuthorizerReady`? Should we combine it into one event because they both related to the authorizer?

dmolokanov (commented):
Could you please add a description to PR?

mqtt/mqtt-broker/src/ready.rs
vadim-kovalyov: Those are two distinct messages coming from EdgeHub. Later we may thing about optimizing this, but for now we need two events.

dmolokanov: Help me to understand the full flow of how these events originated in EHC and pushed to the broker. 
It starts with EdgeHub received its own twin (from storage or via wire). Then it publishes several messages to broker: 1) policy update; 2) configuration update; 3) ...?  
In parallel EdgeHub makes some call to obtain child devices scopes? When it receives them it publishes separate messages with device scopes. That's why we cannot get away with only one event right now?

vadim-kovalyov: yes, policy and bridge config comes from the twin. Device scopes (authorized identities) comes from rest call to the IoT Hub. Those components can be available at different moments in time, so we can't combine them atm.

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -263,15 +263,15 @@ where
...
                     .0
                     .into_iter()
                     .map(|id| (id.identity().into(), id))
                     .collect();
 
+                info!("edgehub authorizer has been updated.");
+
                 // signal that authorizer has been initialized
                 if let Some(mut broker_ready) = self.broker_ready.take() {
                     broker_ready.send(BrokerReadyEvent::AuthorizerReady);
```
dmolokanov, line 277: Could you please add debug messages, that signal has been sent to broker ready component? It will help to investigate issues if broker cannot serve external clients too long

dmolokanov (approved):
:shipit:

PR #3754, opened 2020-10-19
Disable persist proptests on arm32

vadim-kovalyov:
Those tests are running very slow on arm32 and often timeout.


dmolokanov (approved):
I wonder if even need these tests to be proptests?

PR #3766, opened 2020-10-21
Update bridge config from EdgeHub twin

vadim-kovalyov:
Bridge config for MQTT broker would come from EdgeHub twin. In this PR I have:
- added domain model for bridge config.
- added validation for the bridge config.

TODO:
- [x] finish tests


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BrokerConfigValidator.cs
```diff
@@ -58,7 +58,55 @@ public virtual IList<string> ValidateAuthorizationConfig(AuthorizationProperties
...
+
+        static void ValidateBridge(Bridge bridge, int order, List<string> errors)
+        {
+            if (string.IsNullOrEmpty(bridge.Endpoint))
+            {
+                errors.Add($"Bridge {order}: Endpoint must not be empty");
+            }
+
+            foreach (var setting in bridge.Settings)
+            {
```
vipeller, line 89: does the deserialization always resolve at least an empty list for bridge.Settings?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BrokerConfigValidator.cs
```diff
@@ -58,7 +58,55 @@ public virtual IList<string> ValidateAuthorizationConfig(AuthorizationProperties
...
+                if (!IsValidTopicFilter(setting.Topic))
+                {
```
vipeller, line 91: is 'topic' at least an empty string? (or IsvalidTopicFilter can handle null if deserializes to null?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubConfigParser.cs
```diff
@@ -118,7 +119,14 @@ Option<AuthorizationConfig> ParseAuthorizationConfig(BrokerProperties properties
...
         Option<BridgeConfig> ParseBridgeConfig(BrokerProperties properties)
         {
-            return Option.None<BridgeConfig>();
+            IList<string> errors = this.validator.ValidateBridgeConfig(properties.Bridges);
+            if (errors.Count > 0)
+            {
+                string message = string.Join("; ", errors);
+                throw new InvalidOperationException($"Error validating bridge configuration: {message}");
+            }
+
```
vipeller, line 128: If it either throws an exception or returns Some, does it need to use Option?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BridgeConfig.cs
```diff
@@ -1,15 +1,158 @@
...
+        public Direction Direction { get; }
+
+        [JsonProperty("topic")]
+        public string Topic { get; }
+
+        [JsonProperty("inPrefix")]
+        public string InPrefix { get; }
+
+        [JsonProperty("outPrefix")]
+        public string OutPrefix { get; }
```
dylanbronson, line 116: What are these?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BridgeConfig.cs
```diff
@@ -1,15 +1,158 @@
...
+                hashCode = (hashCode * 397) ^ (this.OutPrefix?.GetHashCode() ?? 0);
+                return hashCode;
+            }
+        }
+    }
+
+    public enum Direction
+    {
+        In,
+        Out,
```
dylanbronson, line 155: Why is this not up/down?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BrokerConfigValidator.cs
```diff
@@ -58,7 +58,55 @@ public virtual IList<string> ValidateAuthorizationConfig(AuthorizationProperties
...
+        ///
+        /// Validates bridge config and returns a list of errors (if any).
+        /// </summary>
+        public virtual IList<string> ValidateBridgeConfig(BridgeConfig properties)
+        {
+            Preconditions.CheckNotNull(properties, nameof(properties));
+
+            var order = 0;
+            var errors = new List<string>();
+            foreach (var bridge in properties)
```
dylanbronson, line 72: There are multiple bridges in the properties? Is it the whole hierarchy of bridges? How does this work?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BrokerConfigValidator.cs
vadim-kovalyov: good point. I marked some properties as required and added default values otherwise.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BrokerConfigValidator.cs
vadim-kovalyov: :+1: Added check for null

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubConfigParser.cs
vadim-kovalyov: There are 3 cases here: 
- settings present (Some)
- settings absent (None)
- validation error (exception)

I've added a null/empty check to return handle none case.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BridgeConfig.cs
vadim-kovalyov: discussed offline

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BridgeConfig.cs
```diff
@@ -1,15 +1,158 @@
...
+        public Bridge(string endpoint, IList<Settings> settings)
+        {
+            this.Endpoint = endpoint;
+            this.Settings = settings;
+        }
+
+        [JsonProperty("endpoint", Required = Required.Always)]
+        public string Endpoint { get; }
+
+        [JsonProperty("settings", Required = Required.Always)]
```
ancaantochi, line 58: In the doc with the schema changed I asked if we should actually use subscriptions or topics because we might want to add other settings which are not related to topics. Let’s sync about which name to use

ancaantochi (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/BridgeConfig.cs
vadim-kovalyov: Discussed offline. We can change it later if needed.

PR #3767, opened 2020-10-21
Fixing ValidateMetrics E2E

yophilav:
Currently the `ValidateMetrics` parses [metric readme page](https://github.com/Azure/iotedge/blob/master/doc/BuiltInMetrics.md) to obtain all the metric keys. The test then deploy the system runtime modules and populate its metrics which then are compared against the parsed-document set.
There are a handful of not-so-positive metrics that are not being populated if the system modules operated on a happy path during the E2E test. These not-so-positive metrics are currently causing the `ValidateMetrics` to fail since they are not being populated.


test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
```diff
@@ -76,6 +76,28 @@ protected override async Task Test(CancellationToken cancellationToken)
                 }
             }
 
+            // The following metric should not be populated in a happy E2E path.
+            // We are going to make a list and remove them here to not consider them as a failure.
+            IEnumerable<string> acceptableDegeneratedMetrics = new HashSet<string>
+            {
+                // "edgeagent_direct_method_invocations_count", // BEARWASHERE -- Actually make this pass
```
damonbarry, line 83: Let's not check in any BEARWASHERE comments. ;)

Yes, it should be easy to force this metric by calling "ping" on Edge agent back in `Metrics.ValidateMetrics()`, something like:

```csharp
    var agent = new EdgeAgent(this.runtime.DeviceId, this.iotHub);
    await agent.PingAsync(token);
```

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
```diff
@@ -76,6 +76,28 @@ protected override async Task Test(CancellationToken cancellationToken)
...
+                "edgeAgent_metadata",
```
damonbarry, line 84: This one (`edgeAgent_metadata`) should already be showing up in the metrics. Maybe it's just a case mismatch or something silly like that? @lfitchett

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
```diff
@@ -76,6 +76,28 @@ protected override async Task Test(CancellationToken cancellationToken)
...
+                "edgeAgent_unsuccessful_iothub_syncs_total",
+                "edgehub_client_connect_failed_total",
+                "edgehub_messages_dropped_total",
+                "edgehub_messages_unack_total",
+                "edgehub_offline_count_total",
+                "edgehub_operation_retry_total"
+            };
+
+            foreach (string depopulatingMetric in acceptableDegeneratedMetrics)
```
damonbarry, line 93: We have a couple different confusing terms here (depopulating, degenerated)...can I suggest something more direct like "skipped"?

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
```diff
@@ -106,11 +128,15 @@ protected override async Task Test(CancellationToken cancellationToken)
...
+            string deviceId = Environment.GetEnvironmentVariable("IOTEDGE_DEVICEID");
+
             await this.moduleClient.SendEventAsync(new Message(Encoding.UTF8.GetBytes("Test message to seed metrics")), cancellationToken);
 
             const string methodName = "FakeDirectMethod";
             await this.moduleClient.SetMethodHandlerAsync(methodName, (_, __) => Task.FromResult(new MethodResponse(200)), null);
-            await this.moduleClient.InvokeMethodAsync(Environment.GetEnvironmentVariable("IOTEDGE_DEVICEID"), Environment.GetEnvironmentVariable("IOTEDGE_MODULEID"), new MethodRequest(methodName), cancellationToken);
+            await this.moduleClient.InvokeMethodAsync(deviceId, Environment.GetEnvironmentVariable("IOTEDGE_MODULEID"), new MethodRequest(methodName), cancellationToken);
+            // BEARWASHERE -- Lets see if you can ping a M2M dm to edgeAgent via edgeHub
+            await this.moduleClient.InvokeMethodAsync(deviceId, "$edgeAgent", new MethodRequest("ping"), cancellationToken);
```
damonbarry, line 139: This should work...did it? You could also do it from the test rather than the metricsValidator module; see my earlier comment.

Regardless, please remove the BEARWASHERE comment...

damonbarry (changes requested):

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
yophilav: Nope. This doesn't work.

builds/e2e/e2e.yaml
```diff
@@ -52,7 +52,8 @@ jobs:
     displayName: Linux amd64
 
     pool:
-      vmImage: ubuntu-16.04
+      name: $(pool.name)
+      demands: bearwashere
```
yophilav, line 56: I will come back to this after the test is done.

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
yophilav: Can confirm that works! Thank you!

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
yophilav: From the most recent test run, it seems like this metrics was not populated. 
Per \iotedge\doc\BuiltInMetrics.md, it seems like such matric can possibly not being populated if none of the flags mentioned are set.

yophilav:
I have run the E2E to verify the fix

test/modules/MetricsValidator/src/tests/ValidateDocumentedMetrics.cs
lfitchett: It should definitely be showing up. That metric is always on and we definitely want to verify it shows up

damonbarry (approved):

PR #3770, opened 2020-10-21
[mqtt] RPC command handler (part 2) 

dmolokanov:
Implemented a full flow of an RPC command via the following handlers:
1. `LocalRpcHandler` is an `EventHandler` that receives a publication with RPC command in the body. It deserializes command and sends it to `RemotePump` via mpsc channel.
2. `RemoteUpstreamPumpEventHandler` is an `PumpMessageHandler` that reads the RPC command from a channel. 
    - For`PUB` command it sends publication right away to upstream broker using `PublishHandle` and waits for the response. When a response received it sends a acknowledgemnt message to a `LocalPump`
    - For `SUB`/`UNSUB` commands it makes the corresponding request using `UpdateSubscriptionHandle`
3. `RemoteRpcHandle` is an `EventHandler` that receives a subscription update event. It filters those request which was made for RPC command and sends an acknowledgemnt message to a `LocalPump`
4. `LocalUpstreamPumpEventHandler` is an `PumpMessageHandler` that an acknowledgemnt message from mpsc channel and sends corresponding ACK/NACK message as MQTT publication using `PublishHandle` to local broker


mqtt/mqtt-bridge/src/upstream/events/remote.rs
```diff
@@ -16,15 +28,257 @@ pub enum RemoteUpstreamPumpEvent {
 /// It handles follwing events:
```
vadim-kovalyov, line 28:
```suggestion
/// It handles following events:
```

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -431,9 +522,12 @@ pub enum ClientError {
     #[error("failed to shutdown custom mqtt client: {0}")]
```
vadim-kovalyov, line 522: nit: update all error messages to include "{0}" where there is a #[from]

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -167,6 +167,9 @@ pub enum BridgeError {
     #[error("failed to get publish handle from client.")]
     PublishHandle(#[source] ClientError),
 
+    #[error("failed to get subscribe handle from client.")]
+    UpdateSubscriptionHandle(#[source] ClientError),
```
vadim-kovalyov, line 171: nit: add "{0}" to error messages with "source" or "from"

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -406,17 +378,136 @@ impl<H: EventHandler> MqttClient<H> {
     }
 }
 
+/// A trait extending `MqttClient` with additional handles functionality.
+pub trait MqttClientExt {
```
vadim-kovalyov, line 382: :+1: like this one

vadim-kovalyov (approved):
LGTM!

mqtt/mqtt-bridge/src/upstream/rpc/mod.rs
```diff
@@ -90,23 +88,36 @@ pub enum RpcCommand {
...
+/// It is shared due to `mqtt3::Client` implementation details when
+/// subscription has been made with `UpdateSubscriptionHandle` but the server
+/// response comes back as `mqtt3::Event` type which handled with the event
+/// handler.
+#[derive(Debug, Clone, Default)]
+pub struct RpcSubscriptions(Arc<Mutex<HashMap<String, CommandId>>>);
+
+impl RpcSubscriptions {
+    /// Stores topic filter to command identifier mapping.
+    pub fn insert(&self, id: CommandId, topic_filter: &str) -> Option<CommandId> {
```
ancaantochi, line 114: I think topic_filter should be before id since it is used as the key

mqtt/mqtt-bridge/src/client.rs
ancaantochi, line 382: nit: maybe call it BridgeClient or some other naming without Ext

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -406,17 +378,136 @@ impl<H: EventHandler> MqttClient<H> {
...
+    /// Publish handle type.
+    type PublishHandle;
+
+    /// Returns an instance of publish handle.
+    fn publish_handle(&self) -> Result<Self::PublishHandle, ClientError>;
+
+    /// Update subscription handle type.
+    type UpdateSubscriptionHandle;
```
ancaantochi, line 390: nit: name it SubscriptionHandle

mqtt/mqtt-bridge/src/client.rs
dmolokanov: Actually I was thinking about the opposite: remote "{0}" from each error. Wherever we need a full chain of error we should rely on the framework (tracing) to log with a full stack. What do you think?

mqtt/mqtt-bridge/src/upstream/events/local.rs
```diff
@@ -48,27 +57,133 @@ impl PumpMessageHandler for LocalUpstreamPumpEventHandler {
...
+                    }
+                }
+            }
+            LocalUpstreamPumpEvent::RpcAck(command_id) => {
+                debug!("sending rpc command ack {}", command_id);
 
-                let publication = Publication {
-                    topic_name: CONNECTIVITY_TOPIC.to_owned(),
+                Some(Publication {
+                    topic_name: format!("$edgehub/rpc/ack/{}", command_id),
```
ancaantochi, line 82: extract topic in const

mqtt/mqtt-bridge/src/client.rs
vadim-kovalyov: I don't think it works that way right now, or maybe we have wrong setup somewhere. In my manual tests I couldn't get the full error logs that's why I suggested to add {0} everywhere.

ancaantochi (approved):

PR #3773, opened 2020-10-22
env var name override

huguesBouvier:
Use new env variable name to override config


vadim-kovalyov (approved):

mqtt/mqtt-edgehub/src/settings.rs
```diff
@@ -216,6 +241,28 @@ mod tests {
 
     const DAYS: u64 = 24 * 60 * 60;
 
+    #[test]
```
vadim-kovalyov, line 244: @huguesBouvier , can you please add #[serial] to this test, so it won't interfere with others.

PR #3775, opened 2020-10-22
Fix Edge identity creation in X509 thumbprint E2E test

richma-ms:
The secondary thumbprint is mandatory when creating an X509 self-signed identity, but it seems that the iothub API implementation doesn't enforce it.  So the testcase ended up creating a non-compliant Edge device by not having both thumprints.  This causes initial cache refresh to fail when we parse the ScopeResults from iothub, and prevent the downstream device from connecting.


test/Microsoft.Azure.Devices.Edge.Test/helpers/X509ManualProvisioningFixture.cs
```diff
@@ -88,7 +88,8 @@ async Task<(X509Thumbprint, IdCertificates)> CreateIdentityCertAsync(string devi
 
             return (new X509Thumbprint()
             {
-                PrimaryThumbprint = deviceCert.Thumbprint
+                PrimaryThumbprint = deviceCert.Thumbprint,
+                SecondaryThumbprint = deviceCert.Thumbprint
```
yophilav, line 92: Should the two thumbprints be different? or can it be whatever?

yophilav (commented):
Should the two thumbprints be different?

test/Microsoft.Azure.Devices.Edge.Test/helpers/X509ManualProvisioningFixture.cs
richma-ms: they can be whatever, as long as one of them matches the client cert

yophilav (approved):

PR #3778, opened 2020-10-22
[mqtt] RPC command handler (part 3)

dmolokanov:
`RemoteRpcHandler` receives a new publication and checks if it is the one of twin response, twin desired properties response or direct method messages. In this case, it changes the topic prefix to `$downstream` and forwards it to the local broker.


vadim-kovalyov (approved):
LGTM!

mqtt/mqtt-bridge/src/upstream/events/local.rs
```diff
@@ -92,7 +95,7 @@ impl PumpMessageHandler for LocalUpstreamPumpEventHandler {
                 let doc = doc! { "reason": reason };
                 match doc.to_writer(&mut payload) {
                     Ok(_) => Some(Publication {
-                        topic_name: format!("$edgehub/rpc/nack/{}", command_id),
+                        topic_name: format!("$upstream/rpc/nack/{}", command_id),
```
and-rewsmith, line 98: What was the reason this changed to downstream?

dmolokanov: we discussed with @vipeller that it will cleaner if all requests that go to upstream broker have `$upstream` prefix, all responses `$downstream`. this way it is easier to reason about who is the recepient of the message.

dmolokanov: oh, actually you spotted an error here!

and-rewsmith: On the line for my comment we are sending a publish to $upstream only in the case of nack. This is a response so shouldn't it be $downstream?

and-rewsmith: You wrote the test to match this upstream / downstream pattern so I think I am missing something

and-rewsmith: Your comment didn't load for me. I see it now and understand.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -306,8 +306,8 @@ impl<H: EventHandler> MqttClient<H> {
             None
         }) {
             debug!("handling event {:?}", event);
-            if let Err(e) = self.event_handler.handle(&event).await {
-                error!(err = %e, "error processing event {:?}", event);
+            if let Err(e) = self.event_handler.handle(event).await {
+                error!(err = %e, "error processing event");
             }
```
and-rewsmith, line 311: We aren't doing anything with the Handled level here. I'm guessing we have it for tests, and in case we need to log it in the future?

mqtt/mqtt-bridge/src/upstream/rpc/remote.rs
```diff
@@ -71,28 +112,54 @@ impl RemoteRpcHandler {
...
+            format!(
+                "\\$iothub/{}/methods/post/(?P<params>.*)",
+                DEVICE_OR_MODULE_ID
+            )
+        ])
+        .expect("upstream topic patterns");
+    };
+
+    if UPSTREAM_TOPIC_PATTERNS.is_match(topic_name) {
+        Some(topic_name.replace("$iothub", "$downstream"))
```
and-rewsmith, line 160: Why are we replacing with downstream?

and-rewsmith (approved):
I think it is very confusing how we have upstream/events folder contain EventHandler prefixed type names but they don't implement EventHandler trait. Also, the types in upstream/rpc implement the EventHandler trait and are prefixed by RpcHandler. Since we are in a time crunch I think it is fine if we don't fix it in this PR.

and-rewsmith:
@dmolokanov I think there is a problem with master, so please don't merge this until we confirm that master is working or find the problem

mqtt/mqtt-bridge/src/upstream/rpc/remote.rs
```diff
@@ -71,28 +112,54 @@ impl RemoteRpcHandler {
 impl EventHandler for RemoteRpcHandler {
     type Error = RpcError;
 
-    async fn handle(&mut self, event: &Event) -> Result<Handled, Self::Error> {
-        match event {
-            Event::Publication(_) => {
-                // TODO implement incoming messages translation (C2D and M2M between EdgeHubs in nested edge scenario)
```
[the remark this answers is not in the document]
dmolokanov: I added implementation actually

mqtt/mqtt-bridge/src/client.rs
dmolokanov: It is logged before we consume it

mqtt/mqtt-bridge/src/upstream/rpc/remote.rs
dmolokanov: Per the initial design doc, events that come from upstream should have topic prefixed `$downstream`. It is done to avoid confusion when EHC doesn't know if this publication came from connected child clients or from the cloud.

dmolokanov:
> I think it is very confusing how we have upstream/events folder contain EventHandler prefixed type names but they don't implement EventHandler trait. Also, the types in upstream/rpc implement the EventHandler trait and are prefixed by RpcHandler. Since we are in a time crunch I think it is fine if we don't fix it in this PR.

I agree. It is due to the fact that Event is a very common name and can mean anything. What if we change `EventHandler` to `MqttEventHandler`?

and-rewsmith (approved):

PR #3780, opened 2020-10-22
[k8s] Environment Variables can be null or empty.

darobs:
In config.yaml for iotedged, customer set an environment variable in agent bootstrap like this:
```yaml
agent:
  name: "edgeAgent"
  type: "docker"
  env:
    https_proxy:
```
Technically, this is acceptable for environment variables - a null value actually gets translated to "https_proxy=" in the docker API. This PR allows null values in config.yaml, and in the ModuleSpec.

For JSON, use a deserialization function to read env vars as optional, then transform empty entries into default values.
For YAML config, don't return an error in Yaml::Null case, transform it into ValueKind::Nil.

### DIFF
--- a/edgelet/edgelet-utils/src/ser_de.rs
+++ b/edgelet/edgelet-utils/src/ser_de.rs
@@ -81,7 +81,9 @@ where
     sorted_map.serialize(serializer)
 }
 
-pub fn allow_null_map_values<'de, K, V, D>(deserializer: D) -> StdResult<HashMap<K, V>, D::Error>
+pub fn deserialize_map_with_default_values<'de, K, V, D>(
+    deserializer: D,
+) -> StdResult<HashMap<K, V>, D::Error>
 where
     K: Deserialize<'de> + Eq + std::hash::Hash,
     V: Deserialize<'de> + Default,
@@ -170,8 +172,8 @@ mod tests {
     #[derive(Debug, Deserialize, Serialize)]
     struct Setting {
         #[serde(serialize_with = "serialize_ordered")]
-        #[serde(deserialize_with = "allow_null_map_values")]
-        pub map: HashMap<String, String>,
+        #[serde(deserialize_with = "deserialize_map_with_default_values")]
+        map: HashMap<String, String>,
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


edgelet/edgelet-utils/src/ser_de.rs
```diff
@@ -81,6 +81,63 @@ where
     sorted_map.serialize(serializer)
 }
 
+pub fn allow_null_map_values<'de, K, V, D>(deserializer: D) -> StdResult<HashMap<K, V>, D::Error>
```
arsing, line 84: `deserialize_map_with_default_values` perhaps.

edgelet/edgelet-utils/src/ser_de.rs
```diff
@@ -110,10 +167,11 @@ mod tests {
         options: Options,
     }
 
-    #[derive(Debug, Serialize)]
+    #[derive(Debug, Deserialize, Serialize)]
     struct Setting {
         #[serde(serialize_with = "serialize_ordered")]
-        map: HashMap<String, String>,
+        #[serde(deserialize_with = "allow_null_map_values")]
+        pub map: HashMap<String, String>,
```
arsing, line 174: Why does it need to be `pub` ?

edgelet/edgelet-utils/src/yaml_file_source.rs
```diff
@@ -116,8 +116,8 @@ fn from_yaml_value(uri: Option<&String>, value: Yaml) -> Result<Value, ConfigErr
 
             Ok(Value::new(uri, l))
         }
-
-        // 1. Yaml NULL
+        // 1. Yaml NULL - Option::None will transform into ValueKind::Nil
```
arsing, line 119: Since `None` transforms into `Nil`, fix it on line 129 instead. And remove the paragraph on line 126.

arsing (changes requested):

edgelet/edgelet-utils/src/ser_de.rs
darobs: It probably doesn't.

arsing (approved):

PR #3790, opened 2020-10-23
EdgeHub startup: selecting upstream protocol and async config updater start

vipeller:


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -209,7 +209,11 @@ void RegisterAmqpModule(ContainerBuilder builder)
             TimeSpan configUpdateFrequency = TimeSpan.FromSeconds(configUpdateFrequencySecs);
             bool checkEntireQueueOnCleanup = this.configuration.GetValue("CheckEntireQueueOnCleanup", false);
             bool closeCloudConnectionOnDeviceDisconnect = this.configuration.GetValue("CloseCloudConnectionOnDeviceDisconnect", true);
-            bool isLegacyUpstream = this.configuration.GetValue("mqttBrokerSettings:legacyUpstream", true);
+
+            bool isLegacyUpstream = !experimentalFeatures.Enabled
```
richma-ms, line 213: >experimentalFeatures.Enabled [](start = 37, length = 28)

this is not needed, both experimentalFeatures.Enabled is the master switch and also taken into account when calculating EnableMqttBroker and EnabledNestedEdge

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -97,7 +97,10 @@ static async Task<int> MainAsync(IConfigurationRoot configuration)
             logger.LogInformation("Initializing configuration");
             IConfigSource configSource = await container.Resolve<Task<IConfigSource>>();
             ConfigUpdater configUpdater = await container.Resolve<Task<ConfigUpdater>>();
-            await configUpdater.Init(configSource);
+            var configUpdaterStartupFailed = new TaskCompletionSource<bool>();
+            _ = configUpdater.Init(configSource).ContinueWith(
```
richma-ms, line 101: >_ =  [](start = 12, length = 4)

is this explicit var discard needed?

richma-ms (approved):

PR #3792, opened 2020-10-23
Packaging changes for Module Runtime

massand:
Edge Module Runtime (formerly Edge Security Daemon) now depends on Identity Service (aziot-identity-service) package.
The Identity Service package requires manual installation steps to configure provisioning information and certificates.
The experience will be refined through the introduction of an 'iotedge init' command tool. For now, the install message
clarifies the steps needed to the package installer.

- Deleted debian8 (all builds), centos7 (arm/arm64 builds) and added ubuntu20.04 (all build flavors) packages
- Cleaned up Windows setup scripts
- Removed package.slow CI build
- Updated version for iotedge package to 1.1 (bumped for now - final version TBD)


massand:
Note that the final packaging plan includes changing the names of binary `iotedged` to aziot-edged and package 'iotedge` to `iotedge-aziot`. However, I realized that we cannot perform automated e2e tests easily if we change the entire script for install. So, I have decided to preserve the old binary and package names. I have tasked myself with doing the final rename after we have done e2e validation with nested edge bits in master.

arsing (approved):

PR #3797, opened 2020-10-23
[mqtt] Bridge Simple In-Flight Management

and-rewsmith:
We had a more complex PR for in flight with an in-flight publish handle. This alternative approach was good in that it allowed us to account rpc messages into our in-flight pool. However we ran into undefined behavior with the implementation in our egress message pump (await inside select macro). This behavior is scary, so we should stay on the safe side and just implement in-flight in the egress message pump.

I have implemented this with the type BufferUnordered which allows running a limited set of futures at one time.

### DIFF
--- a/mqtt/mqtt-bridge/src/client.rs
+++ b/mqtt/mqtt-bridge/src/client.rs
@@ -411,27 +411,18 @@ impl ShutdownHandle {
     }
 }
 
-#[async_trait]
-pub trait ClientPublishHandle {
-    async fn publish(&mut self, publication: Publication) -> Result<(), PublishError>;
-}
-
 /// A client publish handle.
 #[derive(Debug, Clone)]
 pub struct PublishHandle(mqtt3::PublishHandle);
 
-#[async_trait]
-impl ClientPublishHandle for PublishHandle {
-    async fn publish(&mut self, publication: Publication) -> Result<(), PublishError> {
+impl PublishHandle {
+    pub async fn publish(&mut self, publication: Publication) -> Result<(), PublishError> {
         self.0.publish(publication).await
     }
 }
 
 mockall::mock! {
-    pub PublishHandle {}
-
-    #[async_trait]
-    pub trait ClientPublishHandle {
+    pub PublishHandle {
         async fn publish(&mut self, publication: Publication) -> Result<(), PublishError>;
     }
 
--- a/mqtt/mqtt-bridge/src/client.rs
+++ b/mqtt/mqtt-bridge/src/client.rs
@@ -461,13 +461,32 @@ impl ShutdownHandle {
     }
 }
 
+#[async_trait]
+pub trait ClientPublishHandle {
+    async fn publish(&mut self, publication: Publication) -> Result<(), PublishError>;
+}
+
 /// A client publish handle.
+#[derive(Debug, Clone)]
 pub struct PublishHandle(mqtt3::PublishHandle);
 
-#[automock]
-impl PublishHandle {
-    pub async fn publish(&mut self, publication: Publication) -> Result<(), PublishError> {
+#[async_trait]
+impl ClientPublishHandle for PublishHandle {
+    async fn publish(&mut self, publication: Publication) -> Result<(), PublishError> {
         self.0.publish(publication).await
+    }
+}
+
+mockall::mock! {
+    pub PublishHandle {}
+
+    #[async_trait]
+    trait ClientPublishHandle {
+        async fn publish(&mut self, publication: Publication) -> Result<(), PublishError>;
+    }
+
+    pub trait Clone {
+        fn clone(&self) -> Self;
     }
 }
 
--- a/mqtt/mqtt-bridge/src/client.rs
+++ b/mqtt/mqtt-bridge/src/client.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::default_trait_access)] // Needed because mock! macro violates
 #![allow(dead_code)] // TODO remove when ready
 use std::{fmt::Display, io::Error, io::ErrorKind, pin::Pin, str, time::Duration};
 
@@ -430,7 +431,7 @@ mockall::mock! {
     pub PublishHandle {}
 
     #[async_trait]
-    trait ClientPublishHandle {
+    pub trait ClientPublishHandle {
         async fn publish(&mut self, publication: Publication) -> Result<(), PublishError>;
     }
 
--- a/mqtt/mqtt-bridge/src/client.rs
+++ b/mqtt/mqtt-bridge/src/client.rs
@@ -1,7 +1,5 @@
 #![allow(dead_code)] // TODO remove when ready
-use std::{
-    collections::HashSet, fmt::Display, io::Error, io::ErrorKind, pin::Pin, str, time::Duration,
-};
+use std::{fmt::Display, io::Error, io::ErrorKind, pin::Pin, str, time::Duration};
 
 use async_trait::async_trait;
 use chrono::Utc;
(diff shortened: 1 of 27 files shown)


mqtt/mqtt-bridge/src/pump/egress.rs
```diff
@@ -59,28 +61,44 @@ where
         info!("starting egress publication processing...");
 
         let mut shutdown = shutdown_recv.fuse();
-        let mut loader = store.loader().fuse();
+
+        // Take the stream of loaded messages and convert to a stream of futures which publish.
+        // Then convert to buffered stream so that we can have multiple in-flight and also limit number of publications.
+        let loader = store.loader();
+        let load_and_publish = loader
+            .filter_map(|loaded| async {
```
and-rewsmith, line 69: There is some build error here, but it doesn't show in rust analyzer

dmolokanov, line 69: It'd be a bit easier to grasp if Ok block or whole body to be moved to a closure or separate method

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -410,16 +410,35 @@ impl ShutdownHandle {
     }
 }
 
+#[async_trait]
+pub trait ClientPublishHandle {
```
dmolokanov, line 414: we dont need this trait anymore, right?

dmolokanov (changes requested):

mqtt/mqtt-bridge/src/client.rs
and-rewsmith: It is needed because of the mock! macro needed for clone to work.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -410,16 +410,35 @@ impl ShutdownHandle {
...
+
+    #[async_trait]
+    trait ClientPublishHandle {
+        async fn publish(&mut self, publication: Publication) -> Result<(), PublishError>;
+    }
+
+    pub trait Clone {
+        fn clone(&self) -> Self;
+    }
+}
```
and-rewsmith, line 440: I need to fix the tests to work with clone

and-rewsmith: Nevermind they still work because rpc doesn't clone the handle

mqtt/mqtt-bridge/src/pump/egress.rs
```diff
@@ -59,26 +69,26 @@ where
         info!("starting egress publication processing...");
 
         let mut shutdown = shutdown_recv.fuse();
-        let mut loader = store.loader().fuse();
+
+        // Take the stream of loaded messages and convert to a stream of futures which publish.
+        // Then convert to buffered stream so that we can have multiple in-flight and also limit number of publications.
+        let loader = store.loader();
+        let publish_handle_stream = stream::iter(iter::repeat(publish_handle.clone()));
```
and-rewsmith, line 76: Lmk if there is a better way. We can't move the publish handle into the map closure.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::default_trait_access)] // Needed because mock! macro violates
```
and-rewsmith, line 1: Putting this above the macro didn't work.

and-rewsmith:
I tested the demo scenario earlier this morning and it worked.

mqtt/mqtt-bridge/src/client.rs
dmolokanov: Ok, I fixed this for `Default`, it seems like we need something more reliable

mqtt/mqtt-bridge/src/pump/egress.rs
dmolokanov: I came up with 
```rust
let publications = store
            .loader()
            .filter_map(|loaded| {
                let publish_handle = publish_handle.clone();
                async {
                    match loaded {
                        Ok((key, publication)) => {
                            Some(try_publish(key, publication, publish_handle))
                        }
                        Err(e) => {
                            error!(err = %e, "failed loading publication from store");
                            None
                        }
                    }
                }
            })
            .buffer_unordered(MAX_IN_FLIGHT);
        pin_mut!(publications);
...

...
async fn try_publish(key: Key, publication: Publication, mut publish_handle: PublishHandle) -> Key {
    debug!("publishing {:?}", key);
    if let Err(e) = publish_handle.publish(publication).await {
        error!(err = %e, "failed publish");
    }

    key
}
```

mqtt/mqtt-bridge/src/client.rs
dmolokanov: I checked docs and we could use something like this 
```rust
/// A client publish handle.
#[derive(Debug, Clone)]
pub struct PublishHandle(mqtt3::PublishHandle);

impl PublishHandle {
    pub async fn publish(&mut self, publication: Publication) -> Result<(), PublishError> {
        self.0.publish(publication).await
    }
}

mock! {
    pub PublishHandle {
        async fn publish(&mut self, publication: Publication) -> Result<(), PublishError>;
    }

    pub trait Clone {
        fn clone(&self) -> Self;
    }
}
```

mqtt/mqtt-bridge/src/pump/egress.rs
```diff
@@ -14,6 +19,11 @@ pub use crate::client::MockPublishHandle as PublishHandle;
 #[cfg(not(test))]
 use crate::client::PublishHandle;
 
+use mqtt3::proto::Publication;
+use std::iter;
+
+const MAX_IN_FLIGHT: usize = 16;
```
dmolokanov, line 25: Check with @ancaantochi if that's should come from configuration

dmolokanov (changes requested):

mqtt/mqtt-bridge/src/pump/egress.rs
and-rewsmith: I'm guessing you didn't try to build this? I tried this Friday and it didn't work.

mqtt/mqtt-bridge/src/client.rs
and-rewsmith: This won't work because the tests won't be able to resolve publish for the MockPublishHandle. From the docs:
https://docs.rs/mockall/0.8.3/mockall/macro.mock.html

> which basically involves repeating the struct's or trait's definitions.

Since we can't define impl in this mock! macro I was thinking that we needed a trait. I just tried it again to confirm and this won't work.

mqtt/mqtt-bridge/src/pump/egress.rs
and-rewsmith: She said you said it shouldn't be configurable. Do we expect customers will want to configure this? Do they currently do so for edgehub? If so, I think we can add it now.

mqtt/mqtt-bridge/src/client.rs
dmolokanov: I've just ran suggested above changes and tests passed.

mqtt/mqtt-bridge/src/pump/egress.rs
dmolokanov: not for edgehub right. for another remote broker can potentially be different limits.

mqtt/mqtt-bridge/src/pump/egress.rs
dmolokanov: actually I did and it compiled at least :)

mqtt/mqtt-bridge/src/pump/egress.rs
and-rewsmith: I was just saying that if it is something commonly configured without the broker/bridge, we should add the configuration now.

and-rewsmith:
Tested and it works

dmolokanov (approved):
LGTM!

PR #3798, opened 2020-10-24
bridge address review comments

ancaantochi:
Address some review comments from a merged PR

### DIFF
--- a/mqtt/mqtt-bridge/src/config_update.rs
+++ b/mqtt/mqtt-bridge/src/config_update.rs
@@ -1,11 +1,12 @@
 use std::collections::HashMap;
 
 use serde::Deserialize;
+use tracing::debug;
 
 use crate::{bridge::BridgeHandle, controller::Error, settings::Direction, settings::TopicRule};
 
-// Keeps the current subscriptions and forwards and calculates the diff with an BridgeUpdate
-// It is used to send a diff to a BridgeHandle and update itself with latest configuration
+/// Keeps the current subscriptions and forwards and calculates the diff with an BridgeUpdate
+/// It is used to send a diff to a BridgeHandle and update itself with latest configuration
 pub struct ConfigUpdater {
     bridge_handle: BridgeHandle,
     current_subscriptions: HashMap<String, TopicRule>,
@@ -21,8 +22,10 @@ impl ConfigUpdater {
         }
     }
 
-    pub async fn send_update(&mut self, bridge_update: &BridgeUpdate) -> Result<(), Error> {
-        let diff = self.diff(&bridge_update);
+    pub async fn send_update(&mut self, bridge_update: BridgeUpdate) -> Result<(), Error> {
+        let diff = self.diff(bridge_update);
+
+        debug!("sending diff {:?}", diff.clone());
 
         self.bridge_handle.send(diff.clone()).await?;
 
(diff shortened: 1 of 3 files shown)


mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -111,7 +114,7 @@ impl BridgeController {
                     }
 
                     if let Some(config) = self.bridge_handles.lock().await.get_mut(UPSTREAM) {
-                        if let Err(e) = config.send_update(&bridge_update).await {
+                        if let Err(e) = config.send_update(bridge_update.to_owned()).await {
```
ancaantochi, line 117: had to add to_owned here because it gets consumed

mqtt/mqtt-bridge/src/controller.rs
```diff
@@ -101,7 +104,7 @@ impl BridgeController {
         };
 
         let updates = async move {
-            while let Some(update) = self.updates_receiver.recv().await {
```
and-rewsmith, line 104: recv() call was busy?

and-rewsmith (approved):

mqtt/mqtt-bridge/src/config_update.rs
```diff
@@ -21,8 +22,10 @@ impl ConfigUpdater {
         }
     }
 
-    pub async fn send_update(&mut self, bridge_update: &BridgeUpdate) -> Result<(), Error> {
-        let diff = self.diff(&bridge_update);
+    pub async fn send_update(&mut self, bridge_update: BridgeUpdate) -> Result<(), Error> {
+        let diff = self.diff(bridge_update);
+
+        debug!("sending diff {:?}", diff.clone());
```
dmolokanov, line 28:
```suggestion
        debug!("sending diff {:?}", diff);
```

mqtt/mqtt-bridge/src/controller.rs
dmolokanov: which is fine? I guess we don't need `bridge_update` anywhere here
```suggestion
                        if let Err(e) = config.send_update(bridge_update).await {
```

mqtt/mqtt-bridge/src/config_update.rs
dmolokanov:
```rust
self.update(diff)
...
    fn update(&mut self, bridge_diff: BridgeDiff) {
        let (local_updates, remote_updates) = bridge_diff.into_parts();
        update_pump(local_updates, &mut self.current_forwards);

        update_pump(remote_updates, &mut self.current_subscriptions)
    }
...

fn update_pump(pump_diff: PumpDiff, current: &mut HashMap<String, TopicRule>) {
    let (added, removed) = pump_diff.into_parts();
    for rule in added {
        current.insert(rule.subscribe_to(), rule);
    }

    for rule in removed {
        current.remove(&rule.subscribe_to());
    }
}
....

fn diff_topic_rules(updated: Vec<TopicRule>, current: &HashMap<String, TopicRule>) -> PumpDiff 
...
 for sub in updated {
        if !current.contains_key(&sub.subscribe_to())
            || current
                .get(&sub.subscribe_to())
                .filter(|curr| *curr == &sub)
                == None
        {
            added.push(sub);
        }
    }
...

```

dmolokanov: Notice how we removed cloning just changing value move semantic from borrowing.

dmolokanov (changes requested):

mqtt/mqtt-bridge/src/controller.rs
ancaantochi: Actually I noticed that to_owned is not needed, not sure why I thought was needed.

mqtt/mqtt-bridge/src/controller.rs
ancaantochi: Denis suggested in the original PR to use next

dmolokanov (approved):
LGTM

PR #3799, opened 2020-10-24
[mqtt] Fix bridge client id error

and-rewsmith:
Fixing the bridge client id to match edgehub expected system component structure.

### DIFF
--- a/mqtt/mqtt-bridge/src/bridge.rs
+++ b/mqtt/mqtt-bridge/src/bridge.rs
@@ -96,10 +96,7 @@ impl Bridge<WakingMemoryStore> {
                     &system_address,
                     settings.keep_alive(),
                     settings.clean_session(),
-                    Credentials::Anonymous(format!(
-                        "{}/$edgeHub/$bridge",
-                        device_id,
-                    )),
+                    Credentials::Anonymous(format!("{}/$edgeHub/$bridge", device_id,)),
                 ))
                 .with_rules(settings.forwards());
             })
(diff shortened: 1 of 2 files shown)


and-rewsmith:
I tested with $edgehub/$upstream/$bridge and didn't see the error. I have since changed it though to [DEVICE_ID[/$edgeHub/$bridge, but didn't test. 

I don't think we will see the error though according because this logic states as long as we end in $bridge and are three segments we should be fine.
https://github.com/Azure/iotedge/blob/master/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/SystemComponentIdProvider.cs#L9

ancaantochi (approved):

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -96,11 +96,7 @@ impl Bridge<WakingMemoryStore> {
                     &system_address,
                     settings.keep_alive(),
                     settings.clean_session(),
-                    Credentials::Anonymous(format!(
-                        "{}/$edgeHub/$bridge/{}",
-                        device_id,
-                        settings.name()
-                    )),
+                    Credentials::Anonymous(format!("{}/$edgeHub/$bridge", device_id,)),
```
dmolokanov, line 99: Not sure if this is the correct fix.  Since we are going to use different MQTT client to connect to the local broker, we need to assign a unique client_id to each of them. What will happen right now for multiple bridges: all local clients will share the same client_id and start to fight for the session in the local broker. The next one will take over and previous connection will be closed. And same for the current one again and again.

PR #3802, opened 2020-10-26
Added FeatureFlagAuthorizer to wrap PolicyAuthorizer

vadim-kovalyov:
Since policy engine requires default configuration, we needed to wrap `PolicyAuthorizer` with a feature flag, so it does not interfere with existing e2e tests. Otherwise all existing e2e tests needs to be fixed to include policy engine configuration.

### DIFF
--- a/mqtt/mqttd/src/broker/bootstrap/edgehub.rs
+++ b/mqtt/mqttd/src/broker/bootstrap/edgehub.rs
@@ -21,7 +21,10 @@ use mqtt_broker::{
 };
 use mqtt_edgehub::{
     auth::PolicyAuthorizer,
-    auth::{EdgeHubAuthenticator, EdgeHubAuthorizer, LocalAuthenticator, LocalAuthorizer},
+    auth::{
+        EdgeHubAuthenticator, EdgeHubAuthorizer, FeatureFlagAuthorizer, LocalAuthenticator,
+        LocalAuthorizer,
+    },
     command::{
         AuthorizedIdentitiesCommand, BridgeUpdateCommand, CommandHandler, CommandHandlerError,
         DisconnectCommand, PolicyUpdateCommand, ShutdownHandle,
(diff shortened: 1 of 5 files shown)


dmolokanov (dismissed):
I believe it is temporal thing and to be removed as soon as you fix tests. If my assumption is not correct please import structs/traits with use block instead of FQN.

vadim-kovalyov:
> I believe it is temporal thing and to be removed as soon as you fix tests. If my assumption is not correct please import structs/traits with use block instead of FQN.

Nothing more permanent than temporary :) But I believe it should go away when MQTT broker flag goes away.

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -32,6 +35,8 @@ use mqtt_edgehub::{
 
 const DEVICE_ID_ENV: &str = "IOTEDGE_DEVICEID";
 
+const POLICY_ENABLED_ENV: &str = "experimentalFeatures:policyEngineEnabled";
```
varunpuranik, line 38: We should just use the experimentalFeatures:mqttBrokerEnabled feature flag.

varunpuranik (changes requested):
Why do we need to add a new flag just for the policy engine?

mqtt/mqttd/src/broker/bootstrap/edgehub.rs
```diff
@@ -52,8 +57,15 @@ pub async fn broker(
     state: Option<BrokerSnapshot>,
     broker_ready: &BrokerReady,
 ) -> Result<Broker<impl Authorizer>> {
-    // TODO: Use AllowAll as bottom level authorizer until Policies are sent over from EdgeHub
-    let authorizer = LocalAuthorizer::new(EdgeHubAuthorizer::new(AllowAll, broker_ready.handle()));
+    let device_id = env::var(DEVICE_ID_ENV).context(DEVICE_ID_ENV)?;
+
+    let authorizer = LocalAuthorizer::new(EdgeHubAuthorizer::new(
```
dmolokanov, line 62: curious if we could just select here in place: 
```rust
env::var(POLICY_ENABLED_ENV).ok().map_or_else(||AllowAll, |_| PolicyAuthorizer::new(device_id, broker_ready.handle()))
```

dmolokanov (approved):

vadim-kovalyov:
No longer required.

PR #3806, opened 2020-10-26
Enable policy engine by default.

vadim-kovalyov:
This will enable policy engine by default when broker is enabled.


varunpuranik (approved):

dmolokanov (approved):
Shouldn't it be a cherry-pick from master though?

PR #3808, opened 2020-10-26
mqtt bridge remove enableupstreambridge flag

ancaantochi:
If the broker is enabled and starts in edgehub mode and GatewayHostname is set, then bridge is enabled.


mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -392,8 +377,7 @@ mod tests {
 
     #[test]
     #[serial(env_settings)]
-    fn from_env_no_upstream_protcol() {
```
and-rewsmith, line 395: why change the test?

and-rewsmith (approved):

PR #3814, opened 2020-10-26
[mqtt] Rework shutdown for bridge and pumps

dmolokanov:
`select!` macros does not support async operation in its branches. As a result execution stops without making any proggress. So I refactored awaiting for pumps and bridge itself.

### DIFF
--- a/mqtt/mqtt-bridge/src/bridge.rs
+++ b/mqtt/mqtt-bridge/src/bridge.rs
@@ -1,5 +1,9 @@
+use futures_util::{
+    future::{self, Either},
+    pin_mut,
+};
 use mqtt3::ShutdownError;
-use tokio::{select, sync::oneshot::Sender};
+use tokio::sync::oneshot::Sender;
 use tracing::{debug, error, info, info_span};
 use tracing_futures::Instrument;
 
@@ -127,13 +131,15 @@ where
     S: StreamWakeableState + Send,
 {
     pub async fn run(self) -> Result<(), BridgeError> {
-        info!("Starting {} bridge...", self.connection_settings.name());
+        info!("Starting bridge...");
 
+        let shutdown_local_pump = self.local_pump.handle();
         let local_pump = self
             .local_pump
             .run()
             .instrument(info_span!("pump", name = "local"));
 
+        let shutdown_remote_pump = self.remote_pump.handle();
         let remote_pump = self
             .remote_pump
             .run()
@@ -144,19 +150,44 @@ where
             self.connection_settings.name()
         );
 
-        select! {
-            _ = local_pump => {
-                // TODO shutdown remote pump
-                // shutdown_handle.shutdown().await?;
-            }
+        pin_mut!(local_pump, remote_pump);
 
-            _ = remote_pump => {
-                // TODO shutdown local pump
-                // shutdown_handle.shutdown().await?;
+        match future::select(local_pump, remote_pump).await {
+            Either::Left((local_pump, remote_pump)) => {
+                if let Err(e) = local_pump {
+                    debug!(error = %e, "local pump exited with error");
+                } else {
+                    debug!("local pump exited");
+                }
+
+                debug!("shutting down remote pump...");
+                shutdown_remote_pump.shutdown().await;
+
+                if let Err(e) = remote_pump.await {
+                    debug!(error = %e, "remote pump exited with error");
+                } else {
+                    debug!("remote pump exited");
+                }
+            }
+            Either::Right((remote_pump, local_pump)) => {
+                if let Err(e) = remote_pump {
+                    debug!(error = %e, "remote pump exited with error");
+                } else {
+                    debug!("remote pump exited");
+                }
+
+                debug!("shutting down local pump...");
+                shutdown_local_pump.shutdown().await;
+
+                if let Err(e) = local_pump.await {
+                    debug!(error = %e, "local pump exited with error");
+                } else {
+                    debug!("local pump exited");
+                }
             }
         }
 
-        debug!("bridge {} stopped...", self.connection_settings.name());
+        debug!("bridge stopped...");
         Ok(())
     }
 
--- a/mqtt/mqtt-bridge/src/bridge.rs
+++ b/mqtt/mqtt-bridge/src/bridge.rs
@@ -3,7 +3,6 @@ use futures_util::{
     pin_mut,
 };
 use mqtt3::ShutdownError;
-use tokio::sync::oneshot::Sender;
 use tracing::{debug, error, info, info_span};
 use tracing_futures::Instrument;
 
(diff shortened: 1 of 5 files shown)


mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -127,13 +131,15 @@ where
     S: StreamWakeableState + Send,
 {
     pub async fn run(self) -> Result<(), BridgeError> {
-        info!("Starting {} bridge...", self.connection_settings.name());
```
and-rewsmith, line 130: We will have to make sure to log names for other bridges when we support them.

mqtt/mqtt-bridge/src/pump/egress.rs
```diff
@@ -49,22 +49,21 @@ where
...
 
-        let mut shutdown = shutdown_recv.fuse();
-
-        // Take the stream of loaded messages and convert to a stream of futures which publish.
-        // Then convert to buffered stream so that we can have multiple in-flight and also limit number of publications.
-        let loader = store.loader();
-        let load_and_publish = loader
+        // Take the stream of loaded messages and convert to a stream of futures
+        // which publish. Then convert to buffered stream so that we can have
+        // multiple in-flight and also limit number of publications.
```
and-rewsmith, line 64: You did this because exceeded 80 columns?

mqtt/mqtt-bridge/src/pump/egress.rs
```diff
@@ -80,15 +79,15 @@ where
                 }
             })
             .buffer_unordered(MAX_IN_FLIGHT);
-        pin_mut!(load_and_publish);
+        pin_mut!(publications);
```
and-rewsmith, line 82: nit: I liked the old name because it signified what was happening.

mqtt/mqtt-bridge/src/pump/ingress.rs
```diff
@@ -33,10 +33,12 @@ where
     }
 
     /// Runs ingress processing.
-    pub(crate) async fn run(mut self) {
+    pub(crate) async fn run(mut self) -> Result<(), IngressError> {
```
and-rewsmith, line 36: I'm guessing you foresee some reason to return Err in egress/ingress?

mqtt/mqtt-bridge/src/pump/ingress.rs
```diff
@@ -33,10 +33,12 @@ where
...
         info!("starting ingress publication processing...",);
```
and-rewsmith, line 37: We have a bad comma here

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -31,8 +35,13 @@ pub fn channel<M: 'static>() -> (PumpHandle<M>, mpsc::Receiver<PumpMessage<M>>)
 }
 
 #[derive(Debug, thiserror::Error)]
-#[error("unable to send command to pump")]
-pub struct PumpError;
+pub enum PumpError {
+    #[error("unable to send command to pump")]
+    Send,
+
+    #[error("error ocurred when running pump")]
```
vadim-kovalyov, line 42: `{0}`?

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -111,34 +120,106 @@ where
...
+                            debug!(error = %e, "ingress processing exited with error");
+                        } else {
+                            debug!("ingress processing exited");
+                        }
+
+                        if let Err(e) = egress.await {
+                            error!(error = %e, "egress processing exited with error");
+                        } else {
+                            info!("egress processing exited")
+                        }
```
and-rewsmith, line 163: There might be a better way to deduplicate this logic. Helper function?

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -111,34 +120,106 @@ where
...
+                if let Err(e) = messages.await {
+                    debug!(error = %e, "pump messages processor exited with error");
+                } else {
+                    debug!("pump messages processor exited");
+                }
+
+                ingress.map_err(|e| PumpError::Run(e.into()))
             }
-        };
+        }?;
```
and-rewsmith, line 219: Seems a bit weird we only raise error in case first-exiting thing is an error. Shouldn't we log and raise in all error cases?

and-rewsmith:
Let me know when you want me to test

mqtt/mqtt-bridge/src/bridge.rs
dmolokanov: Yeah, we log them in the span.

mqtt/mqtt-bridge/src/pump/egress.rs
dmolokanov: Oops, I guess it is left from my experiments

mqtt/mqtt-bridge/src/pump/mod.rs
dmolokanov: there are types differences, I'm not sure how to dedup logic. happy to listen though :)

dmolokanov:
> Let me know if you want me to test

Yes please!

mqtt/mqtt-bridge/src/pump/mod.rs
dmolokanov: good point

mqtt/mqtt-bridge/src/pump/ingress.rs
dmolokanov: yes, that's what we discussed with Anca. what should Pump do when the client errors out? proably a bit premature changes however I need them to return some error.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -144,19 +150,44 @@ where
...
-            _ = local_pump => {
-                // TODO shutdown remote pump
-                // shutdown_handle.shutdown().await?;
-            }
+        pin_mut!(local_pump, remote_pump);
+
+        match future::select(local_pump, remote_pump).await {
+            Either::Left((local_pump, remote_pump)) => {
+                if let Err(e) = local_pump {
+                    debug!(error = %e, "local pump exited with error");
```
ancaantochi, line 158: this should be logged as error, same for remote pump

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -144,19 +150,44 @@ where
...
+                if let Err(e) = local_pump.await {
+                    debug!(error = %e, "local pump exited with error");
+                } else {
+                    debug!("local pump exited");
+                }
             }
         }
 
-        debug!("bridge {} stopped...", self.connection_settings.name());
+        debug!("bridge stopped...");
```
ancaantochi, line 190: log as info

mqtt/mqtt-bridge/src/pump/ingress.rs
ancaantochi: is this going to shutdown the bridge? right now we don't recreate it so this will trigger the broker shutdown I think. I was not planning to recreate the bridge because we discussed about validation so that client doesn't return errors.

ancaantochi (approved):

mqtt/mqtt-bridge/src/pump/egress.rs
and-rewsmith: So we are going to leave it?

PR #3817, opened 2020-10-27
Token auth fixes

richma-ms:
- Make iothub hostname check case-insensitive
- Trigger on-demand refresh for updated identities

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
@@ -137,7 +137,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy.Authenticators
             }
 
             if (string.IsNullOrWhiteSpace(hostName) ||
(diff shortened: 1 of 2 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs
```diff
@@ -137,7 +137,7 @@ internal bool ValidateAudience(string audience, IIdentity identity)
             }
 
             if (string.IsNullOrWhiteSpace(hostName) ||
-                !(this.iothubHostName.Equals(hostName) || this.edgeHubHostName.Equals(hostName)))
+                !(string.Equals(this.iothubHostName, hostName, StringComparison.OrdinalIgnoreCase) || string.Equals(this.edgeHubHostName, hostName, StringComparison.OrdinalIgnoreCase)))
```
varunpuranik, line 140: Is this correct? Are these case insensitive?

richma-ms: is the same reasoning as when we made deviceId/moduleId check case-insensitive, SAS tokens come in as all lower case by convention, so we need to relax our string comparison logic on the hostname as well.

yophilav (approved):

PR #3825, opened 2020-10-27
add client input validation

ancaantochi:
Adding validation for input length in the mqtt client to avoid mqtt3 client user error


mqtt/mqtt-bridge/src/client.rs
```diff
@@ -21,9 +21,10 @@ use crate::{
 };
 
 const DEFAULT_TOKEN_DURATION_MINS: i64 = 60;
-const DEFAULT_MAX_RECONNECT: Duration = Duration::from_secs(5);
+const DEFAULT_MAX_RECONNECT: Duration = Duration::from_secs(60);
```
ancaantochi, line 24: Increased default to 1min because it was retrying too often to reconnect

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -208,6 +208,9 @@ pub enum BridgeError {
     #[error("failed to get publish handle from client.")]
     PublishHandle(#[source] ClientError),
 
+    #[error("failed to validate client settings.")]
```
vadim-kovalyov, line 211:
```suggestion
    #[error("failed to validate client settings: {0}")]
```

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -436,18 +480,22 @@ pub struct UpdateSubscriptionHandle(mqtt3::UpdateSubscriptionHandle);
 
 #[automock]
 impl UpdateSubscriptionHandle {
-    pub async fn subscribe(
-        &mut self,
-        subscribe_to: SubscribeTo,
-    ) -> Result<(), UpdateSubscriptionError> {
-        self.0.subscribe(subscribe_to).await
+    pub async fn subscribe(&mut self, subscribe_to: SubscribeTo) -> Result<(), ClientError> {
+        validate_length(&subscribe_to.topic_filter)?;
```
vadim-kovalyov, line 484: just curious why it is important to validate for such, it seems, edge case? what would happen with the client if we send a long string to it? I assume it will fail in the codec. Is it unrecoverable error?

mqtt/mqtt-bridge/src/client.rs
dmolokanov: Can we add exponential backoff with randomized intervals?

mqtt/mqtt-bridge/src/client.rs
dmolokanov: This is a good point actually. We were focused on the client throwing an error and should we re-create client/pump/bridge. And it seems like it is fine to just log an error in this case. I wonder if we even need this input validation. Lets discuss offline?

mqtt/mqtt-bridge/src/client.rs
ancaantochi: this is the max backoff, before it reaches it is using exponential but it starts with 1 sec not random value. client doesn't expose a setting for starting interval, do you thing should be added?

mqtt/mqtt-bridge/src/client.rs
ancaantochi: If we don't validate the client will get into an error and can't recover without recreating it. I think we can have both validation and re-create of bridge.

ancaantochi: actually not needed for subscribe and publish because handle validate internally before sending to the client

vadim-kovalyov (approved):
All these settings are coming from EH twin, right? If so, we should add the same validation on the EH side, I believe...

PR #3830, opened 2020-10-29
Update Nested Edge Deployment manifests 

yophilav:
Update Nested Edge Deployment manifests to be using the most recent environment variables.


dylanbronson (approved):
LGTM - make sure CollectMetrics is actually on by default though

PR #3833, opened 2020-10-29
[mqtt] Unify log formatting across all rust code

dmolokanov:
Unify log formatting across all rust code:
- all lower case
- EdgeHub log entry format
- `error` as additional variable

Fix some log mistypings

### DIFF
--- a/mqtt/mqttd/src/tracing/format.rs
+++ b/mqtt/mqttd/src/tracing/format.rs
@@ -7,15 +7,15 @@ use tracing_subscriber::fmt::{
 };
 
 /// Marker for `Format` that indicates that the syslog format should be used.
-pub(crate) struct Syslog;
+pub(crate) struct EdgeHub;
 
 /// Custom event formatter.
-pub(crate) struct Format<F = Syslog, T = ChronoLocal> {
+pub(crate) struct Format<F = EdgeHub, T = ChronoLocal> {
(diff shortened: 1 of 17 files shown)


mqtt/mqttd/src/tracing/format.rs
```diff
@@ -42,7 +42,7 @@ where
 
         write!(writer, "<{}> ", fmt_level.syslog_level())?;
         self.timer.format_time(writer)?;
-        write!(writer, "[{}] - [{}{}] ", fmt_level, fmt_ctx, meta.target(),)?;
+        write!(writer, "[{}] [{}{}] - ", fmt_level, fmt_ctx, meta.target(),)?;
```
and-rewsmith, line 45: Why did we add this dash here?

and-rewsmith (approved):

mqtt/mqttd/src/tracing/format.rs
dmolokanov: Isn't it the edgehub log format?
```
<6> 2020-10-27 19:11:19.403 +00:00 [INF] [EdgeHub] - Shutdown complete.
```

and-rewsmith: Yes this is true. I misread the log.

PR #3834, opened 2020-10-29
Recognize the new DPS registration substatus that indicates cloud identity did not change.

arsing:
This commit replaces the "deviceDataUpdated" speculatively added by
1750d178585c1ae69c33893e6e7d15e1269e376b with the actual substatus that
DPS has added. The semantics are still the same.

### DIFF
--- a/edgelet/dps/src/lib.rs
+++ b/edgelet/dps/src/lib.rs
@@ -10,7 +10,6 @@
     clippy::use_self
 )]
 
-pub mod dps;
 pub mod error;
 mod model;
 pub mod registration;
@@ -22,4 +21,4 @@ pub use model::{
 };
 pub use registration::{DpsClient, DpsTokenSource};
 
-pub const DPS_API_VERSION: &str = "2018-11-01";
+pub const DPS_API_VERSION: &str = "2019-04-15";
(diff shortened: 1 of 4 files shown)


edgelet/dps/src/lib.rs
```diff
@@ -22,4 +21,4 @@ pub use model::{
 };
 pub use registration::{DpsClient, DpsTokenSource};
 
-pub const DPS_API_VERSION: &str = "2018-11-01";
+pub const DPS_API_VERSION: &str = "2019-04-15";
```
CindyXing, line 24: Have we confirmed this is the final version DPS is going to use?

edgelet/provisioning/src/provisioning.rs
```diff
@@ -836,12 +850,15 @@ where
                     move |mut prov_result| {
                         debug!("Provisioning result {:?}", prov_result);
                         let reconfigure = match prov_result.reconfigure {
-                            ReprovisioningStatus::DeviceDataUpdated => {
```
CindyXing, line 839: So DPS will not return DeviceDataUpdated anymore or edge will not handle this substatus?

arsing: It was added speculatively by Anca. DPS never returned `DeviceDataUpdated`, and we never had any code to parse it from DPS's response it even if it did, so the code path was never used.

edgelet/dps/src/lib.rs
arsing: Yes.

CindyXing (approved):

damonbarry (approved):

damonbarry (approved):

PR #3838, opened 2020-10-29
Fixed multiple vars issue in policy definition

vadim-kovalyov:
This PR fixes the issue where multiple vars in a string in policy definition didn't work.


dmolokanov (approved):
LGTM

PR #3844, opened 2020-10-30
handle child $edgeHub as indirect client

vipeller:
This is  quick fix to the issue that direct clients (e.g. devices, modules) use a different dialect sending their messages (e.g. twin update) than messages connecting indirectly, where the broker/bridge transforms their topics to have identity information a consistent structure for multilevel messaging.

When a client connects "indirectly" (meaning that the fact of the connection will be noticed by not by the connected client events published by the broker, but that the clients send messages without prior connection), they get a flag, so when edgeHub sends mqtt messages to them, it uses the right dialect.

Child $edgeHub is an exception - the bridge uses its identity to connect, so the parent will receive a connection attempt and sets the flag for $edgeHub as a directly connecting device. However $edgeHub itself is not connected directly (it uses the same bridged channel as other indirectly connected child devices) and it needs the appropriate 'indirect dialect'.

With this change the connected $edgeHub will receive the messages on appropriate topic.

Further improvement can be to monitor every client and set flags according to their subscriptions/messages. Created a ticket to investigate its benefits and develop it if makes sense.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/DeviceProxy.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/DeviceProxy.cs
@@ -44,7 +44,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
             // Although it is a direct client, it uses the indirect topics
             if (identity is ModuleIdentity moduleIdentity)
             {
-                if (string.Equals(moduleIdentity.ModuleId, "$edgeHub"))
+                if (string.Equals(moduleIdentity.ModuleId, Constants.EdgeHubModuleId))
                 {
                     this.isDirectClient = false;
                 }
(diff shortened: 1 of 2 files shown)


vadim-kovalyov (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/DeviceProxy.cs
```diff
@@ -40,6 +40,16 @@ public class DeviceProxy : IDeviceProxy
             this.isActive = new AtomicBoolean(true);
             this.isDirectClient = isDirectClient;
 
+            // when a child edge device connects, it uses $edgeHub identity.
+            // Although it is a direct client, it uses the indirect topics
+            if (identity is ModuleIdentity moduleIdentity)
+            {
+                if (string.Equals(moduleIdentity.DeviceId, "$edgeHub"))
```
ancaantochi: nit: use a const for $edgeHub

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/DeviceProxy.cs
```diff
@@ -40,6 +40,16 @@ public class DeviceProxy : IDeviceProxy
             this.isActive = new AtomicBoolean(true);
             this.isDirectClient = isDirectClient;
 
+            // when a child edge device connects, it uses $edgeHub identity.
+            // Although it is a direct client, it uses the indirect topics
+            if (identity is ModuleIdentity moduleIdentity)
+            {
+                if (string.Equals(moduleIdentity.ModuleId, "$edgeHub"))
```
ancaantochi, line 47: you can use Constants.EdgeHubModuleId from  Microsoft.Azure.Devices.Edge.Hub.Core

ancaantochi (approved):

vadim-kovalyov (dismissed):

vadim-kovalyov (approved):

PR #3846, opened 2020-10-30
Iotedged agent spec: Environment Variables can be null or empty

darobs:
In config.yaml for iotedged, customer set an environment variable in agent bootstrap like this:
```yaml
agent:
  name: "edgeAgent"
  type: "docker"
  env:
    https_proxy:
```
Technically, this is acceptable for environment variables - a null value actually gets translated to "https_proxy=" in the docker API. This PR allows null values in config.yaml, and in the ModuleSpec.

For JSON, use a deserialization function to read env vars as optional, then transform empty entries into default values.
For YAML config, don't return an error in Yaml::Null case, transform it into ValueKind::Nil.

### DIFF
--- a/edgelet/edgelet-core/src/module.rs
+++ b/edgelet/edgelet-core/src/module.rs
@@ -13,7 +13,7 @@ use failure::{Fail, ResultExt};
 use futures::{Future, Stream};
 use serde_derive::Serialize;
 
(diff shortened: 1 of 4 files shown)


darobs:
Closed due to iotedged refactor.

PR #3851, opened 2020-10-30
Fixes for certificate properties

massand:
Server and identity certificates had their FQDNs incorrectly sanitized (dot character removed). Also, IPs were being dropped in certificate properties.
The method names have been refactored to match intent as well.

### DIFF
--- a/edgelet/edgelet-http-workload/src/server/cert/mod.rs
+++ b/edgelet/edgelet-http-workload/src/server/cert/mod.rs
@@ -167,15 +167,26 @@ fn generate_key_and_csr(
     let mut extensions = openssl::stack::Stack::new()?;
     extensions.push(extended_key_usage)?;
 
-    if props.san_entries().is_some() {
+    if props.dns_san_entries().is_some() || props.ip_entries().is_some() {
         let mut subject_alt_name = openssl::x509::extension::SubjectAlternativeName::new();
-        props
-            .san_entries()
-            .expect("san entries unexpectedly empty")
-            .iter()
-            .for_each(|s| {
-                subject_alt_name.dns(s);
-            });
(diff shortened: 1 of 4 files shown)


edgelet/edgelet-http-workload/src/server/cert/mod.rs
```diff
@@ -167,15 +167,26 @@ fn generate_key_and_csr(
...
+    if props.dns_san_entries().is_some() || props.ip_entries().is_some() {
         let mut subject_alt_name = openssl::x509::extension::SubjectAlternativeName::new();
-        props
-            .san_entries()
-            .expect("san entries unexpectedly empty")
-            .iter()
-            .for_each(|s| {
-                subject_alt_name.dns(s);
-            });
+        if props.dns_san_entries().is_some() {
```
arsing, line 172:
```diff
-if some_option.is_some() { some_option.expect("...").iter().for_each(...); }`
+some_option.iter().flatten().for_each("...");
```

arsing (changes requested):

edgelet/edgelet-http-workload/src/server/cert/mod.rs
massand: `flatten()` doesn't work because the option is `Option<&[String]>` and iter is `Iter<&[String]>`. Flatten converts it to a `&&[String]` which isn't an `Iterator`. Any other suggestions to extract the `[String]` from an `Option<&[String]>` inline?

arsing: If it's an `Option<&T>` then use `.into_iter()` instead of `.iter()`

arsing (approved):

PR #3859, opened 2020-10-30
1.0.10: Arm32/Arm64 base image update

and-rewsmith:
Updating base images for security vulnerability

### DIFF
--- a/edge-agent/docker/linux/arm32v7/Dockerfile
+++ b/edge-agent/docker/linux/arm32v7/Dockerfile
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.5.3-linux-arm32v7
+ARG base_tag=1.0.5.4-linux-arm32v7
 FROM azureiotedge/azureiotedge-agent-base:${base_tag}
 
 ARG EXE_DIR=.
--- a/edge-agent/docker/linux/arm32v7/base/Dockerfile
+++ b/edge-agent/docker/linux/arm32v7/base/Dockerfile
@@ -1,4 +1,4 @@
-ARG base_tag=3.1.7-bionic-arm32v7
+ARG base_tag=3.1.9-bionic-arm32v7
 FROM mcr.microsoft.com/dotnet/core/aspnet:${base_tag}
 
 RUN apt-get update && \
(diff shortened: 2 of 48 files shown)


edge-agent/docker/linux/arm32v7/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG base_tag=1.0.5.3-linux-arm32v7
```
and-rewsmith, line 1: Bumping by 2 here to keep versions consistent

edge-agent/docker/linux/arm32v7/Dockerfile
```diff
@@ -1,4 +1,4 @@
...
+ARG base_tag=1.0.5.4-linux-arm32v7
```
yophilav, line 1: The following images are not being updated correctly: 
```
MetricsValidator
ModuleRestarter
NumberLogger
Relayer
TestResultCoordinator
TwinTester
```

yophilav (changes requested):

edge-agent/docker/linux/arm32v7/Dockerfile
and-rewsmith: Good catch I only updated the modules where the base images lived.

yophilav: After you do the changes, could you please rebuild the whole thing?

and-rewsmith: I would expect the list to be more similar to this from the master PR:
```
iotedge-diagnostics-dotnet
MetricsCollector
NetworkController
CloudToDeviceMessageTester
DeploymentTester
DirectMethodReceiver
DirectMethodSender
EdgeHubRestartTester
load-gen
MetricsValidator
ModuleRestarter
NumberLogger
Relayer
TemperatureFilter
TestResultCoordinator
TwinTester
```

yophilav: Why is `NumberLogger` still using `1.0.6.2` ?

yophilav (changes requested):

edge-agent/docker/linux/arm32v7/Dockerfile
and-rewsmith: It was probably an error missed in the last update. I will bump

yophilav (approved):

PR #3865, opened 2020-11-02
Fix subscription issues

ancaantochi:
- remove from topic_mappers when a unsub is received 
- insert top topic mappers updates before sending subscription

### DIFF
--- a/mqtt/mqtt-bridge/src/messages.rs
+++ b/mqtt/mqtt-bridge/src/messages.rs
@@ -89,9 +89,7 @@ impl<S> StoreMqttEventHandler<S> {
     }
 
     fn update_unsubscribed(&mut self, sub: &str) {
-        if self.topic_mappers_updates.contains_key(sub) {
-            self.topic_mappers.remove(sub);
-        } else {
+        if self.topic_mappers.remove(sub).is_none() {
             warn!("unexpected subscription/rejected ack for {}", sub);
         };
     }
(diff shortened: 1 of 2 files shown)


mqtt/mqtt-bridge/src/pump/messages.rs
```diff
@@ -98,28 +98,25 @@ where
...
-                        match subscribe_result {
-                            Ok(_) => match sub.to_owned().try_into() {
-                                Ok(mapper) => {
-                                    self.topic_mappers_updates.insert(&subscribe_to, mapper);
-                                }
-                                Err(e) => {
-                                    error!("topic rule could not be parsed {}. {}", subscribe_to, e)
+                        match sub.to_owned().try_into() {
+                            Ok(mapper) => {
+                                self.topic_mappers_updates.insert(&subscribe_to, mapper);
```
dmolokanov, line 104: should we insert only when sub succeeded?

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -89,9 +89,7 @@ impl<S> StoreMqttEventHandler<S> {
     }
 
     fn update_unsubscribed(&mut self, sub: &str) {
-        if self.topic_mappers_updates.contains_key(sub) {
-            self.topic_mappers.remove(sub);
-        } else {
+        if self.topic_mappers.remove(sub).is_none() {
```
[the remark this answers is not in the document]
ancaantochi: because it is removing the update when the unsubscribe event was sent and by the time it got the ack it was already not there and not removing from topic_mappers

mqtt/mqtt-bridge/src/pump/messages.rs
ancaantochi: it means that the subscribe event was sent to the channel, but was still it doesn't mean it was sent to server

dmolokanov (approved):
LGTM

PR #3869, opened 2020-11-02
Send subscriptions updates

ancaantochi:
mqtt3 change to send subscriptions even if it is already in the current subscrptions


mqtt/mqtt3/src/client/subscriptions.rs
```diff
@@ -251,7 +251,7 @@ impl State {
                 }
             }
 
-            let mut target_subscriptions = current_subscriptions.clone();
+            let mut target_subscriptions = std::collections::BTreeMap::new();
```
vipeller, line 255: why is it an empty collection? there is that second loop starting in 282 line, it checks if there is any item in the current_subscription that is no longer in the target_subscription. The original implementation copied the current_subscriptions, so if no item was removed, then the target_subscription had all items (and it generated no pending_unsubscription). Starting with an empty collection, that loop will find all existing subscription missing, and will generate pending_unsubscription for them?

ancaantochi: thanks! changed because it was not handling correctly the unsub

vipeller (approved):

mqtt/mqtt3/src/client/subscriptions.rs
```diff
@@ -223,66 +223,42 @@ impl State {
...
+                pending_subscriptions.push_back(crate::proto::SubscribeTo {
+                    topic_filter: topic_filter.clone().into_owned(),
+                    qos,
+                });
             }
 
             let mut pending_unsubscriptions: std::collections::VecDeque<_> = Default::default();
-            for topic_filter in current_subscriptions.keys() {
-                if !target_subscriptions.contains_key(topic_filter) {
-                    pending_unsubscriptions.push_back(topic_filter.clone().into_owned());
```
and-rewsmith, line 284: I don't really understand this original logic here. It is saying that for a subscription waiting for an ack to come back, if it does not have a matching subscription waiting to get sent, then it will try to unsubscribe?

and-rewsmith (commented):
The new logic makes sense to me but I had a question about the old logic

PR #3874, opened 2020-11-03
Fixed on-behalf-of authorization

vadim-kovalyov:
When we introduced on-behalf-of authorization logic, we had a wrong assumption about iothub topic structure.
- expected: `$iothub/clients/<device_id/modules/<module_id>/twin/get`
- actual: `$iothub/<device_id/<module_id>/twin/get`

Since the actual format does not contain useful "markers" (`clients`, `modules`), we can't simply split the string and extract device id and module id. We have to use regex now. As a result, the current logic for validation and caching of iothub topics is no longer needed, b/c Regex covers it.

Also, two more checks were added for the on-behalf-of scenario: 
 - check that actor_id is in the auth chain for current edgehub.
 - check that client_id is in the auth chain for current edgehub.

### DIFF
--- a/mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
+++ b/mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
@@ -1,5 +1,7 @@
 use std::{any::Any, cell::RefCell, collections::HashMap, error::Error as StdError, fmt};
 
+use lazy_static::lazy_static;
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use tracing::info;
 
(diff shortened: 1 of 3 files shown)


vipeller (approved):
I focused mainly on the rules

mqtt/mqtt-edgehub/src/auth/authorization/edgehub.rs
```diff
@@ -141,67 +107,97 @@ where
...
+            None => false,
         }
     }
 }
 
-fn get_on_behalf_of_id(topic: &str) -> Option<ClientId> {
-    // topics without the new topic format cannot have on_behalf_of_ids
-    if !topic.starts_with("$iothub/clients") {
-        return None;
+fn get_actor_id(topic: &str) -> Option<ClientId> {
```
dmolokanov, line 166: nit: `extract_client_id` or `extract_actor_id`. you added a notion of an actor which is not common in our codebase. I was a bit confused by this naming

PR #3881, opened 2020-11-03
Apply log info from config.yaml

huguesBouvier:
1. Apply login info from config.yaml when pulling parent image
2. Fix connectivity test. Proxy should only be applied for MQTT
3. Added possibility to customize the address for the diagnostic image.

TEST:
Pulled parent image from a repo with password and from nested edge that doesn't require password
did iotedge check behind a proxy and not behind a proxy
Tried pulling diagnostic image with a custom address field.


edgelet/iotedge/src/check/checks/container_connect_upstream.rs
```diff
@@ -139,16 +139,18 @@ impl ContainerConnectUpstream {
...
 
-        let proxy = settings
-            .agent()
-            .env()
-            .get("https_proxy")
-            .map(std::string::String::as_str);
-        self.proxy = proxy.map(ToOwned::to_owned);
-        if let Some(proxy) = proxy {
-            args.extend(&["--proxy", proxy]);
+        if port == 443.to_string() {
```
lfitchett, line 142: You could also do 
```suggestion
        if &port == "443" {
```

lfitchett: If a customer decides to use a port other than 443 for https this check wouldn't run, but I think that's probably fine, I doubt many customers are doing that.

edgelet/iotedge/src/check/checks/pull_agent_from_upstream.rs
```diff
@@ -36,6 +36,20 @@ impl PullAgentFromUpstream {
...
+        (&settings.agent().config().auth().and_then(docker::models::AuthConfig::username), 
+        &settings.agent().config().auth().and_then(docker::models::AuthConfig::password),
+        &settings.agent().config().auth().and_then(docker::models::AuthConfig::serveraddress))
+        {
+            super::docker(
+                docker_host_arg,
+                vec!["login", serveraddress, "-p", password, "--username", username],
+            )
+            .map_err(|(_, err)| err)
+            .context("Failed to get edge Agent image")?;
```
lfitchett, line 50:
```suggestion
            .context(&format!("Failed to login to {}", serveraddress))?;
```

edgelet/iotedge/src/check/checks/pull_agent_from_upstream.rs
```diff
@@ -36,6 +36,20 @@ impl PullAgentFromUpstream {
             return Ok(CheckResult::Skipped);
         };
 
+
+        if let (Some(username), Some(password), Some(serveraddress)) =
```
lfitchett, line 40:
```suggestion
        if let (Some(username), Some(password), Some(server_address)) = 
```

edgelet/iotedge/src/check/checks/container_connect_upstream.rs
```diff
@@ -139,16 +139,18 @@ impl ContainerConnectUpstream {
...
+                .env()
+                .get("https_proxy")
+                .map(std::string::String::as_str);
+            self.proxy = proxy.map(ToOwned::to_owned);
+            if let Some(proxy) = proxy {
+                args.extend(&["--proxy", proxy]);
+            }
         }
-
+
```
lfitchett, line 153: remove trailing whitespace

lfitchett (changes requested):
Couple minor nits

edgelet/iotedge/src/check/checks/container_connect_upstream.rs
lfitchett: make sure to run `cargo fmt --all` and `cargo clippy --tests` before checking in.

huguesBouvier: Yes I realized I didn't fmt the last bit. Since I run test on VM I don't always have it automated

edgelet/iotedge/src/check/checks/container_connect_upstream.rs
huguesBouvier: Yes, I though about that but I am not sure you can change that port in edgeAgent.

lfitchett (approved):
LGTM

PR #3885, opened 2020-11-03
[mqtt] Fix flaky tests for command handler

dmolokanov:
Fix to make CI build to succeed. 

`CommandHandler` tests used to rely on `init` method which subscribed to required topics and waited for responses.
Since the body of `init` moved to the `run` method instead, tests became flaky because they lost an ability to start verification only when `CommandHandler` initialized.

Changes were made to use another mqtt client to subscribe to the broker and wait for such an event emitted by broker.


vadim-kovalyov (approved):

dmolokanov:
Closing in favor of #3890

PR #3892, opened 2020-11-04
Construct valid auth-chains for children for non-nested mode

richma-ms:
When nested edge is turned off, there's no concept of auth-chains because all identities are immediate children of the Edge device.  But edgehub authorizer (in the broker) needs auth-chains to work even in the single-node configuration, so this change adds logic to construct simple auth chains for the children in the legacy code path.


vadim-kovalyov (approved):

varunpuranik (approved):
Change looks good. We need Unit tests, but that can be a follow up commit. 
Why was this not caught in our automated tests?

vadim-kovalyov:
> Change looks good. We need Unit tests, but that can be a follow up commit.
> Why was this not caught in our automated tests?

@varunpuranik, this surfaced after my yesterday fix. Before, no one checked auth chain for non on-beahlf-of requests. I added that check, but did not know that auth chain is empty if nested edge flag is off, so this bug surfaced today during my tests. We discussed with @richma-ms and agreed that it would be easier and proper to just always return auth chain no matter the state of the nested edge flags.

PR #3893, opened 2020-11-04
bridge topics fix when inPrefix or outPrefix are empty

ancaantochi:
when inPrefix or outPrefix are empty string it should work as None.

Note: I think would be nicer just to deserialize "" as None, but couldn't make it work with custom deserialization when field was missing. For now will go with this fix and will try in the meantime with custom deserialization

### DIFF
--- a/mqtt/mqtt-bridge/src/messages.rs
+++ b/mqtt/mqtt-bridge/src/messages.rs
@@ -65,17 +65,23 @@ impl<S> StoreMqttEventHandler<S> {
                 // maps if local does not have a value it uses the topic that was received,
                 // else it checks that the received topic starts with local prefix and removes the local prefix
                 .map_or(Some(topic_name), |local_prefix| {
-                    let prefix = format!("{}/", local_prefix);
-                    topic_name.strip_prefix(&prefix)
+                    if local_prefix.is_empty() {
+                        topic_name.strip_prefix(local_prefix)
+                    } else {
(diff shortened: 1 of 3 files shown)


varunpuranik (approved):

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -65,17 +65,23 @@ impl<S> StoreMqttEventHandler<S> {
                 // maps if local does not have a value it uses the topic that was received,
                 // else it checks that the received topic starts with local prefix and removes the local prefix
                 .map_or(Some(topic_name), |local_prefix| {
-                    let prefix = format!("{}/", local_prefix);
-                    topic_name.strip_prefix(&prefix)
+                    if local_prefix.is_empty() {
```
dmolokanov, line 68: just as a suggestion: can we make a new type `NonEmptyString` instead and pass it everywhere? with struct API similar to standard https://doc.rust-lang.org/std/num/struct.NonZeroUsize.html

dmolokanov: it this case you'll make sure it is not empty, and whenever you need a string you can extract it. as a convenience, you can make `fn as_str(&self) -> &str` and `impl AsRef<str> for NonEmptyString`

dmolokanov (approved):
LGTM. although we could do a bit better and reduce code duplication with `NonEmptyString` type

PR #3894, opened 2020-11-04
fixed direct method calls for modules

vipeller:
modules were not handled for direct method calls


varunpuranik (approved):
Pls add a test in a follow up

PR #3897, opened 2020-11-04
[mqtt] Fix certificate renewal timer

dmolokanov:
When the server certificate expiration date is longer than 2y interval `tokio::time::delay_for` panics when try to schedule waiting.

`tokio::time::delay_for` and `tokio::time::delay_until` has a bug that prevents to schedule a very long timeout (more than 2 years). To mitigate this issue we split a long interval by the number of small intervals and schedule small intervals and waits for a small one until we reach a target.

### DIFF
--- a/mqtt/mqttd/src/app/edgehub.rs
+++ b/mqtt/mqttd/src/app/edgehub.rs
@@ -11,7 +11,6 @@ use futures_util::{
     future::{self, Either},
     FutureExt,
 };
-use tokio::time;
 use tracing::{debug, error, info};
 
 use mqtt_bridge::{settings::BridgeSettings, BridgeController};
@@ -287,8 +286,8 @@ async fn server_certificate_renewal(renew_at: DateTime<Utc>) {
             "scheduled server certificate renewal timer for {}",
             renew_at
         );
-        let delay = std::time::Duration::from_secs(3600);
-        time::delay_for(delay).await;
+        let delay = delay.to_std().expect("duration must not be negative");
+        crate::time::sleep(delay).await;
 
         info!("restarting the broker to perform certificate renewal");
     } else {
--- a/mqtt/mqttd/src/time.rs
+++ b/mqtt/mqttd/src/time.rs
@@ -1,0 +1,57 @@
+//! `tokio::time::delay_for` and `tokio::time::delay_until` has a bug that
+//! prevents to schedule a very long timeout (more than 2 years).
+//! To mitigate this issue we split a long interval by the number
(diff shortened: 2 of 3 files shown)


mqtt/mqttd/src/app/edgehub.rs
```diff
@@ -287,8 +286,8 @@ async fn server_certificate_renewal(renew_at: DateTime<Utc>) {
             "scheduled server certificate renewal timer for {}",
             renew_at
         );
-        let delay = std::time::Duration::from_secs(3600);
```
dmolokanov, line 290: this is the wrong change that slipped in the previous PR

mqtt/mqttd/src/time.rs
```diff
@@ -0,0 +1,57 @@
+//! `tokio::time::delay_for` and `tokio::time::delay_until` has a bug that
+//! prevents to schedule a very long timeout (more than 2 years).
```
varunpuranik, line 2: How confident are you with this change? If it is risky, we can pull it in after public preview.

varunpuranik (approved):

mqtt/mqttd/src/time.rs
```diff
@@ -0,0 +1,57 @@
+//! `tokio::time::delay_for` and `tokio::time::delay_until` has a bug that
```
vadim-kovalyov, line 1: why not intro a simple while loop like that
```
while renew_at > 0 
do
 interval = min(renew_at, 30d)
 delay_for(interval)
 renew_at -= interval
end

dmolokanov: Yes, we could do that. I thought we may need more general solution

mqtt/mqttd/src/time.rs
dmolokanov: since we have a relatively small interval `30d` I believe it should work fine

mqtt/mqttd/src/time.rs
varunpuranik: Yes, we have similar logic in EdgeHub as well - https://github.com/Azure/iotedge/blob/master/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/CertificateRenewal.cs

dmolokanov: it is coupled with certificates renewal, and we may need more general solution for "long sleep" action. 

since async/await is a generated by the compiler the type will be harder to understand when awaiting sleep and something else at the same time. 
in the end we have the same logic as @vadim-kovalyov mentioned but written with future rather than async method.

PR #3905, opened 2020-11-05
Speedup persist proptest

vadim-kovalyov:
I found that most of the time is spent in input generation, so I reduced it.

On my box this reduced persistence proptests from from ~72s to ~14s.


dmolokanov (approved):
Wow! Love that!

Issue #3795, opened 2020-10-23
New edgeAgent Direct Method GetModuleLogs returns empty payload when management/workload URIs are http-based


ryayl:
## Expected Behavior
The new direct method GetModuleLogs (provided by edgeAgent 1.0.10) is working correctly on a RaspberryPi based system, where the iotedge config file has the following settings for both connect and listen:

```
management_uri: "unix:///var/run/iotedge/mgmt.sock"
workload_uri: "unix:///var/run/iotedge/workload.sock"
```

After sending GetModuleLogs command from the Azure CLI, I get a response including the payload containing the logs. Meanwhile I can observe the following activity in the edgeAgent and edgeDaemon logs:

- edgeAgent receives direct method and generates http request, getting the response: 

```
<7> 2020-10-23 21:47:49.140 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Agent.Edgelet.ModuleManagementHttpClient] - Making a Http call to unix:///var/run/iotedge/mgmt.sock to Get logs for CameraApp
<7> 2020-10-23 21:47:49.141 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Connecting socket /var/run/iotedge/mgmt.sock
<7> 2020-10-23 21:47:49.142 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Connected socket /var/run/iotedge/mgmt.sock
<7> 2020-10-23 21:47:49.142 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Sending request http://mgmt.sock/modules/CameraApp/logs?api-version=2019-11-05&follow=false&tail=100
<7> 2020-10-23 21:47:49.158 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Util.Uds.HttpUdsMessageHandler] - Response received OK
<7> 2020-10-23 21:47:49.158 +00:00 [DBG] [Microsoft.Azure.Devices.Edge.Agent.Edgelet.ModuleManagementHttpClient] - Received a valid Http response from unix:///var/run/iotedge/mgmt.sock for Get logs for CameraApp
```

- edgeDaemon receives the http request and sends the response:
```
Oct 23 22:55:57 rypi iotedged[1099]: 2020-10-23T21:55:57Z [INFO] - Getting logs for module CameraApp...
[...]

PR #3908, opened 2020-11-05
GetModuleLogs works with http endpoints

lfitchett:
This was a simple fix. 

~~The new e2e tests currently do not test the http configuration for mgmt and workload sockets, so there may be a follow up to add tests.~~
Since http is not officially supported for mgmt and workload sockets, no tests will be added.


damonbarry (approved):

damonbarry:
Fixes #3795.

PR #3911, opened 2020-11-05
Adjust manifest sample for using module startupOrder

micahl:
Proposing these changes to version #'s used since my understanding is the feature was part of the 1.0.10 release (of edgeAgent, right?) and only recognized when using schemaVersion of at least 1.1.


doc/ModuleStartupOrder.md
```diff
@@ -58,7 +58,7 @@ The following sample deployment manifest illustrates how startupOrder values of
           "edgeAgent": {
             "type": "docker",
             "settings": {
-              "image": "mcr.microsoft.com/azureiotedge-agent:1.0",
+              "image": "mcr.microsoft.com/azureiotedge-agent:1.0.10",
```
yophilav, line 61: If the release process is done correctly, the docker image tag `1.0.10.1` and `1.0` should be synonymous. This change is uncessary.

yophilav, line 61: If the release process is done correctly, the docker image tag `1.0.10.1` and `1.0` should be synonymous. This change is unnecessary.

doc/ModuleStartupOrder.md
```diff
@@ -116,7 +116,9 @@ The following sample deployment manifest illustrates how startupOrder values of
 
 In the sample deployment manifest shown above:
 
-* The `$edgeHub` module has been assigned a `startupOrder` value of 0.
+* The '$edgeAgent' schemaVersion has been set to 1.1 (or later).
+* The 'edgeAgent' module's image is 1.0.10 (or later).
```
yophilav, line 120: Per comment above, there is no need to update this image version.

doc/ModuleStartupOrder.md
```diff
@@ -116,7 +116,9 @@ The following sample deployment manifest illustrates how startupOrder values of
...
+* The `edgeHub` module has been assigned a `startupOrder` value of 0.
```
yophilav, line 121: @richma-ms The `startupOrder` of `edgeAgent` is also 0, right? We can make an explicit comment here as well.

micahl: There's a note given in an earlier section of the doc describing the behavior for edgeAgent.  I'll reiterate it here as well.

doc/ModuleStartupOrder.md
micahl: Ok, will remove.

doc/ModuleStartupOrder.md
micahl: OK, removing this change.

yophilav (approved):

PR #3914, opened 2020-11-05
change edgehub status to align with iot

huguesBouvier:
change edgehub status to align with iot.
Cherry pick: https://github.com/Azure/iotedge/pull/3915.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
@@ -2,6 +2,7 @@
 namespace Microsoft.Azure.Devices.Edge.Hub.Http.Controllers
 {
     using System;
+    using System.Collections.Generic;
     using System.Net;
     using System.Text;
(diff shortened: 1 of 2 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -212,7 +212,7 @@ public class RegistryController : Controller
 
                 if (!await this.AuthenticateAsync(deviceId, Option.None<string>(), Option.None<string>()))
                 {
-                    await this.SendResponseAsync(HttpStatusCode.Unauthorized);
+                    await this.SendResponseAsync(HttpStatusCode.BadRequest);
```
varunpuranik, line 215: Is this correct? If the entire request is correct but the token is invalid, IoTHub returns 400?

huguesBouvier: What I did here was wrong. I had 2 PR. Forgot to put this on the master master PR.

richma-ms (approved):

PR #3915, opened 2020-11-05
change edgehub status to align with iot

huguesBouvier:
change edgehub status to align with iot

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
@@ -212,7 +212,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Http.Controllers
 
                 if (!await this.AuthenticateAsync(deviceId, Option.None<string>(), Option.None<string>()))
                 {
-                    await this.SendResponseAsync(HttpStatusCode.Unauthorized);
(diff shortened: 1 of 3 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
```diff
@@ -513,10 +522,18 @@ async Task<bool> AuthenticateAsync(string deviceId, Option<string> moduleId, Opt
             return false;
         }
 
-        async Task SendResponseAsync(HttpStatusCode status, string jsonContent = "")
+        async Task SendResponseAsync(HttpStatusCode status, string jsonContent = "", Dictionary<string, string> headers = null)
         {
             this.Response.StatusCode = (int)status;
 
+            if (headers != null)
```
richma-ms, line 529: >if (headers != null) [](start = 12, length = 20)

the team doesn't like null checks, you can avoid this by having an overload instead that takes (HttpStatus, string), and calls this with an empty Dictionary instead of null.

richma-ms (dismissed):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/RegistryController.cs
huguesBouvier: fixed

richma-ms (approved):

PR #3916, opened 2020-11-05
mqtt broker align env variables

ancaantochi:
replace : with __

Todo: make the change in master to only use new env variables format in broker without converting


varunpuranik (approved):

vadim-kovalyov (approved):

vadim-kovalyov:
@ancaantochi please port it to master as well, if you haven't done so yet. Thanks!

PR #3919, opened 2020-11-06
Fix default port and remove inexistent env var

ebertrams:


edge-modules/api-proxy-module/readme.md
```diff
@@ -113,7 +113,7 @@ Next, set each environment variable's value by listing them directly.
 
 | Environment variable  | comments |
 | ------------- |  ------------- |
-| NGINX_DEFAULT_PORT  | Changes the port Nginx listens too. If you change this option, make sure that the port you select is exposed in the dockerfile. Default is 443  |
+| NGINX_DEFAULT_PORT  | Changes the port Nginx listens too. If you update this environment variable, make sure the port you select is also exposed in the module dockerfile and the port binding. Default is 8000.  |
```
varunpuranik, line 116:
```suggestion
| NGINX_DEFAULT_PORT  | Changes the port Nginx listens on. If you update this environment variable, make sure the port you select is also exposed in the module dockerfile and the port binding. Default is 8000.  |
```

varunpuranik: Default port seems to be 443 indeed. Why are we saying 8000?
https://github.com/Azure/iotedge/blob/master/edge-modules/api-proxy-module/docker/linux/amd64/Dockerfile

ebertrams: I think you re right. Not because of the docker file but because of this: https://github.com/Azure/iotedge/blob/f7eaf02a3054a1dc63fcabfc79760b1508632b02/edge-modules/api-proxy-module/src/monitors/config_monitor.rs#L24
Both 443 and 8000 are exposed by the docker file. The default marketplace setting uses 8000, whcih overrides the default 443 setting baked into the image. Will update.

varunpuranik: The idea of the nginx proxy is to multiplex over port 443... so..

varunpuranik: I know dockerfile exposes multiple ports. What I meant is it also exposes 443, which is the default.

huguesBouvier (approved):

PR #3924, opened 2020-11-06
fix for iotedge check

huguesBouvier:
fix for iotedge check.

Test performed:
With proxy nested
On golden path setup, lvl 3: Ran iotedge check with new diagnostic image:
× container on the default network can connect to upstream HTTPS / WebSockets port - Error
× container on the IoT Edge module network can connect to upstream HTTPS / WebSockets port - Error

Upgrade lvl 4 with new edgehub that returns new header:
√ host can connect to and perform TLS handshake with upstream HTTPS / WebSockets port - OK
√ container on the IoT Edge module network can connect to upstream HTTPS / WebSockets port - OK

Without proxy nested
On golden path setup, lvl 4: Ran iotedge check with new diagnostic image:
× container on the default network can connect to upstream HTTPS / WebSockets port - Error
× container on the IoT Edge module network can connect to upstream HTTPS / WebSockets port - Error

Upgrade lvl 5 with new edgehub that returns new header:
√ host can connect to and perform TLS handshake with upstream HTTPS / WebSockets port - OK
√ container on the IoT Edge module network can connect to upstream HTTPS / WebSockets port - OK

With proxy not nested (root)
 container on the default network can connect to upstream HTTPS / WebSockets port - OK
√ container on the IoT Edge module network can connect to upstream HTTPS / WebSockets port - OK

### DIFF
--- a/edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
+++ b/edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
@@ -74,8 +74,9 @@ namespace Diagnostics
         static async Task Upstream(string hostname, string port, string proxy)
         {
             var httpClientHandler = new HttpClientHandler();
-            httpClientHandler.ServerCertificateCustomValidationCallback = (message, cert, chain, sslPolicyErrors) => {
-                    return true;   //Is valid
+            httpClientHandler.ServerCertificateCustomValidationCallback = (message, cert, chain, sslPolicyErrors) =>
+            {
+                    return true; // Is valid
             };
 
             if (proxy != null)
@@ -83,15 +84,15 @@ namespace Diagnostics
                 Environment.SetEnvironmentVariable("https_proxy", proxy);
             }
 
-                var httpClient = new HttpClient(httpClientHandler);
+            var httpClient = new HttpClient(httpClientHandler);
             var logsUrl = string.Format("https://{0}/devices/0000/modules", hostname);
             var httpRequest = new HttpRequestMessage(HttpMethod.Get, logsUrl);
             HttpResponseMessage httpResponseMessage = await httpClient.SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead);
 
             var keys = httpResponseMessage.Headers.GetValues("iothub-errorcode");
-            if (keys.Contains("InvalidProtocolVersion") == false)
+            if (!keys.Contains("InvalidProtocolVersion"))
             {
-                throw new Exception($"Got bad response");
+                throw new Exception($"Wrong value for iothub-errorcode header");
             }
         }
 
--- a/edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
+++ b/edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
@@ -40,6 +40,7 @@ namespace Diagnostics
 
         static async Task MainAsync(string[] args)
         {
+            await Upstream("192.168.0.183", "443", "http://10.0.0.0");
             var config = new ConfigurationBuilder().AddCommandLine(args).Build();
             switch (args[0])
             {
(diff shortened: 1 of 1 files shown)


edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -85,52 +73,31 @@ static async Task EdgeAgent(string managementUri)
...
         static async Task Upstream(string hostname, string port, string proxy)
         {
+            var httpClientHandler = new HttpClientHandler();
+            httpClientHandler.ServerCertificateCustomValidationCallback = (message, cert, chain, sslPolicyErrors) => {
+                    return true;   //Is valid
+            };
+
             if (proxy != null)
             {
-                Uri proxyUri = new Uri(proxy);
```
lfitchett, line 90: It might be better to just have edgelet set this as an env variable in docker instead of a parameter, but if the rust changes are waiting until rc2 this can be changed then.

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -85,52 +73,31 @@ static async Task EdgeAgent(string managementUri)
 
         static async Task Upstream(string hostname, string port, string proxy)
         {
+            var httpClientHandler = new HttpClientHandler();
+            httpClientHandler.ServerCertificateCustomValidationCallback = (message, cert, chain, sslPolicyErrors) => {
+                    return true;   //Is valid
```
lfitchett, line 78: why do you override this? shouldn't we be checking if the client trusts the proxy? Or is it b/c the diagnostics module doesn't retrieve the trust bundle?

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -85,52 +73,31 @@ static async Task EdgeAgent(string managementUri)
...
             {
-                Uri proxyUri = new Uri(proxy);
-                IProxyClient proxyClient = MakeProxy(proxyUri);
+                Environment.SetEnvironmentVariable("https_proxy", proxy);
+            }
 
-                // Setup timeouts
-                proxyClient.ReceiveTimeout = (int)TimeSpan.FromSeconds(60).TotalMilliseconds;
-                proxyClient.SendTimeout = (int)TimeSpan.FromSeconds(60).TotalMilliseconds;
+                var httpClient = new HttpClient(httpClientHandler);
```
lfitchett, line 86: nit: spacing

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -85,52 +73,31 @@ static async Task EdgeAgent(string managementUri)
...
+            var httpRequest = new HttpRequestMessage(HttpMethod.Get, logsUrl);
+            HttpResponseMessage httpResponseMessage = await httpClient.SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead);
 
-                // Get TcpClient to futher work
-                var client = proxyClient.CreateConnection(hostname, int.Parse(port));
-                client.GetStream();
-            }
-            else
+            var keys = httpResponseMessage.Headers.GetValues("iothub-errorcode");
+            if (keys.Contains("InvalidProtocolVersion") == false)
```
lfitchett, line 92:
```suggestion
            if (!keys.Contains("InvalidProtocolVersion"))
```

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -85,52 +73,31 @@ static async Task EdgeAgent(string managementUri)
...
             {
-                TcpClient client = new TcpClient();
-                await client.ConnectAsync(hostname, int.Parse(port));
-                client.GetStream();
+                throw new Exception($"Got bad response");
```
lfitchett, line 94: It might be good to be more specific here. Something like "The client did not receive the expected response from hub

lfitchett (approved):
Couple minor things, but overall fine

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
huguesBouvier: Otherwise we need to extract the trusted CA from config.yaml. I did that because it is not the focus of this test and it is faster.

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
huguesBouvier: Fixed

lfitchett (approved):

edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
varunpuranik: +1 on that

PR #3938, opened 2020-11-10
Fixed an issue where not all variable rules are evaluated

vadim-kovalyov:
**Scenario:**
 - Have a policy with a custom identity matcher.
 - Have two variable rules (deny and allow) for an identity, such that  both rules match a given request identity.
 - But the two rules must be different in resources.
 - The deny rule resources do not match the request.
 - The allow rule resources do match the request.

Expected: "allow" the request.

**Fix:**
 - The problem was in early return from a `eval_variable_rules` func, instead of going through all variable rules that match the identity.

### DIFF
--- a/mqtt/policy/src/core/mod.rs
+++ b/mqtt/policy/src/core/mod.rs
@@ -130,7 +130,7 @@ where
                                 result = effect;
                             }
                         }
-                        // continue to look for other variable rules
+                        // continue to look for other identity variable rules
                         // if no resources matched the current one.
                         if result == &EffectOrd::undefined() {
                             continue;
@@ -714,9 +714,10 @@ pub(crate) mod tests {
     /// - Have two variable rules (deny and allow) for an identity, such that
     ///   both rules match a given request identity.
     /// - But the two rules must be different in resources.
+    /// - Make a request to the allowed resource.
     /// - The deny rule resources do not match the request.
     /// - The allow rule resources do match the request.
-    /// - Expected: "allow" the request.
+    /// - Expected: request allowed.
     ///
     /// This case is created as a result of a discovered bug.
     #[test]
--- a/mqtt/policy/src/core/mod.rs
+++ b/mqtt/policy/src/core/mod.rs
@@ -83,8 +83,8 @@ where
             Some(operations) => match operations.0.get(&request.operation) {
                 // operation exists.
                 Some(resources) => {
-                    // Iterate over and match resources.
-                    // We need to go through all resources and find one with highest priority (smallest order).
+                    // iterate over and match resources.
+                    // we need to go through all resources and find one with highest priority (smallest order).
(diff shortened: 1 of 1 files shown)


mqtt/policy/src/core/mod.rs
```diff
@@ -702,25 +709,130 @@ pub(crate) mod tests {
         assert_matches!(policy.evaluate(&request), Ok(Decision::Allowed));
     }
 
-    /// `TestSubstituter` replaces any value with the corresponding identity or resource
+    /// Scenario:
+    /// - Have a policy with a custom identity matcher
+    /// - Have two variable rules (deny and allow) for an identity, such that
+    ///   both rules match a given request identity.
+    /// - But the two rules must be different in resources.
```
and-rewsmith, line 716: I think we need to say after this line something like:
"We make a request to the allowed resource"

It isn't clear reading this what the resource requested is.

mqtt/policy/src/core/mod.rs
```diff
@@ -702,25 +709,130 @@ pub(crate) mod tests {
...
+            .with_default_decision(Decision::Denied)
+            .with_substituter(TestIdentitySubstituter)
+            .with_matcher(DefaultResourceMatcher)
+            .build()
+            .expect("Unable to build policy from json.");
+
+        let request = Request::new("actor_a", "write", "hello/a").unwrap();
+
+        assert_matches!(policy.evaluate(&request), Ok(Decision::Allowed));
+    }
```
and-rewsmith, line 813: nit: we could probably eliminate duplicated code by paramaterizing test.

mqtt/policy/src/core/mod.rs
```diff
@@ -115,8 +115,8 @@ where
                 return match operations.0.get(&request.operation) {
                     // operation exists.
                     Some(resources) => {
-                        // Iterate over and match resources.
-                        // We need to go through all resources and find one with highest priority (smallest order).
+                        // iterate over and match resources.
```
dmolokanov, line 118: nit: a bit complex function with multiple nested loops and a bunch of nested levels. any chance we can improve that?

dmolokanov (dismissed):
one nit. otherwise lgtm.

mqtt/policy/src/core/mod.rs
vadim-kovalyov: good point, but since json definition is also different, I thought that it wouldn't look that great to have it parametrized via `test_case`. So I just made different tests.

and-rewsmith (approved):

mqtt/policy/src/core/mod.rs
vadim-kovalyov: I tried extracting funcs, can't say it looks better that way. That's why I put as much inline comments as I could.

and-rewsmith (approved):

PR #3939, opened 2020-11-10
Add type to IS queries

gordonwang0:
[This change](https://github.com/Azure/iot-identity-service/commit/1ed1cb1a6144c5562c9a3bb00a5d7b5fc66b8f04) to aziot updates the Identity Service APIs to require a `type` parameter.

This PR updates IS calls to include the type parameter.


edgelet/identity-client/src/client.rs
```diff
@@ -120,7 +120,7 @@ impl IdentityClient {
     ) -> Box<dyn Future<Item = aziot_identity_common::Identity, Error = Error> + Send> {
         let client = self.client.clone();
         let uri = format!(
-            "/identities/modules/{}?api-version={}",
+            "/identities/modules/{}?api-version={}&type=aziot&type=aziot",
```
massand, line 123: typo

edgelet/management/src/apis/identity_api.rs
```diff
@@ -63,6 +63,7 @@ where
 
         let query = ::url::form_urlencoded::Serializer::new(String::new())
             .append_pair("api-version", &api_version.to_string())
+            .append_pair("type", "aziot")
```
massand, line 66: These are code-gen and shouldn't be changed

edgelet/identity-client/src/client.rs
gordonwang0: Fixed

edgelet/management/src/apis/identity_api.rs
gordonwang0: Fixed

massand (approved):

PR #3945, opened 2020-11-10
MQTT: Use FQN for the authenticated identity

vadim-kovalyov:
When broker sends an authentication request to EH Core, it gets back the identity name without hub name prefix. As a result, customers have to use just device name in the policy engine, instead of FQN for device:

**Current:**
```
                        {
                            "identities": [
                                **"vakavali_leaf"**
                            ],
                            "allow": [
                                {
                                    "operations": [
                                        "mqtt:publish",
                                        "mqtt:subscribe"
                                    ],
                                    "resources": [
                                        "#"
                                    ]
                                }
                            ]
                        }
```

**Expected:**
```
                        {
                            "identities": [
                                **"mytesthub.azure-devices.net/vakavali_leaf"**
                            ],
                            "allow": [
                                {
                                    "operations": [
                                        "mqtt:publish",
                                        "mqtt:subscribe"
                                    ],
                                    "resources": [
                                        "#"
                                    ]
                                }
                            ]
                        }
```

**Notes:**
This is a quick fix solution. We may revisit it when we have story for local identities ready.

### DIFF
--- a/mqtt/mqtt-broker/src/auth/mod.rs
+++ b/mqtt/mqtt-broker/src/auth/mod.rs
@@ -84,3 +84,9 @@ impl PartialEq<ClientId> for Identity {
         self.as_str() == other.as_str()
     }
 }
+
+impl PartialEq<String> for &Identity {
+    fn eq(&self, other: &String) -> bool {
+        self.as_str() == other
+    }
+}
--- a/mqtt/mqtt-broker/src/auth/mod.rs
+++ b/mqtt/mqtt-broker/src/auth/mod.rs
@@ -17,8 +17,6 @@ use std::{
 
 use serde::{Deserialize, Serialize};
 
-use crate::ClientId;
-
 /// Authenticated MQTT client identity.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum AuthId {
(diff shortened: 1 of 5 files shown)


vipeller (dismissed):

mqtt/mqtt-broker/src/auth/mod.rs
```diff
@@ -84,3 +84,9 @@ impl PartialEq<ClientId> for Identity {
         self.as_str() == other.as_str()
     }
 }
+
+impl PartialEq<String> for &Identity {
```
dmolokanov, line 88:
```suggestion
impl PartialEq<T: AsRef<str>> for &Identity {
```
it will make ergonomics better

dmolokanov: impl PartialEq<T> for &Identity will be derived automatically

dmolokanov: Do we actually need this implementation? As far as I can see, you use it in one place. Should we create a ClientId instead of String?
```rust
let actor_id = ClientId::from(format!("{}/{}", self.iothub_id, activity.client_id()));
if identity == actor_id {
...
}

dmolokanov (dismissed):
LGTM

dmolokanov (approved):
:shipit:

PR #3946, opened 2020-11-10
Remove "precondition" task in checkin pipelines in favor of native filter

vadim-kovalyov:
In this PR I tried to remove "precondition" task in checkin pipelines in favor of native filter for paths in Azure Pipelines. This will save us some cycles.

If this change goes well, I'm planning to fix the CI pipelines the same way in the following PR.


builds/checkin/api-proxy.yaml
```diff
@@ -3,30 +3,18 @@ pr:
...
-            echo "Detected changes inside builds or edge-modules/api-proxy-module or mqtt/edgelet-client folders"
-            echo "##vso[task.setvariable variable=RUN_PIPELINE;isOutput=true]TRUE"
-          fi
-        displayName: Check changes in sources
-        name: check_files
+      - release/*
+  paths:
+    include:
+      - "edge-modules/api-proxy-module/*"
+      - "mqtt/edgelet-client/*"
```
dmolokanov, line 10: this looks weird. `edgelet-client` is a dependecy of `api-proxy-module`. so changes in the `edgelet-client` should not cause checks running for any other crate. we don't run checks when 3rd party dependency changed...

vadim-kovalyov: I would let @huguesBouvier argue this case :) I just converted the task. 

From my side, I understand what you're saying Denis, but it also makes to have this safeguard against breaking api-proxy. Someone may go and make breaking changes in `edgelet-client`, and fix the broker but forget about api-proxy...

dmolokanov (approved):
Great work! Love that we finally got rid of these annoying checks! Next step: GitHub Actions 🚀

huguesBouvier (approved):

builds/checkin/api-proxy.yaml
huguesBouvier: I believe if we make changes in edgelet-client we should check api proxy still compiles. Otherwise there is a risk of have a code that doesn't

huguesBouvier:
> Great work! Love that we finally got rid of these annoying checks! Next step: GitHub Actions 🚀

+1

PR #3947, opened 2020-11-11
removing "noise" made by API proxy

huguesBouvier:
API proxy is generating a lot of messages, this PR aims to improve that.

Fixes:
- API Proxy: Checking changes in root certificate every 1s is unecessary (https://msazure.visualstudio.com/One/_workitems/edit/8741548)
Now waits for the root certificate before starting poll on the server cert to check if it expires.

- ApiProxy sends reported properties every 5s (https://msazure.visualstudio.com/One/_workitems/edit/8682854)
That bit of code is now removed

- Module iot_edge_api_proxy sends mqtt pings every ~2 seconds. 
 (https://msazure.visualstudio.com/One/_workitems/edit/8718522)
Keep alive moved from 5s to 5min. Ping should happen every 2.5min.

### DIFF
--- a/edge-modules/api-proxy-module/src/monitors/certs_monitor.rs
+++ b/edge-modules/api-proxy-module/src/monitors/certs_monitor.rs
@@ -51,15 +51,13 @@ pub fn start(
             let wait_shutdown = shutdown_signal.notified();
             futures::pin_mut!(wait_shutdown);
 
-            match futures::future::select(time::delay_for(CERTIFICATE_POLL_INTERVAL), wait_shutdown)
-                .await
+            if let Either::Right(_) =
+                futures::future::select(time::delay_for(CERTIFICATE_POLL_INTERVAL), wait_shutdown)
+                    .await
             {
-                Either::Right(_) => {
-                    warn!("Shutting down certs monitor!");
-                    return Ok(());
-                }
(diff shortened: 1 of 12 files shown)


edge-modules/api-proxy-module/src/monitors/certs_monitor.rs
```diff
@@ -76,6 +77,27 @@ pub fn start(
...
+        notify_certs_rotated.notify();
+
+        //Loop to check if server certificate expired.
+        //It is implemented as a polling instead of a delay until certificate expiry, because clocks are unreliable.
+        //If the system clock gets readjusted while the task is sleeping, the system might wake up after the certificate expiry.
+        loop {
+            let wait_shutdown = shutdown_signal.notified();
+            futures::pin_mut!(wait_shutdown);
+
+            match futures::future::select(time::delay_for(CERTIFICATE_POLL_INTERVAL), wait_shutdown)
```
vadim-kovalyov, line 92: nit: it is a bit odd to use such constructs `Either::Left(_) => {}`. It is better to use `if let Either::Right(_) = ...` in that case...

vadim-kovalyov (approved):

vadim-kovalyov (approved):
:ship:

PR #3949, opened 2020-11-11
Enable package builds for Mariner

chzawist:
Builds azure-iotedge and libiothsm-std RPMs for the Mariner distribution.
Builds rust 1.45 from source for Mariner as long as the official Mariner repo does not satisfy iotedge's rust>=1.40 constraint.


arsing:
`sysinfo` must not be downgraded.

edgelet/build/linux/package-mariner.sh
```diff
@@ -0,0 +1,551 @@
...
+curl "https://marineriotedge.file.core.windows.net/mariner-build-env/toolkit-1.0.20201029-x86_64.tar.gz?sv=2019-12-12&ss=bfqt&srt=o&sp=rlx&se=2020-11-20T11:15:13Z&st=2020-11-09T03:15:13Z&spr=https&sig=6wO%2Fv3PlokOq1uBP0t7aFzY%2BmY6%2BYYZ5vxereF1I18U%3D" --output toolkit-1.0.20201018.tar.gz
+mv toolkit-*.tar.gz ./toolkit.tar.gz
+tar xzf toolkit.tar.gz
+cd toolkit
+sudo make clean
+
+# Build Mariner RPM packages
+sudo make build-packages PACKAGE_BUILD_LIST="azure-iotedge libiothsm-std" CONFIG_FILE= -j$(nproc)
+
+DOCKER_VOLUME_MOUNTS=''
```
chzawist, line 45: Remove unused code beyond this

edgelet/build/linux/package-mariner.sh
```diff
@@ -0,0 +1,551 @@
...
+sed -i 's/\(azure-iotedge-[0-9.]+.tar.gz": "\)[a-fA-F0-9]+/\1${TARBALL_HASH}/g' "${PROJECT_ROOT}/SPECS/libiothsm-std/libiothsm-std.signatures.json"
+
+# Copy source tarball to expected locations
+mkdir -p "${PROJECT_ROOT}/SPECS/azure-iotedge/SOURCES/"
+cp "${PROJECT_ROOT}/../azure-iotedge-${VERSION}.tar.gz" "${PROJECT_ROOT}/SPECS/azure-iotedge/SOURCES/"
+mkdir -p "${PROJECT_ROOT}/SPECS/libiothsm-std/SOURCES/"
+cp "${PROJECT_ROOT}/../azure-iotedge-${VERSION}.tar.gz" "${PROJECT_ROOT}/SPECS/libiothsm-std/SOURCES/"
+
+# Mariner package builds may not touch the internet, so provide Cargo dependencies
+curl "https://marineriotedge.file.core.windows.net/mariner-build-env/azure-iotedge-1.0.10-cargo.tar.gz"
```
chzawist, line 32: Use cargo vendor

SPECS/azure-iotedge/0001-set-mgmt-socket-to-var-lib.patch
```diff
@@ -0,0 +1,19 @@
...
+diff --git a/iotedge/src/main.rs b/iotedge/src/main.rs
+index b82f57d6..1f2f72a9 100644
+--- a/iotedge/src/main.rs
++++ b/iotedge/src/main.rs
+@@ -67,7 +67,7 @@ fn run() -> Result<(), Error> {
+             )
+         } else {
+             (
+-                Cow::Borrowed("unix:///var/run/iotedge/mgmt.sock"),
++                Cow::Borrowed("unix:///var/lib/iotedge/mgmt.sock"),
```
chzawist, line 16: This change may not be needed. See the fix for CentOS. If a fix is needed, it can be configured via an environment variable.

arsing:
As discussed in the meeting, if we're stuck using Rust 1.39.0 for Mariner it would be useful to downgrade the branch to 1.39.0 (via `edgelet/rust-toolchain`) so that we don't have one tier 1 OS using a different compiler and set of packages. Otherwise someone could make a change that works on their machine and in the checkin-build CI, so it gets checked in, then a day layer we get the email that the Mariner packages job failed.

If `sysinfo` is the only crate that's broken, we can't downgrade it to 0.9.6 because 0.12.0 has fixes for non-amd64 arches, but we can consider forking 0.12.0 and modifying it to compile with Rust 1.39.0; we've already forked some dependencies from the tokio stack to work with Windows arm32 so this would be one more of those.

edgelet/build/linux/package-mariner.sh
```diff
@@ -0,0 +1,43 @@
...
+sed -i 's/\(azure-iotedge-[0-9.]+.tar.gz": "\)[a-fA-F0-9]+/\1${TARBALL_HASH}/g' "${PROJECT_ROOT}/SPECS/libiothsm-std/libiothsm-std.signatures.json"
+
+# Copy source tarball to expected locations
+mkdir -p "${PROJECT_ROOT}/SPECS/azure-iotedge/SOURCES/"
+cp "${PROJECT_ROOT}/../azure-iotedge-${VERSION}.tar.gz" "${PROJECT_ROOT}/SPECS/azure-iotedge/SOURCES/"
+mkdir -p "${PROJECT_ROOT}/SPECS/libiothsm-std/SOURCES/"
+cp "${PROJECT_ROOT}/../azure-iotedge-${VERSION}.tar.gz" "${PROJECT_ROOT}/SPECS/libiothsm-std/SOURCES/"
+
+# Mariner package builds may not touch the internet, so provide Cargo dependencies
+curl "https://marineriotedge.file.core.windows.net/mariner-build-env/azure-iotedge-1.0.10-cargo.tar.gz?sv=2019-12-12&ss=bfqt&srt=o&sp=rlpx&se=2020-11-13T06:50:00Z&st=2020-11-12T22:50:00Z&spr=https&sig=islDvt5euCCf%2FyhKyAsY18TWepRgPq3pLkV9OQ%2FQbfs%3D" --output azure-iotedge-1.0.10-cargo.tar.gz
```
yophilav, line 32: Should this parameter be something that's configurable upon the deployment?

edgelet/build/linux/package-mariner.sh
```diff
@@ -0,0 +1,43 @@
...
+mv azure-iotedge-1.0.10-cargo.tar.gz "${PROJECT_ROOT}/SPECS/azure-iotedge/SOURCES/"
+
+# Download Mariner toolkit
+curl "https://marineriotedge.file.core.windows.net/mariner-build-env/toolkit-1.0.20201029-x86_64.tar.gz?sv=2019-12-12&ss=bfqt&srt=o&sp=rlpx&se=2020-11-13T06:50:00Z&st=2020-11-12T22:50:00Z&spr=https&sig=islDvt5euCCf%2FyhKyAsY18TWepRgPq3pLkV9OQ%2FQbfs%3D" --output toolkit-1.0.20201018.tar.gz
```
yophilav, line 36: Should this parameter be something that's configurable upon the deployment?

chzawist:
Just force-pushed a copy of yophilav:yophilav/mariner1.0.10.3 to continue that work in progress

SPECS/azure-iotedge/0001-set-mgmt-socket-to-var-lib.patch
chzawist: It turns out the CentOS fix is somewhat different... CentOS 7 is old enough that systemd socket activation does not work quite the same as on more recent distros.

Mariner has a recent version of systemd so socket activation works fine, but this socket file happens to live under `/var/lib` rather than `/var/run`.

SPECS/azure-iotedge/0001-set-mgmt-socket-to-var-lib.patch
```diff
@@ -0,0 +1,19 @@
+This patch changes the default socket location in the iotedge management
```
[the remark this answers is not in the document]
chzawist: Fixed by @ms-mahuber

SPECS/azure-iotedge/azure-iotedge.signatures.json
```diff
@@ -0,0 +1,6 @@
+{
+ "Signatures": {
+  "azure-iotedge-1.0.10-cargo.tar.gz": "4f54b136cd5ec4782409f355163c3b3c8c924fba5397dec2060e634549b2c137",
+  "azure-iotedge-1.0.10.tar.gz": "1eb4bff9cdfcf903314a3c2432ad342c9ff2ac43ee4b96bf289f451879585bd8"
```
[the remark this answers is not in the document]
chzawist: These are sha256-sum hashes, and the pipeline does automatically replace these strings with the correct hashes. Would an empty placeholder be better?

SPECS/azure-iotedge/azure-iotedge.spec
```diff
@@ -0,0 +1,190 @@
...
+This package contains the IoT Edge daemon and CLI tool
+
+%define iotedge_user iotedge
+%define iotedge_group %{iotedge_user}
+%define iotedge_home %{_localstatedir}/lib/iotedge
+%define iotedge_logdir %{_localstatedir}/log/iotedge
+%define iotedge_confdir %{_sysconfdir}/iotedge
+%define iotedge_datadir %{_datadir}/iotedge
+
+%global debug_package %{nil}
```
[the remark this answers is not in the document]
chzawist: Setting this global to %{nil} disables generation of debug packages. When debug packages _are_ generated, they contain the equivalent of .pdbs

SPECS/azure-iotedge/azure-iotedge.spec
```diff
@@ -0,0 +1,190 @@
...
+BuildRequires:  rust >= 1.39.0
+BuildRequires:  cmake
+BuildRequires:  curl
+BuildRequires:  git
+BuildRequires:  shadow-utils
+BuildRequires:  systemd
+Requires:       libiothsm-std
+Requires:       openssl
+Requires:       moby-engine
+Requires:       moby-cli
```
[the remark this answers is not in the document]
chzawist: It is possible to specify version constraints, yes. See the Rust example a few lines above.

SPECS/azure-iotedge/azure-iotedge.spec
```diff
@@ -0,0 +1,190 @@
...
+# Create iotedge user
+if ! /usr/bin/getent passwd iotedge >/dev/null; then
+    %{_sbindir}/useradd -r -g %{iotedge_group} -c "iotedge user" -s /bin/nologin -d %{iotedge_home} %{iotedge_user}
+fi
+
+if /usr/bin/getent group docker >/dev/null; then
+    %{_sbindir}/usermod -a -G docker %{iotedge_user}
+fi
+exit 0
+
```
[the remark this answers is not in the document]
chzawist: No, I believe that's a standard RPM specfile convention.

>exit 0 signifies the end of the %prep section, and therefore, the end of the %setup macro

builds/mariner/SPECS/libiothsm-std/libiothsm-std.spec
```diff
@@ -0,0 +1,61 @@
...
+
+%clean
+rm -rf $RPM_BUILD_ROOT
+
+%files
+%defattr(-,root,root)
+%license LICENSE
+%{_libdir}/libiothsm.so*
+
+%changelog
```
chzawist, line 53: Need to update this changelog as well

edgelet/build/linux/package-mariner.sh
```diff
@@ -0,0 +1,94 @@
...
+BUILD_REPOSITORY_LOCALPATH="$(realpath "${BUILD_REPOSITORY_LOCALPATH:-$DIR/../../..}")"
+EDGELET_ROOT="${BUILD_REPOSITORY_LOCALPATH}/edgelet"
+MARINER_BUILD_ROOT="${BUILD_REPOSITORY_LOCALPATH}/builds/mariner"
+VERSION="$(cat "$EDGELET_ROOT/version.txt")"
+
+pushd "${EDGELET_ROOT}"
+
+# Pull Cargo deps and extract
+echo "Vendoring Rust dependencies"
+curl -o "azure-iotedge-1.0.10.3-cargo-vendor.zip" "https://marineriotedge.file.core.windows.net/mariner-build-env/azure-iotedge-1.0.10.3-cargo-vendor.zip?sv=2019-12-12&ss=bf&srt=o&sp=rl&se=2021-02-01T09:51:53Z&st=2021-01-04T01:51:53Z&spr=https&sig=yKbgAIjgol1nmv%2B3bFP%2BegX43i2bfmc82vhR%2F%2Bs7naw%3D"
```
chzawist, line 17: Avoid use of SAS tokens

edgelet/build/linux/package-mariner.sh
```diff
@@ -0,0 +1,94 @@
...
+popd # EDGELET_ROOT
+
+# Create source tarball, including cargo dependencies and license
+pushd "${BUILD_REPOSITORY_LOCALPATH}"
+tar -czf azure-iotedge-${VERSION}.tar.gz --transform="s,^.*edgelet/,azure-iotedge-${VERSION}/edgelet/," "${EDGELET_ROOT}"
+popd
+
+# Update expected tarball hash
+TARBALL_HASH=$(sha256sum "${BUILD_REPOSITORY_LOCALPATH}/azure-iotedge-${VERSION}.tar.gz" | awk '{print $1}')
+# TODO: inject version into signatures.json and into the SPEC file definition in order to avoid hard-coding
```
chzawist, line 59: Remove TODO comment; we are already using the correct approach

edgelet/build/linux/package-mariner.sh
```diff
@@ -0,0 +1,94 @@
...
+# Prepare toolkit
+pushd ${MARINER_BUILD_ROOT}
+sudo tar xzf toolkit.tar.gz
+pushd toolkit
+sudo make clean
+
+# Download Rust 1.45 RPM instead of using official 1.39 which is too old
+pushd "${BUILD_REPOSITORY_LOCALPATH}"
+mkdir -p "builds/mariner/out/RPMS/x86_64/"
+curl -o "builds/mariner/out/RPMS/x86_64/rust-1.45.2-1.cm1.x86_64.rpm" "https://marineriotedge.file.core.windows.net/mariner-build-env/rust-1.45.2-1.cm1.x86_64.rpm?sv=2019-12-12&ss=bf&srt=o&sp=rl&se=2021-02-01T09:51:53Z&st=2021-01-04T01:51:53Z&spr=https&sig=yKbgAIjgol1nmv%2B3bFP%2BegX43i2bfmc82vhR%2F%2Bs7naw%3D"
```
chzawist, line 88: Avoid use of SAS tokens

builds/mariner/SPECS/azure-iotedge/azure-iotedge.signatures.json
```diff
@@ -0,0 +1,5 @@
+{
+ "Signatures": {
+  "azure-iotedge-1.0.10.3.tar.gz": "0000000000000000000000000000000000000000000000000000000000000000"
```
chzawist, line 3: Fix version numbers throughout the PR

builds/misc/packages.yaml
```diff
@@ -146,6 +146,13 @@ jobs:
...
         displayName: Set Version
+      - task: AzureKeyVault@1
+        displayName: Get secrets
+        inputs:
+          azureSubscription: $(az.subscription)
+          keyVaultName: $(kv.name)
+          secretsFilter: >-
+            marineriotedge-key
       - script: edgelet/build/linux/package-mariner.sh
         displayName: Create libiothsm and iotedged packages
```
yophilav, line 157:
```
- script: edgelet/build/linux/package-mariner.sh
  displayName: Create libiothsm and iotedged packages
  env:
    MARINERIOTEDGE_KEY: $(marineriotedge-key)
```
then in the `edgelet/build/linux/package-mariner.sh`, you can access the value using `$MARINERIOTEDGE_KEY`

chzawist:
The PR should now be complete and in mergeable condition.

Two main points of interest:
- The `sysinfo` crate is no longer downgraded since the Mariner RPM is built with Rust 1.45.
- Vendored cargo dependencies and the Rust 1.45 RPM are downloaded from Azure Blob Storage using a shared key.

chzawist:
[Pipeline results](https://msazure.visualstudio.com/One/_build/results?buildId=38174535&view=results) are green across the board.

edgelet/build/linux/package-mariner.sh
```diff
@@ -0,0 +1,90 @@
...
+
+# Create source tarball, including cargo dependencies and license
+pushd "${BUILD_REPOSITORY_LOCALPATH}"
+tar -czf azure-iotedge-${VERSION}.tar.gz --transform="s,^.*edgelet/,azure-iotedge-${VERSION}/edgelet/," "${EDGELET_ROOT}"
+popd
+
+# Update expected tarball hash
+TARBALL_HASH=$(sha256sum "${BUILD_REPOSITORY_LOCALPATH}/azure-iotedge-${VERSION}.tar.gz" | awk '{print $1}')
+sed -i 's/\("azure-iotedge-[0-9.]\+.tar.gz": "\)\([a-fA-F0-9]\+\)/\1'${TARBALL_HASH}'/g' "${MARINER_BUILD_ROOT}/SPECS/azure-iotedge/azure-iotedge.signatures.json"
+sed -i 's/\("azure-iotedge-[0-9.]\+.tar.gz": "\)\([a-fA-F0-9]\+\)/\1'${TARBALL_HASH}'/g' "${MARINER_BUILD_ROOT}/SPECS/libiothsm-std/libiothsm-std.signatures.json"
```
yophilav, line 59: Please print out the `TARBALL_HASH` to the console

edgelet/build/linux/package-mariner.sh
```diff
@@ -0,0 +1,90 @@
...
+pushd toolkit
+
+# Build Mariner RPM packages
+sudo make build-packages PACKAGE_BUILD_LIST="azure-iotedge libiothsm-std" CONFIG_FILE= -j$(nproc) LOG_LEVEL=trace
+popd
+popd
+
+# Prevent publishing the Rust RPM as an artifact
+pushd "${BUILD_REPOSITORY_LOCALPATH}"
+rm "builds/mariner/out/RPMS/x86_64/rust-1.45.2-1.cm1.x86_64.rpm"
```
yophilav, line 89: Should this be a regex and not a hard-coded value?

builds/mariner/SPECS/azure-iotedge/0001-set-mgmt-socket-to-var-lib.patch
```diff
@@ -0,0 +1,19 @@
+This patch changes the default socket location in the iotedge management
+application to be in /var/lib/, the canonical location for non-debian Linux.
+
+This saves the user from having to specify --host unix:\\<path-to>.socket every
+time they use the iotedge management app.
+
+diff --git a/iotedge/src/main.rs b/iotedge/src/main.rs
+index b82f57d6..1f2f72a9 100644
+--- a/iotedge/src/main.rs
++++ b/iotedge/src/main.rs
```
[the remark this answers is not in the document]
chzawist: Yes, GitHub automatically colors the patch, but really we are adding the entire patch file. This patch changes `/var/run` to `/var/lib` for compatibility with Mariner's systemd socket location.

builds/misc/packages.yaml
```diff
@@ -125,6 +125,70 @@ jobs:
...
+      - task: AzureCLI@2
+        displayName: Download from Azure Blob Storage
+        inputs:
+          azureSubscription: $(az.subscription)
+          scriptType: bash
+          scriptLocation: inlineScript
+          inlineScript: |
+            az --version
+            mkdir -p "builds/mariner/out/RPMS/x86_64/"
+            az storage blob download --account-name "marineriotedge" --container-name "iotedge-build-pipeline" --file "builds/mariner/out/RPMS/x86_64/rust-1.45.2-1.cm1.x86_64.rpm" --name "rust-1.45.2-1.cm1.x86_64.rpm" --account-key "$(marineriotedge-key)"
```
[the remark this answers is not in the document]
chzawist: I actually built them in this pipeline: ([commit](https://github.com/Azure/iotedge/pull/3949/commits/02a12cea77b94a66feb16576a59748f2a1b1585a), [pipeline result](https://msazure.visualstudio.com/One/_build/results?buildId=38132739&view=artifacts&pathAsName=false&type=publishedArtifacts))

Then I downloaded that artifact and uploaded it to blob storage. The original Rust 1.45 specfile and source tarball came from the [wireless5g project](https://dev.azure.com/mariner-org/mariner/_git/wireless5g).

edgelet/build/linux/package-mariner.sh
chzawist: Good idea. Then it will work even if we update Rust in the future.

builds/misc/packages.yaml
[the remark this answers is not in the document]
yophilav: Nice! Let say... there is a rust version update, how can I build, publish & update the rust in blob storage? 
Do you think it is a reasonable thing to do to have rust be build and publish as an artifact as well? 
(I know this is contradict to what I said earlier, sorry).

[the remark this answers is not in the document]
chzawist: I've reverted this to build rust 1.45 from source as long as the official repos do not satisfy the `rust >= 1.45` constraint as declared in the iotedge spec file.

edgelet/build/linux/package-mariner.sh
chzawist: As discussed, we're actually going to publish the rust rpm as an artifact as long as it is built from source.

builds/mariner/SPECS/azure-iotedge/azure-iotedge.spec
```diff
@@ -0,0 +1,186 @@
...
+
+#Source0:       https://github.com/Azure/iotedge/archive/%{version}.tar.gz
+Source0:        %{name}-%{version}.tar.gz
+License:        MIT
+Group:          Applications/File
+URL:            https://github.com/azure/iotedge
+Vendor:         Microsoft Corporation
+Distribution:   Mariner
+
+BuildRequires:  rust >= 1.40.0
```
arsing, line 23: This should be an `=` relation on the exact version listed in `edgelet/rust-toolchain`, otherwise the Mariner packages will be built with a different toolchain than every other package, which is the problem we've already talked about.

builds/mariner/SPECS/rust/rust.spec
```diff
@@ -0,0 +1,60 @@
+Summary:        Rust Programming Language
+Name:           rust
+Version:        1.45.2
```
arsing, line 3: Same here. The version of Rust that is built must be the exact version listed in `edgelet/rust-toolchain`.

builds/mariner/SPECS/azure-iotedge/0001-set-mgmt-socket-to-var-lib.patch
```diff
@@ -0,0 +1,19 @@
...
+@@ -67,7 +67,7 @@ fn run() -> Result<(), Error> {
+             )
+         } else {
+             (
+-                Cow::Borrowed("unix:///var/run/iotedge/mgmt.sock"),
```
arsing, line 15: The correct way to do this is the way it's already done for Debian 8 and CentOS 7, with a compile-time env var.

https://github.com/Azure/iotedge/blob/release/1.0.10/edgelet/Makefile#L120-L123

arsing (changes requested):

builds/mariner/SPECS/azure-iotedge/azure-iotedge.spec
```diff
@@ -20,7 +20,7 @@ URL:            https://github.com/azure/iotedge
 Vendor:         Microsoft Corporation
 Distribution:   Mariner
 
-BuildRequires:  rust >= 1.40.0
+BuildRequires:  rust == 1.41.1
```
arsing, line 23: `s/41/44/`

chzawist:
Comments addressed, [build pass](https://msazure.visualstudio.com/One/_build/results?buildId=38213174&view=results) is green again.

arsing (approved):
Changes look fine to me. I assume someone is testing the packages on Mariner to ensure they work, etc.

arsing:
@yophilav I think you said we have some doc for how to make releases and such, right? If so, we should to update it because this PR adds new versions numbers that need to be kept in sync - the iotedge package's version is hard-coded in the `*.spec` and `*.signatures.json` files, and the content of the rust-toolchain file needs to match the `*.spec` files. And of course we need to update the Rust source tarballs too.

yophilav:
> @yophilav I think you said we have some doc for how to make releases and such, right? If so, we should to update it because this PR adds new versions numbers that need to be kept in sync - the iotedge package's version is hard-coded in the `*.spec` and `*.signatures.json` files, and the content of the rust-toolchain file needs to match the `*.spec` files. And of course we need to update the Rust source tarballs too.
Thank you Arnav! I'll address that in our release document for now. I think we can patch this to be part of the pipeline step in a near future.

chzawist:
I have documented the build pipeline [here](https://microsoft.sharepoint-df.com/teams/AzureEFLOWRD/Wiki/Onboarding%20EFLOW%20Mariner%20for%20Iotedge.aspx).

It contains a short description of how the Mariner build works, as well as steps for updating Rust, Rust dependencies, and azure-iotedge/libiothsm-std version numbers.

Issue #3948, opened 2020-11-11
'iotedge check' fails on Nvidia Jetson NX when checking daemon management endpoint


Chad-McQuillen:
## Expected Behavior
'iotedge check' should not fail when checking for the daemon management endpoint.

## Current Behavior
```
× config.yaml has correct URIs for daemon mgmt endpoint - Error
    One or more errors occurred. (Cannot assign requested address var/run/iotedge/mgmt.sock)
        caused by: docker returned exit code: 1, stderr = One or more errors occurred. (Cannot assign requested address var/run/iotedge/mgmt.sock)
```

## Steps to Reproduce
Run `iotedge check` on an Nvidia Jetson NX Devkit.

### Output of `iotedge check`

<details>
<summary>Click here</summary>

```
√ config.yaml is well-formed - OK
√ config.yaml has well-formed connection string - OK
√ container engine is installed and functional - OK
√ config.yaml has correct hostname - OK
× config.yaml has correct URIs for daemon mgmt endpoint - Error
    One or more errors occurred. (Cannot assign requested address var/run/iotedge/mgmt.sock)
        caused by: docker returned exit code: 1, stderr = One or more errors occurred. (Cannot assign requested address var/run/iotedge/mgmt.sock)
√ latest security daemon - OK
√ host time is close to real time - OK
√ container time is close to host time - OK
‼ DNS server - Warning
    Container engine is not configured with DNS server setting, which may impact connectivity to IoT Hub.
    Please see https://aka.ms/iotedge-prod-checklist-dns for best practices.
    You can ignore this warning if you are setting DNS server per module in the Edge deployment.
‼ production readiness: certificates - Warning
    The Edge device is using self-signed automatically-generated development certificates.
    They will expire in 89 days (at 2021-02-08 18:43:09 UTC) causing module-to-module and downstream device communication to fail on an active deployment.
    After the certs have expired, restarting the IoT Edge daemon will trigger it to generate new development certs.
    Please consider using production certificates instead. See https://aka.ms/iotedge-prod-checklist-certs for best practices.
‼ production readiness: container engine - Warning
    Device is not using a production-supported container engine (moby-engine).
    Please see https://aka.ms/iotedge-prod-checklist-moby for details.
‼ production readiness: logs policy - Warning
    Container engine is not configured to rotate module logs which may cause it run out of disk space.
    Please see https://aka.ms/iotedge-prod-checklist-logs for best practices.
[...]

PR #3950, opened 2020-11-11
Fix Linux Check

lfitchett:
When checking the workload api from a container, Windows needs to remove a leading /. Linux doesn't, and removing the / caused it to incorrectly fail. 

To fix this, the Uri class is used to correctly localize the path.


damonbarry:
Fixes #3948.

damonbarry (approved):

PR #3955, opened 2020-11-11
Prepare for release 1.0.10.2

damonbarry:


lfitchett (dismissed):

varunpuranik (changes requested):
We also need to update this file - https://github.com/Azure/iotedge/blob/release/1.0.10/versionInfo.json

damonbarry:
> We also need to update this file - https://github.com/Azure/iotedge/blob/release/1.0.10/versionInfo.json

@varunpuranik That file is used by Edge agent and Edge hub to report the version in their logs. Since we're not bumping agent and hub, I don't think we should update that file.

varunpuranik (approved):

varunpuranik (approved):

lfitchett (approved):

arsing (approved):

PR #3960, opened 2020-11-12
Improving M2M feedback handling

vipeller:
- m2m pending confirmations are stored by lockToken not by identity - allowing multiple pending messages by identity
- fixed race condition when confirmation message gets back sooner than lockToken gets stored

m2m pending messages were stored by identity - it worked most of the time because only one message is getting sent to a client. However when that message timeouts a new message will be sent, overwriting the previous pending message. This caused problems when the old message with the timeout got confirmed, because edgeHub believed that it was the recent message. With this change every messageId (lockToken) is saved, so feedbacks cannot be mismatched.

Also there was a race condition when the lock token of the sent out message was stored after the 'send' call. It could happened that the confirmation message get back earlier than the lock token got stored, and because of that the confirmation code could not find the related lock token. As a result the confirmed message was never been confirmed, causing resent messages.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ModuleToModuleResponseTimeout.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
+{
+    using System;
+
+    // The role of this class is to help injecting a timeout value without activators
+    public class ModuleToModuleResponseTimeout
```
vadim-kovalyov, line 7: nit: since TimeSpan is a struct, you can make it zero-cost abstraction ;)
```suggestion
    public struct ModuleToModuleResponseTimeout
```

vadim-kovalyov (changes requested):
This change will break authZ for m2m messages. Let's discuss offline.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/ModuleToModuleResponseTimeout.cs
vipeller: It is registered by autofac as an instance and it has problems with structures

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ModuleToModuleMessageHandler.cs
```diff
@@ -85,13 +93,25 @@ public async Task SendModuleToModuleMessageAsync(IMessage message, string input,
             }
 
             bool result;
-            var topic = string.Empty;
             try
             {
+                var currentTime = DateTime.Now;
```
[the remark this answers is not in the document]
vipeller: yes, around daylight savings it could have caused problems (showing the token age e.g. an hour older and remove it)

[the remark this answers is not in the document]
vadim-kovalyov: nit: I believe there is `DateTime.UtcNow`

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/ModuleToModuleMessageHandler.cs
```diff
@@ -85,13 +93,25 @@ public async Task SendModuleToModuleMessageAsync(IMessage message, string input,
             }
 
             bool result;
-            var topic = string.Empty;
             try
             {
+                var currentTime = DateTime.Now.ToUniversalTime();
```
vadim-kovalyov, line 98: nit: I believe there is `DateTime.UtcNow`

vadim-kovalyov (dismissed):

vadim-kovalyov (approved):

vadim-kovalyov (approved):

PR #3961, opened 2020-11-12
Enable topic translation tests

vadim-kovalyov:
There was a bug in feature declaration and translation integration tests did not compile and run for a while. 

The fix is:
- Add optional `proptest` dependency to `mqtt-edgehub` package. That will enable `proptest` feature.
- Import  `mqtt-broker` with `features = ["proptest"]` as a dev-dependency in `mqtt-edgehub`.

The rest of the changes is just to make tests compile after all the recent changes.

### DIFF
--- a/mqtt/mqtt-edgehub/tests/translation.rs
+++ b/mqtt/mqtt-edgehub/tests/translation.rs
@@ -1,32 +1,26 @@
-use std::{
-    error::Error as StdError,
-    sync::atomic::{AtomicU32, Ordering},
-};
-
-use lazy_static::lazy_static;
 use matches::assert_matches;
+use mqtt_edgehub::connection::MakeEdgeHubPacketProcessor;
 use proptest::prelude::*;
-use tokio::{runtime::Runtime, sync::oneshot};
+use tokio::runtime::Runtime;
 
-use futures_util::FutureExt;
 use mqtt3::{
     proto::{ClientId, QoS},
     ReceivedPublication,
 };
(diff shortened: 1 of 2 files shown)


mqtt/mqtt-edgehub/tests/translation.rs
```diff
@@ -1,32 +1,26 @@
-use std::{
-    error::Error as StdError,
-    sync::atomic::{AtomicU32, Ordering},
-};
-
-use lazy_static::lazy_static;
 use matches::assert_matches;
+use mqtt_edgehub::connection::MakeEdgeHubPacketProcessor;
```
dmolokanov, line 2: nit: move this down?

mqtt/mqtt-edgehub/tests/translation.rs
```diff
@@ -331,28 +325,21 @@ async fn ensure_subscribed(client: &mut TestClient) {
     client.subscriptions().recv().await;
 }
 
-fn start_server<N, Z>(broker: Broker<Z>, authenticator: N) -> ServerHandle
```
dmolokanov, line 334: Can we just re-use `server::start_server` ?

dmolokanov (approved):

mqtt/mqtt-edgehub/tests/translation.rs
vadim-kovalyov: we can't, it does not add message handler.

PR #3972, opened 2020-11-16
Added Publish Test Results task to MQTT Pipelines + other pipeline improvements

vadim-kovalyov:
Rust tests don't have native test results support in Azure Pipelines. In this PR I want to add [a 3rd party tool](https://github.com/johnterickson/cargo2junit) that converts cargo test output into junit format (Azure Pipelines compatible). As a result, we can enjoy test stats, can see flaky tests, trends.

image

Drawbacks:
- We have to use unstable cargo feature to generate json output. I believe this should not be a problem, since it does not affect production bits, but only tests.
- Console output in the pipeline becomes less readable, but not that bad, and still searchable. Can use "Tests" tab instead!

Another minor improvements in this PR:
- Don't checkout submodules for MQTT Broker. It doesn't use any.
- Shallow clone (speeds up checkout x4)
- Removed Style Check task from MQTT CI. I can argue that it was redundant.

### DIFF
--- a/mqtt/build/linux/test.sh
+++ b/mqtt/build/linux/test.sh
@@ -28,6 +28,7 @@ usage()
     echo ""
     echo "options"
     echo " -h, --help          Print this help and exit."
+    echo " -t, --target        Target architecture."
     echo " -r, --release       Release build? (flag, default: false)"
     echo " --report            Optional. Generates the xml test report with specified name."
     exit 1;
@@ -57,6 +58,7 @@ process_args()
                 "-r" | "--release" ) RELEASE="--release";;
                 "-c" | "--cargo" ) save_next_arg=2;;
                 "--report" ) save_next_arg=3;;
+                * ) usage;;
             esac
         fi
     done
--- a/mqtt/build/linux/test.sh
+++ b/mqtt/build/linux/test.sh
@@ -30,6 +30,7 @@ usage()
     echo " -h, --help          Print this help and exit."
     echo " -t, --target        Target architecture."
     echo " -r, --release       Release build? (flag, default: false)"
+    echo " -c, --cargo         Path of cargo installation."
     echo " --report            Optional. Generates the xml test report with specified name."
     exit 1;
 }
--- a/mqtt/build/linux/test.sh
+++ b/mqtt/build/linux/test.sh
@@ -29,6 +29,7 @@ usage()
     echo "options"
     echo " -h, --help          Print this help and exit."
     echo " -r, --release       Release build? (flag, default: false)"
+    echo " --report            Optional. Generates the xml test report with specified name."
     exit 1;
 }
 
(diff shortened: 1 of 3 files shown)


mqtt/build/linux/test.sh
```diff
@@ -63,8 +63,14 @@ process_args()
...
 
+# Get cargo2junit to report test results to Azure Pipelines
+$CARGO install cargo2junit
+
 if [[ -z ${RELEASE} ]]; then
-    cd "$PROJECT_ROOT" && $CARGO test --workspace --all-features --target "$TARGET"
+    cd "$PROJECT_ROOT" && $CARGO test --no-fail-fast --workspace --all-features --target "$TARGET" -- -Z unstable-options --format json | tee test-result.json
 else
-    cd "$PROJECT_ROOT" && $CARGO test --workspace --all-features --release --target "$TARGET"
+    cd "$PROJECT_ROOT" && $CARGO test --no-fail-fast --workspace --all-features --release --target "$TARGET" -- -Z unstable-options --format json | tee test-result.json
```
and-rewsmith, line 72: nit: If we can reduce number of columns it will be more readable. Maybe a line split with \ character.

and-rewsmith (dismissed):
Nice job!

mqtt/build/linux/test.sh
```diff
@@ -63,8 +63,14 @@ process_args()
 
 process_args "$@"
 
+# Get cargo2junit to report test results to Azure Pipelines
+$CARGO install cargo2junit
```
dmolokanov, line 67: lets move cargo2junit out of the test.sh script? Or at least cover this with command-line argument. For instance, someone may use this script to run all tests before he pushes the code, but he doesn't want prepare any reports locally

mqtt/build/linux/test.sh
```diff
@@ -29,6 +29,7 @@ usage()
     echo "options"
     echo " -h, --help          Print this help and exit."
     echo " -r, --release       Release build? (flag, default: false)"
```
and-rewsmith, line 31: This could be a good opportunity to fix the usage missing the --target option.

mqtt/build/linux/test.sh
```diff
@@ -44,33 +45,37 @@ process_args()
...
-            CARGO="$arg"
+            REPORT="$arg"
             save_next_arg=0
         else
             case "$arg" in
                 "-h" | "--help" ) usage;;
                 "-t" | "--target" ) save_next_arg=1;;
-                "-r" | "--release" ) save_next_arg=2;;
-                "-c" | "--cargo" ) save_next_arg=3;;
-                * ) usage;;
```
and-rewsmith, line 58: Why remove this call to usage?

vadim-kovalyov: good catch, it was a mistake

mqtt/build/linux/test.sh
and-rewsmith: Just noticed it is also missing --cargo

and-rewsmith (dismissed):

dmolokanov (approved):
Looks great! Thank! Code coverage next?

and-rewsmith (approved):

dmolokanov (approved):
:shipit:

PR #3978, opened 2020-11-17
Update ARM base images to 3.1.10-bionic

yophilav:
Update ARM base images to 3.1.10-bionic


and-rewsmith (approved):
I checked out the repo and nothing was using 3.1.9-bionic or using 1.0.6.3 base image tag anymore, so I think we are good to go.

PR #3980, opened 2020-11-17
[Fix 8419918]: fixed message counts metrics

davilu:
Modified to update message counts metrics after clean up


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -261,28 +261,38 @@ private async Task CleanQueue(bool checkEntireQueueOnCleanup)
...
-                            CheckpointData checkpointData = await this.messageStore.checkpointStore.GetCheckpointDataAsync(endpointSequentialStore.Key, CancellationToken.None);
+                            var (endpointId, priority) = MessageQueueIdHelper.ParseMessageQueueId(messageQueueId);
+                            Events.CleanupTaskStarted(messageQueueId);
+                            CheckpointData checkpointData = await this.messageStore.checkpointStore.GetCheckpointDataAsync(messageQueueId, CancellationToken.None);
                             ISequentialStore<MessageRef> sequentialStore = endpointSequentialStore.Value;
-                            Events.CleanupCheckpointState(endpointSequentialStore.Key, checkpointData);
+                            Events.CleanupCheckpointState(messageQueueId, checkpointData);
                             int cleanupEntityStoreCount = 0;
 
+                            // If cleanup just peek head, message counts is tailOffset-headOffset+1
```
dylanbronson, line 279: It would be clearer for me if the comment was
"If checkEntireQueueOnCleanup is set to false, we only peek the head, so message count is tailOffset-headOffset+1"

edge-hub/core/src/Microsoft.Azure.Devices.Routing.Core/MessageQueueIdHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+            if (idx < 0)
+            {
+                return (messageQueueId, RouteFactory.DefaultPriority);
+            }
+
+            var endpointId = messageQueueId.Substring(0, idx);
+            var priority = messageQueueId.Substring(idx + MessageQueueIdDelimiter.Length);
+            try
+            {
+                return (endpointId, Convert.ToUInt32(priority));
```
vadim-kovalyov, line 28: nit: why not UInt32.TryParse? exceptions are slow and really for exceptional situations...

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -261,28 +261,38 @@ private async Task CleanQueue(bool checkEntireQueueOnCleanup)
...
+                            // otherwise count while iterating over the queue.
+                            var headOffset = 0L;
+                            var tailOffset = sequentialStore.GetTailOffset(CancellationToken.None);
+                            var messageCounts = 0L;
```
dylanbronson, line 283: nit: messageCount ?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -261,28 +261,38 @@ private async Task CleanQueue(bool checkEntireQueueOnCleanup)
...
+                            var (endpointId, priority) = MessageQueueIdHelper.ParseMessageQueueId(messageQueueId);
+                            Events.CleanupTaskStarted(messageQueueId);
+                            CheckpointData checkpointData = await this.messageStore.checkpointStore.GetCheckpointDataAsync(messageQueueId, CancellationToken.None);
                             ISequentialStore<MessageRef> sequentialStore = endpointSequentialStore.Value;
-                            Events.CleanupCheckpointState(endpointSequentialStore.Key, checkpointData);
+                            Events.CleanupCheckpointState(messageQueueId, checkpointData);
                             int cleanupEntityStoreCount = 0;
 
+                            // If cleanup just peek head, message counts is tailOffset-headOffset+1
+                            // otherwise count while iterating over the queue.
```
dylanbronson, line 280: nit: "while iterating over the queue(s)" since there can be multiple

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/DeviceMessageHandlerTest.cs
```diff
@@ -389,6 +390,8 @@ public async Task X509DeviceCanSendMessageTest()
             // send message to x509 device
             Task sendMessageTask = deviceMessageHandler.SendMessageAsync(message, "input1");
             await deviceMessageHandler.ProcessMessageFeedbackAsync(lockToken, FeedbackStatus.Complete);
+
+            await Task.Delay(TimeSpan.FromSeconds(1));
```
vadim-kovalyov, line 394: nit: why not just await sendMessageTask with timeout?

edge-hub/core/test/Microsoft.Azure.Devices.Routing.Core.Test/MessageQueueIdHelperTest.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        [Fact]
+        [Unit]
+        public void TestGetMessageQueueIdWithNonDefaultPriority()
+        {
+            // Message queue with non-default priority should combain endpoint id and priority with MessageQueueIdHelper.MessageQueueIdDelimiter
+            var endpointId = Guid.NewGuid().ToString();
+            uint priority = 1234;
+            var expectMessageQueueId = $"{endpointId}{MessageQueueIdHelper.MessageQueueIdDelimiter}{priority}";
```
vadim-kovalyov, line 27:
```suggestion
            var expectMessageQueueId = $"test_Pri1234";
```
I encourage to use actual strings instead of referencing consts from prod code and string format. https://youtu.be/dJUVNFxrK_4?t=478

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/DeviceMessageHandlerTest.cs
```diff
@@ -323,6 +323,7 @@ public async Task MessageCompletionMismatchedResponseTest()
 
             await deviceMessageHandler.ProcessMessageFeedbackAsync(Guid.NewGuid().ToString(), FeedbackStatus.Complete);
 
+            await Task.Delay(TimeSpan.FromSeconds(1));
```
dylanbronson, line 326: Does this test fail without this? Or is it flaky without this?

edge-hub/core/test/Microsoft.Azure.Devices.Routing.Core.Test/MessageQueueIdHelperTest.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace Microsoft.Azure.Devices.Routing.Core.Test
+{
+    using System;
+    using Microsoft.Azure.Devices.Edge.Util.Test.Common;
+    using Xunit;
+
+    public class MessageQueueIdHelperTest
+    {
+        [Fact]
+        [Unit]
```
dylanbronson, line 11: nit: You can add [Unit] attribute at the top of the class to have it apply to the whole class

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -261,28 +261,38 @@ private async Task CleanQueue(bool checkEntireQueueOnCleanup)
...
+                            var tailOffset = sequentialStore.GetTailOffset(CancellationToken.None);
+                            var messageCounts = 0L;
+
                             async Task<bool> DeleteMessageCallback(long offset, MessageRef messageRef)
                             {
-                                var enqueuedTime = DateTime.UtcNow - messageRef.TimeStamp;
-                                if (checkpointData.Offset < offset &&
-                                     enqueuedTime < messageRef.TimeToLive)
+                                var expiry = messageRef.TimeStamp + messageRef.TimeToLive;
+                                if (offset > checkpointData.Offset && expiry > DateTime.UtcNow)
```
dylanbronson, line 288: When would offset be less than checkpointData.Offset?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/DeviceMessageHandlerTest.cs
davilu: If there is a bug in code `sendMessageTask` might never return

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.Core.Test/DeviceMessageHandlerTest.cs
davilu: The test is flaky without it.

edge-hub/core/test/Microsoft.Azure.Devices.Routing.Core.Test/MessageQueueIdHelperTest.cs
davilu: It sounds reasonable but it's the same. The fact of where actual strings come from is from human calculation which based on coder's understanding of the logic, which is `$"{endpointId}{MessageQueueIdHelper.MessageQueueIdDelimiter}{priority}"` here. I won't say it;s the same code since I'm not asserting `func(a,b)` equals `func(a,b)`. If someone modified original function, the test will catch it. I couldn't agree his opinion UT is for catching bugs in code. IMPO, UT is used for avoid regression. And since `MessageQueueIdDelimiter` is defined as `const`, it could be changed as well. I see benefit hardcode in UT but for long term, it could be issue for maintains.

edge-hub/core/src/Microsoft.Azure.Devices.Routing.Core/MessageQueueIdHelper.cs
davilu: done

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
davilu: done

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
davilu: `CheckPointer` will update  `checkpointData.Offset` which indicates message get sent. But `offset` here is message ref from database which won't be update unless deleted. They are completely separate.

edge-hub/core/test/Microsoft.Azure.Devices.Routing.Core.Test/MessageQueueIdHelperTest.cs
davilu: done.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
richma-ms: we can probably just remove the class member to reduce naming confusion

---
In reply to: [525486168](https://github.com/Azure/iotedge/pull/3980#discussion_r525486168) [](ancestors = 525486168)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -305,10 +315,7 @@ async Task<bool> DeleteMessageCallback(long offset, MessageRef messageRef)
 
                                 if (deleteMessage)
                                 {
-                                    if (checkpointData.Offset < offset && enqueuedTime >= messageRef.TimeToLive)
```
richma-ms, line 308: >if (checkpointData.Offset < offset && enqueuedTime >= messageRef.TimeToLive) [](start = 36, length = 76)

Seems like this is needed?  We only want to increment the counter here for messages that were not sent but have expired.

davilu: Good catch. Fixed.

richma-ms (approved):

PR #3984, opened 2020-11-18
Disable e2e tests

dylanbronson:
Disabling flaky e2e tests to unblock the pipeline


test/Microsoft.Azure.Devices.Edge.Test/Module.cs
```diff
@@ -19,6 +19,7 @@ public class Module : SasManualProvisioningFixture
 
         [Test]
         [Category("CentOsSafe")]
+        [Ignore("Temporarily disabling flaky test while we figure out what is wrong")]
```
yophilav, line 22: Please use ` Assert.Ignore()` instead

dylanbronson: Sure, but why?

yophilav (approved):

PR #3985, opened 2020-11-18
Policy engine proptests

vadim-kovalyov:
In this PR I've added one proptest scenario for Policy engine.

To make that work I had to introduce a method `PolicyBuilder::from_definition(definition)`, that can accept struct, instead of json string.

Proptests guarder by the "proptest" feature.


mqtt/policy/src/core/mod.rs
```diff
@@ -843,4 +833,98 @@ pub(crate) mod tests {
             policy.starts_with(input)
         }
     }
+
+    #[cfg(feature = "proptest")]
+    mod proptests {
```
dmolokanov, line 838: Do you want to split proptest generators and test themselves and move generators to a separate file?

mqtt/policy/src/core/mod.rs
```diff
@@ -304,18 +304,8 @@ pub enum Decision {
...
-            Effect::Allow => Decision::Allowed,
-            Effect::Deny => Decision::Denied,
-            Effect::Undefined => Decision::Denied,
-        }
-    }
-}
-
 #[derive(Debug, Copy, Clone, PartialOrd, PartialEq)]
-enum Effect {
+enum EffectImpl {
```
dmolokanov, line 308: Not sure I understood this change. You have Effect, EffectImpl, and Decision. What are the differences?

mqtt/policy/src/core/mod.rs
```diff
@@ -843,4 +833,98 @@ pub(crate) mod tests {
...
+            /// - All combinations of identity/operation/resource in a statement in the definition
+            ///   should produce expected result.
+            ///   Since some statements can be overridden by the previous ones,
+            ///   we can only safely verify the very first statement.
+            #[test]
+            fn policy_engine_proptest(definition in arb_policy_definition()){
+                use itertools::Itertools;
+
+                // take very first statement, which should have top priority.
+                let statement = &definition.statements()[0];
```
dmolokanov, line 854: I guess it is possible to receive an empty definition from the customer. Do we want to cover that here?

mqtt/policy/src/core/mod.rs
```diff
@@ -843,4 +833,98 @@ pub(crate) mod tests {
...
+
+                // collect all combos of identity/operation/resource
+                // in the statement.
+                let requests = statement.identities()
+                    .iter()
+                    .cartesian_product(statement.operations())
+                    .cartesian_product(statement.resources())
+                    .map(|item| Request::new(item.0.0, item.0.1, item.1)
+                                .expect("unable to create a request")
+                        ).collect::<Vec<_>>();
```
dmolokanov, line 868:
```suggestion
                        );
```
I believe you don't need to allocate a vec only to iterate it later in for loop

dmolokanov (dismissed):
Overall looks good to me. I have some concerns about changing visibility to `pub(crate)` for fields of the definition and some other nits.

mqtt/policy/src/core/mod.rs
```diff
@@ -843,4 +833,98 @@ pub(crate) mod tests {
...
+                    Effect::Deny => Decision::Denied,
+                };
+
+                // collect all combos of identity/operation/resource
+                // in the statement.
+                let requests = statement.identities()
+                    .iter()
+                    .cartesian_product(statement.operations())
+                    .cartesian_product(statement.resources())
+                    .map(|item| Request::new(item.0.0, item.0.1, item.1)
```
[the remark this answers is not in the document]
vadim-kovalyov: yeah... there is a known issue with nested tuples and rustfmt which should be fixed in upcoming release. https://github.com/rust-lang/rustfmt/issues/4355

mqtt/policy/src/core/mod.rs
vadim-kovalyov: Good suggestion, but at the moment I don't see a strong reason to do that.

mqtt/policy/src/core/mod.rs
vadim-kovalyov: `Effect` is a part of policy definition. It can be "undefined" in certain cases. We used to have two `Effect` enums, in mod.rs and `builder.rs`. They are a bit different, but probably can be combined together. I have some ideas I'll explore in the following PR.

`Decision` is a result of policy evaluation. It looks similar, but conceptually a different enum.

mqtt/policy/src/core/mod.rs
vadim-kovalyov: It is possible, and in that case every request will return default decision. We don't need to cover that in proptests though, I believe we have a unit test for that.

mqtt/policy/src/core/mod.rs
vadim-kovalyov: I needed to materialize the collection, b/c I can't continue to use borrowed statement after I create a policy from it. Policy consumes the definition.

mqtt/policy/src/core/mod.rs
```diff
@@ -843,4 +833,98 @@ pub(crate) mod tests {
...
+                // take very first statement, which should have top priority.
+                let statement = &definition.statements()[0];
+                let expected = match statement.effect() {
+                    Effect::Allow => Decision::Allowed,
+                    Effect::Deny => Decision::Denied,
+                };
+
+                // collect all combos of identity/operation/resource
+                // in the statement.
+                let requests = statement.identities()
```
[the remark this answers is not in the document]
vadim-kovalyov: oh, cool!

mqtt/policy/src/core/mod.rs
dmolokanov: Can you please put some docs in the comments for each of them covering differences, especially Effect/EffectImpl?

mqtt/policy/src/core/mod.rs
dmolokanov: ah, that's right!

mqtt/policy/src/core/mod.rs
```diff
@@ -859,13 +859,13 @@ pub(crate) mod tests {
...
                 // collect all combos of identity/operation/resource
                 // in the statement.
-                let requests = statement.identities()
-                    .iter()
-                    .cartesian_product(statement.operations())
-                    .cartesian_product(statement.resources())
-                    .map(|item| Request::new(item.0.0, item.0.1, item.1)
-                                .expect("unable to create a request")
-                        ).collect::<Vec<_>>();
+                let requests = iproduct!(
```
dmolokanov, line 862: neat!

dmolokanov (approved):
:shipit:

PR #3988, opened 2020-11-18
Try to stabilize policy engine e2e tests.

vadim-kovalyov:
I noticed that one of the policy e2e tests is flaky. 

The goal of the test is to verify that we can update the policy "on the fly".

Previous version of the test was unstable b/c there is no reliable way to notify the test when policy update actually finished, and when DeviceClient tries to connect again, it can still be authorized (when exception is expected) because new policy did not take effect.

So I'm changing a logic to be the opposite: 
- initial deployment denies the connection for device1
- DeviceClient for device1 tries to connect and fails (expected exception)
- update the deployment to allow the connection for device2
- DeviceClient for device2 is able to connect. (And this is the trick, since DeviceClient has retry strategy, it will retry until policy is updated and takes effect).

I needed to use different device ids since there are some issues with SDK if I try to reuse one.

### DIFF
--- a/test/Microsoft.Azure.Devices.Edge.Test/AuthorizationPolicy.cs
+++ b/test/Microsoft.Azure.Devices.Edge.Test/AuthorizationPolicy.cs
@@ -5,7 +5,6 @@ namespace Microsoft.Azure.Devices.Edge.Test
     using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
-    using Microsoft.Azure.Devices.Client;
     using Microsoft.Azure.Devices.Client.Exceptions;
     using Microsoft.Azure.Devices.Edge.Test.Common;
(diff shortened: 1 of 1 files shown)


dylanbronson (approved):
Nice clean way of switching it up to be less flaky!
I'm afraid you'll have to fix merge conflict, as I just disabled this test.
(Also, I assume you've run this locally for testing)

vadim-kovalyov:
> Nice clean way of switching it up to be less flaky!
> I'm afraid you'll have to fix merge conflict, as I just disabled this test.
> (Also, I assume you've run this locally for testing)

vadim-kovalyov:
> Nice clean way of switching it up to be less flaky!
> I'm afraid you'll have to fix merge conflict, as I just disabled this test.
> (Also, I assume you've run this locally for testing)

Yep, I verified them on my box, although it is hard to repro the flakiness.

PR #3991, opened 2020-11-19
outPrefix and inPrefix None when empty string

ancaantochi:


dmolokanov (approved):
LGTM. I wonder if we need to introduce `NonEmptyString` type. There are several occasions already we need that.

PR #3992, opened 2020-11-19
Policy engine: getting rid of EffectImpl

vadim-kovalyov:
This is a small refactoring to policy engine internals. No functional changes. Just addressing some feedback from previous PRs.

Before:
We had two similar enums:
- `Effect` (used in `PolicyDefinition`)
- `EffectImpl` - internal struct for calculating the policy decision. Very similar to `Effect` but had additional `Undefined` variant.

Now:
- Instead of `Undefined` variant, I can use `Option<Effect>`. So `EffectImpl` can be retired.


dmolokanov (approved):
Neat! I like this more! 👍

PR #3995, opened 2020-11-19
Try to stabilize policy e2e tests.

vadim-kovalyov:
Second attempt...

The goal of the test is to verify that we can update the policy "on the fly".

Previous version of the test was unstable b/c there is no reliable way to notify the test when policy update actually finished, and when DeviceClient tries to connect again, it can still be unauthorized.

So I've introduced explicit retry policy, b/c default retry policy in the `DeviceClient` does not handle authN/authZ exceptions.

### DIFF
--- a/test/Microsoft.Azure.Devices.Edge.Test/AuthorizationPolicy.cs
+++ b/test/Microsoft.Azure.Devices.Edge.Test/AuthorizationPolicy.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Azure.Devices.Edge.Test
     using Microsoft.Azure.Devices.Edge.Test.Helpers;
     using Microsoft.Azure.Devices.Edge.Util;
     using Microsoft.Azure.Devices.Edge.Util.Test.Common.NUnit;
+    using Microsoft.Azure.Devices.Edge.Util.TransientFaultHandling;
     using NUnit.Framework;
 
     [EndToEnd]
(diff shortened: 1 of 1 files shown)


test/Microsoft.Azure.Devices.Edge.Test/AuthorizationPolicy.cs
```diff
@@ -134,34 +135,24 @@ public async Task AuthorizationPolicyUpdateTest()
...
-                CertificateAuthority.GetQuickstart(),
-                this.iotHub,
-                token,
-                Option.None<string>());
+            // There is no reliable way to signal when the policy
+            // is updated in $edgehub, so need to retry several times.
+            //
+            // DefaultProgressive => 55 sec max.
+            LeafDevice leaf = null;
+            await RetryPolicy.DefaultProgressive.ExecuteAsync(async () =>
```
dylanbronson, line 143: So now we basically just poll for a while?

dylanbronson (approved):
LGTM as long as it's passing locally

test/Microsoft.Azure.Devices.Edge.Test/AuthorizationPolicy.cs
vadim-kovalyov: yep...

dylanbronson (approved):

PR #3998, opened 2020-11-20
Extract client io object creation from bridge into mqtt-util crate 

and-rewsmith:
We are going to be adding some rust test modules which will need to generate tokens and get the trust bundle. This logic currently exists in the bridge. We should reuse this code so I have made a mqtt-util crate we can ref from our new rust test modules. I then refactored the bridge to use this mqtt-util crate also.


mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -156,92 +159,6 @@ impl ConnectionSettings {
...
-
-    pub fn generation_id(&self) -> &str {
-        &self.generation_id
-    }
-
-    pub fn workload_uri(&self) -> &str {
-        &self.workload_uri
-    }
-}
-
```
and-rewsmith, line 244: I thought it was a good idea to move these settings into the util crate since test modules will be using these too.

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -1,191 +1,27 @@
 #![allow(clippy::default_trait_access)] // Needed because mock! macro violates
 use core::{convert::TryInto, num::TryFromIntError};
```
and-rewsmith, line 2: Where should core go?

mqtt/mqtt-util/src/client_io.rs
```diff
@@ -0,0 +1,404 @@
...
+pub enum ClientIoSource {
+    Tcp(TcpConnection<SasTokenSource>),
+    Tls(TcpConnection<SasTokenSource>),
+}
+
+pub trait ClientIo: AsyncRead + AsyncWrite + Send + Sync + 'static {}
+
+impl<I> ClientIo for I where I: AsyncRead + AsyncWrite + Send + Sync + 'static {}
+
+type BridgeIoSourceFuture =
```
ancaantochi, line 33: nit: rename to ClientIoSourceFuture

mqtt/mqtt-util/src/client_io.rs
```diff
@@ -0,0 +1,404 @@
+use core::{convert::TryInto, num::TryFromIntError};
+use std::{
+    io::{Error, ErrorKind},
+    pin::Pin,
+    str,
+};
+
+use anyhow::Result;
```
ancaantochi, line 8: I think it's not needed to use anyhow::Result and use std::result::Result

ancaantochi (approved):

PR #3999, opened 2020-11-20
Reduce mutability in the broker

vadim-kovalyov:
No functional changes. In preparation to some other work, I'd like to push out some minor refactoring to reduce mutability in the code.


dmolokanov (approved):
Love that!

PR #4004, opened 2020-11-21
Module runtime binary and package rename

massand:
This change finalizes the binary and package names for the Edge Module runtime. 
- Reflected all the final binary naming and file locations as per the [packaging document](https://github.com/Azure/iot-identity-service/blob/main/docs/packaging.md)
- Removed deb8 and added Ubuntu 20.04 package support

Validated manually E2E on Ubuntu 18.04 only.


gordonwang0 (approved):
Assuming this is just a rename (I didn't look through the entire change), I approve.

PR #4006, opened 2020-11-23
Drop session on re-authorize

vadim-kovalyov:
Currently, when authorization policy changes, broker re-evaluates all the existing connected or offline sessions to see if they violate the new rules. The problem is that for a persistent session the broker does not drop it, but just disconnects the client. As a result, there is a possibility for some existing subscriptions to evade authorization.

In this PR instead of disconnecting the client, we will drop the whole session if it violates any of the connect or subscribe rules.


mqtt/mqtt-broker/tests/integration.rs
```diff
@@ -0,0 +1,87 @@
...
+    let broker_handle = broker.handle();
+    let mut server_handle = start_server(broker, DummyAuthenticator::with_id("device_1"));
+
+    // connect a client with persistent session.
+    let mut client = TestClientBuilder::new(server_handle.address())
+        .with_client_id(ClientId::IdWithExistingSession("device_1".into()))
+        .build();
+
+    // connect another client with persistent session that is always authorized.
+    let root = TestClientBuilder::new(server_handle.address())
```
dylanbronson, line 34: Can you name this root_client or something similar so we know it's also a client?

vadim-kovalyov: Sure, I'll do that in the next pr for session cleanup.

dmolokanov (approved):
LGTM

PR #4008, opened 2020-11-23
Re-enable non-flaky tests

and-rewsmith:
These tests aren't flaky. I ran the CI pipeline 20 times and they didn't fail. There is one failing edge agent connection test I added some notes for, but this is consistently failing due to outstanding work on the hub side.


dylanbronson (approved):
LGTM if they're not failing

PR #4009, opened 2020-11-24
Prevent stackflow when syncing circularly nested hierarchies

richma-ms:
The stack overflow comes from us trying to clean up the children authchains when we detect that the hierarchy has exceeded maximum depth.  But the cleanup logic is also recursive, so this change limits our recursion maxDepth * 2, which would cover all valid scenarios.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
@@ -382,13 +382,23 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
                 }
             }
 
-            void RemoveAuthChain()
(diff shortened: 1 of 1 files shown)


huguesBouvier (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ServiceIdentityTree.cs
```diff
@@ -382,13 +382,23 @@ void UpdateAuthChainFromParent(ServiceIdentityTreeNode parentNode, int traveled)
                 }
             }
 
-            void RemoveAuthChain()
+            void RemoveAuthChain() => this.RemoveAuthChainRecursive(0);
+
+            void RemoveAuthChainRecursive(int traveled)
             {
+                if (traveled >= 2 * MaxNestingDepth)
```
varunpuranik, line 389: Test?

PR #4013, opened 2020-11-24
Update E2E test files for new configuration format

gordonwang0:


test/Microsoft.Azure.Devices.Edge.Test.Common/windows/EdgeDaemon.cs
```diff
@@ -64,14 +64,20 @@ public async Task InstallAsync(Option<string> packagesPath, Option<Uri> proxy, C
 
         public async Task ConfigureAsync(Func<DaemonConfiguration, Task<(string, object[])>> config, CancellationToken token, bool restart)
         {
-            string configYamlPath =
-                Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData) + @"\iotedge\config.yaml";
+            // These tests do not run on Windows; use empty file paths.
```
massand, line 67: Better to wipe the Windows and Kubernetes E2E tests than update them with stubs. That is what I have been doing throughout the code.

gordonwang0: Removed Windows tests.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -49,8 +66,8 @@ public string[] GetInstallCommandsFromLocal(string path)
                 SupportedPackageExtension.Rpm => new[]
                 {
                     "set -e",
-                    $"yum install -y {string.Join(' ', packages)}",
-                    "pathToSystemdConfig=$(systemctl cat iotedge | head -n 1)",
+                    $"rpm --nodeps -i {string.Join(' ', packages)}",
```
massand, line 69: Nit question: why switch this?

gordonwang0: The version numbers between the aziot-identity-service and aziot-edge packages were incompatible:
```
Error: Package: aziot-edge-1.1.0-0.1.dev20201123.5.el7.x86_64 (/aziot-edge-1.1.0-0.1.dev20201123.5.el7.x86_64)
           Requires: aziot-identity-service >= 1.1.0-0.1.dev20201123.5
```

`yum` was refusing to install aziot-edge without an appropriately numbered aziot-identity-service, so I changed it to `rpm --nodeps` which ignores the dependencies.

builds/e2e/templates/e2e-setup.yaml
```diff
@@ -63,6 +65,34 @@ steps:
...
+        $artifacts_link | ConvertFrom-JSON
+    $download_link = $artifacts.artifacts | where {$_.name -eq '$(identityServiceArtifactName)'} | `
+        Select-Object -ExpandProperty archive_download_url
+
+    Invoke-WebRequest -Headers $github_headers $download_link -OutFile iot-identity-service.zip
+    Expand-Archive -Path iot-identity-service.zip -DestinationPath iot-identity-service
+
+    $packages = Get-ChildItem -Recurse iot-identity-service -Filter $(identityServicePackageFilter)
+    $packagePath = Convert-Path '$(System.ArtifactsDirectory)/$(artifactName)'
+    Copy-Item $packages -Destination $packagePath
```
massand, line 93: This is fine for now. We should discuss long term plans after this PR. This assumes that the aziot commit id is always coming from mainline (which isn't always true if we decided to do preview versions). Possible solutions - 
1) Stand up a new pipeline for iot-identity-service (which is redundant with Github Actions).; Maybe we switch everything to Github Actions?
2) Allow passing in branch and commit id as parameters into the pipeline. If not specified, we can do what you just did here.

WDYT?

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -88,13 +105,13 @@ public string[] GetInstallCommandsFromLocal(string path)
         {
             SupportedPackageExtension.Deb => new[]
             {
-                "apt-get purge --yes libiothsm-std iotedge"
+                "dpkg --purge libiothsm-std aziot-edge aziot-identity-service iotedge"
```
massand, line 108: Since `libiothsm-std` and `iotedge` pakages are never installed, we don't need to clean them up.

builds/e2e/templates/e2e-setup.yaml
gordonwang0: Option 2 makes sense IMO. We'll probably need to discuss further with other test automation folks.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
gordonwang0: They may be installed on the test machines because a previous pipeline ran against `master`.

massand (changes requested):
Minor changes

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
massand: The cleanup from previous tests should handle it. The new tests should cleanup their own installed packages.

gordonwang0: Currently, the test tear down on master doesn't uninstall packages. It only deletes created devices.

I remember running into some issues when running on Pipelines from having both iotedge (from a previous run) and aziot-edge installed. Purging the packages before and after the run fixed that.

massand (approved):
:shipit:

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
massand: It is a tad annoying to cleanup old packages, but I see no way out of this for now. Thanks for patiently explaining!

test/Microsoft.Azure.Devices.Edge.Test/SetupFixture.cs
```diff
@@ -96,6 +117,24 @@ public async Task BeforeAllAsync()
...
                     }
+
+                    // Remove packages installed by this run.
+                    await this.daemon.UninstallAsync(token);
+
+                    // Delete test certs, keys, etc.
+                    Directory.Delete("/etc/aziot/e2e_tests", true);
+
+                    // Restore backed up config files.
+                    foreach (string file in this.configFiles)
```
vadim-kovalyov, line 128: @gordonwang0, I noticed that this piece of code has been removed from master but still in release/1.2. Can you please check that this is alright?

PR #4017, opened 2020-11-25
Add message cleanup interval environment variable to edgeHub

dylanbronson:
Adding 'messageCleanupIntervalSecs' environment variable to edgeHub in order to control the interval at which we cleanup our messages (if that wasn't obvious from its name :) ) in MessageStore.

Making it non-optional for MessageStore and changing checkEntireQueueOnCleanup to be non-optional as well. Their defaults are chosen in the DependencyManager if no values is supplied by the user.

Also removing a delay between cleaning queues, as it seems unnecessary.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -38,14 +38,14 @@ public class MessageStore : IMessageStore
         readonly IStoreProvider storeProvider;
         TimeSpan timeToLive;
 
-        public MessageStore(IStoreProvider storeProvider, ICheckpointStore checkpointStore, TimeSpan timeToLive, bool checkEntireQueueOnCleanup = false)
+        public MessageStore(IStoreProvider storeProvider, ICheckpointStore checkpointStore, TimeSpan timeToLive, bool checkEntireQueueOnCleanup = false, int messageCleanupIntervalSecs = -1)
```
varunpuranik, line 41: Option?

varunpuranik (dismissed):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -192,25 +192,28 @@ public MessageWrapper(IMessage message, DateTime timeStamp, int refCount)
         class CleanupProcessor : IDisposable
         {
             const int CleanupBatchSize = 10;
-            static readonly TimeSpan CleanupTaskFrequency = TimeSpan.FromMinutes(30); // Run once every 30 mins.
-            static readonly TimeSpan MinCleanupSleepTime = TimeSpan.FromSeconds(30); // Sleep for 30 secs
+            static readonly TimeSpan DefaultCleanupTaskFrequency = TimeSpan.FromMinutes(30); // Check every 30 min that it is still running
```
richma-ms, line 195: >DefaultCleanupTaskFrequency [](start = 37, length = 27)

nit: this should probably be renamed to avoid future confusion, right now the naming implies that the cleanup task runs every 30min.  Maybe change it to something like CleanupTaskCheckFrequency

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/storage/MessageStore.cs
```diff
@@ -380,9 +382,23 @@ async Task<bool> DeleteMessageCallback(long offset, MessageRef messageRef)
...
+                    // Must wait MinCleanupSleepTime, even if given interval is lower.
+                    return TimeSpan.FromSeconds(Math.Max(this.messageCleanupIntervalSecs, MinCleanupSleepTime.TotalSeconds));
+                }
+                else
+                {
+                    // If MessageCleanupIntervalSecs isn't given, we will wait TTL/2 or 30 min - whichever is lesser
+                    // But we have a hard cap floor of MinCleanupSleepTime as the minimum interval
+                    double seconds = this.messageStore.timeToLive.TotalSeconds / 2 < DefaultCleanupTaskFrequency.TotalSeconds
+                        ? this.messageStore.timeToLive.TotalSeconds / 2
+                        : DefaultCleanupTaskFrequency.TotalSeconds;
```
richma-ms, line 398: I think this is just Math.Min()?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -208,6 +208,7 @@ void RegisterAmqpModule(ContainerBuilder builder)
             int configUpdateFrequencySecs = this.configuration.GetValue("ConfigRefreshFrequencySecs", 3600);
             TimeSpan configUpdateFrequency = TimeSpan.FromSeconds(configUpdateFrequencySecs);
             bool checkEntireQueueOnCleanup = this.configuration.GetValue("CheckEntireQueueOnCleanup", false);
+            int messageCleanupIntervalSecs = this.configuration.GetValue("MessageCleanupIntervalSecs", -1);
```
richma-ms, line 211: >-1 [](start = 103, length = 2)

what if we default the value to 1800 here?  Wouldn't that make the MessageStore logic simpler as we wouldn't need to separate hardcode another default value.

dylanbronson: This will change the current behavior where TTL/2 controls cleanup interval, but it doesn't matter since this env var addition will solve that control anyway

richma-ms (approved):

PR #4018, opened 2020-11-25
Remove new package names from test agents

gordonwang0:
On the iiot branch, the package names were renamed to aziot-*. Remove these packages (if installed) when running tests on the master branch.


gordonwang0:
Will add to cleanup of tests on iiot.

PR #4019, opened 2020-11-25
Session cleanup for expired offline sessions

vadim-kovalyov:
For a long time we had a session expiration setting (default is 60d) that has not been used. In this PR I introduced a cleanup process for expired sessions in the Broker. 

Key notes:
- Only offline sessions that past expiration date are removed. 
- I've added `last_active` field into ~~`SessionState`~~ `OfflineSession` to keep track of when the session was last active.
- I've added a new setting: `cleanup_interval`, that is the frequency of the cleanup job. Default is `1d`.
- There are a lot of affected files in this PR, but most of them are just plumbing of `last_active` field. The most interesting parts are in `broker.rs` and `cleanup.rs`.
-


mqtt/mqtt-broker/src/session/state/mod.rs
```diff
@@ -2,6 +2,7 @@ mod map;
 mod queue;
 mod set;
 
+use chrono::{DateTime, Utc};
```
dmolokanov, line 5: this is misplaced

mqtt/mqtt-broker/src/session/offline.rs
```diff
@@ -35,6 +36,10 @@ impl OfflineSession {
         self.state.subscriptions()
     }
 
+    pub fn last_active(&self) -> &DateTime<Utc> {
```
dmolokanov, line 39:
```suggestion
    pub fn last_active(&self) -> DateTime<Utc> {
```
since `DateTime<Tz>` implements Copy

dmolokanov (commented):
You added `last_active` field for the session in any state, but you don't update it for every client activity: client published a message, subscribed/unsubscribed, sent a ping, etc.. 

I wonder if we need to make it more consistent. I see a couple of options: 
1. make sure this field always represents the timestamp of the last client activity which is too much work I guess. 
2. make this field as `Offline` session attribute only in a way it represents the timestamp of when the session became inactive.

You may have other ideas. Let's discuss offline

mqtt/mqtt-broker/src/settings/mod.rs
```diff
@@ -93,12 +97,21 @@ impl SessionConfig {
...
+    pub fn cleanup_interval(&self) -> Duration {
+        self.cleanup_interval
+    }
 }
 
 impl Default for SessionConfig {
     fn default() -> Self {
         SessionConfig::new(
             Duration::from_secs(60 * DAYS),
+            Duration::from_secs(DAYS), // 1d
```
dmolokanov, line 114:
```suggestion
            Duration::from_secs(1*DAYS), 
```
?

mqtt/mqtt-generic/src/settings.rs
```diff
@@ -176,6 +176,7 @@ mod tests {
                     RetainedMessagesConfig::new(1000, Duration::from_secs(60 * DAYS)),
                     SessionConfig::new(
                         Duration::from_secs(60 * DAYS),
+                        Duration::from_secs(DAYS), // 1d
```
dmolokanov, line 179:
```suggestion
                        Duration::from_secs(1*DAYS), 
```
?

mqtt/mqttd/src/app/cleanup.rs
```diff
@@ -0,0 +1,36 @@
...
+    let tick = tick_cleanup(cleanup_interval, expiration, broker_handle);
+    tokio::spawn(tick);
+
+    Ok(())
+}
+
+async fn tick_cleanup(period: StdDuration, expiration: Duration, broker_handle: BrokerHandle) {
+    info!("cleaning up expired sessions every {:?}", period);
+    let start = Instant::now() + period;
+    let mut interval = tokio::time::interval_at(start, period);
```
dmolokanov, line 27:
```suggestion
    let mut interval = time::interval_at(start, period);
```

mqtt/mqtt-generic/src/settings.rs
dmolokanov: there are more places with similar suggestion

dmolokanov (commented):
can/should we make the cleanup task a sidecar?

mqtt/mqtt-broker/src/settings/mod.rs
vadim-kovalyov: Clippy doesn't like that.

mqtt/mqtt-generic/src/settings.rs
vadim-kovalyov: Clippy doesn't like that.

vadim-kovalyov:
> can/should we make the cleanup task a sidecar?

We probably could, but in that case we need to refactor both snapshotter and cleanup. We can investigate that later.

mqtt/mqttd/src/app/mod.rs
```diff
@@ -107,7 +112,13 @@ pub trait Bootstrap {
...
 
     /// Returns update interval for snapshotter.
-    fn snapshot_interval(&self, settings: &Self::Settings) -> std::time::Duration;
+    fn snapshot_interval(&self, settings: &Self::Settings) -> Duration;
+
+    /// Returns session expiration.
+    fn session_expiration(&self, settings: &Self::Settings) -> Duration;
+
+    /// Returns session cleanup interval.
+    fn session_cleanup_interval(&self, settings: &Self::Settings) -> Duration;
```
and-rewsmith, line 121: nit: If these can be obtained from the Settings do we need them in bootstrap? 

The pattern we have now is to pass Settings into bootstrap, then don't touch settings again. But this seems to cause some duplicated code. Maybe we should be extracting from Settings directly for simple cases?

mqtt/mqttd/src/app/cleanup.rs
```diff
@@ -0,0 +1,36 @@
...
+pub async fn start_cleanup(
+    broker_handle: BrokerHandle,
+    cleanup_interval: StdDuration,
+    expiration: StdDuration,
+) -> Result<()> {
+    info!("starting cleanup job...");
+
+    let expiration = Duration::from_std(expiration)?;
+    let tick = tick_cleanup(cleanup_interval, expiration, broker_handle);
+    tokio::spawn(tick);
```
and-rewsmith, line 19: Do we need to handle shutdown?

and-rewsmith (approved):
LGTM with some minor comments. Why would we have to refactor snapshotter now for us to make this a sidecar?

mqtt/mqtt-broker/tests/integration.rs
```diff
@@ -2,18 +2,160 @@
...
+    // assert that offline session is removed.
+    // and only "online" client's session remains.
+    let (_, mut sessions) = server_handle.shutdown().await.into_parts();
+    assert_eq!(1, sessions.len());
+
+    let (client_info, _, _, _) = sessions.remove(0).into_parts();
+    assert_eq!("online_client", client_info.client_id().as_str());
+
+    // dispose a client.
+    online_client.shutdown().await;
```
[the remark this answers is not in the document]
vadim-kovalyov: `server_handle` implements drop that signals broker to stop. It is kind of important for tests to cleanup resources b/c multiple tests run in parallel.

You might be wondering why not implement `Drop` for `TestClient`. The problem is that `shutdown()` for `TestClient` is async, and Drop cannot be async.

mqtt/mqttd/src/app/cleanup.rs
vadim-kovalyov: good point. I kept it similar to the Snapshotter tick and skipped the shutdown. We can revisit that later.

mqtt/mqttd/src/app/mod.rs
vadim-kovalyov: Yeah, it seems weird, but here is what I understand. We have an `App<B>` struct that is a generic on `Bootstrap` which has associated type `Bootstrap::Settings`. So `App` does not know the type of Settings to extract any values from it, so `Bootstrap` provides "getters" via its interface. 

As an alternative, we can intro `Settings` trait and bound `Bootstrap::Settings` by it. And both generic and edgehub settings have to implement the `Settings` trait. Which will be the same amount of code, imo.

mqtt/mqtt-broker/src/session/state/mod.rs
```diff
@@ -68,18 +72,30 @@ impl SessionState {
...
+            last_active,
+        )
+    }
+
+    pub fn into_snapshot(self, last_active: DateTime<Utc>) -> SessionSnapshot {
+        SessionSnapshot::from_parts(
+            self.client_info,
+            self.subscriptions,
+            self.waiting_to_be_sent.into_inner(),
+            last_active,
```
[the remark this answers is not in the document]
vadim-kovalyov: `SessionSnapshot` is used for persistence and need only subset of the data. In addition to that (and the main reason why `SessionSnapshot` was introduced) `SessionState` has `SessionConfig` and uses `BoundedQueue`. It is possible, that queue settings can change offline, and when broker comes online it needs to go over stored messages and drop some. That's done during conversion from `SessionSnapshot` to `SessionState`

PR #4020, opened 2020-11-25
Fix cleanup processor

dylanbronson:
Adding bool.TrueString to metrics.Set call so that labels are not null. Without this, cleanup processor fails every time it sets metrics and throws an error.

This fixes the issue.


davilu (approved):
Nice catch.

PR #4028, opened 2020-11-30
Enable PlugAndPlay for MQTT broker

dylanbronson:
Currently, PlugAndPlay devices cannot connect through Edge's MQTT Broker. This is because the broker does not store the modelId of devices/modules for later use.

This PR enables PlugAndPlay through the broker by storing any given ModelId for the identity in the MetadataStore when that identity is authenticated by EdgeHub Core. The broker will ask EH Core for authentication, and the EH Core's AuthAgentProtocolHead will do the authentication and store the ModelId in the MetadataStore, if is authenticated.

The metadataStore is already checked for a ModelId when EH is connecting to the cloud on behalf of an identity, so no change is needed there.

This change does NOT enable PlugAndPlay for the Mqtt Bridge, so nested scenario + Mqtt broker will NOT work yet. More work will need to be done to send the ModelId (and all of the username add-ons) upstream through the broker

Additionally, in a PR immediately after this one, I will add E2E tests for this case


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/AuthAgentListenerTest.cs
```diff
@@ -345,6 +346,31 @@ public async Task DeniesBadContentLengthLongBody()
...
+                content.version = "2020-04-20";
+                content.username = "testhub/device/api-version=2018-06-30&model-id=dtmi:test:modelId;1";
+                // [SuppressMessage("Microsoft.Security", "CS002:SecretInNextLine", Justification="Synthetic password used in tests")]
+                content.password = "somepassword";
+
+                dynamic response = await PostAsync(content, URL);
+
+                Assert.Equal(200, (int)response.result);
+                var modelId = (await metadataStore.GetMetadata("device")).ModelId;
+                Assert.True(modelId.HasValue);
```
vadim-kovalyov, line 370: nit: do you want to compare model id with what you set above?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/AuthAgentListenerTest.cs
```diff
@@ -345,6 +346,31 @@ public async Task DeniesBadContentLengthLongBody()
...
+            IEntityStore<string, string> store = storeProvider.GetEntityStore<string, string>("productInfo");
+            var metadataStore = new MetadataStore(store, "productInfo");
+            using (var sut = new AuthAgentProtocolHead(authenticator, metadataStore, usernameParser, credFactory, sysIdProvider, config))
+            {
+                await sut.StartAsync();
+
+                dynamic content = new ExpandoObject();
+                content.version = "2020-04-20";
+                content.username = "testhub/device/api-version=2018-06-30&model-id=dtmi:test:modelId;1";
+                // [SuppressMessage("Microsoft.Security", "CS002:SecretInNextLine", Justification="Synthetic password used in tests")]
```
vadim-kovalyov, line 363: nit: is this still needed?

vadim-kovalyov (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentController.cs
```diff
@@ -57,7 +59,8 @@ await this.GetIdsFromUsername(request)
                                     async creds => (await this.AuthenticateAsync(creds), Option.Some(creds)),
                                     () => Task.FromResult((false, Option.None<IClientCredentials>())));
 
-                return this.Json(GetAuthResult(isAuthenticated, credentials));
+                var authResult = await GetAuthResult(isAuthenticated, credentials, this.metadataStore);
```
vipeller, line 62: Would not be better to create a new StoreMetadata(credentials, this.metadataStore) method, and call that separately (if isAuthenticated == true), so then no need to change the GetAuthResult()? I feel it should not be the concern of GetAuthResult() to do other things than create the object that represents the result.

dylanbronson: Good call. I'll make the change.

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/AuthAgentListenerTest.cs
dylanbronson: oops forgot to do that

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/AuthAgentListenerTest.cs
dylanbronson: Looks like it is based on the fact that it was committed like this for CredScan

vipeller (approved):

vadim-kovalyov (approved):

PR #4029, opened 2020-11-30
Fixing ValidateMetrics E2E (#3767)

yophilav:
Currently the `ValidateMetrics` parses [metric readme page](https://github.com/Azure/iotedge/blob/master/doc/BuiltInMetrics.md) to obtain all the metric keys. The test then deploy the system runtime modules and populate its metrics which then are compared against the parsed-document set.
There are a handful of not-so-positive metrics that are not being populated if the system modules operated on a happy path during the E2E test. These not-so-positive metrics are currently causing the `ValidateMetrics` to fail since they are not being populated.

Cherry-picked: https://github.com/Azure/iotedge/commit/f443dde70a7aea9ad6716cf61cc66f89b6642783


dylanbronson (approved):
LGTM!

PR #4031, opened 2020-12-01
Test module for generic mqtt telemetry

and-rewsmith:
Adding the skeleton for the generic mqtt test module. This module will exist in connectivity and longhaul. This module exists on the lowest edge node and one node up. The lowest edge node will send to the upper node which will send a message back. When the lowest edge will then receive this relayed message and report to the TRC.

More specifically:
```
/// Module will run in one of two modes. The behavior depends on this mode.
///
/// 1: Test module runs on the lowest node in the topology.
///     - Spawn a thread that publishes messages continuously to upstream edge.
///     - Receives same message routed back from upstream edge and reports the result to the TRC.
///
/// 2: Test module runs on middle node in the topology.
///     - Receives a message from downstream edge and relays it back to downstream edge.
```

In addition to the skeleton I have also added the new() implementation for most of the abstractions.


mqtt/mqtt-broker-tests-util/src/client.rs
```diff
@@ -292,3 +297,70 @@ where
...
+) -> ClientIoSource {
+    let credentials = Credentials::Provider(credential_provider_settings);
+    let trust_bundle_source = TrustBundleSource::new(credentials.clone());
+    let token_source = SasTokenSource::new(credentials.clone());
+    let addr = "edgeHub:8883";
+    let tcp_connection = TcpConnection::new(addr, Some(token_source), Some(trust_bundle_source));
+    let io_source = ClientIoSource::Tls(tcp_connection);
+
+    io_source
+}
```
and-rewsmith, line 366: We could make an abstraction for all this logic or leave it as functions.

test/modules/GenericMqttTester/src/message_handler.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl SendBackMessageHandler {
+    pub fn new() -> Self {
+        let (publication_sender, publication_receiver) =
+            mpsc::unbounded_channel::<ReceivedPublication>();
+        let (shutdown_send, shutdown_recv) = mpsc::channel::<()>(1);
+        let shutdown_handle = ShutdownHandle::new(shutdown_send);
+
+        // needs to run in a separate thread and receive messages through
+        // channel so we don't block polling the client
```
and-rewsmith, line 56: Both message handlers need to do some async thing so both will follow this pattern.

test/modules/GenericMqttTester/src/tester.rs
```diff
@@ -0,0 +1,60 @@
...
+            publish_handle,
+            shutdown_handle,
+            message_handler,
+            shutdown_recv,
+        })
+    }
+
+    pub fn run() -> (JoinHandle<Result<(), MessageTesterError>>, ShutdownHandle) {
+        todo!()
+    }
```
and-rewsmith, line 59: Run method might (depending on settings) spawn new thread that will send messages continuously. It will also run its message handler. 

When shutdown signal is received then it will shutdown this thread and its message handler then exit.

and-rewsmith:
I am a bit concerned about the implications of building this module in our build pipeline. I think we need to look into improvements for the build pipeline so it can support more rust modules without bloating the time.

and-rewsmith:
I think we need to add a checkin for all rust test modules (only this one currently). I can do this in a separate PR.

test/modules/GenericMqttTester/src/message_handler.rs
```diff
@@ -0,0 +1,88 @@
...
+        let message_send_join_handle = tokio::spawn(Self::send_messages_back(
+            publication_receiver,
+            shutdown_recv,
+        ));
```
and-rewsmith, line 60: Need to find a way to get this join handle exposed so failures will be observable

test/modules/GenericMqttTester/src/message_handler.rs
```diff
@@ -0,0 +1,63 @@
...
+            publication_sender,
+            shutdown_handle,
+        }
+    }
+}
+
+impl MessageHandler for SendBackMessageHandler {
+    fn run(self) -> (JoinHandle<Result<(), MessageTesterError>>, ShutdownHandle) {
+        todo!()
+    }
```
and-rewsmith, line 58: Will spawn a thread to receive messages from channel and hande them. Returns this join handle and a shutdown handle that will stop the thread.

test/modules/GenericMqttTester/src/tester.rs
```diff
@@ -0,0 +1,61 @@
...
+        let client = client::create_client_from_module_env()
+            .map_err(MessageTesterError::ParseEnvironment)?;
+        let publish_handle = client
+            .publish_handle()
+            .map_err(MessageTesterError::PublishHandle)?;
+
+        let message_handler: Box<dyn MessageHandler> = match settings.test_scenario() {
+            TestScenario::Initiate => Box::new(SendBackMessageHandler::new(publish_handle.clone())),
+            TestScenario::Relay => Box::new(ReportResultMessageHandler::new()),
+        };
```
and-rewsmith, line 43: Is boxing this the correct approach?

test/modules/GenericMqttTester/src/message_handler.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/// Responsible for receiving publications and reporting result to the TRC.
+pub struct ReportResultMessageHandler {}
+
+impl ReportResultMessageHandler {
+    pub fn new() -> Self {
+        todo!()
+    }
+}
```
and-rewsmith, line 24: I am not yet sure how the report result handler will operate so skipping new() impl

test/modules/GenericMqttTester/src/message_handler.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    fn publication_sender_handle(&self) -> UnboundedSender<ReceivedPublication> {
+        todo!()
+    }
+}
+
+/// Responsible for receiving publications and sending them back to the downstream edge.
+pub struct SendBackMessageHandler {
+    publication_sender: UnboundedSender<ReceivedPublication>,
```
and-rewsmith, line 38: I think we need the receiver stored here too.

test/modules/GenericMqttTester/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum MessageTesterError {
+    #[error("could not get client publish handle")]
+    ParseEnvironment(#[source] VarError),
+
+    #[error("could not get client publish handle")]
+    PublishHandle(#[source] PublishError),
+
+    #[error("could not get client publish handle")]
```
and-rewsmith, line 21: need to fix error messages

test/modules/GenericMqttTester/src/message_handler.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn publication_sender_handle(&self) -> UnboundedSender<ReceivedPublication> {
+        todo!()
+    }
+}
+
+/// Responsible for receiving publications and sending them back to the downstream edge.
+pub struct RelayingMessageHandler {
+    publication_sender: UnboundedSender<ReceivedPublication>,
+    publication_receiver: UnboundedReceiver<ReceivedPublication>,
+    shutdown_handle: ShutdownHandle,
```
and-rewsmith, line 40: Also need to store the receiver.

test/modules/GenericMqttTester/src/tester.rs
```diff
@@ -0,0 +1,65 @@
...
+            settings,
+            client,
+            publish_handle,
+            shutdown_handle,
+            message_handler,
+            shutdown_recv,
+        })
+    }
+
+    pub fn run() -> Result<(), MessageTesterError> {
```
and-rewsmith, line 58: Should contain mut self, but this is fixed in the next PR

test/modules/GenericMqttTester/src/tester.rs
vadim-kovalyov: Seems fine to me. I can't think of another way to have polymorphism based on the runtime parameters without dynamic dispatching.

vadim-kovalyov (commented):
Nice work, seems clean and understandable. The only question I have is a circular dependency in mqtt-broker-tests-util.

test/modules/GenericMqttTester/src/message_handler.rs
```diff
@@ -0,0 +1,85 @@
...
+    shutdown_recv: Receiver<()>,
+    publish_handle: PublishHandle,
+}
+
+impl RelayingMessageHandler {
+    pub fn new(publish_handle: PublishHandle) -> Self {
+        let (publication_sender, publication_receiver) =
+            mpsc::unbounded_channel::<ReceivedPublication>();
+        let (shutdown_send, shutdown_recv) = mpsc::channel::<()>(1);
```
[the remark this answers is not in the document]
and-rewsmith: I tried using oneshot but couldn't derive clone for the ShutdownHandle. How do I get around this?

test/modules/GenericMqttTester/src/message_handler.rs
```diff
@@ -0,0 +1,85 @@
...
+    async fn run(self) -> Result<(), MessageTesterError>;
+
+    /// Sends a publication to be handled by the message handler
+    fn publication_sender_handle(&self) -> UnboundedSender<ReceivedPublication>;
+
+    // Get the shutdown handle to stop the run() method
+    fn shutdown_handle(&self) -> ShutdownHandle;
+}
+
+/// Responsible for receiving publications and reporting result to the TRC.
```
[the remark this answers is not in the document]
and-rewsmith: Discussed offline. TestResultCoordinator. Will add full name.

mqtt/mqtt-broker-tests-util/Cargo.toml
```diff
@@ -17,4 +17,5 @@ tracing = "0.1"
 tracing-subscriber = "0.1"
 
 mqtt3 = { path = "../mqtt3", features = ["serde1"] }
-mqtt-broker = { path = "../mqtt-broker" }
\ No newline at end of file
+mqtt-broker = { path = "../mqtt-broker" }
```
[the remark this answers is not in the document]
and-rewsmith: It does seem like a circular dependency to me, but everything builds. I think it might have something to do with tests util being a dev dependency for the broker?

test/modules/GenericMqttTester/src/message_handler.rs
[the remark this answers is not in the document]
vadim-kovalyov: oh, right, then mpsc is fine.

test/modules/GenericMqttTester/src/tester.rs
and-rewsmith: I mixed the cases up, but it is fixed in the next PR. If I make more changes to this branch I will fix this too.

vadim-kovalyov (approved):

PR #4035, opened 2020-12-02
Fix E2E tests on iiot branch

gordonwang0:


test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
```diff
@@ -146,6 +149,28 @@ async Task InternalStartAsync(CancellationToken token)
...
             await WaitForStatusAsync(ServiceControllerStatus.Running, token);
+
+            // Waiting for the processes to enter the "Running" state doesn't guarantee that
+            // they are fully started and ready to accept requests. Therefore, this function
+            // must wait until a request can be processed.
+            while (true)
+            {
+                var request = System.Diagnostics.Process.Start("iotedge", "list");
+
+                if (request.WaitForExit(1000))
```
massand, line 160: >WaitForExit [](start = 28, length = 11)

This would be true even if the process exited with an error. Should we check for non-zero exit code and retry?

gordonwang0: Fixed

test/Microsoft.Azure.Devices.Edge.Test/SetupFixture.cs
```diff
@@ -142,4 +146,49 @@ public async Task BeforeAllAsync()
...
+        }
+
+        public static async Task<TestCertificates> GenerateCertsAsync(string deviceId, CancellationToken token)
+        {
+            string scriptPath = Context.Current.CaCertScriptPath.Expect(
+                () => new System.InvalidOperationException("Missing CA cert script path"));
+            (string certPath, string keyPath, string password) = Context.Current.RootCaKeys.Expect(
+                () => new System.InvalidOperationException("Missing root CA"));
+
+            await OsPlatform.Current.InstallRootCertificateAsync(certPath, keyPath, password, scriptPath, token);
```
massand, line 170: The CertificateAuthority constructor appears to do a similar root CA install. How is this different?

gordonwang0: This is using the lower-level OsPlatform instead of CertificateAuthority. Fixed to used CertificateAuthority

massand:
:shipit:

PR #4045, opened 2020-12-03
Fix default settings for edge sockets

gordonwang0:
Change /var/run/aziot/edged so that it's owned by iotedge and not root.

### DIFF
--- a/edgelet/contrib/config/linux/config.yaml
+++ b/edgelet/contrib/config/linux/config.yaml
@@ -69,8 +69,8 @@ hostname: "<ADD HOSTNAME HERE>"
 # Nested Edge parent hostname
 ###############################################################################
 #
-# Optional. This entry should only be specified for the host name of the 
-# parent edge device for nested edge sceanrio. This value is injected into 
+# Optional. This entry should only be specified for the host name of the
+# parent edge device for nested edge sceanrio. This value is injected into
 # Edge Agent and Edge Hub modules as environment value 'IOTEDGE_GATEWAYHOSTNAME',
 # and injected into other module as environment value 'IOTEDGE_PARENTHOSTNAME'.
 # The parent_hostname value is converted to lower case before it is used.
@@ -127,8 +127,8 @@ hostname: "<ADD HOSTNAME HERE>"
 ###############################################################################
 
 connect:
-  management_uri: "unix:///run/aziot/edged/aziot-edged.mgmt.sock"
-  workload_uri: "unix:///run/aziot/edged/aziot-edged.workload.sock"
+  management_uri: "unix:///var/lib/aziot/edged/aziot-edged.mgmt.sock"
+  workload_uri: "unix:///var/lib/aziot/edged/aziot-edged.workload.sock"
 
 ###############################################################################
 # Listen settings
@@ -157,8 +157,8 @@ connect:
 ###############################################################################
 
 listen:
-  management_uri: "unix:///run/aziot/edged/aziot-edged.mgmt.sock"
-  workload_uri: "unix:///run/aziot/edged/aziot-edged.workload.sock"
+  management_uri: "unix:///var/lib/aziot/edged/aziot-edged.mgmt.sock"
+  workload_uri: "unix:///var/lib/aziot/edged/aziot-edged.workload.sock"
 
 ###############################################################################
 # Home Directory
@@ -180,11 +180,11 @@ homedir: "/var/lib/aziot/edged"
 # Additional container network configuration such as enabling IPv6 networking
 # and providing the IPAM settings can be achieved by specifying the relevant
 # configuration in the network settings.
-# 
-# Content Trust is enabled by specifying the location of root CAs for each registry 
-# using the fields below 
 #
-# content_trust - configuration to enable Content Trust feature 
+# Content Trust is enabled by specifying the location of root CAs for each registry
+# using the fields below
+#
+# content_trust - configuration to enable Content Trust feature
 # ca_certs - CA certs for each registry is specified under this configuration
 #
 # 'contoso1.azurecr.io' - Login server name of registry 1
@@ -197,7 +197,7 @@ homedir: "/var/lib/aziot/edged"
 #   ca_certs :
 #     'contoso1.azurecr.io' : "/path/to/root_ca_contoso1.crt"
 #     'contoso2.azurecr.io' : "/path/to/root_ca_contoso2.crt"
-# 
+#
 # network:
 #   name: "azure-iot-edge"
 #   ipv6: true
@@ -216,7 +216,7 @@ homedir: "/var/lib/aziot/edged"
 moby_runtime:
   uri: "unix:///var/run/docker.sock"
   # content_trust:
-  #   ca_certs: 
+  #   ca_certs:
   #     'contoso1.azurecr.io' : "/path/to/root_ca_contoso1.crt"
   #     'contoso2.azurecr.io' : "/path/to/root_ca_contoso2.crt"
   #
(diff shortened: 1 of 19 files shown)


arsing:
You shouldn't need the change in `debian/postinst`. For distros that use socket activation (ie everything except CentOS 7), the `aziot-edged.service` already has a `Requires` and `After` dependency on the socket units. Socket units already automatically create their parent directory when they start; that's why you can see the i-i-s repo doesn't have either of the two changes you've made here.

So you only need to change the CentOS 7-specific service unit, which you've done.

gordonwang0:
You're right, I don't need to change debian/postinst. I'll revert that.

I think I still need to change both aziot-edged.service and debian/aziot-edged.service though. It's true that the parent directories are all created, but they're created as owned by root with somewhat restrictive permissions. Because of this, aziot-edged isn't able to access the sockets in these directories. The RuntimeDirectory= changes in the .service files fix this by creating the parent directories owned by User=iotedge (as also specified in the .service files).

arsing:
>but they're created as owned by root with somewhat restrictive permissions. Because of this, aziot-edged (which runs under the iotedge user) isn't able to access the sockets in these directories.

I don't follow. For socket activation the `aziot-edged` process doesn't bind to any socket file. It gets the fd from systemd.

gordonwang0:
Without the change in this PR, here's what happens:
```
$ sudo systemctl start aziot-edged
2020-12-04T15:17:32Z [INFO] - Starting management API...
2020-12-04T15:17:32Z [INFO] - Starting workload API...
2020-12-04T15:17:32Z [INFO] - Starting watchdog with 60 second frequency...
2020-12-04T15:17:32Z [ERR!] - The daemon could not start up successfully: Could not start management service
2020-12-04T15:17:32Z [ERR!] -         caused by: An error occurred for path /run/aziot/edged/aziot-edged.mgmt.sock
2020-12-04T15:17:32Z [ERR!] -         caused by: Permission denied (os error 13)
aziot-edged.service: Main process exited, code=exited, status=1/FAILURE
aziot-edged.service: Failed with result 'exit-code'.
```

Before I start aziot-edged, the directory /var/run/aziot/edged does not exist. Upon starting aziot-edged, the directory gets created:
```
$ ls -all /var/run/aziot/edged/
drwxr-xr-x 2 root    root     80 Dec  4 07:17 .
drwxr-xr-x 3 root    root    120 Dec  4 07:17 ..
srw-rw---- 1 iotedge iotedge   0 Dec  4 07:17 aziot-edged.mgmt.sock
srw-rw-rw- 1 iotedge iotedge   0 Dec  4 07:17 aziot-edged.workload.sock
```

But the directory is created owned by root. That's the cause of the Permission denied error from the aziot-edged startup. To fix this, I either need to give the w permission to all users, or change the .service files so that /var/run/aziot/edged is owned by iotedge (what this PR does).

arsing:
Your `aziot-edged` is ignoring the socket activation and trying to bind the socket itself. That is wrong. Find out why it's doing that and fix it.

gordonwang0:
The issue was that the default config was using unix:/// URIs instead of fd:// URIs. Changing it to fd:// URIs works.

Should we still fix this though? Otherwise, this prevents people from using "unix:///run/aziot/edged/aziot-edged.mgmt.sock" in place of "fd://aziot-edged.mgmt.socket", despite the fact that they should be equivalent.

Also, I feel like we need to rename something: the fd URI needs to be "fd://aziot-edged.mgmt.socket" (to match the name of the .socket files), but everywhere else the socket is simply "aziot-edged.mgmt.sock".

Thoughts?

arsing:
>Should we still fix this though? Otherwise, this prevents people from using "unix:///run/aziot/edged/aziot-edged.mgmt.sock" in place of "fd://aziot-edged.mgmt.socket". I feel like these two should be equivalent.

The easiest way is to just remove the `fd://` scheme entirely and make it automatic based on the env vars, like we did in i-i-s.

>Also, I feel like we need to rename something: the fd URI needs to be "fd://aziot-edged.mgmt.socket" (to match the name of the .socket files), but everywhere else the socket is "aziot-edged.mgmt.sock".

Socket units must be called `.socket`, and socket files are usually called `.sock`, so it's correct as it is.

edgelet/contrib/config/linux/config.yaml
```diff
@@ -127,8 +127,8 @@ hostname: "<ADD HOSTNAME HERE>"
 ###############################################################################
 
 connect:
-  management_uri: "unix:///run/aziot/edged/aziot-edged.mgmt.sock"
-  workload_uri: "unix:///run/aziot/edged/aziot-edged.workload.sock"
+  management_uri: "unix:///var/lib/aziot/edged/aziot-edged.mgmt.sock"
```
arsing, line 130: What's the reason to move them back to homedir?

gordonwang0: I was going through contrib/centos/aziot-edge.spec and the sockets are created in %{iotedge_home}. I think they were only moved out on debian.

arsing: As I said on Teams:

>iotedged used to bind those sockets under its homedir, ie `/var/lib/iotedge` in that case. I think Rajeev changed it to be `/run/aziot/edged`, ie the same as if it was using socket activation (which should be fine)

If `RuntimeDirectory=` works on CentOS, there is no reason to move them back to the homedir.

arsing: Also:

>I was going through contrib/centos/aziot-edge.spec and the sockets are created in %{iotedge_home}.

That is pointless. Even if the socket files are created up-front they're not usable as sockets. Not only are they created as text files, even if they were created as socket files they would need to be deleted first because `bind()` needs to create them, which is what the code already does. Just delete that from the spec file.

gordonwang0: RuntimeDirectory= doesn't work on CentOS 7. So I'll have to try something else

Since iotedge is hardcoded to use "unix:///run/aziot/edged/aziot-edged.mgmt.sock" I'll have to change this anyways.

arsing: Okay, that's what I was looking for.

In that case, yes, the socket should go in homedir.

arsing: You also need to fix https://github.com/Azure/iotedge/blob/iiot/edgelet/Makefile#L99 so that `iotedge` uses the right default.

gordonwang0: Done. I've changed the sockets to be in /var/lib.

Also, it seems like creating /var/lib/aziot/edged/aziot-edge.*.sock is necessary in the .spec file. Removing it causes rpmbuild to fail: Installed (but unpackaged) file(s) found

arsing: Because they're being packaged. https://github.com/Azure/iotedge/blob/iiot/edgelet/Makefile#L131-L132 Don't package them.

gordonwang0: Fixed

arsing: Ahh, the reason this was originally done is because the code will not bind the sockets with the right permissions otherwise - https://github.com/Azure/iotedge/blob/iiot/edgelet/edgelet-http/src/unix.rs#L20 If the socket file already exists, then the existing mode is preserved. But if the file doesn't exist, then it's bound with the default umask.

That'll probably make the sockets reachable to all users, which is fine for workload but not for mgmt.

Threading down the permissions through the call chain is not trivial. For now, remove your last commit please.

gordonwang0: Done

arsing (approved):

PR #4046, opened 2020-12-03
PlugAndPlay broker e2e tests

dylanbronson:
We just enabled PlugAndPlay for the MQTT broker. This PR adds e2e tests for this scenario

Also adding in AMQP tests for broker while I'm at it.

I've kept the DeviceClient tests disabled. I believe that in order to fix these, we should switch from the HTTP request we're making now to using the Registry API (which is newly available)


test/Microsoft.Azure.Devices.Edge.Test/PlugAndPlay.cs
```diff
@@ -75,30 +82,71 @@ public async Task PlugAndPlayDeviceClient()
...
                     });
                 },
                 token);
+
             EdgeModule filter = deployment.Modules[LoadGenModuleName];
             await filter.WaitForEventsReceivedAsync(deployment.StartTime, token);
             await this.ValidateModule(this.runtime.DeviceId, LoadGenModuleName, TestModelId);
         }
 
+        EdgeConfigBuilder AddBrokerToDeployment(EdgeConfigBuilder builder)
```
vadim-kovalyov, line 119: this is interesting approach. Is there a way we can make it part of the e2e framework? Could be useful for everyone. Don't have to be done in this pr, but maybe the next one? //cc @damonbarry

vadim-kovalyov (approved):

test/Microsoft.Azure.Devices.Edge.Test/PlugAndPlay.cs
dylanbronson: Probably the most clear way to do this (that I can think of) would be to add a version of this method to the runtime.DeployConfigurationAsync method. Then we can add an attribute with test data + parameter to every test (similar to this test) to give each test the option of turning this functionality on or not...
This is similar to how we deal with amqp or mqtt today

Another option is we could have an extra stage in the pipeline that runs all the tests again, but this time with a different deployment that has broker on. Requires no test code change, but I'm not sure how the test output would look

PR #4051, opened 2020-12-04
Align base images for mqtt perf tests with production images.

vadim-kovalyov:
~~Since we use `busybox` as a base image for generic mqtt broker (used in perf tests), we need to use `alpine` and `armv7-unknown-linux-musleabihf` for building the image. If we try to build for `armv7-unknown-linux-gnueabihf`, it will fail to start on `busybox` since glibc is missing.~~

We need to align base images for mqtt perf tests with what we use in production.

### DIFF
--- a/mqtt/docker/linux/amd64/Dockerfile
+++ b/mqtt/docker/linux/amd64/Dockerfile
@@ -1,4 +1,5 @@
-FROM busybox
+ARG base_tag=3.1.4-alpine3.11
+FROM mcr.microsoft.com/dotnet/core/aspnet:${base_tag}
 
 ADD ./x86_64-unknown-linux-musl/release/mqttd /usr/local/bin/mqttd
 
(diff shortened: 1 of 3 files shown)


and-rewsmith:
I don't think this will work because we don't have a target for alpine/arm in our build script.

mqtt/docker/linux/arm32v7/Dockerfile
```diff
@@ -1,16 +1,13 @@
-ARG base_tag=3.1.10-bionic-arm32v7
-FROM mcr.microsoft.com/dotnet/core/aspnet:${base_tag}
+# Use the same base image as prod edgehub images
+ARG base_tag=1.0.6.4-linux-arm64v8
```
and-rewsmith, line 2: typo: should be arm32v7

mqtt/docker/linux/amd64/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM busybox
+# Use the same base image as prod edgehub images
+ARG base_tag=3.1.4-alpine3.11
+FROM mcr.microsoft.com/dotnet/core/aspnet:${base_tag}
```
and-rewsmith, line 3: If we also use hub base here then the test image approach is the same across platforms and the environments will always be the same. The tradeoff is an increased image size. I don't have a strong opinion here so it is your choice.

and-rewsmith (approved):
I queued another build images pipeline run for the most recent change.

PR #4064, opened 2020-12-08
PlugAndPlay E2E test cleanup

dylanbronson:

Previously, we needed to use a preview version of IoTHub that supported PnP. Since then, they've gone GA, and we can use our normal IoTHub and normal EventHub endpoint. We can also get rid of the HTTP request we manually made and use the ServiceSDK's GetTwinAsync call that now contains the ModelID.


builds/e2e/templates/e2e-setup.yaml
```diff
@@ -12,10 +12,8 @@ steps:
       TestContainerRegistryPassword,
       TestDpsGroupKeySymmetric,
       TestEventHubCompatibleEndpoint,
-      TestPreviewEventHubCompatibleEndpoint,
```
yophilav, line 15: Are these secrets removed from KV? (Please do if you haven't done it already)

yophilav (approved):
LGTM. Please make sure to 
1. Run the test to make sure it works
2. Merge to other branch {release/1.0.10, release/1.2}

PR #4070, opened 2020-12-08
Cherry-pick PlugAndPlay E2E test cleanup (#4064)

dylanbronson:
Cherry-pick from main branch to release/1.2

Previously, we needed to use a preview version of IoTHub that supported PnP. Since then, they've gone GA, and we can use our normal IoTHub and normal EventHub endpoint. We can also get rid of the HTTP request we manually made and use the ServiceSDK's GetTwinAsync call that now contains the ModelID.


yophilav (approved):
:shipit:

PR #4071, opened 2020-12-08
Removing preview iothub and eventhub endpoint from PnP tests for 1.0.10

dylanbronson:
Cherry-picking part of this commit from main branch to 1.0.10: https://github.com/Azure/iotedge/pull/4064

Previously, we needed to use a preview version of IoTHub that supported PnP. Since then, they've gone GA, and we can use our normal IoTHub and normal EventHub endpoint.


yophilav (approved):
:shipit:

PR #4075, opened 2020-12-09
Generic Mqtt Test Module: Fill in skeleton which sends messages back and forth 

and-rewsmith:
Filled in the skeleton added by the prior PR:
https://github.com/Azure/iotedge/pull/4031

I believe there is only one significant change to the skeleton. The previous MessageHandler interface led to a lot of code duplication so I made a wrapping struct with the shared logic (MessageChannel).


test/modules/GenericMqttTester/docker/amd64/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+COPY $EXE_DIR/generic-mqtt-tester ./
+
+RUN apt-get update && apt-get install -y libssl-dev
+
+# Add an unprivileged user account for running the module
+RUN useradd -ms /bin/bash moduleuser
+USER moduleuser
+
+CMD echo "$(date --utc +"[%Y-%m-%d %H:%M:%S %:z]"): Starting Module" && \
+/app/generic-mqtt-tester
```
and-rewsmith, line 16: Adding simple dockerfile for now, will change and add others when I make a PR for the build pipeline.

mqtt/mqtt-util/src/client_io.rs
```diff
@@ -147,12 +147,9 @@ impl ClientIoSource {
                 })?;
             }
 
-            info!("Ssl connector");
             let config = SslConnector::builder(SslMethod::tls())
                 .map(|mut builder| {
-                    info!("Ssl connector built for tls");
                     if let Some(trust_bundle) = server_root_certificate {
-                        info!("Ssl connector detected trust bundle {:?}", trust_bundle);
```
and-rewsmith, line 155: Accidentally added these debug logs in one of my prior PRs

test/modules/GenericMqttTester/src/main.rs
```diff
@@ -12,10 +20,44 @@ async fn main() -> Result<()> {
 
     let settings = Settings::new()?;
 
-    todo!()
+    let tester = MessageTester::new(settings).await?;
+    let tester_shutdown = tester.shutdown_handle();
+
+    // TODO: We should be able to change this pattern to spawn only shutdown future (which blocks), but this doesn't work, Need to figure out why.
```
and-rewsmith, line 26: There is a significant problem with the below two futures. Ideally we should spawn the blocking shutdown future and not spawn the test future. This doesn't work and the client does not get polled.

mqtt/mqtt-broker-tests-util/src/client.rs
```diff
@@ -299,9 +299,9 @@ where
     }
 }
 
-pub fn create_client_from_module_env(address: String) -> Result<Client<ClientIoSource>, VarError> {
```
vadim-kovalyov, line 302: I think it was fine with `&str`.  `TcpConnection::new()` needs `impl Into<String>` anyways

test/modules/GenericMqttTester/src/message_channel.rs
```diff
@@ -0,0 +1,160 @@
...
+    shutdown_handle: MessageHandlerShutdownHandle,
+    shutdown_recv: Receiver<()>,
+    message_handler: Box<H>,
+}
+
+impl<H> MessageChannel<H>
+where
+    H: MessageHandler + ?Sized + Send,
+{
+    pub fn new(message_handler: Box<H>) -> Self {
```
vadim-kovalyov, line 109: nit: box it yourself
```suggestion
    pub fn new(message_handler: H) -> Self {
```

test/modules/GenericMqttTester/src/message_channel.rs
```diff
@@ -0,0 +1,160 @@
...
+
+use mqtt3::{
+    proto::{Publication, QoS},
+    PublishHandle, ReceivedPublication,
+};
+
+use crate::{MessageTesterError, BACKWARDS_TOPIC};
+
+#[derive(Debug, Clone)]
+pub struct MessageHandlerShutdownHandle(Sender<()>);
```
vadim-kovalyov, line 18: nit: just `ShutdownHandle`?

test/modules/GenericMqttTester/src/message_channel.rs
and-rewsmith: We need to match on settings and return a trait object, which I believe needs to be a box. Therefore I think it is cleaner to pass it directly into this function as Box<> rather than unbox and rebox. What do you think?

test/modules/GenericMqttTester/src/message_channel.rs
and-rewsmith: I was trying to differentiate between the MessageTesterShuttdownHandle and this one. Now that you mention it, I feel like it should be named MessageChannelShutdownHandle.

mqtt/mqtt-broker-tests-util/src/client.rs
and-rewsmith: I am guessing you meant it was fine with ```String```? Clippy complained that it wasn't consumed, but maybe that was because I had unecessary clones. Going to check again.

and-rewsmith: Because I am logging it as the last usage, which doesn't consume, I think we need &str here.

test/modules/GenericMqttTester/src/tester.rs
```diff
@@ -29,42 +70,211 @@ pub struct MessageTester {
...
+                break;
+            }
+            Either::Right((publish, _)) => {
+                publish.map_err(MessageTesterError::Publish)?;
+            }
+        };
+
+        time::delay_for(Duration::from_secs(1)).await;
+
+        seq_num += 1;
```
vadim-kovalyov, line 226: what are the termination condition of the loop? just shutdown signal? I'm just curios if we should worry about overflowing seq_num.

vadim-kovalyov (approved):

test/modules/GenericMqttTester/src/tester.rs
and-rewsmith: since it runs every second it will be 130+ years

PR #4081, opened 2020-12-11
Refactor EdgeHub config parsing to be version dependent

richma-ms:
This change refactors the twin JSON deserialization logic in EdgeHub so that we have completely separate implementation for each version.  By doing this we untangle the schema validation of the individual versions.  And in the cases of forward compatible changes, this will allow older runtimes to safely ignore unknown JSON properties from newer schema formats.

Note that in the case of changes that backwards compatible but not forward compatible, there's nothing we can do to prevent the older Edge runtimes from running into JSON deserialization errors.

e.g.
- Edge 1.0.9 will not be able to be able to ignore routes with priority/TTL as defined by 1.0.10/1.1.x, because it's a format change of the existing routes property
- Edge 1.1 will be able to ignore MQTT broker changes introduced by 1.2, as everything in 1.2 is purely additive


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubConfigParser.cs
```diff
@@ -26,32 +29,68 @@ public EdgeHubConfigParser(RouteFactory routeFactory, BrokerPropertiesValidator
...
+                else if (version == Core.Constants.SchemaVersion_1_1)
+                {
+                    var desiredProperties = JsonConvert.DeserializeObject<EdgeHubDesiredProperties_1_1>(twinJson);
+                    edgeHubConfig = this.GetEdgeHubConfig(desiredProperties);
+                }
+                else if (version == Core.Constants.SchemaVersion_1_2)
+                {
+                    var desiredProperties = JsonConvert.DeserializeObject<EdgeHubDesiredProperties_1_2>(twinJson);
+                    edgeHubConfig = this.GetEdgeHubConfig(desiredProperties);
+                }
```
richma-ms, line 64: @varunpuranik if we want to take this change release/1.1, we'd need to remove this block and everything that references EdgeHubDesiredProperties_1_2 in this change.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/EdgeHubConfigParser.cs
```diff
@@ -26,32 +29,76 @@ public EdgeHubConfigParser(RouteFactory routeFactory, BrokerPropertiesValidator
...
+            else
+            {
+                throw new InvalidSchemaVersionException("EdgeHub config missing SchemaVersion");
+            }
+
+            return edgeHubConfig;
         }
 
-        static ReadOnlyDictionary<string, RouteConfig> ParseRoutes(EdgeHubDesiredProperties desiredProperties, RouteFactory routeFactory)
+        public EdgeHubConfig GetEdgeHubConfig(EdgeHubDesiredProperties_1_0 desiredProperties)
```
vadim-kovalyov, line 86: do these need to be public?

vadim-kovalyov (approved):
LGTM!

PR #4084, opened 2020-12-11
EdgeHub awaits for Twin in non-MQTT broker scenario

yophilav:
TL;DR
Currently the protocol head could start up before the route configuration is setup on edgeHub. This causes the messages which are sent before the route is config to be permanently lost.

Background:
Historically, the edgeHub tries to download its twin to get the routings before it starts up the protocol heads, then it was waiting till the twin arrived, then it had the routings, then it started the protocol heads. With the new broker and nested scenario, the problem is that it needs the protocol heads up to get the twin. The twin comes through the MQTT broker, and one of the protocol heads handle the messages from the broker. The problem arises because protocol heads are up, but no twin available yet.

Fix:
If the upstream is not via MQTT broker, then await the twin task; otherwise, let it runs.

Todo:
This code change does not handle the scenario:
IoTHub <--MQTT/BRIDGE-- EdgeDevice <--AMQP-- DeviceA

Since we are not waiting on the configUpdate (that would be a dead lock, because configUpdate() needs the new mqtt/broker to be started), the code goes ahead and starts every protocol head i.e. AMQP listener for connecting clients. This can cause an AMQP message to be sent (into the void) before a route is setup on the DeviceA.


vipeller (dismissed):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/ExperimentalFeatures.cs
```diff
@@ -28,6 +28,16 @@ public static ExperimentalFeatures Create(IConfiguration experimentalFeaturesCon
             return experimentalFeatures;
         }
 
+        public static bool IsViaBrokerUpstream(ExperimentalFeatures experimentalFeatures, bool hasGatewayHostname)
```
varunpuranik, line 31: Nit: this doesn't belong in ExperimentalFeatures. Pls move it back to DependencyManager.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -97,10 +97,20 @@ static async Task<int> MainAsync(IConfigurationRoot configuration)
...
+
+            _ = configDownloadTask.ContinueWith(
+                                            _ => configUpdaterStartupFailed.SetResult(false),
+                                            TaskContinuationOptions.OnlyOnFaulted);
+
+            if (!ExperimentalFeatures.IsViaBrokerUpstream(
+                    experimentalFeatures,
+                    string.IsNullOrEmpty(configuration.GetValue<string>(Constants.ConfigKey.GatewayHostname))))
+            {
+                await configDownloadTask;
```
varunpuranik, line 112: I am confused, why can't we await the configUpdater.Init here?

vipeller: The problem is that when edgeHub uses the broker to talk upstream (e.g. to pull the twins), then it needs the protocol heads running.
On the other hand, if protocol heads start up, then clients can start connecting before the configUpdater finishes pulling the config (e.g. the route settings).
This was a bug before (clients started to send messages, but routing was not set), that is why it was waiting before for configUpdater, and then it started the protocol heads.
So now that we want to actieve:
- for 'old' protocol head it should wait to avoid sending messages to early
- for 'new' protocol head, it should spin up before the configUpdater gives up, so it can pull the twin.

The question is that for the 'new' protocol head it isn't a problem that clients connect too early - but it is not, because in this case the client facing interface is the mqtt broker and it will deny client connections until edgehub reports that it is ready. So at the end the system will work for internal communication (e.g. to pull the twin), but clients cannot connect before the twin is pulled for configuration

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/ExperimentalFeatures.cs
vipeller: it was moved here because there was a code duplication, so somewhere a common class is needed to be shared (this is called from dependency manager and from program.cs. In program.cs it is needed to decide if configUpdater should be awaited or not - if the communication is ViaBroker, then it should no wait.
Although the duplication is not big (so would not be a problem to copy), because it depends on config settings, odds are big that with the change of the config (e.g. it is not experimental feature anymore), one place will be fixed but not the other, seemed safer to merge. And because almost all components of that condition touches experimentalFetaures. and this was a static class already, it seemed a good candidate to move the condition here.
I think it makes sense the question that "is mqtt broker enabled' is very similar to the question that 'is the upstream communication goes through the new broker"

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
varunpuranik: But what about the other protocol heads - AMQP and HTTP?

vipeller (approved):

PR #4085, opened 2020-12-11
Update base images for Linux ARM & Windows AMD

yophilav:
1. Update ARM base image usage to version 1.0.5.6 (3.1.10-bionic-arm*)
2. Update Windows AMD64 base image to 3.1.10-nanoserver-1809


ancaantochi (approved):
why using version 1.0.5.6 instead of 1.0.6?

PR #4086, opened 2020-12-11
Locking down broker environment variables to only documented ones

dylanbronson:
Currently, the user is free to change whatever environment variables they want from the edgeHub container environment variables. This PR scopes it down to 4 public ones: 
mqttBroker__max_queued_messages
mqttBroker__max_queued_bytes
mqttBroker__max_inflight_messages
mqttBroker__when_full

These are now the only broker variables that a user can set for the broker from the edgeHub container. The others are taken from the default config file we have hard coded into the broker.
We use the config::Source trait to accomplish the env var mapping


vadim-kovalyov (approved):

mqtt/mqtt-edgehub/src/settings.rs
```diff
@@ -21,11 +23,43 @@ lazy_static! {
         s.merge(File::from_str(DEFAULTS, FileFormat::Json)).expect(
             "Unable to load default broker config. Check default.json has invalid json format.",
         );
-        s.try_into()
-            .expect("Unable to load default broker config. Check default.json to match BrokerConfig structure.")
+        let settings = s.try_into()
+            .expect("Unable to load default broker config. Check default.json to match BrokerConfig structure.");
+        debug!("broker config settings: {:?}", settings);
```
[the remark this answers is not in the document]
dylanbronson: (I moved it but same idea in new version) I think this is helpful to have in debug logs to see what settings broker starts with

mqtt/mqttd/src/app/edgehub.rs
```diff
@@ -51,7 +51,9 @@ impl Bootstrap for EdgeHubBootstrap {
 
     fn load_config<P: AsRef<Path>>(&self, path: P) -> Result<Self::Settings> {
         info!("loading settings from a file {}", path.as_ref().display());
-        Ok(Self::Settings::from_file(path)?)
+        let settings = Self::Settings::from_file(path);
```
vadim-kovalyov, line 54:
```suggestion
        let settings = Self::Settings::from_file(path)?;
```

mqtt/mqttd/src/app/edgehub.rs
```diff
@@ -51,7 +51,9 @@ impl Bootstrap for EdgeHubBootstrap {
...
+        debug!("broker settings are: {:?}", settings);
+        Ok(settings?)
```
vadim-kovalyov, line 56:
```suggestion
        Ok(settings)
```

vadim-kovalyov (approved):

PR #4088, opened 2020-12-12
Make edgelet's certificate CA configurable

massand:
Previously, the edgelet workload logic had hardcoded cert_ids for Edge CA (similar in essence to former workload CA) and trust bundle. The binaries assumed that `iotedge init` would initialize these. Configuration is still required for the trust bundle and Edge CA. However, the Edge CA is now dynamically generated with a shorter default expiration (90 days). This provides flexibility in chaining it to an existing CA in the Certificate Service. 

Changes:
- Fixed workload identity/server certificate request logic to leverage Edge CA
- Added Edge CA auto-generation logic when it is expired or missing
- Added new configurations for the Edge (workload) CA certificate and trust bundle

Verified using the following Certificate Service TOML structure (note owner CA still needs to be pre-generated by `iotedge init`):
```
[cert_issuance]
aziot-edged-owner-ca = { method = "self_signed" }
aziot-edged-ca = { method = "local_ca" }

[cert_issuance.local_ca]
cert = "aziot-edged-owner-ca"
pk = "aziot-edged-owner-ca"

[preloaded_certs]
"aziot-edged-owner-ca" = "file:///path/to/owner_ca.pem"
"iotedge-trust-bundle" = "file:///path/to/owner_ca.pem"

```
..and the following edgelet config.yaml additions:
```
edge_ca_cert: "aziot-edged-ca"
edge_ca_key: "aziot-edged-ca"
trust_bundle_cert: "aziot-trust-bundle"
```


daprilik (commented):
Is there a reason you've changed all the http route handlers (in edgelet-http-workload) to accept a `key_connector` instead of a `key_client` directly? From what I can tell, you seem to end up recreating a new `Arc<key client>` in many of the handlers, which seems a bit silly, since the point of an `Arc` is to share a single instance instead of constantly remaking it.

massand:
> Is there a reason you've changed all the http route handlers (in edgelet-http-workload) to accept a `key_connector` instead of a `key_client` directly? From what I can tell, you seem to end up recreating a new `Arc<key client>` in many of the handlers, which seems a bit silly, since the point of an `Arc` is to share a single instance instead of constantly remaking it.

Good catch. I switched because the `key_connector` is used for firing up `key_client` and `key_openssl_engine`, depending on the handler. But, I could pass a simple reference up the call stack.

edgelet/contrib/config/linux/config.yaml
```diff
@@ -233,3 +233,18 @@ moby_runtime:
...
+###############################################################################
+# Edge CA Settings 
+###############################################################################
+#
+# Optional. This entry should only be specified to override the certificate
+# and keypair name of the CA certificate created by IoT Edge to issue identity
+# and server certificates for Edge modules. The name corresponds to the 
+# cert-id in the Certificate Service and key-id in the Key Service that are
+# pre-configured manually or by invoking the iotedge init command.
+###############################################################################
```
gordonwang0, line 246: Would be helpful to list what the default names are if these options are not provided (same in debian/config.yaml below).

gordonwang0 (commented):
This will probably affect the E2E tests, since I hardcoded the names "iotedge-trust-bundle" and "aziot-edged-device-ca" in the test framework.

edgelet/contrib/config/linux/config.yaml
massand: Sounds good. I will update the default cert names in the tests once they go in.

gordonwang0: Got it. I'll approve then.

gordonwang0 (approved):

gordonwang0 (dismissed):

edgelet/iotedge/src/init/import/mod.rs
```diff
@@ -20,6 +20,7 @@ use std::path::Path;
 
 use config::Config;
 
+use edgelet_core::{AZIOT_EDGED_CA_ALIAS, TRUST_BUNDLE_ALIAS};
```
arsing, line 23: To be consistent, `TRUST_BUNDLE_ALIAS` should be `aziot-edged-trust-bundle` instead of `iotedge-trust-bundle` ?

Also, should `AZIOT_EDGED_CA_ALIAS` be `aziot-edged-ca` instead of `aziot-edged-device-ca` ?

edgelet/edgelet-core/src/settings.rs
```diff
@@ -191,6 +194,9 @@ pub struct Settings<T> {
     #[serde(default, skip_serializing)]
     #[cfg_attr(not(debug_assertions), serde(skip_deserializing))]
     pub endpoints: Endpoints,
+    pub edge_ca_cert: Option<String>,
+    pub edge_ca_key: Option<String>,
+    pub trust_bundle_cert: Option<String>,
```
arsing, line 199: Move these above the `endpoints` field to be consistent with the other services' configs.

edgelet/edgelet-http-workload/src/server/cert/mod.rs
```diff
@@ -30,6 +30,24 @@ mod server;
...
 pub use self::server::ServerCertHandler;
 
+// 5 mins
+const AZIOT_EDGE_CA_CERT_MIN_DURATION_SECS: i64 = 5 * 60;
+
+// 180 days
+const AZIOT_EDGE_CA_CERT_MAX_DURATION_SECS: u64 = 180 * 24 * 3600;
+
+// Workload CA alias
+const IOTEDGED_CA_ALIAS: &str = "iotedged-workload-ca";
```
arsing, line 40: Why does this exist instead of using `settings.edge_ca_cert` ?

arsing (changes requested):

edgelet/edgelet-http-workload/src/server/cert/mod.rs
massand: Oops

edgelet/edgelet-http-workload/src/server/cert/mod.rs
```diff
@@ -80,69 +98,125 @@ fn refresh_cert(
...
+> {
+    let (workload_ca_private_key, workload_ca_public_key) = {
+        let workload_ca_key_pair_handle =
+            std::ffi::CString::new(ca_key_pair_handle.0.clone()).unwrap();
+        let workload_ca_public_key = key_engine
+            .load_public_key(&workload_ca_key_pair_handle)
+            .unwrap();
+        let workload_ca_private_key = key_engine
+            .load_private_key(&workload_ca_key_pair_handle)
+            .unwrap();
```
daprilik, line 208: Unless they provably won't panic, these `.unwrap()`s should probably be replaced with some proper error handling, right?

massand: Fixed

edgelet/edgelet-http-workload/src/server/cert/mod.rs
massand: Fixed

edgelet/edgelet-core/src/settings.rs
massand: Done

edgelet/iotedge/src/init/import/mod.rs
massand: Yes - thanks for catching that inconsistency

arsing (dismissed):

massand:
@arsing I made a quick fix to the UT after your approval. I have not made any functional changes. Please review. Thanks.

arsing (approved):

gordonwang0 (approved):

PR #4096, opened 2020-12-15
Build Pipeline Optimization: Move only the rust executable into build context

and-rewsmith:
I thought I made this optimization a while ago when I did it for the mqtt image build, but I forgot. This change will copy only the executable into the build context instead of the whole target folder. This reduces the build context size and will trim some time off the build pipeline.


and-rewsmith:
release build:
https://msazure.visualstudio.com/One/_build/results?buildId=37692383&view=results

regular build:
https://msazure.visualstudio.com/One/_build/results?buildId=37692347&view=results

vadim-kovalyov (approved):
Thanks!

PR #4098, opened 2020-12-15
if no initial twin with base configuration, stop edgeHub

vipeller:
On initial startup (or when no persistent store is set up), there is no previous state for some configuration values needed for EdgeHub to work correctly. In a nested scenario (or even not nested case), it can happen that edgeHub starts up disconnected and it cannot pull its initial configuration. In this case edgeHub will stop (and gets restarted for a next try)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/upstream/BrokeredCloudConnectionProvider.cs
```diff
@@ -27,10 +27,33 @@ public Task<Try<ICloudConnection>> Connect(IClientCredentials clientCredentials,
             return this.Connect(clientCredentials.Identity, connectionStatusChangedHandler);
         }
 
-        public Task<Try<ICloudConnection>> Connect(IIdentity identity, Action<string, CloudConnectionStatus> connectionStatusChangedHandler)
+        public async Task<Try<ICloudConnection>> Connect(IIdentity identity, Action<string, CloudConnectionStatus> connectionStatusChangedHandler)
```
ancaantochi, line 30: Are these changes related to the config missing?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/upstream/BrokeredCloudConnectionProvider.cs
```diff
@@ -27,10 +27,33 @@ public Task<Try<ICloudConnection>> Connect(IClientCredentials clientCredentials,
...
+            {
+                return new Try<ICloudConnection>(new Exception("Bridge is not connected upstream"));
+            }
+
             var cloudProxy = new BrokeredCloudProxy(identity, this.cloudProxyDispatcher, connectionStatusChangedHandler);
-            return Task.FromResult(new Try<ICloudConnection>(new BrokeredCloudConnection(cloudProxy)));
+            return new Try<ICloudConnection>(new BrokeredCloudConnection(cloudProxy));
+        }
+
+        async Task<bool> IsConnected()
```
ancaantochi, line 41: can an event be used to the dispacher that gets triggered when it connect? or if it only has to wait for a short time and then throw you can use the RetryStrategy from utils

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/config/ConfigUpdater.cs
```diff
@@ -63,6 +63,16 @@ public async Task Init(IConfigSource configProvider)
                     {
                         Events.GettingConfig();
                         await pullTask;
+
+                        if (!this.currentConfig.HasValue)
```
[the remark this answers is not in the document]
vipeller: changed

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/upstream/BrokeredCloudConnectionProvider.cs
vipeller: The problem it solves came out testing the missing twin problem. The original problem was that when a new child edge device is starting up (so it has no stored twins) and the parent is not ready (so it cannot acquire the twin), then this problems happened (this PR fixes that problem). A secondary problem was that sometimes during startup it happens that edgeHub starts executing upstream operation when the infrastructure is not ready yet. It polluted the logs with error messages quite often, so adding this extra check for the connection status mitigates this.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/upstream/BrokeredCloudConnectionProvider.cs
vipeller: Changed to retry strategy. Using events seems more complex for me - normally it should be connected, so normally the tests pass (and the method returns true). As described above, this trying loop just mitigates the noise mainly at startup, when it can happen that edgehub-core has not received the connection event from the broker (from upstream bridge)

ancaantochi (approved):

PR #4105, opened 2020-12-16
cherry pick sha256 based support for thumbprint authentication

vipeller:


varunpuranik (changes requested):
I will confirm that we want to take this for 1.0.10 before we merge it.

varunpuranik (approved):

PR #4106, opened 2020-12-16
Add iotedge-init-import command to migrate pre-1.2 config to 1.2+ config.

arsing:
This is a subcommand of the `iotedge init` subcommand, however the latter has
not yet been implemented and always fails.

Other changes:

- Fix compilation warnings, missing lint preamble and missing copyright headers
  in support-bundle crate.

- Remove some unused error kinds from edgelet-core related to parsing
  the manual provisioning connection string, since that is no longer done by
  edged.

---

TODO:

- Update submodule reference once https://github.com/Azure/iot-identity-service/pull/85 is merged.

- Test.


edgelet/edgelet-core/src/module.rs
```diff
@@ -139,15 +139,15 @@ impl ModuleRuntimeState {
 
 #[derive(serde_derive::Deserialize, Debug, serde_derive::Serialize)]
 pub struct ModuleSpec<T> {
-    name: String,
+    pub name: String,
```
daprilik, line 142: (general comment, not necessarily part of this PR)

If we are making all these struct fields `pub` (IMO, as they should have been originally), then we aught to remove the getter/setter interface entirely. This shouldn't be difficult, though it will be a bit tedious... I guess we can do it as part of the Tokio 0.1 to 0.3/1.0 rewrite effort?

edgelet/support-bundle/src/support_bundle.rs
```diff
@@ -1,5 +1,16 @@
 // Copyright (c) Microsoft. All rights reserved.
 
+// TODO:
```
daprilik, line 3: ehh, I feel like it's probably fine lumping this TODO in with the rest of these changes. You know as well as I do that the moment you put a TODO like this into the codebase, unless we actually make a work-item for it, it'll just linger for a long while...

edgelet/iotedge/src/main.rs
```diff
@@ -2,11 +2,11 @@
 
 #![deny(rust_2018_idioms, warnings)]
 #![deny(clippy::all, clippy::pedantic)]
-#![allow(clippy::similar_names)]
+#![allow(clippy::let_unit_value, clippy::similar_names)]
+
+mod init;
```
daprilik, line 7: this should be in lib.rs

edgelet/iotedge/src/init/import/old_config/mod.rs
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+mod agent;
+pub(crate) use agent::*;
```
daprilik, line 4: (unrelated to the PR at hand)

Why use `pub(crate)` instead of just `pub` when working with `bin` crates? IMO it just makes things more verbose for no great reason, since there aren't any upstream crates to "hide" these details from.

edgelet/iotedge/src/init/import/old_config/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+#[derive(Debug, serde_derive::Deserialize)]
+pub(crate) struct DeviceCertificate {
+    pub(crate) device_ca_cert: String,
+    pub(crate) device_ca_pk: String,
+    pub(crate) trusted_ca_certs: String,
+}
+
+fn default_auto_generated_ca_lifetime_days() -> u16 {
```
daprilik, line 66: `const DEFAULT_AUTO_GENERATED_CA_LIFETIME_DAYS: u16 = 90`?

or at the very least, make this a `const fn`

edgelet/iotedge/src/init/import/old_config/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+mod provisioning;
+pub(crate) use provisioning::*;
+
+use std::path::PathBuf;
+
+use url::Url;
+
+#[derive(Debug, serde_derive::Deserialize)]
+pub(crate) struct Config {
+    pub(crate) provisioning: Provisioning,
```
daprilik, line 18: if the struct is `pub(crate)`, marking individual fields as `pub(crate)` as opposed to just `pub` isn't necessary.

ditto for every other instance of this.

daprilik (commented):
Core migrate logic seems fine. Other than that, mostly just style questions/nits

edgelet/edgelet-core/src/module.rs
arsing: Yes. Removing the getters will mean a lot more ancillary changes. Otherwise I would've removed them; I'm not a fan of trivial getters either.

massand: I am fine with both for now. I added trivial getters for my PR because I knew they will be rolled into structs eventually.

edgelet/iotedge/src/init/import/old_config/mod.rs
arsing: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=048936fced270e6262c38750d9fcc86b vs https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c9416b3a5cf256e6f40c254d44d805a0

edgelet/iotedge/src/init/import/old_config/mod.rs
arsing: `#[serde(default)]` needs a path that can be invoked as a function.

edgelet/iotedge/src/init/import/old_config/mod.rs
arsing: I think it's better to be precise. The field is only written within the crate, so the field should be `pub(crate)`. Otherwise it's a slippery slope - if a `mod` is `pub(crate)` should all the items in the `mod` be made `pub` ?

edgelet/iotedge/src/init/import/old_config/mod.rs
arsing: Done.

edgelet/iotedge/src/main.rs
arsing: Done.

edgelet/support-bundle/src/support_bundle.rs
massand: Would the replacement for this be tokio 0.1 compatible? If not, it isn't worth the switch now maybe. 

---
In reply to: [544641313](https://github.com/Azure/iotedge/pull/4106#discussion_r544641313) [](ancestors = 544641313)

arsing: :effort:, but fine. Done.

edgelet/iotedge/src/init/import/old_config/mod.rs
daprilik: Ah, sure, that's actually pretty reasonable actually. Though I'm pretty sure you can work around that by marking top-level modules as `pub(crate)` and calling it a day, but sure.

edgelet/iotedge/src/init/import/old_config/mod.rs
daprilik: I mean... yeah, that's what I do 😶
But whatever, it's not a dealbreaker or anything.

arsing: Feels like a footgun. You're one refactor away from accidentally increasing your API surface when you didn't mean to.

massand (approved):
:shipit:

edgelet/iotedge/src/init/import/old_config/mod.rs
daprilik: I mean, when I write libraries, I tend to keep an eye on the `cargo doc` output, which makes those cases pretty obvious...

In any case, I wrote out some more thoughts on the matter, but after reading them over I feel like it kinda just boils down to personal preference, and idk if there's a clear best answer. My gut feeling is that `pub(crate)` should stay at the item level, not really the field level, but like I said, that's just a gut feeling.

edgelet/iotedge/src/init/import/mod.rs
```diff
@@ -0,0 +1,701 @@
...
+const AZIOT_EDGED_HOMEDIR_PATH: &str = "/var/lib/aziot/edged";
+
+/// The ID used for the device ID key (symmetric or X.509 private) and the device ID cert.
+const DEVICE_ID_ID: &str = "device-id";
+
+/// The ID used for the device CA X.509 private key and the device CA cert.
+const DEVICE_CA_ID: &str = "device-ca";
+
+/// The ID used for the trust bundle certs file.
+const TRUST_BUNDLE_ID: &str = "trust-bundle";
```
arsing, line 23: @massand Are these correct? I haven't been following your and Gordon's work but I thought I saw some mention of hard-coded IDs for these two?

edgelet/iotedge/src/init/import/mod.rs
```diff
@@ -0,0 +1,701 @@
...
+
+/// The default value of the edged connect and listen management URIs.
+const AZIOT_EDGED_MANAGEMENT_URI: &str = "unix:///var/lib/aziot/edged/aziot-edged.mgmt.sock";
+
+/// The default value of the edged connect and listen workload URIs.
+const AZIOT_EDGED_WORKLOAD_URI: &str = "unix:///var/lib/aziot/edged/aziot-edged.workload.sock";
```
arsing, line 29: This is wrong. Needs to be `/var/run`, but still be `/var/lib` for CentOS.

edgelet/iotedge/src/init/import/mod.rs
massand: This​ PR migrates the hardcoded certid/keyid for the CA just fine - this is correct. This should work fine for the E2E tests and migrating non-quickstart scenarios (not sure if it handles the quickstart certs - in which case we should generate the `device-ca` and `trust-bundle` certs in `iotedge init` in all cases. I can add more migrate logic to my PR once you merge your PR.

massand: Never mind the E2E tests - we don't migrate in E2E yet. The E2E tests set the same certificates manually without `iotedge init` today and can continue to do so.

massand: Actually, the latest names in my PR #4088  are as follows:
```
pub const AZIOT_EDGED_CA_ALIAS: &str = "aziot-edged-device-ca";
pub const AZIOT_TRUST_BUNDLE_ALIAS: &str = "aziot-trust-bundle";
```

Could you change it to match that?

massand: Or if that PR doesn't go in today, we need to match the current names in the codebase - 
aziot-edged-device-ca
iotedge-trust-bundle

arsing: Fixed. I made it use the existing constant in `edgelet_core`, and added another constant in `edgelet_core` for the trust bundle.

edgelet/iotedge/src/init/import/mod.rs
arsing: Fixed by copying over the existing endpoints. Rajeev confirmed that reusing the old endpoints should be fine.

massand (approved):
:shipit:

PR #4107, opened 2020-12-16
[CI] enable MqttConnectorTest in CI pipeline

dmolokanov:
`MqttMiniServer` had multiple places with loops that try to make a progress forever even when the client disconnected.
`AuthProtocolHead.CloseAsync` method calls `IWebHost.StopAsync` which blocks not only the execution of the current test but also other tests that were running in parallel with it.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
@@ -100,8 +100,13 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
             {
                 // FIXES a bug when the the process is stuck on WebHost.StopAsync() method call https://github.com/dotnet/extensions/issues/1363
                 // lets not wait forever but cancel after a timeout
-                var cts = new CancellationTokenSource(TimeSpan.FromSeconds(15));
-                this.CloseAsync(cts.Token).Wait();
+                var timeout = TimeSpan.FromSeconds(15);
+                var closingTask = this.CloseAsync(new CancellationTokenSource(timeout).Token);
+                var timeoutTask = Task.Delay(timeout);
+                if (Task.WhenAny(closingTask, timeoutTask).Result == timeoutTask)
+                {
+                    Events.ClosingTaskTimeoutOut(timeout);
+                }
             }
         }
 
@@ -140,7 +145,8 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
                 Closing,
                 Closed,
                 ClosedWhenNotRunning,
-                StartedWhenAlreadyRunning
+                StartedWhenAlreadyRunning,
+                ClosingTaskTimeoutOut
             }
 
             public static void Starting() => Log.LogInformation((int)EventIds.Starting, "Starting AUTH head");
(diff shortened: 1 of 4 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
```diff
@@ -100,8 +100,13 @@ public void Dispose()
             {
                 // FIXES a bug when the the process is stuck on WebHost.StopAsync() method call https://github.com/dotnet/extensions/issues/1363
                 // lets not wait forever but cancel after a timeout
-                var cts = new CancellationTokenSource(TimeSpan.FromSeconds(15));
-                this.CloseAsync(cts.Token).Wait();
+                var timeout = TimeSpan.FromSeconds(15);
+                var closingTask = this.CloseAsync(new CancellationTokenSource(timeout).Token);
```
vadim-kovalyov, line 104: if `CloseAsync` calls `IWebHost.StopAsync` and blocks everything, how does this change helps? or the problem was in blocking call `.Wait()`?

dmolokanov: Yes, the problem is with blocking `Wait`. On a machine, with a limited number of cores, it was quite easy to block all threads in a thread pool. That's how I reproduced the issue: run tests in parallel on a VM with 2 cores

dmolokanov: With this change, we request stop and if it hasn't finished within some interval (15 sec), stop waiting and continue. Hopefully, GC will clean up resources soon: when the test finished or when our container restarted in production

vadim-kovalyov: @dmolokanov, I wonder what would happen on a 1 core machine. `.Result` call on line 106 is also blocking, and isn't it possible that it can deadlock itself so even timer will not be able to wake up? I'm not sure about dotnet, but we recently had the same problem with @and-rewsmith and tokio blocking calls.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
```diff
@@ -95,13 +95,15 @@ public async Task CloseAsync(CancellationToken token)
         }
 
         public void Dispose()
+        {
+            DisposeAsync().ConfigureAwait(false).GetAwaiter().GetResult();
+        }
```
and-rewsmith, line 100: If the solution was to switch to IAsyncDisposable, and for this we only need to implement DisposeAsync, why do we need this block?

vadim-kovalyov (approved):
Nice!

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
vadim-kovalyov: It's a good point. It is suggested in the docs for a class to implement both, but we don't really need sync version of `Dispose`. We don't need it by two reasons: 
- aspnet DI will use IAsyncDisposable if available
- Tests must use `async using` to avoid deadlocks.
I wonder if by removing this we'll force consumers to use `await using` only - what would be a good thing.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/authentication/AuthAgentProtocolHead.cs
```diff
@@ -95,13 +95,15 @@ public async Task CloseAsync(CancellationToken token)
         }
 
         public void Dispose()
+        {
+            DisposeAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
and-rewsmith, line 99: How do we decide whether or not to wrap wrap in Task.Run? From what I looked up wrapping in Task.Run() is likely to block a threadpool thread but will not result in async deadlock. I don't really understand this though.

and-rewsmith: https://stackoverflow.com/questions/59018541/when-to-use-task-run-getawaiter-getresult-and-getawaiter-getresult

vadim-kovalyov (approved):

PR #4109, opened 2020-12-17
Remove metrics v0

ancaantochi:
Metrics V0 were already deprecated and now removed


doc/[Old]Metrics.md
```diff
@@ -1,5 +1,5 @@
 
-# **!!! IMPORTANT NOTE !!! The metrics described here are deprecated and will be removed in subsequent releases. They are superseded by [edge metrics V2](BuiltInMetrics.md).**
+# **!!! IMPORTANT NOTE !!! The metrics described here were removed in release 1.1.0. They are superseded by [edge metrics V2](BuiltInMetrics.md).**
```
Comment (varunpuranik) [inline doc/[Old]Metrics.md:2]: Should we even keep this page? Or just remove it?
@veyalla

varunpuranik (approved):

doc/[Old]Metrics.md
Comment (ancaantochi) [reply doc/[Old]Metrics.md:2]: will follow up with Venkat regarding this file

PR #4110, opened 2020-12-17
Fix policy test flakiness

vadim-kovalyov:
There was another issue in Policy tests that make them flaky. `LeafDevice.CreateAsync` appears to be not idempotent, so there is a race condition, when LeafDevice tries to connect before the policy is updated, and in case of failed authorization, it tries to delete and re-create the device. That screws up cached symmetric keys in EH. And since EH does not refresh the device scopes (and keys) more often that 2min, the test fails after several retries.

The solution is to not to use `LeafDevice` - not to remove the device in case of authorization error, but just retry again.


dmolokanov (approved):
LGTM

PR #4140, opened 2020-12-21
[8999293] modified protocol head start up order based on if MQTT broker is enabled

davilu:
modified to always start MQTT broker related protocol head along with get twin task. 
modified to start original hub core related protocol head after twin get pulled.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
```diff
@@ -122,21 +115,23 @@ static async Task<int> MainAsync(IConfigurationRoot configuration)
...
+            using (IProtocolHead edgeHubProtocolHead = await GetEdgeHubProtocolHeadAsync(logger, configuration, experimentalFeatures, container, hosting))
             using (var renewal = new CertificateRenewal(certificates, logger))
             {
                 try
                 {
-                    await protocolHead.StartAsync();
+                    await Task.WhenAll(mqttBrokerProtocolHead.StartAsync(), configDownloadTask);
                     await Task.WhenAny(cts.Token.WhenCanceled(), renewal.Token.WhenCanceled(), configUpdaterStartupFailed.Task);
+                    await edgeHubProtocolHead.StartAsync();
                 }
```
vipeller, line 127: Is the order right? The way I read the code is that it starts mqttbroker (or the empty protocol head) and the configDownload (this part is ok)
But then it waits for shutdown (or till the certificate expires or the config download failed) and then starts the rest of the protocol head.

vipeller (changes requested):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Service/Program.cs
davilu: Nice catch. updated.

vipeller (approved):

PR #4144, opened 2020-12-21
Fixed hanging CI tests for MqttBrokerConnector

vipeller:
Fixed two tests that was either flaky or got hung:
- WhenReconnectThenResubscribes: this was not waiting for the second action to complete and sometimes checked the result 'too early'. Added a 'milestone' to wait at, so it goes to the assert part when the broker signals that it executed a certain action
- OfflineSendGetSentAfterReconnect: it never broke on a local desktop or raspberry pi with 4 cores enabled, however it got hung sometimes in CI or raspberry with 2 core enabled. The test becomes stable even after a tiny wait time between dropping the connection and sending the message (which needs to be handled with connection off)

Note, that this can be a sign of a real bug (not test bug) that if a 'send' operation happens in a very narrow window when the connection got dropped (not even noticed by the mqtt library yet) and a message gets sent in that window, then the message will not be resent later by the library, so no ack will be delivered to the connector, and it will keep waiting.

there are multiple options to mitigate the problem:
- change the client library: that is a bigger piece of work, also we don't know what problems those libraries would have, and this one has been tested for a while.
- try to find the problem in the source of the current client library (if it is in there)
- live with the bug but add code in the connector to handle it: currently the code waits for the ack indefinitely, assuming that either an error or the ack will arrive. This code could have a timer and gives up waiting after a while.

The latest change is small, however this communication happens inside the same container and odds are slim that it gets disconnected, even slimmer than in that sensitive window a message gets sent. On the other hand creating an expiring task for every message takes resources. Because of that I would not make that chage.

### DIFF
--- a/edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
+++ b/edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
@@ -266,7 +266,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test
             Assert.Equal(2, broker.ConnectionCounter);
         }
 
-        [Fact(Skip = "Fails in CI pipeline. Temporarily disabling while we investigate what is wrong")]
+        [Fact]
         public async Task WhenReconnectsThenResubscribes()
         {
             using var broker = new MiniMqttServer();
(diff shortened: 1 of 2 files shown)


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
```diff
@@ -322,6 +336,8 @@ public async Task OfflineSendGetSentAfterReconnect()
                 broker.DropActiveClient();
             }
 
+            await Task.Delay(TimeSpan.FromMilliseconds(200)); // give time to settle with dropped connection
```
vadim-kovalyov, line 339: do we have a work item for a better fix?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
```diff
@@ -287,17 +287,31 @@ public async Task WhenReconnectsThenResubscribes()
...
-            Assert.True(await milestone.WaitAsync(TimeSpan.FromSeconds(5)));
+            Assert.True(await milestoneConnected.WaitAsync(TimeSpan.FromSeconds(5)));
+            Assert.True(await milestoneSubscribed.WaitAsync(TimeSpan.FromSeconds(5)));
 
-            var expected = consumers.SelectMany(c => c.Subscriptions).OrderBy(s => s);
-            Assert.Equal(expected, broker.Subscriptions.OrderBy(s => s));
+            var expected = consumers.SelectMany(c => c.Subscriptions).OrderBy(s => s).ToArray();
+            var actual = broker.Subscriptions.OrderBy(s => s).ToArray();
+
+            Assert.Equal(expected, actual);
```
vadim-kovalyov, line 311: I'm not sure if I follow the logic of this test. Specifically, where is the guarantee that you'll wait until client re-subscribes before doing assertion?

vadim-kovalyov (approved):
Looks good, but I still have few questions. Feel free to answer and merge if it's not an issue.

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
vipeller: the logic for waiting for re-subscribe is in the broker.OnSubscribe(). First it counts how many subscription it should wait for, then every subscription decreases the counter and at the end the semaphore releases, that is what a bit lower the code is waiting for.

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
vipeller: I am creating one

PR #4151, opened 2020-12-22
Fix changelog for 1.2-rc3 release

and-rewsmith:
We need to duplicate some items as they affect edge agent also.


and-rewsmith:
Some of the items I listed under edgehub features are things that existed in 1.0.10, but we didn't have for 1.2 public preview (pnp and sha256). Let me know if you think it is better if they are under bug fixes.

varunpuranik (approved):

Issue #3764, opened 2020-10-20
Edge running on AKS is not updating module images with new tag and version in IotHub


ms-amakumar:
Hey all,

We have an edge device running on AKS and our release pipeline on every deployment updates the image tag as well as module version. We are using automatic deployments. I see that the deployments are being applied on the devices, but the module image is actually not updating iot hub. 

I have been using Helm chart from below
https://edgek8s.blob.core.windows.net/staging

Screenshot for reference:
![image](https://user-images.githubusercontent.com/63478804/96665600-1a7e0800-130a-11eb-80fb-e81e149d3fd9.png)

Edge agent version: Version - 1.0.10-dev.35780923 (0d09dc260e49baed02df9021a0046610d081bc0a)
Edge agent image: azureiotedge/azureiotedge-agent:0.1.0-beta11
Iotedged image: azureiotedge/azureiotedge-iotedged:0.1.0-beta11

Also, Iotedged pod seems to be generating every log as error.
 
![image](https://user-images.githubusercontent.com/63478804/96665622-279af700-130a-11eb-9ea0-7614999459a0.png)

## Expected Behavior
Tell us what should happen
Module images should be correctly updated whenever a new deployment is made with different image tag.

## Current Behavior
Module images are still showing old values while the deployment is already applied to edge device.

## Steps to Reproduce
Provide a detailed set of steps to reproduce the bug.
1. Setup an edge device in AKS cluster
2. Create a deployment and apply it to the device.
3. Create another deployment with same priority and update the module tag to a higher version.
4. Validate image tag in Iot Hub.

### Output of `iotedge check`

<details>
<summary>Click here</summary>

```

<Paste here between the triple backticks>

```
</details>

### Device Information
* Host OS [e.g. Ubuntu 16.04, Ubuntu 18.04, Windows IoT Core]:  Ubuntu 18.04
* Architecture [e.g. amd64, arm32, arm64]: amd64
* Container OS [e.g. Linux containers, Windows containers]:  Linux

### Runtime Versions
* iotedged [run `iotedge version`]: azureiotedge/azureiotedge-iotedged:0.1.0-beta11
* Edge Agent [image tag (e.g. 1.0.0)]: azureiotedge/azureiotedge-agent:0.1.0-beta11
* Edge Hub [image tag (e.g. 1.0.0)]: 1.0.0
* Docker/Moby [run `docker version`]: Docker

[...]

PR #4152, opened 2020-12-23
[k8s] ensure agent shuts down on critical watch failures.

darobs:
Sometimes the initial List in a List/Watch API request will time out, and the exception is swallowed instead of being handled.
This is what happened in issues like #3764 .

The main program has a CancellationTokenSource, created to shut down the main process from SIGTERM events and other critical failures. This is now passed to the k8s watches, so the watch can cancel the CTS when a fatal error occurs.

The list of fatal errors:
- OnEvent: the event handler receives a fatal exception.
- OnClosed: An exception is thrown while attempting watch restart.
- OnError: Always triggers shutdown.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesEnvironmentOperator.cs
```diff
@@ -37,43 +38,67 @@ public void Stop()
...
         public void Dispose() => this.Stop();
 
-        void StartListPods() =>
+        void StartListPods(CancellationTokenSource shutdownCts) =>
             this.client.ListNamespacedPodWithHttpMessagesAsync(this.deviceNamespace, watch: true)
-                .ContinueWith(this.OnListPodsCompleted);
+                .ContinueWith(this.OnListPodsCompleted, shutdownCts);
 
-        async Task OnListPodsCompleted(Task<HttpOperationResponse<V1PodList>> task)
+        async Task OnListPodsCompleted(Task<HttpOperationResponse<V1PodList>> task, object shutdownCtsObject)
```
pmzara, line 45: Why the parameter shutdownCtsObject is an object and not CancellationTokenSource?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentOperator.cs
```diff
@@ -51,36 +51,23 @@ public void Stop()
...
-                        Events.EdgeDeploymentWatchFailed(ex);
-                        throw ex;
-                    }));
+                    onEvent: async (type, item) => await this.EdgeDeploymentOnEventHandlerAsync(type, item, shutdownCts),
+                    onClosed: () => this.RestartWatch(shutdownCts),
+                    onError: (ex) => this.HandleError(ex, shutdownCts)));
         }
 
-        internal async Task EdgeDeploymentOnEventHandlerAsync(WatchEventType type, EdgeDeploymentDefinition item)
+        internal async Task EdgeDeploymentOnEventHandlerAsync(WatchEventType type, EdgeDeploymentDefinition item, Option<CancellationTokenSource> shutdownCts)
```
pmzara, line 70: Why use Option<CancellationTokenSource> and not only CancellationTokenSource? In some places, you are using CancellationTokenSource and in others, you are using Option<CancellationTokenSource>. Any specific reason to do this?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesEnvironmentOperator.cs
darobs: The [ContinueWith](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.continuewith?view=net-5.0#System_Threading_Tasks_Task_ContinueWith_System_Action_System_Threading_Tasks_Task_System_Object__System_Object_) on line 43 function is defined as `Action<Task,Object>` I'm probably being overly cautious. about the parameter type.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/EdgeDeploymentOperator.cs
darobs: I'm using the Option type to manage the potentially null object passed into `OnListEdgeDeploymentsCompleted`.  Logically, it shouldn't ever be null, but I think this is once again me being overly cautious.

pmzara (approved):

PR #4153, opened 2020-12-23
`iotedge check` update

daprilik:
Updates `iotedge check` to work with the new underlying `aziot` architecture.

This PR builds on-top of the functionality introduced in the downstream `aziot` PR at https://github.com/Azure/iot-identity-service/pull/83

* * *

In a nutshell, many of the checks that used to be in `iotedge check` have been ported over into the new `aziot check` tool. Please see the associated downstream PR for more details. That being said, we don't want to make users invoke multiple `check` commands, so this PR also includes functionality whereby `iotedge` will shell-out to `aziot` when running `check` and `check-list`, transparently displaying `aziot`'s check results alongside `iotedge`-specific checks.

As with the downstream PR, this PR should be reviewed along two axes:

1. Refactoring work around the actual `iotedge check/check-list` subcommands
2. The actual checks themselves (i.e: confirming that all removed checks have indeed been migrated / aren't applicable anymore).

I would **really appreciate it** if the reviewers actually pulled these changes locally and tested them out, as subtle things such as check output formatting and/or checking check pass/failure cases aren't easy to spot by looking at the code alone.


edgelet/iotedge/src/check/checks/storage_mounted_from_host.rs
```diff
@@ -95,15 +95,8 @@ fn storage_mounted_from_host(
...
-        .unwrap_or(
-            // Hard-code the value here rather than using the tempfile crate. It needs to match .Net Core's implementation,
-            // and needs to be in the context of the container user instead of the host running `iotedge check`.
-            if cfg!(windows) {
-                r"C:\Windows\Temp"
-            } else {
-                "/tmp"
-            },
-        );
+        // Hard-code the value here rather than using the tempfile crate.
```
arsing, line 98: Keep the rest of the original comment. It applies to both Windows and Unix.

edgelet/iotedge/src/main.rs
```diff
@@ -64,6 +61,16 @@ fn run() -> Result<(), Error> {
                 .env("IOTEDGE_HOST")
                 .default_value(default_mgmt_uri),
         )
+        .arg(
```
arsing, line 64: Why does this need to be configurable? Just for local dev? If so, might be useful to make it `#[cfg(debug_assertions)]`

edgelet/iotedge/src/check/mod.rs
```diff
@@ -215,82 +200,89 @@ impl Check {
...
-        let result: Vec<&'static str> = Check::checks()
-            .iter()
-            .flat_map(|(_, section_checks)| section_checks)
-            .map(|check| check.id())
-            .collect();
+        // get all the aziot checks by shelling-out to aziot
+        {
+            let aziot_check_out = std::process::Command::new(aziot_bin)
+                .arg("check-list")
+                .arg("-j")
```
arsing, line 225: `check-list` ought to be changed to be consistent with `check` and use `--output json` instead of `--json`

edgelet/iotedge/src/check/mod.rs
```diff
@@ -334,190 +343,306 @@ impl Check {
...
-                            if self.verbose {
-                                for cause in err.iter_causes() {
-                                    write_lines(
-                                        stdout,
-                                        "        caused by: ",
-                                        "                   ",
-                                        cause.to_string().lines(),
-                                    )?;
+            match aziot_check.spawn() {
+                Err(err) => {
```
arsing, line 565: Put the `Err` arm after the `Ok` arm please, to be consistent with the rest of the codebase.

edgelet/iotedge/src/check/mod.rs
```diff
@@ -215,82 +200,89 @@ impl Check {
...
-            .collect();
+        // get all the aziot checks by shelling-out to aziot
+        {
+            let aziot_check_out = std::process::Command::new(aziot_bin)
+                .arg("check-list")
+                .arg("-j")
+                .output();
+
+            match aziot_check_out {
+                Err(_) => {
```
arsing, line 229: Put the `Err` arm after the `Ok` arm please, to be consistent with the rest of the codebase.

arsing (changes requested):

edgelet/iotedge/src/main.rs
daprilik: Yes, I've primarily been using it for local dev. I don't imagine any end users would have to manually set it.

Instead of customizing it via CLI flag, how about I read it from an `AZIOT_BIN` env var instead? In that case, should I hide it behind `#[cfg(debug_assertions)]`, or would it be fine to leave it active in release builds as well?

edgelet/iotedge/src/check/mod.rs
daprilik: Fair enough, though given that this isn't a flag that users will ever be expected to use directly, I don't think this bit of inconsistency is too big of a deal. Plus, it'll be a bit of a PITA to update `aziot check` now that the first PR has already been merged :wink:

edgelet/iotedge/src/main.rs
daprilik: Oh, better yet, I can just use [`option_env!`](https://doc.rust-lang.org/std/macro.option_env.html) instead. We already use it for `IOTEDGE_HOST`. No `#[cfg(debug_assertions)]` needed :smile: 

Sidenote, I didn't know `option_env!` was a thing - very cool!

edgelet/iotedge/src/check/mod.rs
arsing: Yeah, sorry about that.

daprilik: Sooooo I _should_ make the change then :smile:

arsing: Yes.

massand (dismissed):

massand:
:shipit:

massand:
Hold off on any submodule update before merge because I might have exposed a bug in the module runtime with the latest aziot version. I'll post a fix for it shortly. If you aren't updating the `aziot` submodule, you should be fine.

massand:
#4240 fixes the above issue. If you don't update the submodule, I can do so in my PR.

daprilik:
> Hold off on any submodule update before merge because I might have exposed a bug in the module runtime with the latest aziot version. I'll post a fix for it shortly. If you aren't updating the `aziot` submodule, you should be fine.

I still need to fix https://github.com/Azure/iotedge/pull/4153#discussion_r555611080, so I'll end up bumping the submodule regardless.

massand:
One note on testing these `iotedge check` changes locally (@micahl can provide guidance): `iotedge check` compares installed version with the latest versions [here](https://github.com/Azure/azure-iotedge/blame/master/latest-versions.json). We should coordinate on adding `aziot-identityd` and adding the version check in this PR perhaps?

arsing:
https://github.com/Azure/iot-identity-service/blob/main/aziot/src/internal/check/checks/mod.rs#L34-L35
________________________________

huguesBouvier (changes requested):
Should be tested in nested configuration. Check it out with me

edgelet/iotedge/src/init/import/mod.rs
```diff
@@ -268,7 +268,7 @@ fn execute_inner(
                     ),
             }) => (
                 aziot_identityd_config::Provisioning {
-                    dynamic_reprovisioning: *dynamic_reprovisioning,
+                    always_reprovision_on_startup: *dynamic_reprovisioning,
```
arsing, line 271: No. `always_reprovision_on_startup` and `dynamic_reprovisioning` are not the same thing.

`always_reprovision_on_startup` needs to be set to what it was set in the old config.

daprilik: @massand said this is the fix :cry:

arsing: I don't know why he said that, but it doesn't change what I said.

daprilik: Wait, so what should the fix be for the `ProvisioningType::Manual` cases?

daprilik: @arsing

daprilik: Also, on that note, it looks like I need to change the test-files as well.

I can't say I'm a fan of how these changes are getting bundled in with this unrelated PR, but if it's happening, I could use some guidance on what to do.

arsing: https://github.com/Azure/iotedge/blob/release/1.0.10/edgelet/edgelet-core/src/settings.rs#L426-L430

daprilik: Gotcha. So I'll just set them to true then. Cool.

daprilik: It is done

edgelet/iotedge/src/init/import/mod.rs
```diff
@@ -253,7 +253,7 @@ fn execute_inner(
     ) = {
         let old_config::Provisioning {
             provisioning,
-            dynamic_reprovisioning,
+            dynamic_reprovisioning: _,
```
arsing, line 256: Add a comment on this line along the lines of `// TODO: Migrate this to edged config when support for dynamic reprovisioning is reinstated in edged.`

@massand Do you have a work item for it already?

arsing (dismissed):

edgelet/iotedge/src/init/import/mod.rs
daprilik: added the comment

huguesBouvier (changes requested):
Best careful, the run on the pipeline is not compiling:
 use edgelet_docker::{Settings, UPSTREAM_PARENT_KEYWORD};
  |                                ^^^^^^^^^^^^^^^^^^^^^^^ no `UPSTREAM_PARENT_KEYWORD` in the root

daprilik:
> Best careful, the run on the pipeline is not compiling:
> use edgelet_docker::{Settings, UPSTREAM_PARENT_KEYWORD};
> | ^^^^^^^^^^^^^^^^^^^^^^^ no `UPSTREAM_PARENT_KEYWORD` in the root

Yes indeed, I realized that and force pushed a fix.

huguesBouvier (approved):

arsing (approved):

massand (approved):
I had hit some check errors due to diagnostics module misbehaving. But, the other checks seemed fine. Let us test this after merge, since it will be easier. Thanks!

PR #4154, opened 2020-12-23
Set default auth mode to Scope (#4095)

ancaantochi:
cherry-pick #4095
Authentication mode is set as Scope by default. This is a breaking change for leaf devices which are not in scope of edge device. Customer can set the env variable to AuthenticationMode to CloudAndScope if they need to, going forward this option will be deprecated. Will follow up with Emmanuel to have this breaking change documented.


varunpuranik (changes requested):
Let's confirm if this should go to 1.0.10 or 1.1 branch.

PR #4156, opened 2020-12-24
Added #!/bin/bash to certGet.sh (to work in zsh)

vadim-kovalyov:


dmolokanov (approved):
What does it complain about in `zsh'?

PR #4158, opened 2020-12-30
Rust test module checkin

and-rewsmith:
We are adding a rust test module and it would be nice to have a checkin gate for all rust test modules.


and-rewsmith:
We need this yaml to be in master branch before we activate the checkin.

dmolokanov (dismissed):
Overall looks good except the fact we started to add too many required pipelines which will trigger for **any** change in the repo (I'm talking about the `check_run_pipeline` job). I guess we should reconsider the approach

vadim-kovalyov:
> Overall looks good except the fact we started to add too many required pipelines which will trigger for **any** change in the repo (I'm talking about the `check_run_pipeline` job). I guess we should reconsider the approach

@dmolokanov I'm not sure I understand the concern and what is the alternative approach.

builds/checkin/rust-test-modules.yaml
```diff
@@ -0,0 +1,72 @@
...
+        displayName: Publish test results
+        inputs:
+          testResultsFormat: "JUnit"
+          testResultsFiles: "**/test-results.xml"
+          failTaskOnFailedTests: true
+        condition: succeededOrFailed()
+
+  ################################################################################
+  - job: style_check
+  ################################################################################
```
vadim-kovalyov, line 57: nit: does it make sense to have style check separate for this small tool? I believe since we removed Windows support and now build for one platform, it might be better to combine style+build+test. (like it's done in watchdog)

vadim-kovalyov (approved):
LGTM! the one question about combining style+build, but I don't have strong preference though.

dmolokanov:
> > Overall looks good except the fact we started to add too many required pipelines which will trigger for **any** change in the repo (I'm talking about the `check_run_pipeline` job). I guess we should reconsider the approach
> 
> @dmolokanov I'm not sure I understand the concern and what is the alternative approach.

@vadim-kovalyov Now we have a bunch of required checkins per project (mqtt, edgelet, api-proxy, dotnet, watchdog, libiothsm, etc) and each of them has a first job that verifies whether changes are made in the corresponding folders. Because all of them are required they all run for **any** change in the repo. As a result, we have too many pipelines triggered per change in the repo to run the very first job.

My question is that we probably don't need checkings to be that granular? Maybe we can combine api-proxy, watchdog, test modules into one pipeline since they all are relatively small modules? Or something similar...

vadim-kovalyov (approved):

vadim-kovalyov (approved):

and-rewsmith:
> As a result, we have too many pipelines triggered per change in the repo to run the very first job.

@dmolokanov It seems like a tradeoff imo. With this way It uses more agents, but has the benefit of shorter checkins. If the repeated agent-grabs are causing issues I think we should consolidate all the checkins you mentioned.

PR #4160, opened 2020-12-31
ARM Base Image Update Pipeline

yophilav:
First take on automating ARM Base Image update.

Use case scenario: 
1. tools/BaseImageUpdate/baseImage.ps1 is a helper script that a developer can use on his powershell to update the ARM base image version.  i.e. `Update-ARM-BaseImages -NewASPNetCoreVersion "2.1.23"`.
2. A developer send out a Pull request with base image changes.
3. A developer uses a pipeline `builds/misc/base-images-publish.yaml` to kick off the build & base image publication.
4. (Optional) As a good measure, once (3) is completed, a developer can run the build image pipeline to verify a successful publication.


builds/misc/templates/base-image-publish/base-image-build-and-publish.yaml
```diff
@@ -0,0 +1,31 @@
...
+      exit -1;
+    }
+
+    for ( $idx=0; $idx -lt $info.count; $idx++ )
+    {
+      $eachInfo = $info[$idx]
+      $buildCommand = "sh -c ""$scriptPath"" -d ""$($eachInfo.Item1.ToString())"" -i ""$($eachInfo.Item2.ToString())"" -v ""$($versionTags[$idx])"" "
+      echo "Command: $buildCommand"
+      # BEARWASHERE -- Don't run this unless you really want to release.
+      # $buildCommand
```
yophilav, line 29: I intentionally leave this here for the moment as I'm expecting some comment & code change from this PR. I don't want to accidentally release a base image while testing the code change.

builds/misc/templates/base-image-publish/base-image-config.yaml
```diff
@@ -0,0 +1,17 @@
+steps:
+- pwsh: |
+    $info = New-Object System.Collections.ArrayList
+    # Tuple format:  (ModuleRepoPath,                           BaseImageName,                   ReturningVstsVar)
+    $info.AddRange((
+      [Tuple]::Create("edge-agent",                             "azureiotedge-agent-base",       "edgeAgentTag"),
+      [Tuple]::Create("edge-hub",                               "azureiotedge-hub-base",         "edgeHubTag"),
+      [Tuple]::Create("edge-modules/SimulatedTemperatureSensor","azureiotedge-module-base",      "moduleTag"),
+      [Tuple]::Create("test/modules/TestAnalyzer",              "azureiotedge-module-base-full", "fullModuleTag")
```
huguesBouvier, line 9: shouldn't the API proxy be in that list?

builds/misc/templates/base-image-publish/base-image-overview.yaml
```diff
@@ -0,0 +1,26 @@
...
+    $info = '$(serialize_info)' | ConvertFrom-Json
+
+    # Fetch tags & version from each module Dockerfiles {../$(arch)/base/Dockerfile, ../$(arch)/Dockerfile}
+    foreach ($each in $info)
+    {
+      $pathPrefix = "$(Build.SourcesDirectory)/$($each.Item1.ToString())/docker/$(os)/$(arch)"
+      $unberlyingFilePath = $pathPrefix + '/base/Dockerfile' | Resolve-path
+      $underlyingTag = $($(Get-Content $unberlyingFilePath | Select-String "ARG base_tag=") -split "=")[1]
+      $imageFilePath = $pathPrefix + '/Dockerfile' | Resolve-path
+      Set-Variable -Name "$($each.Item3.ToString())" -Value "$($($(Get-Content $imageFilePath | Select-String "ARG base_tag=") -split "=")[1])"
```
and-rewsmith, line 14: I think we can disambiguate what Item3  and this value is by creating a variable for it with a label before this line. We could use this approach in other places as well.

builds/misc/templates/base-image-publish/base-image-overview.yaml
```diff
@@ -0,0 +1,26 @@
...
+      Write-Output "$($each.Item2.ToString()):$(Get-Variable -Name "$($each.Item3.ToString())" -ValueOnly) <= $underlyingTag"
```
and-rewsmith, line 15: It is possible for this to get written to console, but this should fail the build if it evaluates to false right?

builds/misc/templates/base-image-publish/base-image-overview.yaml
```diff
@@ -0,0 +1,26 @@
...
+      Write-Output "##vso[task.setvariable variable=$($each.Item3.ToString())]$(Get-Variable -Name "$($each.Item3.ToString())" -ValueOnly)"
```
and-rewsmith, line 16: Why do we set the var to itself? Or am I missing something?

builds/misc/templates/base-image-publish/base-image-overview.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+      # Extract version number from the *-linux* base image
+      # Uses (ReturningVstsVar) + 'Version' to return the version string as VSTS variable
+      #
+      # Note:
+      #    The output {$(edgeAgentTagVersion), $(edgeHubTagVersion), $(moduleTagVersion), $(fullModuleTagVersion)} are used in base-image-build-and-publish.yaml
+      Write-Output "##vso[task.setvariable variable=$($each.Item3.ToString())Version]$($(Get-Variable -Name "$($each.Item3.ToString())" -ValueOnly).split("-")[0])"
+    }
+  displayName: Overview ARM64
```
and-rewsmith, line 25: Is that a typo? Why just arm64?

builds/misc/base-images-publish.yaml
```diff
@@ -0,0 +1,55 @@
...
+        - run-base-image-release -equals true
+
+    variables:
+      os: linux
+      arch: arm32v7
+
+    steps:
+    - template: templates/base-image-publish/base-image-config.yaml
+    - template: templates/base-image-publish/base-image-overview.yaml
+    - template: ../e2e/templates/e2e-clean-directory.yaml
```
and-rewsmith, line 24: Why do we need to clean this dir?

builds/misc/templates/base-image-publish/base-image-setup-librocksdb.yaml
```diff
@@ -0,0 +1,18 @@
+steps:
+- pwsh: |
+    $artifactLocale = "$(Build.SourcesDirectory)/librocksdb.so"
+    Invoke-WebRequest -Uri $("https://edgebuild.blob.core.windows.net/librocksdb/librocksdb.so.$(arch)") -OutFile $artifactLocale
```
and-rewsmith, line 4: We will need to remember to update this file whenever we increase the rocksdb version

builds/misc/templates/base-image-publish/base-image-setup-librocksdb.yaml
```diff
@@ -0,0 +1,18 @@
...
+    $info = '$(serialize_info)' | ConvertFrom-Json
+
+    # Copy librocksdb.so to the same directory where base image is built
+    # iotedge/edge-agent/docker/linux/arm64v8/base
+    foreach ($each in $info)
+    {
+      $filePath = "$(Build.SourcesDirectory)/$($each.Item1.ToString())/docker/$(os)/$(arch)/base/librocksdb.so"
+      Copy-Item $artifactLocale -Destination "$filePath"
+
+      $isSucceeded = $(if ([System.IO.File]::Exists($filePath)) {"Succeeded"} else {"Failed"})
```
and-rewsmith, line 15: Why do we need to check this explicitly? Wouldn't the copy step fail otherwise?

builds/misc/templates/base-image-publish/base-image-config.yaml
yophilav: API proxy uses `arm32v7/haproxy:1.8.23` for ARM32 and `haproxy:1.8.23-alpine ` for AMD64. 
I don't think they are relevant to ARM base images `3.1.10-bionic-arm*` that we are updating here.

With that said, I have a few follow up question: 
1. How do we known when we should upgrade the base image for HAProxy e.g. OS security patch, bug fixes?
2. Are the HAProxy base image somehow related to the Bionic arm image releases? Are they release about the same frequency?

builds/misc/templates/base-image-publish/base-image-overview.yaml
and-rewsmith, line 14: Why do we set item3 here? Where do we use it later? It seems from the build script all we care about is the var name based on this with the "Version" suffix.

and-rewsmith (commented):
Nice work! This will really help us all out. I think step 4 you listed in the description is a great idea, and we should always do it before merging PRs which use this new process.

Has this been ran to completion yet? What changes are you planning on making in following PRs?

builds/misc/templates/base-image-publish/base-image-overview.yaml
```diff
@@ -0,0 +1,26 @@
+steps:
+- pwsh: |
+    # Note: With in the same job, you can access $(info), but it would interpret $(info) as string similar to `echo "$info"` which is undesirable for our case, we want an Object.
+    # We worked around this by using the serialization & deserailzation
+    $info = '$(serialize_info)' | ConvertFrom-Json
+
+    # Fetch tags & version from each module Dockerfiles {../$(arch)/base/Dockerfile, ../$(arch)/Dockerfile}
+    foreach ($each in $info)
```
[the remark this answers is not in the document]
yophilav: `$eachInfo` ? lol.

builds/misc/templates/base-image-publish/base-image-setup-librocksdb.yaml
yophilav: Yup. 
TODO: Create a documentation about this tool (and address this concern).

builds/misc/templates/base-image-publish/base-image-setup-librocksdb.yaml
yophilav: It may or may not (depending on the return code). I'd rather have it double check here and print out to console.

builds/misc/templates/base-image-publish/base-image-overview.yaml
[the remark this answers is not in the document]
and-rewsmith: I think something that indicates that this info is specific to one base image would help people unfamiliar with the whole process. Maybe baseImageContext?

builds/misc/base-images-publish.yaml
yophilav: Removed the file.

builds/misc/templates/base-image-publish/base-image-overview.yaml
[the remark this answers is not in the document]
yophilav: Changed to `config`
For more explicit & convention on the procedure, I'll defer to a user manual (which is to be written).

builds/misc/templates/base-image-publish/base-image-overview.yaml
yophilav: That would be a lot of duplication say... explicitly assigning {`edgeAgentTag`, `edgeHubTag`, `moduleTag`, `fullModuleTag`}. 

Note these variable are only set as env vars for this instance of `pwsh`, they are gone after this task is done.

yophilav: Per the comment above, this is setting a variable name (value of item3) as a local var to be used in the next line 2 lines of code.

builds/misc/templates/base-image-publish/base-image-overview.yaml
yophilav: The wonder of VSTS variable. This is actually setting the variable to pipeline's JOB scope. Which is different that task env var. As it turns out, we don't need this var, so I'm deleting it.

builds/misc/templates/base-image-publish/base-image-overview.yaml
yophilav: Refactoring plunder! (a typo)

builds/misc/templates/base-image-publish/base-image-overview.yaml
and-rewsmith: > That would be a lot of duplication say... explicitly assigning {edgeAgentTag, edgeHubTag, moduleTag, fullModuleTag}.

Yeah this is what I had in mind actually. It would have definitely helped me understand what this was doing better.

builds/misc/templates/base-image-publish/base-image-overview.yaml
and-rewsmith: I see the output is:
> azureiotedge-agent-base:1.0.6.4-linux-arm32v7 <= 3.1.10-bionic-arm32v7

Is this less than or equals sign supposed to be an arrow?

builds/misc/templates/base-image-publish/base-image-overview.yaml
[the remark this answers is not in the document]
and-rewsmith: nit: Still think we could make it more clear that this config is specific to one base image. The user manual makes sense, but I think the code can still be more self-documenting.

builds/misc/templates/base-image-publish/base-image-overview.yaml
yophilav: Nope I'm not unrolling a loop with 5/6 lines repeating the same code over and over. 
What you request will turn this loop into this: 
```
      $pathPrefix = "$(Build.SourcesDirectory)/edge-agent/docker/$(os)/$(arch)"
      $unberlyingFilePath = $pathPrefix + '/base/Dockerfile' | Resolve-path
      $underlyingTag = $($(Get-Content $unberlyingFilePath | Select-String "ARG base_tag=") -split "=")[1]
      $imageFilePath = $pathPrefix + '/Dockerfile' | Resolve-path
      Set-Variable -Name "edgeAgentTag" -Value "$($($(Get-Content $imageFilePath | Select-String "ARG base_tag=") -split "=")[1])"
      Write-Output "azureiotedge-agent-base:$(Get-Variable -Name "edgeAgentTag" -ValueOnly) 🠘 $underlyingTag"


      $pathPrefix = "$(Build.SourcesDirectory)/edge-hub/docker/$(os)/$(arch)"
      $unberlyingFilePath = $pathPrefix + '/base/Dockerfile' | Resolve-path
      $underlyingTag = $($(Get-Content $unberlyingFilePath | Select-String "ARG base_tag=") -split "=")[1]
      $imageFilePath = $pathPrefix + '/Dockerfile' | Resolve-path
      Set-Variable -Name "edgeHubTag" -Value "$($($(Get-Content $imageFilePath | Select-String "ARG base_tag=") -split "=")[1])"
      Write-Output "azureiotedge-hub-base:$(Get-Variable -Name "edgeHubTag" -ValueOnly) 🠘 $underlyingTag"


      $pathPrefix = "$(Build.SourcesDirectory)/edge-modules/SimulatedTemperatureSensor/docker/$(os)/$(arch)"
      $unberlyingFilePath = $pathPrefix + '/base/Dockerfile' | Resolve-path
      $underlyingTag = $($(Get-Content $unberlyingFilePath | Select-String "ARG base_tag=") -split "=")[1]
      $imageFilePath = $pathPrefix + '/Dockerfile' | Resolve-path
      Set-Variable -Name "moduleTag" -Value "$($($(Get-Content $imageFilePath | Select-String "ARG base_tag=") -split "=")[1])"
      Write-Output "azureiotedge-module-base:$(Get-Variable -Name "moduleTag" -ValueOnly) 🠘 $underlyingTag"


      $pathPrefix = "$(Build.SourcesDirectory)/test/modules/TestAnalyzer/SimulatedTemperatureSensor/docker/$(os)/$(arch)"
      $unberlyingFilePath = $pathPrefix + '/base/Dockerfile' | Resolve-path
      $underlyingTag = $($(Get-Content $unberlyingFilePath | Select-String "ARG base_tag=") -split "=")[1]
      $imageFilePath = $pathPrefix + '/Dockerfile' | Resolve-path
      Set-Variable -Name "fullModuleTag" -Value "$($($(Get-Content $imageFilePath | Select-String "ARG base_tag=") -split "=")[1])"
      Write-Output "azureiotedge-module-base-full:$(Get-Variable -Name "fullModuleTag" -ValueOnly) 🠘 $underlyingTag"
```
I'd rather have a list somewhere and formulate the necessary variable from the list; rather than have multiple duplication of code everywhere you need it.

and-rewsmith: Oh I didn't mean unrolling, I just meant adding labels to make it more readable.

builds/misc/templates/base-image-publish/base-image-overview.yaml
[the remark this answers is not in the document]
yophilav: I see where you're coming from. Without a paragraph explaining that each "base image" has an underlying OS base image associate with them. I'm not sure how else I can explain it. The user should read the code to see what it is doing. Otherwise, you will run into the maintenance hassle between (code comment vs code vs document). I drew the line that anything conceptual & user manual are going to be explained in document. For more detailed implementation, let the user read the code. Naming convention can help so much when a user doesn't have an understand of what we are trying to do.

builds/misc/templates/base-image-publish/base-image-overview.yaml
and-rewsmith: For example these lines took me a few minutes to understand:
```
      Set-Variable -Name "$($eachConfig.Item3.ToString())" -Value "$($($(Get-Content $imageFilePath | Select-String "ARG base_tag=") -split "=")[1])"
      Write-Output "$($eachConfig.Item2.ToString()):$(Get-Variable -Name "$($eachConfig.Item3.ToString())" -ValueOnly) 🠘 $underlyingTag"
```

I think having labels above these two lines for the values you are using within them would help a lot.

and-rewsmith (approved):
LGTM. Someone more familiar with devops might have some insight on how to clean it up further.

tools/BaseImageUpdate/baseImage.ps1
```diff
@@ -0,0 +1,86 @@
...
+    $fileLocale = $(Get-ChildItem -Recurse -Filter "Dockerfile" )
+
+    # Replace the underlying ASP .Net Core to the new version
+    # Assuming the ARM64 & ARM32 both use the same *-bionic-arm* ASP .Net Core image tag
+    $baseAspNetLocale = $($($fileLocale | Convert-Path) -like "*\base\*" | Resolve-path)
+    foreach ($file in $baseAspNetLocale)
+    {
+        (Get-Content $file.Path) |
+        Foreach-Object { $_ -replace "ARG base_tag=.*.-bionic-arm", "ARG base_tag=$NewASPNetCoreVersion-bionic-arm" } |
+        Set-Content $file.Path
```
damonbarry, line 67: I worry about encoding problems here. I wonder if powershell might insert a BOM in some cases, or change the encoding to utf16 or something. Should we `Get-Content` and `Set-Content` with the `-Encoding UTF8` argument?

damonbarry (changes requested):
In general, I wonder why the pipeline is broken into so many little templates? There's _a lot_ of extra complexity in here to serialize/deserialize and pass values between templates using the `##vso...` syntax, and to check for expected input values at every step along the way. I think the code would be way simpler and easier to read if `base-image-config`, `base-image-overview`, `base-image-*-librocksdb`, and `base-image-build-and-publish` were all one script.

Templates are for reusing tasks in different pipelines, but these tasks aren't shared with any other pipeline.

tools/BaseImageUpdate/baseImage.ps1
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $(cd $(Build.SourcesDirectory)) 
+    }
+    catch
+    {
+        $curDir = pwd
+        $iotedgeDir=$($curDir -split("iotedge", 2))
+        $iotedgeDir = $iotedgeDir[0] + "iotedge"
+        cd $iotedgeDir
+    }
```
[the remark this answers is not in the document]
yophilav: Yup. It seems like `$env:BUILD_SOURCESDIRECTORY` is not defined neither $PSScriptRoot

builds/misc/templates/base-image-publish/base-image-docker-login.yaml
```diff
@@ -0,0 +1,9 @@
+steps:
+- pwsh: |
+    $artifactLocale = "$(Build.SourcesDirectory)/DOCKERIO_PASS"
+    $env:DOCKERIO_PASS | Out-File -Encoding Utf8 $artifactLocale
+    cat $artifactLocale | docker login docker.io --username $(dockerio.username) --password-stdin
+    Remove-Item $artifactLocale -ErrorAction Ignore
+  displayName: Docker login
+  env:
+    DOCKERIO_PASS: $(dockerio.password)
```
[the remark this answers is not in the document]
yophilav: This is a tricky one. I have tested the login with the Dokcer@2 task but I couldn't figure out how to get the `connection` to go properly given that we are not logging into Azure Container Registry. We are logging into `https://hub.docker.com/r/azureiotedge/` (aka docker.io)

tools/BaseImageUpdate/baseImage.ps1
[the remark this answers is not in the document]
damonbarry: `$env:BUILD_SOURCESDIRECTORY` is only defined in Azure pipelines builds. `$PSScriptRoot` is an automatic variable that contains the directory from which a script is being run, since PowerShell 3.0. We're already using this logic in other scripts.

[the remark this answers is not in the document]
yophilav: > $PSScriptRoot is an automatic variable that contains the directory from which a script is being run

Ah I see. That's why it's returning empty when I tried running it striaght on the pwsh.

builds/misc/templates/base-image-publish/base-image-docker-login.yaml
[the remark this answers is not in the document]
damonbarry: You have to create a new Docker service connection in Azure pipelines. In the example I gave earlier, someone created a service connection called "iotedge-edgebuilds-acr" that connects to edgebuilds. You'd have to do the same, only for docker.io. Something like `containerRegistry: docker.io` won't work because there's no service connection called "docker.io" in our Azure pipelines...you have to create one.

builds/misc/templates/base-image-publish/base-image-setup.yaml
```diff
@@ -0,0 +1,50 @@
+steps:
```
damonbarry, line 1: This is better, but you're still distributing the work across multiple templates so you're still doing all the serialization/deserialization. Templates are good for sharing code, but they're bad for organizing code because passing information between templates requires a lot of effort. And you're not sharing this logic with any other pipelines so there's not a compelling reason to use templates. If you want organization, you can use powershell functions inside a single script.

`base-image-build-and-publish.yaml`, for example, is only about 7 lines of code if you remove the deserialization and input checking logic. You could easily merge the powershell scripts from `base-image-setup.yaml`, `base-image-build-and-publish.yaml`, and `base-image-remove-librocksdb.yaml` into one script, and it would make things a lot easier to read and maintain.

And since

tools/BaseImageUpdate/baseImage.ps1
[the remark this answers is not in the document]
damonbarry: Oh right, it's only available inside a script :)

builds/misc/templates/base-image-publish/base-image-setup.yaml
yophilav: 😆 That how it was all started by one giant yaml script. Looks like this pipeline has come a full circle: dev -> create one giant yaml script -> thinking the script is too large -> overly refactored scripts -> decided the first version is the most readible one -> create one giant script.

damonbarry: Haha...you definitely need organization and factoring, it's just that breaking up the powershell script into different yaml templates isn't ideal. If you need some organization you can break up the powershell code into functions instead, but still keep it all in one script.

Also, some amount of templates are fine since you don't want to duplicate the logic in each job, so that's why I suggested you just merge `base-image-setup.yaml`, `base-image-build-and-publish.yaml`, and `base-image-remove-librocksdb.yaml` into one yaml template with one powershell script.

Another option is to put the powershell into a ps1 file (which means you could also run it on dev desktop if needed), then call the ps1 script from each job. Either is fine with me.

yophilav: Alright, I merge the templates into a template since it's under 100... it should be ok? 
I'm not sure how I feel about having *.ps1 scripts with each executing only a sliver of a for-loop.

builds/misc/templates/base-image-publish/base-image-build-and-publish.yaml
```diff
@@ -1,31 +1,67 @@
...
-    if ($config.count -ne $versionTags.count)
+    # Get all the tag for publication
+    Write-Output "`n`nBase Image Info"
+    foreach ($eachConfig in $config)
     {
-      echo '$config and $versionTags must have the same count.'
-      exit -1;
+      # Fetch tags & version from each module Dockerfiles {../$(arch)/base/Dockerfile, ../$(arch)/Dockerfile}
+      # Steps: 
+      #  1. Go to a respective Dockerfile as defined in base-image-config.yaml for an underlying base image
```
damonbarry, line 23: Nit: There are a couple places in comments where you still refer to the deleted yaml files...

builds/misc/templates/base-image-publish/base-image-build-and-publish.yaml
```diff
@@ -1,31 +1,67 @@
...
+      #  5. Output the read values. Format:  <BaseImageName>:<ReferringTag> 🠘 <UnderLyingBaseImageTag>
+      $pathPrefix = "$(Build.SourcesDirectory)/$($eachConfig.Item1)/docker/$(os)/$(arch)"
+      $unberlyingFilePath = $pathPrefix + '/base/Dockerfile' | Resolve-path
+      $underlyingTag = $($(Get-Content $unberlyingFilePath | Select-String "ARG base_tag=") -split "=")[1]
+      $imageFilePath = $pathPrefix + '/Dockerfile' | Resolve-path
+      Set-Variable -Name "$($eachConfig.Item3)" -Value "$($($(Get-Content $imageFilePath | Select-String "ARG base_tag=") -split "=")[1])"
+      Write-Output "$($eachConfig.Item2):$(Get-Variable -Name "$($eachConfig.Item3)" -ValueOnly) 🠘 $underlyingTag"
+
+      # Extract version number from the *-linux* base image
+      # Uses (ReturningVstsVar) + 'Version' to return the version string as VSTS variable
```
damonbarry, line 36: `ReturningVstsVar` is no longer a thing...

damonbarry (dismissed):
👍👍

damonbarry (approved):

PR #4163, opened 2021-01-04
Rust clippy fix

and-rewsmith:
Fixing clippy

### DIFF
--- a/mqtt/mqtt-bridge/src/persist/waking_state/mod.rs
+++ b/mqtt/mqtt-bridge/src/persist/waking_state/mod.rs
@@ -247,7 +247,9 @@ mod tests {
     }
 
     #[test_case(WakingMemoryStore::default())]
-    //https://github.com/rust-lang/rust-clippy/issues/6353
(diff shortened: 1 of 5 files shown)


and-rewsmith:
I think it will start complaining about tests next so converting to draft for now

mqtt/mqtt-bridge/src/persist/waking_state/mod.rs
```diff
@@ -275,6 +277,8 @@ mod tests {
         // insert an element to wake the stream, then wait for the other thread to complete
         let mut state_borrow = state.borrow_mut();
         state_borrow.insert(key1, pub1).unwrap();
+        drop(state_borrow);
```
dmolokanov, line 280: why?

and-rewsmith: I thought it made this more clear that the allow was because of a false positive, but I guess it is ok as it was before.

dmolokanov (approved):
Thanks!

PR #4165, opened 2021-01-04
EFLOW on E2E test

yophilav:
Introducing E2E test on EFLOW machine


test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
```diff
@@ -17,40 +17,51 @@ public class EdgeDaemon : IEdgeDaemon
...
-                case "Raspbian":
-                    os = "debian";
-                    version = "stretch";
-                    packageExtension = SupportedPackageExtension.Deb;
-                    break;
-                case "CentOS":
-                    os = os.ToLower();
-                    version = version.Split('.')[0];
-                    packageExtension = SupportedPackageExtension.Rpm;
+                platformInfo = await Process.RunAsync("lsb_release", "-sir", token);
```
damonbarry, line 31: This works, but if we already ran `cat /etc/os-release`, do we still need to run `lsb_release -sir`? Doesn't the first command have everything we need?

ms-mahuber: That is actually a great suggestion. From the /etc/os-release output, I would take the lines "ID=xxx" and "VERSION=yyy" and feed the contents into the "os" and "version" variables. @damonbarry , does that sound good to you?

damonbarry: Sounds good, as long as we're confident `/etc/os-release` is available on all our tested platforms.

ms-mahuber: It is, and this stackoverflow post gives me high confidence that it is a better solution in overall:
https://stackoverflow.com/questions/47838800/etc-lsb-release-vs-etc-os-release

Preparing and testing according changes.

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
```diff
@@ -17,37 +17,48 @@ public class EdgeDaemon : IEdgeDaemon
...
-                platformInfo = platformInfo[0].Split(' ');
+                version = Array.Find(platformInfo, element => element.StartsWith("VERSION="));
             }
 
-            string os = platformInfo[0].Trim();
-            string version = platformInfo[1].Trim();
-            SupportedPackageExtension packageExtension;
+            if (os == null || version == null)
+            {
+                throw new NotImplementedException($"Failed to gather operating system information (OS: '{os}', Version: '{version}')");
```
damonbarry, line 34: You're trying to print `{os}` and `{version}` when one or both of them are null. Won't that cause a null-ref exception?

ms-mahuber: Thanks Damon!

I'm not a c# programmer, I thought it would print out as null. I've fixed that and ensured that potential double quotes are trimmed from the version and os variables (which can potentially happen according to what Bear and me have seen, e.g. the CentOS we're using).

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
```diff
@@ -17,37 +17,49 @@ public class EdgeDaemon : IEdgeDaemon
...
+            string os, version;
+            string[] platformInfo = await Process.RunAsync("cat", @"/etc/os-release", token);
+            os = Array.Find(platformInfo, element => element.StartsWith("ID="));
+            version = Array.Find(platformInfo, element => element.StartsWith("VERSION_ID="));
+
+            // VERSION_ID is desired but it is an optional field
+            if (version == null)
             {
-                platformInfo = platformInfo[0].Split(' ');
+                version = Array.Find(platformInfo, element => element.StartsWith("VERSION="));
```
yophilav, line 29: The `VERSION` field can have any of the following values:
```
VERSION="18.04.4 LTS (Bionic Beaver)"
VERSION="1.0.20201124"  
VERSION="7 (Core)"
```
We may want to sterile them first before setting the C# `version`

ms-mahuber: Done in new revision

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/EdgeDaemon.cs
damonbarry: Actually, it might just print an empty string...I don't remember.

damonbarry (approved):

damonbarry (approved):

damonbarry (dismissed):

richma-ms (approved):

PR #4170, opened 2021-01-05
Fix E2E test pipelines for latest iot-identity-service

gordonwang0:


test/Microsoft.Azure.Devices.Edge.Test.Common/DaemonConfiguration.cs
```diff
@@ -209,6 +230,19 @@ public void RemoveCertificates()
             this.config[Service.Certd].Document.RemoveIfExists("preloaded_certs.iotedge-trust-bundle");
         }
 
+        public void AddPrincipal(string name, uint uid)
```
massand, line 233: For completeness, should we add an optional type parameter for local identity testing?

gordonwang0: Fixed.

massand (approved):

PR #4172, opened 2021-01-05
Update current config when plan is empty (#4166)

ancaantochi:
cherry-pick #4166


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/Agent.cs
```diff
@@ -146,6 +146,11 @@ public async Task ReconcileAsync(CancellationToken token)
 
                         if (plan.IsEmpty)
                         {
+                            if (this.currentConfig.Version != deploymentConfigInfo.Version)
```
nyanzebra, line 149: Should we have test cases for both possibilities? Where versions are the same and different? From what I can tell below, it looks like we have the test for both being the same. Feel free to resolve if I missed something and both cases are covered.

ancaantochi: by default currentConfig.Version is -1 so it tests for different, but it looks like there is no test covering when they are equal. This is a cherry-pick, I will add the test in a different PR in master

vadim-kovalyov (approved):

PR #4174, opened 2021-01-06
Fixing test case for API proxy

huguesBouvier:
Fixing test case for API proxy


gordonwang0:
This seems to be fixing the same issue that I found in #4170.

Should I use this fix instead? The fix I made was to pin
```
syn = "=1.0.57"
```

gordonwang0 (approved):

and-rewsmith:
@gordonwang0 I think pinning syn is fine. The reason we couldn't do that here is because the test-case crate internally uses syn, so they are the ones that need to pin it. Also, we weren't even using test-case crate here so it can just be removed.

PR #4179, opened 2021-01-07
Use native precondition for MQTT CI

vadim-kovalyov:
## Problem:
Right now if someone wants to run MQTT CI on an arbitrary commit - they can't. The commit must contain changes in mqtt or builds folder.

## Solution:
Out other CI pipelines don't use "precondition task" to check if they need to run or not. I'm removing the precondition check from MQTT CI as well, but instead I'm adding "native" precondition check for `builds` and `mqtt` folders. This way it will be possible to start the pipeline manually, but it won't be auto-triggered unnecessarily.


varunpuranik (approved):

dmolokanov (approved):
:shipit:

PR #4181, opened 2021-01-07
Remove conflicting packages left by iiot runs

gordonwang0:
iiot usually cleans up its packages on completion, but won't be able to do this if it crashes, times out, or is canceled. Update tests to remove these packages if encountered.


yophilav (approved):

yophilav:
Please don't forget to cherry-pick to release/1.0.10 & release/1.2 as well. 
(They share the same test agent and all of them need to delete the artifact.)

PR #4184, opened 2021-01-07
Fix CI tests after default authentication mode changed to Scope

ancaantochi:
CI tests were failing because the authentication mode was changed to Scope only. A few changes to the tests to make it work:
- leaf devices should have scope set to the edge device scope
- fixture should use edgeHub sas key to connect to cloud because IotHub doesn't authorize to call scope API with device sas key


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/ProtocolHeadFixture.cs
```diff
@@ -78,7 +78,19 @@ async Task StartProtocolHead()
...
+            string iothub = ConnectionStringHelper.GetHostName(edgeDeviceConnectionString);
+            RegistryManager rm = RegistryManager.CreateFromConnectionString(iotHubConnectionString);
+            var edgeHubModule = await rm.GetModuleAsync(edgeDeviceId, "$edgeHub");
+            if (edgeHubModule.Authentication.Type == AuthenticationType.None)
+            {
+                edgeHubModule.Authentication.Type = AuthenticationType.Sas;
+                edgeHubModule = await rm.UpdateModuleAsync(edgeHubModule);
+            }
+
+            ConfigHelper.TestConfig[EdgeHubConstants.ConfigKey.IotHubConnectionString] = $"HostName={iothub};DeviceId={edgeDeviceId};ModuleId=$edgeHub;SharedAccessKey={edgeHubModule.Authentication.SymmetricKey.PrimaryKey}";
```
and-rewsmith, line 93: How is this connection string we are setting different from the ```edgeDeviceConnectionString``` we get from keyvault above?

test/modules/CloudToDeviceMessageTester/Settings.cs
```diff
@@ -54,7 +54,8 @@ class Settings
                     ApiVersion = Preconditions.CheckNonWhiteSpace(apiVersion, nameof(apiVersion)),
                     ModuleGenerationId = Preconditions.CheckNonWhiteSpace(moduleGenerationId, nameof(moduleGenerationId)),
                     IotHubHostName = Preconditions.CheckNonWhiteSpace(iotHubHostName, nameof(iotHubHostName)),
-                    TransportType = transportType
+                    TransportType = transportType,
+                    EdgeDeviceId = deviceId
```
and-rewsmith, line 58: nit: both should be edgeDeviceId?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.E2E.Test/ProtocolHeadFixture.cs
ancaantochi: keyvault has device connection string and here it's using edgeHub connection string

test/modules/CloudToDeviceMessageTester/Settings.cs
ancaantochi: deviceId is read from env variable and for leaf device is using that name and appends "-leaf", so deviceId is actually edgeDeviceId

and-rewsmith (approved):

PR #4189, opened 2021-01-08
fixed subscription restore from client state when mqtt client reconnects

vipeller:
When an MQTT client used persistent session, the previous subscriptions did not get restored. If the same client connected to IoTHub directly, its subscriptions (e.g. for receiving twin responses) would have been restored.
In case of c# sdk it would not be a problem, because the sdk automatically resubscribes when it is able to restore the connection to edgeHub.

In case of python sdk however this does not happen. As a result, if a device connects using the python sdk and edgeHub gets restarted, the client will not be receiving m2m messages, twin patches, etc anymore.

The fix seems big but the main code in "MessagingServiceClient.cs", the rest is just the result of a changed interface and some plumbing.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
```diff
@@ -71,10 +71,13 @@ public Task ProcessMethodResponseAsync(IMessage message)
             return Task.CompletedTask;
         }
 
-        public void BindDeviceProxy(IDeviceProxy deviceProxy)
+        public void BindDeviceProxy(IDeviceProxy deviceProxy, Option<Action> initWhenBound)
         {
             this.underlyingProxy = Preconditions.CheckNotNull(deviceProxy);
-            this.connectionManager.AddDeviceConnection(this.Identity, this);
+
+            var connectionCreation = this.connectionManager.AddDeviceConnection(this.Identity, this);
```
varunpuranik, line 78: Can we wrap all of this async logic inside an async void method?

varunpuranik (dismissed):
Change looks good. I am just concerned about all the async calls from sync context. Might be better to make it all async (with Tasks), and then call it from an async void method.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/ClientConnectionHandler.cs
```diff
@@ -50,7 +50,7 @@ public Task<IDeviceListener> GetDeviceListener()
                                         // TODO: Implement plug and play for AMQP
                                         IDeviceListener dl = await this.connectionProvider.GetDeviceListenerAsync(this.identity, Option.None<string>());
                                         var deviceProxy = new DeviceProxy(this, this.identity);
-                                        dl.BindDeviceProxy(deviceProxy);
+                                        dl.BindDeviceProxy(deviceProxy, Option.None<Action>());
```
ancaantochi, line 53: I prefer having an interface instead of passing option of Action

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/StringIdentity.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace Microsoft.Azure.Devices.Edge.Hub.Mqtt
+{
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Identity;
```
ancaantochi, line 4: unnecessary

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/StringIdentity.cs
```diff
@@ -0,0 +1,18 @@
...
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Azure.Devices.ProtocolGateway.Identity;
+
+    public class StringIdentity : IDeviceIdentity
```
ancaantochi, line 8: nit: rename to SimpleIdentity or AuthenticatedIdentity or even DeviceIdentity

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Amqp/ClientConnectionHandler.cs
vipeller: changed it to interface

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/StringIdentity.cs
vipeller: renamed

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Mqtt/StringIdentity.cs
vipeller: removed

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/device/DeviceMessageHandler.cs
vipeller: moved out

varunpuranik (approved):

PR #4190, opened 2021-01-08
iotedge-init-import: Fix import of listen URIs when original config used socket activation.

arsing:
When the original config used socket activation, the `fd://` URI cannot be
directly reused for the new config, because the socket units have been renamed.
Instead, the new config needs to use the new socket unit names.

Also, any existing Edge modules would have volume mounts for
`/var/{lib,run}/iotedge/{mgmt,workload}.sock` that cannot be changed without
deleting and recreating the modules. Since we don't want the user to have to
delete the modules manually before upgrade, nor do we want the upgrade to
delete the modules for them (since that can cause data loss), we decided to
revert the new socket units to the old socket file paths.

Therefore, aziot-edged still has a new homedir `/var/lib/aziot/edged`,
but the sockets are bound under `/var/run/iotedge` (`/var/lib/iotedge` when
not using systemd socket-activation) as before.

Lastly, this change fixes the existing init-import test cases
that incorrectly used the new default connect URIs as
the old config's connect URIs instead of the old default connect URIs.


edgelet/iotedge/src/init/import/old_config/mod.rs
```diff
@@ -37,14 +37,14 @@ pub(crate) struct Config {
 
 #[derive(Debug, serde_derive::Deserialize)]
 pub(crate) struct Connect {
-    pub(crate) workload_uri: Url,
     pub(crate) management_uri: Url,
+    pub(crate) workload_uri: Url,
```
daprilik, line 41: nit: any reason you shuffled around the order in the struct?
not a blocker, just curious

daprilik (approved):

edgelet/iotedge/src/init/import/old_config/mod.rs
arsing: It was, and still is, driving me crazy.

The usual order we think of is management -> workload, and the default config file has management -> workload, but the struct in edgelet-core has workload -> management, and the code that uses the settings in iotedged does workload -> management.  It didn't matter until `iotedge init import` arrived and started writing out configs, which is why all the output `edged.yaml` in the test cases have workload -> management. But because the default config file has management -> workload I wanted to keep the input config files of the test cases management -> workload.

With this PR, I added a function that returns two URIs (the defaults), so I had to pick what order it returns them in. The code that uses the defaults would look weird if it parsed the old config in one order and overrode with the defaults in the opposite order. So I just made everything related to the old config in the management -> workload order.

daprilik (approved):

edgelet/iotedge/test-files/init/import/not-socket-activated-varlib/device-id
```diff
@@ -0,0 +1 @@
+aziot-identity-service|aziot-ide
```
massand, line 1: nit: missing newline

edgelet/iotedge/test-files/init/import/not-socket-activated-varrun/device-id
```diff
@@ -0,0 +1 @@
+aziot-identity-service|aziot-ide
```
massand, line 1: nit: missing newline

edgelet/iotedge/test-files/init/import/socket-activated-one-custom/device-id
```diff
@@ -0,0 +1 @@
+aziot-identity-service|aziot-ide
```
massand, line 1: nit: missing newline

scripts/linux/runE2ETest.sh
```diff
@@ -14,9 +14,10 @@ function clean_up() {
     systemctl stop aziot-edged || true
 
     echo 'Remove IoT Edge and config file'
-    rm -rf /var/lib/aziot/edged/
-    rm -rf /etc/systemd/system/aziot-edged.service.d/
-    rm -rf /etc/aziot/edged/config.yaml
+    rm -rf /var/lib/aziot/
+    rm -rf /var/lib/iotedge/
+    rm -rf /etc/aziot/
+    rm -rf /etc/systemd/system/aziot-*.service.d/
```
massand, line 20: Is removal of non-iotedged service folders intentional here?

massand: Never min d - Is see the trcE2Etest.sh and this would be consistent with that

massand (commented):
Few nits

edgelet/iotedge/test-files/init/import/not-socket-activated-varlib/device-id
arsing: https://github.com/Azure/iot-identity-service/pull/45#discussion_r505589143

gordonwang0:
Running E2E tests on this. The CentOS and Pi jobs should work.
https://msazure.visualstudio.com/One/_build/results?buildId=38925564&view=results

daprilik (approved):

gordonwang0 (approved):
Tests look ok.

PR #4195, opened 2021-01-08
Move 'Download aziot-identity-service' to separate file

gordonwang0:
Move the script that downloads aziot-identity-service from the E2E test YAML so it can be shared with the Connectivity test pipeline.


massand (approved):
:shipit:

PR #4196, opened 2021-01-08
Added timeout sending messages to MQTT broker to prevent infinite waiting

vipeller:
This is for being on the safe side: when edgeHub sends a message to the mqtt broker, it waits for ACK. There was no timeout for this. The odds are slim that no ack will come back: this is a local communication, and on top of that, the mqtt client library should resend the message if no ack. Also, when the mqtt broker disconnects, that cancels all the pending operations (so no call would be waiting forever). Yet, to avoid a situation when the code gets stuck in a Send call, a timeout has been added.


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
```diff
@@ -248,6 +248,34 @@ public async Task SendAsyncCancelsWhenDisconnecting()
             await Assert.ThrowsAsync<TaskCanceledException>(async () => await producer.Connector.SendAsync("boo", Encoding.ASCII.GetBytes("hoo")));
         }
 
+        // Note, that this test tests a corner case scenario, because the connection between edgeHub and the broker
+        // should not be broken, and if it is, then the MQTT library should resend the message automatically, which
+        // ultimately deliver an ACK. We added an internal timeout anyway to prevent infinite waiting (and freeze a
+        // processing loop). This timeout is 10 seconds. As a result, this test completes after 10 seconds.
```
nyanzebra, line 254: Could we have the 10 sec timeout be configurable? Then for testing we can lower the timeout. Or if we find 10 sec to be more than enough we can lower it as well for production.

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
```diff
@@ -248,6 +248,34 @@ public async Task SendAsyncCancelsWhenDisconnecting()
...
+                };
+
+            var producer = new ProducerStub();
+
+            sut = new ConnectorBuilder()
+                .WithProducer(producer)
+                .Build();
+
+            await sut.ConnectAsync(HOST, broker.Port);
+            await Assert.ThrowsAsync<TimeoutException>(async () => await producer.Connector.SendAsync("boo", Encoding.ASCII.GetBytes("hoo")));
```
vadim-kovalyov, line 276: so, this test will run for at least 10 sec? is there a way to parametrize the timeout? Also, if timeout happens, we don't remove the message from `pendingAcks`, is that ok?

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
vipeller: added setting for publish-ack timeout

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/MqttBrokerConnectorTest.cs
vipeller: - fixed the pendingAcks issue and added a property to change timeout

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -189,7 +194,21 @@ public async Task<bool> SendAsync(string topic, byte[] payload, bool retain = fa
...
-            var result = await tcs.Task.TimeoutAfter(TimeSpan.FromSeconds(AckTimeoutSecs));
+            bool result;
+
+            try
+            {
+                result = await tcs.Task.TimeoutAfter(this.AckTimeout);
+            }
+            catch
+            {
+                lock (this.guard)
```
vadim-kovalyov, line 205: why not use `ConcurrentDictionary`?

nyanzebra (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: There is a place when need to keep the dictionary locked until another operation is finished, so that is why I use a regular lock. Because of that, every other place needs to be locked, so at the end (because of the manual locks) a ConcurrentDictionary would not give any benefit.

vadim-kovalyov (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -32,10 +32,14 @@ public class MqttBrokerConnector : IMqttBrokerConnector
 
         AtomicBoolean isRetrying = new AtomicBoolean(false);
 
+        public TimeSpan AckTimeout { get; set; }
```
varunpuranik, line 35: Pls make it read only

varunpuranik (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: This value needs to be modified after instantiation, that is the reason this property is not read-only. The full story:
- The ack timeout was added to avoid that very slim possibility that on a local connection the broker (or the mqtt library) does not ack back a publish operation, and the call get stuck forever. The very first implementation used a 10 sec hardcoded value.
- Because of that the related unit test needed to wait for this 10 sec to expire. Reacting to the PR feedbacks, I added this property, so the unit test can lower the value.

Other options could have been:
- add the ack timeout via constructor parameter. I did not do that for two reasons:
  - 'too visible' change, considering that it is only to make a test run shorter
  - passing numeric values in constructor is not too autofac friendly. Currently this class does not have a custom activator and creating one just for the sake of a faster unit test did not seem reasonable.
- make the originally constant value a read-only private variable and modify it from the unit test via reflection. Although I liked that idea because it would have left the class interface clean, according to previous experiences, tricks like that are not always popular.
- So I ended up adding this property which can be used after the class is instantiated, so the unit test can change the timeout before it publishes. The drawback that it is still to 'visible', when the only reason is to speed up a test.
- (also considered creating the unit test only for myself to see that the code was working, then disable with a skip="some explanation")

PR #4208, opened 2021-01-12
[k8s] Allow non standard ports for container registry.

darobs:
The conversion from the AuthConfig username and password
to ImagePullName was allowing invalid characters to be set as the name.
Specifically, valid server addresses can include port numbers and http/s
prefixes.
This change gets the only the server name and port (if port is set) and
incorparates that into the image pull secret name. This will allow
two registries on the same server name with different ports.
A side-effect of this is that it will throw the exception in the command
execution, which should be slightly easier to debug.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/ImagePullSecretName.cs
```diff
@@ -8,17 +8,37 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment
...
+                stripped = stripped.Substring(HttpsStart.Length);
+            }
+            else if (serverAddress.StartsWith(HttpStart))
+            {
+                stripped = stripped.Substring(HttpStart.Length);
+            }
+
+            string[] nameparts = stripped.Split('/', 2);
+            return nameparts[0].Replace(':', '-');
+        }
```
zasherif, line 40: Is it more standard to use uri.host instead of coding one? https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=netframework-4.8

darobs: You know what stinks?  Registry addresses aren't URIs.  I did attempt to use the URI class here, eventually, I had to settled on implementing `ConvertToHostName` from the Moby project.

zasherif: re: "Registry addresses aren't URIs". That's a bit weird because if I put all the examples you have - in ValidServerAddresses method - in a URI validator, the are actually valid URIs. But since you tried that URI class and it did not work, it lgtm.

darobs: Using the URI class failed on this test: "my-acr.io:8080" the URI class interprets that as protocol="my-acr.io" with hostname="8080" :(

zasherif (dismissed):
lgtm after fixing some failing checks

ggjjj (approved):
:shipit:

PR #4219, opened 2021-01-13
Enable package builds for Mariner

chzawist:
Includes PR "Update Mariner RPM version numbers automatically during build" (#4202)


damonbarry:
@yophilav This PR is from January...do we need it anymore?

PR #4220, opened 2021-01-13
Fix E2E test module test flakiness from moby issue

and-rewsmith:
I noticed that our module tests are flaky in new e2e because of a moby bug (lingering container reference is left in the iotedge docker network even though the container doesn't exist). The way to fix this is to restart the docker service after we uninstall iotedge (which should also work on CentOs). The same mitigation was made in the old e2e tests:
https://github.com/Azure/iotedge/blob/master/scripts/linux/runE2ETest.sh#L25


and-rewsmith:
Passing E2E:
https://msazure.visualstudio.com/One/_build/results?buildId=38411666&view=results

damonbarry (approved):

PR #4221, opened 2021-01-13
Separate commands and add better logging for e2e tests

dylanbronson:
Separating commands for uninstalling everything edge related on agents before our end to end tests run. Previously, if one fails, they all might fail. This change will make it so that each is independent. We probably don't need to uninstall libiothsm-std anymore, but this is a fail safe to get the agent back to a good state in case some agent is used by an older pipeline.


gordonwang0 (approved):

test/Microsoft.Azure.Devices.Edge.Test.Common/linux/PackageManagement.cs
```diff
@@ -110,11 +110,17 @@ public string[] GetInstallCommandsFromMicrosoftProd(Option<Uri> proxy)
...
-                "dpkg --purge libiothsm-std aziot-edge aziot-identity-service iotedge"
+                "dpkg --purge aziot-edge",
+                "dpkg --purge aziot-identity-service",
+                "dpkg --purge iotedge",
+                "dpkg --purge libiothsm-std"
             },
             SupportedPackageExtension.Rpm => new[]
             {
-                "yum remove -y libiothsm-std aziot-edge aziot-identity-service iotedge"
+                "yum remove -y --remove-leaves aziot-edge ",
```
gordonwang0, line 120: Nit: extra space at the end here.

gordonwang0 (approved):

gordonwang0 (dismissed):

gordonwang0 (approved):

PR #4225, opened 2021-01-13
Upgrade Ubuntu hosted agents

damonbarry:
I took a pass though our Azure pipelines YAML files to update any Ubuntu 16.04 hosted agents to 18.04. Many of these pipelines define our Tier 1 support, and since Ubuntu 16.04 won't be supported much longer we want to pin to the OS version we support.

Note that any Docker images containing our Rust components will still be built on 16.04-based agents for now, because there's some non-trivial work to update `cross` dependencies. That's okay because we aren't officially releasing any of those images for 1.1.0.

This PR also fixes a problem with a setup step in the end-to-end tests pipeline on Windows; it was trying to use docker to do some cleanup before docker is actually installed (usually it will be installed from the previous run but you can't always count on that, e.g. for a new agent, or if someone logged onto the agent and uninstalled IoT Edge).

### DIFF
--- a/builds/checkin/dotnet.yaml
+++ b/builds/checkin/dotnet.yaml
@@ -12,7 +12,7 @@ jobs:
 ################################################################################
     displayName: Check pipeline preconditions (changes ARE NOT in either edgelet, docs, or mqtt folder)
     pool:
-      vmImage: ubuntu-latest
+      vmImage: ubuntu-18.04
     steps:
       - bash: |
           git log -m -1 --name-only --first-parent --pretty="" | egrep -i -v '^(edgelet|doc|mqtt)'
--- a/builds/checkin/dotnet.yaml
+++ b/builds/checkin/dotnet.yaml
@@ -30,7 +30,7 @@ jobs:
     dependsOn: check_run_pipeline
     condition: eq(dependencies.check_run_pipeline.outputs['check_files.RUN_PIPELINE'], 'true')
     pool:
-      vmImage: ubuntu-latest
+      vmImage: ubuntu-18.04
     steps:
       - template: ../templates/install-dotnet2.yaml
       - template: ../templates/install-dotnet3.yaml
--- a/builds/checkin/dotnet.yaml
+++ b/builds/checkin/dotnet.yaml
@@ -12,7 +12,7 @@ jobs:
 ################################################################################
     displayName: Check pipeline preconditions (changes ARE NOT in either edgelet, docs, or mqtt folder)
     pool:
-      vmImage: "ubuntu-16.04"
+      vmImage: ubuntu-latest
     steps:
       - bash: |
           git log -m -1 --name-only --first-parent --pretty="" | egrep -i -v '^(edgelet|doc|mqtt)'
@@ -30,7 +30,7 @@ jobs:
     dependsOn: check_run_pipeline
     condition: eq(dependencies.check_run_pipeline.outputs['check_files.RUN_PIPELINE'], 'true')
     pool:
-      vmImage: "ubuntu-16.04"
+      vmImage: ubuntu-latest
     steps:
       - template: ../templates/install-dotnet2.yaml
       - template: ../templates/install-dotnet3.yaml
--- a/builds/ci/dotnet.yaml
+++ b/builds/ci/dotnet.yaml
@@ -14,7 +14,7 @@ jobs:
     displayName: Linux
     timeoutInMinutes: 120
     pool:
-      vmImage: ubuntu-16.04
+      vmImage: ubuntu-latest
     variables:
       testEnvironment: linux
     steps:
--- a/builds/ci/dotnet.yaml
+++ b/builds/ci/dotnet.yaml
@@ -14,7 +14,7 @@ jobs:
     displayName: Linux
     timeoutInMinutes: 120
     pool:
-      vmImage: ubuntu-latest
+      vmImage: ubuntu-18.04
     variables:
       testEnvironment: linux
     steps:
--- a/builds/e2e/templates/e2e-clear-docker-cached-images.yaml
+++ b/builds/e2e/templates/e2e-clear-docker-cached-images.yaml
@@ -38,18 +38,6 @@ steps:
 
     if ($IsWindows)
     {
-        Get-Command docker -ErrorAction SilentlyContinue
-        if (!$?) {
-            Write-Output 'Docker not installed, skipping...'
-            Exit 0
-        }
-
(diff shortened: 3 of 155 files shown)


arsing:
Even though it works, it'll probably be confusing to have half the jobs on`-18.04` and the other half on `-latest`. The OS update doesn't need to happen as often, so I feel manually updating it every time we need to is better.

Besides, I'm not sure having a floating baseline OS is a good idea for an LTS branch especially, since `-latest`'s presumed upgrade to `-22.04` would happen in the middle of the LTS period.

damonbarry:
> Even though it works, it'll probably be confusing to have half the jobs on`-18.04` and the other half on `-latest`. The OS update doesn't need to happen as often, so I feel manually updating it every time we need to is better.
> 
> Besides, I'm not sure having a floating baseline OS is a good idea for an LTS branch especially, since `-latest`'s presumed upgrade to `-22.04` would happen in the middle of the LTS period.

Good point about LTS. I'm happy to fix them all to the same version. Stay tuned...

gordonwang0:
I'm using 20.04 in the E2E pipeline
https://github.com/Azure/iotedge/blob/84cfc9a3e1efed25b484bbd34fbd32e8ac6c43f2/builds/e2e/e2e.yaml#L57

arsing (approved):

damonbarry:
> I'm using 20.04 in the E2E pipeline. Should that be updated to whichever we decide here?

@gordonwang0 Yeah probably.

edgelet/build/linux/install.sh
```diff
@@ -69,26 +69,26 @@ process_args "$@"
 
 install_rust
 
-# Add trusty repo to get older version of libc6-armhf-cross
-sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu/ trusty main universe"
-
 # Install OpenSSL, curl and uuid and valgrind
 sudo apt-get update || :
 sudo apt-get install -y \
     pkg-config \
```
yophilav, line 75: Per https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md, it looks like the `pkg-config` is preinstalled in 18.04

mqtt/build/linux/install.sh
```diff
@@ -69,26 +69,26 @@ process_args "$@"
 
 install_rust
 
-# Add trusty repo to get older version of libc6-armhf-cross
-sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu/ trusty main universe"
-
 # Install OpenSSL, curl and uuid and valgrind
 sudo apt-get update || :
 sudo apt-get install -y \
     pkg-config \
```
yophilav, line 75: `pkg-config` comes preinstalled. (see comment above)

builds/checkin/dotnet.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
     dependsOn: check_run_pipeline
     condition: eq(dependencies.check_run_pipeline.outputs['check_files.RUN_PIPELINE'], 'true')
     pool:
-      vmImage: "ubuntu-16.04"
+      vmImage: ubuntu-18.04
     steps:
       - template: ../templates/install-dotnet2.yaml
       - template: ../templates/install-dotnet3.yaml
```
yophilav, line 36: Is there any particular reason we need to manually install .NetCoreSdk 2.2.207, and 3.1.202 manually? 

Can we simply just use the pre-installed .NetSDK listed here? 
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md#net-core-sdk
Note: We for sure don't need the `- template: ../templates/install-dotnet3.yaml ` since 3.1.202 comes preinstalled.

builds/ci/dotnet.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     displayName: Linux
     timeoutInMinutes: 120
     pool:
-      vmImage: ubuntu-16.04
+      vmImage: ubuntu-18.04
```
yophilav, line 17: On Line35, we don't need `../templates/install-dotnet3.yaml`. (May be we can also get rid of the `../templates/install-dotnet2.yaml`?! ) 
Please investigate if we can dotnet2 and (hopefully remove it from the rest of the code base).

builds/misc/templates/image-linux-rust-jobs.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
 ################################################################################
     displayName: Linux Rust ${{ parameters.archName }}
     pool:
-      vmImage: 'ubuntu-16.04'
+      vmImage: ubuntu-16.04
```
yophilav, line 13: Did you mean to update this to 18.04?

builds/e2e/templates/e2e-clear-docker-cached-images.yaml
```diff
@@ -1,6 +1,23 @@
...
+        Get-Command docker -ErrorAction SilentlyContinue
+        if ($?) {
+            $moby = Get-Service 'iotedge-moby' -ErrorAction SilentlyContinue
+            if (($moby -ne $null) -and ($moby.Status -eq 'Running')) {
+                exit 0
+            }
+        }
+
+        Write-Output 'Docker not installed, skipping...'
+        echo "##vso[task.setvariable variable=hasDocker]No"
```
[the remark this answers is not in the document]
damonbarry: Output variables are of type string not bool. So `ne('true')` and `ne('false')` would both return `true`.

builds/checkin/dotnet.yaml
damonbarry: Not sure, but that's outside the scope of this change.

builds/misc/templates/image-linux-rust-jobs.yaml
damonbarry: No, these images fail to build if you upgrade the agent to 18.04. So I put them back to 16.04, and we can tackle it later.

edgelet/build/linux/install.sh
damonbarry: I've been going back and forth on this. `pkg-config` is (apparently) a prerequisite for the logic in this script, and if 18.04 already has it then it will just be a no-op. So leaving it in doesn't impact the script's correctness.

I really want to keep this PR focused on upgrading agents. If we want to make a pass through the install script to streamline it or remove cruft, that can be done separately.

builds/ci/dotnet.yaml
damonbarry: Again, that's not the focus of this PR.

arsing (approved):

yophilav (approved):

PR #4227, opened 2021-01-14
Generic mqtt test module trc integration

and-rewsmith:
Integrating the mqtt test module with the trc. I also have made a separate abstraction for message initiator and improved the shutdown logic.

In the next PR I am going to fix the folder and crate names to adhere to standard format. I didn't do it here because the work was done in a separate branch and I didn't want to make the diff confusing.


test/modules/GenericMqttTester/src/main.rs
```diff
@@ -43,7 +43,7 @@ async fn main() -> Result<()> {
 }
 
 fn init_logging() {
-    let subscriber = Subscriber::builder().with_max_level(Level::TRACE).finish();
+    let subscriber = Subscriber::builder().with_max_level(Level::INFO).finish();
```
and-rewsmith, line 46: @vadim-kovalyov The reason I bumped this to info was to stop hyper from spitting out tons of logs. I tried to do this the another way first with the ```with_env_filter``` method passing in the string ```trace,hyper=info``` but this didn't work.

From my investigation in a console app I think the problem is because hyper isn't a direct dependency, but exists in the crate trc-client which the generic-mqtt-tester takes a dependency on. Am I right? What is the correct way I can filter this more granuarly?

test/modules/GenericMqttTester/src/message_initiator.rs
```diff
@@ -0,0 +1,110 @@
...
+                }
+                Either::Right((publish, _)) => {
+                    publish.map_err(MessageTesterError::Publish)?;
+                }
+            };
+
+            self.report_message_sent(seq_num).await?;
+            seq_num += 1;
+
+            time::delay_for(Duration::from_secs(1)).await;
```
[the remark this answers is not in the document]
and-rewsmith: Yeah I think that's a great idea

test/modules/GenericMqttTester/config/default.json
```diff
@@ -1,4 +1,4 @@
 {
     "test_scenario": "Initiate",
-    "test_result_coordinator_url": "http://testResultCoordinator:5001/api/testoperationresult"
+    "trc_url": "http://testResultCoordinator:5001/api/testoperationresult"
```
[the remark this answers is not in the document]
and-rewsmith: I had it like this previously, but on the last PR I got feedback to shorten it. The logic is that we give necessary context in the Cargo.toml description.

test/modules/GenericMqttTester/src/main.rs
vadim-kovalyov: I'm not sure what is the correct way for managing different log level for different components. I have an open PBI for that for the broker :) But I'm totally fine for bumping max level to INFO.

test/modules/GenericMqttTester/src/lib.rs
```diff
@@ -1,53 +1,85 @@
...
+            .await
+            .map_err(MessageTesterError::SendShutdownSignal)
+    }
+}
 
 #[derive(Debug, thiserror::Error)]
 pub enum MessageTesterError {
-    #[error("could not parse expected env vars")]
-    ParseEnvironment(#[source] VarError),
+    #[error("could not parse expected env vars: {0:?}")]
```
vadim-kovalyov, line 41: 👍🏻

vadim-kovalyov (approved):
LGTM!

dylanbronson (approved):

PR #4230, opened 2021-01-14
Fix Add-ons pipeline

gordonwang0:
Changes the use of $(version) to $(Build.BuildNumber) if $(version) is empty.

### DIFF
--- a/builds/misc/addons-release.yaml
+++ b/builds/misc/addons-release.yaml
@@ -1,5 +1,5 @@
 name: $(version)
-jobs: 
+jobs:
 ################################################################################
   - job: linux_API_proxy_module
 ################################################################################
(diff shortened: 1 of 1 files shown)


builds/misc/addons-release.yaml
```diff
@@ -20,11 +20,11 @@ jobs:
           name: API Proxy
           imageName: azureiotedge-api-proxy
           project: api-proxy-module
-          version: $(version)
+          version: $(Build.BuildNumber)
```
damonbarry, line 23: I would leave this one as it was, since it was working. If you change this and someone actually enters a value for the "version" variable, this change would ignore it.

builds/misc/addons-release.yaml
```diff
@@ -20,11 +20,11 @@ jobs:
...
       # Check API proxy
       - task: ComponentGovernanceComponentDetection@0
         inputs:
-          dockerImagesToScan: '$(registry.address)/microsoft/azureiotedge-api-proxy:$(version)-linux-amd64,$(registry.address)/microsoft/azureiotedge-api-proxy:$(version)-linux-arm32v7,$(registry.address)/microsoft/azureiotedge-api-proxy:$(version)-linux-arm64v8' 
+          dockerImagesToScan: '$(registry.address)/microsoft/azureiotedge-api-proxy:$(Build.BuildNumber)-linux-amd64,$(registry.address)/microsoft/azureiotedge-api-proxy:$(Build.BuildNumber)-linux-arm32v7,$(registry.address)/microsoft/azureiotedge-api-proxy:$(Build.BuildNumber)-linux-arm64v8'
```
damonbarry, line 27: Again if someone actually passes in "version"...you should probably default "version" to $(Build.BuildNumber) if it's empty, otherwise use version.

damonbarry (changes requested):

builds/misc/addons-release.yaml
```diff
@@ -1,12 +1,18 @@
...
   - job: linux_API_proxy_module
 ################################################################################
     displayName: linuxAPIProxy
     pool:
       vmImage: 'ubuntu-18.04'
     steps:
+      - bash: |
+            build_version='$(version)'
+            if [ -z "$build_version" ] then
+              echo '##vso[task.setvariable variable=buildVersion]$(Build.BuildNumber)'
```
damonbarry, line 13: `buildVersion` needs to be set to `$(version)` in the else case.

gordonwang0: Just realized that :D
Fixed

damonbarry (dismissed):

damonbarry (dismissed):

damonbarry (approved):

PR #4232, opened 2021-01-14
Making NetworkStatusTimeline optional in TestResultCoordinator

dylanbronson:
TestResultCoordinator uses NetworkStatusTimeline to generate DirectMethod reports. 
Now that we will be using TRC for longhaul, NetworkStatusTimeline should be optional, since there will be no networkController. This change makes NetworkStatusTimeline optional. It can only be absent if we're in an online scenario, though

### DIFF
--- a/test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
+++ b/test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
@@ -2,6 +2,7 @@
 namespace TestResultCoordinator.Reports
 {
     using System;
+    using System.IO;
     using System.Threading.Tasks;
     using Microsoft.Azure.Devices.Edge.ModuleUtil;
     using Microsoft.Azure.Devices.Edge.ModuleUtil.NetworkController;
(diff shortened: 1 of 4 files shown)


test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
```diff
@@ -183,11 +184,18 @@ internal TestReportGeneratorFactory(ITestOperationResultStorage storage, Network
...
+            try
+            {
+                return Option.Some(await NetworkStatusTimeline.CreateAsync(
+                    new StoreTestResultCollection<TestOperationResult>(this.Storage.GetStoreFromSource("networkController"), BatchSize),
+                    tolerancePeriod));
+            }
+            catch (InvalidDataException)
+            {
+                return Option.None<NetworkStatusTimeline>();
+            }
```
and-rewsmith, line 198: Letting this fail and then returning None seems like it could lead to some unintended side effects. What if we give the TRC some configuration that could tell it that it shouldn't be dealing with this network status timeline?

dylanbronson:
Unnecessary. Closing

PR #4234, opened 2021-01-15
Add TestMode to TestResultCoordinator and make it able to run without test duration

dylanbronson:
We are adapting TestResultCoordinator to be able to use for Long haul. Therefore, we need to make it so that it can run indefinitely and send reports every period of time rather than running once.

This PR introduces TestMode to the TRC so that we can know if we're in Connectivity or LongHaul mode. 
Depending on that we will use different timings for TRC's reporting service


test/modules/TestResultCoordinator/Services/TestResultReportingService.cs
```diff
@@ -22,25 +22,39 @@ class TestResultReportingService : IHostedService, IDisposable
...
         readonly ITestOperationResultStorage storage;
         readonly TestResultReportingServiceSettings serviceSpecificSettings;
+        readonly TimeSpan sendReportFrequency;
         Timer timer;
 
         public TestResultReportingService(ITestOperationResultStorage storage)
         {
             this.storage = Preconditions.CheckNotNull(storage, nameof(storage));
-            this.delayBeforeWork = Settings.Current.TestStartDelay + Settings.Current.TestDuration + Settings.Current.DurationBeforeVerification;
+            if (TestMode.Connectivity.Equals(Settings.Current.TestMode))
```
vadim-kovalyov, line 31: nit: consider if new switch expression syntax would make it more compact/readable? https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression

test/modules/TestResultCoordinator/TestMode.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft. All rights reserved.
+namespace TestResultCoordinator
+{
+    enum TestMode
```
vadim-kovalyov, line 4: maybe it would make sense to add some comments on what is the difference b/w these two modes.

test/modules/TestResultCoordinator/Services/TestResultReportingService.cs
```diff
@@ -22,25 +22,39 @@ class TestResultReportingService : IHostedService, IDisposable
...
         {
             this.logger.LogInformation("Test Result Reporting Service running.");
 
-            // Specify -1 ms to disable periodic run
             this.timer = new Timer(
                 this.DoWorkAsync,
                 null,
                 this.delayBeforeWork,
-                TimeSpan.FromMilliseconds(-1));
+                this.sendReportFrequency);
```
and-rewsmith, line 57: The sendReportFrequency for the connectivity tests shouldn't be relevant because the test is over after delayBeforeWork elapses.

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -150,11 +163,9 @@ static Settings Create()
 
         public bool OptimizeForPerformance { get; }
 
-        public TimeSpan TestDuration { get; }
-
         public TimeSpan TestStartDelay { get; }
 
-        public TimeSpan DurationBeforeVerification { get; }
+        public TimeSpan SendReportFrequency { get; }
```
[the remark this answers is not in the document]
and-rewsmith: Although after looking at the usage in the case of connectivity, it currently only kicks in after the delayBeforeWork which is when the test finishes.

and-rewsmith (commented):
Looking good, but have a question if we should make some Longhaul specific settings for report send frequency.

test/modules/TestResultCoordinator/Settings.cs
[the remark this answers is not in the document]
dylanbronson: I was originally thinking that it could be useful for connectivity tests, but honestly I don't think it is. If it is ever is necessary in the future we can add it then.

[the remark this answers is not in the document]
dylanbronson: Modified to be longhaul specific setting

and-rewsmith (approved):
Looking good to me

vadim-kovalyov (approved):

test/modules/TestResultCoordinator/Settings.cs
```diff
@@ -75,13 +75,26 @@ class Settings
...
+                        break;
+                    }
+
+                case TestMode.LongHaul:
+                    {
+                        this.LongHaulSpecificSettings = Option.Some(new TestResultCoordinator.LongHaulSpecificSettings()
+                        {
+                            SendReportFrequency = sendReportFrequency.Expect<ArgumentException>(() => throw new ArgumentException("SendReportFrequency must be supplied if in LongHaul test mode"))
+                        });
+                        break;
```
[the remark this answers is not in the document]
dylanbronson: dotnet forces you to

builds/e2e/longhaul.yaml
```diff
@@ -25,6 +25,7 @@ variables:
   #run.linux.arm64v8.docker: 'true/false'
   #longhaul.desiredModulesToRestartCSV: ''
   #longhaul.restartIntervalInMins: ''
+  #longHaul.sendReportFrequency: '$(longHaul.sendReportFrequency)'
```
and-rewsmith, line 28: I think this should be:
```#longHaul.sendReportFrequency: 'xx:xx:xx.x'```

File: scripts/linux/trcE2ETest.sh [hunk 7]
```diff
@@ -50,6 +50,7 @@ function usage() {
     echo ' -listenWorkloadUri                       Customize listen workload socket'
     echo ' -desiredModulesToRestartCSV              CSV string of module names for long haul specifying what modules to restart. If specified, then "restartIntervalInMins" must be specified as well.'
     echo ' -restartIntervalInMins                   Value for long haul specifying how often a random module will restart. If specified, then "desiredModulesToRestartCSV" must be specified as well.'
+    echo ' -sendReportFrequency                     Value for long haul specifying how often TRC will send reports to LogAnalytics.'
```
and-rewsmith, line 53: nit: Looking back, I think we added too many configurable options to the pipelines which we never really change. The downside of this is that we need to maintain a list of variables in each pipeline we create. I feel it could be better to avoid making things configurable at the pipeline level and just rely on defaults in this script, or just hardcode it in the deployment files. 

Even if we still have this configurable from the pipeline level, a default might be a good idea anyways. Your call.

dylanbronson: There's a default within the dotnet code. So if the parameter is not filled in by the pipeline, it will take the default. We can remove the stuff from the yamls if we want. I'm just going to keep it for now. It's pretty simple to remove if it's annoying us later

and-rewsmith: I think that if the parameter is not filled in by the pipeline, it won't take the dotnet default. This is because in our deployment template we give the value a placeholder. If the pipeline doesn't fill in the placeholder I would expect an exception from the dotnet app.

dylanbronson: Yeah you're right

and-rewsmith (approved):

builds/e2e/longhaul.yaml
```diff
@@ -120,6 +123,7 @@ jobs:
           metricsCollector.scrapeFrequencyInSecs: '$(metricsCollector.scrapeFrequencyInSecs)'
           metricsCollector.uploadTarget: '$(metricsCollector.uploadTarget)'
           metricsCollector.hostPlatform: 'linux_amd64_moby'
+          test.testMode: 'LongHaul'
```
and-rewsmith, line 126: If you hardcode this in the deploy.yaml we can eliminate duplication

PR #4238, opened 2021-01-15
upstream/downstream messages

ancaantochi:
Mqtt bridge integration tests: 
- upstream/downstream messages
- settings update
- subscription retry when rejected
- connection failure tests


mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -0,0 +1,418 @@
...
+
+fn setup_brokers<Z, T>(
+    local_authorizer: Z,
+    upstream_authorizer: T,
+) -> (ServerHandle, ServerHandle, BrokerHandle)
+where
+    Z: Authorizer + Send + 'static,
+    T: Authorizer + Send + 'static,
+{
+    let (local_server_handle, _) = setup_local_broker(local_authorizer);
```
and-rewsmith, line 345: nit: might be a good practice to define this broker handle and return it. I see we ignore some of the items currently returned now so we could similarly ignore the local broker handle.

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -0,0 +1,418 @@
...
+        Some(ReceivedPublication{payload, .. }) if payload == Bytes::from("from local")
+    );
+
+    assert_matches!(
+        local_client.publications().recv().await,
+        Some(ReceivedPublication{payload, .. }) if payload == Bytes::from("from upstream")
+    );
+
+    upstream_client.shutdown().await;
+    local_client.shutdown().await;
```
and-rewsmith, line 117: should we also shutdown bridge controller and brokers?

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -0,0 +1,418 @@
...
+    controller_handle
+        .send_update(BridgeControllerUpdate::from_bridge_topic_rules(
+            "$upstream",
+            subs.as_ref(),
+            forwards.as_ref(),
+        ))
+        .unwrap();
+
+    // delay to propagate the update
+    tokio::time::delay_for(Duration::from_secs(2)).await;
```
and-rewsmith, line 180: If we can use signals here instead of a wait that will be better, but I don't think it is possible.

mqtt/mqtt-broker/src/lib.rs
```diff
@@ -25,7 +25,7 @@ mod snapshot;
 mod state_change;
 mod stream;
 mod subscription;
-mod tls;
+pub mod tls;
```
and-rewsmith, line 28: Another option is to move this to the util crate

and-rewsmith (approved):
LGTM with some minor comments

mqtt/mqtt-bridge/tests/integration.rs
dmolokanov: @vadim-kovalyov Do you remember implementing awaiting for a signal?

dmolokanov (approved):

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -0,0 +1,418 @@
...
+
+async fn setup_bridge_controller(
+    local_address: String,
+    upstream_address: String,
+    subs: Vec<Direction>,
+) -> BridgeControllerHandle {
+    let credentials = Credentials::PlainText(AuthenticationSettings::new(
+        "bridge".into(),
+        "pass".into(),
+        "bridge".into(),
```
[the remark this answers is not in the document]
ancaantochi: because it needs to create the upstream client without iotedged calls so I added the function to provide plain credentials and the trust-bundle

and-rewsmith (approved):

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -0,0 +1,438 @@
...
+    local_client.subscriptions().recv().await;
+
+    let mut upstream_client = TestClientBuilder::new(upstream_server_handle.address())
+        .with_client_id(ClientId::IdWithExistingSession("upstream_client".into()))
+        .build();
+
+    upstream_client
+        .subscribe("upstream/temp/#", QoS::AtLeastOnce)
+        .await;
+    // wait to receive subscription ack
```
vadim-kovalyov, line 152: nit: IMO it is more readable to have a newline before inline comments.

mqtt/mqtt-bridge/tests/integration.rs
vadim-kovalyov: I implemented awaiting for an authorizer updates. This is different.

vadim-kovalyov (commented):
Sorry, I'm late to the party :)

PR #4240, opened 2021-01-15
Retry getting device information on startup

massand:
When `aziot-edged` service starts up, it auto-starts the `aziot-identityd` service. The provisioning process is asynchronous in `aziot-identityd` and should be treated as such by `aziot-edged`.
This fix ensures the `aziot-edged` service loops on a fixed retry interval, if there is an error returned during the `get_device_info` call.


edgelet/aziot-edged/src/lib.rs
```diff
@@ -254,56 +257,65 @@ where
...
-                &runtime,
-                cfg.clone(),
-                make_shutdown_signal(),
-                &mut tokio_runtime,
-            )?;
-
-            if code != StartApiReturnStatus::Restart {
-                break;
-            }
+                    InitializeErrorReason::DpsProvisioningClient,
```
arsing, line 263: While you're at it, can you fix this error reason too?

daprilik (dismissed):

arsing (approved):

PR #4242, opened 2021-01-15
Update long haul deployment to use TRC

and-rewsmith:
We are now switching to use the TRC in longhaul. Since the connectivity test script contained logic relevant to the TRC, I discussed with Bear and David and we thought it was appropriate to move the logic for longhaul into the connectivity script and rename it. 

 This PR gates the use of the TRC behind a feature flag because there are TRC code changes necessary for reporting to run correctly. Once we make the necessary changes to the testdashboard and the TRC we can remove the feature flag and remove the longhaul logic from runE2ETest.sh.


e2e_deployment_files/connectivity_deployment.template.json
```diff
@@ -513,6 +531,12 @@
               },
               "TransportType": {
                 "value": "Mqtt"
+              },
+              "TwinUpdateSize": {
+                "value": "<TwinUpdateSize>"
+              },
+              "TwinUpdateFrequency": {
+                "value": "<TwinUpdateFrequency>"
```
and-rewsmith, line 539: I added these params to connectivity as we might want to configure them in the future.

scripts/linux/runE2ETest.sh
```diff
@@ -506,7 +506,6 @@ function process_args() {
 }
 
 function get_hash() {
-    # TODO: testHelper.sh needs to be shared across build pipelines
```
and-rewsmith, line 509: This script doesn't exist so I removed all references to it

scripts/linux/trcE2ETest.sh
```diff
@@ -77,18 +86,25 @@ function get_artifact_file() {
 
     local filter
     case "$fileType" in
-        'aziot_edge' ) filter='aziot-edge_*.deb';;
-        'aziot_is' ) filter='aziot-identity-service_*.deb';;
+        'aziot_edge' ) filter='*/aziot-edge_*.deb';;
+        'aziot_is' ) filter='*/aziot-identity-service_*.deb';;
```
and-rewsmith, line 90: This was a bug from the iiot merge

dylanbronson (approved):
This looks good to me as long as it deploys correctly!

and-rewsmith:
Discussed with Dylan the current status. Deploys everything but there is a problem with name resolution in the docker network, Since this is behind a feature flag I think we can merge now and fix in a separate PR.

PR #4244, opened 2021-01-15
Config changes for content trust with cert service

ggjjj:
Config.yaml previously had a mapping of registry hostname and root CA path. With new runtime, it would be a mapping of registry hostname and the cert ID of its corresponding root CA. 
In certd.toml, the mapping of Cert of the root CA and its file path must be configured.


edgelet/edgelet-docker/src/notary.rs
```diff
@@ -17,54 +17,94 @@ use crate::{Error, ErrorKind};
...
-        .into());
+    if cert_buf.is_empty() {
+        return Err(
+            ErrorKind::InitializeNotary("root ca pem string content is empty".to_owned()).into(),
+        );
     }
 
-    // Create notary home directory
+    // Directory structure example
+    // home directory : /home
```
arsing, line 34: I suggest using the actual homedir here, `/var/lib/aziot/edged`

edgelet/edgelet-docker/src/settings.rs
```diff
@@ -493,25 +492,11 @@ mod tests {
         {
             assert_eq!(
                 content_trust_map.get("contoso1.azurcr.io"),
-                Some(&std::path::PathBuf::from("/path/to/root_ca_contoso1.crt"))
+                Some(&String::from("content-trust-contoso1.azurecr.io"))
```
arsing, line 495: `Some("content-trust-contoso1.azurecr.io")`

Same for the statement below.

If it complains about `Option<&String> cannot be compared with Option<&str>`, then do:

```rust
assert_eq!(
    content_trust.get("contoso1.azurecr.io").as_deref(),
    Some("content-trust-contoso1.azurecr.io")
);
```

arsing (changes requested):

edgelet/edgelet-docker/src/settings.rs
arsing: Oh, actually instead of `.as_deref()` it should be `.map(AsRef::as_ref)`

`.as_deref()` would convert `&Option<String>` to `Option<&str>`, whereas you want to convert `Option<&String>` to `Option<&str>`

arsing (approved):

PR #4245, opened 2021-01-15
Add DirectMethodLongHaulReport and separate DM reports

dylanbronson:
Now that we are moving long haul to be handled by the TestResultCoordinator and not the analyzer, we need to handle DM reports differently. 

With the long haul tests, we can get a status code of "0". This is not a concept in the connectivity tests. 
This PR makes a new type of report, DirectMethodLongHaulReport and turns the old DirectMethodReport into DirectMethodConnectivityReport. This was easier and cleaner than cramming specific long haul logic that doesn't fit well into the already-complicated DirectMethodReportGenerator. This is a much clearer separation, as the reports are different, and the generation is much different as well.

We can also simplify the report generation, as we have no need for a timeline, since we assume the connection is always on.

The new DMLongHaul logic is identical to the logic that was in the analyzer: if we get anything that's not a 200 or a 0 we fail the test. We have some tolerance for 0's based on the percentage of 0's (it must be < 1%). We now have receiver results too, so we use those just as a sanity check to make sure there are as many receiver results as there are success results.


test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/NetworkStatusTimelineTest.cs
```diff
@@ -6,6 +6,7 @@ namespace Modules.Test.TestResultCoordinator.Reports.DirectMethod
     using System.Linq;
     using global::TestResultCoordinator.Reports;
     using global::TestResultCoordinator.Reports.DirectMethod;
+    using global::TestResultCoordinator.Reports.DirectMethod.Connectivity;
```
and-rewsmith, line 9: Do we need extra tests for the longhaul direct methods or no?

test/modules/Modules.Test/TestResultCoordinator/TestReportUtilTest.cs
```diff
@@ -308,16 +352,41 @@ public void ParseReportMetadataList_ParseDirectMethodTestReportMetadataWithoutRe
...
             Assert.Equal(TestOperationResultType.DirectMethod, reportMetadata.TestOperationResultType);
-            Assert.Equal(TestReportType.DirectMethodReport, reportMetadata.TestReportType);
+            Assert.Equal(TestReportType.DirectMethodConnectivityReport, reportMetadata.TestReportType);
             Assert.Equal("directMethodSender1.send", reportMetadata.SenderSource);
             Assert.False(reportMetadata.ReceiverSource.HasValue);
             Assert.Equal(new TimeSpan(0, 0, 0, 0, 5), reportMetadata.TolerancePeriod);
         }
 
+        [Fact]
+        public void ParseReportMetadataList_ParseDirectMethodLongHaulTestReportMetadataWithoutReceiverSource()
```
and-rewsmith, line 366: Can you explain the context around these tests? In what scenario is a report made without the receiver? Is it when the direct method is initially sent?

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public string SenderSource { get; }
+        public Option<string> ReceiverSource { get; }
+        public ulong SenderSuccesses { get; }
+        public Option<ulong> ReceiverSuccesses { get; }
+        public ulong StatusCodeZero { get; }
+        public ulong Unknown { get; }
+
+        public override string Title => this.ReceiverSource.HasValue ?
+            $"DirectMethod LongHaul Report for [{this.SenderSource}] and [{this.ReceiverSource.OrDefault()}] ({this.ResultType})" : $"DirectMethod Report for [{this.SenderSource}] ({this.ResultType})";
```
and-rewsmith, line 42: What is the OrDefault() on receiver source resolve to?

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
```diff
@@ -0,0 +1,61 @@
...
+            this.StatusCodeZero = statusCodeZero;
+            this.Unknown = unknown;
+        }
+
+        public string SenderSource { get; }
+        public Option<string> ReceiverSource { get; }
+        public ulong SenderSuccesses { get; }
+        public Option<ulong> ReceiverSuccesses { get; }
+        public ulong StatusCodeZero { get; }
+        public ulong Unknown { get; }
```
and-rewsmith, line 39: What does unknown come from?

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public override string Title => this.ReceiverSource.HasValue ?
+            $"DirectMethod LongHaul Report for [{this.SenderSource}] and [{this.ReceiverSource.OrDefault()}] ({this.ResultType})" : $"DirectMethod Report for [{this.SenderSource}] ({this.ResultType})";
+
+        public override bool IsPassed => this.IsPassedHelper();
+
+        public bool IsPassedHelper()
+        {
+            bool noUnknownResults = this.Unknown > 0 ? false : true;
```
and-rewsmith, line 48: I think it can be more readable to fail immediately if there are unknown, then remove the unknown results from the below logic.

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReportGenerator.cs
```diff
@@ -0,0 +1,113 @@
...
+                        break;
+                }
+            }
+
+            await this.ReceiverTestResults.ForEachAsync(async r =>
+            {
+                ulong receiverResults = 0;
+                while (await r.MoveNextAsync())
+                {
+                    // ReceiverSource will always be there if ReceiverTestResults is so it's safe to put OrDefault
```
and-rewsmith, line 85: What default will this give if we are wrong? It might be better to try to unwrap and throw an exception if none.

and-rewsmith (commented):
Looking good, just a few questions and minor comments. Have we tested this end to end yet?

test/modules/Modules.Test/TestResultCoordinator/TestReportUtilTest.cs
dylanbronson: EdgeAgent ping tests is the only one I know with no receiver

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
dylanbronson: I believe empty string if receiver is not there, which i guess it pretty awkward looking. I was copy pasting so I can make it better

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
dylanbronson: Any failure (500 most notably)

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
dylanbronson: oh nvm it's ok - it'll only run that first part if receiver source has a value. So the default portion of ordefault can never run

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
```diff
@@ -0,0 +1,394 @@
...
+
+            var senderStoreData = GetSenderStoreData(senderSource, resultType, senderStoreValues, statusCodes, timestamps, guid);
+            for (int i = 0; i < senderStoreData.Count; i += batchSize)
+            {
+                int startingOffset = i;
+                mockSenderStore.Setup(s => s.GetBatch(startingOffset, batchSize)).ReturnsAsync(senderStoreData.Skip(startingOffset).Take(batchSize));
+            }
+
+            var receiverStoreData = GetReceiverStoreData(receiverSource, resultType, receiverStoreValues, timestamps, guid);
+            for (int j = 0; j < senderStoreData.Count; j += batchSize)
```
and-rewsmith, line 251: should this be ```j < senderStoreData.count``` or ```j < receiverStoreData.count```

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
```diff
@@ -0,0 +1,394 @@
...
+                receiverResults,
+                resultType);
+
+            Guid guid = Guid.NewGuid();
+
+            var senderStoreData = GetSenderStoreData(senderSource, resultType, senderStoreValues, statusCodes, timestamps, guid);
+            for (int i = 0; i < senderStoreData.Count; i += batchSize)
+            {
+                int startingOffset = i;
+                mockSenderStore.Setup(s => s.GetBatch(startingOffset, batchSize)).ReturnsAsync(senderStoreData.Skip(startingOffset).Take(batchSize));
```
and-rewsmith, line 247: Do we really use all these setups or is this just for safety?

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
```diff
@@ -0,0 +1,394 @@
...
+    using global::TestResultCoordinator.Reports.DirectMethod.LongHaul;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil.TestResults;
+    using Microsoft.Azure.Devices.Edge.Storage;
+    using Microsoft.Azure.Devices.Edge.Util;
+    using Moq;
+    using Newtonsoft.Json;
+    using Xunit;
+
+    public class DirectMethodLongHaulReportGeneratorTest
```
and-rewsmith, line 19: nit: I think it could increase readability to order the tests in the order of the DirectMethodLongHaulReportGenerator args.

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
```diff
@@ -0,0 +1,394 @@
...
+                    TestDescription,
+                    Guid.NewGuid().ToString(),
+                    "senderSource",
+                    mockSenderResults.Object,
+                    Option.Some("receiverSource"),
+                    Option.None<ITestResultCollection<TestOperationResult>>(),
+                    "resultType1"));
+
+            Assert.Equal("Provide both receiverSource and receiverTestResults or neither.", ex.Message);
+        }
```
and-rewsmith, line 90: Is this a duplicate test?

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
```diff
@@ -0,0 +1,394 @@
...
+
+            var mockSenderStore = new Mock<ISequentialStore<TestOperationResult>>();
+            var senderResults = new StoreTestResultCollection<TestOperationResult>(mockSenderStore.Object, batchSize);
+            var mockReceiverStore = new Mock<ISequentialStore<TestOperationResult>>();
+            var receiverResults = Option.Some<ITestResultCollection<TestOperationResult>>(
+                new StoreTestResultCollection<TestOperationResult>(mockReceiverStore.Object, batchSize));
+
+            var reportGenerator = new DirectMethodLongHaulReportGenerator(
+                TestDescription,
+                Guid.NewGuid().ToString(),
```
and-rewsmith, line 39: will it throw if this guid isn't provided?

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
```diff
@@ -0,0 +1,394 @@
...
+                senderSource,
+                senderResults,
+                Option.Some(receiverSource),
+                receiverResults,
+                resultType);
+
+            Guid guid = Guid.NewGuid();
+
+            var senderStoreData = GetSenderStoreData(senderSource, resultType, senderStoreValues, statusCodes, timestamps, guid);
```
and-rewsmith, line 243: nit: I think the styleguide says to specify type when it is ambiguous

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportData.cs
```diff
@@ -0,0 +1,73 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Net;
+
+    class DirectMethodLongHaulReportData
+    {
+        public static IEnumerable<object[]> GetCreateReportData =>
+        new List<object[]>
+        {
+            new object[]
```
and-rewsmith, line 14: nit: iirc you are just following the existing pattern, but is there a reason why we didn't make some sort of abstraction for this test data?

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
dylanbronson: We actually use them when creating reports

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
dylanbronson: good catch. fixed

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
dylanbronson: yes. that test exists too - TestConstructorThrowsWhenTrackingIdIsNotProvided

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
and-rewsmith: What I meant was that it seems like it should take elements in batch sizes starting from each index, which I would guess leads to duplicates. So I was assuming it doesn't use all these setup calls. Regardless I think it is fine.

and-rewsmith (approved):
LGTM

PR #4247, opened 2021-01-16
Added unsubscribe through nested levels on client-disconnect 

vipeller:
The purpose of this change is that in nested scenario, when a client at child level disconnects, it unsubscribes from IotHub topics. The un-subscriptions bubble up to parents and it turn on the idle timer (which functionality was already implemented). After some idle time the parents disconnects the client at their level, too.
This play is needed as not dedicated connect/disconnect message is travelling between parent and child edge-devices to keep the interface more simple.

Important parts of the changes:

- The root of the change is at ConnectionHandler.cs, which calls a newly created 'RemoveSubscriptions' method on the DeviceListener when a client disconnects. This new method needed to be added at several places (because it is an interface change), and also needed to bubble up to the SubscriptionProcessor (through RoutingEdgehub). This method is added to avoid calling the existing 'RemoveSubscription' several times. Its main implementation is in SubscriptionProcessorBase().

- In SubscriptionProcessor.cs not all the subscription had related implementation for 'remove', these are added there. Also, these removals not forwarded to ICouldProxy in every case (where necessary), so 2 new methods have been added (for twin and c2d subscriptions)

- The CloudProxy.cs already had an implementation for Idle Timeout. This timer was turned off if a client made a subscription, so it was expected that IoTHub would send something. Once the timer was off, the idle timeout could never happen. This part has been changed, so now it monitors if there is an active subscription, and if not, it restarts the idle timer.

- In EdgeHub-Bridge there is a code which tries to figure out the client subscriptions according to the broker notification. THe notifications work sort of ok when there is no 'in-name-of' subscription (so a client subscribes in its own name). In case of nested scenarion, edgeHub subscribes in name of every client and decoding the related notifications is quite painful. There is a new code implemented for that in SubscriptionChangeHandler.cs. The modification was needed because the previous implementation supported only subscriptions but not unsubscribe - which was adequate that time as the rest of the system did not support unsubscribe.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,8 +91,86 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
-            // TODO: the following solution is to unblock the case when in a nested scenario a child edgehub
-            // subscribes in the name of a device/module. However changes needed to support unsubscribe
+            var id1 = match.Groups["id1"];
+            var id2 = match.Groups["id2"];
+
+            // For indirect clients the subscription change will be reported through edgeHub. That is a
+            // slower method to reconcile so we handle the direct case separately with a faster method
+            if (id2.Success && string.Equals(id2.Value, Constants.EdgeHubModuleId))
+            {
+                await this.HandleIndirectChanges(subscriptionList);
```
varunpuranik, line 101: Nit: suggest renaming to HandleNestedSubChanges / HandleDirectSubChanges..

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -97,34 +180,94 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+                        else
+                        {
+                            Events.CouldNotObtainListenerForSubscription(subscription.ToString(), connection.Id);
+                        }
+                    }
+                }
+            }
+        }
+
+        static bool IsMatchWithIds(Match match, Group id1, Group id2)
```
varunpuranik, line 257: I think I follow.. but can you add a comment on what this does.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,8 +91,86 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+            }
+            else
+            {
+                await this.HandleDirectChanges(id1, id2, subscriptionList);
+            }
+
+            return true;
+        }
+
+        async Task HandleDirectChanges(Group id1, Group id2, List<string> subscriptionList)
```
varunpuranik, line 111: This id1 and id2 makes this hard to follow. Can we expand these before we pass them around..

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
```diff
@@ -162,6 +162,38 @@ public bool RemoveSubscription(string id, DeviceSubscription deviceSubscription)
...
+                {
+                    var subscriptions = d.Subscriptions.Keys.ToArray();
+                    foreach (var deviceSubscription in subscriptions)
+                    {
+                        d.Subscriptions.AddOrUpdate(
+                            deviceSubscription,
+                            false,
+                            (_, old) =>
+                            {
+                                if (old != false)
```
ancaantochi, line 184: if (old)

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
```diff
@@ -162,6 +162,38 @@ public bool RemoveSubscription(string id, DeviceSubscription deviceSubscription)
...
 
+        public IReadOnlyCollection<DeviceSubscription> RemoveSubscriptions(string id)
+        {
+            if (!this.devices.TryGetValue(Preconditions.CheckNonWhiteSpace(id, nameof(id)), out ConnectedDevice device))
+            {
+                throw new ArgumentException($"A connection for {id} not found.");
+            }
+
+            var toRemove = new List<DeviceSubscription>();
+            device.DeviceConnection.Filter(d => d.IsActive)
```
ancaantochi, line 173: the update is not thread safe, it looks like it is possible to remove a subscription if in the meantime was added from a different thread. I don't think there is a lock on a higher level to make sure it's thread safe

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
```diff
@@ -21,6 +22,15 @@ namespace Microsoft.Azure.Devices.Edge.Hub.CloudProxy
 
     class CloudProxy : ICloudProxy
     {
+        [Flags]
+        enum Subscriptions
```
ancaantochi, line 26: Subscriptions enum and the add/remove operations could be encapsulated in a class

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,45 +91,211 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+                                         dev => this.identityProvider.Create(dev, mod),
+                                         () => throw new ArgumentException("Invalid topic structure for subscriptions - Module name matched but no Device name")),
+                               () => deviceId.Match(
+                                         dev => this.identityProvider.Create(dev),
+                                         () => throw new ArgumentException("Invalid topic structure for subscriptions - no Device name matched")));
+
+            var listener = default(IDeviceListener);
+            var maybeListener = await this.connectionRegistry.GetDeviceListenerAsync(identity, true);
+
+            if (maybeListener.HasValue)
```
ancaantochi, line 120: you can use GetorElse

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,45 +91,211 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+            {
+                listener = maybeListener.Expect(() => new Exception($"No device listener found for {identity.Id}"));
+
+                foreach (var subscriptionPattern in this.subscriptionPatterns)
+                {
+                    var subscribes = false;
+
+                    foreach (var subscription in subscriptionList)
```
ancaantochi, line 128: nit: extract a method

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,45 +91,211 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+
+                            if (IsMatchingIds(subscribedDevId, subscribedModId, deviceId, moduleId))
+                            {
+                                subscribes = true;
+                                break;
+                            }
+                        }
+                    }
+
+                    try
```
ancaantochi, line 144: try - catch could be moved inside AddOrRemoveSubscription

ancaantochi:
Still looking at SubscriptionChangehandler

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,45 +91,211 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+            var identity = moduleId.Match(
+                               mod => deviceId.Match(
+                                         dev => this.identityProvider.Create(dev, mod),
+                                         () => throw new ArgumentException("Invalid topic structure for subscriptions - Module name matched but no Device name")),
+                               () => deviceId.Match(
+                                         dev => this.identityProvider.Create(dev),
+                                         () => throw new ArgumentException("Invalid topic structure for subscriptions - no Device name matched")));
+
+            var listener = default(IDeviceListener);
+            var maybeListener = await this.connectionRegistry.GetDeviceListenerAsync(identity, true);
```
ancaantochi, line 118: nit: rename GetOrCreateDeviceListeberAsync

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,45 +91,211 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+                        var maybeListener = await this.connectionRegistry.GetDeviceListenerAsync(identity, false);
                         if (maybeListener.HasValue)
                         {
                             try
                             {
                                 var listener = maybeListener.Expect(() => new Exception($"No device listener found for {identity.Id}"));
-                                await AddOrRemoveSubscription(listener, true, subscriptionPattern.Subscrition);
+                                await AddOrRemoveSubscription(listener, true, subscriptionPattern.Subscription);
+
+                                var subscriptions = default(HashSet<DeviceSubscription>);
```
ancaantochi, line 208: var subscriptions = new HashSet<DeviceSubscription>();

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,45 +91,211 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+        //     from all iothub related topics.
+        // To solve 1), the connectionRegistry.GetDeviceListenerAsync() creates the device if it has never
+        // been seen before.
+        // To solve 2), we need to take all known devices and unsubscribe from all topics not listed in the
+        // current notification.
+        async Task HandleNestedSubscriptionChanges(List<string> subscriptionList)
+        {
+            // Collect all added subscriptions to this dictionary for every client. At the end we are going to
+            // unsubscribe from everything not in this collection
+            var affectedClients = new Dictionary<IIdentity, HashSet<DeviceSubscription>>();
```
ancaantochi, line 178: would extract collecting affectedClient in a method

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,45 +91,211 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+            // so when a client unsubscribes, that is only a shorter list of subscriptions sent.
+            // The information we want to gain is that what the unsubscriptions are, and the answer is that what we have not seen
+            // previously as subscribed, that is unsubscribed. E.g. if a client subscribed to 'methods' only, then we say that
+            // it unsubscribed from twin responses, c2d, m2m, etc. The underlaying layers take care of the optimization doing
+            // nothing when unsubscribed twice from a topic.
+            // As a first step we need all nested connections, because it is possible that the subscribtion list above did not
+            // cover all clients. It can happen when a client had only a single subscription before, but now it unsubscribed.
+            // In that case that client will not be contained by the subscription event sent by the broker, because it has no
+            // subscription at all.
+            var nestConnections = await this.connectionRegistry.GetNestedConnectionsAsync();
```
ancaantochi, line 241: this part till end could also be extracted in a separate method

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
```diff
@@ -86,45 +91,211 @@ public async Task<bool> HandleAsync(MqttPublishInfo publishInfo)
...
+                    subscriptions = new HashSet<DeviceSubscription>();
+                }
+
+                var maybeListener = await this.connectionRegistry.GetDeviceListenerAsync(connection, false);
+
+                foreach (var subscription in this.allSubscriptions)
+                {
+                    if (!subscriptions.Contains(subscription))
+                    {
+                        if (maybeListener.HasValue)
```
ancaantochi, line 260: check listener exists before iterating

ancaantochi:
added a few comments to SubscriptionChangeHandler

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
vipeller: This will be a bit long, sorry:

let's say that there is a lock that Remove/AddSubscription and Subscriptions use. If there is a parallel call, that let's say A) thread turns on C2D and B) thread turns off, then two things can happen:

1) A thread gets the lock first, it turns on C2D. Then B turns it off
2) B thread turns off everything, then A turns C2D on.

At this level (RemoveSubscriptions), without the lock, the same two cases can happen:
1) A thread turns C2D on before B thread gets there in the loop, so at the end, everything will be turned off
2) A thread turns off C2D after B thread got there in the loop, so at the end C2D will be turned on.

The problem in both cases (with or without lock) is that the toRemove array may have an item which is at the end is not turned off. But a local locking will not solve this, because even if we ensure that at the time of exiting the function toRemove represents the current state, right after the return something can turn a subscription back.

This problem existed before RemoveSubscriptions() was added:

Let's say that a thread removed C2D using the old ConnectionManager.RemoveSubscription() method. In the code it happens in SubscriptionProcessorBase::RemoveSubscription() method. Now in ConnectionManager the subscription is turned off.
Then let's say that another thread turned this back using AddSubscription(), so now C2D is on according to ConnectionManager.
What we cannot guarantee that in SubscriptionProcessorBase, the subsequent HandleSubscription() call will be in the same order. It can happen that for the second thread it gets executed sooner, so it will subscribe at IoTHub for C2D, and then the first thread unsubscribes. At the end, ConnectionManager and IoTHub will be out of sync.

Adding RemoveSubscriptions() did not bring it a new problem just made an existing one more visible. I was not too concerned about this problem for two reasons:
- it is not too realistic having sudden subscriptions/unsubscriptions on the same topic from a single client. (although it can happen)
- RemoveSubscriptions() is added for more convinient support when a client disconnects and we want to remove all subscriptions. Because of this, in real life it is hard to receive a subscription in the middle of RemoveSubscriptions(), as the client disconnected.

It still might worth to fix the original problem, but I would create a ticket for that instead and leave this part of the code as is.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: That sounds more correct, changed.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: I could use ForEachAsync (as there is an async call in the block), but for me it looks less readable:
```
await maybeListener.ForEachAsync(
  async listener =>  
  {  
    ...  
    try  
    {  
      await AddOrRemoveSubscription(listener, subscribes, subscriptionPattern.Subscription);  
    }  
    ...  
  }  
},  
() =>  
{  
  Events.CouldNotObtainListener(identity.Id);  
  return Task.CompletedTask;  
});  
```

It adds an extra await at the beginning and a Task.CompletedTask at the end

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: It needs to print an error message for every subscription, so if there is no listener, and there is C2D and Methods subscription, we would see an error for both C2D and Methods. 
I can do something like this
```
                foreach (var subscription in this.allSubscriptions)
                {
                    if (!subscriptions.Contains(subscription))
                    {
                        await maybeListener.Match(
                            async listener => await AddOrRemoveSubscription(listener, false, subscription),
                            () =>
                            {
                                Events.CouldNotObtainListenerForSubscription(subscription.ToString(), connection.Id);
                                return Task.CompletedTask;
                            });
                    }
                }
```

which looks strange because of the Task.CompletedTask

vipeller: I've changed Events.CouldNotObtainListenerForSubscription() to return Task.Completed, so now the code can be 

```
foreach (var subscription in this.allSubscriptions)
                {
                    if (!subscriptions.Contains(subscription))
                    {
                        await maybeListener.Match(
                            listener => AddOrRemoveSubscription(listener, false, subscription),
                            () => Events.CouldNotObtainListenerForSubscription(subscription.ToString(), connection.Id));
                    }
                }
```

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
```diff
@@ -162,6 +162,38 @@ public bool RemoveSubscription(string id, DeviceSubscription deviceSubscription)
...
+                .ForEach(d =>
+                {
+                    var subscriptions = d.Subscriptions.Keys.ToArray();
+                    foreach (var deviceSubscription in subscriptions)
```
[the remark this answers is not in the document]
vipeller: changed

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
vipeller: changed

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudProxy.cs
vipeller: moved logical operation into a class

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: moved into the method

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: I turned Events.CloudNotObtainListener into an async call and using with async Match

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: moved the declaration into TryGetValue() as inline declaration

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: extracted

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: extracted

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/SubscriptionChangeHandler.cs
vipeller: extracted

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
ancaantochi: As discussed, we should open a bug and fix it separately

ancaantochi (approved):

PR #4249, opened 2021-01-16
[mqtt] Retry non-iothub subscription when rejected by server

dmolokanov:
When edge deployed on several levels in nested scenarios there can be a lag between all rules are applied. In this case, lower level subscriptions can be rejected by the upper level. The lower level should keep retrying until it is subscribed.


mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -128,10 +129,14 @@ where
...
         let remote_topic_mappers_updates = TopicMapperUpdates::new(topic_filters);
 
+        let (retry_send, retry_recv) = mpsc::unbounded_channel();
+
         let rpc_subscriptions = RpcSubscriptions::default();
         let rpc = RemoteRpcMqttEventHandler::new(rpc_subscriptions.clone(), local_pump.handle());
-        let messages =
+        let mut messages =
             StoreMqttEventHandler::new(local_store, remote_topic_mappers_updates.clone());
+        messages.set_retry_sub_sender(retry_send);
```
vadim-kovalyov, line 138: nit: is it possible to just add retry_sender to ctor?

mqtt/mqtt-bridge/src/pump/mod.rs
```diff
@@ -16,7 +16,7 @@ use futures_util::{
     join, pin_mut,
 };
 use mockall::automock;
-use parking_lot::Mutex;
+use parking_lot::RwLock;
```
vadim-kovalyov, line 19: just curious why this change in this PR and not separate?

vadim-kovalyov (approved):
LGTM!

mqtt/mqtt-bridge/src/pump/mod.rs
dmolokanov: I started to use more read operations with changes in this PR: reading when retry timeout expired. so they can come together

mqtt/mqtt-bridge/src/pump/builder.rs
dmolokanov: not every pump needs to retry subscriptions thus `retry_send` can be optional. not sure if optional argument is good for ctor

vadim-kovalyov: I see. The ideal solution would be builder :) but maybe overkill here. So i'm fine with "setter"

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -76,19 +98,28 @@ impl<S> StoreMqttEventHandler<S> {
...
         if self.topic_mappers.remove(sub).is_none() {
             warn!("unexpected subscription/rejected ack for {}", sub);
         };
     }
+
+    fn handle_rejected(&mut self, sub: SubscribeTo) {
+        self.topic_mappers.remove(&sub.topic_filter);
+        if let Some(sender) = &mut self.retry_sub_send {
+            if sender.send(sub).is_err() {
+                warn!("unable to schedule subscription retry. channel closed");
```
nyanzebra, line 119: Can we log the err too?

mqtt/mqtt-bridge/src/pump/builder.rs
nyanzebra: There is a macro trait for builder pattern :) https://github.com/colin-kiegel/rust-derive-builder

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -150,6 +181,32 @@ where
     }
 }
 
+pub async fn retry_subscriptions(
+    retries: UnboundedReceiver<SubscribeTo>,
```
nyanzebra, line 185: Why are we choosing unbounded?

mqtt/mqtt-bridge/src/messages.rs
dmolokanov: in this case the error is the `sub` itself. the only way error occurred is when the channel is closed https://docs.rs/tokio/1.0.2/tokio/sync/mpsc/struct.UnboundedSender.html#method.send

mqtt/mqtt-bridge/src/pump/builder.rs
dmolokanov: well, some time ago we decided not to use too many derive macros: it makes code more difficult to read and navigate and it is super easy to implement it manually.

we write code once and read it many times...

mqtt/mqtt-bridge/src/messages.rs
dmolokanov: no specific reason. we could use either of the channels in this case

nyanzebra: Are we expecting a lot of simultaneous retries? As unbounded will take up more memory:

```
Note that the amount of available system memory is an implicit bound to the channel. Using an unbounded channel has the ability of causing the process to run out of memory. In this case, the process will be aborted.
```

PR #4254, opened 2021-01-20
[k8s] Allow alphanumerics to start K8s object names.

darobs:
An automated system created a PVC with a GUID name,  It should be valid, but the GUID started with a number, and the number was stripped from the PVC name when doing a volume mount, leaving the pod pending.

This change allows DNS labels (object names) to start with alphanumerics.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubeUtils.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubeUtils.cs
@@ -86,13 +86,20 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
             return output.ToString();
         }
 
+        /// DNS label (as per RFC 1035
+        public static string SanitizeDNSValue(string name) => SanitizeDNSValue(name, IsAlpha);
+
         // DNS label (as per RFC 1035/1123)
-        public static string SanitizeDNSValue(string name)
+        public static string SanitizeDNSLabel(string name) => SanitizeDNSValue(name, IsAlphaNumeric);
+
+        private static string SanitizeDNSValue(string name, Func<char, bool> startCriteria)
         {
-            // The name returned from here must conform to following rules (as per RFC 1035):
+            // The name returned from here must conform to following rules (as per RFC 1035/1123):
             //  - length must be <= 63 characters
             //  - must be all lower case alphanumeric characters or '-'
-            //  - must start and end with an alphanumeric character
+            //  - (RFC 1123) must start and end with an alphanumeric character
+            //  - (RFC 1035) must start with an alphabet
+            //  - must end with an alphanumeric character
             if (string.IsNullOrEmpty(name))
             {
                 throw new InvalidKubernetesNameException($"DNS Name '{name}' is null or empty");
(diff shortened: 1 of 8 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubeUtils.cs
```diff
@@ -86,13 +86,19 @@ public static string SanitizeK8sValue(string name)
...
 
-        // DNS label (as per RFC 1035)
-        public static string SanitizeDNSValue(string name)
+        /// DNS label (as per RFC 1035
+        public static string SanitizeDNSValue(string name) => SanitizeDNSValue(name, IsAlpha);
+
+        // DNS label (as per RFC 1035/1123)
+        public static string SanitizeDNSLabel(string name) => SanitizeDNSValue(name, IsAlphaNumeric);
+
+        private static string SanitizeDNSValue(string name, Func<char, bool> startCriteria)
```
zasherif, line 95: Is system.uri.checkhostname sufficient? https://docs.microsoft.com/en-us/dotnet/api/system.uri.checkhostname?redirectedfrom=MSDN&view=net-5.0

darobs: Back when we wrote this, we debated on using these tools, but we had distinct requirements, specifically, we needed to drop the "$" from the system modules, "$edgeHub" -> edgeHub, so sanitizing rather than checking made more sense.

zasherif: Okay. Thank you!

edgelet/edgelet-utils/src/lib.rs
```diff
@@ -190,7 +202,41 @@ mod tests {
             sanitize_dns_label("$a23456789-123456789-123456789-123456789-123456789-123456789-1234")
```
ggjjj, line 202: By this any chars $, *, \n, etc will be removed and passed with the sanitized output that can be used as object names?

edgelet/edgelet-utils/src/lib.rs
```diff
@@ -59,7 +59,19 @@ const DNS_MAX_SIZE: usize = 63;
...
 ///  - must end with an alphanumeric character
 pub fn sanitize_dns_label(name: &str) -> String {
-    name.trim_start_matches(|c: char| !c.is_ascii_alphabetic())
+    sanitize_dns_label_with(name, char::is_ascii_alphabetic)
+}
+
+/// The name returned from here must conform to following rules (as per RFC 1035/1123):
+///  - length must be <= 63 characters
+///  - must be all lower case alphanumeric characters or '-'
+///  - must start and end with an alphanumeric character
```
ggjjj, line 68: do we need to catch any exception if the input does not adhere this standard?

edgelet/edgelet-utils/src/lib.rs
darobs: Yes, the object names in k8s have some restrictions, and our module names have different restrictions. Rather than rejecting all module names that don't match allowed k8s names, we did what we do for Docker module names, strip out the offending characters, or "sanitize" them. This function was primarily designed to convert "$edgeAgent" to "edgeagent", so sanitizing makes more sense than rejecting names. 

But there were other places where we derive an object name from module settings, and we used the same function, but other objects likes PVCs are less restrictive in their names. We need to sanitize this names differently.

edgelet/edgelet-utils/src/lib.rs
darobs: This function should create a conforming string or an empty string.  Some times empty strings are OK, like in label values. but when empty string are not they're caught in this function: edgelet/edgelet-kube/src/convert/mod.rs:27

ggjjj (approved):

PR #4266, opened 2021-01-21
Generic mqtt test module: Integrate into build pipeline

and-rewsmith:
Adding build logic for the mqtt test module. I added some base images for arm, which I don't think we will need to update at all so I pushed some base images with the tag below. If we need to start updating these we should conform with the branch-consistent versioning scheme for our dotnet base images.
```
edgebuilds.azurecr.io/microsoft/azureiotedge-module-base-rust:1.0.0-[ARCH]
```

While I was touching the build logic I also refactored the edgehub build logic. The problem before was that the logic that moved the artifacts around was in a bunch of yamls, which was not conducive to local builds. This new way will allow us to run a series of scripts to build edgehub locally. Something like this:
```
scripts/linux/buildBranch.sh
scripts/linux/cross-platform-rust-build.sh --os alpine --arch amd64 --build-path mqtt/mqttd
scripts/linux/cross-platform-rust-build.sh --os alpine --arch amd64 --build-path edgehub/watchdog
scripts/linux/consolidate-build-artifacts.sh --artifact-name "edge-hub"
scripts/linux/buildImage.sh -r "$(registry.address)" -u "$(registry.user)" -p "$(registry.password)" -i "${{ parameters.imageName }}" -n "${{ parameters.namespace }}" -P "${{ parameters.project }}" -v "${{ parameters.version }}"
```


builds/misc/templates/build-rust-artifacts.yaml
```diff
@@ -15,19 +29,19 @@ steps:
...
       filePath: scripts/linux/cross-platform-rust-build.sh
-      arguments: --os ubuntu18.04 --arch arm32v7 --build-path mqtt/mqttd
+      arguments: --os ubuntu18.04 --arch arm32v7 --build-path test/modules/generic-mqtt-tester
   - task: Bash@3
-    displayName: Build MQTT Broker - arm64
+    displayName: Build generic mqtt tester - arm64
     inputs:
       filePath: scripts/linux/cross-platform-rust-build.sh
-      arguments: --os ubuntu18.04 --arch aarch64 --build-path mqtt/mqttd
+      arguments: --os ubuntu18.04 --arch aarch64 --build-path test/modules/generic-mqtt-tester
```
and-rewsmith, line 46: I would try to avoid duplicating some of this code, but since display name cannot contain dynamic text I think it is better to have duplication here and an understandable description we can see in the build pipeline

test/modules/generic-mqtt-tester/docker/arm64v8/Dockerfile
```diff
@@ -0,0 +1,14 @@
+ARG base_tag=1.0.0-linux-arm64v8
```
and-rewsmith, line 1: From the description:
> I added some base images for arm, which I don't think we will need to update at all so I pushed some base images with the tag below. If we need to start updating these we should conform with the branch-consistent versioning scheme for our dotnet base images.

builds/misc/templates/build-rust-artifacts.yaml
damonbarry: I see a few different places in our yaml files where we build the displayName from variables...is this what you're talking about?

Examples:
```yaml
displayName: Linux Rust ${{ parameters.archName }}
displayName: Build ($(Build.Configuration))
```

and-rewsmith: For the second one I think this might work because it is available as soon as the build starts, but I wouldn't expect the first to work based on this:
https://github.com/MicrosoftDocs/azure-devops-docs/issues/2327

Also I couldn't find the first example in our repo.

scripts/linux/consolidate-build-artifacts.sh
```diff
@@ -0,0 +1,125 @@
...
+process_args "$@"
+
+cd $BUILD_REPOSITORY_LOCALPATH
+
+TARGET_ARM32V7="armv7-unknown-linux-gnueabihf"
+TARGET_ARM64V8="aarch64-unknown-linux-gnu"
+TARGET_AMD64_MUSL="x86_64-unknown-linux-musl"
+
+case "$ARTIFACT_NAME" in
+generic-mqtt-tester)
```
damonbarry, line 70: In the script you allow the caller to specify which artifact they want to consolidate, but then the only caller (the yaml file) always calls the script twice, once for each artifact. So do we really need the extra logic to just consolidate one artifact or the other?

builds/misc/templates/build-rust-artifacts.yaml
damonbarry: IMO it's your call...

The first example is from the `release/1.1` branch, doesn't exist in master.

scripts/linux/consolidate-build-artifacts.sh
and-rewsmith: If we are using it for local builds you won't want to call it for both, because the artifacts for both might not be available.

damonbarry (commented):
LGTM, although I'll leave it to someone more familiar with the edge hub/broker/watchdog stuff to sign off...

dylanbronson (commented):
Looks pretty good to me as long as you've verified it's working. Can you also update the documentation on building edgeHub either at the edgeHub README or here https://github.com/Azure/iotedge/blob/master/doc/devguide.md under build branch?

test/modules/generic-mqtt-tester/docker/arm64v8/base/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM arm64v8/ubuntu:18.04
+
+RUN apt-get update && apt-get install -y libssl-dev 
+
+# Add an unprivileged user account for running the module
+RUN useradd -ms /bin/bash moduleuser
+USER moduleuser
```
darobs, line 7: We set USER in base and in module's Dockerfile. Not really a problem, but not necessary.

darobs (approved):
LGTM, too.  One minor comment.

doc/devguide.md
```diff
@@ -68,6 +68,17 @@ The syntax of the "filter" argument is described [here](https://docs.microsoft.c
 
 The end-to-end tests are documented [here](../test/README.md).
 
+## Build Edge Hub Container Locally
```
darobs, line 71: Thanks for adding this!

darobs (approved):

and-rewsmith:
Passing build images:
https://msazure.visualstudio.com/One/_build/results?buildId=38818035&view=results

Passing build images release:
https://msazure.visualstudio.com/One/_build/results?buildId=38884610&view=results

vadim-kovalyov (approved):
Very nice!

PR #4269, opened 2021-01-21
Add twin legacy result to TestResultCoordinator

dylanbronson:
Adding twin test result to TestResultCoordinator. This was decided as the best way to migrate longhaul to the TRC. It keeps all the onus of validation on the twin tester module (just like longhaul does currently). 
This is a very minimal change that will not cause any regression in longhaul testing, but also does not add anything new to TRC. 
In order to migrate to the TRC, we need to add a minimal report to the TRC. This report has just the status of the sender results and the IsPassed attribute, which is the most important part


e2e_deployment_files/long_haul_deployment.template.json
```diff
@@ -207,92 +204,17 @@
...
-              "TransportType": {
-                "value": "Amqp"
-              },
-              "TwinUpdateSize": {
-                "value": "<TwinUpdateSize>"
-              },
-              "TwinUpdateFrequency": {
-                "value": "<TwinUpdateFrequency>"
+              "EdgeHubRestartFailureTolerance": {
+                "value": "<EdgeHubRestartFailureTolerance>"
```
and-rewsmith, line 209: There is also TwinUpdateFailureThreshold, but I don't think it needs to be configurable from the pipeline

and-rewsmith, line 209: There is also TwinUpdateFailureThreshold, but I don't think it needs to be configurable from the pipeline

e2e_deployment_files/long_haul_deployment.template.json
```diff
@@ -207,92 +204,17 @@
...
-              },
-              "testStartDelay": {
-                "value": "<TestStartDelay>"
-              },
-              "testDuration": {
-                "value": "<TestDuration>"
-              },
-              "ServiceClientConnectionString": {
-                "value": "<IoTHubConnectionString>"
-              },
```
and-rewsmith, line 240: Don't we need this connection string?

and-rewsmith (approved):
Mostly LGTM, but don't we need some connection string for the twin testers? Has this been tested or are we blocked by recent changes to longhaul deployment?

e2e_deployment_files/long_haul_deployment.template.json
dylanbronson: Think it's actually in there it's just that the diff makes it look like it's gone. 
I went from 4 to 2 twintester modules, so the diff looks a bit funky

yophilav (approved):
:shipit:

yophilav (approved):

PR #4270, opened 2021-01-22
Added checking for null values in arrays

vipeller:
which is not allowed in reported properties by IotHub


varunpuranik (approved):

varunpuranik:
We should align on the error returned to the client as well.

PR #4274, opened 2021-01-23
Fix GetModules error when calling using iso datetime format

pmzara:
Calling GetModules using DirectMethod with parameters where the DateTime format is ISO causes the API to return an error because the library that parses the JSON payload converts the DateTime parameters to a different format.


test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
```diff
@@ -61,6 +61,96 @@ public async Task TestGetModuleLogs()
             Assert.AreEqual(expected, response.Payload.Replace("\r\n", "\n"));
         }
 
+        [Test]
+        public async Task TestGetModuleLogsSinceDuration()
```
damonbarry, line 65: Is it possible to write unit tests instead of end-to-end tests? This is a very high-level way to test low-level functionality, and will add as much as 15 minutes to the end-to-end run time.

For example, could tests be added to `edge-agent\test\Microsoft.Azure.Devices.Edge.Agent.Core.Test\requests\RequestHandlerBaseTest.cs`?

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/SerDeExtensions.cs
```diff
@@ -20,6 +20,20 @@ public static T FromJson<T>(this string json)
                 : JsonConvert.DeserializeObject<T>(json);
         }
 
+        public static T FromJson<T>(this string json, JsonSerializerSettings jsonSerializerSettings)
```
damonbarry, line 23: This new function is mostly a duplicate of the existing `FromJson<T>` method. Could we reduce duplication by adding the `jsonSerializerSettings` argument to the original method, but with a default value? I _think_ that calling `JsonConvert.DeserializeObject<T>(json, (JsonSerializerSettings?)null)` is the same as calling `JsonConvert.DeserializeObject<T>(json)`. See [here](https://github.com/JamesNK/Newtonsoft.Json/blob/7c3d7f8da7e35dde8fa74188b0decff70f8f10e3/Src/Newtonsoft.Json/JsonConvert.cs#L718-L721).

damonbarry (changes requested):

edge-util/src/Microsoft.Azure.Devices.Edge.Storage/SerDeExtensions.cs
pmzara: I removed the extra function and added the default value to the new parameter.

test/Microsoft.Azure.Devices.Edge.Test/EdgeAgentDirectMethods.cs
pmzara: It's not possible to use unit tests on this case, because what's causing the problem is the API call with invalid DateTime formats. If we only add tests to check the DateTime format after the serialization we can't reproduce the problem without calling the API.

PR #4276, opened 2021-01-24
Nested E2E : LeafDevice to GrandParent

yophilav:
- Change the test agent capability from `run-idle` to `run-nested-e2e` to better reflect the feature
- Introduce the NestedEdgeConfig component to EdgeDevice for nested edge context configuration
- Introduce nested edge device connection to grandparent verification test
- Update the Keyvault configuration to use E2E test IoTHub instead of dev IotHub


huguesBouvier (commented):
Can you run a full nested E2E + partial(if you want) E2E for regression?

richma-ms:
>                 Context.Current.NestedEdge);

do these need to be updated as well to use the new fields?

---
Refers to: test/Microsoft.Azure.Devices.Edge.Test/Device.cs:79 in 2df777c. [](commit_id = 2df777c5aa1eff2c9c1ada027920f648dd610c02, deletion_comment = False)

richma-ms:
>                 Context.Current.NestedEdge);

same here

---
Refers to: test/Microsoft.Azure.Devices.Edge.Test/Device.cs:108 in 2df777c. [](commit_id = 2df777c5aa1eff2c9c1ada027920f648dd610c02, deletion_comment = False)

test/Microsoft.Azure.Devices.Edge.Test/helpers/Context.cs
```diff
@@ -109,6 +109,7 @@ IEnumerable<Registry> GetAndValidateRegistries()
             this.TestTimeout = TimeSpan.FromMinutes(context.GetValue("testTimeoutMinutes", 5));
             this.Verbose = context.GetValue<bool>("verbose");
             this.ParentHostname = Option.Maybe(Get("parentHostname"));
+            // FQDNS for the Edge Device
```
richma-ms, line 112: >// FQDNS for the Edge Device [](start = 12, length = 28)

formatting is a little weird here, maybe put it behind the line you're trying to comment?

test/Microsoft.Azure.Devices.Edge.Test/helpers/SasManualProvisioningFixture.cs
```diff
@@ -23,19 +21,19 @@ protected virtual async Task SasProvisionEdgeAsync(bool withCerts = false)
                 CancellationToken token = cts.Token;
                 DateTime startTime = DateTime.Now;
 
-                EdgeDevice device = await EdgeDevice.GetOrCreateIdentityAsync(
-                    Context.Current.DeviceId.GetOrElse(DeviceId.Current.Generate()),
-                    Context.Current.ParentDeviceId,
+                this.device = await EdgeDevice.GetOrCreateIdentityAsync(
+                    DeviceId.Current.Generate(),
```
richma-ms, line 25: >DeviceId.Current.Generate(), [](start = 20, length = 28)

this looks like we won't ever use the DeviceId from the context anymore, is that intentional?

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeDevice.cs
```diff
@@ -14,20 +15,50 @@ public class EdgeDevice
...
 
-        EdgeDevice(Device device, bool owned, IotHub iotHub)
+        readonly NestedEdgeConfig nestedEdgeConfig;
+
+        public class NestedEdgeConfig
+        {
+            public readonly string DeviceHostname;
+            public readonly bool IsNestedEdge;
+            public readonly Option<string> ParentDeviceId;
+            public readonly string ParentHostname;
```
richma-ms, line 25: these seem like they can just be readonly getters

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeDevice.cs
```diff
@@ -14,20 +15,50 @@ public class EdgeDevice
         readonly IotHub iotHub;
         readonly bool owned;
 
-        EdgeDevice(Device device, bool owned, IotHub iotHub)
+        readonly NestedEdgeConfig nestedEdgeConfig;
+
+        public class NestedEdgeConfig
```
richma-ms, line 20: >NestedEdgeConfig [](start = 21, length = 16)

actually I'm not seeing why this class is needed, looks like the only thing that's being used out of here is the ParentDeviceId?

test/Microsoft.Azure.Devices.Edge.Test/helpers/SasManualProvisioningFixture.cs
yophilav: We are still doing the same thing when the `NestedEdgeConfig(5)` is called here :
https://github.com/Azure/iotedge/blob/2df777c5aa1eff2c9c1ada027920f648dd610c02/test/Microsoft.Azure.Devices.Edge.Test/helpers/ManualProvisioningFixture.cs#L77

I'm basically consolidate the parameter checking logic to the same spot in `NestedEdgeConfig` and then read-only throughout the test cases (instead of read & check them every test case).

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeDevice.cs
yophilav: Yup

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeDevice.cs
yophilav: donezo

richma-ms (approved):

huguesBouvier (changes requested):
Can you run a full nested E2E + partial(if you want) E2E for regression?

yophilav:
> Can you run a full nested E2E + partial(if you want) E2E for regression?

Done

SingleNode E2E:
https://msazure.visualstudio.com/One/_build/results?buildId=39670038&view=logs&j=4323b27c-bad1-54aa-e27b-49b281e4d21f&t=2ea499ca-cfe0-556e-a400-a67ccaeb494e

Nested E2E: 
https://msazure.visualstudio.com/One/_build/results?buildId=39671938&view=logs&j=e8afecc0-c259-5cc1-62e0-5c5d1c2a3db4&t=f401ce77-605a-59dc-34fb-0bcc6cca03c5

huguesBouvier (approved):

huguesBouvier (changes requested):
Check dylan run from friday

PR #4278, opened 2021-01-25
Feature/mqtt bridge persistence

nyanzebra:
# MQTT Bridge Persistence
This is a draft PR and not necessarily indicative of the final code. The purpose of this pr is to get some early feedback on the design of the RingBuffer and interface changes within the code.

I am adding comments around different locations in the code to help point people to where I want some feedback, but please feel free to look through the rest and see if there is anything you wish to comment on :)


mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -69,18 +72,24 @@ impl BridgeHandle {
...
+impl Bridge<RingBuffer> {
     pub fn new_upstream(
         system_address: &str,
         device_id: &str,
         settings: &ConnectionSettings,
     ) -> Result<Self, BridgeError> {
         const BATCH_SIZE: usize = 10;
+        const FILE_NAME: &'static str = "ring_buffer.txt";
+        const FLUSH_OPTIONS: FlushOptions = FlushOptions::AfterXBytes(1024);
+        const MAX_FILE_SIZE: usize = 1024 * 1024;
```
nyanzebra, line 92: All of these default settings here and within the builder should probably go away in favour of some sort of settings or config. Need to look into what to edit to pass along such fields/values.

mqtt/mqtt-bridge/src/bridge.rs
```diff
@@ -117,7 +133,7 @@ impl Bridge<WakingMemoryStore> {
 
 impl<S> Bridge<S>
 where
-    S: StreamWakeableState + Send,
+    S: StreamWakeableState + Send + Sync,
```
nyanzebra, line 136: I have gone ahead and commented out the in-memory solution for storage and have moved to ringbuffer which can be send and sync

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -58,19 +58,25 @@ impl TryFrom<TopicRule> for TopicMapper {
...
 /// Handle events from client and saves them with the forward topic
-pub struct StoreMqttEventHandler<S> {
+pub struct StoreMqttEventHandler<S>
+where
+    S: StreamWakeableState + Send + Sync,
+{
     topic_mappers: HashMap<String, TopicMapper>,
     topic_mappers_updates: TopicMapperUpdates,
-    store: PublicationStore<S>,
+    store: Arc<PublicationStore<S>>,
```
nyanzebra, line 67: This Arc can likely be avoided by making PublicationStore cloneable, it should also already be Send and Sync as the member of PublicationStore is PublicationInner which is an Arc<Mutex>

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -122,10 +128,20 @@ impl<S> StoreMqttEventHandler<S> {
     }
 }
 
+async fn push_publication<S>(
```
nyanzebra, line 131: Small helper fn I was working with while trying to make sure everything was at least Send. Can probably remove or extend this to be drop in below.

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -77,45 +77,125 @@ where
...
-                },
-                |extracted| Poll::Ready(Some(Ok(extracted))),
-            )
+            let poll;
+            {
+                let batch_future = self.next_batch();
+                let mut batch_future = Box::pin(batch_future);
+                poll = batch_future.as_mut().poll(cx);
+            }
+            println!("poll {:?}", poll);
```
nyanzebra, line 86: Currently debugging a bug I noticed while testing, so feel free to ignore these printlns

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -77,45 +77,125 @@ where
...
-                    state_lock.set_waker(cx.waker());
-                    Poll::Pending
-                },
-                |extracted| Poll::Ready(Some(Ok(extracted))),
-            )
+            let poll;
+            {
+                let batch_future = self.next_batch();
+                let mut batch_future = Box::pin(batch_future);
+                poll = batch_future.as_mut().poll(cx);
```
nyanzebra, line 84: This is all to make async play nice with a poll fn.

mqtt/mqtt-bridge/src/persist/publication_store.rs
```diff
@@ -6,22 +6,55 @@ use parking_lot::Mutex;
...
+// impl PublicationStore<Sled> {
+//     pub fn new_db(
+//         path: String,
+//         tree_name: String,
+//         flush_options: FlushOptions,
+//         batch_size: usize,
+//     ) -> Self {
+//         Self::new(Sled::new(path, tree_name, flush_options), batch_size)
+//     }
+// }
```
nyanzebra, line 45: I have commented out both Sled and Memory. My plan is to remove memory altogether, and once I have tests working and if there is some extra time I will try to add sled as an alternative option.

mqtt/mqtt-bridge/src/persist/publication_store.rs
```diff
@@ -6,22 +6,55 @@ use parking_lot::Mutex;
...
     offset: u64,
     loader: MessageLoader<S>,
 }
 /// Persistence implementation used for the bridge
 pub struct PublicationStore<S>(Arc<Mutex<PublicationStoreInner<S>>>);
 
-impl PublicationStore<WakingMemoryStore> {
-    pub fn new_memory(batch_size: usize) -> PublicationStore<WakingMemoryStore> {
-        Self::new(WakingMemoryStore::default(), batch_size)
+unsafe impl<S> Send for PublicationStore<S> {}
```
nyanzebra, line 27: This might not be needed anymore, I started with this before making all the where constraints be Send and Sync in most places.

mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -37,19 +40,36 @@ pub struct Builder<S> {
...
-impl Default for Builder<WakingMemoryStore> {
+// impl Default for Builder<WakingMemoryStore> {
+//     fn default() -> Self {
+//         Self {
+//             local: PumpBuilder::default(),
+//             remote: PumpBuilder::default(),
+//             store: Box::new(|| PublicationStore::new_memory(0)),
+//         }
+//     }
+// }
```
nyanzebra, line 51: Like mentioned elsewhere, I plan to remove in memory implemenation

mqtt/mqtt-bridge/src/persist/waking_state/memory.rs
```diff
@@ -4,6 +4,8 @@ use std::{
     task::Waker,
```
nyanzebra, line 4: Like mentioned elsewhere I plan on removing the memory implementation

mqtt/mqtt-bridge/src/persist/storage/sled/mod.rs
```diff
@@ -0,0 +1,164 @@
+use std::{
```
nyanzebra, line 1: After I get all the tests working and if I have some spare time, I will try to incorporate sled as an option, I have this code here so that if someone has some good ideas for what kind of settings or async implementations might work well they can post.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/block.rs
```diff
@@ -0,0 +1,201 @@
...
+    inner: Vec<u8>,
+}
+
+impl Data {
+    pub fn new(data: Vec<u8>) -> Self {
+        Self { inner: data }
+    }
+}
+
+pub(crate) fn validate(block: &BlockHeaderWithHash, data: &Data) -> RingBufferResult<()> {
```
nyanzebra, line 94: Let's make sure that the validation of block looks 100% correct :) So if I can get some eyeballs looking over this that will be very helpful :)

mqtt/mqtt-bridge/src/persist/storage/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        amount: usize,
+    ) -> Pin<Box<dyn Future<Output = StorageResult<VecDeque<(Key, Value)>>> + Send + 'a>>;
+    fn remove_future<'a>(
+        &self,
+        key: &Key,
+    ) -> Pin<Box<dyn Future<Output = StorageResult<()>> + Send + 'a>>;
+    fn set_waker(&mut self, waker: Waker);
+    fn waker(&mut self) -> &mut Option<Waker>;
+}
+
```
nyanzebra, line 92: This interface can likely go away, it is a useful interface for my development purposes if there is some minor changes I want to test I don't have to break the entire code base, just patch it up below in StreamWakeableState.

mqtt/mqtt-bridge/src/persist/storage/error.rs
```diff
@@ -0,0 +1,35 @@
...
+pub enum StorageError {
+    #[error("RingBuffer error occurred")]
+    RingBuffer(#[from] RingBufferError),
+    #[error("Database error occurred")]
+    Database(#[from] SledError),
+    #[error("Serialization error occurred")]
+    Serialization(#[from] BincodeError),
+}
+
+// impl From<RingBufferError> for StorageError {
```
nyanzebra, line 19: Was trying different ways to create Errors, and seems we use the thiserror crate. Will remove these comments, but needed them while I was figuring out thiserror.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+/// It is possible that W could wrap around
+/// and cause us to lose data due to variable
+/// sized data being written.
+/// If W + data size > R and should write precede read,
+/// the data could become corrupted.
+/// Either this is prevented by scanning the file for
+/// the next block and checking if the write will fit.
+/// Or, W + data must not pass both D and R.
+/// Not passing D helps with guaranteeing all data read was handled.
+/// Not passing R helps avoid corruption.
```
nyanzebra, line 59: This is the strategy I thought might work well in an async env where we don't want to lose any data we haven't handled yet (unless there is a bit flip or internal HW error that prevents getting the data). 

If there are any questions on this, let me know, I want to put rust docs through the code I am adding so it will be easy to re-read and test how to use.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+pub struct RingBuffer {
+    inner: Arc<Mutex<MmapMut>>,
+    max_file_size: usize,
+    flush_options: FlushOptions,
+    pointers: Arc<FilePointers>,
+    waker: Option<Waker>,
```
nyanzebra, line 65: I am not super confident in my async await ability... but I think now that I have changed the interface on StreamWakeableState we don't need this field or the set_waker fns anymore? Can someone confirm this for me?

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+struct RingBufferWriteFuture {
+    pointers: Arc<FilePointers>,
+    flush_options: FlushOptions,
+    mmap: Arc<Mutex<MmapMut>>,
+    flush_state: Arc<FlushState>,
+    value: Vec<u8>,
+    max_file_size: usize,
+}
+
+impl Future for RingBufferWriteFuture {
```
nyanzebra, line 120: Each operation is implemented as a future, that way we can do two things:
1. apply back pressure to read or write if either one is moving faster. By this I mean, if we are writing a lot and not reading as much or vice versa, we will be in a pending state in the async execution and give the other operation a chance to catch up. Not sure how often pending is checked, though, so that needs to be thought about.
2. The other part about making these futures is it allows us to be prepared to switch to anything like AsyncWrite or other async traits with little effort.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+    fn insert_future<'a>(
+        &self,
+        value: &Vec<u8>,
+    ) -> Pin<Box<dyn Future<Output = StorageResult<()>> + Send + 'a>> {
+        let future = RingBufferWriteFuture {
+            pointers: self.pointers.clone(),
+            flush_options: self.flush_options,
+            mmap: self.inner.clone(),
+            flush_state: self.flush_state.clone(),
+            value: value.clone(),
```
vadim-kovalyov, line 375: can we move the buffer instead of clone?

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
vadim-kovalyov: If RingBuffer is the one who implements `StreamWakeableState`, then you probably still need waker. And I believe it should be used in the following way:
- If a Write future can't do the job (return "Pending" state), you need to clone the waker and remember it in the ring buffer.
- The write future then goes to sleep.
- Next time when the "read" operation is performed, you need to check if there is a waker, and if there is one - wake it.

It seems that you would need two wakers - one for "write" and one for "delete".

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
nyanzebra: Yeah cloning the data is more than I want, let me update this :)

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+        let total_size = block_size + data_size;
+        // Check to see if we might corrupt data if we write, if so, return pending.
+        // As trying later might be successful.
+        if write_index < read_index {
+            if write_index + total_size > read_index {
+                println!("write pending");
+                return Poll::Pending;
+            }
+        }
+        cx.waker().wake_by_ref();
```
vadim-kovalyov, line 148: I believe waker should not be called from the poll method of the future. Instead you need to "remember" the waker somehow. By "remember" I mean that you need to store it in the struct that implements `StreamWakeableState`. 

Next time a `batch` (or rather `remove`) method is called on the `StreamWakeableState`, at the end of the reading, you would wake a waker if there is one.

mqtt/mqtt-bridge/src/persist/waking_state/mod.rs
```diff
@@ -21,35 +20,142 @@ pub mod memory;
 /// Thus all implementations will share the below test suite.
 #[async_trait]
 pub trait StreamWakeableState {
-    fn insert(&mut self, key: Key, value: Publication) -> Result<(), PersistError>;
+    async fn insert(&self, key: Key, value: Publication) -> Result<(), PersistError>;
```
vadim-kovalyov, line 23: This is something we should probably chat about as a team. 

Having async_trait will create an extra level of indirection (it desugars into Box<dyn Future...>). If we are settling on a single implementation for storage, it might be worth to stick to concrete future types kept on stack.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+            .update(0, 0, timer.elapsed().as_millis() as usize);
+        if should_flush {
+            self.flush_state.reset(&self.flush_options);
+        }
+
+        Poll::Ready(Ok(()))
+    }
+}
+
+impl Storage<usize, Vec<u8>> for RingBuffer {
```
nyanzebra, line 365: One thing I thought about is how to guarantee we don't read some data and have it in memory but write over it on disk from input thinking everything is fine when in reality we might crash. I think the right thing to do is something like this:
1. Have each read flag the data as 'in progress' or something
2. On read completion of the data (most likely remove operation) update the flag to 'done'
3. Have a BatchResults struct that keep track of these
4. Once all are done the BatchResults can update the read pointer

If we were to crash that will be okay because we can come back up without having to worry about writing over read after we did a batch. It would also allow us to avoid duplicates after crashing. I still need to ponder this idea a bit... but will push after I have something that looks reasonable to discuss.

mqtt/mqtt-bridge/src/persist/storage/mod.rs
vadim-kovalyov: Yes, this seems unnecessary trait at the end. But just in case, in my previous comment I mentioned that we should probably try to stay away from async traits and try not to use dynamic dispatching

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
nyanzebra: So, I would need check both wakers in the "read" operation?

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
nyanzebra: Ahh okay, so wake once the condition for blocking has passed

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
nyanzebra: Upon thinking on this more, I think I can just use two read pointers, one a begin and one an end. The begin one will be updated in "remove" and the end updated in "batch". The "remove" operation will continue to read ahead and check if there are valid entries after a removal. If there are valid entries that were marked for removal, then continue to move up the read begin pointer. In "batch" the end read pointer will be at the end of the last element in the batch.

vadim-kovalyov: I like this idea.

nyanzebra: As discussed offline, we can have a single read pointer and error if the key for remove does not match the pointer. As we are only expecting to maintain in-order transfer as per MQTT

PR #4281, opened 2021-01-26
Nested E2E - rotate SAS key of leaf device

richma-ms:
This change adds a testcase that rotates the SAS key of the leaf device at runtime, and verifies that the client can use the new credentials to connect through the hierarchy to IoT Hub.

### DIFF
--- a/builds/e2e/templates/e2e-run.yaml
+++ b/builds/e2e/templates/e2e-run.yaml
@@ -21,11 +21,17 @@ steps:
     #filter out tests not usable on nested edge
     if ('$(nestededge)' -eq 'true')
     {
+      $filter += '&Category!=SingleNodeOnly'
+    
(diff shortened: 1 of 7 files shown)


builds/e2e/templates/e2e-run.yaml
```diff
@@ -21,11 +21,17 @@ steps:
     #filter out tests not usable on nested edge
     if ('$(nestededge)' -eq 'true')
     {
+      $filter += '&Category!=SingleNodeOnly'
```
huguesBouvier, line 24: Shouldn't the filter be &Category=NestedEdgeOnly ?

huguesBouvier: I forgot, please add the link of your pipeline run. Thanks!

huguesBouvier (approved):

PR #4286, opened 2021-01-26
Drop Ubuntu 16.04, add 18.04

damonbarry:
This change updates our YAML pipelines **for the 1.1 release** to generate and consume bits built on/for Ubuntu 18.04 instead of 16.04. Specifically:
- Connectivity, End-to-end, Long Haul and Stress pipelines + downstream scripts: replace references to Ubuntu 16.04 artifacts with 18.04
- Edgelet Packages pipeline + downstream scripts: don't build Ubuntu 16.04 artifacts
- Remove Ubuntu 16.04 cross-compile scripts
- Various markdown files, tests, comments, and schema examples: change Ubuntu 16.04 references to 18.04

Note: this change does not update the arm-based docker containers that the K8s betas use for iotedged. They're still 16.04-based in this branch, and will be updated later.


arsing:
If you're dropping Ubuntu 16.04 support, you should drop Debian 8 support too. (Should be just in `package.sh` and any YAMLs that reference it.)

damonbarry:
> If you're dropping Ubuntu 16.04 support, you should drop Debian 8 support too. (Should be just in `package.sh` and any YAMLs that reference it.)

Yeah, and Rasbian Stretch (which exited LTS around the same time as Debian 8). But I plan to handle those separately...just focused on Ubuntu in this PR.

arsing (approved):

PR #4288, opened 2021-01-26
Fixes for long haul pipeline

gordonwang0:
Makes parameters for long haul similar to connectivity.

### DIFF
--- a/scripts/linux/runE2ETest.sh
+++ b/scripts/linux/runE2ETest.sh
@@ -15,6 +15,7 @@ function clean_up() {
 
     echo 'Remove IoT Edge and config file'
     rm -rf /var/lib/aziot/edged/
+    rm -rf /etc/systemd/system/aziot-edged.service.d/
     rm -rf /etc/aziot/edged/config.yaml
 
     echo 'Remove docker containers'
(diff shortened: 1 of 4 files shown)


scripts/linux/trcE2ETest.sh
```diff
@@ -875,7 +880,7 @@ elif [[ "${TEST_NAME,,}" == "${CONNECTIVITY_TEST_NAME,,}" ]]; then
     NETWORK_CONTROLLER_RUNPROFILE=${NETWORK_CONTROLLER_RUNPROFILE:-Offline}
 
     is_build_canceled=$(is_cancel_build_requested $DEVOPS_ACCESS_TOKEN $DEVOPS_BUILDID)
-    if [ $is_build_canceled -eq 1 ]; then
+    if [ "$is_build_canceled" -eq '1' ]; then
```
yophilav, line 883: is `"`  synonymous as `'` in bash?

scripts/linux/runE2ETest.sh
```diff
@@ -15,6 +15,7 @@ function clean_up() {
 
     echo 'Remove IoT Edge and config file'
     rm -rf /var/lib/aziot/edged/
+    rm -rf /etc/systemd/system/aziot-edged.service.d/
```
yophilav, line 18: We will likely need this in trcE2ETest.sh as well.

scripts/linux/trcE2ETest.sh
gordonwang0: No. " interpolates, ' doesn't. So "$is_build_canceled" will be string value of $is_build_canceled and '1' is literal string '1'

scripts/linux/runE2ETest.sh
gordonwang0: Added

yophilav (approved):

scripts/linux/trcE2ETest.sh
yophilav: Ok. I'm just making sure we are not comparing `1`  to `'1'`

PR #4303, opened 2021-01-28
Improve Nested Edge E2E Test Pipeline

yophilav:
- Change the capability name from `run-idle` to `run-nested-e2e`
- Change privileges & permission of VSTS upon installation
- Organized Keyvault secrets
- Introduce log rotation for nested edge pipeline
Note: Unfortunatly, I cannot automate the azure-iot extension installation of Azure CLI. The user needs to manually login to the VM and install that themselves. The extension installation requires tty to verify the options.


builds/e2e/templates/nested-get-device-name.yaml
```diff
@@ -4,7 +4,7 @@ steps:
     inputs:
       targetType: inline
       script: |
-        device_name=$(az vm show -d  -g "iotedge-deploy" -n $(hostname) --query fqdns)
+        device_name=$(ip route get 8.8.8.8 | sed -n '/src/{s/.*src *\([^ ]*\).*/\1/p;q}')
```
dylanbronson, line 7: what is the change here?

builds/e2e/templates/nested-get-secrets.yaml
```diff
@@ -4,10 +4,22 @@ steps:
...
+        edgebuild-blob-core-connection-string,
+        edgebuild-service-principal-secret,
+        edgebuild-1-PAT
+
   - task: AzureKeyVault@1
-    displayName: 'Nested Edge Azure Key Vault'
+    displayName: 'E2E Azure Key Vault'
     inputs:
       azureSubscription: $(az.subscription)
       KeyVaultName: $(azure.keyVault)
```
dylanbronson, line 22: Both these tasks use the same kv, $(azure.keyVault)? Should we just combine them entirely?

builds/e2e/templates/nested-get-device-name.yaml
yophilav: Getting the public IP address instead of using the FQDN (which is empty string for this VM).

builds/e2e/templates/nested-get-secrets.yaml
yophilav: That's the unfortunately reality. I wanted to separate secrets into 2 KV : E2E & the common KV. 
Given the development of how Nested LH & E2E pipeline tangle, it isn't going to be easy to separate the two.
I'm going to group the secrets as you suggested.

huguesBouvier (approved):

PR #4304, opened 2021-01-28
Adds a ring buffer to MQTT bridge

nyanzebra:
# Purpose
This adds a ring buffer implementation to the mqtt-bridge persistence code. As this can provide configurable storage in the bride. The ring buffer allows for variable sized data to be read and written. For integrity, ordering, and manageability there are block headers that exist before every data entry.

This is not including integration changes to utilize the ring buffer. This is solely on the ring buffer implementation itself. Future prs will come for integration and any additional integration tests required.


mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+            }
+        }
+
+        let start = write_index;
+        let end = start + block_size;
+
+        // Check if an existing block header is present. If the block is there
+        // and if the overwrite flag is not set then we shouldn't write.
+        let result = deserialize_block(self.inner.clone(), start, end);
+        if result.is_ok() {
```
vadim-kovalyov, line 300: nit: why not if let or match?

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+
+        self.order.store(order + 1, Ordering::SeqCst);
+
+        self.flush_state
+            .update(1, total_size, timer.elapsed().as_millis() as usize);
+        if should_flush {
+            self.flush_state.reset(&self.flush_options);
+        }
+
+        self.wake_up_task();
```
vadim-kovalyov, line 341: not a big deal, but we might call the waker  right after we whore the `write` pointer and before flush.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+            waker.wake();
+        }
+    }
+}
+
+#[allow(dead_code)]
+impl Drop for RingBuffer {
+    fn drop(&mut self) {
+        let mmap = self.inner.lock();
+        mmap.flush().expect("Failed to flush");
```
vadim-kovalyov, line 464: do we really want to panic here? I'm not saying this is bad, just curious.

vadim-kovalyov (commented):
I did a quick pass, and so far it looks good to me, you are on the right track. Just few minor styling things. I'll try to do one more thorough review later tomorrow.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
nyanzebra: The way I am looking at it is that the RingBuffer exists for the lifetime of the bridge. So, dropping is something that should only happen in rare cases where something else panicked. That is why I added this custom Drop to try and flush at the very last moment if a chance exists, and if we cannot, well we are already panicking so what is one more :)

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
nyanzebra: I feel there is a typo above :) Yeah, can move up

mqtt/mqtt-bridge/src/persist/storage/mod.rs
```diff
@@ -0,0 +1,60 @@
+pub mod ring_buffer;
+pub mod serialize;
+
+use super::StorageError;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
nyanzebra: I had to these in a lot of places as the code is only being used in tests for this PR. If I can get away with a allow on the mod that would probably be nicer

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -19,3 +22,12 @@ pub enum PersistError {
     #[error("Attempted to remove entry which does not exist")]
     RemovalForMissing,
 }
+
+// This might replace the PersistError or merge with it.
+#[derive(Debug, thiserror::Error)]
+pub enum StorageError {
+    #[error("RingBuffer error occurred")]
```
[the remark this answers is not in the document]
nyanzebra: Oh good idea!

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
vadim-kovalyov: 0_0 oops

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
vadim-kovalyov: I hope nothing in the bridge should panic :) and panic inside a panic will cause app to terminate immediately, which is the worst case for us. If we can't flush, we already potentially loosing data. Abrupt termination can cause other bridges to loose data. So I would rather prefer ignoring (and maybe logging) the error. Not sure how logging would work here though... @dmolokanov, any thoughts?

nyanzebra: Good point on immediate termination. I would expect logging to not be guaranteed if we are terminating the process either... So, I can also see just swallowing the error as maybe the best choice...

dmolokanov: Yep, I agree. Let's try to "safely" clear resources and log an error with the highest level.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+        }
+
+        let mut start = read_index;
+        let mut vdata = VecDeque::new();
+        for _ in 0..count {
+            // This case shouldn't be pending, as we must have gotten something we can process.
+            if start >= write_index {
+                break;
+            }
+            let block;
```
dmolokanov, line 363: didn't see this before. do you know how it works? compiler converts it to something like this?

```rust
let block = { ..., binary_deserialize(bytes)? }
```

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+    use std::{
+        fs::{remove_dir_all, remove_file},
+        panic,
+    };
+
+    const FLUSH_OPTIONS: FlushOptions = FlushOptions::Off;
+    const FILE_NAME: &'static str = "test_file";
+    const MAX_FILE_SIZE: usize = 1024;
+
+    fn cleanup_test_file(file_name: String) {
```
dmolokanov, line 484: we use `tempfile` crate for these temporary files

dmolokanov (changes requested):
Impressive work! I didn't go to the pointers arithmetic details, so I hope you tested that well :)
Since we use it in our hot path we can invest some time to optimize allocations. And we will improve readability together!

I probably missed part of discussions. Are we going with the approach of fixed size blocks potentially with gaps, or you implemented the one which is where blocks can have arbitraty size?

mqtt/mqtt-bridge/src/persist/storage/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+    AfterEachWrite,
+    AfterXWrites(usize),
+    AfterXBytes(usize),
+    AfterXMilliseconds(usize),
+    Off,
+}
+
+#[allow(dead_code)]
+pub struct FlushState {
+    writes: AtomicUsize,
```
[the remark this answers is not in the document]
vadim-kovalyov: I believe you're referring to `FlushOptions`. Where`FlushState` keeps track of writes/times and being updated with every write, and needs to be thread safe.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/block.rs
```diff
@@ -0,0 +1,280 @@
...
+};
+use super::error::{BlockError, RingBufferError};
+
+/// A utility fn to calculate any entity that derives Hash.
+#[allow(dead_code)]
+pub(crate) fn calculate_hash<T>(t: &T) -> u64
+where
+    T: Hash + ?Sized,
+{
+    let mut hasher = DefaultHasher::new();
```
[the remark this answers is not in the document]
vadim-kovalyov: It makes sense what you're saying. But before we jump to optimization we need to consider possible attack vectors. If we are sure that payload cannot be fully controlled by an adversary, we can use unsafe and fast hashing.

I did a quick search, and found this interesting SO post - https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed. There are some funny collisions for some algos :D

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/block.rs
```diff
@@ -0,0 +1,280 @@
...
+
+/// + ------------+------+---------+
+/// | BlockHeader | hash | data... |
+/// +-------------+------+---------+
+///
+/// The BlockHeader contains attributes meaningful for data
+/// storage and validation.
+#[allow(dead_code)]
+#[derive(Copy, Clone, Debug, Deserialize, Hash, Serialize)]
+#[repr(C)]
```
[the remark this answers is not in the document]
vadim-kovalyov: My guess is to make sure the layout is predictable and never changes.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+    order: AtomicUsize,
+    pointers: Arc<FilePointers>,
+    waker: Option<Waker>,
+}
+
+unsafe impl Send for RingBuffer {}
+unsafe impl Sync for RingBuffer {}
+
+#[allow(dead_code)]
+fn find_pointers_post_crash(
```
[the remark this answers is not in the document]
nyanzebra: Sounds good to me! I want this to be as readable as possible so if there are any parts that look like they need work let me know!

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
nyanzebra: Good plan will switch to putting an appropriate log message

mqtt/mqtt-bridge/src/persist/storage/mod.rs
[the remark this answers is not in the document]
nyanzebra: Yes, you are correct Vadim in that FlushState and the pointers are atomic so that the operations from StreamWakeableState can be thread safe. However, because we need to assign waker and the PublicationStore and else use StreamWakeableState currently in a Mutex these could be non-atomic. My intention is to try to spend some time to think on the interface and see if can make the ring buffer be thread safe and be able to remove locks elsewhere. 

If this proves to be a larger problem then I think we can rely on the Mutexes elsewhere and remove atomics and Mutexes here.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/block.rs
```diff
@@ -0,0 +1,280 @@
...
+use std::{
+    collections::hash_map::DefaultHasher,
+    fmt::Debug,
+    hash::{Hash, Hasher},
+};
+use crate::persist::storage::{
+    serialize::{binary_serialize, binary_serialize_size},
+    StorageResult,
+};
+use super::error::{BlockError, RingBufferError};
```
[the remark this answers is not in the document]
nyanzebra: Yes :)

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/block.rs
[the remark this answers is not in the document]
nyanzebra: I tried to write things in a way so that it would be fairly easy to change implementation. As I put things in a function to calculate hash, so we can modify as need be.

I agree the hashing algorithm that is default in rust is not the fastest, though I think we can stick with it for now and if we need to tune performance and validate security concerns on different hash algorithm we can :)

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/block.rs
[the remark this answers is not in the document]
nyanzebra: Yes that is correct, I  want to make sure things are both small and conformant. The repr C should make sure things are not unnecessarily aligned and allow us to be sure of exact byte to byte representation if we need to look at the raw file.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/block.rs
```diff
@@ -0,0 +1,280 @@
...
+
+    use std::collections::hash_map::DefaultHasher;
+
+    use super::*;
+
+    fn random_data(data_length: usize) -> Vec<u8> {
+        (0..data_length).map(|_| rand::random::<u8>()).collect()
+    }
+
+    mod serialize {
```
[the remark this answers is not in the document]
nyanzebra: I can merge them if you like :)

This helped me organize things when testing locally and then minimize tests that I wasn't immediately looking at.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+    start: usize,
+    end: usize,
+    file_size: usize,
+) -> Vec<u8> {
+    let mmap = mmap.lock();
+    // Need to wrap around
+    if end > file_size {
+        let file_split = end - file_size;
+        let mut v = vec![];
+        v.append(&mut Vec::from(&mmap[start..file_size]));
```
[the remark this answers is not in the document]
nyanzebra: Yes, this part should be optimized to avoid copies. Went with simple for now to be sure I got all the bytes for wrap around case. Can spend some time eliminating copies :)

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+        .open(&PathBuf::from(file_name))
+}
+
+#[allow(dead_code)]
+fn deserialize_block(
+    mmap: Arc<Mutex<MmapMut>>,
+    start: usize,
+    end: usize,
+) -> BincodeResult<BlockHeaderWithHash> {
+    let bytes = mmap_read(mmap.clone(), start, end);
```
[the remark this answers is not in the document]
nyanzebra: Yes, I think here we can just return a reference to the deserialized object. We can make it generic and constrain it on deserialize

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+        }
+        FlushOptions::AfterXMilliseconds(xmillis_elapsed) => {
+            flush_state.millis_elapsed.load(Ordering::SeqCst) >= xmillis_elapsed
+        }
+        FlushOptions::Off => false,
+    }
+}
+
+#[allow(dead_code)]
+fn read_data_maybe_wrap_arond(
```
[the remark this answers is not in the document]
nyanzebra: Whoops @_@

mqtt/mqtt-bridge/src/persist/storage/serialize.rs
```diff
@@ -0,0 +1,27 @@
+use bincode::{deserialize, serialize, serialized_size, Result as BincodeResult};
```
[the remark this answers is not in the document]
nyanzebra: Yes we can move it. My original intent was to allow potentially different storage solutions use these helpers, but that could easily be done at a later time if we need other solutions :)

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -1,10 +1,13 @@
-use serde::{Deserialize, Serialize};
-
 mod loader;
 mod publication_store;
+mod storage;
```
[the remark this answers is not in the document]
nyanzebra: I think that is a fair point. I agree it should be moved

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mmap.rs
```diff
@@ -0,0 +1,29 @@
+use super::error::RingBufferError;
```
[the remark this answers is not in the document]
nyanzebra: Sounds good to me :)

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/error.rs
```diff
@@ -0,0 +1,32 @@
+use std::io::Error as IOError;
```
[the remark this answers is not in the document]
nyanzebra: Sure will add newline after each enum option :)

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+    flush_options: FlushOptions,
+    flush_state: Arc<FlushState>,
+    inner: Arc<Mutex<MmapMut>>,
+    max_file_size: usize,
+    order: AtomicUsize,
+    pointers: Arc<FilePointers>,
+    waker: Option<Waker>,
+}
+
+unsafe impl Send for RingBuffer {}
```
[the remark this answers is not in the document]
nyanzebra: I think Send and Sync require unsafe? https://doc.rust-lang.org/nomicon/send-and-sync.html

If I am misinterpreting this please let me know!

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
nyanzebra: Oh nice! This is exactly what I need!

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/block.rs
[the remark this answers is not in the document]
dmolokanov: yeah, makes sense.

mqtt/mqtt-bridge/src/persist/storage/ring_buffer/mod.rs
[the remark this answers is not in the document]
dmolokanov: need to be more clear: why we implement it ourselves instead of asking the compiler to derive implementation for us?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+    mmap: &'a mut MmapMut,
+    max_file_size: usize,
+    order: usize,
+    pointers: FilePointers,
+    waker: Option<Waker>,
+}
+
+#[allow(dead_code)]
+impl<'a> RingBuffer<'a> {
+    fn new(mmap: &'a mut MmapMut, max_file_size: usize, flush_options: FlushOptions) -> Self {
```
nyanzebra, line 77: Changed to mmap, for new but feel like having it as a path makes more sense...

mqtt/mqtt-bridge/src/persist/storage/mod.rs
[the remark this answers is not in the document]
vadim-kovalyov: did it work? did you try mod level allow?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+        save_data(self.mmap, &data, start, self.max_file_size, should_flush)?;
+
+        end = start + data_size;
+        self.pointers.write = end % self.max_file_size;
+
+        self.wake_up_task();
+
+        self.order += 1;
+
+        self.flush_state
```
vadim-kovalyov, line 153: nit:
```rust
        if should_flush {
            self.flush_state.reset(&self.flush_options);
        } else {
            self.flush_state
                .update(1, total_size, timer.elapsed().as_millis() as usize);
        }
```

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+        let data_hash = calculate_hash(&data);
+
+        let write_index = self.pointers.write;
+        let read_index = self.pointers.read;
+        let order = self.order;
+        let key = write_index;
+
+        let block_header =
+            BlockHeaderWithHash::new(data_hash, data_size, order, read_index, write_index);
+        let block_size = serialized_block_size()?;
```
vadim-kovalyov, line 105: can this be static?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+
+        let total_size = block_size + data_size;
+        // Check to see if we might corrupt data if we write, if so, return err of full.
+        // There are two cases for this:
+        // 1. write has wrapped around and is now behind read
+        if write_index < read_index && write_index + total_size > read_index {
+            return Err(StorageError::RingBuffer(RingBufferError::Full));
+        }
+        // 2. write has reached the end (or close enough) but read has not moved
+        if read_index == 0 && write_index + total_size > self.max_file_size {
```
vadim-kovalyov, line 115: I think this needs to be checked for read_index other than 0 as well... something like this
```rust
        if write_index > read_index
            && write_index + total_size > self.max_file_size
            && write_index + total_size % self.max_file_size > read_index
        {
            return Err(StorageError::RingBuffer(RingBufferError::Full));
        }
```

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+        Ok(Key { offset: key as u64 })
+    }
+
+    fn batch(&self, count: usize) -> StorageResult<VecDeque<(Key, Publication)>> {
+        let write_index = self.pointers.write;
+        let read_index = self.pointers.read;
+        let block_size = serialized_block_size()?;
+
+        // If read would go into where writes are happening then we don't have data to read.
+        if read_index == write_index {
```
vadim-kovalyov, line 168: what if we filled up the whole buffer exactly, so write_index wrapped around and become equal to read_index again? 
- do we need to keep "wrapped" flag?
- or we should disallow write_index to wrap and reach read_index by checking `write_index < read_index && write_index + total_size **>=** read_index`

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+        let mut vdata = VecDeque::new();
+        for _ in 0..count {
+            // This case shouldn't be pending, as we must have gotten something we can process.
+            if start >= write_index {
+                break;
+            }
+
+            let end = start + block_size;
+            if end > self.max_file_size {
+                return Err(RingBufferError::WrapAround.into());
```
vadim-kovalyov, line 182: I'm not sure I understand why we return error on wrap around?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+
+        let start = key;
+        let mut end = start + block_size;
+        if end > self.max_file_size {
+            return Err(StorageError::RingBuffer(RingBufferError::WrapAround));
+        }
+
+        let maybe_block = load_block_header(self.mmap, start, block_size, self.max_file_size);
+        if maybe_block.is_err() {
+            return Err(StorageError::RingBuffer(RingBufferError::NonExistantKey));
```
vadim-kovalyov, line 236: nit: I'm just trying to avoid `.unwrap()` as much as possible.
```rust
        let mut block = match load_block_header(self.mmap, start, block_size, self.max_file_size) {
            Ok(block) => block,
            Err(e) => {
                return Err(StorageError::Serialization(e));
            }
        };

        if block.inner().hint() != BLOCK_HINT {
            return Err(StorageError::RingBuffer(RingBufferError::NonExistantKey));
        }
```

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+        }
+
+        let mut block = maybe_block.unwrap();
+        if block.inner().hint() != BLOCK_HINT {
+            return Err(StorageError::RingBuffer(RingBufferError::NonExistantKey));
+        }
+
+        let data_size;
```
vadim-kovalyov, line 244: nit:
```rust
        let data_size = {
            let inner_block = block.inner_mut();
            inner_block.set_should_not_overwrite(false);
            inner_block.data_size()
        };
```

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+                end += 1;
+                continue;
+            }
+        } else {
+            start += 1;
+            end += 1;
+            continue;
+        }
+
+        break;
```
vadim-kovalyov, line 340: not needed?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+    order: usize,
+    pointers: FilePointers,
+    waker: Option<Waker>,
+}
+
+#[allow(dead_code)]
+impl<'a> RingBuffer<'a> {
+    fn new(mmap: &'a mut MmapMut, max_file_size: usize, flush_options: FlushOptions) -> Self {
+        let file_pointers =
+            find_pointers_post_crash(&mmap, max_file_size).expect("Failed to init file pointers");
```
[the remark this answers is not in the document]
nyanzebra: So, we should new with Result? I am okay with that, just felt weird

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: I want it to be :) I need a way to get it to be static through bincode though... maybe this is something lazy_static can do? Haven't used it before, so if you know of a way to do this correctly, let me know!

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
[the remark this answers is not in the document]
vadim-kovalyov: Yeah, we do that in few places as well. I think it is fine for new() to return a Result. C'tor methods should have a right to fail :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Not sure on the above situation here... The code allows for a wrap around.
Imagine the '-' is just random garbage data and the '/' is a block and data.
[//:-------://]

We support this situation where we have some free space at the beginning and the end, but cannot fit entirely at the end. So, the `write_index + total_size > self.max_file_size` case seems odd to me. But the second part you suggest `write_index + total_size % self.max_file_size > read_index` should probably just be the first check I have i.e:
```rust
 if write_index < read_index && write_index + total_size > read_index {
            return Err(StorageError::RingBuffer(RingBufferError::Full));
 }
```
should be
```rust
 if write_index + total_size % self.max_file_size > read_index{
            return Err(StorageError::RingBuffer(RingBufferError::Full));
 }
```

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Aye aye will do!

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: * Good catch! Let me add a test for this very situation and it would appear some 'wrapped' flag might be necessary or as you said if we reach the max_file_size with the write, it should stay there until the next write. This combined with the change mentioned above in insert on correcting the bound checking with a write should work. Either case, more tests!
* I think write should still be allowed to wrap and reach read. That way we can at least optimize for space.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
vadim-kovalyov: you can add this somewhere at the top of the rign_buffer/mod.rs
```rust
lazy_static! {
    static ref BLOCK_SIZE: usize =
        serialized_block_size().expect("failed to calculate ring buffer block size");
}
```
and use it like this:
```rust
let block_size = *BLOCK_SIZE;
```

nyanzebra (commented):
One thing I found is we might still need a read begin and read end. Coupled with a flag for whether we have read or not.

Ex:
* We insert 10 items.
* We call remove(0)
* Item is marked as removed (updates the read pointer)
* We insert more until we write over the first item
* We call batch (we never read the first item!)

By having a 'read begin', 'read end', and a 'has read' flag we can get around this. The steps for are as follows:
* batch will update read end (we still can not read past the write pointer)
* batch will also update the flag 'has read'
* remove will proceed **only** if the 'has read' flag is set
* remove will update the read begin
* if read begin and read end are equal at the end of the remove call then we unset 'has ready', it must be done at the end of the call to not potentially reach the case where begin == end but all data was read.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Actually, I misspoke, we do want to break after we have found the first block

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
dmolokanov: or with `map_err`
```rust
        let mut block = match load_block_header(self.mmap, start, block_size, self.max_file_size)
           .map_err(|e| Err(StorageError::Serialization(e))?;

        if block.inner().hint() != BLOCK_HINT {
            return Err(StorageError::RingBuffer(RingBufferError::NonExistantKey));
        }
```

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+        }
+
+        let should_flush = self.should_flush();
+        save_block_header(self.mmap, &block, start, self.max_file_size, should_flush)?;
+
+        end += data_size;
+
+        self.pointers.read = end % self.max_file_size;
+
+        self.flush_state
```
dmolokanov, line 258: I think @vadim-kovalyov has some concerns in another method above. Should we extract common code to a separate method?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+            }
+
+            let end = start + block_size;
+            if end > self.max_file_size {
+                return Err(RingBufferError::WrapAround.into());
+            }
+
+            let bytes = &self.mmap[start..end];
+            // this is unused memory
+            if bytes.iter().map(|&x| x as usize).sum::<usize>() == 0 {
```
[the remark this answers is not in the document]
nyanzebra: I am not sure if this is a possible case, at least given what publication looks like. One option that I figured out while testing is to scan only 8 bytes for the 'hint' and if it is not immediately after data then it means it is "empty".

I will be updating this pr with bug fixes I have found over the last few days. Some a bit sneaky.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+        // 2. write has reached the end (or close enough) but read has not moved
+        if read_index == 0 && write_index + total_size > self.max_file_size {
+            return Err(StorageError::RingBuffer(RingBufferError::Full));
+        }
+
+        let mut start = write_index;
+
+        // Check if an existing block header is present. If the block is there
+        // and if the overwrite flag is not set then we shouldn't write.
+        let result = load_block_header(self.mmap, start, block_size, self.max_file_size);
```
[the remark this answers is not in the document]
nyanzebra: There are several possibilities: 
1. There is no data, this will deserialize as an empty block
2. There is a block and we deserialize
3. There may have been a block but the data is not deserializable into a block, which means this data would be lost. We have an option of trying to do a partial deserialize and hope to recover the data... but if the block is corrupt i.e. non deserializable... then seems it likely lost cause.

So, the reason I didn't care about err here is it seemed to be only possible if it was too late to do anything about. If we add versioning or different blocks this will 100% need to be handled.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/flush.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub millis_elapsed: usize,
+}
+
+#[allow(dead_code)]
+impl FlushState {
+    pub(crate) fn new() -> Self {
+        Self {
+            writes: usize::default(),
+            bytes_written: usize::default(),
+            millis_elapsed: usize::default(),
```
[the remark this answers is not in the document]
nyanzebra: Yes, we can change. I had this initially as atomic but for this pr can be duration. 

I still want to move locking here and remove locks up in the code... but can do later

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/block.rs
```diff
@@ -0,0 +1,332 @@
...
+        return Err(RingBufferError::Validate(BlockError::BlockHash {
+            found: actual_block_hash,
+            expected: block_hash,
+        })
+        .into());
+    }
+
+    let inner_block = block.inner;
+    let actual_data_size = inner_block.data_size();
+    let data_size = binary_serialize_size(&*data)?;
```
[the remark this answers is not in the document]
nyanzebra: Hmm weird.. should have gotten some warning about spurious deref

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,1129 @@
...
+    mem::size_of,
+    path::Path,
+    task::Waker,
+    time::{Duration, Instant},
+};
+
+use bincode::Result as BincodeResult;
+use error::BlockError;
+use memmap::MmapMut;
+use mqtt3::proto::Publication;
```
vadim-kovalyov, line 21: 1st party crate squeezed in :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,1129 @@
...
+    pointers: FilePointers,
+    // A waker for updating any pending batch after an insert.
+    waker: Option<Waker>,
+}
+
+impl RingBuffer {
+    pub(crate) fn new(
+        file_path: &Path,
+        max_file_size: usize,
+        flush_options: &FlushOptions,
```
vadim-kovalyov, line 92: nit: do we need ref if `FlushOptions` is Copy?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,1129 @@
...
+            let inner_block = block.inner();
+            let data_size = inner_block.data_size();
+            let index = inner_block.write_index();
+            start += block_size;
+            let end = start + data_size;
+            let data = load_data(&self.mmap, start, data_size, self.max_file_size)?;
+            start = end % self.max_file_size;
+            self.pointers.read_end = start;
+
+            validate(&block, &binary_serialize(&data)?)?;
```
vadim-kovalyov, line 236: Maybe we can avoid unnecessary serialization here. But this is something we can look into a bit later.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -0,0 +1,1129 @@
...
+        flush_options: &FlushOptions,
+    ) -> StorageResult<Self> {
+        let file = create_file(file_path).map_err(RingBufferError::MmapCreate)?;
+        file.set_len(max_file_size as u64)
+            .map_err(RingBufferError::MmapCreate)?;
+        let mmap = unsafe { MmapMut::map_mut(&file).map_err(RingBufferError::MmapCreate)? };
+        let (file_pointers, order) = find_pointers_and_order_post_crash(&mmap, max_file_size)?;
+
+        Ok(Self {
+            can_read_from_wrap_around: false,
```
vadim-kovalyov, line 101: I think `find_pointers_and_order_post_crash()` should also determine and return `can_read_from_wrap_around` flag as well. It is possible that the buffer was "full" at the time of shutdown/crash. So next time you start the ring buffer, you would think that it is "empty" since read==write.

vadim-kovalyov (dismissed):
I think this PR looks good to be merged. Great job of implementing a ring buffer and commenting the code! There are few things that I pointed out in the latest review, but they can be addressed in the following PRs. I think we should merge this PR asap to unblock your integration work.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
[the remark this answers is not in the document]
dmolokanov: ok

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -19,3 +22,13 @@ pub enum PersistError {
     #[error("Attempted to remove entry which does not exist")]
     RemovalForMissing,
 }
+
+// This might replace the PersistError or merge with it.
+#[derive(Debug, thiserror::Error)]
+pub enum StorageError {
+    #[error("RingBuffer error occurred {0}")]
```
dmolokanov, line 29: nit: what I did in my previous PR: I added `Caused by: {0}` into the error description in the `mqtt-bridge` crate. It'd be nice if you also did that for the code/error you added. it is fine if do it in the next PR.

```suggestion
    #[error("RingBuffer error occurred. Caused by: {0}")]
```

dmolokanov (approved):
Great job @nyanzebra! I really like how detailed your comments are!

dmolokanov (dismissed):

vadim-kovalyov (approved):

dmolokanov (approved):
:shipit:

PR #4306, opened 2021-01-28
cleaning identities

huguesBouvier:
Adding cleaning of identities and improved method of reverting capabilities.
Pipeline should not get stuck now.


builds/e2e/nested-e2e.yaml
```diff
@@ -107,25 +107,50 @@ stages:
...
         - run-idle -equals true 
-        - status -equals unlocked_$(Build.BuildId)  
-    strategy:
-      parallel: 3          
+        - status -equals unlocked_$(Build.BuildId)          
     steps:
-      - template: templates/nested-get-secrets.yaml     
-      - template: templates/unlock-test-agent.yaml      
+      - template: templates/nested-get-secrets.yaml  
+      - template: templates/unlock-test-agent.yaml
```
yophilav, line 135: I think it would be cleaner here instead of having a *.yaml with parameter input, you do a bash function that takes parameter array input. This would prevent the duplication of code.

scripts/linux/nested-edge-deploy-agent.sh
```diff
@@ -103,6 +103,8 @@ function setup_iotedge() {
     az iot edge set-modules --device-id ${DEVICE_ID} --hub-name ${IOT_HUB_NAME} --content ${deployment_working_file} --output none
 
     echo "Setup /var/secrets/aziot/keyd/device-id"
+    sudo mkdir -p /var/secrets/aziot
```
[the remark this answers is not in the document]
huguesBouvier: Done

builds/e2e/nested-e2e.yaml
huguesBouvier: Maybe, there is only 3 lvls. It is pretty subjective at that point. It would still be 2 scripts instead of 5.
The major downside is that you don't get separate red when there is a failure

builds/e2e/nested-e2e.yaml
```diff
@@ -107,25 +107,50 @@ stages:
...
     displayName: Clean up identities and unlock agents
     timeoutInMinutes: 1    
     variables:    
-      deviceLvl5Name: $[ dependencies.SetupVM_level5.outputs['deployIoTEdge.deviceName'] ]
-      deviceLvl5ConnectionString: $[ dependencies.SetupVM_level5.outputs['createIdentity.connectionString'] ]     
-      deviceLvl4Name: $[ dependencies.SetupVM_level4.outputs['deployIoTEdge.deviceName'] ]
-      deviceLvl4ConnectionString: $[ dependencies.SetupVM_level4.outputs['createIdentity.connectionString'] ] 
+      deviceLvl5DeviceId: $[ dependencies.SetupVM_level5.outputs['createIdentity.parentDeviceId'] ] 
+      deviceLvl4DeviceId: $[ dependencies.SetupVM_level4.outputs['createIdentity.parentDeviceId'] ] 
+      deviceLvl5IoTHubName: $[ dependencies.SetupVM_level5.outputs['createIdentity.iotHubName'] ]
```
davilu, line 120: Do we have use case for two different hubs?

huguesBouvier: no we don't. Out of scope for now

builds/e2e/nested-e2e.yaml
```diff
@@ -107,25 +107,50 @@ stages:
...
+      - template: templates/nested-clean-identity.yaml
+        parameters:
+          iotHubName: $(deviceLvl5IoTHubName)
+          deviceId: $(deviceLvl5DeviceId)
+          lvl: 5    
+      - template: templates/nested-clean-identity.yaml
+        parameters:
+          iotHubName: $(deviceLvl4IoTHubName)
+          deviceId: $(deviceLvl4DeviceId) 
+          lvl: 4
```
richma-ms, line 156: hmm gut feeling wonders if it'd be slightly safer to run cleanup in reverse order, though I can't seem to think of any concrete difference off the top of my head.

richma-ms (approved):

yophilav (approved):

builds/e2e/nested-e2e.yaml
davilu: So we won't need 2 variables, right? And hub name shouldn't require get back from job but load from config or key vault.

PR #4316, opened 2021-02-01
[mqtt] implement rpc integration test

dmolokanov:
Implements an integration test that emulates how  `EHCore` interacts with `upstream` via RPC commands.


vadim-kovalyov (approved):
LGTM! Just one question - do we want to have rpc tests in a separate module (separate from integration)?

PR #4318, opened 2021-02-02
Make sure we send will on broker shutdown.

vadim-kovalyov:
In this PR i've fixed the issue in the subj and added few more tests.


dmolokanov (approved):
lgtm

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -454,6 +454,7 @@ where
 
             // Ungraceful disconnect - send the will
             if let Some(will) = session.into_will() {
```
dmolokanov, line 456: should we make a new `fn try_send_will(&self, session)` that will contain logic and logging?

dmolokanov (approved):
looks good apart from some nits. I removed the label in case if you decide to fix them

dmolokanov (approved):
:shipit:

PR #4323, opened 2021-02-02
Fix partially succeeding Build Images Pipeline

and-rewsmith:
Here is the fix:

https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file
> When <clear /> is present for a given node, NuGet ignores previously defined configuration values for that node. Read more about how settings are applied.


dylanbronson (approved):
LGTM thanks!

PR #4324, opened 2021-02-02
Relax artifact download requirements to allow partially succeeded builds

and-rewsmith:
IMO, we should always allow partially succeeded builds when downloading artifacts. This will block the possibility that we don't notice tests are running old bits.


and-rewsmith:
Verified that this works to pull partially succeeded builds

dylanbronson (approved):
LGTM

PR #4326, opened 2021-02-03
Make sure we send will on broker shutdown. (#4318)

vadim-kovalyov:
In this PR i've fixed the issue in the subj and added few more tests.


and-rewsmith (approved):
LGTM (cherrypick)

PR #4327, opened 2021-02-03
Get deployment file from checked out branch

and-rewsmith:
We don't want to have to run the build images pipeline when changing things like deployment files. This change gets it from the checked out repo.


and-rewsmith:
Successful longhaul deployment:
https://msazure.visualstudio.com/One/_build/results?buildId=39060848&view=logs&j=7745b47e-e2a2-549a-b47a-b50c2a0d5e6c

builds/e2e/templates/longhaul-deploy.yaml
```diff
@@ -4,6 +4,7 @@ parameters:
   test.buildId: ''
```
dylanbronson, line 4: Can you change the other reference to Build.SourcesDirectory in this file (line 94) to be the parameter. It's a bit confusing otherwise to use both

dylanbronson: and 96

dylanbronson (commented):
nit: Checkout path might be somewhat confusing. I'll leave it up to you, but I feel that something like CheckedOutRepo or RepoPath might be more clear

builds/e2e/templates/longhaul-deploy.yaml
and-rewsmith: yeah good catch

and-rewsmith: I also updated the connectivity tests to read from this checkout path

dylanbronson (approved):
LGTM just make sure to test with the newest changes

PR #4329, opened 2021-02-03
Translate hostConfig.PidMode=host into podspec.hostPID=true.

darobs:
In Edge on k8s, translate `createOptions.HostConfig.PidMode=host` to `hostPID=true` in podSpec. Leaving it unset will not set the field, and setting it to anything other than "host" (case-insensitive) will set it to false.


zasherif (approved):
lgtm

PR #4335, opened 2021-02-03
Revert to use only root CA as trust bundle

gordonwang0:
Reverts a previous change that added the intermediate certs to the trust bundle. Only put the root CA cert in the trust bundle. The E2E tests were failing when intermediate certs were in the trust bundle, possibly because it wasn't validating the full chain.


dylanbronson (approved):
LGTM thank you !

PR #4338, opened 2021-02-03
Fix mqtt image build

and-rewsmith:
The mqtt image build broke when I refactored the way we copied built rust artifacts into the docker build context. It referenced a yaml template that I deleted. So the solution is to move to the new way of copying rust artifacts (i.e. the copy artifacts script).


and-rewsmith:
Passing mqtt image build:
https://msazure.visualstudio.com/One/_build/results?buildId=39095937&view=results

scripts/linux/consolidate-build-artifacts.sh
```diff
@@ -107,18 +111,36 @@ edge-hub)
...
+    # make build context structure
+    mkdir "$ARTIFACTS_DEST"
+    mkdir -p "$ARTIFACTS_DEST/$TARGET_AMD64_MUSL/release"
+    mkdir -p "$ARTIFACTS_DEST/$TARGET_ARM32V7/release"
+    mkdir -p "$ARTIFACTS_DEST/$TARGET_ARM64V8/release"
+
+    # copy mqtt artifacts
+    cp "$MQTT_ARTIFACTS_SOURCE/target/$TARGET_AMD64_MUSL/release/mqttd" "$ARTIFACTS_DEST/$TARGET_AMD64_MUSL/release" || true
+    cp "$MQTT_ARTIFACTS_SOURCE/target/$TARGET_ARM32V7/release/mqttd" "$ARTIFACTS_DEST/$TARGET_ARM32V7/release" || true
+    cp "$MQTT_ARTIFACTS_SOURCE/target/$TARGET_ARM64V8/release/mqttd" "$ARTIFACTS_DEST/$TARGET_ARM64V8/release" || true
```
and-rewsmith, line 139: This ```|| true``` pattern lets the step fail. That is useful for local builds, because you don't want to build the rust artifacts for every arch. It is also useful for this mqtt image build because they have separate jobs that build for each arch

I had a misunderstanding from before where I thought it wouldn't fail out because I removed the ```set error -e``` from the top of the script. Turns out it still does and we need || true.  This won't cause any issues because if artifacts are truly missing the docker build will fail

vadim-kovalyov (approved):

PR #4339, opened 2021-02-03
[mqtt] Implement shutdown/restart integration test

dmolokanov:
Adds an integration test that verifies the bridge can work properly after being restarted.
Moves RPC integration tests to appropriate module `rpc`.
Makes `mqtt_util` types exported as the high-level crate types.

### DIFF
--- a/mqtt/mqtt-bridge/tests/integration.rs
+++ b/mqtt/mqtt-bridge/tests/integration.rs
@@ -185,7 +185,7 @@ async fn bridge_settings_update() {
         .unwrap();
 
     // delay to propagate the update
-    // tokio::time::delay_for(Duration::from_secs(2)).await;
+    tokio::time::delay_for(Duration::from_secs(2)).await;
 
     // send upstream
     local_client
--- a/mqtt/mqtt-bridge/tests/integration.rs
+++ b/mqtt/mqtt-bridge/tests/integration.rs
@@ -1,30 +1,24 @@
-use std::{any::Any, convert::Infallible, time::Duration, vec};
+mod common;
 
-use bson::{doc, spec::BinarySubtype};
+use std::{any::Any, convert::Infallible, time::Duration};
+
 use bytes::Bytes;
 use futures_util::StreamExt;
 use matches::assert_matches;
+
 use mqtt3::{
     proto::{ClientId, QoS},
     ReceivedPublication,
 };
 use mqtt_bridge::{
-    settings::{BridgeSettings, Direction, TopicRule},
-    BridgeController, BridgeControllerHandle, BridgeControllerUpdate,
+    settings::{Direction, TopicRule},
+    BridgeControllerUpdate,
 };
 use mqtt_broker::{
     auth::{Activity, AllowAll, Authorization, Authorizer, Operation},
-    sidecar::Sidecar,
-    BrokerBuilder, BrokerHandle, ServerCertificate, SystemEvent,
+    SystemEvent,
 };
-use mqtt_broker_tests_util::{
-    client::TestClientBuilder,
-    server::{start_server, start_server_with_tls, DummyAuthenticator, ServerHandle},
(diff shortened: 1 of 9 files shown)


mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -1,30 +1,24 @@
-use std::{any::Any, convert::Infallible, time::Duration, vec};
+mod common;
```
and-rewsmith, line 1: Why declare the mod here? Isn't it usually in some lib.rs? But can we even have this under the tests folder?

mqtt/mqtt-util/src/lib.rs
```diff
@@ -1 +1,19 @@
...
+    clippy::must_use_candidate,
+    clippy::missing_errors_doc
+)]
+
+mod client_io;
+
+pub use client_io::{
+    AuthenticationSettings, ClientIoSource, CredentialProviderSettings, Credentials,
+    SasTokenSource, TcpConnection, TrustBundleSource,
+};
```
and-rewsmith, line 19: Good change :)

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -221,7 +219,7 @@ async fn bridge_settings_update() {
 #[tokio::test]
 async fn subscribe_to_upstream_rejected_should_retry() {
     let (mut local_server_handle, _, mut upstream_server_handle, upstream_broker_handle) =
-        setup_brokers(AllowAll, DummySubscribeAuthorizer(false));
+        common::setup_brokers(AllowAll, DummySubscribeAuthorizer(false));
 
     let subs = vec![
```
and-rewsmith, line 224: we duplicate these subs quite a bit, so we could define outside scope of an individual test for reuse

mqtt/mqtt-bridge/tests/rpc.rs
```diff
@@ -0,0 +1,108 @@
...
+                subtype: BinarySubtype::Generic,
+                bytes: payload,
+            },
+        );
+    }
+
+    let mut payload = Vec::new();
+    command.to_writer(&mut payload).unwrap();
+    payload.into()
+}
```
and-rewsmith, line 108: Assuming this test was only moved and not changed?

and-rewsmith (commented):
Looking good. I spotted some minor improvements

mqtt/mqtt-bridge/tests/integration.rs
dmolokanov: yeah, each integration test module is a separate compilation unit. so we need to declare share modules in each test module. at some point, we may want all our integration tests to be a single compilation unit but there will get different issues :)

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -189,7 +185,7 @@ async fn bridge_settings_update() {
         .unwrap();
 
     // delay to propagate the update
-    tokio::time::delay_for(Duration::from_secs(2)).await;
+    // tokio::time::delay_for(Duration::from_secs(2)).await;
```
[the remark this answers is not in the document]
dmolokanov: ouch! I haven't noticed myself. thanks!

mqtt/mqtt-bridge/tests/integration.rs
dmolokanov: while this is true for production code, for the test data it's better to have it inside the test instead.

mqtt/mqtt-bridge/tests/rpc.rs
vadim-kovalyov: 👍🏻

and-rewsmith (approved):

vadim-kovalyov (approved):
🚢

PR #4349, opened 2021-02-05
Stabilize PriorityQueueModuleToHub test

dylanbronson:
PriorityQueueModuleToHub end to end tests were flaky because we were getting some results from other tests.
This change makes sure all results we receive are from the current test by matching up a trackingId

### DIFF
--- a/test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
+++ b/test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
@@ -10,6 +10,7 @@ namespace Microsoft.Azure.Devices.Edge.Test
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Common;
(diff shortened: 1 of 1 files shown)


test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
```diff
@@ -54,7 +55,7 @@ public async Task PriorityQueueModuleToModuleMessages()
         }
 
         [Test]
-        [Category("Unstable")]
+        [Category("UnstableOnArm")]
```
and-rewsmith, line 58: Why is it unstable on arm?

and-rewsmith (approved):

test/Microsoft.Azure.Devices.Edge.Test/PriorityQueues.cs
dylanbronson: All the trc tests are unstable on arm32. Gets some resource unavailable exception. Haven't dove deep on it

PR #4350, opened 2021-02-05
add bridge restart test

ancaantochi:
- test for the case when upstream bridge is stopped and it should be recreated by controller
- return error in handle events if the client gets an error
- retry to create upstream bridge when new fails

### DIFF
--- a/mqtt/mqtt-bridge/src/client.rs
+++ b/mqtt/mqtt-bridge/src/client.rs
@@ -165,23 +165,21 @@ where
         let default_topics = self.event_handler.subscriptions();
         self.subscribe(default_topics).await?;
 
-        self.handle_events().await;
+        self.handle_events().await?;
         Ok(())
     }
 
-    async fn handle_events(&mut self) {
+    async fn handle_events(&mut self) -> Result<(), ClientError> {
         debug!("polling bridge client...");
 
-        while let Some(event) = self.client.try_next().await.unwrap_or_else(|e| {
-            // TODO: handle the error by recreating the connection
-            error!(error = %e, "failed to poll events");
-            None
-        }) {
+        while let Some(event) = self.client.try_next().await? {
             debug!("handling event {:?}", event);
             if let Err(e) = self.event_handler.handle(event).await {
                 error!(error = %e, "error processing event");
             }
         }
+
+        Ok(())
     }
 
     async fn subscribe(&mut self, topics: Vec<String>) -> Result<(), ClientError> {
--- a/mqtt/mqtt-bridge/src/controller/bridges.rs
+++ b/mqtt/mqtt-bridge/src/controller/bridges.rs
@@ -57,6 +57,10 @@ impl Bridges {
         self.bridges.push(Box::pin(task));
     }
 
+    pub(crate) fn is_empty(&self) -> bool {
+        self.bridges.is_empty()
+    }
+
     pub(crate) async fn send_update(&mut self, update: BridgeUpdate) {
         if let Some(config) = self.config_updaters.get_mut(update.name()) {
             if let Err(e) = config.send_update(update).await {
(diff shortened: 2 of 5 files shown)


mqtt/mqtt-bridge/src/controller/mod.rs
```diff
@@ -181,6 +196,7 @@ impl BridgeControllerHandle {
 pub enum BridgeControllerMessage {
     BridgeControllerUpdate(BridgeControllerUpdate),
     Shutdown,
+    ShutdownBridge(String),
```
vadim-kovalyov, line 198: nit: I would add some comments to each variant here.

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -479,3 +479,77 @@ async fn bridge_forwards_messages_after_restart() {
     upstream_client.shutdown().await;
     local_client.shutdown().await;
 }
+
+#[tokio::test]
+async fn recreate_upstream_bridge_when_fails() {
```
vadim-kovalyov, line 484: It would help if we can write up the scenario as a comment for the tests, similar like we do in `mqtt-broker/tests/compliance.rs`. Even if we can get the idea from the name, with comments it is easier to understand how test is working, what clients are being used, what ops performed and what is expected.

vadim-kovalyov (dismissed):
LGMT! code looks good, I have just a small wish to have more comments for the test scenario.

vadim-kovalyov (approved):
👍🏻 thanks!

mqtt/mqtt-bridge/src/client.rs
```diff
@@ -165,23 +165,21 @@ where
...
-    async fn handle_events(&mut self) {
+    async fn handle_events(&mut self) -> Result<(), ClientError> {
         debug!("polling bridge client...");
 
-        while let Some(event) = self.client.try_next().await.unwrap_or_else(|e| {
-            // TODO: handle the error by recreating the connection
-            error!(error = %e, "failed to poll events");
-            None
-        }) {
+        while let Some(event) = self.client.try_next().await? {
```
dmolokanov, line 175: I think we had a discussion already and agreed this code will never error out. Could you tell what changed? Did we miss something?

mqtt/mqtt-bridge/src/controller/bridges.rs
```diff
@@ -67,6 +71,16 @@ impl Bridges {
         }
     }
 
+    pub(crate) async fn shutdown_bridge(&mut self, name: &str) {
+        debug!("sending shutdown request to {} bridge...", name);
+
+        if let Some(bridge_handle) = self.bridge_handles.remove(name) {
+            bridge_handle.shutdown().await;
+        } else {
+            debug!("bridge {} not found", name);
```
dmolokanov, line 80: warn/error instead of debug?

mqtt/mqtt-bridge/src/controller/mod.rs
```diff
@@ -82,9 +82,13 @@ impl Sidecar for BridgeController {
         }
 
         loop {
-            let wait_bridge_or_pending = if bridges.is_terminated() {
-                // if no active bridges available, wait only for a new messages arrival
-                Either::Left(future::pending())
+            let wait_bridge_or_pending = if (bridges.is_terminated() || bridges.is_empty()) {
```
dmolokanov, line 85: I guess there should warnings for the redundant parenthesizes for if-expresssions.

mqtt/mqtt-bridge/src/controller/mod.rs
```diff
@@ -99,12 +103,16 @@ impl Sidecar for BridgeController {
                     bridges.shutdown_all().await;
                     break;
                 }
+                Either::Left((BridgeControllerMessage::ShutdownBridge(name), _)) => {
+                    info!("bridge {} shutdown requested", name);
+                    bridges.shutdown_bridge(name.as_ref()).await;
```
dmolokanov, line 108: nit
```suggestion
                    bridges.shutdown_bridge(&name).await;
```

mqtt/mqtt-bridge/src/controller/mod.rs
dmolokanov, line 85: not sure I understood the change. it used to be `future::pending` to make the sidecar running until `BridgeControllerMessage::Shutdown` received in case all bridges stopped.
now with your change, it is ready with some value right away which should cause the loop consuming CPU without a good reason. could you please verify it doesn't behave in the way I described?

PR #4356, opened 2021-02-05
Persist in-flight queue on broker restart

vadim-kovalyov:
We have an issue where we are losing in-flight qos1 messages on restart. In this PR I have included `SessionState::waiting_to_be_acked` queue into `SessionSnapshot` and made sure it is being persisted on restart.

Main changes are in:
- mqtt/mqtt-broker/src/persist.rs
- mqtt/mqtt-broker/src/session/state/mod.rs
- tests

The rest is just plumbing.

You might ask that `SessionState::waiting_to_be_acked` covers only qos1, and qos2 is not covered completely and still prone to the same issue. But we are not going to support qos2 for GA and will fix it separately later.

### DIFF
--- a/mqtt/mqtt-broker/src/persist.rs
+++ b/mqtt/mqtt-broker/src/persist.rs
@@ -4,7 +4,7 @@ use std::os::unix::fs::symlink;
 use std::os::windows::fs::symlink_file;
 use std::{
     cmp,
-    collections::HashMap,
+    collections::{HashMap, VecDeque},
     error::Error as StdError,
     fs::{self, OpenOptions},
     io::{Read, Write},
@@ -19,9 +19,13 @@ use flate2::{read::GzDecoder, write::GzEncoder, Compression};
 use serde::{ser::SerializeMap, Deserialize, Deserializer, Serialize, Serializer};
 use tracing::{debug, info, info_span};
 
-use mqtt3::proto::Publication;
+use mqtt3::proto::{Publication, Publish};
 
-use crate::{subscription::Subscription, BrokerSnapshot, ClientInfo, SessionSnapshot};
+use crate::{
+    proto::{PacketIdentifierDupQoS, QoS},
+    subscription::Subscription,
+    BrokerSnapshot, ClientInfo, SessionSnapshot,
+};
 
 /// sets the number of past states to save - 2 means we save the current and the pervious
 const STATE_DEFAULT_PREVIOUS_COUNT: usize = 2;
@@ -158,15 +162,20 @@ impl FileFormat for VersionedFileFormat {
     }
 }
 
+/// Actual representation of broker state data that is saved to disk.
 #[derive(Deserialize, Serialize)]
 struct ConsolidatedState {
     #[serde(serialize_with = "serialize_payloads")]
     #[serde(deserialize_with = "deserialize_payloads")]
     payloads: HashMap<u64, Bytes>,
-    retained: HashMap<String, SimplifiedPublication>,
+    retained: HashMap<String, PublicationRef>,
     sessions: Vec<ConsolidatedSession>,
 }
 
+/// In case a single message being delivered to multiple subscribers,
+/// we don't want to duplicate the payload for every session.
+/// So, when we convert from `BrokerSnapshot` to `ConsolidatedState` we
+/// store all the payloads separate, and reference them in `PublicationRef`
 impl From<BrokerSnapshot> for ConsolidatedState {
     fn from(state: BrokerSnapshot) -> Self {
         let (retained, sessions) = state.into_parts();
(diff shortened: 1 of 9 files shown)


mqtt/mqtt-broker/src/persist.rs
```diff
@@ -272,22 +294,68 @@ impl From<ConsolidatedState> for BrokerSnapshot {
...
+    waiting_to_be_sent: VecDeque<PublicationRef>,
+    waiting_to_be_acked: VecDeque<InFlightPublicationRef>,
     last_active: DateTime<Utc>,
 }
 
+/// Represents a queued publication
+/// that has a reference to a payload in `ConsolidatedState`
+/// but not yet in-flight.
+#[derive(Deserialize, Serialize)]
+struct PublicationRef {
```
dmolokanov, line 311: Not sure it is a good name for queued publication. Usually `Ref` suffix is used when there is some owned type `Foo` and `FooRef` which is a reference for `Foo`. For instance [X509](https://docs.rs/openssl/0.10.32/openssl/x509/struct.X509.html) and [X509Ref](https://docs.rs/openssl/0.10.32/openssl/x509/struct.X509Ref.html)

In this case, `PublicationRef` does not reference `Publication`

mqtt/mqtt-broker/src/persist.rs
```diff
@@ -272,22 +294,68 @@ impl From<ConsolidatedState> for BrokerSnapshot {
...
-struct SimplifiedPublication {
+struct InFlightPublicationRef {
     topic_name: String,
-    qos: crate::proto::QoS,
+    packet_identifier_dup_qos: PacketIdentifierDupQoS,
     retain: bool,
     payload: u64,
 }
 
+#[allow(clippy::mutable_key_type)]
```
dmolokanov, line 328: is it still an issue?

mqtt/mqtt-broker/src/session/state/queue.rs
```diff
@@ -129,8 +129,8 @@ impl LimitReached {
 impl Display for LimitReached {
     fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
         match self {
-            Self::Memory(_) => write!(f, "out of memory limits"),
-            Self::QueueLength(_) => write!(f, "out of queue length limits"),
+            Self::Memory(_) => write!(f, "message queue reached configured size limits, check 'max_queued_size' settings"),
```
dmolokanov, line 132: should we include limit number?

mqtt/mqtt-broker/src/snapshot.rs
```diff
@@ -39,28 +40,33 @@ impl SessionSnapshot {
...
         Self {
             client_info,
             subscriptions,
             waiting_to_be_sent,
+            waiting_to_be_acked,
             last_active,
         }
     }
 
+    #[allow(clippy::type_complexity)]
```
dmolokanov, line 55: well, I agree with clippy. just as an idea: can we get a mutable reference instead and make mem::take or mem replace for each field instead?

dmolokanov (commented):
I'm not sure I understand what is happening when the broker starts: we load messages in-flight from storage and then what? we re-send them with `dup` flag? or we just have them to remove from the session's state when receiving ack?

mqtt/mqtt-broker/src/persist.rs
vadim-kovalyov: @dmolokanov , any better suggestion? Previous name, IMO, was even more confusing. Every time I get back to this code I have to go over and recall what is responsible for what :)

The only other good name I can think of is `PayloadRef`, since this struct has a reference to the payload. If you don't have strong objections - I'd leave it as is.

mqtt/mqtt-broker/src/persist.rs
vadim-kovalyov: yes, it complains about `HashMap<Bytes, u64>` where it considers `Bytes` as mutable....

mqtt/mqtt-broker/src/session/state/queue.rs
vadim-kovalyov: good point

vadim-kovalyov:
> I'm not sure I understand what is happening when the broker starts: we load messages in-flight from storage and then what? we re-send them with `dup` flag? or we just have them to remove from the session's state when receiving ack?

@dmolokanov, when the broker starts, all persisted sessions are restored as "offline", since there are no connected clients. When (if) client is connected again, the session is converted `into_online` (offline.rs:56), and there we re-queue all inflight with `dup` flag.

mqtt/mqtt-broker/src/snapshot.rs
vadim-kovalyov: Yeah, it's not super pretty, and I was thinking about that as well. The main problem here is that we're trying to deconstruct it from other modules that don't have access to the private fields. 

I've tried to see how it's done in other places. I believe `pub(crate)` on fields is one approach, and having sub-struct `Parts` with pub fields is another one. 

Taking fields one by one with mem::tak is the new one that you're suggesting. But, IMO, it is even less prettier.

dmolokanov (approved):
lgtm

PR #4362, opened 2021-02-08
Updates bridge for RB integration

nyanzebra:
This sets up the bridge code for easier integration later with the RingBuffer code. 

One thing that I think can be thought a bit more about is whether we want both local and remote pumps to have the same storage settings or if they can/should be different.


mqtt/mqtt-bridge/config/default.json
```diff
@@ -5,6 +5,11 @@
         "subscriptions": [],
         "forwards": []
     },
+    "storage": {
+        "max_file_size": 1048576,
+        "batch_size": 10,
+        "file_name": "/tmp/mqttd/test.txt"
+    },
```
nyanzebra, line 12: I plan to add flush options later once RingBuffer comes in

mqtt/mqtt-bridge/src/controller/bridges.rs
```diff
@@ -39,7 +39,7 @@ pub(crate) struct Bridges {
 impl Bridges {
     pub(crate) async fn start_bridge<S>(&mut self, bridge: Bridge<S>, settings: &ConnectionSettings)
     where
-        S: StreamWakeableState + Send + 'static,
+        S: StreamWakeableState + Send + 'static + Sync,
```
nyanzebra, line 42: Even after format it didn't put lifetime last.. not sure if this is a style issue or not, but will change if it is

mqtt/mqtt-bridge/src/controller/mod.rs
```diff
@@ -109,20 +105,13 @@ impl Sidecar for BridgeController {
...
-                            ) {
-                                Ok(bridge) => {
-                                    bridges.start_bridge(bridge, upstream_settings).await;
-                                }
-                                Err(e) => {
-                                    error!(err = %e, "failed to create {} bridge", name);
-                                }
-                            }
-                        }
+                        start_bridge(
```
nyanzebra, line 108: Put common code in a helper

mqtt/mqtt-bridge/src/controller/mod.rs
```diff
@@ -153,6 +142,31 @@ async fn process_update(update: BridgeControllerUpdate, bridges: &mut Bridges) {
...
+    system_address: &str,
+    device_id: &str,
+    bridges: &mut Bridges,
+) {
+    if let Some(upstream_settings) = settings.upstream() {
+        let bridge_result = bridge_builder::<WakingMemoryStore>()
+            .with_system_address(String::from(system_address))
+            .with_device_id(String::from(device_id))
+            .with_connection_settings(upstream_settings.clone())
+            .build();
```
nyanzebra, line 156: I have made some builders for memory and rb, only memory is here until rb pr goes in. Let me know if there is a preference for doing something else here :)

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -1,16 +1,15 @@
+use super::StorageError;
+use crate::persist::{waking_state::StreamWakeableState, Key};
+use futures_util::stream::Stream;
+use mqtt3::proto::Publication;
+use parking_lot::Mutex;
 use std::{
```
nyanzebra, line 6: Whoops, need to fix the uses here :)

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -126,6 +127,86 @@ async fn send_message_upstream_downstream() {
     local_client.shutdown().await;
 }
 
+#[ignore = "This test can be activated once RingBuffer is merged in."]
+#[tokio::test]
+async fn send_message_upstream_with_crash_is_lossless() {
```
nyanzebra, line 132: The goal for this test is to simulate a crash and then come back up and if rb is configured then the messages pick off where they were before crash. If there are ideas on how to improve this test, let me know!

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -251,6 +259,27 @@ struct UpstreamSettings {
     subscriptions: Vec<Direction>,
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct StorageSettings {
+    max_file_size: usize,
+    file_name: String,
```
dmolokanov, line 265: is it just a file name or a full path to a file? if later let's make it `file_path: PathBuf`

mqtt/mqtt-bridge/tests/common.rs
```diff
@@ -75,6 +75,7 @@ pub async fn setup_bridge_controller(
     local_address: String,
     upstream_address: String,
     subs: Vec<Direction>,
+    id: &'static str,
```
dmolokanov, line 78:
```suggestion
    id: &str,
```
?

dmolokanov (dismissed):

mqtt/mqtt-bridge/src/bridge/builder.rs
```diff
@@ -0,0 +1,234 @@
...
+
+use crate::{
+    bridge::{Bridge, BridgeError},
+    client::MqttClientConfig,
+    persist::{PublicationStore, WakingMemoryStore},
+    pump::Builder,
+    settings::{ConnectionSettings, StorageSettings},
+};
+
+pub(crate) struct Yes;
```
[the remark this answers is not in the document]
vadim-kovalyov: I also haven't seen this pattern before. This is definitely an interesting approach where you restrict  method access and check required data via type system, although looks a bit verbose. I also think that we should reduce visibility of Yes/No structs if possible. Apart form that I don't have strong preferences.

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -45,12 +45,11 @@ where
         Self(inner)
     }
 
-    fn next_batch(&mut self) -> Result<VecDeque<(Key, Publication)>, PersistError> {
+    fn next_batch(&mut self) -> Result<VecDeque<(Key, Publication)>, StorageError> {
```
vadim-kovalyov, line 48: just curious why you're replacing one error type with another?

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -134,11 +132,12 @@ mod tests {
         };
 
         // insert elements
-        let mut state_lock = state.lock();
-        state_lock.insert(key1, pub1.clone()).unwrap();
-        state_lock.insert(key2, pub2).unwrap();
-        drop(state_lock);
-
+        let key1;
```
vadim-kovalyov, line 135: I don't have objections about this syntax, but just curious how is this better than inline declaration + drop(lock)?

mqtt/mqtt-bridge/src/persist/mod.rs
```diff
@@ -1,11 +1,13 @@
-use serde::{Deserialize, Serialize};
-
 mod loader;
 mod publication_store;
-mod waking_state;
+pub mod waking_state;
```
vadim-kovalyov, line 3: why exposing this?

mqtt/mqtt-bridge/src/persist/waking_state/memory/mod.rs
```diff
@@ -1,35 +1,36 @@
...
-use tracing::debug;
+pub mod error;
 
+use super::StorageResult;
+use crate::persist::{waking_state::StreamWakeableState, Key};
+use error::MemoryError;
 use mqtt3::proto::Publication;
-
-use crate::persist::{waking_state::StreamWakeableState, Key, PersistError};
+use std::{cmp::min, collections::VecDeque, task::Waker};
```
vadim-kovalyov, line 7: rearrange please :)

mqtt/mqtt-bridge/src/settings.rs
vadim-kovalyov: As I suggested above, let's make it `folder_path` instead.

mqtt/mqtt-bridge/tests/integration.rs
vadim-kovalyov: It would be great if you can write up the scenario as a test comment. Take a look how it's done for other tests in this file.

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -122,6 +123,89 @@ async fn send_message_upstream_downstream() {
...
+            upstream_client.publications().next().await,
+            Some(ReceivedPublication{payload, .. }) if payload == Bytes::from("from local")
+        );
+
+        // send upstream but disconnect afterwards
+        local_client
+            .publish_qos1("to/temp/1", "from local again", false)
+            .await;
+
+        controller_handle.shutdown();
```
vadim-kovalyov, line 183: So, there is a catch here... the main problem is that clients (`TestClient`) that you are using are "active". It means they do work in the background task and fetch/push messages as fast as they can. They also hide some protocol complexity by taking care of acks, etc.

So, in this case there is no guarantee that the upstream client did not receive the message prior to shutdown. It might have received that and kept it in it's local memory. Later on line 195 you could just get it from local memory and not the bridge as you might think.

What you need to do is to use `PacketStream` instead. `PacketStream` is "passive", and performs work only when asked. It also allows to have more control over the packets, like you're responsible for sending acks yourself with `PacketStream`. So, you can make one of the client to be `PacketStream` and don't send ack when you receive a publication. That will keep it in the bridge. 

Then you can restart the bridge, reconnect the client and see if message is being re-delivered from the storage

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -122,6 +123,89 @@ async fn send_message_upstream_downstream() {
...
+        controller_task.await.expect("controller task");
+    }
```
vadim-kovalyov, line 185: Help me understand please what is the benefit of nested scopes here? :)

mqtt/mqtt-bridge/tests/integration.rs
nyanzebra: Yes, will do

mqtt/mqtt-bridge/src/bridge/mod.rs
```diff
@@ -69,55 +68,17 @@ impl BridgeHandle {
...
-            local_pump,
-            remote_pump,
-        })
-    }
-}
-
 impl<S> Bridge<S>
 where
-    S: StreamWakeableState + Send,
+    S: StreamWakeableState + Send + Sync,
```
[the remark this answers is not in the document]
nyanzebra: Not needed, I stripped down an earlier version to get this out. These are all going away :)

mqtt/mqtt-bridge/src/bridge/builder.rs
```diff
@@ -0,0 +1,234 @@
...
+pub(crate) struct No;
+
+pub(crate) struct UpstreamBridgeBuilder<
+    Kind,
```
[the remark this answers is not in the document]
nyanzebra: Sure, sounds good to me

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -47,6 +49,7 @@ impl BridgeSettings {
         clean_session: bool,
         keep_alive: Duration,
     ) -> Result<Self, ConfigError> {
+        let mut me = Self::new()?;
```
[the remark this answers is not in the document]
nyanzebra: Works for me, will update

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -251,6 +259,27 @@ struct UpstreamSettings {
...
+    pub fn max_file_size(&self) -> usize {
+        self.max_file_size
+    }
+
+    pub fn file_name(&self) -> &str {
+        &self.file_name
+    }
+
+    pub fn batch_size(&self) -> usize {
+        self.batch_size
```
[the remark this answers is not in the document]
nyanzebra: Didn't know these existed! This is exactly what I want!

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: Should be more clear now, I had started with file_name for a path. However, what I think works better is a directory. Then we can create however many files we want and guarantee they are bunched together

mqtt/mqtt-bridge/tests/integration.rs
nyanzebra: Yeah, I found this out the hard way today :P I saw another test that just shutdowns the controller handle and the associated task and then restarts. I have now copied that version

mqtt/mqtt-bridge/tests/integration.rs
nyanzebra: I had done this to force dropping of any resources, from other comment, I see there is a problem with this kind of 'disconnect' and what I really want is to shutdown and then restart the handle like some of the other tests do

mqtt/mqtt-bridge/src/persist/waking_state/memory/error.rs
```diff
@@ -0,0 +1,7 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum MemoryError {
+    #[error("Bad key for removal")]
```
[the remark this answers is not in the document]
nyanzebra: Yes, of course!

mqtt/mqtt-bridge/src/pump/builder.rs
```diff
@@ -72,115 +72,123 @@ where
...
+                pump_handle,
+                remote_sub_handle,
+                remote_topic_mappers_updates,
+            );
+
+            let remote_pump = Pump::new(remote_messages_send, client, remote_store, messages)?;
+
+            Ok((local_pump, remote_pump))
+        } else {
+            Err(BridgeError::UnsetStorage)
```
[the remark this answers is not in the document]
nyanzebra: So, part of the problem now is that I am adding some new settings (StorageSettings) and I want to have a conversation on this. So, there are several options that come out of having these new settings. One, what should the default storage be? Should it be MemoryStore or  RingBuffer? Should people have to specify explicitly? Should Memory also start requiring storage settings for configuration?

mqtt/mqtt-bridge/src/persist/loader.rs
nyanzebra: It was more of an easy pattern to copy paste, as I do the same in for loops in other parts of the code since keys are returned from inserts now. I have no preference on this, I just copied over what I already started with, if want the other way I will change :)

mqtt/mqtt-bridge/src/persist/loader.rs
nyanzebra: It doesn't really matter which one we choose, Persist and Storage were always going to merge based on what I was coding/aiming for. I just picked the one I was already used to typing for things. All the StorageErrors can become PersistErrors, just be ready for the explosion in ring buffer code  :)

mqtt/mqtt-bridge/config/default.json
```diff
@@ -5,6 +5,11 @@
         "subscriptions": [],
         "forwards": []
     },
+    "storage": {
+        "max_file_size": 1048576,
+        "batch_size": 10,
+        "file_name": "/tmp/mqttd/test.txt"
```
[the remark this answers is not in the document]
nyanzebra: Yeah, I have found that directory works a lot better

mqtt/mqtt-bridge/src/bridge/builder.rs
```diff
@@ -0,0 +1,234 @@
...
+        debug!("created bridge {}...", connection_settings.name());
+
+        Ok(Bridge {
+            local_pump,
+            remote_pump,
+        })
+    }
+}
+
+pub(crate) fn bridge_builder<Kind>() -> UpstreamBridgeBuilder<Kind, No, No, No, No> {
```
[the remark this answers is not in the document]
dmolokanov: oops, yes.

mqtt/mqtt-bridge/src/bridge/builder.rs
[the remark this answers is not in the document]
dmolokanov: I guess we cannot use Type parameters with the visibility of less than the type it uses. And this a bit bothering when implementation details (Yes/No types) become a part of the "public" API of the struct.

mqtt/mqtt-bridge/src/pump/builder.rs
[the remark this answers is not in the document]
dmolokanov: Good question. Let's discuss it offline?

nyanzebra:
```rust
error: this argument is passed by value, but not consumed in the function body
   --> mqtt-bridge/src/upstream/rpc/local.rs:204:54
    |
204 |     fn it_captures_command_id(topic: &str, expected: Option<CommandId>) {
    |                                                      ^^^^^^^^^^^^^^^^^ help: consider taking a reference instead: `&Option<CommandId>`
    |
note: the lint level is defined here
   --> mqtt-bridge/src/lib.rs:2:22
    |
2   | #![deny(clippy::all, clippy::pedantic)]
    |                      ^^^^^^^^^^^^^^^^
    = note: `#[deny(clippy::needless_pass_by_value)]` implied by `#[deny(clippy::pedantic)]`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value
```
I am getting this clippy error ... but this is the code it is complaining about...

```rust
 #[test_case(r"$upstream/rpc/foo", Some("foo".into()); "when word")]
    #[test_case(r"$upstream/rpc/CA761232-ED42-11CE-BACD-00AA0057B223", Some("CA761232-ED42-11CE-BACD-00AA0057B223".into()); "when uuid")]
    #[test_case(r"$downstream/rpc/ack/CA761232-ED42-11CE-BACD-00AA0057B223", None; "when ack")]
    #[test_case(r"$iothub/rpc/ack/CA761232-ED42-11CE-BACD-00AA0057B223", None; "when wrong topic")]
    #[test_case(r"$iothub/rpc/ack/some id", None; "when spaces")]
    #[allow(clippy::needless_pass_by_value)]
    fn it_captures_command_id(topic: &str, expected: Option<CommandId>) {
        assert_eq!(capture_command_id(topic), expected)
    }
    ```

any ideas?

File: mqtt/mqtt-bridge/src/messages.rs [hunk 24]
```diff
@@ -147,8 +147,19 @@ where
 
                 if let Some(publication) = forward_publication {
                     debug!("saving message to store");
-                    self.store.push(&publication).map_err(BridgeError::Store)?;
 
+                    let result = self.store.push(&publication).map_err(BridgeError::Store);
```
vadim-kovalyov, line 151: may I suggest:
```rust

                if let Some(publication) = forward_publication {
                    debug!("saving message to store");

                    return match self.store.push(&publication) {
                        Ok(_) => Ok(Handled::Fully),
                        // If we are full we are dropping the message on ground
                        Err(PersistError::RingBuffer(RingBufferError::Full)) => Ok(Handled::Fully),
                        Err(err) => Err(BridgeError::Store(err))
                    };
                }
```

mqtt/mqtt-bridge/tests/integration.rs
```diff
@@ -128,6 +133,100 @@ async fn send_message_upstream_downstream() {
...
+        upstream_client.publications().next().await,
+        Some(ReceivedPublication{payload, .. }) if payload == Bytes::from("from local")
+    );
+
+    // send upstream but disconnect afterwards
+    local_client
+        .publish_qos1("to/temp/1", "from local again", false)
+        .await;
+
+    controller_handle.shutdown();
```
vadim-kovalyov, line 205: I believe there is no guarantee that this test does what you want. There is no way to prove that the message has not been already sent to upstream before shutdown. Maybe what you should do is this:
- shutdown upstream_server_handle and wait for upstream broker task - - it will guarantee that the upstream bridge loses connection to the upstream broker.
- publish a retain message - it will get to the bridge storage.
- [optional] - if you want to verify local restart - you can restart bridge controller at this step.
- restart the upstream broker.
- re-connect and re-subscribe the upstream client. - since message was "retained", it is guaranteed that re-connected client will receive it even if it re-subscribed with delay.
- assert that message received.

vadim-kovalyov (changes requested):
The changes looks good. But I believe the test needs to be fixed. Please see my comments on the test.

vadim-kovalyov (approved):
🚢

PR #4364, opened 2021-02-08
update aziot submodule

daprilik:
I wasn't 100% sure what to set for the new KeyUsage parameter in `create_key_if_not_exists`, so let me know if that needs to be changed.

### DIFF
--- a/edgelet/edgelet-http-workload/src/server/mod.rs
+++ b/edgelet/edgelet-http-workload/src/server/mod.rs
@@ -122,7 +122,7 @@ fn get_master_encryption_key(
         .create_key_if_not_exists(
             "iotedge_master_encryption_id",
             aziot_key_common::CreateKeyValue::Generate,
-            &[aziot_key_common::KeyUsage::Encrypt],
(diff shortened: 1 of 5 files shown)


massand (approved):

edgelet/edgelet-http-workload/src/server/mod.rs
```diff
@@ -121,7 +121,8 @@ fn get_master_encryption_key(
     key_client
         .create_key_if_not_exists(
             "iotedge_master_encryption_id",
-            aziot_key_common::CreateKeyValue::Generate { length: 32 },
+            aziot_key_common::CreateKeyValue::Generate,
+            &[aziot_key_common::KeyUsage::Encrypt],
```
arsing, line 125: It's used for derivation in `get_derived_enc_key_handle`, so this needs to be `::Derive`

arsing (approved):

massand (approved):
Just familiarized myself with the new key service API usage parameters, after @arsing rightfully pointed out the usage. Thanks! LGTM

PR #4367, opened 2021-02-08
Generic Mqtt Test Module: Clean up logging and fix report format

and-rewsmith:
Two changes:
- The logging can be improved by specifying the sequence number of the received message.
- Fixed the report format. The test type needs to be a string not an int, or report generation will fail.

### DIFF
--- a/test/modules/generic-mqtt-tester/src/lib.rs
+++ b/test/modules/generic-mqtt-tester/src/lib.rs
@@ -16,6 +16,7 @@ use std::{
     io::{self},
 };
 
+use bytes::Buf;
 use mqtt3::{PublishError, ReceivedPublication, UpdateSubscriptionError};
 use tokio::{sync::mpsc::error::SendError, sync::mpsc::Sender, task::JoinError};
 use trc_client::ReportResultError;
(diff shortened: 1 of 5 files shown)


test/modules/generic-mqtt-tester/src/message_channel.rs
```diff
@@ -141,7 +149,6 @@ where
             match future::select(received_pub, shutdown_signal).await {
                 Either::Left((received_publication, _)) => {
                     if let Some(received_publication) = received_publication {
-                        info!("received publication {:?}", received_publication);
```
and-rewsmith, line 144: We can remove this log since it is also getting logged when received by mqtt3 client.

test/modules/generic-mqtt-tester/src/lib.rs
```diff
@@ -91,3 +92,7 @@ pub enum MessageTesterError {
     #[error("received rejected subscription: {0}")]
     RejectedSubscription(String),
 }
+
+pub fn parse_sequence_number(publication: &ReceivedPublication) -> u32 {
+    publication.payload.slice(0..4).get_u32()
```
dylanbronson, line 97: Does this parse if an int is in the first 5 characters?

and-rewsmith: The payload is structured as a u32 then potentially a bunch of garbage data to simulate large sized messages. This gets serialized to bytes. The deserialization logic needs to get the first 4 bytes (size of u32) to get the u32 out of the serialized value.

dylanbronson (approved):
lgtm

PR #4375, opened 2021-02-09
Make Edge Agent's Prometheus parser more in line with upstream's spec.

### LINKED ISSUE #4361: Prometheus Regex doesn't match prometheus exposition format
Unfortunately the RegEx doesn't handle the optional trailing commas - which are ugly but allowed:

https://github.com/Azure/iotedge/blob/6d81b94d413484e41243469e2430bff75ae514ee/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs#L19-L24

Prometheus EBNF
https://prometheus.io/docs/instrumenting/exposition_formats/
```
metric_name [
  "{" label_name "=" `"` label_value `"` { "," label_name "=" `"` label_value `"` } [ "," ] "}"
] value [ timestamp ]
```

That means that a exposed metric like
```
edgeagent_total_disk_read_bytes{iothub=""lefitche-hub-3.azure-devices.net"",} 32908854
```
is absolutely valid. However the current parser RegEx doesn't match.

Please note that the Parser code seems to be duplicated in some projects - e.g. mcr.microsoft.com/azuremonitor/containerinsights/ciprod:iot-0.1.3.3

Please let me know if I was unclear or if you need further information.

arsing:
The previous regex did not allow whitespace between the tokens or
trailing commas after the last label value. It was also more permissive
with regard to allowed characters in metric names, label names and label values.

The regex also disallowed metrics with their own timestamps. With this change,
the timestamp is used if present, otherwise it defaults to the current time
as before.

The one visible change for metrics that already parsed successfully is that
escapes in the tag value are now evaluated by the parser, so that the published
metrics do not retain the unnecessary double-escaping.

Fixes #4361


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs
```diff
@@ -13,51 +13,135 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Util
...
+                    {
+                        continue;
+                    }
+
+                    double metricValue;
+                    if (!double.TryParse(match.Groups["metricvalue"]?.Value, out metricValue))
                     {
-                        double metricValue;
-                        if (!double.TryParse(match.Groups["metricvalue"]?.Value, out metricValue))
+                        continue;
```
and-rewsmith, line 85: This will silently over a matched metric if the value cannot be parsed to a double. Should we log a warning/error in this case?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs
```diff
@@ -13,51 +13,135 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Util
...
+                    @")?" +
+                    @"\}" +
+                    $"{Whitespace}*" +
+                @")?" +
+                MetricValue +
+                @"(?:" +
+                    $"{Whitespace}+" +
+                    MetricTimestamp +
+                @")?" +
+            @"$",
```
and-rewsmith, line 58: This is a lot more readable, nice change :)

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs
```diff
@@ -13,51 +13,135 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Util
...
+                            string result = string.Empty;
+                            bool escaped = false;
+                            foreach (char c in value.Value)
+                            {
+                                if (escaped)
+                                {
+                                    if (c == 'n')
+                                    {
+                                        result += "\n";
+                                    }
```
and-rewsmith, line 104: Do we have to handle other special characters according to prometheus spec? Seems like this approach could be error prone.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs
```diff
@@ -13,51 +13,135 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Util
...
+                    Dictionary<string, string> tags = tagNames.Zip(tagValues, (k, v) => new { k, v })
+                        .ToDictionary(x => x.k, x => x.v);
+
+                    string metricTimestampString = match.Groups["metrictimestamp"]?.Value;
+                    if (metricTimestampString.Length != 0)
+                    {
+                        long metricTimestamp;
+                        if (!long.TryParse(metricTimestampString, out metricTimestamp))
+                        {
+                            continue;
```
and-rewsmith, line 134: Same comment as above about silent continue

and-rewsmith (dismissed):
LGTM with some minor comments. We might want to log warning/error in the cases where we silently continue.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs
```diff
@@ -13,51 +13,135 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Util
     public static class PrometheusMetricsParser
     {
         // Extracts data from prometheus format. Read about prometheus format here:
```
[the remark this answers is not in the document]
arsing: No. The regex works fine after all.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs
arsing: That is existing code.

https://github.com/Azure/iotedge/blob/9b4666fa5912692405be7150ef127170970d1aab~2/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs

It looks new because I reduced the indentation. See the diff with whitespace changes ignored: https://github.com/Azure/iotedge/pull/4375/files?w=1 )

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs
arsing: The spec (and golang parser) only all `\\`, `\"` and `\n` escapes. With the exception of `\n`, the rest are just `result += c;`, so I didn't write explicit code for them. The regex already enforces that only those three escapes will be accepted anyway.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs
arsing: Since the existing code already silently swallowed metrics that didn't match the regex at all, as well as metrics with bad metric values (as I explained in your previous comment), I figured it would make sense for this too.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Diagnostics.Test/util/PrometheusMetricsParserTest.cs
```diff
@@ -196,5 +205,79 @@ public void TestNaN()
             Assert.Equal("somevalue.net", metrics.First().Tags["iothub"]);
             Assert.True(double.IsNaN(metrics.First().Value));
         }
+
+        [Fact]
+        public void TestInterstitialWhitespaceAndTrailingCommaAndEscapedTagValues()
```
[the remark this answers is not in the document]
arsing: Done.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Diagnostics/util/PrometheusMetricsParser.cs
arsing: I've added warnings to all three cases (regex doesn't match, metric value is malformed, metric timestamp is malformed).

and-rewsmith (approved):

PR #4377, opened 2021-02-09
Use parenthostname in network controller to set up rules

ancaantochi:
Modified network controller to use IOTEDGE_PARENTHOSTNAME to set the network rules so that connectivity is set for connecting to parent device. If not in nested environment the parent env won't be set so it uses IOTEDGE_IOTHUBHOSTNAME.


dylanbronson (approved):
LGTM. Can you add a description though, please

PR #4381, opened 2021-02-09
Added test to see if edgeHub keeps track on connected devices and their subscriptions

vipeller:


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/NestedConnectionTrackingTest.cs
```diff
@@ -0,0 +1,587 @@
...
+                            }
+                        }
+                        while (!subscriptionsAreOk && DateTime.Now - started < TimeSpan.FromSeconds(5));
+
+                        Assert.True(subscriptionsAreOk);
+                    }
+                }
+            }
+
+            // delay helyett inkabb azt kellene, hogy vizsgalni a cloud proxy call-t, hogy bekapcsolja-e a szubscriptiont (vagy ki)
```
vadim-kovalyov, line 335: ? :D

edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/NestedConnectionTrackingTest.cs
```diff
@@ -0,0 +1,587 @@
...
+                var newClient = new TestClient(isDirect, identity);
+
+                result.Add(newClient);
+            }
+
+            return result;
+        }
+
+        [Fact]
+        public async Task ConnectionsAndSubscriptionsAreBeingTracked()
```
vadim-kovalyov, line 189: it would be great to describe the scenario in method comments. It is very hard to follow such a huge test.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/handlers/DirectMethodHandler.cs
```diff
@@ -85,11 +85,13 @@ async Task<bool> HandleMethodResponse(Match match, MqttPublishInfo publishInfo)
...
             var res = match.Groups["res"];
 
+            var isDirect = string.Equals(match.Groups["dialect"].Value, MqttBrokerAdapterConstants.DirectTopicPrefix);
+
             var identity = id2.Success
                                 ? this.identityProvider.Create(id1.Value, id2.Value)
                                 : this.identityProvider.Create(id1.Value);
 
-            var maybeListener = await this.connectionRegistry.GetOrCreateDeviceListenerAsync(identity);
+            var maybeListener = await this.connectionRegistry.GetOrCreateDeviceListenerAsync(identity, isDirect);
```
vadim-kovalyov, line 94: will it affect identity translation scenario when a module publishes to $iothub/... dialect on-behalf-of a brownfield device?

vipeller: The role of this flag is to figure out that in what dialect EdgeHubCore should answer to a client.

Historically, there were only 'direct clients'. These clients showed up with a connection event from the broker, and handling that even in EdgeHub core, it created the Listener object for it (which is later was returned by the above call, but the method name was GetDeviceListenerAsync()).

Later we added nested clients. Nested clients are connected to a child edge, and when they connect, the broker in child edge generates the connection event - however no event will be generated in parent edge, and because of that, the DeviceListener object (which partially represent a client in edgeHub) will not be created. The parent edge will learn about the client only because it starts talking, e.g. sends a telemetry message. These handler objects (like the DirectMethodHandler whose code we are looking at) cannot call GetDeviceListenerAsync(), because it does not exist. So the method had been changed to GetOrCleateDeviceListenerAsync(), and I don't remember the details but the easy or obvious or practical solution was to get an isDirect parameter with a default false value. The logic behind this is that direct clients always get the connection event first, so their DeviceListener and other object will be created upfront, and this "surprise call" when those object are not ready to use, happens only for indirect clients (making IsDirect = false).

This all worked nicely until this test was written. However the test does not make a "connected" event for direct clients like in the real world, so those clients were created as indirect, and because of the effect of that flag, edgeHub started talking the indirect dialect (when it uses $iothub as prefix)

Once fix would have been generating the connected events to flag direct clients, however for me it seemed more general solution to decide this based on how the client talks. If it uses $edgehub as prefix, that means that the topic translation kicked in, which means that the client talked the "old dialect". As indirect clients connect through edgeHub, which always talks the "new dialect" upstream, this code treats old dialect as direct (and as a result it will use $edgehub as prefix in the responses, so the topic translator in the broker will turn the messages into old style.

vadim-kovalyov (approved):
🚀

PR #4382, opened 2021-02-09
bridge integration tests - address review comments

ancaantochi:
Addressed review comments from #4350


mqtt/mqtt-bridge/src/controller/mod.rs
```diff
@@ -82,13 +82,9 @@ impl Sidecar for BridgeController {
...
-            let wait_bridge_or_pending = if (bridges.is_terminated() || bridges.is_empty()) {
-                if (bridges.is_empty()) {
-                    // if no active bridges available, restart upstream bridge
-                    Either::Left(future::ready(Some((UPSTREAM.into(), Ok(Ok(()))))))
-                } else {
-                    Either::Left(future::ready(None))
-                }
+            let wait_bridge_or_pending = if bridges.is_terminated() {
+                // if bridges were terminated just wait for messages
+                Either::Left(future::pending())
```
dmolokanov, line 87: I just remembered one more reason for `pending` here. `BridgeController` is a sidecar and the sidecar should run until it is implicitly stopped. In case it is exited with error or success the whole process tears down all resources (broker, command handler and corresponding tasks) and exits.

dmolokanov (approved):
LGTM. Thanks for double-checking this!

PR #4383, opened 2021-02-09
nested connectivity pipeline

ancaantochi:
updated nested-connectivity.yaml to work with nested pipeline. Running connectivity stage doesn't need to lock the agent because it deploys and waits for the result.
Made changes to existing templates to be able to reuse for nested-connectivity.

E2E: https://msazure.visualstudio.com/One/_build/results?buildId=39567602&view=results
isa95: https://msazure.visualstudio.com/One/_build/results?buildId=39576447&view=results

TODO: investigate why tests are failing for mqtt, that can be fixed separately.


huguesBouvier (changes requested):
Can you please test this PR on the following pipelines:
nested Long haul
nested E2E tests
isa95 smoke test
And past the result in the PR?
Thanks!

huguesBouvier (approved):

huguesBouvier (approved):

PR #4393, opened 2021-02-10
Consolidate env vars for EdgeHub and Broker

vadim-kovalyov:
There are few issues with env vars that we have and below are the improvements I suggest. 	
- Not consistent naming - I suggest to land on PascalCase, since it's already most popular format for our existing vars. Please note this is pure documentation update. Our configuration implementation will be case insensitive.
- We use `:` in many settings where it is not well-supported on all platforms. Specifically it is hard to use on Linux - I suggest to move to `__` as a separator. This is a supported separator for dotnet core config provider and is back-compatible with `:`. So this, again, can be just documentation change, since both `:` and `__` will continue working in dotnet. For rust (broker code) we will also support both `:` and `__`.
- Some variables in broker needs to be changed from snake_case to PascalCase, while we have time before release. Those vars are:
  - `MqttBroker__MaxQueuedMessages`
  - `MqttBroker__MaxQueuedBytes`
  - `MqttBroker__MaxInflightMessages`
  - `MqttBroker__WhenFull`

Some other changes:
- Added support for `StorageFolder` env for Broker, so EH and Broker will store their state to the same folder.
- Updated `EnvironmentVariables.md`


mqtt/mqtt-edgehub/src/settings.rs
```diff
@@ -27,33 +26,65 @@ lazy_static! {
...
+            result.insert("broker.session.max_inflight_messages".into(), val.into());
+        }
+        if let Ok(val) = host_env.get::<String>("maxqueuedmessages") {
+            dbg!(&val);
             result.insert("broker.session.max_queued_messages".into(), val.into());
         }
-
-        if let Ok(val) = env::var("mqttBroker__max_queued_bytes") {
+        if let Ok(val) = host_env.get::<String>("maxqueuedbytes") {
+            dbg!(&val);
```
dmolokanov, line 74: ?

mqtt/mqtt-edgehub/src/settings.rs
```diff
@@ -27,33 +26,65 @@ lazy_static! {
...
+        host_env.merge(Environment::with_prefix("MqttBroker_").separator(":"))?;
+        host_env.merge(Environment::with_prefix("MqttBroker_").separator("__"))?;
+
+        dbg!(host_env.collect()?);
+
         let mut result: HashMap<String, config::Value> = HashMap::new();
-        if let Ok(val) = env::var("mqttBroker__max_queued_messages") {
+
+        // session
+        if let Ok(val) = host_env.get::<String>("maxinflightmessages") {
```
dmolokanov, line 65: Can we get `Value` from the config instead of getting a string and then converting to the `Value`?

mqtt/mqtt-broker/src/settings/mod.rs
```diff
@@ -160,21 +160,21 @@ impl Default for RetainedMessagesConfig {
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 pub struct SessionPersistenceConfig {
-    path: PathBuf,
+    folder_path: PathBuf,
```
dmolokanov, line 163: curious why not just `path`? it belongs to the persistence config section thus describes the path to persisted data. `file_path` is bad, I agree

doc/EnvironmentVariables.md
```diff
@@ -37,8 +37,8 @@
...
 |-----------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------|-------------------------------|
-| amqpSettings:enabled                    | Whether the AMQP protocol head should be enabled                                                                                                   | bool                                                                  | true                          |
-| amqpSettings:port                       | The port for the AMQP protocol head to listen on                                                                                                   | int32                                                                 | 5671                          |
+| amqpSettings__enabled                    | Whether the AMQP protocol head should be enabled                                                                                                   | bool                                                                  | true                          |
```
dmolokanov, line 40: Well, I'm confused here. In the description, it is said PascalCase is proposed, but changes in the doc do not reflect this statement.

shouldn't it be like
```suggestion
| AmqpSettings__Enabled                    | Whether the AMQP protocol head should be enabled                                                                                                   | bool                                                                  | true                          |
```

dmolokanov (changes requested):
There are several places where `dbg!` macros left

mqtt/mqtt-edgehub/src/settings.rs
vadim-kovalyov: :( sorry about that

doc/EnvironmentVariables.md
vadim-kovalyov: you're right, it should be fixed as well.

mqtt/mqtt-edgehub/src/settings.rs
vadim-kovalyov: good idea. It calls `T::deserialize` inside. I wonder how that would work if you call deserialize on self type :)

mqtt/mqtt-broker/src/settings/mod.rs
vadim-kovalyov: I just wanted to emphasize that his is a folder path, not a file path. IMO it is unclear from just "path".

varunpuranik (approved):

dmolokanov (approved):
:shipit:

PR #4398, opened 2021-02-10
Handle thumbprint auth through API proxy for WebSocket

richma-ms:
Add logic in the websocket middleware to check for client cert in the custom header that API proxy uses to forward the payload.

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/middleware/WebSocketHandlingMiddleware.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/middleware/WebSocketHandlingMiddleware.cs
@@ -91,7 +91,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Http.Middleware
                     }
                     catch (Exception ex)
                     {
-                        Events.InvalidCertificate(ex);
+                        Events.InvalidCertificate(ex, remoteEndPoint.ToString());
                         throw;
                     }
                 }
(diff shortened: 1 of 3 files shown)


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/middleware/WebSocketHandlingMiddleware.cs
```diff
@@ -109,6 +135,9 @@ enum EventIds
 
             public static void WebSocketRequestNoListener(string traceId, string correlationId) =>
                 Log.LogDebug((int)EventIds.BadRequest, Invariant($"No listener found for request {traceId}. CorrelationId {correlationId}"));
+
+            public static void InvalidCertificate(Exception ex) =>
+                Log.LogWarning((int)EventIds.InvalidCertificate, Invariant($"Invalid client certificate: {ex.Message}"));
```
huguesBouvier, line 140: Could you display the client ID? If we have several client might be useful to know the culprit.

huguesBouvier (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/middleware/WebSocketHandlingMiddleware.cs
```diff
@@ -72,6 +75,28 @@ async Task ProcessRequestAsync(HttpContext context, IWebSocketListener listener,
             var remoteEndPoint = new IPEndPoint(context.Connection.RemoteIpAddress, context.Connection.RemotePort);
 
             X509Certificate2 cert = await context.Connection.GetClientCertificateAsync();
+
+            if (cert == null)
+            {
+                // If the connection came through the API proxy, the client cert
+                // would have been forwarded in a custom header
+                if (context.Request.Headers.TryGetValue(Constants.ClientCertificateHeaderKey, out StringValues clientCertHeader) && clientCertHeader.Count > 0)
```
varunpuranik, line 83: Where is cert handling code for the non-WS scenario?

richma-ms: https://github.com/Azure/iotedge/blob/master/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Http/controllers/HttpRequestAuthenticator.cs#L45

huguesBouvier (approved):

PR #4406, opened 2021-02-11
1.50 Rust toolchain update

vadim-kovalyov:
In this PR I propose to change the strategy of using `stable` rust toolchain, and instead pin to a specific version and update it manually with every rust release.

For the last year or so we've been having build breaks on every single toolchain update. It is usually disruptive not only for those who work on the broker, but for others who can be stuck on a broken mqtt build pipeline.

The argument against this approach is that we will **forget** to update periodically, so I'll create a recurring task for the team to check if a new version of toolchain is available, and first available dev will take care of update :)

PS.
There is one clippy warning that I have suppressed for now: `clippy::unnecessary_wraps`. The fix will require some more changes that I would like to do in a separate PR. This PR is just to unblock people.


and-rewsmith:
I remember suggesting this a long time ago, but many people had strong opinions that this was a necessary forcing function to get us to stay up to date. I'm personally fine with this, but others might disagree.

vadim-kovalyov:
> I remember suggesting this a long time ago, but many people had strong opinions that this was a necessary forcing function to get us to stay up to date. I'm personally fine with this, but others might disagree.

Yeah, you were right :)

and-rewsmith (approved):

and-rewsmith (approved):

PR #4411, opened 2021-02-12
Stabilize BlockHeaders across architectures

nyanzebra:
This updates block and ring buffer to use u32 instead of usize. The problem with usize is that it is of different size on different architectures, which is why tests pass on 64bit but fail on 32bit. 

I have also changed the hashing algorithm to rustc-hash, which is pretty fast and doesn't have the caveat of the DefaultHasher in std that can be subject to change. Will post a link to a working build on 32bit here to prove correctness.

Some good ideas from all of this:
1. Add 32bit arch tests to PR pipeline for rust if we want to hit those targets
2. Maybe see if kcov can be added for code coverage reports


nyanzebra:
https://dev.azure.com/msazure/One/_build/results?buildId=39399443&view=results

vadim-kovalyov (approved):
As discussed, let's merge it for now to unblock the pipeline and create a bug to fix this properly.

PR #4413, opened 2021-02-12
TestResultCoordinator: Ignore recent unmatched results if in longhaul mode

and-rewsmith:
For longhaul, we cannot take the same approach as connectivity tests where we stop the test, allow everything in-flight to settle, then generate the result. Instead we need to be able to handle things that are in-flight, meaning some results will be unmatched.

This change allows the TRC to ignore any recently-received, unmatched expected results from the report generation. It accomplishes this through a filtering abstraction which ignores any expected result (and its actual result pair if present) that arrives in the last ignore-time. The default ignore-time is set to 1 minute.

There are really three conceptual changes in this PR:
1. The addition of the filtering abstraction, which includes changes in the settings, report generator factory, etc.
2. The conversion of ITestResultCollection to IAsyncEnumerable. There was a TODO in the ITestResultCollection class indicating we should convert this when we were on dotnet 3.1. But the main reason I did it was because of some convenience methods this provided for my filtering abstraction. If I didn't do it I think I'd be forced to do something hacky to extend the manual implementation of IAsyncEnumerable. This seemed messy so I decided to switch to IAsyncEnumerable, which bloated the diff size.
3. The conversion of the longhaul direct method reports to not allow for sender results without receiver results. This scenario won't exist in longhaul so we can eliminate some tests/logic.


test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
```diff
@@ -10,19 +10,14 @@ class DirectMethodLongHaulReport : TestResultReportBase
             string testDescription,
             string trackingId,
             string senderSource,
-            Option<string> receiverSource,
+            string receiverSource,
             string resultType,
             ulong senderSuccesses,
-            Option<ulong> receiverSuccesses,
+            ulong receiverSuccesses,
```
and-rewsmith, line 16: For longhaul this isn't optional

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReportGenerator.cs
```diff
@@ -39,13 +35,13 @@ class DirectMethodLongHaulReportGenerator : ITestResultReportGenerator
             this.ResultType = Preconditions.CheckNonWhiteSpace(resultType, nameof(resultType));
         }
 
-        internal Option<string> ReceiverSource { get; }
+        internal string ReceiverSource { get; }
```
and-rewsmith, line 38: For longhaul this isn't optional

test/modules/TestResultCoordinator/Reports/DirectMethod/Connectivity/NetworkStatusTimeline.cs
```diff
@@ -22,15 +22,16 @@ class NetworkStatusTimeline
         NetworkControllerStatus initialNetworkControllerStatus;
 
         public static async Task<NetworkStatusTimeline> CreateAsync(
-            ITestResultCollection<TestOperationResult> networkControllerTestOperationResults,
```
and-rewsmith, line 25: reminder: need to revert this

test/modules/TestResultCoordinator/Reports/StoreTestResultCollection.cs
```diff
@@ -2,75 +2,88 @@
...
+                {
+                    this.resultQueue.Enqueue(values.Item2);
+                    lastLoadedKey = values.Item1;
+                }
+
+                return lastLoadedKey;
+            }
         }
     }
 }
```
and-rewsmith, line 89: No code changes except converting to IAsyncEnumerable

PR #4414, opened 2021-02-12
Adding mqtt broker to long haul

dylanbronson:
Currently our nested long haul does not run with mqtt broker on. This PR splits the pipeline into 2 paths - 1 that sets up 3 AMQP vms and 1 that sets up 3 MQTT vms with the broker on. 
That way we can test our MQTT scenarios.
I've taken away logic for arm32 and arm64 since we were skipping it anyway. We will add it back in a later PR when we have arm32 and arm64 agents up and running. 
Example run: https://dev.azure.com/msazure/One/_build/results?buildId=39491736&view=results


builds/e2e/nested-longhaul.yaml
```diff
@@ -17,209 +17,35 @@ stages:
...
-      - SetupVM_level4
-      - Deploy_Longhaul_Linux_Amd64
-      - Deploy_Longhaul_Linux_Arm32
-      - Deploy_Longhaul_Linux_Arm64
+      - SetupVM_level5_amqp
+      - SetupVM_level5_mqtt
+      - SetupVM_level4_amqp
+      - SetupVM_level4_mqtt
+      - Deploy_Longhaul_Linux_Amd64_amqp
+      - Deploy_Longhaul_Linux_Amd64_mqtt
```
nyanzebra, line 39: Just for my understanding, do these account for level 3? I see level 3 in our pool below but not as a depends, so just curious about that :)

e2e_deployment_files/nested_long_haul_amqp_deployment.template.json
```diff
@@ -2,7 +2,7 @@
   "modulesContent": {
     "$edgeAgent": {
       "properties.desired": {
-        "schemaVersion": "1.0",
+        "schemaVersion": "1.1",
```
nyanzebra, line 5: Why not put both agent and hub at 1.2?

dylanbronson: For edgeHub, 1.2 is not publicly available and is just used in development, but we need it for broker. To be honest, I don't know what edgeAgent schema 1.2 is or what it means (or if it even exists), but 1.1 is documented, which is why I chose this. But I should definitely ask someone who knows the status of edgeAgent schema versions because if 1.2 does exist we should be testing with it

builds/e2e/nested-longhaul.yaml
dylanbronson: "Deploy_longhaul_linux_amd64_[amqp/mqtt]" are the names for the level 3 deployment. I'll change the name to reflect that

e2e_deployment_files/nested_long_haul_amqp_deployment.template.json
nyanzebra: Yeah, the reason why I ask is due to how I think our build pipelines work for creating images. As I think it creates edgehub and edgeagent. But I think Hugues/Bear might know best :)

builds/e2e/templates/nested-parent-vm-setup.yaml
```diff
@@ -1,20 +1,24 @@
+parameters:
+  transportType: 'amqp'
```
yophilav, line 2: I think you mean 
`  transportType: ''`

builds/e2e/nested-longhaul.yaml
```diff
@@ -17,235 +17,65 @@ stages:
...
-      deviceLvl4AgentName: $[ dependencies.SetupVM_level4.outputs['lock_test_agent.agentName'] ]       
-      amd64AgentName: $[ dependencies.Deploy_Longhaul_Linux_Amd64.outputs['lock_test_agent.agentName'] ]  
-      arm32AgentName: $[ dependencies.Deploy_Longhaul_Linux_Arm32.outputs['lock_test_agent.agentName'] ]  
-      arm64AgentName: $[ dependencies.Deploy_Longhaul_Linux_Arm64.outputs['lock_test_agent.agentName'] ]  
+      deviceLvl5AmqpAgentName: $[ dependencies.SetupVM_level5_amqp.outputs['lock_test_agent.agentName'] ]
+      deviceLvl5MqttAgentName: $[ dependencies.SetupVM_level5_mqtt.outputs['lock_test_agent.agentName'] ]
+      deviceLvl4AmqpAgentName: $[ dependencies.SetupVM_level4_amqp.outputs['lock_test_agent.agentName'] ]
+      deviceLvl4MqttAgentName: $[ dependencies.SetupVM_level4_mqtt.outputs['lock_test_agent.agentName'] ]
+      amd64AmqpAgentName: $[ dependencies.Deploy_Longhaul_Linux_Amd64_amqp.outputs['lock_test_agent.agentName'] ]
+      amd64MqttAgentName: $[ dependencies.Deploy_Longhaul_Linux_Amd64_mqtt.outputs['lock_test_agent.agentName'] ]
```
yophilav, line 48: You technically can use a matrix strategy to parallelize the run time and shrink the code base of the clean up process. It would require more work to do so... 
I'm gonna leave this here. You can make a call if you want to implement it or not.
https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#matrix

e2e_deployment_files/nested_long_haul_mqtt_deployment.template.json
```diff
@@ -0,0 +1,626 @@
...
+                "value": "<TestDuration>"
+              },
+              "testStartDelay": {
+                "value": "<TestStartDelay>"
+              },
+              "trackingId": {
+                "value": "<TrackingId>"
+              },
+              "transportType": {
+                "value": "Amqp"
```
yophilav, line 83: Do you intent to deploy an AMQP module under MQTT deployment? 
What is an expected behavior?

builds/e2e/templates/nested-parent-vm-setup.yaml
```diff
@@ -1,20 +1,24 @@
...
       parentDeviceId: ''
+      transportType: ${{ parameters.transportType }}
     pool:
      name: $(pool.name)
      demands:
-       - agent-group -equals $(agent.group)
+       - agent-group -equals $(agent.group)-${{ parameters.transportType }}
        - Agent.OS -equals Linux
        - Agent.OSArchitecture -equals X64
        - run-idle -equals true
```
yophilav, line 24: 1. `- run-idle -equals true` should no longer exist. It was meant to exist as my prove-of-concept for a locking scheme. Looks like the code got copied before I clean them up. Please remove. 
2. Please add a remark regarding `$(agent.group)-${{ parameters.transportType }}`'s vnet. Where are they? what VNET value?

scripts/linux/runE2ETest.sh
```diff
@@ -102,7 +102,7 @@ function get_long_haul_deployment_artifact_file() {
     local nestedEdgeTest=$(printenv E2E_nestedEdgeTest)
 
     if [[ ! -z "$nestedEdgeTest" ]]; then
-      path="$E2E_TEST_DIR/artifacts/core-linux/e2e_deployment_files/nested_long_haul_deployment.template.json"
+      path="$E2E_TEST_DIR/artifacts/core-linux/e2e_deployment_files/nested_long_haul_amqp_deployment.template.json"
```
yophilav, line 105: Nit: Please follow the naming convention of the deployment file
nestededge_{layer}Deployment_{protocol}.json

builds/e2e/nested-longhaul.yaml
```diff
@@ -17,235 +17,65 @@ stages:
...
-          testResultCoordinator.storageAccountConnectionString: '$(EdgeLonghaulStorageAccountConnString)'
-          quickstart.package.name: '$(quickstart.package.name)'
-          aziotis.artifact.name: '$(aziotis.artifact.name)'
-          aziotis.package.filter: '$(aziotis.package.filter)'
+    parameters:
+      transportType: amqp
+  - template: templates/nested-longhaul-deploy-amd64.yaml
+    parameters:
+      transportType: amqp
+  - template: templates/nested-parent-vm-setup.yaml
```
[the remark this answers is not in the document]
dylanbronson: We discussed this offline. It would require a refactor to the template yamls to make them into steps, not jobs, that is not worth it at the moment. AMQP/MQTT is already parallelized so we don't gain too much anyway

builds/e2e/templates/nested-parent-vm-setup.yaml
dylanbronson: Was trying to get default param. I don't think it works like this though

builds/e2e/nested-longhaul.yaml
dylanbronson: Discussed this offline - this would also require a large refactor. A lot of the jobs have dependencies that don't work quite well with matrix

e2e_deployment_files/nested_long_haul_mqtt_deployment.template.json
dylanbronson: For now let's keep this. It's good to see that AMQP works while broker is on. May want to change in future

scripts/linux/runE2ETest.sh
```diff
@@ -102,7 +102,7 @@ function get_long_haul_deployment_artifact_file() {
     local nestedEdgeTest=$(printenv E2E_nestedEdgeTest)
 
     if [[ ! -z "$nestedEdgeTest" ]]; then
-      path="$E2E_TEST_DIR/artifacts/core-linux/e2e_deployment_files/nested_long_haul_deployment.template.json"
+      path="$E2E_TEST_DIR/artifacts/core-linux/e2e_deployment_files/nestededge_bottomLayerBaseDeployment_long_haul_amqp.template.json"
```
davilu, line 105: name convention: `neste_dedge_bottom_layer_base_deployment_long_haul_amqp`?

scripts/linux/trcE2ETest.sh
```diff
@@ -739,7 +743,7 @@ function run_longhaul_test() {
 
 	local hash
 	hash=$(tr -dc 'a-zA-Z0-9' < /dev/urandom | head -c 8)
-	local device_id="$RELEASE_LABEL-Linux-$image_architecture_label-longhaul-$hash"
+	local device_id="$RELEASE_LABEL-Linux-$image_architecture_label-longhaul-$TRANSPORT_TYPE-$hash"
```
davilu, line 746: It's not necessary since `hash` is already random.

builds/e2e/nested-longhaul.yaml
```diff
@@ -17,235 +17,65 @@ stages:
...
+    parameters:
+      transportType: mqtt
+  - template: templates/nested-longhaul-deploy-amd64.yaml
```
davilu, line 28: Even we just test on `AMD64` now but should better keep the other 2 platforms `ARM32` and `ARM64` template in case we need to add them later.

davilu (dismissed):

scripts/linux/trcE2ETest.sh
dylanbronson: good point!

nyanzebra (dismissed):

scripts/linux/runE2ETest.sh
dylanbronson: I was following how the other deployment files did it - nestededge_middleLayerBaseDeployment.json

davilu (approved):

yophilav (approved):

nyanzebra (approved):

PR #4416, opened 2021-02-12
Fix unnecessary `Result` wraps.

vadim-kovalyov:
This is a follow up PR to fix some new 1.50 Clippy warnings. 

The main complain was about unnecessary `Result` wraps for the code that never returns `Err`


dmolokanov (approved):
🚀

PR #4424, opened 2021-02-16
Update tests for authorization required with KS and CS

gordonwang0:
Updates tests to set principals for the required auth with KS and CS.

Changes git submodule of aziot to a cargo dependency.

### DIFF
--- a/edgelet/Cargo.toml
+++ b/edgelet/Cargo.toml
@@ -15,5 +15,3 @@ members = [
     "management",
     "systemd",
 ]
-
-exclude = [ "aziot" ]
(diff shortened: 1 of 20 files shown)


arsing (approved):

edgelet/Cargo.toml
```diff
@@ -15,5 +15,3 @@ members = [
     "management",
     "systemd",
 ]
-
-exclude = [ "aziot" ]
```
massand, line 19: Curious - Doesn't this increase the build time?

massand (approved):
:shipit:

PR #4425, opened 2021-02-16
Removing arm from API proxy checking

huguesBouvier:
Removing arm from checkin from API proxy.

The checks are performed during core image built in.

Removing it because compilation time is about 1h.


vadim-kovalyov (approved):
👍🏻

PR #4429, opened 2021-02-16
File-based ring buffer

nyanzebra:
This makes a few changes to the RingBuffer implementation.

1. Switch to a (sync) File based RingBuffer. This allows for us to use large files.
2. Add a metadata file for speeding up initialization. As with large files, scanning byte by byte is not sufficient.
3. u32 to u64, as File IO api uses u64
4. BlockHeader -> BlockHeaderV1 

UTs have been updated accordingly, and these changes have been tested within the context of integration and verified data was being sent for over a day on my local setup.


mqtt/mqtt-bridge/src/persist/waking_state/mod.rs
```diff
@@ -6,6 +6,7 @@ use mqtt3::proto::Publication;
 use crate::persist::{Key, PersistError};
 
 pub mod memory;
+#[allow(dead_code)]
```
nyanzebra, line 9: Thought one of these was already here, let me figure out if I put another one by accident...

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/block.rs
```diff
@@ -27,32 +31,37 @@ pub const BLOCK_HINT: u32 = 0xdead_beef;
 /// validation.
 #[derive(Copy, Clone, Debug, Deserialize, Hash, Serialize)]
 #[repr(C)]
-pub(crate) struct BlockHeader {
+pub(crate) struct BlockHeaderV1 {
```
nyanzebra, line 34: Rename to V1 for versioning. Let me know if a different versioning format would be preferable!

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -35,9 +37,25 @@ pub type StorageResult<T> = Result<T, StorageError>;
...
-    read_end: u32,
+    write: u64,
+    read_begin: u64,
+    read_end: u64,
+}
+
+// Metadata that is saved to state file for rapidly picking up where
+// previous `RingBuffer` state was at.
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub(crate) struct RingBufferMetadata {
```
nyanzebra, line 48: When I was testing with large files, I found that initialization took a **long** time... I decided some sort of extra saved state would be useful for quickly getting back to a *close enough* location. Afterwards we can scan for a precise location. This way we don't have to scan large files, only a small portion.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -353,248 +427,293 @@ impl RingBuffer {
...
         .write(true)
         .create(true)
         .open(file_path)
 }
 
+fn save_ring_buffer_metadata(metadata: &RingBufferMetadata, file: &mut File) -> BincodeResult<()> {
+    let buf = bincode::serialize(metadata)?;
+    file.seek(SeekFrom::Start(0))?;
+    file.write_all(&buf)?;
+    file.flush()?;
```
nyanzebra, line 456: We always flush for metadata, as the alternative of having to scan the whole file (especially if large) felt unacceptable to me.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -353,248 +427,293 @@ impl RingBuffer {
...
-) -> StorageResult<(FilePointers, u128)> {
-    let mut block: BlockHeaderWithHash;
+    // mmap: &MmapMut,
+    file: &mut File,
+    max_file_size: u64,
+    best_guess_metadata: RingBufferMetadata,
+) -> RingBufferMetadata {
+    let block_size = *SERIALIZED_BLOCK_SIZE;
+
+    let mut start = best_guess_metadata.file_pointers.write;
```
nyanzebra, line 477: We should only need to do half the work as what this function did previously. This is because the metadata, if any, found us a location of a block close to where we left off.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -353,248 +427,293 @@ impl RingBuffer {
...
+    file.seek(SeekFrom::Start(0))?;
+    file.write_all(&buf)?;
+    file.flush()?;
+    Ok(())
+}
+
+fn retrieve_ring_buffer_metadata(file: &mut File) -> BincodeResult<RingBufferMetadata> {
+    let mut buf = vec![];
+    file.read_to_end(&mut buf)?;
+    if buf.is_empty() {
```
nyanzebra, line 463: If empty we will just have an empty/default metadata, which is fine as it means we can scan block at pos 0. As this should only be empty if we haven't had a successful operation yet.

nyanzebra:
Need to add in a checksum. Let me know if want it in this PR or in a separate one :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -165,154 +217,176 @@ impl RingBuffer {
...
-            validate(&block, &bincode::serialize(&data)?)?;
-            let key = Key { offset: index };
+            // Validate the block and data. This should be data section in the mmap
+            // after a `BlockHeaderWithHash`.
+            validate(&block, &bincode::serialize(&publication)?)?;
 
-            vdata.push_back((key, data));
+            #[allow(clippy::cast_possible_truncation)]
+            let key = Key {
+                offset: index as u32,
```
vadim-kovalyov, line 321: why u32?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/error.rs
```diff
@@ -21,8 +21,17 @@ pub enum RingBufferError {
...
+    #[error("Unable to create file. Caused by {0}")]
+    FileCreate(std::io::Error),
+
+    #[error("File IO error occurred. Caused by {0}")]
+    FileIO(std::io::Error),
+
+    #[error("File metadata unavailable")]
+    FileMetadata(std::io::Error),
+
+    #[error("File cannot be truncated.")]
```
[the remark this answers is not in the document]
vadim-kovalyov: btw, you don't have to address it in this PR, it is fine to fix it in the "integration" pr

vadim-kovalyov (approved):

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/block.rs
```diff
@@ -147,11 +163,11 @@ pub(crate) fn validate(block: &BlockHeaderWithHash, data: &[u8]) -> StorageResul
         .into());
     }
 
-    let inner_block = block.inner;
+    let BlockVersion::Version1(inner_block) = block.inner;
     let actual_data_size = inner_block.data_size();
 
     #[allow(clippy::cast_possible_truncation)]
```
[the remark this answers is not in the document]
nyanzebra: Ah you are right, had a bunch of these kind of clippy issues, so just added at most casts to resolve clippy issues in batch. Let me double check these :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/flush.rs
```diff
@@ -33,25 +33,26 @@ use serde::Deserialize;
...
 pub struct FlushState {
-    pub writes: u32,
-    pub bytes_written: u32,
+    pub writes: u64,
+    pub bytes_written: u64,
     pub elapsed: Duration,
 }
 
 impl FlushState {
     pub(crate) fn new() -> Self {
```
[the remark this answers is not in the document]
nyanzebra: Fair point

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: From previous fix pr, this is u32. This will be resolved as u64 again in my other integration pr

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
vadim-kovalyov: can metadata follow the same flushing strategy? In normal circumstances the buffer and metadata should always be in sync. The only issue is when process crashes/killed. Two things that should ensure consistency:
- treat metadata as an optimizatiopn but not the source of truth - that's what you're already doing.
- add checksum to metadata, disregard corrupted metadata and fallback to full scan (unfortunately).

What do you think?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: This happens frequently at each operation call. This is to a separate file as well. The problem I found when testing with larger files was the scanning of the file for where the pointers should be. Which could be very far off. This caused the time it took to initialized things be over a min depending on file size. Which feels really bad to me. If it is something we can be okay with, then this can be removed.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -165,154 +217,176 @@ impl RingBuffer {
...
         let mut start = read_index;
         let mut vdata = VecDeque::new();
         for _ in 0..count {
-            let end = start + block_size;
-            let bytes = &self.mmap[start as usize..end as usize];
-            // this is unused memory
-            if bytes.iter().map(|&x| x as usize).sum::<usize>() == 0 {
-                break;
-            }
+            #[allow(clippy::cast_possible_truncation)]
```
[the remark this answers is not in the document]
nyanzebra: For the block_size as usize? Yes, as block_size is u64 :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -353,248 +427,293 @@ impl RingBuffer {
...
+    // Read is found by scanning blocks marked for overwrite
+    // and taking the value of the first non-overwrite block.
 
+    // if we have more write_updates than read_updates and write == read
+    // then we must be in wrap_around case.
+    let mut write_updates = 0;
+    let mut read_updates = 0;
     #[allow(clippy::cast_possible_truncation)]
-    let block_size = *SERIALIZED_BLOCK_SIZE as u32;
+    let mut block = match load_block_header(file, start, block_size as usize, max_file_size) {
```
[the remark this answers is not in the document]
nyanzebra: This will not work as we are short-circuiting the method with the return and not for assigning to the block. I am okay to refactoring this into something else though :) Just let me know!

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -609,8 +728,9 @@ mod tests {
 
     use super::*;
 
-    const FLUSH_OPTIONS: FlushOptions = FlushOptions::Off;
-    const MAX_FILE_SIZE: u32 = 1024;
+    const FLUSH_OPTIONS: FlushOptions = FlushOptions::AfterEachWrite;
+    const MAX_FILE_SIZE: u64 = 1024;
+    const MAX_FILE_SIZE_NON_ZERO: NonZeroU64 = unsafe { NonZeroU64::new_unchecked(MAX_FILE_SIZE) };
```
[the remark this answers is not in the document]
nyanzebra: unwrap is not const... so we can try this later with a newer version of rust if it becomes const.

Exact error message: `std::option::Option::<T>::unwrap is not yet stable as a const fn`

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
dmolokanov: mins is a problem, I agree. several seconds is pretty much ok. do you know why it takes so long to process a large file? reading from SSD should be 100x MB/sec

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
dmolokanov: Yeah, I agree. Lets merge 2 threads https://github.com/Azure/iotedge/pull/4429/files#r577213662

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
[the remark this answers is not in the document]
dmolokanov: that's right. I guess it's not gonna happen soon :( https://github.com/rust-lang/rust/issues/51999

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Likely because it is more than just reading, it is the process of deserializing each block header and checking to see if it is a block closer to where the pointers should be or if we found the last block.

nyanzebra: I can spend some time to try and optimize the initial function after this pr and see if can get away from having to do this.. as it could very well be the case that I can try to avoid deserializing and it might be plenty fast after that.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Good point, didn't think about it like that. Let me go ahead and make that change too. As mentioned in another thread, it might be possible to get away from needing the metadata file altogether, but I was seeing long times for init. Might be having to deserialize a bunch, but not sure.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/error.rs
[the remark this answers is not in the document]
nyanzebra: Met somewhere in the middle :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Okay, so I made some changes locally to test without the hint file. I ran two tests on my M1 macbook with file sizes of 0.5GiB (it_errs_on_insert_when_full and it_errs_on_insert_when_full_and_had_batched_and_removed). These tests always start with fresh files. This means it would only use the initial first loop from earlier code (scanning for a block, but not scanning for most recent). These combined took 611s (10 min!!!). So, I am okay with trying more to get this going faster, as I still suspect deserialization as a large amount of the slowness... but this is the reason why I added the metadata file hint.

vadim-kovalyov: as an option, we can write metadata only on graceful shutdown. On crash - we will have to re-scan the file. Although, we still need to try to optimize the scanning later.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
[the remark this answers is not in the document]
vadim-kovalyov: I think we should avoid that. If we **must** cast block_size to usize, then we need to make SERIALIZED_BLOCK_SIZE to be usize, and cast it to u64 when needed w/o clippy warning suppression.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
vadim-kovalyov: As I mentioned above: another option is to write metadata only on graceful shutdown, and re-scan the file on crash. Checksum on metadata is good to have in any case.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -353,248 +428,293 @@ impl RingBuffer {
...
+                };
             }
         };
     }
 }
 
 fn load_block_header(
-    mmap: &MmapMut,
-    mut start: usize,
+    // mmap: &MmapMut,
```
vadim-kovalyov, line 583: remove comment?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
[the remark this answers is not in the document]
vadim-kovalyov: Okay, I'm taking my words back. I've looked at `load_block_header` and it always casts `size as u64`. So no point in casting `block_size as usize`. Let's fix that here.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Done :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -58,52 +76,89 @@ pub(crate) struct FilePointers {
...
+            create_file(metadata_file_path).map_err(RingBufferError::FileCreate)?;
+
+        // Retrieving from metadata file is only best effort.
+        let best_guess_metadata = retrieve_ring_buffer_metadata(&mut metadata_file)?;
+
+        // For correctness, need to scan after best guess and see if can get more accurate.
+        let metadata =
+            find_pointers_and_order_post_crash(&mut file, max_file_size, best_guess_metadata);
+
+        file.seek(SeekFrom::Start(0)).unwrap();
```
vadim-kovalyov, line 146: test code? :)

vadim-kovalyov (approved):

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Whoops~ removing :P

vadim-kovalyov (approved):

mqtt/mqtt-bridge/src/persist/loader.rs
```diff
@@ -201,8 +201,7 @@ mod tests {
         let mut state_lock = state.lock();
         let num_elements = 10_usize;
         for i in 0..num_elements {
-            #[allow(clippy::cast_possible_truncation)]
-            let key = Key { offset: i as u32 };
+            let key = Key { offset: i as u64 };
```
dmolokanov, line 204: offset is already u64, right?
```suggestion
            let key = Key { offset: i };
```

dmolokanov (dismissed):

dmolokanov (approved):
:shipit:

PR #4430, opened 2021-02-17
Update agent version to 1.1 in config.yaml

damonbarry:
Since 1.1 is released, this change updates the default agent tag in config.yaml to 1.1.


yophilav (approved):
As a follow up to this PR, maybe we should also update the default value for E2E test to be pulling for 1.1 images along with the documentation?

PR #4431, opened 2021-02-17
Upgrade hosted Ubuntu agents to 18.04

damonbarry:
Replace some remaining references to Ubuntu 16.04 with 18.04.

Note our release pipeline for docker images (`images-release.yaml`) calls `install.sh` with the `--package-arm` argument, which hasn't been updated for 18.04. That's a bigger effort that would need to be resolved before this can be merged (or handled in a separate PR).


vadim-kovalyov (commented):
> Replace some remaining references to Ubuntu 16.04 with 18.04

@damonbarry I think we need to fix `scripts/linux/generic-rust/install.sh` to have this line instead
```
sudo apt-get install --yes --target-release bionic-updates libssl-dev
```

damonbarry:
> @damonbarry I think we need to fix `scripts/linux/generic-rust/install.sh` to have this line instead
> 
> ```
> sudo apt-get install --yes --target-release bionic-updates libssl-dev
> ```

Yeah I moved this PR to draft status when I saw all the build failures. 😉 I think it might require a lot more changes than just "xenial-updates" --> "bionic-updates" but I'll give it a try...

scripts/linux/generic-rust/install.sh
```diff
@@ -77,26 +77,27 @@ process_args "$@"
...
-# Add trusty repo to get older version of libc6-armhf-cross
-sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu/ trusty main universe"
-
 # Install OpenSSL, curl and uuid and valgrind
 sudo apt-get update || :
 sudo apt-get install -y \
     pkg-config \
     uuid-dev curl \
     libcurl4-openssl-dev \
+    libssl-dev \
```
vadim-kovalyov, line 86: we need to make sure that this installs 1.1.1 version, and not 1.1.0 which is not supported anymore, AFAIK.

damonbarry: Not supported by whom?

Looks like the regular bionic distribution has this:
```
bionic (18.04LTS) (libdevel): Secure Sockets Layer toolkit - development files
1.1.1-1ubuntu2.1~18.04.7 [security]: amd64 i386
1.1.0g-2ubuntu4 [ports]: arm64 armhf ppc64el s390x
```
...whereas bionic-updates has this:
```
bionic-updates (libdevel): Secure Sockets Layer toolkit - development files
1.1.1-1ubuntu2.1~18.04.7: amd64 arm64 armhf i386 ppc64el s390x
```
So I could add bionic-updates to get 1.1.1 across all CPU architectures that we care about.

vadim-kovalyov: I believe there was an internal conversation about that a while back... I can try and dig that up...  Also, on the website they say it is not supported anymore - https://www.openssl.org/news/vulnerabilities-1.1.0.html

I believe that's why we were using `xenial-updates` repository for 16.04.

damonbarry: Looking through our YAML pipelines, this script always runs on Ubuntu amd64. So we should be okay, given that `bionic` and `bionic-updates` repositories provide the same version of libssl-dev for amd64. I will leave it the way it is here, unless you feel differently.

Other options:
- Specify `libssl-dev=1.1.1*` to make sure we get the right version (the command will fail if the default repository can't provide the right version)
- Specify `libssl-dev/bionic-updates` to always get this package from the updates repo

vadim-kovalyov: It sounds good to me to leave it as is, if we always run on Ubuntu amd64. I would just feel a bit more comfortable if more folks took a look at this change, like @arsing, @myagley, @and-rewsmith (prior contributors on that file).

arsing: Just `libssl-dev` is correct.

1. AFAIU, the difference between `foo` and `foo-updates` is for non-security-critical updates. `foo` won't serve packages with security issues even if `foo-updates` has a newer version.

  In general, LTS distros backport security fixes without bumping the version number, so regardless of whether upstream consideres 1.1.0 to be out of support or not, if the distro's repo serves 1.1.0 it's supported and secure.

1. There is no binary incompatibility between 1.1.0 and 1.1.1, so compiling against 1.1.0 and running against 1.1.1 is fine anyway.

arsing (dismissed):
It seems fine to me, but it depends on whether all the modified builds still work or not.

vadim-kovalyov (dismissed):

damonbarry:
> It seems fine to me, but it depends on whether all the modified builds still work or not.

Besides the pipelines that GitHub automatically runs against this PR (which are passing), I've confirmed these pipelines pass:
```
builds/ci/dotnet.yaml
builds/ci/edgelet.yaml
builds/ci/mqtt.yaml
builds/misc/images.yaml
```

There are a few that I didn't run:
`builds/e2e/stresstest.yaml`: quasi-abandoned right now, so I only updated it for completeness
`builds/misc/images-release.yaml`: designed to be run only at release time. so difficult to verify right now, but I don't anticipate any problems (especially because `builds/misc/images.yaml` passes)
`builds/service/service-deployment.yaml`: designed to run against the current release branch (1.1) not master, so I'm just keeping this one in sync with the version in `release/1.1`
`builds/e2e/nested-connectivity.yaml`: #4383 is much more comprehensive and makes the same change to this pipeline, so I've reverted my change here.

arsing (approved):

PR #4433, opened 2021-02-17
Fix casing mistakes in env var doc

richma-ms:
Found some casing mistakes a few days back, taking the time now to do a pass on the env vars to fix them.


vadim-kovalyov (approved):
🚀

PR #4436, opened 2021-02-17
Parallelize upstream REST calls during device scope cache refresh

richma-ms:
During the breadth-first expansion of the nested hierarchy, each call to get the device scope of a child Edge were previously serialized based on the queue implementation.  We can make this more optimized by concurrently firing off all currently enqueued calls, then merge the results together.  This effectively makes it so that we're fetching the hierarchy a full level at a time, instead of one device at a time.


dylanbronson (approved):
LGTM

PR #4437, opened 2021-02-18
Iotedge system commands

lfitchett:
This adds commands for managing the host runtime services. It contains restart, status, set-log-level and restart.


edgelet/iotedge/src/system.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use crate::error::{Error, ErrorKind};
+use aziotctl_common::{
+    get_status, get_system_logs as logs, restart, set_log_level as log_level, ServiceDefinition,
+    SERVICE_DEFINITIONS as IS_SERVICES,
+};
+
+lazy_static! {
+    static ref IOTEDGED: ServiceDefinition = {
+        let sockets: &'static [&'static str] = option_env!("IOTEDGE_HOST").map_or(&[], |_host| {
```
lfitchett, line 13: @arsing I wasn't sure if this is right. I know CentOS doesn't use the sockets, but is CentOS the only distro that doesn't set "IOTEDGE_HOST"?

arsing: CentOS is the only distro that *does* set `IOTEDGE_HOST`

daprilik:
It looks like you may have accidentally re-added the `edgelet/hsm-sys` submodule. That code lives in i-i-s, and shouldn't be re-added to iotedge

edgelet/iotedge/src/lib.rs
```diff
@@ -16,6 +16,9 @@
     clippy::use_self
 )]
 
+#[macro_use]
+extern crate lazy_static;
```
daprilik, line 20: There is hardly ever a reason to use `extern crate` in Rust 2018. Just import `lazy_static::lazy_static` when you need to use it.

edgelet/iotedge/src/main.rs
```diff
@@ -388,6 +422,37 @@ fn run() -> Result<(), Error> {
...
+                    if let Ok(level) = log::Level::from_str(log_level) {
+                        System::set_log_level(level)
+                    } else {
+                        eprintln!(
+                            r#"Error: log_level must be one of "trace", "debug", "info", "warn", or "error""#
+                        );
+                        std::process::exit(1);
+                    }
+                } else {
+                    eprintln!("Error: missing value for log_level");
```
daprilik, line 446: instead of doing this error handling manually, just pass `.required()` to clap when defining the `log_level` flag.

edgelet/iotedge/src/system.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use std::ffi::OsStr;
+
+use crate::error::{Error, ErrorKind};
+use aziotctl_common::{
+    get_status, get_system_logs as logs, restart, set_log_level as log_level, ServiceDefinition,
+    SERVICE_DEFINITIONS as IS_SERVICES,
+};
```
daprilik, line 9: `crate` imports should be in a separate group after external deps (in this case: aziotctl_common)

edgelet/iotedge/src/system.rs
```diff
@@ -0,0 +1,65 @@
...
+            .chain(IS_SERVICES.iter().copied())
+            .collect();
+
+        service_definitions
+    };
+}
+
+pub struct System;
+
+impl System {
```
daprilik, line 35: is there a reason this needs to be a struct + impl block vs. just having a set of freestanding functions?

edgelet/iotedge/src/system.rs
```diff
@@ -0,0 +1,65 @@
...
+            &["aziot-edged.mgmt.socket", "aziot-edged.workload.socket"]
+        });
+
+        ServiceDefinition {
+            service: "aziot-iotedgd.service",
+            sockets,
+        }
+    };
+    static ref SERVICE_DEFINITIONS: Vec<&'static ServiceDefinition> = {
+        let iotedged: &ServiceDefinition = &IOTEDGED;
```
daprilik, line 23: since IOTEDGED is only used from this `static ref`, you aught to just inline the definition.

daprilik (changes requested):

edgelet/iotedge/src/system.rs
lfitchett: It was cleaner to export and matched the export pattern of the other files better

edgelet/iotedge/src/system.rs
lfitchett: What do you mean by inline it? SERVICE_DEFINITIONS is a vec of refs, so the Service_Definition (IOTEDGED in this case) needs to be static in order to live long enough.

daprilik: ahhhh, true. okay, carry on.

edgelet/iotedge/src/system.rs
daprilik: alright, fair enough.

daprilik (dismissed):
:+1: 

Don't merge just yet, Arnav should also sign off on it.

edgelet/iotedge/src/system.rs
```diff
@@ -0,0 +1,69 @@
...
+
+lazy_static! {
+    static ref IOTEDGED: ServiceDefinition = {
+        let sockets: &'static [&'static str] = option_env!("IOTEDGE_HOST").map_or(
+            &["aziot-edged.mgmt.socket", "aziot-edged.workload.socket"],
+            |_host| &[],
+        );
+
+        ServiceDefinition {
+            service: "aziot-iotedgd.service",
```
arsing, line 22: `aziot-edged.service`

(You didn't test this, did you? :) )

edgelet/iotedge/src/system.rs
```diff
@@ -0,0 +1,69 @@
...
+use aziotctl_common::{
+    get_status, get_system_logs as logs, restart, set_log_level as log_level, ServiceDefinition,
+    SERVICE_DEFINITIONS as IS_SERVICES,
+};
+
+use crate::error::{Error, ErrorKind};
+
+lazy_static! {
+    static ref IOTEDGED: ServiceDefinition = {
+        let sockets: &'static [&'static str] = option_env!("IOTEDGE_HOST").map_or(
```
arsing, line 16: Add a comment like:

```rust
// Use the presence of IOTEDGE_HOST to infer whether this is being built for CentOS 7 or not.
// CentOS 7 doesn't use socket activation.
```

edgelet/iotedge/src/system.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    pub fn set_log_level(level: log::Level) -> Result<(), Error> {
+        log_level(&SERVICE_DEFINITIONS, level).map_err(|err| {
+            eprintln!("{:#?}", err);
+            Error::from(ErrorKind::System)
+        })
+    }
+
+    pub fn get_system_status() -> Result<(), Error> {
```
arsing, line 63: So what happens on CentOS 7 if `aziot-edged.service` is `inactive` ? Does this succeed even though there are no sockets around to start it up?

arsing (changes requested):

edgelet/iotedge/src/system.rs
arsing: (If the answer is "yes", don't fix that here by special-casing `IOTEDGED`. Fix it in i-i-s for any `ServiceDefinition` with 0 `socket`s.)

edgelet/iotedge/src/system.rs
lfitchett: I tested for the aziot commands, but I didn't have a 1.2 edge install handy 😅

edgelet/iotedge/src/system.rs
lfitchett: yes: https://github.com/Azure/iot-identity-service/pull/119/files

edgelet/iotedge/src/system.rs
lfitchett: ok, I installed a current version of iotedge 1.2 and tested the commands. It all works, but it brought an interesting question: 

If iotedge fails to start, it will never start IS. This means if both edged and identityd are unable to start, status wouldn't start reporting the identityd failure until edged was fixed (b/c identityd is never started).

I think this is fine, but its something to note maybe

arsing: If edged doesn't start specifically because it crashes before trying to start IS, then that's likely an edged-specific problem anyway. And if edged doesn't start because it couldn't get IS to start, then both would be `failed`. So it should be fine.

arsing (approved):

PR #4439, opened 2021-02-18
Consolidate log levels b/w EdgeHub and Broker

vadim-kovalyov:
## Context:	
We have `RuntimeLogLevel`  env var that can be set to one of the following: `fatal`, `error`, `warning`, `info`, `debug`, `verbose`.
- We have BROKER_LOG and RUST_LOG undocumented env vars that affect broker logs.
- Both dotnet core and rust support some sort of logging customization via so called log filters.	
    - For rust it looks like  "warn,test::foo=info,test::foo::bar=debug"
    - For dotnet we use Serilog, but I don't see a way how to configure it for edgehub container, apart from log levels.

## Proposal in this PR is:
- To keep the current `RuntimeLogLevel` environment and it's values for back-compat reasons,
- Make Broker respect `RuntimeLogLevel` only and disregard RUST_LOG or any other env vars.
- Add the ability to configure `RuntimeLogLevel` with additional "rust" style filters. For example:
	- RuntimeLogLevel="debug" - everything is set to debug.
	- RuntimeLogLevel="info,mqtt_broker::broker=debug" - everything is set to info, but code in mqtt_broker::broker module is set to debug.
	- RuntimeLogLevel="warning,mqtt_bridge::bridge=debug" - everything is set to warning, but code in mqtt_bridge::bridge module is set to debug.

I think this approach has the most flexibility for debugging, back-compatible and future proof.


dmolokanov (approved):
LGTM. should we add some tests to verify conversion works?

vadim-kovalyov:
> LGTM. should we add some tests to verify conversion works?

I'm open for suggestions on how to do that? :)

varunpuranik (approved):

dmolokanov:
> > LGTM. should we add some tests to verify conversion works?
> 
> I'm open for suggestions on how to do that? :)

Extract parsing to appropriate abstraction (method?) and write a unit test? 🤔

nyanzebra (approved):
Let's add a few small tests, but LGTM!

mqtt/mqttd/src/tracing/edgehub.rs
```diff
@@ -19,16 +19,12 @@ const EDGEHUB_2_RUST_LOG_LEVELS: [(&str, &str); 4] = [
 /// The following values are allowed: fatal, error, warning, info, debug, verbose.
 ///
 /// To make it work with rust log levels, we do a simple pre-processing.
+/// E.g: this string: `warning,mqtt_broker::broker=debug` becomes `warn,mqtt_broker::broker=debug`.
 pub fn init() {
     let mut log_level = env::var(EDGE_HUB_LOG_LEVEL_ENV)
         .map_or_else(|_| "info".into(), |level| level.to_lowercase());
```
dmolokanov, line 25:
```suggestion
        .map_or_else(|_| "info".into(), |level| sanitize_log_level(level.to_lowercase()));
```
nit

mqtt/mqttd/src/tracing/edgehub.rs
```diff
@@ -40,3 +36,48 @@ pub fn init() {
     let filter = log_level.parse().unwrap_or(LevelFilter::Info);
     let _ = LogTracer::builder().with_max_level(filter).init();
 }
+
+// make sure to replace all edgehub-specific log levels to rust-compatible
+fn sanitize_log_level(log_level: impl Into<String>) -> String {
+    let mut log_level: String = log_level.into();
```
dmolokanov, line 42: should we make lowercase inside this method?

dmolokanov (approved):
:shipit:

PR #4441, opened 2021-02-18
Switch to CRC over hash

nyanzebra:
Since we only actually care about data validation and not for malicious activity, we can use a checksum algorithm like crc32. This should give us some speedup too.

Show it works in CI: https://dev.azure.com/msazure/One/_build/results?buildId=39569477&view=results


mqtt/mqtt-bridge/Cargo.toml
```diff
@@ -35,6 +35,7 @@ mqtt3 = { path = "../mqtt3", features = ["serde1"] }
 mqtt-broker = { path = "../mqtt-broker" }
 mqtt-util = { path = "../mqtt-util" }
 rustc-hash = "1.1.0"
+crc32fast = "1.2.1"
```
vadim-kovalyov, line 38: are we settling on this less popular but faster (as author claims) crate? I don't see a problem per se, but we just need to make sure it works well on all platforms we support, cause it looks like it uses specific SIMD instructions.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/block.rs
```diff
@@ -16,15 +14,15 @@ lazy_static! {
         let v1 = BlockHeaderV1::new(0, 0, 0, 0, 0);
         let versioned_block = BlockVersion::Version1(v1);
 
-        bincode::serialized_size(&BlockHeaderWithHash::new(versioned_block)).unwrap()
+        bincode::serialized_size(&BlockHeaderWithCrc::new(versioned_block).unwrap()).unwrap()
```
vadim-kovalyov, line 17:
```suggestion
        bincode::serialized_size(&BlockHeaderWithCrc::new(versioned_block).expect("unable to create sample data block")).expect("unable to serialize sample data block")
```

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/block.rs
```diff
@@ -141,24 +139,29 @@ impl BlockHeaderWithHash {
...
+/// A utility fn to calculate any crc over serialized objects.
+pub(crate) fn calculate_crc<T>(t: &T) -> BincodeResult<u32>
 where
-    T: Hash + ?Sized,
+    T: Serialize + ?Sized,
 {
-    let mut hasher = rustc_hash::FxHasher::default();
-    t.hash(&mut hasher);
-    hasher.finish()
+    let data = bincode::serialize(t)?;
```
vadim-kovalyov, line 147: just an idea for future optimization: serialization allocates extra buffer. If we already have struct with stable and predictable layout (repr(C)), we could do some unsafe magic to transmute &T to &[u8]. But I believe we have to know the size of the `BlockVersion`. So, instead of calculating the size of `BlockHeaderWithCrc`, we can calculate the size of `BlockVersion` and add sizeof(u32) to get `BlockHeaderWithCrc` size.

I know @dmolokanov doesn't like unsafe, so this is just an option for the future, no need to fix it now. We have other low handing apples :)

vadim-kovalyov (dismissed):
The change LGTM! Just curious to see why you've chosen this particular crate.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/block.rs
dmolokanov: I think it is actually a good idea to avoid unnecessarily work

mqtt/mqtt-bridge/Cargo.toml
nyanzebra: ? it has 12 million downloads: https://crates.io/search?q=checksum&sort=downloads. I picked it because it was so popular :)

nyanzebra: Will order deps :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/block.rs
nyanzebra: Yeah, I have been trying to avoid unsafe magic like that if I can, but maybe in order to get good perf I will have to get my spellbooks out

mqtt/mqtt-bridge/Cargo.toml
nyanzebra: Also, I think this means it will not use simd for our use cases (we don't use nightly):

```rust
#![cfg_attr(not(feature = "std"), no_std)]
#![cfg_attr(
    all(feature = "nightly", target_arch = "aarch64"),
    feature(stdsimd, aarch64_target_feature)
)]
```

If everything works fine, we can just stick to this version as long as there are not major issues with it. Otherwise we can swap to something else :)

vadim-kovalyov (approved):
🚢

PR #4442, opened 2021-02-18
Speed up init for ring buffer

nyanzebra:
Removes need for metadata file.
Makes some fns generic over traits for read and write.
Uses BufReader (uses page_size crate to have capacity optimally configured) and checking for empty blocks to speed up init.
Makes 2 writes into 1 for insert.

Some #s for a 1GiB file and writing to fill it up.
```
init time 10.656031667s
time to insert 14510024 messages (size: 74): 31.852353167s
```
Here we can see it takes a little over 10 seconds now to scan the entire file at init time. It still takes a **long** time to write... Ideally both of these numbers get better. Maybe something moving towards async will help with these times. As for the times above, those are from my macbook, will put rpi times here too :).


mqtt/mqtt-bridge/Cargo.toml
```diff
@@ -17,6 +17,7 @@ humantime = "2.0"
 humantime-serde = "1.0"
 lazy_static = "1.4"
 mockall = "0.8"
+page_size = "0.4"
```
vadim-kovalyov, line 20: is this used?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -427,19 +389,15 @@ fn retrieve_ring_buffer_metadata(file: &mut File) -> BincodeResult<RingBufferMet
...
-) -> RingBufferMetadata {
-    let block_size = *SERIALIZED_BLOCK_SIZE;
-
-    let mut start = best_guess_metadata.file_pointers.write;
-    let mut end = start + block_size;
-    let mut read = best_guess_metadata.file_pointers.read_begin;
-    let mut write = best_guess_metadata.file_pointers.write;
-    let mut order = best_guess_metadata.order;
-    let can_read_from_wrap_around = best_guess_metadata.can_read_from_wrap_around;
+fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> RingBufferMetadata {
```
vadim-kovalyov, line 392: do we need to check the integrity of blocks on first load?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -516,152 +469,207 @@ fn find_pointers_and_order_post_crash(
...
-        start %= file_size;
+    max_size: u64,
+) -> BincodeResult<BlockHeaderWithCrc>
+where
+    T: Read + Seek,
+{
+    if start > max_size {
+        start %= max_size;
     }
     let end = start + size as u64;
```
vadim-kovalyov, line 531:
```suggestion
    let end = start + size;
```

vadim-kovalyov: Same in load_data

vadim-kovalyov (approved):
LGTM!

vadim-kovalyov (approved):

Issue #2747, opened 2020-03-27
iotedge check crashes on 1.0.9 under Yocto Linux and Rust 1.41.0


surlemur:
## Expected Behavior
`iotedge check` should output the result of some checks

## Current Behavior
`iotedge check` crashes (output with RUST_BACKTRACE=full):

```
root@edge:~# iotedge check
thread 'tokio-runtime-worker-1' panicked at 'byte index 6 is out of bounds of `temp1`', src/libcore/str/mod.rs:2052:9
stack backtrace:
   0:   0x91d76e - <unknown>
   1:   0x931b4e - <unknown>
   2:   0x91c2c8 - <unknown>
   3:   0x91f61a - <unknown>
   4:   0x91f420 - <unknown>
   5:   0x91fa9e - <unknown>
   6:   0x91f7fc - <unknown>
   7:   0x9316b0 - <unknown>
   8:   0x92e6c0 - <unknown>
   9:   0x765238 - <unknown>
  10:   0x7666f8 - <unknown>
  11:   0x75c098 - <unknown>
  12:   0x5c10b0 - <unknown>
  13:   0x5c8544 - <unknown>
  14:   0x5242b4 - <unknown>
  15:   0x5202e6 - <unknown>
  16:   0x5260b6 - <unknown>
  17:   0x8b14ce - <unknown>
  18:   0x8b2adc - <unknown>
  19:   0x92115a - <unknown>
  20:   0x8b1b0e - <unknown>
  21:   0x8ad602 - <unknown>
  22:   0x8acbae - <unknown>
  23:   0x89da0c - <unknown>
  24:   0x89dbd0 - <unknown>
  25:   0x89db04 - <unknown>
  26:   0x89ed56 - <unknown>
  27:   0x8b0eb4 - <unknown>
  28:   0x8b0ffc - <unknown>
  29:   0x8b2320 - <unknown>
  30:   0x8b2a86 - <unknown>
  31:   0x92115a - <unknown>
  32:   0x8b3a30 - <unknown>
  33:   0x90f596 - <unknown>
  34:   0x91cd44 - <unknown>
  35:   0x915966 - <unknown>
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Canceled', src/libcore/result.rs:1188:5
stack backtrace:
   0:   0x91d76e - <unknown>
   1:   0x931b4e - <unknown>
   2:   0x91c2c8 - <unknown>
   3:   0x91f61a - <unknown>
[...]

PR #4443, opened 2021-02-18
Update sysinfo package in iotedge CLI

damonbarry:
This change upgrades the sysinfo package in the iotedge CLI to match the version used by iotedged. It fixes a bug called out in #2747.

Note that master branch is on a newer version of sysinfo (0.14.10). For 1.1 I'm choosing to bring iotedge to the same version (0.12.0) as iotedged to minimize churn--this change aligns sysinfo across both components without changing iotedged.

### DIFF
--- a/edgelet/Cargo.lock
+++ b/edgelet/Cargo.lock
@@ -4,1002 +4,1076 @@
 name = "adler32"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d2e7343e7fc9de883d1b0341e0b13970f764c14101234857d2ddafa1cb1cac2"
 
 [[package]]
 name = "aho-corasick"
 version = "0.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d6531d44de723825aa81398a6415283229725a00fa30713812ab9323faa82fc4"
 dependencies = [
- "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr",
 ]
 
 [[package]]
 name = "ansi_term"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
 dependencies = [
- "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


arsing:
The 0.14 upgrade was done in https://github.com/Azure/iotedge/commit/e9dc6c1902b043bfaebf95d03c55e36a5c363413 . Is that fix not needed for 1.1 ? (cc @lfitchett )

damonbarry:
> The 0.14 upgrade was done in [e9dc6c1](https://github.com/Azure/iotedge/commit/e9dc6c1902b043bfaebf95d03c55e36a5c363413) . Is that fix not needed for 1.1 ? (cc @lfitchett )

Well, it looks like Lee's commit fixed two different problems: (1) the hwmon-related crash (fixed by GuillaumeGomez/sysinfo@13a224956ed57be800612f29379d2df1d37d4c67 and released in sysinfo 10.0, AFAICT), and (2) a problem with metrics that Lee mentioned in his commit description. Whether 1.1 needs the 2nd fix is not something we've talked about. But I believe upgrading to sysinfo 0.12.0 fixes the 1st problem. And it limits the scope of the change to iotedge CLI.

arsing:
>Well, it looks like Lee's commit fixed two different problems: (1) the hwmon-related crash (fixed by GuillaumeGomez/sysinfo@13a2249 and released in sysinfo 10.0, AFAICT), and (2) a problem with metrics that Lee mentioned in his commit description.

Lee's commit that I linked to updated from 0.12 to 0.14. It fixes only the second problem. The first one was fixed by a different commit by Lee which updated from 0.9 to 0.12. I'm already aware of that commit, and it is the change you currently have in your PR. I'm asking if the Docker CPU fix is also needed for 1.1 or not, because if it is you should bump to 0.14 directly instead of 0.12

damonbarry:
> Lee's commit that I linked to updated from 0.12 to 0.14. It fixes only the second problem.

That's true for iotedged, but not iotedge CLI. I've been looking at this from the iotedge CLI angle.

> I'm asking if the Docker CPU fix is also needed for 1.1

You're right, I think the Docker CPU fix is needed since metrics is a critical feature for 1.1. Stay tuned...

arsing:
Oh right. The hwmon fix was also needed for `support-bundle`, and that commit had the fix indeed. My bad.

damonbarry:
Closing this PR in favor or #4444.

PR #4445, opened 2021-02-19
Fixing iothub status error check

huguesBouvier:
Currently, iotedge check verify that upstream connection on 443 is valid by connection and checking the error status.
That additional error status check ensures that the connection is rejected by edgeHub and not stopping at the API proxy.

That additional check should have worked with edgeHub and IotHub, however, it happens some IotHub (in preview state in this case) don't return that error status and thus fails the check.

Since that specific check is designed to verify connection is not stopping at the API proxy, there is no need to verify for that header on lvl5 or when not nested. Hence that check is removed when there is no parent.

Tested:
When there is a parent
When there is no parent


edge-modules/iotedge-diagnostics-dotnet/src/Program.cs
```diff
@@ -91,10 +91,13 @@ static async Task Upstream(string hostname, string port, string proxy)
                 var httpRequest = new HttpRequestMessage(HttpMethod.Get, logsUrl);
                 HttpResponseMessage httpResponseMessage = await httpClient.SendAsync(httpRequest, HttpCompletionOption.ResponseHeadersRead);
 
-                var keys = httpResponseMessage.Headers.GetValues("iothub-errorcode");
-                if (!keys.Contains("InvalidProtocolVersion"))
+                if (isNested != null)
```
richma-ms, line 94: >if (isNested != null) [](start = 16, length = 21)

for the non-nested case we prob still need to at least check that the response isn't a 404 or a timeout.

huguesBouvier: That's the same logic as before. It trigger an exception for timeout I tested that. If it gets a 404, it still means it succeeded in connecting to iotHub since something replied. I am a bit wary of changing the logic so close it release since I can't test all the outliers.
In this PR I want to focus on fixing a specific bug.

richma-ms (approved):

PR #4448, opened 2021-02-19
Add mqtt generic client to nested long haul

dylanbronson:
Adding generic mqtt client to Nested Long Haul deployment
The Relayer is on the middle layer
and the Initiator is on the bottom layer. The initiator publishes messages upstream, and the relayer receives those messages and then re-publishes them so the initiator can receive them back. The initiator sends results to the TestResultCoordinator


dmolokanov (approved):
:shipit:

PR #4449, opened 2021-02-19
Adds storage settings

nyanzebra:
Adds basic storage settings that are specific for both Memory and RingBuffer, see specific config.jsons for examples.


mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -15,6 +22,8 @@ pub struct BridgeSettings {
     remotes: Vec<ConnectionSettings>,
 
     messages: MessagesSettings,
+
+    storage: Option<StorageSettings>,
```
dmolokanov, line 26: Isn't it should be for ring buffer by default?

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -251,13 +266,64 @@ struct UpstreamSettings {
...
+
+    pub fn flush_options(&self) -> &FlushOptions {
+        &self.flush_options
+    }
+}
+
+impl Default for RingBufferSettings {
+    fn default() -> Self {
+        Self {
+            max_file_size: unsafe { NonZeroU64::new_unchecked(4 * 1024 * 1024 * 1024) },
```
dmolokanov, line 313: lets move it to a const

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -251,13 +266,64 @@ struct UpstreamSettings {
...
+            directory: PathBuf::from("/tmp/mqttd/"),
+            flush_options: FlushOptions::AfterEachWrite,
```
dmolokanov, line 315: Well, I doubt it is a good default

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -251,13 +266,64 @@ struct UpstreamSettings {
...
+    pub fn directory(&self) -> &PathBuf {
+        &self.directory
+    }
+
+    pub fn flush_options(&self) -> &FlushOptions {
+        &self.flush_options
+    }
+}
+
+impl Default for RingBufferSettings {
```
dmolokanov, line 310: iirc all default settings we read from default.json . can we do that similar way?

dmolokanov (changes requested):
We used to have defaults.json file from mqtt-edgehub crate the source of default values for settings. Could you please make a corresponding change?

vadim-kovalyov (commented):
I agree with @dmolokanov that we need to use default.json for that.

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: I was planning on not having the default for RingBuffer here, rather in the bridge creation. So, if there are no settings for storage passed into the bridge, then it defaults to RingBuffer. That sound okay?

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: Shouldn't we allow for opting out of durability for speed?

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: Don't think we even need this, as if we go with the defaults.json option then these defaults for settings are useless right?

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: Just removed the Default here as storage settings are a part of bridge settings that can use defaults.json

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: Just removed the Default here as storage settings are a part of bridge settings that can use defaults.json

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: Just removed the Default here as storage settings are a part of bridge settings that can use defaults.json

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: Just added to defaults.json

vadim-kovalyov (approved):

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: Found that adding to defaults.json results in a lot of problems as the data being represented is an enum. So, during config merge it can end up trying to put memory and ring_buffer, which is obviously bad... so have opted for a default on StorageSettings and setting it in deserialize if no other version is found. 

If there is a better way to make this work, please let me know!

vadim-kovalyov (approved):

dmolokanov (approved):
Great!

PR #4451, opened 2021-02-20
fix blocked pipeline

ancaantochi:
Sending messages to cloud gets blocked when a leaf device is disabled, removed from scope or removed as a device because edgehub receives Unauthorized exceptions and will keep retrying.

To fix the issue:
- when AuthenticationMode is Scope and a new flag is true it will drop messages for devices which are not in scope or in scope and disabled. New flag that was added to drop messages is false by default
- CloudEndpoint gets a CloudProxy when trying to send a message. Before it was returning Option.None if connection failed, now it will return Try which will contain the exception when connection failed.  Existing code handles the case if exception is thrown during send message the cloud proxy is closed. Next time when it tries to get the cloud connection it should resync the cache.
- CloudConnectionProvider calls new method from device scope cache which returns DeviceInvalidStateException when device is disabled or not in scope (deleted device is treated as not is scope). device scope cache might get a response that the device is in scope, but when trying to connect it will receive Unauthorized exception, it that case it should force refresh the cache.
- device scope cache was changed to treat as device removed from scope when HTTP error codes 400, 401, 403, 404 (before was only 400)

Tests: 
CI passed: https://msazure.visualstudio.com/One/_build/results?buildId=39795314&view=results


davilu (approved):

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -88,8 +87,16 @@ public void Register(ContainerBuilder builder)
...
 
-            this.RegisterCommonModule(builder, optimizeForPerformance, storeAndForward, metricsConfig);
-            this.RegisterRoutingModule(builder, storeAndForward, experimentalFeatures);
+            if (!Enum.TryParse(this.configuration.GetValue("AuthenticationMode", string.Empty), true, out AuthenticationMode authenticationMode))
+            {
+                authenticationMode = AuthenticationMode.Scope;
+            }
+
+            bool trackDeviceState = authenticationMode == AuthenticationMode.Scope
+                && this.configuration.GetValue("TrackDeviceState", false);
```
ancaantochi, line 96: true by default

builds/e2e/connectivity.yaml
```diff
@@ -4,6 +4,7 @@ pr: none
 variables:
   images.artifact.name.linux: 'core-linux'
   vsts.project: $(System.TeamProjectId)
+  NugetSecurityAnalysisWarningLevel: warn
```
varunpuranik, line 7: ?

ancaantochi: I added to be able to run the connectivity tests, it was failing to run the pipeline otherwise.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -46,7 +49,9 @@ public class CloudConnectionProvider : ICloudConnectionProvider
             TimeSpan operationTimeout,
             bool useServerHeartbeat,
             Option<IWebProxy> proxy,
-            IMetadataStore metadataStore)
+            IMetadataStore metadataStore,
+            bool scopeAuthenticationOnly = true,
+            bool trackDeviceState = false)
```
varunpuranik, line 54: Do these have default values because of tests?

varunpuranik, line 54: Change trackDeviceState default to true

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -177,6 +196,102 @@ public async Task<Try<ICloudConnection>> Connect(IIdentity identity, Action<stri
...
+                    else
+                    {
+                        // old behavior
+                        Events.ErrorCreatingCloudConnection(identity, ex);
+                        return Try<ICloudConnection>.Failure(ex);
+                    }
+                }
+                else
+                {
+                    // try with cached device credentials if auth mode is not Scope or identity is for edgeHub
```
varunpuranik, line 280: When will this code get called? Walking back, it seems like this shoudl be called only if trackingDeviceState is true?

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -120,6 +123,21 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId, s
             return serviceIdentityResult;
         }
 
+        Exception MapException(DeviceScopeApiException ex)
+        {
+            switch (ex.StatusCode)
+            {
+                case HttpStatusCode.Unauthorized:
+                case HttpStatusCode.Forbidden:
+                    return new DeviceInvalidStateException($"Deviced not in scope: [{ex.StatusCode}: {ex.Message}].");
```
varunpuranik, line 132:
```suggestion
                    return new DeviceInvalidStateException($"Device not in scope: [{ex.StatusCode}: {ex.Message}].");
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -120,6 +123,21 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId, s
...
+                case HttpStatusCode.BadRequest:
+                case HttpStatusCode.NotFound:
+                    return new DeviceInvalidStateException($"Deviced not found: [{ex.StatusCode}: {ex.Message}].");
```
varunpuranik, line 135:
```suggestion
                    return new DeviceInvalidStateException($"Device not found: [{ex.StatusCode}: {ex.Message}].");
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -120,6 +123,21 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId, s
...
+                default:
+                    return new TimeoutException($"Request failed: [{ex.StatusCode}: {ex.Message}].");
```
varunpuranik, line 137: Why not include the original exception as inner exception? With just ex.Message, we will lose the callstack. Otherwise pls make the exception text more verbose, so that we know exactly where it was thrown.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceInvalidStateException.cs
```diff
@@ -0,0 +1,28 @@
...
+            : base(message)
+        {
+        }
+
+        public DeviceInvalidStateException(string message, Exception innerException)
+            : base(message, innerException)
+        {
+        }
+
+        protected DeviceInvalidStateException(SerializationInfo info, StreamingContext context)
```
varunpuranik, line 23: Do  you need this? If not, then pls remove.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -17,27 +17,40 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
 
     public sealed class DeviceScopeIdentitiesCache : IDeviceScopeIdentitiesCache
     {
+        static readonly TimeSpan DefaultRefreshDelay = TimeSpan.FromMinutes(5);
```
varunpuranik, line 20: 5 mins seems too much. What do we use in 1.2? I would suggest 1 min or 2 mins.

varunpuranik: For me, this is just to prevent repeated calls to the cloud (say in the span of a second)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -17,27 +17,40 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
...
         readonly IDictionary<string, StoredServiceIdentity> serviceIdentityCache;
         readonly Timer refreshCacheTimer;
         readonly TimeSpan refreshRate;
+        readonly TimeSpan refreshDelay;
         readonly AsyncAutoResetEvent refreshCacheSignal = new AsyncAutoResetEvent();
         readonly object refreshCacheLock = new object();
 
         Task refreshCacheTask;
 
+        DeviceScopeIdentitiesCache(
```
varunpuranik, line 33: Nit: we probably don't need 2 pvt ctors. We can always use the other one.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -74,12 +101,73 @@ await serviceIdentity
...
+            }
+            catch (Exception e)
+            {
+                Events.ErrorRefreshingCache(e, id);
+            }
+        }
+
+        bool ShouldNotRefresh(StoredServiceIdentity storedServiceIdentity, bool refreshIfOutOfDate) => !refreshIfOutOfDate || storedServiceIdentity.Timestamp + this.refreshDelay > DateTime.UtcNow;
+
+        void VerifyServiceIdentity(StoredServiceIdentity storedServiceIdentity) => this.VerifyServiceIdentity(storedServiceIdentity.ServiceIdentity);
```
varunpuranik, line 116: Nit: remove

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -74,12 +101,73 @@ await serviceIdentity
...
                 Events.ErrorRefreshingCache(e, id);
+                throw;
             }
         }
 
+        public async Task VerifyServiceIdentityState(string id, bool refreshIfOutOfDate = false)
+        {
+            Option<StoredServiceIdentity> storedServiceIdentity = await this.GetStoredServiceIdentity(id);
+            // if stored service identity is up to date, use it, otherwise refresh
+            await storedServiceIdentity.Filter(ssi => this.ShouldNotRefresh(ssi, refreshIfOutOfDate))
```
varunpuranik, line 161: Nit: the flow here is a bit confusing - suggest the following flow - 
- Get service identity
- Refresh if needed (does not validate)
- Validate it

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -74,12 +101,73 @@ await serviceIdentity
...
+            }
             catch (Exception e)
             {
+                // Refresh failed
                 Events.ErrorRefreshingCache(e, id);
+                throw;
             }
         }
 
+        public async Task VerifyServiceIdentityState(string id, bool refreshIfOutOfDate = false)
```
varunpuranik, line 157: Nit: rename refreshIfOutOfDate to refreshCacheEntry or refreshCachedIdentity

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
varunpuranik, line 161: We should log here if the cached identity is valid, or if we refreshed it, or if it was just refreshed.. debug logs

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -74,12 +101,73 @@ await serviceIdentity
...
+            {
+                Events.ErrorRefreshingCache(e, id);
+            }
+        }
+
+        bool ShouldNotRefresh(StoredServiceIdentity storedServiceIdentity, bool refreshIfOutOfDate) => !refreshIfOutOfDate || storedServiceIdentity.Timestamp + this.refreshDelay > DateTime.UtcNow;
+
+        void VerifyServiceIdentity(StoredServiceIdentity storedServiceIdentity) => this.VerifyServiceIdentity(storedServiceIdentity.ServiceIdentity);
+
+        void VerifyServiceIdentity(Option<ServiceIdentity> serviceIdentity) => serviceIdentity.ForEach(
```
varunpuranik, line 118: Should probably log the status of Verify, if it doesn't get logged anywhere else.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
```diff
@@ -88,8 +87,16 @@ public void Register(ContainerBuilder builder)
...
 
-            this.RegisterCommonModule(builder, optimizeForPerformance, storeAndForward, metricsConfig);
-            this.RegisterRoutingModule(builder, storeAndForward, experimentalFeatures);
+            if (!Enum.TryParse(this.configuration.GetValue("AuthenticationMode", string.Empty), true, out AuthenticationMode authenticationMode))
+            {
+                authenticationMode = AuthenticationMode.Scope;
+            }
+
+            bool trackDeviceState = authenticationMode == AuthenticationMode.Scope
+                && this.configuration.GetValue("TrackDeviceState", true);
```
varunpuranik, line 96: Trying to think what would be the most intuitive and future proof env var name here.. 
naming is hard :)

varunpuranik (approved):

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
ancaantochi: it is called if it is not scope only and scope auth failed or if it is edgeHub identity and scope auth failed. trackDeviceState is always false if auth mode is not scope

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
ancaantochi: I think TrackDeviceState might be too generic, maybe something like MessagesDropWhenDeviceInvalidState, MessagesDropWhenOutOfScope or MessagesHandlingWhenDeviceInvalidState and have an enum of Drop, keep. Also should we use State or Status? I think Status is more appropiate.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
ancaantochi: yes. I changed to use explicit values

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
ancaantochi: it's 2 min in 1.2,  updated here to 2 mins

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -163,11 +176,19 @@ public async Task<Try<ICloudConnection>> Connect(IIdentity identity, Action<stri
                     .GetOrElse(
                         async () =>
                         {
-                            Events.ServiceIdentityNotFound(identity);
-                            Option<IClientCredentials> clientCredentials = await this.credentialsCache.Get(identity);
-                            return await clientCredentials
-                                .Map(cc => this.Connect(cc, connectionStatusChangedHandler))
-                                .GetOrElse(() => throw new InvalidOperationException($"Unable to find identity {identity.Id} in device scopes cache or credentials cache"));
+                            // allow to use credential cache when auth mode was set to Scope and identity is for edgeHub
```
davilu, line 179:
```suggestion
                            // allow to use credential cache when auth mode was set to Scope or identity is for edgeHub
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -177,6 +198,102 @@ public async Task<Try<ICloudConnection>> Connect(IIdentity identity, Action<stri
...
+                        .Map(cc => this.Connect(cc, connectionStatusChangedHandler))
+                        .GetOrElse(() => throw new InvalidOperationException($"Unable to find identity {identity.Id} in device scopes cache or credentials cache"));
+                }
+            }
+            catch (Exception e)
+            {
+                Events.ErrorCreatingCloudConnection(identity, e);
+                return Try<ICloudConnection>.Failure(e);
+            }
+        }
```
davilu, line 295:
```suggestion
        async Task<Try<ICloudConnection>> TryRecoverCloudConnection(IIdentity identity, Action<string, CloudConnectionStatus> connectionStatusChangedHandler, bool wasRefreshed, Exception ex)
        {
            if (wasRefreshed)
            {
                Events.ErrorCreatingCloudConnection(identity, ex);
                return Try<ICloudConnection>.Failure(ex);
             }
             else
             {
                // recover: try to update out of date cache and try again
                return await this.ConnectInternalWithDeviceStateTracking(identity, connectionStatusChangedHandler, true);
             }
        }
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -120,6 +123,21 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId, s
             return serviceIdentityResult;
         }
 
+        Exception MapException(DeviceScopeApiException ex)
+        {
+            switch (ex.StatusCode)
+            {
+                case HttpStatusCode.Unauthorized:
+                case HttpStatusCode.Forbidden:
+                    return new DeviceInvalidStateException($"Device not in scope: [{ex.StatusCode}: {ex.Message}].");
```
davilu, line 132:
```suggestion
                    return new DeviceInvalidStateException($"Device not in scope: [{ex.StatusCode}: {ex.Message}].", ex);
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -120,6 +123,21 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId, s
...
+                case HttpStatusCode.BadRequest:
+                case HttpStatusCode.NotFound:
+                    return new DeviceInvalidStateException($"Device not found: [{ex.StatusCode}: {ex.Message}].");
```
davilu, line 135:
```suggestion
                    return new DeviceInvalidStateException($"Device not found: [{ex.StatusCode}: {ex.Message}].", ex);
```

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
```diff
@@ -97,6 +97,14 @@ public async Task<Option<ICloudProxy>> GetCloudConnection(string id)
                 .Map(c => (ICloudProxy)new RetryingCloudProxy(id, () => this.TryGetCloudConnection(id), c));
         }
 
+        public async Task<Try<ICloudProxy>> GetCloudConnectionTry(string id)
```
davilu, line 100:
```suggestion
        public async Task<Try<ICloudProxy>> TryGetCloudConnection(string id)
```

ancaantochi: renamed it to TryGetCloudConnection to be consistent with other GetCloudConnection.

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.Service/DependencyManager.cs
varunpuranik: Yes, I agree. But in a way TrackDeviceState/Status makes it more generic, in case we want to add more changes to how device deletions are handled (say remove twin/DMs/Subs, etc)

edge-hub/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/CloudConnectionProvider.cs
```diff
@@ -163,11 +176,19 @@ public async Task<Try<ICloudConnection>> Connect(IIdentity identity, Action<stri
...
-                            Option<IClientCredentials> clientCredentials = await this.credentialsCache.Get(identity);
-                            return await clientCredentials
-                                .Map(cc => this.Connect(cc, connectionStatusChangedHandler))
-                                .GetOrElse(() => throw new InvalidOperationException($"Unable to find identity {identity.Id} in device scopes cache or credentials cache"));
+                            // allow to use credential cache when auth mode is not Scope only (could be CloudAndScope or Cloud) or identity is for edgeHub
+                            if (!this.scopeAuthenticationOnly || this.edgeHubIdentity.Id.Equals(identity.Id))
+                            {
+                                Events.ServiceIdentityNotFound(identity);
+                                Option<IClientCredentials> clientCredentials = await this.credentialsCache.Get(identity);
+                                return await clientCredentials
```
varunpuranik, line 184: Nit: Expect is a better pattern here?

varunpuranik (dismissed):

davilu (dismissed):

davilu (approved):

PR #4452, opened 2021-02-20
support M2M subscription with both cases of /input/ included or not

vipeller:
M2M messages are being sent in the form of:

 "devices/device_1/modules/module_a/inputs/some_parameters"

because of that structure, a client can subscribe either for

 "devices/device_1/modules/module_a/#" or
 "devices/device_1/modules/module_a/inputs/#"

the C# SDK choses the first form, and because of that the current implementation handles that. However, it turned out that there are clients using the second form, and that is ignored.

this PR makes handle both versions

### DIFF
--- a/mqtt/mqtt-edgehub/src/topic/translation.rs
+++ b/mqtt/mqtt-edgehub/src/topic/translation.rs
@@ -519,7 +519,7 @@ mod tests {
         assert_eq!(
             c2d.to_internal("devices/device_1/modules/module_a/foo/#", &client_id),
             None
-        );        
+        );
 
         // M2M incoming
         assert_eq!(
--- a/mqtt/mqtt-edgehub/src/topic/translation.rs
+++ b/mqtt/mqtt-edgehub/src/topic/translation.rs
@@ -262,7 +262,7 @@ translate_c2d! {
     // Module-to-Module inputs
     module_to_module_inputs {
         to_internal {
-            format!("devices/{}/modules/{}/(inputs/)?#", DEVICE_ID, MODULE_ID),
+            format!("devices/{}/modules/{}/(#|inputs/.*)", DEVICE_ID, MODULE_ID),
             {|captures: regex::Captures<'_>, _| format!("$edgehub/{}/{}/+/inputs/#", &captures["device_id"], &captures["module_id"])}
         },
         to_external {
@@ -513,6 +513,14 @@ mod tests {
 
         assert_eq!(
             c2d.to_internal("devices/device_1/modules/module_a/inputs/#", &client_id),
+            Some("$edgehub/device_1/module_a/+/inputs/#".to_owned())
+        );
+
+        assert_eq!(
+            c2d.to_internal(
+                "devices/device_1/modules/module_a/inputs/route_1/#",
+                &client_id
+            ),
             Some("$edgehub/device_1/module_a/+/inputs/#".to_owned())
         );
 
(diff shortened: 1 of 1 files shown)


nyanzebra:
Do we need to add/update any integration tests later to make sure subscriptions are handled regardless of the C# sdk default? That way we have protection against any potential drift in code :)

vipeller:
@nyanzebra maybe it would be a good idea to make at least some basic operations using the c++ (or other) client libraries. I am not quite sure what the module uses which produced the problem. I am adding a ticket for this.

mqtt/mqtt-edgehub/src/topic/translation.rs
```diff
@@ -262,7 +262,7 @@ translate_c2d! {
     // Module-to-Module inputs
     module_to_module_inputs {
         to_internal {
-            format!("devices/{}/modules/{}/#", DEVICE_ID, MODULE_ID),
+            format!("devices/{}/modules/{}/(inputs/)?#", DEVICE_ID, MODULE_ID),
```
vadim-kovalyov, line 265: do we want to be on a safe side and do "starth_with" approach
```suggestion
            format!("devices/{}/modules/{}/(inputs/)?(?P<params>.*)", DEVICE_ID, MODULE_ID),
            &captures["device_id"], &captures["module_id"])}
            {|captures: regex::Captures<'_>, _| format!("$edgehub/{}/{}/+/inputs/{}", &captures["device_id"], &captures["module_id"], &captures["params"])}
```

varunpuranik:
@vipeller - which clients are using the 2nd form?

mqtt/mqtt-edgehub/src/topic/translation.rs
vipeller: The idea is good, but implementing it (just a bit) trickier:
The M2M message format is something like this:

devices/device_1/modules/module_a/inputs/route_1/%24.cdid=device_1&%24.cmid=module_a

Some valid subscriptions are:

devices/device_1/modules/module_a/#
devices/device_1/modules/module_a/inputs/#
devices/device_1/modules/module_a/inputs/route_1/#

What makes the problem a bit tricky is that EdgeHubCore is not an mqtt broker, but it needs to figure out the intention of the client, that it wants to receive M2M messages. So it does not do real topic filtering, still needs to know when the client is ready. The way it is implemented in EdgeHubCore using the old MQTT endpoint is that it uses a regular expression that is supposed to recognize when the clients want M2M:

"^devices/.+/modules/.+/#$"

Note, that it captures all of the above examples. However it also captures some subscription which would never lead to receiving M2M messages, e.g.

devices/device_1/modules/module_a/foo/#

the proposed change also captures this.

The other thing to consider that EdgeHubBridge does not need to know all the details of the subscription. EdgeHubBridge just does a simple string comparison that signals that the client wants M2M messages. This way it is not smart. The translated string could be "wants_m2m/device_1/module_a" even if the client gave a complex subscription which filters on route and other parameters. Because of that, it is not needed to capture the <params> and send it towards EdgeHubCore. Currently edgeHubCore fits this expression:

"^((?<dialect>(\$edgehub)|(\$iothub)))/(?<id1>[^/\+\#]+)/(?<id2>[^/\+\#]+)/\+/inputs/\#$"

or as an example:
$edgehub/device_1/modules/module_a/inputs/#

For this reason I propose this regex to capture the intention of M2M subscription of the client:

format!("devices/{}/modules/{}/(#|inputs/.*)", DEVICE_ID, MODULE_ID)

vadim-kovalyov (approved):

vipeller:
@varunpuranik vadim checked that it was a JS client

PR #4453, opened 2021-02-20
Re-enable E2E proxy test

massand:
- Re-enable E2E pipeline job to run tests in proxy environment
- Fixed some tests to run using Websockets transport (which is generally the case for E2E tests)
- Improve UploadSupportBundle test reliability in self-host environment by reducing requested log time range
- Fixed typo in traces within UploadSupportBundle product code
- Filtered out Nested Edge tests and temporarily filtered out tests that timeout in proxy environment

### DIFF
--- a/builds/e2e/templates/e2e-run.yaml
+++ b/builds/e2e/templates/e2e-run.yaml
@@ -41,7 +41,12 @@ steps:
     elseif ($test_type -eq 'http_proxy')
     {
       #Disable tests that don't work in proxy environment. Renable post-investigation.
-      $filter += '&FullyQualifiedName!~PriorityQueue&FullyQualifiedName!~PlugAndPlay'
+      $filter += '&FullyQualifiedName!~PriorityQueue&FullyQualifiedName!~PlugAndPlay&FullyQualifiedName!~ValidateMetrics'
+
+      #Disable tests that timeout in proxy self-host environment
+      $filter += '&FullyQualifiedName!~TestGetModuleLogs&FullyQualifiedName!~TestUploadSupportBundle'
+
+      #Disable nested edge tests
       $filter += '&Category!=NestedEdgeOnly'
     }
     else
--- a/builds/e2e/templates/e2e-run.yaml
+++ b/builds/e2e/templates/e2e-run.yaml
@@ -38,9 +38,15 @@ steps:
     {
       $filter = 'Category=nestededge_isa95'
     }
+    elseif ($test_type -eq 'http_proxy')
+    {
+      #Disable tests that don't work in proxy environment. Renable post-investigation.
+      $filter += '&FullyQualifiedName!~PriorityQueue&FullyQualifiedName!~PlugAndPlay'
+      $filter += '&Category!=NestedEdgeOnly'
+    }
     else
     {
-      $filter += '&Category!=NestedEdgeOnly' 
+      $filter += '&Category!=NestedEdgeOnly'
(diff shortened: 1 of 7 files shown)


gordonwang0 (dismissed):
Please run the pipeline on this PR to make sure it works before merging.

builds/e2e/templates/e2e-run.yaml
```diff
@@ -38,9 +38,20 @@ steps:
     {
       $filter = 'Category=nestededge_isa95'
     }
+    elseif ($test_type -eq 'http_proxy')
```
gordonwang0, line 41: I think it's better to create a new group for tests to disable behind the proxy instead of disabling tests by name individually.

test/Microsoft.Azure.Devices.Edge.Test/IoTEdgeCheck.cs
```diff
@@ -100,11 +100,15 @@ public async Task IoTEdge_check()
...
 
-            // TODO: iotedge check currently has 2 legitimate errors in it. When we update the aziot-submodule, the errors
-            // will go away, and we should change this assertion to empty string instead of passing on 2 errors.
-            // This is better than disabling the test, because it still tests iotedge check to some extent, and it will
-            // remind us with a failed run to update this test.
-            Assert.AreEqual(string.Empty, errors_number);
+            // When test runs behind proxy, MQTT and AMQP checks legitimately fail in three different connectivity checks
+            if (Context.Current.EdgeProxy.HasValue)
+            {
+                Assert.AreEqual("6 check(s) raised errors.", errors_number);
```
gordonwang0, line 106: This is kinda hacky. I think it's better to use the `--dont-run` option to `iotedge check` and disable the checks that shouldn't run behind a proxy there.

test/Microsoft.Azure.Devices.Edge.Test/PlugAndPlay.cs
```diff
@@ -29,10 +29,10 @@ public class PlugAndPlay : SasManualProvisioningFixture
         const string TestModelId = "dtmi:edgeE2ETest:TestCapabilityModel;1";
         const string LoadGenModuleName = "loadGenModule";
 
-        [TestCase(Protocol.Mqtt, false)]
-        [TestCase(Protocol.Amqp, false)]
-        [TestCase(Protocol.Mqtt, true)]
-        [TestCase(Protocol.Amqp, true)]
+        [TestCase(Protocol.MqttWs, false)]
```
gordonwang0, line 32: Why were these changed?

massand: Mentioned this in the description- Fixed some tests to run using Websockets transport (which is generally the case for E2E tests). If this assumption is incorrect, I can revert this. But, this should work for the normal (non-proxy) tests the same.

builds/e2e/templates/e2e-run.yaml
massand: Agreed on categorizing tests. I am still investigating some tests before adding test categories to filter them out.

test/Microsoft.Azure.Devices.Edge.Test/IoTEdgeCheck.cs
massand: Fixed

gordonwang0 (approved):

PR #4454, opened 2021-02-20
completing nested LH

huguesBouvier:
Adding image/edgelet/identity service selection ability + a bit of clean up.
https://dev.azure.com/msazure/One/_build/results?buildId=39675166&view=results
Confirmed all module deployed on lvl 3.

Fixing longHaul test duration.


dylanbronson (approved):
LGTM if we're going with the testDuration route

PR #4455, opened 2021-02-20
TestResultCoordinator: Ignore recent results if in longhaul mode

and-rewsmith:
For longhaul, we cannot take the same approach as connectivity tests where we stop the test, allow everything in-flight to settle, then generate the result. Instead we need to be able to handle things that are in-flight, meaning some results will be unmatched.

This change allows the TRC to ignore any recently-received, unmatched expected results from the report generation. It accomplishes this through a filtering abstraction which ignores any result that arrives in the last ignore-time, with the exception of actual results matching an expected result not in the ignore time. The default ignore-time is set to 1 minute.

There are really three conceptual changes in this PR:
1. The addition of the filtering abstraction, which includes changes in the settings, report generator factory, etc.
2. The conversion of ITestResultCollection to IAsyncEnumerable. There was a TODO in the ITestResultCollection class indicating we should convert this when we were on dotnet 3.1. But the main reason I did it was because of some convenience methods this provided for my filtering abstraction. If I didn't do it I think I'd be forced to do something hacky to extend the manual implementation of IAsyncEnumerable. This seemed messy so I decided to switch to IAsyncEnumerable, which bloated the diff size.
3. The conversion of the longhaul direct method reports to not allow for sender results without receiver results. This scenario won't exist in longhaul so we can eliminate some tests/logic.

The cleanest way to make the filtering abstraction was to inject some comparer, however there was not a comparer implemented for direct methods. The reason for this is because the report generator uses a heuristic about how the receiver will only report successes. I made a direct method comparer anyway that can be used in the filter abstraction which compares results more thoroughly (includes sequence number). We should at some point add this comparer to the direct method report generators.


test/modules/Modules.Test/TestResultCoordinator/TestReportUtilTest.cs
```diff
@@ -362,31 +361,6 @@ public void ParseReportMetadataList_ParseDirectMethodConnectivityTestReportMetad
...
-
-            Assert.Single(results);
-            var reportMetadata = results[0] as DirectMethodLongHaulReportMetadata;
-            Assert.NotNull(reportMetadata);
-            Assert.Equal("edge agent ping", reportMetadata.TestDescription);
-            Assert.Equal(TestOperationResultType.DirectMethod, reportMetadata.TestOperationResultType);
-            Assert.Equal(TestReportType.DirectMethodLongHaulReport, reportMetadata.TestReportType);
-            Assert.Equal("directMethodSender1.send", reportMetadata.SenderSource);
-            Assert.False(reportMetadata.ReceiverSource.HasValue);
-        }
```
and-rewsmith, line 388: This test no longer applies because longhaul has to have both sender/receiver.

test/modules/TestResultCoordinator/Reports/CloudTwinTestResultCollection.cs
```diff
@@ -2,82 +2,102 @@
...
+                    }.ToTestOperationResult();
+                }
+                catch (Exception e)
+                {
+                    Logger.LogError(e, $"Failed to get twin for {this.moduleId}");
+                    return null;
+                }
             }
         }
     }
```
and-rewsmith, line 103: No code changes, just convert to IAsyncEnumerable

test/modules/TestResultCoordinator/Reports/DirectMethod/DirectMethodTestResultComparer.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            DirectMethodTestResult dmResult1 = JsonConvert.DeserializeObject<DirectMethodTestResult>(value1.Result);
+            DirectMethodTestResult dmResult2 = JsonConvert.DeserializeObject<DirectMethodTestResult>(value2.Result);
+
+            return dmResult1.TrackingId == dmResult2.TrackingId
+                && dmResult1.SequenceNumber == dmResult2.SequenceNumber
+                && dmResult1.HttpStatusCode == dmResult2.HttpStatusCode;
+        }
+    }
+}
```
and-rewsmith, line 35: Made this comparer so that we can use in the filter abstracton. Before this PR, it wasn't needed because we make assumptions about how dm tests operate (i.e. receiver never sends any failure result).

I made this comparer and only integrated into the filter, not the report generator which works as before.

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReportGenerator.cs
```diff
@@ -77,20 +73,16 @@ public async Task<ITestResultReport> CreateReportAsync()
...
-                    receiverResults++;
-                }
+                // ReceiverSource will always be there if ReceiverTestResults is so it's safe to put OrDefault
+                this.ValidateDataSource(this.ReceiverTestResults.Current, this.ReceiverSource);
+                DirectMethodTestResult dmReceiverTestResult = JsonConvert.DeserializeObject<DirectMethodTestResult>(this.ReceiverTestResults.Current.Result);
+                receiverResults++;
+            }
 
-                receiverSuccesses = Option.Some(receiverResults);
-            });
```
and-rewsmith, line 93: Longhaul direct methods will always have receiver results

test/modules/TestResultCoordinator/Reports/ITestResultCollection.cs
```diff
@@ -1,16 +0,0 @@
...
-    // TODO: Replace ITestResultCollection<T> with IAsyncEnumerable<T> when project is moved to .Net Core 3.1
-    public interface ITestResultCollection<T> : IDisposable
-    {
-        T Current { get; }
-
-        Task<bool> MoveNextAsync();
-
-        void Reset();
-    }
-}
```
and-rewsmith, line 16: Deleted because we converted to IAsyncEnumerable

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
```diff
@@ -10,19 +10,14 @@ class DirectMethodLongHaulReport : TestResultReportBase
             string testDescription,
             string trackingId,
             string senderSource,
-            Option<string> receiverSource,
```
and-rewsmith, line 13: Longhaul has to have receiver source.

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReportGenerator.cs
```diff
@@ -20,16 +21,11 @@ class DirectMethodLongHaulReportGenerator : ITestResultReportGenerator
             string testDescription,
             string trackingId,
             string senderSource,
-            ITestResultCollection<TestOperationResult> senderTestResults,
-            Option<string> receiverSource,
-            Option<ITestResultCollection<TestOperationResult>> receiverTestResults,
```
and-rewsmith, line 25: longhaul has to have receiver results

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReportMetadata.cs
```diff
@@ -9,19 +9,18 @@ class DirectMethodLongHaulReportMetadata : TestReportMetadataBase, ITestReportMe
         public DirectMethodLongHaulReportMetadata(
             string testDescription,
             string senderSource,
-            string receiverSource = "")
+            string receiverSource)
             : base(testDescription)
         {
             this.SenderSource = senderSource;
-            this.ReceiverSource = string.IsNullOrEmpty(receiverSource) ? Option.None<string>() : Option.Some(receiverSource);
```
and-rewsmith, line 16: receiver source must exist for longhaul

test/modules/TestResultCoordinator/Reports/StoreTestResultCollection.cs
```diff
@@ -2,75 +2,88 @@
...
+                {
+                    this.resultQueue.Enqueue(values.Item2);
+                    lastLoadedKey = values.Item1;
+                }
+
+                return lastLoadedKey;
+            }
         }
     }
 }
```
and-rewsmith, line 89: no code changes, just convert to IAsyncEnumerable

test/modules/TestResultCoordinator/Reports/TestReportGeneratorFactory.cs
```diff
@@ -42,13 +47,20 @@ internal TestReportGeneratorFactory(ITestOperationResultStorage storage, Network
...
+
                         return new CountingReportGenerator(
                             metadata.TestDescription,
                             trackingId,
                             metadata.ExpectedSource,
-                            expectedTestResults,
+                            expectedTestResults.GetAsyncEnumerator(),
                             metadata.ActualSource,
-                            actualTestResults,
+                            actualTestResults.GetAsyncEnumerator(),
```
and-rewsmith, line 63: These generators took our custom enumerator type before, so we pass in the IAsyncEnumerator.

and-rewsmith:
I tested it manually and it works. Image tag:
edgebuilds.azurecr.io/andsmi/azureiotedge-test-result-coordinator:0

@dylanbronson We could add tests for this, but I didn't have time. Your call as I know we are on a tight schedule.

test/modules/TestResultCoordinator/Reports/TestResultFilter.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// <summary>
+    /// Longhaul tests need to filter out recent results, so recent unmatched results
+    /// that will soon be matched don't result in false failures.
+    ///
+    /// This abstraction filters out any result with a CreatedAt
+    /// within the configurable ignore threshold, whith
+    /// the exception of actual results matching a
+    /// non-filtered expected result.
+    /// </summary>
```
and-rewsmith, line 20: If there is an unexpected actual result at the very beginning, followed by expected/actual pairs, this will lead to bad filtering. But this implementation is consistent with our report generators which suffer from the same issue. I have never seen this happen, and don't expect it to be possible.

test/modules/TestResultCoordinator/Reports/TestResultFilter.cs
```diff
@@ -0,0 +1,84 @@
...
+            List<TestOperationResult> expectedResults = await expectedTestResultsEnumerable.ToListAsync();
+            List<TestOperationResult> actualResults = await actualTestResultsEnumerable.ToListAsync();
+            List<TestOperationResult> expectedResultsOutput = new List<TestOperationResult>();
+            List<TestOperationResult> actualResultsOutput = new List<TestOperationResult>();
+
+            DateTime startIgnoringAt = DateTime.UtcNow.Subtract(unmatchedResultTolerance);
+
+            int expectedResultCounter = 0;
+            int actualResultCounter = 0;
+            while (expectedResultCounter < expectedResults.Count && expectedResults[expectedResultCounter].CreatedAt < startIgnoringAt)
```
and-rewsmith, line 45: There are a couple other approaches we could have taken for the filter.

*Split the expected results list based on ignore threshold, then for every expected result find its corresponding actual result if exists*
I didn't do this because we would need to have hashing or some other optimization to avoid O(n^2). Perf for the TRC is not a priority, but we potentially will have tons of results so we should avoid this if we can.

*Iterate backwards and pop elements off expected/actual lists*
The current implementation mimics the way the report generators iterate over results. These report generators assume:
- having an unexpected actual result at the very beginning is impossible
- having duplicate actual results is possible
- having trailing actual results with no matching expected results is possible

I think there are some logic issues with these assumptions (i.e. inconsistent views on when unexpected actual results are possible). But making these assumptions allows the iteration logic to be a bit simpler because we can assume that the first actual result will have an expected result match. So if the first expected and actual don't match, then we can keep over expected and the match will eventually be found.

If we iterate backwards, we can't leverage these assumptions. There is ambiguity as to whether the last actual result can be matched by iterating backwards over the expected results.

dylanbronson (dismissed):
Nice job. This was not an easy fix, but this seems like the best way to do it

test/modules/Modules.Test/TestResultCoordinator/LongHaulFilteringTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            IAsyncEnumerable<TestOperationResult> actualResults = new StoreTestResultCollection<TestOperationResult>(mockActualStore.Object, 10);
+            List<TestOperationResult> list = new List<TestOperationResult>();
+            (var filteredExpectedTestResults, var filteredActualTestResults) =
+                await filter.FilterResults(unmatchedResultTolerance, expectedResults, actualResults);
+            Assert.Equal(2, await filteredExpectedTestResults.CountAsync());
+            Assert.Equal(2, await filteredActualTestResults.CountAsync());
+        }
+
+        [Fact]
+        public async Task FilterTestIncludeDuplicatesAndIgnoreResultsAfterTolerance()
```
and-rewsmith, line 36: We might want to specify the duplicates are actual results.

test/modules/Modules.Test/TestResultCoordinator/LongHaulFilteringTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            IAsyncEnumerable<TestOperationResult> actualResults = new StoreTestResultCollection<TestOperationResult>(mockActualStore.Object, 10);
+            List<TestOperationResult> list = new List<TestOperationResult>();
+            (var filteredExpectedTestResults, var filteredActualTestResults) =
+                await filter.FilterResults(unmatchedResultTolerance, expectedResults, actualResults);
+            Assert.Equal(3, await filteredExpectedTestResults.CountAsync());
+            Assert.Equal(5, await filteredActualTestResults.CountAsync());
+        }
+
+        [Fact]
+        public async Task FilterTestActualResultWithNoMatch()
```
and-rewsmith, line 52: We should specify that the no match is only at the end

test/modules/Modules.Test/TestResultCoordinator/LongHaulFilteringTest.cs
```diff
@@ -0,0 +1,152 @@
...
+    using global::TestResultCoordinator;
+    using global::TestResultCoordinator.Reports;
+    using Microsoft.Azure.Devices.Edge.ModuleUtil;
+    using Microsoft.Azure.Devices.Edge.Storage;
+    using Microsoft.Azure.Devices.Edge.Util.Test.Common;
+    using Moq;
+    using Xunit;
+
+    [Unit]
+    public class LongHaulFilteringTest
```
and-rewsmith, line 17: nit: We could also add tests for:
- no actual results
- no expected results

and-rewsmith (commented):
Tests look good. I think we could add some more tests, but in the interest of time I think we should make some minor naming edits for clarity then merge. I can do the naming edits and also add some more context to the docstring for the filter.

dylanbronson (dismissed):
lgtm nice job

dylanbronson (approved):

PR #4456, opened 2021-02-22
Test Dashboard: Update dev ops sync job and restore existing tests

and-rewsmith:
We need to add the new pipelines and release branches to the dev ops sync job we have running in a container instance. I have already built the image and deployed the new container.


tools/IoTEdgeDevOps/DevOpsLibTest/BuildExtensionTest.cs
```diff
@@ -39,9 +43,9 @@ public void TestDisplayName()
             Assert.AreEqual("Edgelet CI", BuildDefinitionId.EdgeletCI.DisplayName());
             Assert.AreEqual("Edgelet Packages", BuildDefinitionId.EdgeletPackages.DisplayName());
             Assert.AreEqual("Edgelet Release", BuildDefinitionId.EdgeletRelease.DisplayName());
-            Assert.AreEqual("End-to-End Test", BuildDefinitionId.EndToEndTest.DisplayName());
+            Assert.AreEqual("New E2E Test", BuildDefinitionId.EndToEndTest.DisplayName());
```
nyanzebra, line 46: nit: Do we need to have the 'New' in the name?

nyanzebra (approved):

tools/IoTEdgeDevOps/DevOpsLibTest/BuildExtensionTest.cs
and-rewsmith: We used to differentiate between multiple E2E pipelines with this tag, but if we remove it, we would need to change many things.

PR #4457, opened 2021-02-22
Test Dashboard: Add test name to TestInfo

and-rewsmith:
Now that we are adding various nested pipelines, we need to be able to add more fields to TestInfo so that the test dashboard can parse what kind of environment the test is running in.


builds/e2e/templates/longhaul-deploy.yaml
```diff
@@ -89,7 +89,7 @@ steps:
         testInfo="$testInfo,EdgeletSourceBranch=${{ parameters['edgelet.source.branch'] }}"
         testInfo="$testInfo,ImagesSourceBranch=${{ parameters['images.source.branch'] }}"
         testInfo="$testInfo,HostPlatform=${{ parameters['metricsCollector.hostPlatform'] }}"
-        testInfo="$testInfo,TestName=$testName"
+        testInfo="$testInfo,TestName=longhaul (single-node)"
```
and-rewsmith, line 92: I already added testname param to all other pipelines than single-node longhaul. This is hardcoded because single-node longhaul is the only one that uses this deploy yaml.

I am going to make a change to hardcode it elsewhere too,

and-rewsmith: Unfortunately, I cannot hardcode it elsewhere as nested-longhaul-deploy is used for the following TestNames:
- longhaul (nested-broker)
- longhaul (nested-non-broker)

Nested connectivity hasn't landed yet, but when it does, I think they are using the same connectivity deploy yaml. So the desired implementation is that all pipelines except single-node longhaul will have a pipeline variable indicating TestName.

yophilav: Is the white space treated nicely with the query & the backend? 
Also would `testInfo="$testInfo,TestName='longhaul (single-node)'"`  Note the `'` around the quote be a safer practice?

and-rewsmith: Yeah whitespace is fine. Philip and I wanted a super simple custom parse format so decided not to do quotes

yophilav (approved):

PR #4461, opened 2021-02-22
[iotedge check] Skip latest version check in nested scenarios

daprilik:
Nested deployments typically do not have access to the public internet, and therefore cannot perform the HTTP request to fetch the latest version information from GitHub. This PR addresses this issue by checking if `iotedge check` is running in a nested deployment, and tweaking the behavior of the `aziot-edged-version` check accordingly.

* * *

This PR ended up being a bit gnarlier than expected, as the version-check HTTP request was being made prior to loading the config.yaml, which must be read from to determine whether or not there is a nested deployment. This seems to have been done due to historical reasons - i.e: the asynchronous HTTP request logic was added prior to the refactor that enabled individual checks to spawn async tasks on the tokio runtime. The code has now been moved directly into the relevant check, which makes much more sense from a code-organization standpoint.


edgelet/iotedge/src/check/checks/aziot_edged_version.rs
```diff
@@ -18,8 +24,113 @@ impl Checker for AziotEdgedVersion {
...
+        check: &mut Check,
+        tokio_runtime: &mut tokio::runtime::Runtime,
+    ) -> CheckResult {
+        let settings = if let Some(settings) = &check.settings {
+            settings
+        } else {
+            return CheckResult::Skipped;
+        };
+
+        if settings.parent_hostname().is_some() {
```
arsing, line 38: No reason to skip it if `check.expected_aziot_edged_version` is `Some()`

edgelet/iotedge/src/check/checks/aziot_edged_version.rs
```diff
@@ -18,8 +24,113 @@ impl Checker for AziotEdgedVersion {
...
+            return CheckResult::Ignored;
```
arsing, line 39: Add a comment.

```rust
// This is a nested Edge device so it may not be able to access aka.ms or github.com.
// In the best case the request would be blocked immediately, but in the worst case it may take a long time to time out.
// The user didn't provide the expected version on the command line either. So just ignore this check.
```

edgelet/iotedge/src/check/checks/aziot_edged_version.rs
```diff
@@ -18,8 +24,113 @@ impl Checker for AziotEdgedVersion {
...
+                hyper_client
+                    .call(request)
+                    .then(|response| -> Result<_, Error> {
+                        let response = response.context(ErrorKind::FetchLatestVersions(
+                            FetchLatestVersionsReason::GetResponse,
+                        ))?;
+                        Ok(response)
+                    })
+                    .and_then(move |response| match response.status() {
+                        hyper::StatusCode::MOVED_PERMANENTLY => {
```
arsing, line 81: I know this is the original code, but this will fail if aka.ms starts returning 3xx instead of 301 specifically. (I'm surprised it returns 301 instead of 302, actually, given that aka.ms redirects are mutable.)

Change it to:

```rust
if response.status().is_redirection() {
    ...
}
else {
    ...
}
```

edgelet/iotedge/src/lib.rs
```diff
@@ -48,5 +48,5 @@ pub trait Command {
 
 #[derive(Debug, Deserialize)]
 pub struct LatestVersions {
-    pub aziot_edged: String,
+    pub iotedged: String,
```
arsing, line 51: Why?

edgelet/iotedge/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-#![deny(rust_2018_idioms, warnings)]
-#![deny(clippy::all, clippy::pedantic)]
+// #![deny(rust_2018_idioms, warnings)]
```
arsing, line 3: Revert this.

edgelet/iotedge/src/check/mod.rs
```diff
@@ -40,7 +36,7 @@ pub struct Check {
     diagnostics_image_name: String,
     dont_run: BTreeSet<String>,
     aziot_edged: PathBuf,
-    latest_versions: Result<super::LatestVersions, Option<Error>>,
+    expected_aziot_edged_version: Option<String>,
```
arsing, line 39: Why?

arsing (changes requested):

edgelet/iotedge/src/lib.rs
daprilik: @massand seems to have changed it as part of his PR (probably as part of a larger search + replace).

The actual JSON blob still uses `iotedge` as the key, so using `aziot_edged` isn't valid.

edgelet/iotedge/src/lib.rs
daprilik: oops, thought I caught that earlier, will do :sweat_smile:

edgelet/iotedge/src/lib.rs
arsing: You're making the commit to master, not to 1.1. The "actual JSON blob" will use `"aziot-edged"` when it's released.

arsing: Actually, on that note, we should change the URL too, to `https://aka.ms/latest-aziot-edge-stable`

@veyalla, @micahl: We don't want `iotedge check` on 1.1 to think 1.2 is a newer version, right? So `latest-iotedge-stable` will say `"1.1"` even after 1.2 is released?

daprilik: Ah, fair enough. I guess I'd just seen that the check was failing, and thought to quickly "fix" it alongside these other changes.

But yes, this does bring up a good point that we should have some way of ensuring that we don't break existing code by renaming the JSON field.

edgelet/iotedge/src/check/checks/aziot_edged_version.rs
```diff
@@ -18,8 +24,116 @@ impl Checker for AziotEdgedVersion {
...
+                settings
+            } else {
+                return CheckResult::Skipped;
+            };
+
+            if settings.parent_hostname().is_some() {
+                // This is a nested Edge device so it may not be able to access aka.ms or github.com.
+                // In the best case the request would be blocked immediately, but in the worst case it may take a long time to time out.
+                // The user didn't provide the `expected_aziot_edged_version` param on the CLI, so we just ignore this check.
+                return CheckResult::Ignored;
```
huguesBouvier, line 49: Thanks! That's correct. In ISA95 environment, bottom layer is locked with not internet access.

huguesBouvier (approved):
Thank you!

edgelet/iotedge/src/lib.rs
daprilik: @arsing I think we should leave this change in for now, since updating this field name will have to be coordinated with any changes to the aka.ms URL + JSON structure in the lead-up to 1.2. 

Moreover, at the moment, leaving the field as "aziot_edged" is plain wrong on two accounts:
a) the current aka.ms URL doesn't match the JSON, and 
b) the JSON field will eventually be `"aziot-edged"`, not `"aziot_edged"` as it is currently.

I'd like to merge this PR to fix the particular issue at hand, and open another one in the future after the PMs have decided how we want to tackle the latest-version JSON question.

arsing (approved):
Please fix your PR description before merging. (It becomes the commit message.)

PR #4470, opened 2021-02-24
fix download aziotis

ancaantochi:
connectivity pipeline fix


gordonwang0 (approved):

and-rewsmith (changes requested):
We aren't taking this right? Requesting changes to make sure not accidentally merged.

huguesBouvier:
> We aren't taking this right? Requesting changes to make sure not accidentally merged.

Why is it wrong? It looks like a legit change.

PR #4474, opened 2021-02-24
Re-introduce the test.deploymentFileName

yophilav:
- Re-introduce the test.deploymentFileName


and-rewsmith (approved):
We should make sure to update all longhaul pipelines with this pipeline var.

yophilav:
> We should make sure to update all longhaul pipelines with this pipeline var.

It was done before this PR was even out! :D

dylanbronson (approved):

PR #4478, opened 2021-02-24
Self checkout for Connectivity test pipeline

yophilav:
- Self checkout for Connectivity test pipeline

### DIFF
--- a/builds/e2e/connectivity.yaml
+++ b/builds/e2e/connectivity.yaml
@@ -98,7 +98,10 @@ jobs:
             echo "deploymentFileName=$(deploymentFileName)"
             echo "test.runtimeLogLevel=$(test.runtimeLogLevel)"
       - template: templates/set-run-flag.yaml
-      - checkout: none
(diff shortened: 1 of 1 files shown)


and-rewsmith (approved):

builds/e2e/connectivity.yaml
```diff
@@ -98,7 +98,10 @@ jobs:
             echo "deploymentFileName=$(deploymentFileName)"
             echo "test.runtimeLogLevel=$(test.runtimeLogLevel)"
       - template: templates/set-run-flag.yaml
-      - checkout: none
+      - checkout: self
+        clean: true
+        fetchDepth: 100
+        submodules: recursive
```
and-rewsmith, line 104: We should make sure to match the pattern in other pipelines. Might be a good idea to create a template in the future.

yophilav: I talked with Damon about this before. It's matched with SN LH pipeline right now.

PR #4481, opened 2021-02-25
Added a separate message pump for messages from upstream

vipeller:
The reason of the change is that when there is a big load of messages (from devices/modules), the processing queue can grow long. In this case, if there are messages from upstream (e.g twin result, direct method call), those messages get at the end of the queue and takes some time to process them. This change creates a separate queue for messages from upstream, increasing the reaction time of the system in certain scenarios.

The root of the change is in the method ForwardPublish(). This method gets calld by the mqtt library when a message arrives, and this puts the messages into a processing queue and returns immediately (giving back the control to the mqtt library as soon as possible). Now the change checks for messages from upstream (those all start with string "$downstream", and puts them into a separate queue.

This separate queue gets processed by a specific processing loop, implemented by UpstreamLoop(). Before this change, there was only a single loop, now there is an UpstreamLoop() and DownstreamLoop(), where UpstreamLoop() processes message coming from upstream. This loop uses a simple message handler class called "BrokeredCloudProxyDispatcher" as all messages from upstream goes through that class. DownstreamLoop() check all available message handlers (including BrokeredCloudProxyDispatcher), because it is not known which handler would process the messages.

Note, that it is intentional leaving the BrokeredCloudProxyDispatcher class for DownstreamLoop(), so if some message breaks the convention (not starting with $downstream), that still will be handled.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -305,71 +310,137 @@ void TriggerReconnect(object sender, EventArgs e)
...
+                                                SingleWriter = true
+                                            }));
+        }
 
-                                        if (!accepted)
-                                        {
-                                            Events.MessageNotForwarded(publishInfo.Topic, publishInfo.Payload.Length);
-                                        }
-                                    }
+        async Task DownstreamLoop()
```
varunpuranik, line 352: Can DownstreamLoop and UpstreamLoop be refactored.. much of the code is identical..

varunpuranik:
Can we add tests for this?

varunpuranik (approved):
Signing off.. but also get ack from Anca.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
```diff
@@ -305,71 +310,137 @@ void TriggerReconnect(object sender, EventArgs e)
...
+        {
+            Events.DownstreamForwardingLoopStarted();
+            while (await this.downstreamPublications.Expect(ChannelIsBroken).Reader.WaitToReadAsync())
```
vadim-kovalyov, line 355: nit: extract var `Channel<MqttPublishInfo> channel = this.downstreamPublications.Expect(ChannelIsBroken)`

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: made UpstreamLoop shorter, now the difference is more visible

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/brokerConnection/MqttBrokerConnector.cs
vipeller: extracted

vadim-kovalyov (approved):
LGTM! but I'd let @ancaantochi take a look

PR #4482, opened 2021-02-25
Improve DirectMethod Reporting for long haul

dylanbronson:
Previously, the DirectMethod report only reports 200's, 0's, and unknowns.
The 0's are a known problem having to do with edgeHub restarts.
Unknowns were its own bucket.
This change buckets status codes together so we know exactly which (and how many) codes we got


test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReportGenerator.cs
```diff
@@ -72,14 +74,22 @@ public async Task<ITestResultReport> CreateReportAsync()
                         senderSuccesses++;
                         break;
                     default:
```
yophilav, line 76: From `test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs ` comment: ` // fail if we find anything that is not a 200 or 0 (most notably, 500's)`. it shouldn't be possible to have `200` in the `other` right? This switch case still think `200` should be part of the dict.

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
```diff
@@ -2,7 +2,10 @@
 namespace TestResultCoordinator.Reports.DirectMethod.LongHaul
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
     using Microsoft.Azure.Devices.Edge.Util;
+    using Org.BouncyCastle.Math.EC.Rfc7748;
```
yophilav, line 8: What's this for?

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReportGenerator.cs
```diff
@@ -55,15 +56,16 @@ class DirectMethodLongHaulReportGenerator : ITestResultReportGenerator
...
+            long senderSuccesses = 0;
+            Option<long> receiverSuccesses = Option.None<long>();
+            long statusCodeZero = 0;
+            Dictionary<int, long> other = new Dictionary<int, long>();
             while (await this.SenderTestResults.MoveNextAsync())
             {
                 this.ValidateDataSource(this.SenderTestResults.Current, this.SenderSource);
                 DirectMethodTestResult dmSenderTestResult = JsonConvert.DeserializeObject<DirectMethodTestResult>(this.SenderTestResults.Current.Result);
-                switch ((int)dmSenderTestResult.HttpStatusCode)
+                int statusCode = (int)dmSenderTestResult.HttpStatusCode;
```
yophilav, line 67: Hmm.... I have a mixed feeling about this. One hand, you can just have the dict with key type `<System.Net.HttpStatusCode ,long>` or you could have `<int, long>`. Personally, I'd prefer to have `<System.Net.HttpStatusCode ,long>` as it carries out how to interpret the key through the code. I'll leave it up to you to decide.

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportGeneratorTest.cs
```diff
@@ -330,10 +330,10 @@ public async Task TestCreateReportAsyncWithEmptyResults()
...
 
-            Assert.Equal(0UL, report.ReceiverSuccesses.Expect<ArgumentException>(() => throw new ArgumentException("impossible")));
-            Assert.Equal(0UL, report.SenderSuccesses);
-            Assert.Equal(0UL, report.StatusCodeZero);
-            Assert.Equal(0UL, report.Unknown);
+            Assert.Equal(0L, report.ReceiverSuccesses.Expect<ArgumentException>(() => throw new ArgumentException("impossible")));
+            Assert.Equal(0L, report.SenderSuccesses);
+            Assert.Equal(0L, report.StatusCodeZero);
+            Assert.Equal(0L, report.Other.Sum(x => x.Value));
             Assert.True(report.IsPassed);
```
yophilav, line 337: From the look there is no test case that verify the sorting of `statusCode` key here? From the PR, it looks like it should work. You can also add a new test case here too. I'll leave it up to you to decide.

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReportGenerator.cs
dylanbronson: I don't think 200 will ever show up in the dict, since it will go to the case 200: senderSuccesses++ lines

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
dylanbronson: removed

test/modules/Modules.Test/TestResultCoordinator/Reports/DirectMethod/DirectMethodLongHaulReportData.cs
```diff
@@ -69,5 +69,27 @@ class DirectMethodLongHaulReportData
...
+                        new DateTime(2020, 1, 1, 9, 10, 22, 10),
+                        new DateTime(2020, 1, 1, 9, 10, 23, 10),
+                        new DateTime(2020, 1, 1, 9, 10, 24, 10),
+                        new DateTime(2020, 1, 1, 9, 10, 24, 15)
+                    },
+                    7,
+                    false,
+                    3L, 1L,
+                    new Dictionary<HttpStatusCode, long> { { HttpStatusCode.InternalServerError, 2 }, { HttpStatusCode.ServiceUnavailable, 1 } }
+                };
```
yophilav, line 93: Nice work! Thank you!

yophilav (approved):

PR #4483, opened 2021-02-25
Remove unused tasks from images release pipeline

damonbarry:
The images-release pipeline installs Rust and cross on the agent, but they're not currently being used. So this change trims them from the pipeline. It also replaces the nuget feed error with a warning, as we've done in the other pipelines.

I verified that, without at least the nuget change, the current pipeline fails. With these changes, the pipeline passes.


builds/misc/images-release.yaml
```diff
@@ -17,18 +19,9 @@ jobs:
...
-        inputs:
-          filePath: scripts/linux/generic-rust/install.sh
-          arguments: --project-root "edgelet" --package-arm
-      - bash: "echo \"##vso[task.setvariable variable=PATH;]$HOME/.cargo/bin:$PATH\""
-        displayName: "Modify path"
       - template: ../templates/install-dotnet2.yaml
       - template: ../templates/install-dotnet3.yaml
       - template: ../templates/dotnet3-globaljson.yaml # use dotnet 3 as primary install for build
-      - bash: "cargo install --git https://github.com/arsing/cross.git --branch set-path"
-        displayName: "Install cross (fork with docker fix)"
```
and-rewsmith, line 31: These steps were in the original build images release pipeline definition (the GUI one). I took a look to see why it was originally there, but don't see why:
https://msazure.visualstudio.com/One/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=31987

Any idea?

damonbarry: I think it may have been related to IoT Edge on K8s, which is supported in 1.1 but not currently in master and release/1.2.

yophilav (approved):

PR #4485, opened 2021-02-25
Remove sendReportFrequency setting from connectiviity deployment file

yophilav:
- Remove sendReportFrequency setting from connectiviity deployment file


and-rewsmith (approved):
Was this same mistake made in nested connectivity deployment file?

PR #4487, opened 2021-02-25
opentelemetry instrumentation initial metrics in MQTT broker

nlcamp:
Added instrumentation to MQTT broker for a single metric:  mqtt.broker.client.messages.received. This metric is intended to represent the total number of client messages received by the broker.

Testing needs to be done to confirm that performance of the broker is not impacted negatively.

### DIFF
--- a/mqtt/mqtt-otel/src/lib.rs
+++ b/mqtt/mqtt-otel/src/lib.rs
@@ -9,14 +9,17 @@ use std::pin::Pin;
 
 use futures::stream::Stream;
 use futures::StreamExt;
-use opentelemetry::metrics::{self, noop::NoopMeterProvider, MetricsError};
+use opentelemetry::global;
+use opentelemetry::metrics as otel_metrics;
+use opentelemetry::metrics::{noop::NoopMeterProvider, MetricsError};
 use opentelemetry::sdk::metrics::PushController;
-use opentelemetry::{global, KeyValue};
 use opentelemetry_prometheus::PrometheusExporter;
 use prometheus::{Encoder, TextEncoder};
 use std::convert::Infallible;
 use std::result::Result;
 use std::time::Duration;
+
+pub mod metrics;
 
 async fn serve_req(
     _req: Request<Body>,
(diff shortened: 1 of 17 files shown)


vadim-kovalyov:
> Added instrumentation to MQTT broker for a single metric: mqtt.broker.client.messages.received. This metric is intended to represent the total number of client messages received by the broker.
> 
> Testing needs to be done to confirm that performance of the broker is not impacted negatively.

Noel, before the review, is there a high level design or documentation for broker metrics? It would be helpful for us (broker v-team) to understand the design/north star solution to see the whole picture.

mqtt/mqtt-otel/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+        .unwrap();
+
+    Ok(response)
+}
+
+pub fn init_prometheus_metrics_exporter() -> Result<PrometheusExporter, MetricsError> {
+    opentelemetry_prometheus::exporter().try_init()
+}
+
+pub fn create_prometheus_server(
```
varunpuranik, line 42: This will not integrate very well with the EdgeHub container. We need to discuss how to do that..

varunpuranik (commented):
Can we have a design for the MQTT Broker metrics and how they will be added? We also need to make sure they integrate well with the existing EdgeHub metrics.

nlcamp:
> > Added instrumentation to MQTT broker for a single metric: mqtt.broker.client.messages.received. This metric is intended to represent the total number of client messages received by the broker.
> > Testing needs to be done to confirm that performance of the broker is not impacted negatively.
> 
> @nlcamp, before the review, is there a high level design or documentation for broker metrics? It would be helpful for us (broker v-team) to understand the design/north star solution to see the whole picture.

@vadim-kovalyov We skipped formal documentation so we could quickly demonstrate a single metric flowing end to end using open-telemetry APIs. However now that we have that, it does seem like a good idea to pause and provide some high level design documentation. We'll start with high level block diagrams and sequence diagrams and go from there. I'll put the PR back into Draft status until this is completed.

nlcamp:
> Can we have a design for the MQTT Broker metrics and how they will be added? We also need to make sure they integrate well with the existing EdgeHub metrics.

Will do. See reply to @vadim-kovalyov 's comment [here](https://github.com/Azure/iotedge/pull/4487#issuecomment-793042135)

mqtt/mqtt-otel/src/lib.rs
nlcamp: Will include mention/depiction of this in the design docs/diagrams.  These docs/diagrams should help with the discussion about the path forward integrating with the EdgeHub container.

nlcamp:
Closing PR until larger observability design process is completed.

PR #4494, opened 2021-02-25
Longhaul: Enforce test duration as connectivity only setting

and-rewsmith:
Longhaul and nested longhaul currently take test duration, which is a setting that should only be used for connectivity.  Single node longhaul is in a working state right now because empty string is passed for test duration. Nested longhaul twin tests break though because test duration is passed in the wrong format.


e2e_deployment_files/nestededge_bottomLayerBaseDeployment_long_haul_amqp.template.json
```diff
@@ -413,7 +392,7 @@
                 "value": "<TestInfo>"
               },
               "testMode": {
-                "value":  "<TestMode>"
+                "value": "<TestMode>"
```
and-rewsmith, line 395: My vscode autoformatter did this.

and-rewsmith:
Longhaul:
https://msazure.visualstudio.com/One/_build/results?buildId=39803973&view=results

Nested longhaul:
https://msazure.visualstudio.com/One/_build/results?buildId=39803962&view=results

and-rewsmith:
Nested longhaul:
https://msazure.visualstudio.com/One/_build/results?buildId=39805487&view=logs&j=670e6388-39b6-51dc-7138-1d8b97a898e1&t=8cd17f3a-4738-5607-3052-bfd5713bcdb1

and-rewsmith:
I removed the pipeline var from all longhaul pipelines

scripts/linux/trcE2ETest.sh
```diff
@@ -220,8 +219,8 @@ function prepare_test_from_artifacts() {
     fi
 
     if [[ "${TEST_NAME,,}" == "${CONNECTIVITY_TEST_NAME,,}" ]]; then
+        sed -i -e "s@<TestDuration>@$TEST_DURATION@g" "$deployment_working_file"
```
yophilav, line 222: As a preventive measure, I think it would be a good idea to have the shell script check if the parameter is accidentally passed in here. If it is in the wrong "TestMode" it would yell at you.

and-rewsmith: I agree, and we should also do so for all the test-specific vars. I'd rather scope this PR to just fixing the twin tests though. My goal is to see if nested twin tests have any serious issues before we release.

and-rewsmith: I looked into this more and I think we should have a second PR where we get rid of TEST_MODE, use TEST_NAME instead. Then in the same PR we can add these checks based on TEST_NAME.

dylanbronson (approved):
LGTM

PR #4495, opened 2021-02-25
Fixes from bugbash

arsing:
- Change config.toml.edge.template from r-- to rw-

  This allows the user's editor to write it in-place instead of overwriting it.

- Add example of registry auth for bootstrap agent to
  the super-config template.

- Fix `iotedge system` to show its help instead of printing an error.

- Put `sudo` explicitly in the postinst suggestions.

- Update aziotctl-common and use new config helpers.

---

TODO:

- Revert i-i-s deps back to git refs when https://github.com/Azure/iot-identity-service/pull/127 is merged.


varunpuranik (approved):

edgelet/iotedge/src/main.rs
```diff
@@ -247,7 +247,6 @@ fn run() -> Result<(), Error> {
         .subcommand(
             SubCommand::with_name("system")
                 .about("Manage system services for IoT Edge.")
-                .unset_setting(AppSettings::SubcommandRequiredElseHelp)
```
daprilik, line 250: Do we know why this was here to begin with?

daprilik (approved):

edgelet/iotedge/src/main.rs
arsing: I assume Lee copied it from the subcommand above it, `init`, which did need this. At the time both `iotedge init` and `iotedge init import` were valid invocations.

daprilik: Ahh, that makes sense. Alright then.

daprilik (approved):

PR #4497, opened 2021-02-25
Cherry-pick Improve DirectMethod Reporting for long haul

dylanbronson:
Previously, the DirectMethod report only reports 200's, 0's, and unknowns.
The 0's are a known problem having to do with edgeHub restarts.
Unknowns were its own bucket.
This change buckets status codes together so we know exactly which (and how many) codes we got

cherry picked from https://github.com/Azure/iotedge/commit/3b7af90370bcc1bf36d23b0ce750321808dde6eb


yophilav (approved):

vadim-kovalyov:
@yophilav is this still relevant?

vadim-kovalyov:
Please reopen if still relevant.

PR #4501, opened 2021-02-26
[iotedge check] bugbash fixes

daprilik:
- Temporarily remove aziot-edged version check until we sort out where the version info for 1.2 will live
- Remove references to config.yaml, replace with broader "configuration" terminology

### DIFF
--- a/edgelet/iotedge/src/check/checks/well_formed_config.rs
+++ b/edgelet/iotedge/src/check/checks/well_formed_config.rs
@@ -12,10 +12,10 @@ pub(crate) struct WellFormedConfig {}
 
 impl Checker for WellFormedConfig {
     fn id(&self) -> &'static str {
-        "config-yaml-well-formed"
+        "config-toml-well-formed"
     }
     fn description(&self) -> &'static str {
-        "config.yaml is well-formed"
+        "config.toml is well-formed"
     }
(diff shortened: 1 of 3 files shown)


edgelet/iotedge/src/check/checks/well_formed_config.rs
```diff
@@ -12,10 +12,10 @@ pub(crate) struct WellFormedConfig {}
 
 impl Checker for WellFormedConfig {
     fn id(&self) -> &'static str {
-        "config-yaml-well-formed"
+        "config-toml-well-formed"
```
arsing, line 15: It's not a single TOML here either. Just make it `config` and `"configuration"` like in your i-i-s PR.

daprilik: Yep, sorry, that's my pre-lunch hunger manifesting itself.

arsing (approved):

arsing:
A clippy failure because the field read by the latest-version check is no longer used.

daprilik:
I decided to disable the check "from the inside", which keeps all the changes / clippy annotations limited to one file, thereby making it easy to revert the change in the future.

arsing (approved):

PR #4467, opened 2021-02-23
Move NestedEdgeEnabled out of experimental features.

vadim-kovalyov:
### NOTE: After this PR is merged, Nested Edge would be enabled by default.
There are two main changes in this PR:
- Enable nested edge by default
- Move `NestedEdgeEnabled` settings out of `experimentalFeatures` section.

### Before:
To enable nested edge/broker you would need to have the following env vars in your deployment manifest:
```json
                    "edgeHub": {
                        "type": "docker",
                        "status": "running",
                        "restartPolicy": "always",
                        "settings": {
                            "image": "edgerelease.azurecr.io/microsoft/azureiotedge-hub:1.2.0-rc3-linux-amd64",
                            "createOptions": "{\"HostConfig\": {\"PortBindings\": {\"1883/tcp\": [{\"HostPort\": \"1883\"}], \"8883/tcp\": [{\"HostPort\": \"8883\"}],\"443/tcp\": [{\"HostPort\": \"443\"}],\"5671/tcp\": [{\"HostPort\": \"5671\"}]}}}"
                        },
                        "env": {
                            "experimentalFeatures__nestedEdgeEnabled": {
                                "value": "true"
                            },
                            "experimentalFeatures__mqttBrokerEnabled": {
                                "value": "true"
                            },
                            "experimentalFeatures__enabled": {
                                "value": "true"
                            }
                        }
                    }
```

### After
With this change, the Nested Edge would be on by default. So if you want to run nested + broker, here is how the deployment manifest would look like:
```json
                    "edgeHub": {
                        "type": "docker",
                        "status": "running",
                        "restartPolicy": "always",
                        "settings": {
                            "image": "edgerelease.azurecr.io/microsoft/azureiotedge-hub:1.2.0-rc3-linux-amd64",
                            "createOptions": "{\"HostConfig\": {\"PortBindings\": {\"1883/tcp\": [{\"HostPort\": \"1883\"}], \"8883/tcp\": [{\"HostPort\": \"8883\"}],\"443/tcp\": [{\"HostPort\": \"443\"}],\"5671/tcp\": [{\"HostPort\": \"5671\"}]}}}"
                        },
                        "env": {
                            "experimentalFeatures__mqttBrokerEnabled": {
                                "value": "true"
                            },
                            "experimentalFeatures__enabled": {
                                "value": "true"
                            }
                        }
                    }
```

if for some reason you want to disable nested edge, here's how to do that:
```json
                    "edgeHub": {
                        "type": "docker",
[...]

PR #3762, opened 2020-10-20
[Manifest Signing] Verify Twin Signatures for EA and EH

ggjjj:
Manifest Signing Feature flow in Edge: 
1. The deployment manifest is signed using Manifest signer client (exists in the sample folder of the solution). It adds integrity section for each module in this format below. Signature is computed based on the algorithm used and the hash of the content of the each module's desired properties. 
```json
"$edgeAgent": {
      "properties.desired": {
        "modules": {},
        "runtime": {},
        "schemaVersion": "1.0",
        "systemModules": {},
        "integrity": {
          "header": {
            "signercert": ["signercert1stpart", "signercert2ndpart"],
            "intermediatecacert": ["intermediatecacert1stpart", "intermediatecacert2ndpart"]
          },
          "signature": {
            "bytes": "signaturebytes",
            "algorithm": "ES256"
          }
        }
      }
    },
    "$edgeHub": {
      "properties.desired": {
        "routes": {},
        "schemaVersion": "1.0",
        "storeAndForwardConfiguration": {},
        "integrity": {
         "header": {
            "signercert": ["signercert1stpart", "signercert2ndpart"],
            "intermediatecacert": ["intermediatecacert1stpart", "intermediatecacert2ndpart"]
          },
          "signature": {
            "bytes": "signaturebytes",
            "algorithm": "ES256"
          }
        }
      }
    },
```
2. A signer cert is used to signed the deployment manifest is also added to the integrity section
3. When a signed deployment is deployed, the twins received by the modules are checked for valid signature before consuming the twin.  

What is this PR?
- Twin Signature validation checks on receiving full twin and patches for EA and EH 
- Schema change for EA deployment config
- Schema change for Edge hub config
- Added Test cases 
- Updated Manifest Signer client by removing manifest version 
- Manifest Signer client added as a sample( removed from tools) and now shares the JSON canonicalization from Edge Util
- Modified code coverage settings to exclude the open source code of JSON canonicalization 

What is not in this PR?
- iotedged workload API for manifest trust bundle and modules using the API for chaining verification

PR #4505, opened 2021-02-26
Port missing changes from master to release/1.2

and-rewsmith:
Lots of test automation changes were missing from 1.2 as they were not ported over. **There were however some changes to product code that Bear and I found that weren't included in 1.2. Some of these we thought should have already been included:**

- https://github.com/Azure/iotedge/pull/4469
	- nginx changes by hugues
	- isn't currently included in 1.2. Shouldn't it be? We cherry picked this.
- https://github.com/Azure/iotedge/pull/4481
	- message pump change to edgehub core made by viktor
	- isn't currently included in 1.2. Shouldn't it be? We cherry picked this.

There were also some commits that we wanted to call out:
- https://github.com/Azure/iotedge/pull/4467
	- master was updated with this change then reverted in a later PR. 1.2 doesn't include the original commit or the revert, which makes sense. Therefore we didn't cherry pick this to 1.2.
- https://github.com/Azure/iotedge/pull/3762
	- manifest signing changes by gaya
	- shouldn't be in 1.2, right? We didn't cherry pick this.


ggjjj (commented):
Manifest signing changes were merged after master was released to merge. It shouldn't be in 1.2

vadim-kovalyov (commented):
Can't say for all of these changes, I would let others confirm, but I can confirm that #4467 is not needed, since it has problems and was reverted.

dylanbronson:
#4490 and #4455 are not strictly needed for 1.2, and they don't touch any product code, but if you want to run long haul for 1.2 you should take them

vipeller:
#4481 is a performance improvement, does not cause much problem if missing, but it is better to have, so thanks for merging it

huguesBouvier (approved):
It should be. I was waiting the result of arm32 testing. I had a PR for it actually. Thanks

and-rewsmith:
I'm going to wait until @varunpuranik approves, then will merge

PR #4507, opened 2021-02-26
Adding event hub long haul mode to trc tests

dylanbronson:
Add EventHubLongHaul mode to TRC tests so that we can pass on tests that involve using EventHub. We've seen these tests have incredibly long delays. 

This PR makes the change to implement this solution: ignore expected messages (== ones sent successfully from loadGen) that do not match, only match on actual messages (== ones received from eventHub), and also ensure that we are still receiving from EventHub by 1) Using our own timestamps in the TRC for messages we receive and 2) making sure the last message received for the actual store was within an hour (we can define that tolerance however we want)


test/modules/TestResultCoordinator/Reports/CountingReport.cs
```diff
@@ -19,7 +19,8 @@ class CountingReport : TestResultReportBase
             ulong totalExpectCount,
             ulong totalMatchCount,
             ulong totalDuplicateResultCount,
-            IReadOnlyList<TestOperationResult> unmatchedResults)
+            IReadOnlyList<TestOperationResult> unmatchedResults,
+            Option<bool> stillReceivingFromEventHub)
```
nyanzebra, line 23: Could the `stillReceivingFromEventHub` be a default value? Looks like most of the cases in this pr are the same. Also, does this need to be optional if it is a boolean? As with option, we will have 3 states: Some(true), Some(false), None. I am sure you picked this for a reason, would you mind explaining to me what each state is representing? would help me a lot! :)

test/modules/TestResultCoordinator/Services/PartitionReceiverHandler.cs
```diff
@@ -62,10 +62,8 @@ public async Task ProcessEventsAsync(IEnumerable<EventData> events)
                         {
                             if (long.TryParse(sequenceNumberFromEvent.ToString(), out long sequenceNumber))
                             {
-                                DateTime enqueuedtime = GetEnqueuedTime(deviceIdFromEvent.ToString(), moduleIdFromEvent.ToString(), eventData);
-
                                 // TODO: remove hardcoded eventHub string in next line
-                                var testResult = new MessageTestResult((string)moduleIdFromEvent + ".eventHub", enqueuedtime)
+                                var testResult = new MessageTestResult((string)moduleIdFromEvent + ".eventHub", DateTime.UtcNow)
```
nyanzebra, line 66: Why did this change to UtcNow?

test/modules/TestResultCoordinator/Reports/CountingReportGenerator.cs
```diff
@@ -112,12 +117,30 @@ public async Task<ITestResultReport> CreateReportAsync()
...
+            else
+            {
+                // If we are are using EventHub to receive messages, we see an issue where EventHub can accrue large delays after
+                // running for a while. Therefore, if we are using EventHub with this counting report, we do two things.
+                // 1. We ignore extra expected results (skipping the last hasExpectedResults loop above)
+                // 2. We make sure that the last result we got from EventHub (i.e. the lastLoadedResult) is within 1 hour.
+                //    This is an arbitrary tolerance period that we have defined, and can be tuned as needed.
+                // TODO: There is either something wrong with the EventHub service or something wrong with the way we are using it,
+                // Because we should not be accruing such large delays. If we move off EventHub, we should fix this as well.
+                if (lastLoadedResult.CreatedAt > DateTime.UtcNow - TimeSpan.FromHours(1))
```
nyanzebra, line 138: Is it worthwhile making this a configurable value through settings?

dmolokanov:
In addition, let's save/log the following parameters 
* a number of expected messages 
* a number of messages received from the EventHub
* timestamp of the latest message received from EventHub
*  values of the `stillReceivingFromEventHub`

In future, it will give us some ideas about: how big lag we have for EventHub messages, and whether we calculated `stillReceivingFromEventHub` correctly

test/modules/TestResultCoordinator/Reports/CountingReportGenerator.cs
dylanbronson: You're right it could be worthwhile to make this configurable. So I changed this to be configurable through settings, but I did not actually make a change to the scripts/yamls so that we can change it from our pipeline. I did this because we are already overloaded with configurable options through the pipeline. It makes things difficult to debug. Especially for a setting like this where we will need to "tune it" in the beginning but then likely never set it again, I think it is better to have it so the user can't change it from the pipeline. If we want to change it we can easily make the change to the default value

test/modules/TestResultCoordinator/Services/PartitionReceiverHandler.cs
dylanbronson: Now, we will use our own timestamps in the TRC rather than using EventHub's timestamps. We weren't really doing anything with the eventhub timestamps, and they don't really tell us much. 
I think it's better to use our own timestamps to signal when the TRC actually receives the message (not when eventHub receives the message). It also helps with the logic I'm implementing here. Now we can say "have we received a message from eventhub within the last <EventHubDelayTolerance>," whereas otherwise that is more difficult to answer

test/modules/TestResultCoordinator/Reports/CountingReport.cs
dylanbronson: Making it an option here gives us a benefit of having the 3 states: true, false, and N/A. If it is None, then we are explicitly saying that we can't have a true or a false in this situation. I added a comment to make this more clear

test/modules/Modules.Test/TestResultCoordinator/Reports/CountingReportTest.cs
```diff
@@ -29,7 +30,9 @@ public void TestConstructorSuccess()
                 {
                     new TestOperationResult("expectedSource", "resultType1", "332", new DateTime(2019, 12, 4, 10, 15, 15)),
                     new TestOperationResult("expectedSource", "resultType1", "734", new DateTime(2019, 12, 4, 10, 15, 18)),
-                });
+                },
+                Option.None<bool>(),
+                new DateTime(2019, 12, 4, 10, 15, 19));
```
nyanzebra, line 35: Can this be a constant?

nyanzebra (approved):
:shipit:

dmolokanov (dismissed):
Overall LGTM! Let's wait a bit more to verify in long haul

dmolokanov (approved):
🚀

PR #4509, opened 2021-02-26
Reverting to nginx image (#4469)

huguesBouvier:
Reverting to nginx since they updated their base image and because there has been a report of the arm32 image crashing and I was not able to reproduce it/fix it(9345043). So I am reverting for release rc4 until I can fix the issue.

https://dev.azure.com/msazure/One/_build/results?buildId=39734886&view=results
https://dev.azure.com/msazure/One/_build/results?buildId=39734935&view=results


nyanzebra (approved):

varunpuranik (changes requested):
This PR is obsolete, I suppose? Move to bionic instead?

PR #4517, opened 2021-03-01
Increase the single node E2E pipeline timeout

yophilav:
Increase the single node E2E pipeline timeout


test/Microsoft.Azure.Devices.Edge.Test/Provisioning.cs
```diff
@@ -37,6 +37,7 @@ string DeriveDeviceKey(byte[] groupKey, string registrationId)
 
         [Test]
         [Category("CentOsSafe")]
+        [Category("Unstable")]
```
nyanzebra, line 40: Do we have any associated task to stabilize these? If so, is it worth having a small comment with the task number?

nyanzebra (approved):

PR #4518, opened 2021-03-01
Fix wrap-around bug

nyanzebra:
# Description
This is a bug fix for a bug I have recently found. I assume I have just gotten really lucky and that is why it took a while to find. 

# The problem
When RingBuffer wraps around it is possible to write a smaller block than what was previously there.
This means we have a buffer that looks like this:
+------------+
|  2  | 1  |0000 |
+------------+

Where 2 is smaller than previous block and when the next insert happens and we check to see if can overwrite the block we will end up seeing garbage and may or may not do it.

# The fix
We will allow for writes if we cannot determine if block header is valid. Other existing conditions will allow for protection against most corruptions, we just need to allow for write in the case of wrap around and writing a smaller block before hand. We do not want to skip beyond the fragmented block/data to try to avoid the problem as it will result in lots of fragmentation. Instead we want to do as described above and just allow the write and continue on. As we do not allow for write to pass read or read to pass write.

# Todo (later prs)
- [ ] Should merge into release 1.2


mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -952,6 +966,59 @@ mod tests {
...
+            topic_name: "test".to_owned(),
+            qos: QoS::AtMostOnce,
+            retain: true,
+            payload: Bytes::new(),
+        };
+
+        let block_size = *SERIALIZED_BLOCK_SIZE;
+
+        let result = bincode::serialize(&publication);
+        assert_matches!(result, Ok(_));
```
vadim-kovalyov, line 982: looks a bit odd, I think we just do `.unwrap()` or better `.expect("...")` in tests...

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -225,8 +225,22 @@ impl StreamWakeableState for RingBuffer {
...
+            if bin_block.iter().map(|&x| u64::from(x)).sum::<u64>() != 0
+                && inner.hint() != BLOCK_HINT
+            {
+                // This is a bad block, where we wrote something smaller
+                // and this is okay, we just need to allow the write.
+                allow_write = true;
+            }
             let should_not_overwrite = inner.should_not_overwrite();
-            if should_not_overwrite {
+            if !allow_write && should_not_overwrite {
```
vadim-kovalyov, line 243: I'm not sure I completely understand the issue. Is it that you were able to deserialize block header from junk data and `should_not_overwrite` gave you false positive?

vadim-kovalyov: Also, I'm more and more skeptical about this check at all. I understand we still need `should_not_overwrite` flag for crash recovery, but do we need that for normal work? can we just rely on pointers?

vadim-kovalyov (commented):
It looks good to me overall, but I have some questions and would like @dmolokanov to take a look as well.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -225,8 +225,22 @@ impl StreamWakeableState for RingBuffer {
         let result = load_block_header(&mut self.file, start, block_size, self.max_file_size);
         if let Ok(block_header) = result {
             let BlockVersion::Version1(inner) = block_header.inner();
+            // It is possible that we end up writing a smaller block on wrap-around,
+            // where the file pointers will end up being in the middle of random data.
+            // To avoid this, we need to check if we have a block of data, if we do,
+            // we will need to just allow the write. Above cases will prevent us from
+            // accidentally corrupting.
+            let mut allow_write = false;
```
dmolokanov, line 233: In general, mutating is error-prone esp. in multithreaded runtime. I usually try to prioritize immutable over mutable operations whenever possible.

```rust
let allow_write = bin_block.is_zeroed() && inner.hint() != BLOCK_HINT;
```

dmolokanov (dismissed):
Can we make in-memory storage by default while continuing testing ring buffer?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: The issue is that yes, we deserialized from junk data, which is okay since we had data there and we are now writing new data which does not cover the whole of previous data. We can get whatever value happens to map over that region of data, whether it is true or false. 

It can be that write goes faster than read. The `should_not_overwrite` flag prevents us from writing when we shouldn't. But, in the case of normal read and write operations without crash, we can likely get away with just the pointers. This would mean we can drop this check altogether, which should make this look nicer too

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Should I fix all of these throughout the tests in this PR or just these?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: As what Vadim mentioned above, I think this section can be dropped entirely.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: I can also open up another pr to get all of these as well if that is preferred :)

nyanzebra: Went ahead and changed these for this test, then if we want to change the others later we can :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Have removed this check entirely and it operates well without it :)

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -31,7 +33,16 @@ impl BridgeSettings {
...
+        config.merge(Environment::new().separator(ENV_VAR_SEPARATOR))?;
+
+        if let Ok(default_storage_memory) =
+            config.get_bool(&DEFAULT_STORAGE_MEMORY_VAR.to_ascii_lowercase())
+        {
+            if default_storage_memory {
+                config.set("storage.type", "memory")?;
+                config.set("storage.max_size", 1024 * 1024 * 1024)?;
+            }
+        }
```
nyanzebra, line 45: I have added an env var to default over to memory storage, let me know if there is a better way or name that you can think of here... just went with something simple

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -44,7 +55,16 @@ impl BridgeSettings {
 
         config.merge(File::from_str(DEFAULTS, FileFormat::Json))?;
         config.merge(File::from(path.as_ref()))?;
-        config.merge(Environment::new())?;
+        config.merge(Environment::new().separator(ENV_VAR_SEPARATOR))?;
```
nyanzebra, line 58: Added to try to be in line with https://github.com/Azure/iotedge/blob/master/doc/EnvironmentVariables.md Vadim, I know you recently worked on env vars, so if there is a style I should change to, let me know!

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -326,6 +348,46 @@ impl RingBufferSettings {
...
 }
 
+#[derive(Deserialize)]
+#[serde(untagged)]
+enum NonZeroU64OrOther {
+    U64(u64),
+    STR(String),
+}
+
+fn deserialize_with_u64<'de, D>(deserializer: D) -> Result<NonZeroU64, D::Error>
```
nyanzebra, line 358: Should allow for conversion of string or int now, which will make env var or json easy to use

PR #4525, opened 2021-03-02
[mqtt] Promote client lifecycle logs to info

dmolokanov:
Right now we don't log client lifecycle events when "INFO" is the max log level. It makes it really hard to debug in these circumstances.

`log::info!` when:
- [x] client connected
- [x] client disconnected gracefully
- [x] broker drops connection for the client
- [x] broker refuses to connect a client

### DIFF
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -349,7 +349,7 @@ where
         // section 3.2 and it MUST close the Network Connection.
         let auth_id = match connreq.auth() {
             Auth::Identity(auth_id) => {
-                info!(
+                debug!(
                     "client {} successfully authenticated: {}",
                     client_id, auth_id
                 );
@@ -376,7 +376,7 @@ where
         let activity = Activity::new(client_info, operation);
         match self.authorizer.authorize(&activity) {
             Ok(Authorization::Allowed) => {
-                info!("successfully authorized: {}", &activity);
+                debug!("successfully authorized: {}", &activity);
             }
             Ok(Authorization::Forbidden(reason)) => {
                 warn!("not authorized: {}; reason: {}", &activity, reason);
@@ -994,7 +994,7 @@ where
                 match session.subscribe_to(subscribe_to) {
                     Ok((qos, subscription)) => {
                         if let Some(subscription) = subscription {
-                            info!("subscribed to {} with qos {:?}", subscription.filter(), qos);
+                            debug!("subscribed to {} with qos {:?}", subscription.filter(), qos);
                             subscriptions.push(subscription);
                         }
 
--- a/mqtt/mqtt-broker/src/broker.rs
+++ b/mqtt/mqtt-broker/src/broker.rs
@@ -306,6 +306,8 @@ where
                 debug!("dropping connection due to connect packet error");
                 let message = Message::Client(client_id, ClientEvent::DropConnection);
                 try_send!(connreq.handle_mut(), message);
+
+                info!("client connection refused");
             };
         };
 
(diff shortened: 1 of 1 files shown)


vadim-kovalyov (changes requested):
I'm not sure I agree with these changes. I had quite different experience with the logs when I was bugbashing nested edge + bridge. I think we need to discuss that offline.

nyanzebra (commented):
I will approve after addressing Vadim's comment and confirming with Viktor that these log changes are everything that he wants :) Feel free to message me if need the approval

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -306,6 +306,8 @@ where
                 debug!("dropping connection due to connect packet error");
                 let message = Message::Client(client_id, ClientEvent::DropConnection);
                 try_send!(connreq.handle_mut(), message);
+
+                info!("client connection refused");
```
varunpuranik, line 310: Is this useful? Do you want to include clientId and reason why it was refused?

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -374,7 +376,7 @@ where
         let activity = Activity::new(client_info, operation);
         match self.authorizer.authorize(&activity) {
             Ok(Authorization::Allowed) => {
-                debug!("successfully authorized: {}", &activity);
+                info!("successfully authorized: {}", &activity);
```
varunpuranik, line 379: Does this need to be info? Feels like it is only useful when not authorized?

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -428,7 +430,7 @@ where
             }
         }
 
-        debug!("connect handled.");
+        info!("connect handled.");
```
varunpuranik, line 433: Again - with multiple clients, I feel it is useful to include the client_id as well.

mqtt/mqtt-broker/src/broker.rs
```diff
@@ -992,8 +994,10 @@ where
                 match session.subscribe_to(subscribe_to) {
                     Ok((qos, subscription)) => {
                         if let Some(subscription) = subscription {
+                            info!("subscribed to {} with qos {:?}", subscription.filter(), qos);
```
varunpuranik, line 997: client_id?

mqtt/mqtt-broker/src/broker.rs
dmolokanov: client_id is in the context of trace event. also before each refuse_connection call, there is a warning message

varunpuranik: I am not sure i follow.. 
> client_id is in the context of trace event.
Isn't this change to address the case when only info logging is enabled?
>also before each refuse_connection call, there is a warning message
and that has the info? Why do we still need this then?

dmolokanov: https://github.com/Azure/iotedge/pull/4525/files#diff-c37c4acce39447d43b2fbcf9c1698aceee26bc6e065afb0f165b85a89f2be923L334-L341

1. before each `refuse_connection!` call we have a warn message explaining what went wrong. so I don't think we need to provide refuse connection message again with info level.
2. since we are using tracing a not logging we can use trace event context and put additional data into it like client_id of the client which packet we process. so we don't need to pass client_id to every log message because it is already there

vadim-kovalyov: Can you please make this `warn!`

vadim-kovalyov (commented):
I'm fine with adding 2 new logs if that helps debugging connectivity issues, but I'm not sure I agree with changing some `debug!` logs to `info!`. For me it is unnecessary and does not help to solve immediate issues.

mqtt/mqtt-broker/src/broker.rs
vadim-kovalyov: Or rather why not update the `debug!` log above? (line 306). Also, it would be good to include reason code.

dmolokanov:
Closing this PR since the whole log should be revised.

PR #4526, opened 2021-03-02
Fix license type in aziot-edge.spec

gordonwang0:
Fix typo of license type in aziot-edge.spec.

### DIFF
--- a/edgelet/contrib/centos/aziot-edge.spec
+++ b/edgelet/contrib/centos/aziot-edge.spec
@@ -10,7 +10,7 @@ Name:           aziot-edge
 Version:        @version@
 Release:        @release@%{?dist}
 
-License:        Proprietary
+License:        MIT
 Summary:        Azure IoT Edge Module Runtime
 URL:            https://github.com/azure/iotedge
 


arsing:
There's no need to do this here. Do it in i-i-s. The a-i-s package already conflicts with iotedge, so just need to add libiothsm-std to the list.

gordonwang0:
I did it here because aziot-edge is more equivalent to iotedge than a-i-s. It shouldn't be a problem to install a-i-s with iotedge+libiothsm, but installing aziot-edge should remove iotedge+libiothsm.

There's another PR I made in i-i-s that moves the conflict to this package. https://github.com/Azure/iot-identity-service/pull/139

arsing:
>I did it here because aziot-edge is more equivalent to iotedge than a-i-s. It shouldn't be a problem to install a-i-s with iotedge+libiothsm, but installing aziot-edge should remove iotedge+libiothsm.

a-i-s is the equivalent of iotedge because they both do provisioning, and only one should be able to do the provisioning. Installing a-i-s absolutely must remove iotedge. The way it was implemented is correct.

gordonwang0:
Got it. I'll fix it then.

There's a typo in the aziot-edge.spec file for the license. Would still like to fix that.

arsing (approved):
Update the PR title and OP before kodiak merges it.

PR #4527, opened 2021-03-02
MQTT Perf Tests improvements.

vadim-kovalyov:
There are few small improvements I propose for MQTT perf tests:
- use `--security-opt apparmor=unconfined` flag to disable apparmor that sometimes causes problems restarting the containers
- use `docker rm -f` && `docker run` instead of `docker restart`
- save image name into local file, for ease of use in later steps.
- make all stages always run even if previous failed.

### DIFF
--- a/builds/misc/templates/mqtt-perf-setup.yaml
+++ b/builds/misc/templates/mqtt-perf-setup.yaml
@@ -67,7 +67,7 @@ jobs:
         commands: |
           echo "$(registry.password)" | docker login $(registry.address) --username $(registry.user) --password-stdin 2>/dev/null
           docker pull $(cat brokerImageName.txt)
-          docker run -d --network=host --security-opt apparmor=docker-default --name=mqtt-broker $(cat brokerImageName.txt)
+          docker run -d --network=host --security-opt apparmor=unconfined --name=mqtt-broker $(cat brokerImageName.txt)
           docker logout
       condition: eq(variables['arch'], 'arm32v7')
       
--- a/builds/misc/templates/mqtt-perf-setup.yaml
+++ b/builds/misc/templates/mqtt-perf-setup.yaml
@@ -42,8 +42,8 @@ jobs:
         runOptions: commands
         commands: |
           echo "$(registry.password)" | docker login $(registry.address) --username $(registry.user) --password-stdin 2>/dev/null
-          docker pull $$(cat brokerImageName.txt)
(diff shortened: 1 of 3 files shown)


builds/misc/templates/mqtt-perf-setup.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
         sshEndpoint: iotedge-mqtt-perf-amd64-conn
         runOptions: commands        
         commands: |
-          docker stop mqtt-broker ; docker system prune -a -f
+          echo $(brokerImageName) > brokerImageName.txt ; docker stop mqtt-broker ; docker system prune -a -f
```
varunpuranik, line 33: why do we need to do this?

builds/misc/templates/mqtt-perf-setup.yaml
```diff
@@ -42,8 +42,8 @@ jobs:
         runOptions: commands
         commands: |
           echo "$(registry.password)" | docker login $(registry.address) --username $(registry.user) --password-stdin 2>/dev/null
-          docker pull $(brokerImageName)
-          docker run -d --network=host --name=mqtt-broker $(brokerImageName)
+          docker pull $(cat brokerImageName.txt)
+          docker run -d --network=host --security-opt apparmor=unconfined --name=mqtt-broker $(cat brokerImageName.txt)
```
varunpuranik, line 46: I am okay with this.. but what problems are we running into with the default profile?

varunpuranik (approved):

builds/misc/templates/mqtt-perf-setup.yaml
vadim-kovalyov: We have a "Cleanup" step down the pipeline, and it needs an image name to re-create the container. The problem is, to pass env var from one job to another, you need to write a an ugly echo command on every step. Saving the image name into the local file seems a good workaround.

```
echo "##vso[task.setvariable variable=brokerImageName;]$brokerImageName"

builds/misc/templates/mqtt-perf-setup.yaml
vadim-kovalyov: Sometimes, `docker rm -f` or `docker restart` failed with permission denied issue. Googling showed that this is b/c of apparmor. I'm not sure about the details, but I have not had this issue again since I disabled apparmor.

PR #4528, opened 2021-03-03
Enable the Dps Tests

yophilav:
The DPS tests were previously disabled to avoid a known DPS issue. 
Now the DSP issue is fixed. Time to bring back the tests!

### DIFF
--- a/builds/e2e/e2e.yaml
+++ b/builds/e2e/e2e.yaml
@@ -39,7 +39,7 @@ jobs:
       identityServiceArtifactName: packages_debian-9-slim_arm32v7
       identityServicePackageFilter: aziot-identity-service_*_armhf.deb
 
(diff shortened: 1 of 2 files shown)


and-rewsmith (approved):
Commiting directly to 1.2 imo is a bit dangerous. Might be better in the future to merge to master first and then cherry pick to release branches.

Having a consistent cherry pick strategy will make it easier come release time.

yophilav:
> Commiting directly to 1.2 imo is a bit dangerous. Might be better in the future to merge to master first and then cherry pick to release branches.
> 
> Having a consistent cherry pick strategy will make it easier come release time.

Well.... The thing is... Master didn't get its DPS disabled. Only release1.2 did (the fix turn around was <4hr). lol

PR #4532, opened 2021-03-03
Ignore validity in cert API requests

massand:
Our current logic within `iotedge config import` discards the `auto_generated_edge_ca_expiry_days`. We need to respect the old value from the config.yaml during import. This change also ignores the module certificate validity information, since this is already determined by CS policy, (which is also set during `iotedge config` as well).

Changes:
- Ignore validity in cert API requests
- Add support for old quickstart default expiration
- Update test and template TOMLs


arsing:
Wait, we want to do this for module certs too? I thought we were only going to do this for the Edge CA cert. If modules want a smaller-lifetime cert than the default / certd-configured value, why not let them have that?

massand:
The expiration is a CA side policy. Which is why [CSRs](https://docs.rs/openssl/0.10.32/openssl/x509/struct.X509ReqBuilder.html) don't have the ability for expiration to be specified.  Unless, we want to expose a new CS API to set the expiration, how else would we pass this?

arsing:
>Which is why CSRs don't have the ability for expiration to be specified.

Oh, I never noticed that. I thought it was optional and the CA could ignore it.

So yes, you're only limiting the artificial limitation within edged. That's fine then.

---

Related: We should add the ability to certd config to specify cert issuances on wildcard IDs, so that the user can override the default for all module certs and such. It would also be nice to let them specify a wildcard for all module certs but still provide a specific issuance for a particular module cert. That'd be a bit complicated because it has to consider exact matches before wildcards though, so we'd have to change the issuance table from a `BTreeMap` to a `Vec`, build the table in config-file order, and expect the user to put exact matches before wildcards (and stronger wildcards before looser ones).

edgelet/contrib/config/linux/template.toml
```diff
@@ -28,6 +28,21 @@
 # trust_bundle_cert = "file:///var/secrets/trust-bundle.pem"
 
 
+# ==============================================================================
+# Quickstart certificate expiration
```
varunpuranik, line 32: I think we should put this next to the Edge CA cert section.

massand: There is a limitation of our current TOML parser where individual elements need to precede the tables. So, individual properties need to be at the top.

arsing: Put it *in* the `edge_ca` section? It only applies to the Edge CA after all.

massand: I went back and forth on this one, but ultimately I went with this approach because it is mutually exclusive of the other options in `edge_ca`. Putting it under `edge_ca` makes the auto-generation behavior seem related to the customer-specified cert, when its not. Previously, in 1.1, this section was called `certificates` where the trust-bundle, device CA and auto-generated workload CA properties were all bundled up in one. If `trust_bundle_cert` is split out, so should this.

arsing: I think both Varun and Micah asked for `trust_bundle_cert` to be moved into `[edge_ca]` too actually, for exactly that reason.

massand: We should not conflate these two issues. Values that are mutually exclusive should not be co-located in the configuration. The config file clearly states that `edge_ca` is a customer-defined configuration, while `auto_generated_edge_ca_expiry_days` is a configuration, if there is no customer-defined `edge_ca`. The relationship between `trust_bundle` and `edge_ca` is different from the `auto_generated_edge_ca_expiry_days` and `edge_ca` relationship in that `edge_ca` value needs to be chained upto `trust_bundle` to be useful, while `auto_generated_edge_ca_expiry_days` is independent and mutually exclusive of both super-config values.

arsing:
```rust
#[serde(untagged)]
enum EdgeCa {
    Explicit {
        cert: Url,
        pk: Url,
    },
    QuickStart {
        auto_generated_expiry_days: u16,
    },
}
```

massand: Two questions:
- Are you suggesting this for super-config only? Or do we want to keep an `[edge_ca]` section in edged.toml
- Will `Explicit` enum value contain `trust_bundle_cert` too then?

arsing: I've only been talking about the super-config.

---

If I personally had to choose whether `trust_bundle_cert` is in `[edge_ca]` or not, I would choose not to put it there. I only mentioned Varun and Micah asking for it. I believe their argument is that the only time the user cares about the `trust_bundle_cert` is when they're setting it up with nested Edge, in which case it'd be nice to have it bear the other property they have to set up for nested Edge - the Edge CA.

(In 1.1 there was a reason to colocate them even for non-nested Edge - the user had to make sure the trust bundle cert includes the device CA. This is no longer needed for 1.2, so apparently only the nested Edge use case remains.)

massand: Ok. I have replaced `EdgeCa` struct with the suggested enum. Thanks.

edgelet/iotedge/src/config/super_config.rs
```diff
@@ -30,9 +30,15 @@ pub(super) struct Config {
...
-    pub(super) cert: Url,
-    pub(super) pk: Url,
+#[serde(untagged)]
+pub(super) enum EdgeCa {
+    Explicit {
+        cert: Url,
+        pk: Url,
+    },
+    Quickstart {
+        auto_generated_edge_ca_expiry_days: u16,
```
arsing, line 40: Make it a `u32` to match the `expiry_days`

edgelet/iotedge/src/config/import/old_config/mod.rs
```diff
@@ -209,7 +209,4 @@ homedir: '/var/lib/iotedge'
 moby_runtime:
   uri: 'unix:///var/run/docker.sock'
   network: 'azure-iot-edge'
-
```
arsing, line 212: This was the old default, so keep it.

edgelet/iotedge/src/config/apply.rs
```diff
@@ -207,37 +207,46 @@ fn execute_inner(
...
+        }
+        super_config::EdgeCa::Quickstart {
+            auto_generated_edge_ca_expiry_days,
+        } => {
+            certd_config.cert_issuance.certs.insert(
+                edgelet_core::AZIOT_EDGED_CA_ALIAS.to_owned(),
+                aziot_certd_config::CertIssuanceOptions {
+                    method: aziot_certd_config::CertIssuanceMethod::SelfSigned,
+                    common_name: Some(IOTEDGED_COMMONNAME.to_owned()),
+                    expiry_days: Some(auto_generated_edge_ca_expiry_days.into()),
```
arsing, line 239: Remove this `.into()` when you change the type to `u32`

arsing (changes requested):

edgelet/iotedge/src/config/apply.rs
massand: Fixed. The `into()` got moved into the `import\mod.rs` where the original config type is `u16`

edgelet/iotedge/src/config/import/old_config/mod.rs
massand: Fixed. I also reflected the change in the import test files

edgelet/iotedge/src/config/super_config.rs
massand: Fixed.

arsing (approved):

arsing (approved):

PR #4533, opened 2021-03-03
Update postrm to delete iotedge user on purge

gordonwang0:
Updates postrm so that apt purge fully removes all files used by aziot-edge and removes iotedge user.

This only applies to Debian, as CentOS (rpm) doesn't have a purge command.

### DIFF
--- a/edgelet/contrib/debian/postrm
+++ b/edgelet/contrib/debian/postrm
@@ -21,6 +21,7 @@ case "$1" in
         done
 
         /usr/sbin/userdel iotedge
+        rm -rf /var/lib/aziot/edged
 
         if [ -d /var/lib/aziot ] && [ -z "$(ls -A /var/lib/aziot)" ]; then
             rm -rf /var/lib/aziot
--- a/edgelet/contrib/debian/postrm
+++ b/edgelet/contrib/debian/postrm
@@ -7,10 +7,7 @@ case "$1" in
 
         if [ -d /etc/aziot ]; then
             rm -rf /etc/aziot/edged
-
-            if [ -z "$(ls -A /etc/aziot)" ]; then
-                rm -rf /etc/aziot
-            fi
+            rm -f /etc/aziot/config.toml
         fi
 
         rm -rf /var/log/aziot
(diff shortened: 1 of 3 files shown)


edgelet/contrib/debian/postrm
```diff
@@ -3,22 +3,32 @@ set -e
...
     purge)
-        if [ -f /etc/aziot/edged/config.toml ]; then
-            rm /etc/aziot/edged/config.toml
-        fi
-        if [ -d /etc/aziot/edged/config.d ]; then
-            rm -r /etc/aziot/edged/config.d
-        fi
-        if [ -d /var/lib/aziot/edged ]; then
-            rm -rf /var/lib/aziot/edged
+        systemctl daemon-reload
```
arsing, line 6: Why do you need this?

edgelet/contrib/debian/postrm
```diff
@@ -3,22 +3,32 @@ set -e
...
+
+        if [ -d /etc/aziot ]; then
+            rm -rf /etc/aziot/edged
```
arsing, line 9: Can you just delete `/etc/aziot/edged` and `/etc/aziot/config.toml` and let a-i-s's postrm handle the rest? Do postrm's run in a defined order so that a-i-s's will always run after this one?

edgelet/contrib/debian/postrm
gordonwang0: This clears the uninstalled package from commands like systemctl status, etc.

edgelet/contrib/debian/postrm
gordonwang0: It's possible to uninstall aziot-edge without installing aziot-identity-service, right? Not that I would expect many people to do that, but currently this postrm is taking that into account.

gordonwang0: Although now that I think about it, yes, it's fine to just delete `/etc/aziot/edged/` and `/etc/aziot/config.toml`. Fixed.

arsing (dismissed):

arsing (approved):

gri6507:
Is there a comparable fix for iotedge 1.1.x?

PR #4538, opened 2021-03-04
Introduce nested edge device connection to grandparent verification test

yophilav:
- Introduce nested edge device connection to grandparent verification test


test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeDevice.cs
```diff
@@ -14,20 +15,51 @@ public class EdgeDevice
         readonly IotHub iotHub;
         readonly bool owned;
 
-        EdgeDevice(Device device, bool owned, IotHub iotHub)
+        readonly NestedEdgeConfig nestedEdgeConfig;
+
+        public class NestedEdgeConfig
+        {
+            public readonly string DeviceHostname;
```
ancaantochi, line 22: these should be properties: public string DeviceHostname { get; }

test/Microsoft.Azure.Devices.Edge.Test.Common/EdgeDevice.cs
```diff
@@ -51,27 +83,30 @@ public class EdgeDevice
 
         public static async Task<Option<EdgeDevice>> GetIdentityAsync(
             string deviceId,
-            Option<string> parentDeviceId,
             IotHub iotHub,
             CancellationToken token,
-            bool takeOwnership = false)
+            bool takeOwnership = false,
+            NestedEdgeConfig nestedEdgeConfig = default(NestedEdgeConfig))
```
ancaantochi, line 89: preferred to use Option instead of null

test/Microsoft.Azure.Devices.Edge.Test/AuthorizationPolicy.cs
```diff
@@ -160,7 +160,7 @@ public async Task AuthorizationPolicyUpdateTest()
                 $"HostName={this.IotHub.Hostname};" +
                 $"DeviceId={leaf.Id};" +
                 $"SharedAccessKey={leaf.Authentication.SymmetricKey.PrimaryKey};" +
-                $"GatewayHostName={Context.Current.Hostname.GetOrElse(Dns.GetHostName().ToLower())}";
+                $"GatewayHostName={this.device.NestedEdge.DeviceHostname}";
```
ancaantochi, line 163: is the ToLower() required, just wondering if this can cause the test issue

huguesBouvier (changes requested):
Hey, ran the E2E test on this PR: https://dev.azure.com/msazure/One/_build/results?buildId=40273660&view=results
But got the following issues:
1. only a subset of test are running => Remove your debug only
2. Some test are failing:TransparentGateway

You can compare with a recent run on master: https://dev.azure.com/msazure/One/_build/results?buildId=40275354&view=logs&j=e8afecc0-c259-5cc1-62e0-5c5d1c2a3db4

builds/e2e/templates/e2e-run.yaml
```diff
@@ -42,7 +42,10 @@ steps:
       $filter += '&Category!=NestedEdgeOnly' 
     } 
 
-    sudo --preserve-env dotnet test $testFile --logger:trx --testcasefilter:$filter 
+    # sudo --preserve-env dotnet test $testFile --logger:trx --testcasefilter:$filter
+
+    # BEARWASHERE -- DEBUG ONLY
```
huguesBouvier, line 47: Bear should not be in production code :)

yophilav:
Do you believe in Code Fermentation? 
Without changing a thing, it seems like my code is now passing on this test run: 
https://msazure.visualstudio.com/One/_build/results?buildId=40699021&view=results

huguesBouvier (changes requested):
Looking at your run on nested E2E, several tests are failing. Let's dig into them

huguesBouvier (approved):

PR #4539, opened 2021-03-04
Fix flaky proptest

vadim-kovalyov:
It looks like that in prev version of regex it is possible that `.*` can eval into a digit, which in turn will lead to a valid input :)
```
".*[0-9]{9}.*(k|K|m|M|g|G)?.*"
```


mqtt/mqtt-broker/src/settings/size.rs
```diff
@@ -257,7 +257,7 @@ mod tests {
         }
 
         #[test]
-        fn it_cannot_parse_input(input in r".*[0-9]{9}.*(k|K|m|M|g|G)?.*") {
+        fn it_cannot_parse_input(input in r"[^0-9]+[0-9]{9}[^0-9]+(k|K|m|M|g|G)?.*") {
```
nyanzebra, line 260: Was this validated with another regex validator to double check it matches only what we expect? Would it be worthwhile to have this kind of test be fuzzed?

dmolokanov (approved):
I guess it fails on arm32? :)

mqtt/mqtt-broker/src/settings/size.rs
vadim-kovalyov: I did a quick check of the regex, and at least I was not able to repro the issue with the previously reported failed value. I believe it is very tiny feature to warrant fuzz tests.

PR #4540, opened 2021-03-04
Drop messages when device is not in scope and auth mode is Scope

ancaantochi:
changed https://github.com/Azure/iotedge/commit/756b83b9062624a3d0c2e55f59a56195cc6704b8 picked from release/1.1 and added chain validation
added service identity validation to BrokeredCloudConnectionProvider


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
```diff
@@ -166,6 +169,21 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string deviceId, s
             return serviceIdentityResult;
         }
 
+        Exception MapException(DeviceScopeApiException ex)
```
davilu, line 172: As discussed, not sure if upstream will return the same response as Hub's. At least manual test is required for nested edge scenario.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -188,6 +195,77 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string id)
...
+
+            if (refreshCachedIdentity)
+            {
+                Events.RefreshingAuthChain(authChain);
+
+                await this.RefreshServiceIdentities(ids);
+                authChain = (await this.serviceIdentityHierarchy.TryGetAuthChain(id)).Value;
+            }
+
+            ids = AuthChainHelpers.GetAuthChainIds(authChain);
```
varunpuranik, line 245: Move it inside the if condition above?

varunpuranik (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/ServiceProxy.cs
ancaantochi: tested in nested env and it works with amqp or mqtt older protocol head, with new mqtt broker in nested env it won't detect when there is unauthorized so tracking that as a separate issue that can be solved with MQTT5 support.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -188,6 +197,67 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string id)
...
+        }
+
+        async Task<string> VerifyServiceIdentityAuthChainState(string id, bool refreshCachedIdentity = false)
+        {
+            if (refreshCachedIdentity)
+            {
+                Events.RefreshingServiceIdentity(id);
+
+                var authChainTry = await this.serviceIdentityHierarchy.TryGetAuthChain(id);
+                var actorDeviceId = AuthChainHelpers.GetAuthParent(id, authChainTry.Ok());
```
davilu, line 236: Looks incorrect. `actorDeviceId` should always be `edgeDeviceId`, not parent.

ancaantochi: I renamed the variable to onBehalfOfDeviceId because that's what it actually is.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
```diff
@@ -27,6 +27,18 @@ public static Option<string> GetAuthTarget(Option<string> authChain)
...
         }
 
+        public static Option<string> GetAuthParent(string targetId, Option<string> authChain)
+        {
+            return authChain.Match(
+                chain =>
+                {
+                    string[] authChainIds = GetAuthChainIds(chain);
+                    // The auth target is second element after the target
+                    return authChainIds.FirstOption(id => id != targetId);
```
varunpuranik, line 37: This seems a bit loose, if you are making assumptions about how the chains should be, then it will be better to validate that. Something like 
if(authChainIds[0] == targetId) return authChainIds[1];

varunpuranik (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/DeviceScopeIdentitiesCache.cs
```diff
@@ -188,6 +197,67 @@ public async Task<Option<ServiceIdentity>> GetServiceIdentity(string id)
...
+            }));
+        }
+
+        async Task<string> VerifyServiceIdentityAuthChainState(string id, bool refreshCachedIdentity = false)
+        {
+            if (refreshCachedIdentity)
+            {
+                Events.RefreshingServiceIdentity(id);
+
+                var authChainTry = await this.serviceIdentityHierarchy.TryGetAuthChain(id);
```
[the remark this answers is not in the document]
ancaantochi: yes, I tested on same level and it works to refresh modules or leaf device

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
ancaantochi: I changed to return second if available, without passing the targetId. I think it should not validate here if target is first, just return the value if it's there.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
```diff
@@ -27,6 +27,18 @@ public static Option<string> GetAuthTarget(Option<string> authChain)
...
         }
 
+        public static Option<string> GetAuthParent(Option<string> authChain)
+        {
+            return authChain.Match(
+                chain =>
+                {
+                    string[] authChainIds = GetAuthChainIds(chain);
+                    // The auth target is second element after the target
+                    return authChainIds.Skip(1).FirstOption(id => true);
```
varunpuranik, line 37: Check if chain has >0 elements?

varunpuranik (approved):

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/AuthChainHelpers.cs
davilu: What is parent used for?

ancaantochi: it's not needed to check because skip will return empty collection and then parent is None.
it's used for making the onbehalfof call  to refresh the identity

PR #4542, opened 2021-03-04
Make comments in the config.yaml for provisioning and certificates clearer.

arsing:
- Grammar and capitalization ("ID", "DPS", etc).

- Consistent wording to refer to certificates and private keys.
  ("Path of the X certificate", "Path of the private key of the X certificate").

- Clarify that all certificate and private key paths are specified as file URIs,
  with examples.

- Split DPS provisioning sections into one section per attestation method,
  and dispense with the awkward "Required if... Optional if..." prologue
  on every field.

- Clarify what `always_reprovision_on_startup` and `dynamic_reprovisioning` do
  more clearly, especially their defaults.

- Consistently document `always_reprovision_on_startup` and
  `auto_generated_ca_lifetime_days` in all config files.


edgelet/contrib/config/linux/config.yaml
```diff
@@ -16,90 +16,114 @@
...
-#     dynamic_reprovisioning - Optional. A flag to opt-in to the dynamic re-provisioning 
-#                              feature. If enabled, IoT Edge on detecting a possible 
-#                              device re-provisioning event will shut down the daemon.
-#                              This is so that on the next daemon startup, the device is 
-#                              set up with the new provisioning information of the device in 
-#                              in IoT Hub.
-#                              For the external provisioning mode specifically, the daemon 
-#                              will notify the external provisioning endpoint about the
-#                              re-provisioning event before shutting down.
+# Miscellaenous settings
```
gordonwang0, line 74: Spelling, should be `Miscellaneous`.

Also should be fixed in the other two files.

arsing: Fixed.

gordonwang0 (approved):

micahl (approved):
We need to add similar descriptions to the config.toml.  Even if we had a guided walk-through or full UI for configuration, it would still be valuable to have our config.toml document the properties and their expected/accepted value.

varunpuranik (approved):

PR #4544, opened 2021-03-04
Add an IoT Edge focused doc on packaging changes and the refactoring

micahl:


doc/packaging.md
```diff
@@ -0,0 +1,334 @@
...
+
+All configuration is done through the `config.yaml` file. The daemon must be restarted via `systemctl restart` to apply configuration changes.
+</td>
+<td>
+
+```
+/etc/aziot/
+├── config.toml.edge.template
+└── edged/
+    └── config.toml
```
arsing, line 108: `config.toml.default`

doc/packaging.md
```diff
@@ -0,0 +1,334 @@
...
+
+iotedge config apply
+```
+
+After installing the `aziot-edge` package, copy the `/etc/aziot/config.toml.edge.template` to `/etc/aziot/config.toml` and edit the file to provide provisioning information.  Then run `iotedge config apply` to apply the configuration to the services. This performs initialization for both the IS+KS+CS+TPMS components installed by the `aziot-identity-service` package and the MR component installed by the `aziot-edge` package.
+
+## Updating from `iotedge` to `aziot-edge`
+
+```sh
+apt remove iotedge libiothsm-std
```
arsing, line 291: `apt remove` isn't necessary because the packages conflict, so we don't need them to do this. Then you also won't need the warning below about not `purge`ing the package.

doc/packaging.md
```diff
@@ -0,0 +1,334 @@
...
+
+```sh
+iotedge config import
+```
+
+Notes on behavior:
+
+- The device provisioning method is parsed from `/etc/iotedge/config.yaml` and translated into the provisioning information in `/etc/aziot/config.toml`.
+
+- User-provided certificates like device ID, device CA and trust bundle, and their corresponding private key files, will be added as preloaded keys and certs in `config.toml`. The files themselves will not be moved, because they are managed by the user rather than belonging in our services' directories.
```
arsing, line 314: Would be to good to mention either in this point or a new one that the imported device CA will now be referred to as the "Edge CA".

doc/packaging.md
```diff
@@ -0,0 +1,334 @@
...
+
+  This assumes that Microsoft's implementation of `libiothsm-std` is being used where the certs and keys are stored as files on disk. This is a reasonable assumption since we are not aware of any external Edge customers that have written their own `libiothsm-std` implementations which store keys and certs differently.
+
+- The master identity key and master encryption key are two symmetric keys dynamically generated by `iotedged` for internal use. They will not me be imported into the new services.
+
+  Note that this means Edge Agent, Edge Hub and other modules will not be able to decrypt any data when running against the new services that they previously encrypted using the workload API with the old service. This is not a problem for Edge Agent and Edge Hub, in the sense that they ignore the undecryptable data and start with empty storage, so they recover at the cost of data loss.
+
+- Certs like workload CA and module server certs that are created dynamically by `iotedged`, and can be regenerated trivially without any problems, will not be imported into the new services.
```
arsing, line 322: The reasons are different for them so it would be good to separate them:

```
- The workload CA cert is no longer used with aziot-edged and will not be imported.

- Module identity certs and module server certs will not be imported. Modules will get new certs signed directly by the Edge CA (nee the device CA) when they request them.
```

doc/packaging.md
```diff
@@ -0,0 +1,334 @@
...
+
+- The `import` does not detect and modify the access rules to the TPM. The `/etc/udev/rules.d/tpmaccess.rules` would still need to be manually updated to allow the `aziottpm` access to `tpm0` instead of `iotedge`.
```
arsing, line 324:
```diff
-The `import` does not...
+The `import` command does not...
```

arsing (changes requested):

doc/packaging.md
micahl: Fixed

doc/packaging.md
micahl: Split as recommended

doc/packaging.md
micahl: Added as a new point below this one

doc/packaging.md
micahl: Removed the remove and purged the purge

doc/packaging.md
micahl: Fixed

arsing (approved):

PR #4546, opened 2021-03-04
Fix: CloudConnection did not forward close() call to cloud proxy

vipeller:
The main change is in BrokeredCloudConnection.cs where it did not forward the call to CloudProxy. The rest of the change is only about to being able to log the Identity in case of failure.
The removed "TODO" about the subscription is not needed anymore as the proper place was in ConnectionHander.cs/RemoveConnectionAsync() to remove existing subscriptions.


edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/upstream/BrokeredCloudProxy.cs
```diff
@@ -14,22 +14,22 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
...
-        IIdentity identity;
         Action<string, CloudConnectionStatus> connectionStatusChangedHandler;
 
         AtomicBoolean isActive = new AtomicBoolean(true);
         AtomicBoolean twinNeedsSubscribe = new AtomicBoolean(true);
 
         public BrokeredCloudProxy(IIdentity identity, BrokeredCloudProxyDispatcher cloudProxyDispatcher, Action<string, CloudConnectionStatus> connectionStatusChangedHandler)
         {
-            this.identity = Preconditions.CheckNotNull(identity);
+            this.Identity = Preconditions.CheckNotNull(identity);
```
vadim-kovalyov, line 24: nit: why change this to prop?

vadim-kovalyov (approved):
LGTM

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/upstream/BrokeredCloudConnection.cs
```diff
@@ -4,26 +4,58 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
...
         {
-            // TODO, in order to tear down the connection higher level,
-            // it should unsubscribe from all topics
             this.IsActive = false;
-            return Task.FromResult(true);
+            var result = default(bool);
+
+            try
+            {
+                result = await this.CloudProxy.Match(
```
varunpuranik, line 35: Nit: suggest: .Map().GetOrElse(true)

varunpuranik, line 35: test?

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/upstream/BrokeredCloudConnection.cs
```diff
@@ -4,26 +4,58 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
...
     using Microsoft.Azure.Devices.Edge.Hub.Core.Cloud;
+    using Microsoft.Azure.Devices.Edge.Hub.Core.Identity;
     using Microsoft.Azure.Devices.Edge.Util;
+    using Microsoft.Extensions.Logging;
 
     public class BrokeredCloudConnection : ICloudConnection
     {
+        IIdentity identity;
+
         public BrokeredCloudConnection(BrokeredCloudProxy cloudProxy)
```
varunpuranik, line 15: Why is this not a ICloudProxy? We should avoid passing in actual types.. if possible.

edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter/upstream/BrokeredCloudConnection.cs
```diff
@@ -4,26 +4,58 @@ namespace Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter
...
         {
+            Preconditions.CheckNotNull(cloudProxy);
+
             this.IsActive = true;
             this.CloudProxy = Option.Some(cloudProxy as ICloudProxy);
+            this.identity = cloudProxy.Identity;
```
varunpuranik, line 21: This seems odd... can we just pass in the Identity separately, like we do for CloudConnection?

nyanzebra (commented):
Let's put at least 1 UT covering the change, if we need more thorough tests, we can add those in another pr if that is easier or not a high priority. Will approve after that :)

nyanzebra (approved):
LGTM 🚢

PR #4550, opened 2021-03-04
Update postrm to delete iotedge user on purge (#4533)

gordonwang0:
Updates postrm so that apt purge fully removes all files used by aziot-edge and removes iotedge user.

This only applies to Debian, as CentOS (rpm) doesn't have a purge command.


arsing:
The lockfile changes remind me - we should change all the git deps from `branch = "main"` to `branch = "release/1.2"` in this branch. But it doesn't have to be in this PR.

arsing (approved):

PR #4551, opened 2021-03-05
Add Helm Charts to images build.

darobs:
Helm charts can be pushed into container registries, and this will allow us to publish Helm charts in the same way we publish our Docker images.

The first step to releasing Helm charts through a container registry is to make them a part of our images build. This PR creates a build script and adds a Helm step to the build YAML for our images job.

This also contains a `moveHelmCharts.sh`, which parallels the `moveImages.sh` script, so that we can push built images into release repos.


builds/misc/images.yaml
```diff
@@ -393,6 +393,22 @@ jobs:
       - powershell: docker system prune -a -f
         displayName: Clean Machine
 
+################################################################################
+  - job: linux_helm_charts
+################################################################################
+    displayName: Helm Charts
+    pool:
+      vmImage: ubuntu-18.04
```
damonbarry, line 401: Check out the latest changes to images.yaml in `release/1.1`...we're using a 1ES-hosted pool instead of MS-hosted pool for agents. So you'll want to update this `pool` declaration to look like the ones in `release/1.1`. Something like:

```yaml
    pool:
      name: $(pool.linux.name)
      demands:
        - ImageOverride -equals agent-aziotedge-ubuntu-18.04-docker
```

damonbarry (approved):

PR #4556, opened 2021-03-06
Re-add dynamic provisioning support

massand:
- Replaced `auto_reprovision_on_startup` and `dynamic_reprovision` flags with one `auto_reprovisioning_mode` configuration (see config.toml docs for details)

- Changed import to only set `auto_reprovisioning_mode = AlwaysOnStartup` if `always_reprovision_on_startup = true`.

 - `iotedge config import` sets `auto_reprovisioning_mode = Dynamic` if old config had `auto_reprovision_on_startup = true` and `dynamic_reprovisioning = true` (added comment within `iotedge config import` logic for reason)

- aziot-edged calls `reprovision()` on startup if `auto_reprovision_mode = AlwaysOnStartup`

TODO:

- [X] Import existing config.yaml setting into super-config.toml


massand:
@arsing / @varunpuranik - My thoughts on restoring the `dynamic_reprovisioning` configuration under the root of the configuration: Previously this setting sat under `[provisioning]` in config.yaml. However, this is technically Edge behavior that happened to call into `reprovision()` when provisioning was part of iotedged. Now, it could fall under `[agent]` (if you consider it the only trigger) or under root  (if we foresee other factors that contribute to reprovisioning). However, I am not sure if there is a nested Edge scenario where this request could originate from Edge Hub in the future, but I maybe over-thinking it. WDYT? 

I wanted to close on this before implementing the rest of the `iotedge config` changes.

varunpuranik:
> @arsing / @varunpuranik - My thoughts on restoring the `dynamic_reprovisioning` configuration under the root of the configuration: Previously this setting sat under `[provisioning]` in config.yaml. However, this is technically Edge behavior that happened to call into `reprovision()` when provisioning was part of iotedged. Now, it could fall under `[agent]` (if you consider it the only trigger) or under root (if we foresee other factors that contribute to reprovisioning). However, I am not sure if there is a nested Edge scenario where this request could originate from Edge Hub in the future, but I maybe over-thinking it. WDYT?
> 
> I wanted to close on this before implementing the rest of the `iotedge config` changes.

I think this setting belongs at the top level, and we can put it in the provisioning section (under the comments, and not under [provisioning]). The fact that we use Agent for this is implementation detail, and I don't think we should put it under [agent]. For nested scenarios, it is possible that the information for this flows from EdgeHub, but I don't think we want EdgeHub to actively participate in determining this.
Does that make sense?

edgelet/edgelet-docker/src/runtime.rs
```diff
@@ -736,6 +736,7 @@ impl ModuleRuntime for DockerModuleRuntime {
         // be emitted from Identity Service
         let provisioning = ProvisioningInfo {
             r#type: "ProvisioningType".into(),
+            //TODO: read from settings
```
arsing, line 739: Since it's covered by the comment on line 735 anyway, and especially since line 741 is also already returning a fake value, I guess this comment isn't necessary.

arsing (approved):

arsing (approved):

massand:
> @massand - have you verified that the reprovisioning scenario works now with this change?

Yes - the reprovisioning has been tested, I just added another fix for a cleaner shutdown which I missed previously.

micahl (approved):
The consolidated auto_reprovisioning_mode looks good to me.

massand:
@arsing - there is a cargo version bump on i-i-s and one more change in import needed make after this goes in Azure/iot-identity-service#167. Please stand by your review. 
cc @gordonwang0

edgelet/edgelet-docker/src/runtime.rs
massand: Fixed

edgelet/iotedge/src/config/import/mod.rs
```diff
@@ -215,19 +219,22 @@ fn execute_inner(old_config_file: &Path) -> Result<Vec<u8>, std::borrow::Cow<'st
                         },
                     ),
                 always_reprovision_on_startup,
```
arsing, line 221: We had talked about differentiating between `true` and unset, so that unset maps to `false` rather than `true`.

ie you need to change the `old_config` type to be an `Option<bool>`

arsing (changes requested):

edgelet/iotedge/src/config/import/mod.rs
massand: Fixed

arsing (approved):

PR #4559, opened 2021-03-08
Merge Broker and Bridge settings

vadim-kovalyov:
In this PR I'm trying to consolidate Broker and Bridge settings together in `mqtt-edgehub` crate. Even if the Bridge settings structs are defined in `mqtt-bridge`, we need to manage "reading" of the settings in `mqtt-edgehub`, so we can control how we read (file vs env vars) and what we read (which edgehub-specific env vars can override settings).

Changes in this PR:
- Include Bridge settings into Edgehub settings.
- Move and fix tests.
- Restrict bridge settings to the following variables:
  - `UsePersistentStorage` (existing var) - defines whether to use disk or in memory storage for the bridge.
  - `MqttBridge__StorageMaxFileSize` - defines maximum size (in bytes) of the MQTT Bridge storage. This storage is separate from edgehub messages / twins storage. 
  - `MqttBridge__StorageMaxMessages` - defines max in-memory queued messages in the MQTT Bridge. 
  - `MqttBridge__StorageFlushOptions` - for now only two: "AfterEachWrite" and "Off". **There is a bug in reading "AfterXWrites(u64)" variants from env var, I'd like to fix that in a separate PR**
  - `StorageFolder` - path where to write storage file(s)
- Make sure `$upstream` bridge settings can still be read from env vars


mqtt/mqtt-edgehub/src/settings.rs
```diff
@@ -75,8 +76,83 @@ impl Source for BrokerEnvironment {
...
+            result.insert("bridge.module_id".into(), val);
+        }
+        if let Ok(val) = host_env.get::<Value>("iotedge_modulegenerationid") {
+            result.insert("bridge.generation_id".into(), val);
+        }
+        if let Ok(val) = host_env.get::<Value>("iotedge_workloaduri") {
+            result.insert("bridge.workload_uri".into(), val);
+        }
+
+        // storage rign buffer
```
nyanzebra, line 139: minor: typo

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -284,6 +268,7 @@ pub enum StorageSettings {
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 pub struct MemorySettings {
+    #[serde(deserialize_with = "deserialize_nonzerouusize")]
     max_size: NonZeroUsize,
```
dmolokanov, line 272: Let's make it HumanSize instead?

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -299,6 +284,7 @@ impl MemorySettings {
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 pub struct RingBufferSettings {
+    #[serde(deserialize_with = "deserialize_nonzerou64")]
     max_file_size: NonZeroU64,
```
dmolokanov, line 288: Let's make it HumanSize instead?

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -326,181 +312,53 @@ impl RingBufferSettings {
...
-    fn from_file_reads_storage_settings_without_explicit_storage() {
-        let settings = BridgeSettings::from_file("tests/config.json").unwrap();
-        let storage_settings = settings.storage();
-        // Should exist from default.json.
-        assert_matches!(storage_settings, StorageSettings::RingBuffer(_));
-        if let StorageSettings::RingBuffer(rb) = storage_settings {
-            assert_eq!(rb.max_file_size(), NonZeroU64::new(33_554_432).unwrap());
-            assert_eq!(*rb.directory(), PathBuf::from("/tmp/mqttd/"));
-            assert_eq!(*rb.flush_options(), FlushOptions::AfterEachWrite);
+fn deserialize_nonzerou64<'de, D>(deserializer: D) -> Result<NonZeroU64, D::Error>
```
dmolokanov, line 315: serde doesn't support NonZero types deserializing?

mqtt/mqtt-edgehub/src/settings.rs
```diff
@@ -75,8 +76,83 @@ impl Source for BrokerEnvironment {
...
+        }
+        if let Ok(val) = host_env.get::<Value>("storageflushoptions") {
+            result.insert("bridge.storage.flush_options".into(), val);
+        }
+        if let Ok(val) = host_env.get::<Value>("storagefolder") {
+            result.insert("bridge.storage.directory".into(), val);
+        }
+
+        // storage in memory
+        if let Ok(val) = host_env.get::<Value>("storagesize") {
```
dmolokanov, line 154:
```suggestion
        if let Ok(val) = host_env.get::<Value>("storagememorysize") {
```
As a user, I'm confused with MaxStorageSize and MaxStorageFileSize. Why not use only one value to rule them all? :)
`MaxStorageMemorySize` and `MaxStorageFileSize` look cleaner

mqtt/mqtt-edgehub/Cargo.toml
```diff
@@ -29,6 +29,7 @@ mqtt3 = { path = "../mqtt3", features = ["serde1"] }
 mqtt-bridge = { path = "../mqtt-bridge" }
 mqtt-broker = { path = "../mqtt-broker" }
 mqtt-policy = { path = "../mqtt-policy" }
+mqtt-util = { path = "../mqtt-util" }
```
dmolokanov, line 32: I guess it is not strictly required?

dmolokanov (commented):
GH shows some empty files for me `./mqtt/mqtt-edgehub/tests/settings/settings.rs` maybe there are some more.

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: While we are here, are there any other forms of number format that we want to support? This will handle string and int format, not sure there are others? Like 2Kb or something?

mqtt/mqtt-bridge/src/settings.rs
nyanzebra: The point of these is to allow for multiple formats of the value (string/int) and then take the value and make it NonZeroX. I found this is nice for testing with env::set where things are strings and also gave some flexibility into how one wants to organize their values.

mqtt/mqtt-edgehub/src/settings.rs
nyanzebra: At least when I made the changes for settings on memory and ringbuffer, I wanted it to be clear that the storage values have slightly different side effects, i.e. memory vs disk. I think combining into one value might look nicer and should be hard to destroy in Mordor 😸 Let's just make sure we call out somewhere in the documentation that depending on storage type these can use different resources?

vadim-kovalyov: @dmolokanov, what you say makes sense, but one thing to keep in mind, is that those two variables are actually different units. Storage file size is in bytes (or Humansize), where memory size is in number of publications (not bytes). So I believe combining them will be confusing.

mqtt/mqtt-bridge/src/settings.rs
vadim-kovalyov: It looks like `HumanSize` will not work as it is right now, since it is backed by `usize`, where we need `u64`. We can add another version of `HumanSize(u64)` but I would rather do it in a separate PR.

mqtt/mqtt-edgehub/Cargo.toml
vadim-kovalyov: Some structs are now used in mqtt-edgehub (parts of BridgeSettings)

mqtt/mqtt-bridge/config/default.json
```diff
@@ -1,16 +0,0 @@
-{
```
vadim-kovalyov, line 1: this file is no longer needed. Default values are defined in `mqtt-generic` or `mqtt-edgehub` crates, and they can be different for "generic" or "edgehub" features.

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -783,4 +789,51 @@ mod tests {
             panic!("Should not reach here");
         }
     }
+
+    fn test_bridge_settings() -> BridgeSettings {
```
vadim-kovalyov, line 793: since we moved "how we read settings" logic from `mqtt-bridge` to `mqtt-edgehub`, here we just create test data instead of reading it from `tests/config.json`

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -1,78 +1,36 @@
...
-        let mut config = Config::new();
-
-        config.merge(File::from_str(DEFAULTS, FileFormat::Json))?;
-        config.merge(File::from(path.as_ref()))?;
-        config.merge(Environment::new())?;
-
-        config.try_into()
-    }
-
-    pub fn from_upstream_details(
```
vadim-kovalyov, line 52: with a little refactoring and new ctor functions this has been moved to tests crate, since it is used only for tests.

mqtt/mqtt-bridge/src/settings.rs
```diff
@@ -326,181 +289,53 @@ impl RingBufferSettings {
     }
 }
 
-#[cfg(test)]
-mod tests {
```
vadim-kovalyov, line 330: all these test cases moved to `mqtt-edgehub/src/settings.rs`, not as is, but with similar coverage.

mqtt/mqtt-util/src/client_io.rs
```diff
@@ -309,41 +309,30 @@ pub enum Credentials {
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 pub struct CredentialProviderSettings {
-    #[serde(rename = "iotedge_iothubhostname")]
```
vadim-kovalyov, line 312: this annotation no longer needed, since we do env var mappings in `BridgeEnvironment`

doc/EnvironmentVariables.md
```diff
@@ -84,7 +84,10 @@
...
+| MqttBroker__MaxQueuedMessages           | Maximum number of messages that can be in the outgoing queue for a single session                                                                  | int32                                                                 | 1000                          |
+| MqttBroker__MaxQueuedBytes              | Maximum number of bytes that can be in the outgoing queue for a single session                                                                     | int32 (arm), int64 (amd64)                                            | 0 (no limit)                  |
+| MqttBroker__WhenFull                    | Defines what to do in case either max queued messages or max queued bytes is reached                                                               | drop_new, drop_old                                                    | drop_new                      |
+| MqttBridge__StorageMaxFileSize          | If `UsePersistentStorage` is set, defines maximum size (in bytes) of the MQTT Bridge storage. This storage is separate from edgehub messages / twins storage.   | int64                                                    | 33554432 (32mb)               |
```
varunpuranik, line 91: We need to think through how this will work when we have multiple files for different priority queues. 
So for now, I suggest we rename this to StorageMaxDiskSize or StorageMaxDiskUsage. 
Or alternatively we can remove it altogether till we figure out how to support it for multiple priorities (in case we need to implement it in a different way).

doc/EnvironmentVariables.md
```diff
@@ -84,7 +84,10 @@
...
+| MqttBridge__StorageFlushOptions         | If `UsePersistentStorage` is set, defines MQTT Bridge storage flush options.                                                                                    | AfterEachWrite, Off                                      | AfterEachWrite                |
+| MqttBridge__StorageMaxMessages          | If `UsePersistentStorage` is not set, defines max in-memory queued messages in the MQTT Bridge.                                                                 | int32 (arm), int64 (amd64)                               | 1000                          |
```
varunpuranik, line 93: Same for this.

vadim-kovalyov: @varunpuranik, I have renamed the vars. I used the name `MqttBridge__StorageMaxMemoryMessages` without words "Usage" or "Size" in it, for memory option, to emphasize that this is __a number of messages__, and not bytes.

doc/EnvironmentVariables.md
```diff
@@ -84,7 +84,10 @@
...
+| MqttBroker__WhenFull                    | Defines what to do in case either max queued messages or max queued bytes is reached                                                               | drop_new, drop_old                                                    | drop_new                      |
+| MqttBridge__StorageMaxDiskUsage         | If `UsePersistentStorage` is set, defines maximum size (in bytes) of the MQTT Bridge storage. This storage is separate from edgehub messages / twins storage.   | int64                                                    | 33554432 (32mb)               |
+| MqttBridge__StorageFlushOptions         | If `UsePersistentStorage` is set, defines MQTT Bridge storage flush options.                                                                                    | AfterEachWrite, Off                                      | AfterEachWrite                |
+| MqttBridge__StorageMaxMemoryMessages    | If `UsePersistentStorage` is not set, defines max in-memory queued messages in the MQTT Bridge.                                                                 | int32 (arm), int64 (amd64)                               | 1000                          |
```
varunpuranik, line 93: Actually thinking about this a bit more - one use case we have seen is limiting the memory used by these messages when using in memory store.

varunpuranik: Does it make sense to make the max usage env vars storage type agnostic (in-memory or disk)?

vadim-kovalyov: @varunpuranik, it does make sense, but the issue is that those two vars are different units. Disk Usage is in bytes, where in-memory usage is in number of messages. This is how in-memory store was implemented. We can change that, but I'd rather do it in a separate PR.

doc/EnvironmentVariables.md
```diff
@@ -84,7 +84,10 @@
...
-| MqttBroker__MaxQueuedBytes              | Maximum number of bytes that can be in the outgoing queue for a single session                                                                     | int32 (arm), int64 (amd64)                                            | 0 (no limit)                         |
-| MqttBroker__WhenFull                    | Defines what to do in case either max queued messages or max queued bytes is reached                                                               | drop_new, drop_old                                                    | drop_new                       |
\ No newline at end of file
+| MqttBroker__MaxInflightMessages         | Maximum number of QoS 1 messages that can be in the process of being sent simultaneously for a single session                                      | int32                                                                 | 16                            |
```
nyanzebra, line 87: Can we do smaller than int32? Seems like it allows for a very large possible set of messages in flight :)

vadim-kovalyov: We probably should. But let's do that in a separate PR. I haven't touched this setting in the current PR (just whitespace change in the docs)

nyanzebra (approved):

varunpuranik (approved):

dmolokanov (approved):
:shipit:

PR #4571, opened 2021-03-09
Add EFLOW Job to Long Haul Test Pipeline

josephknierman:
Adds a new job for the long haul test pipeline to run on a eflow agent.


smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
```diff
@@ -162,6 +162,11 @@ public Task Install()
                 commandName = "apt-get";
                 commandArgs = $"--yes install {PackageName}";
             }
+            else if (this.archivePath.Contains(".rpm") && !this.archivePath.Contains(".dep"))
+            {
+                commandName = "sudo";
+                commandArgs = $"rpm --force -U {this.archivePath}";
```
damonbarry, line 168: This command shouldn't require sudo...the tests themselves run as sudo. Notice the commands in the other branches of this "if" block (apt-get and dpkg) don't require sudo...

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
```diff
@@ -253,6 +258,13 @@ public async Task Configure(DeviceProvisioningMethod method, Option<string> agen
                 doc.ReplaceOrAdd("listen.management_uri", uris.ListenManagement);
                 doc.ReplaceOrAdd("listen.workload_uri", uris.ListenWorkload);
             }
+            else if (hostname.Contains("-EFLOW-"))
```
damonbarry, line 261: So we require that the agent's hostname contains "-EFLOW-"? That seems sketchy...is there a better way to differentiate eflow from others when building config.yaml?

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
josephknierman: Changed command name to be rpm instead of sudo

smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
josephknierman: Added section that checks for a specific string in /etc/os-subrelease to determine if the system is EFLOW.

damonbarry (approved):

PR #4586, opened 2021-03-10
Yophilav/fix pipeline err for1.1.1 release 1.1

yophilav:
- Fix Connectivity test pipeline's IS downgrade conflict.
Note: We don't need to implement this fix. Just have a test agent that strictly running 1.1 (agent-group=2). And a separate group running 1.2 (agent-group=1).

### DIFF
--- a/smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
+++ b/smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
@@ -151,26 +151,9 @@ namespace IotEdgeQuickstart.Details
 
             Console.WriteLine($"Installing debian package '{PackageName}' from {this.archivePath ?? "apt"}");
 
-            string commandName;
-            string commandArgs;
-
-            // Use apt-get if a package name is given, or dpkg if a package file is given.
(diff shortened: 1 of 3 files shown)


smoke/IotEdgeQuickstart/details/IotedgedLinux.cs
```diff
@@ -157,15 +157,23 @@ public Task Install()
             // Use apt-get if a package name is given, or dpkg if a package file is given.
             // We'd like to use apt-get for both cases, but older versions of apt-get (e.g.,
             // in Raspbian) can't accept a package file.
-            if (string.IsNullOrEmpty(this.archivePath))
+            if (System.Environment.Is64BitOperatingSystem)
```
damonbarry, line 160: Why are you checking for 64-bit OS here? And can you add a code comment to explain the rationale?

Also, this could be written more succinctly like this:
```csharp
if (System.Environment.Is64BitOperatingSystem || string.IsNullOrEmpty(this.archivePath))
{
    commandName = "apt-get";
    commandArgs = $"--yes install {PackageName}";
}
else
{
    commandName = "dpkg";
    commandArgs = $"--force-confnew -i {this.archivePath}";
}
```

PR #4587, opened 2021-03-11
Set RPM dependency

gordonwang0:


edgelet/contrib/centos/aziot-edge.spec
```diff
@@ -10,14 +10,14 @@ Name:           aziot-edge
...
-License:        Proprietary
+License:        MIT
 Summary:        Azure IoT Edge Module Runtime
 URL:            https://github.com/azure/iotedge
 
 %{?systemd_requires}
 BuildRequires:  systemd
 Requires(pre):  shadow-utils
-Requires:       aziot-identity-service >= @version@-@release@
+Requires:       aziot-identity-service >= 1.2.0~rc4-1
```
arsing, line 20: Make it `=` just in case there's some version shenanigans with the final release (`1.2.0~rc4-1` sorts higher than `1.2.0-1` or something)

gordonwang0: Sure, although I don't plan to merge this PR. Just made it so I could run the pipeline to generate test packages.

PR #4589, opened 2021-03-11
add guid to module and device name

ancaantochi:
Invoke dm tests were failing in CI, same deviceid and moduleid was used and if multiple pipelines are running could delete a module that was used by another run.


yophilav (approved):
Thank you!

PR #4590, opened 2021-03-11
Add optional proxy argument to iotedge

massand:
The proxy-uri parameter can be passed in via command line argument. If that isn't specified, the settings (if valid) will be searched for Edge Agent's https_proxy environment variable.
Either way, the value will be passed along to aziotctl, where the connection check will occur using the proxy URI.

Note that this depends on https://github.com/Azure/iot-identity-service/pull/158

### DIFF
--- a/edgelet/iotedge/src/check/mod.rs
+++ b/edgelet/iotedge/src/check/mod.rs
@@ -45,6 +45,7 @@ pub struct Check {
 
     // These optional fields are populated by the checks
     iothub_hostname: Option<String>, // populated by `aziot check`
+    proxy_uri: Option<String>,       // populated by `aziot check`
     settings: Option<Settings>,
     docker_host_arg: Option<String>,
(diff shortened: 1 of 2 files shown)


edgelet/iotedge/src/check/mod.rs
```diff
@@ -45,6 +45,7 @@ pub struct Check {
 
     // These optional fields are populated by the checks
     iothub_hostname: Option<String>, // populated by `aziot check`
+    proxy_uri: Option<String>,       // populated by `aziot check`
```
gordonwang0, line 48: Should this be a Uri type, such as `hyper::Uri`?

massand: The string is passed verbatim to aziotctl and never used in iotedge. So, the type could be serialized upfront here.

gordonwang0 (approved):

PR #4572, opened 2021-03-09
Nested Longhaul: Fix upstream protocol and test name

and-rewsmith:
Fixes:
- Upstream protocol is incorrectly configured. It assigns the correct upstream protocol, but always reports amqp to log analytics. I fixed this and standardized the naming for upstream protocol across our yamls as ```upstream.protocol```.
- The addition of test name I made has a problem, as I thought we had two nested longhaul pipelines. Turns out these are duplicates and we have one nested longhaul pipeline that runs two longhaul tests. One for amqp upstream and one for mqtt upstream with broker. Therefore we need to plumb this information down not through pipeline vars, but hardcoded values in the nested longhaul yaml similar to upstream protocol.

PR #4591, opened 2021-03-11
Nested Pipelines: Fix wrong casing for upstream protocol

and-rewsmith:
This PR fixes two things, but really they are the same issue:
- Existing bug in ISA95 smoke tests
- Bug I introduced yesterday in #4572 

The problem for both is that upstream protocol is not passed, or passed incorrectly.

This problem was masked by my testing, so I also made a change to fail the pipeline run if an error occurs making the identity. I do this with ```set -e```.
https://stackoverflow.com/questions/1378274/in-a-bash-script-how-can-i-exit-the-entire-script-if-a-certain-condition-occurs

### DIFF
--- a/builds/e2e/isa-95-smoke-test.yaml
+++ b/builds/e2e/isa-95-smoke-test.yaml
@@ -27,6 +27,7 @@ stages:
       identityServiceArtifactName: packages_ubuntu-18.04_amd64
       identityServicePackageFilter: aziot-identity-service_*_amd64.deb
       level: '5'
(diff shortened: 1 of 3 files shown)


and-rewsmith:
Validating changes.
- Nested longhaul: https://msazure.visualstudio.com/One/_build/results?buildId=40278992&view=results
- ISA 95: https://msazure.visualstudio.com/One/_build/results?buildId=40279011&view=results

huguesBouvier (approved):

PR #4593, opened 2021-03-12
Allow null/empty settings in bridge config

dmolokanov:
When using faulty bridge settings, edgehub cannot parse the twin and therefore does not send the correct signals to open its ports. Ideally edgehub should recognize there are faulty bridge settings, log an error, and open the ports.
```json
"bridges": [
  {
    "endpoint": "$upstream",
    "settings": []
  }
]
```


nyanzebra:
Do we need any additional testing?

PR #4594, opened 2021-03-12
Fix: don't unsubscribe when there is no subscription registered

vipeller:
There was an optimization on ConnectionManager so it did not do any action for subscription/unsubscription if the state did not change, to avoid excess communication to iothub. The status of a given subscription (e.g. for C2D messages) can be represented by three states: true, false or missing. Missing means that the subscription was never set, so the dictionary that stores the state has no information about it. When this happens that means that nothing has set the subscription yet, which means that no subscription exists to that topic. In this case when that subscription gets removed, that does not mean status change.

The original implementation handled this as a status change and executed a remove operation. This should not have caused any problems, however a bug in the C# SDK breaks something and it blocks subsequent subscriptions.

This fix makes the optimization work right, and it mitigates the occurrence of bug in the SDK till they fix it

### DIFF
--- a/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
+++ b/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.Core/ConnectionManager.cs
@@ -122,7 +122,7 @@ namespace Microsoft.Azure.Devices.Edge.Hub.Core
             device.DeviceConnection.Filter(d => d.IsActive)
                 .ForEach(d =>
                 {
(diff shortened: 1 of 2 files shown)


nyanzebra (approved):
LGTM, is there some item on the C# SDK team (public bug in changelog or something) that mentions this kind of bug? As it will be good to be aware of and avoid similar problems :)

vipeller:
this is the issue report about the bug this PR tries to avoid:
https://github.com/Azure/azure-iot-sdk-csharp/issues/1822

PR #4598, opened 2021-03-15
Add `iotedge config mp` to create a super-config with a manual-provisioning connection string.

arsing:
Other changes:

- Define `IOTEDGE_{CONNECT,LISTEN}_{MANAGEMENT,WORKLOAD}_URI` env vars
  at compile time so that code can know these target-OS-specific values.
  Previously the `IOTEDGE_HOST` env var was used just for the value of
  `IOTEDGE_CONNECT_MANAGEMENT_URI`.

- Implement `Default` for `edgelet_core::{Connect,Listen,MobyRuntime}`
  and make them not-required in the super-config template. Also make the `agent`
  property optional and hard-code its default in the code.


edgelet/Makefile
```diff
@@ -78,6 +86,10 @@ distclean:
 	rm -rf .cargo
 	rm -rf $(vendordir)
 
+deb: CONNECT_MANAGEMENT_URI=unix:///var/run/iotedge/mgmt.sock
+deb: CONNECT_WORKLOAD_URI=unix:///var/run/iotedge/mgmt.sock
```
gordonwang0, line 90: This should be `workload.sock`.

edgelet/edgelet-core/src/settings.rs
```diff
@@ -27,6 +27,34 @@ impl Connect {
...
+    #[allow(clippy::option_if_let_else)]
+    fn default() -> Self {
+        const DEFAULT_MANAGEMENT_URI: &str =
+            if let Some(value) = option_env!("IOTEDGE_CONNECT_MANAGEMENT_URI") {
+                value
+            } else {
+                "unix:///var/run/iotedge/mgmt.sock"
+            };
+        const DEFAULT_WORKLOAD_URI: &str =
+            if let Some(value) = option_env!("IOTEDGE_CONNECT_MANAGEMENT_URI") {
```
gordonwang0, line 41: This should be `IOTEDGE_CONNECT_WORKLOAD_URI`.

edgelet/edgelet-core/src/settings.rs
```diff
@@ -49,6 +77,35 @@ impl Listen {
...
+    #[allow(clippy::option_if_let_else)]
+    fn default() -> Self {
+        const DEFAULT_MANAGEMENT_URI: &str =
+            if let Some(value) = option_env!("IOTEDGE_CONNECT_MANAGEMENT_URI") {
+                value
+            } else {
+                "fd://aziot-edged.mgmt.socket"
+            };
+        const DEFAULT_WORKLOAD_URI: &str =
+            if let Some(value) = option_env!("IOTEDGE_CONNECT_MANAGEMENT_URI") {
```
gordonwang0, line 91: This should be `IOTEDGE_CONNECT_WORKLOAD_URI`.

gordonwang0 (changes requested):

edgelet/Makefile
arsing: Fixed.

edgelet/edgelet-core/src/settings.rs
arsing: Fixed.

edgelet/edgelet-core/src/settings.rs
arsing: Fixed.

edgelet/iotedge/src/system.rs
```diff
@@ -13,27 +13,30 @@ use crate::error::{Error, ErrorKind};
...
-        // CentOS 7 doesn't use socket activation.
-        let sockets: &'static [&'static str] = option_env!("IOTEDGE_HOST").map_or(
-            &["aziot-edged.mgmt.socket", "aziot-edged.workload.socket"],
-            |_host| &[],
-        );
+        // If IOTEDGE_LISTEN_MANAGEMENT_URI isn't set at compile-time, assume socket activation is being used.
+        //
+        // This doesn't matter for release builds since those always have IOTEDGE_LISTEN_MANAGEMENT_URI set.
+        // It's only useful for developers' builds.
+        let uses_socket_activation = option_env!("IOTEDGE_LISTEN_MANAGEMENT_URI").map_or(true, |value| value.starts_with("fd://"));
```
massand, line 20: It is best to be explicit here and check for cfg_attr(debug_assertions)

arsing: It's not about debug-vs-not. It's about `make rpm` / `make deb` vs not.

massand: I get it now. The comment threw me off as it read "release builds" (I suggest "package build") and "developers' build" (I suggest "non-RPM/DEB builds")

arsing: Oh right. Fixed.

massand (approved):

gordonwang0 (approved):

gordonwang0 (approved):

PR #4602, opened 2021-03-15
[mqtt] handle RPC sub requests

dmolokanov:
For some reasons when EdgeHub starts up it make several concurrent attempts to subscribe to $iothub topics. It causes first attempt to succeed and the second one to fail with a timeout.

The changes to treat RPC subsctioption commands as individual even for the same topic.


vipeller (approved):

mqtt/mqtt-bridge/src/upstream/rpc/mod.rs
```diff
@@ -110,17 +112,56 @@ impl Display for RpcCommand {
 /// response comes back as `mqtt3::Event` type which handled with the event
 /// handler.
 #[derive(Debug, Clone, Default)]
-pub struct RpcSubscriptions(Arc<Mutex<HashMap<String, CommandId>>>);
+pub struct RpcSubscriptions(Arc<Mutex<HashMap<String, VecDeque<CommandId>>>>);
```
nyanzebra, line 115: How many ids and what all operations do you expect here? Mainly ask if we can unify on Vec or VecDeque so we could pass a reference to the collection of CommandIds. Not a requirement, but might be nice :)

mqtt/mqtt-bridge/src/upstream/rpc/mod.rs
```diff
@@ -110,17 +112,56 @@ impl Display for RpcCommand {
...
     /// Removes topic filter to command identifier mapping and returns
     /// `CommandId` if exists.
     pub fn remove(&self, topic_filter: &str) -> Option<CommandId> {
-        self.0.lock().remove(topic_filter)
+        let mut inner = self.0.lock();
+
+        inner
+            .remove_entry(topic_filter)
+            .and_then(|(topic, mut existing)| {
+                existing.pop_front().map(|id| {
```
nyanzebra, line 139: Is this part necessary? Why do we need to re-insert into the hashmap? Are we not able to borrow the entry itself and then pop?

mqtt/mqtt-bridge/src/upstream/rpc/mod.rs
dmolokanov: I should not be the case at all! But for a number of reasons we cannot rely on EdgeHub to guaranty unique RPC request per topic.

I guess we cannot move out vec of CommandId out of shared collection while not holding a mutex. so making a copy of a small Vec may be a good enough.

mqtt/mqtt-bridge/src/upstream/rpc/mod.rs
dmolokanov: there is a bug here :) 

answering your question: the intention here is to get a list of commands and remove collection from the hashmap if empty. my expectation that there is no more than 2 items in the collection (ideally only one). so I guessed I can remove entry and insert if needed.

mqtt/mqtt-bridge/src/upstream/rpc/mod.rs
nyanzebra: Sounds good to me, just want to avoid big copies 😸

mqtt/mqtt-bridge/src/upstream/rpc/mod.rs
nyanzebra: Okay okay got it 😺

mqtt/mqtt-bridge/tests/common.rs
```diff
@@ -72,29 +72,30 @@ where
...
 pub async fn setup_bridge_controller(
+    device_id: &str,
     local_address: String,
     upstream_address: String,
     subs: Vec<Direction>,
     storage_dir_override: &PathBuf,
 ) -> (BridgeControllerHandle, JoinHandle<()>) {
     let credentials = Credentials::PlainText(AuthenticationSettings::new(
-        "bridge".into(),
+        device_id.into(),
```
nyanzebra, line 82: Yeah.. these tests can conflict with one another...

nyanzebra (approved):

PR #4603, opened 2021-03-15
Fix: throw transient error when edgeHubCore is disconnected from the broker

vipeller:
so messages will be resent instead of dropping them


edge-hub/core/test/Microsoft.Azure.Devices.Edge.Hub.MqttBrokerAdapter.Test/ConnectionEventTest.cs
```diff
@@ -121,6 +122,43 @@ void Spy(RpcPacket packet)
             }
         }
 
+        [Fact]
+        public async Task WhenNoConnectionMessagesDontGetDropped()
```
ancaantochi, line 126: I think the test should just check that it throws IotHubException when it is disconnected. 
There should be an integration test in CloudEndpointTest to check that the message is not dropped.

ancaantochi (approved):

PR #4604, opened 2021-03-15
Longhaul: Make log rotation less restrictive on amd64

and-rewsmith:
Current log rotation settings for longhaul are configured for pis. We need to have ample logs, especially on amd64 where we do most of our testing. I have added logic to the trcE2Escript.sh to set the log size differently depending on platform.


test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -156,7 +156,7 @@ internal static async Task UploadLogsAsync(string iotHubConnectionString, Uri bl
             ServiceClient serviceClient = ServiceClient.CreateFromConnectionString(iotHubConnectionString);
             CloudToDeviceMethod uploadLogRequest =
                 new CloudToDeviceMethod("UploadModuleLogs")
-                    .SetPayloadJson($"{{ \"schemaVersion\": \"1.0\", \"sasUrl\": \"{blobContainerWriteUri.AbsoluteUri}\", \"items\": [{{ \"id\": \".*\", \"filter\": {{}} }}], \"encoding\": \"gzip\" }}");
```
and-rewsmith, line 159: This change actually allows the logs to get sent up and readable. But having them not compressed might cause issues.

scripts/linux/trcE2ETest.sh
```diff
@@ -866,10 +867,12 @@ quickstart_working_folder="$working_folder/quickstart"
 image_architecture_label=$(get_image_architecture_label)
 optimize_for_performance=true
 log_upload_enabled=true
+log_rotation_max_file="50"
```
and-rewsmith, line 870: This is my justification for this number:

~15 modules * 10MB per file * 50 files = 7.5 GB total logs

huguesBouvier (approved):

huguesBouvier (approved):

PR #4611, opened 2021-03-16
Remove case sensitivity in hostname check

huguesBouvier:
Series of fix + improvement on iotedge check.
PR following: https://github.com/Azure/iotedge/pull/4608

Removing case sensitivity when checking hostname.
It created issue, like being unable to use fqdn if your hostname has a capital letter (if you use azure VMs).

### DIFF
--- a/edgelet/iotedge/src/check/checks/hostname.rs
+++ b/edgelet/iotedge/src/check/checks/hostname.rs
@@ -86,8 +86,12 @@ impl Hostname {
         //
         // Instead, we punt on this check and assume that everything's fine if config_hostname is identical to the device hostname,
         // or starts with it.
-        if config_hostname != machine_hostname
-            && !config_hostname.starts_with(&format!("{}.", machine_hostname))
+        //
+        // Azure FQDN don't support capital letter so we lower case the hostname before doing the check.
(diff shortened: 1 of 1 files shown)


massand:
The hostname is needed to generate server certificates with the right DNS entries needed for mutual TLS auth by leaf devices. So, I believe this requirement cannot be changed in 1.2 GA. 

The capitalization issues with hostname in checks can be resolved with `to_lowercase`

massand:
Please change the PR description to match the change - i.e. fixed case sensitivity. Thanks.

edgelet/iotedge/src/check/checks/hostname.rs
```diff
@@ -86,8 +86,12 @@ impl Hostname {
         //
         // Instead, we punt on this check and assume that everything's fine if config_hostname is identical to the device hostname,
         // or starts with it.
+        //
+        // Azure FQDN don't support capital letter so we lower case the hostname before doing the check.
         if config_hostname != machine_hostname
```
massand, line 91: to_lowercase() these as well.

huguesBouvier: Ok, though that does't change the logic.

massand (approved):

PR #4618, opened 2021-03-17
Shorten all k8s watch timeouts.

darobs:
Watches on certain clusters seemed to be stopping and not recovering as we expected.  Investigating the code, I realized I had missed a few cases where a failure could occur, and so I fixed the code to ensure all failures were handled.  

YET, the problem still occurred.  We learned that the C# K8s client seems to have trouble with longer timeouts for List/Watch on certain clusters.  So, we shortened the timeout to below the value that seems to be where the problem lies.
Now, instead of using a default timeout, we set all watch timeouts to 3 minutes.


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/KubernetesEnvironmentOperator.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes
 
     public class KubernetesEnvironmentOperator : IKubernetesEnvironmentOperator
     {
+        const int TimeoutSeconds = 180;
```
pmzara, line 15: Why this number? Any chance to use some parameter or config value?

darobs: 3 minutes is consistently short enough to avoid the bug in the C# k8s SDK, but allowing it to be configurable is a really good idea.

pmzara (approved):

PR #4623, opened 2021-03-18
TwinTester: Give more time for updates to come in

and-rewsmith:
Even in single node longhaul, sometimes updates arrive late due to transient network conditions. In nested longhaul the updates also have to travel through the whole chain. So bumping this time will help us eliminate false failures in the tests.


test/modules/TwinTester/Settings.cs
```diff
@@ -29,8 +29,8 @@ static Settings Create()
                 configuration.GetValue<string>("TargetModuleId", moduleId),
                 configuration.GetValue<int>("TwinUpdateSize", 1),
                 configuration.GetValue<TimeSpan>("TwinUpdateFrequency", TimeSpan.FromSeconds(10)),
-                configuration.GetValue<TimeSpan>("TwinUpdateFailureThreshold", TimeSpan.FromMinutes(1)),
-                configuration.GetValue<TimeSpan>("EdgeHubRestartFailureTolerance", TimeSpan.FromMinutes(1)),
+                configuration.GetValue<TimeSpan>("TwinUpdateFailureThreshold", TimeSpan.FromMinutes(2)),
+                configuration.GetValue<TimeSpan>("EdgeHubRestartFailureTolerance", TimeSpan.FromMinutes(2)),
```
and-rewsmith, line 33: We want to give twin updates two minutes to come in, to avoid transient network issues failing the tests.

and-rewsmith: EdgeHub restart tolerance needs to also increase. This is because when edgehub restarts it will start closing upstream subscriptions on behalf of test modules. Therefore, we want to give any update a full two minutes to come through without the edgehub restart process interfering.

yophilav: Do we have a stat on this?

and-rewsmith: What do you mean by a stat?

yophilav (approved):

PR #4627, opened 2021-03-18
Fix erasing unsent data

nyanzebra:
What I am guessing is happening and will hopefully be able to replicate and see is, when read is at end of buffer and write is at beginning and we crash we won't get read at the right spot, instead it will be 0. We will then continue to read what is newly written and eventually write over previous data at old read pointer location. This means data loss.

### DIFF
--- a/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
+++ b/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
@@ -404,7 +404,9 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
     let mut start = inner.write_index();
     let mut end = start + block_size;
     let mut read = 0;
+    let mut write = 0;
     let mut order = inner.order();
+    let mut should_update_write = true;
 
     // if we have more write_updates than read_updates and write == read
     // then we must be in wrap_around case.
@@ -417,13 +419,13 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
         if inner.hint() != BLOCK_HINT {
             return RingBufferMetadata {
                 file_pointers: FilePointers {
-                    write: start,
+                    write,
                     read_begin: read,
                     read_end: read,
                 },
                 order,
                 // Check to see if wrap around case.
-                can_read_from_wrap_around: (start == read && write_updates > read_updates),
+                can_read_from_wrap_around: (write == read && write_updates > read_updates),
             };
         }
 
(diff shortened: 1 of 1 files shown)


nyanzebra:
Made this pr from phone so if have to fix anything, I will have to wait until home to make non-trivial changes

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -417,13 +419,13 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
         if inner.hint() != BLOCK_HINT {
             return RingBufferMetadata {
                 file_pointers: FilePointers {
-                    write: start,
+                    write,
```
and-rewsmith, line 422: When do we encounter memory without a block?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -417,13 +419,13 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
...
                     read_begin: read,
                     read_end: read,
                 },
                 order,
                 // Check to see if wrap around case.
-                can_read_from_wrap_around: (start == read && write_updates > read_updates),
+                can_read_from_wrap_around: (write == read && write_updates > read_updates),
```
and-rewsmith, line 428: What is the point of this ```can_read_from_wrap_around``` arg?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -436,19 +438,15 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
...
-            return RingBufferMetadata {
-                file_pointers: FilePointers {
-                    write: start,
-                    read_begin: read,
-                    read_end: read,
-                },
-                order,
-                // Check to see if wrap around case.
-                can_read_from_wrap_around: (start == read && write_updates > read_updates),
-            };
```
and-rewsmith, line 451: So this was the problem? We stopped this loop at the last write meaning it is possible that the read pointer is incorrect?

and-rewsmith (commented):
This change makes sense, and I would recommend adding a test, but seems like it might not be relevant to the bug I filed today. In that instance I don't believe edge hub stopped or crashed, unless you have reason to believe that the ring buffer crashed internally for some reason?

and-rewsmith:
Also the style check is failing

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Yeah, I think this was an issue, not sure if this is relevant to the issue you saw though

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: If we write and meet back up at read we shouldn't block read operations, so need to check if pointers are at same location and that there were more writes than read :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: If the file is empty then this can happen. Or more precisely, not full.

PR #4628, opened 2021-03-18
Add check version for agent image

huguesBouvier:
This PR add a version check for the agent image if the device is nested with another iotedge device.

When the device is nested, the iotedge version need to be 1.2 or above.

This PR also update the default template to 1.2.


edgelet/iotedge/src/check/checks/check_agent_image.rs
```diff
@@ -0,0 +1,226 @@
...
+        Ok(CheckResult::Ok)
+    }
+}
+
+fn check_agent_image_version(agent_image: &str) -> Result<CheckResult, failure::Error> {
+    // We don't match the repo mcr.microsoft.com because in nested edge we expect the repo to be parent_hostname:443
+    let re = Regex::new(r".*?/azureiotedge-agent:(?P<Major>\d+).(?P<Minor>\d+).*")
+        .context("Failed to create regex")?;
+
+    if let Some(caps) = re.captures(&agent_image) {
```
arsing, line 94: Just to be clear, currently if the image does *not* match the regex, you treat it as valid. Is that what you wanted?

edgelet/iotedge/src/check/checks/check_agent_image.rs
```diff
@@ -0,0 +1,226 @@
...
+                "mcr.microsoft.com/azureiotedge-agent:1.2.0-rc4-linux-arm64v8",
+                1,
+            ),
+            ("$upstream:4443/azureiotedge-agent:3.0.9.5-linux-arm32v7", 1),
+            ("randomImage/randomImage:1.0", 1),
+        ];
+
+        for agent_image in &array {
+            let result = if let Ok(CheckResult::Ok) = check_agent_image_version(&agent_image.0) {
+                1
```
arsing, line 218: Use `true` and `false` instead of `1` and `0`

edgelet/iotedge/src/check/checks/check_agent_image.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::check_agent_image_version;
+    use crate::check::CheckResult;
+
+    #[test]
+    fn test_check_agent_image_version() {
+        let array = vec![
```
arsing, line 131:
```rust
let test_cases = &[
```

edgelet/iotedge/src/check/checks/check_agent_image.rs
huguesBouvier: Hey, yes. If it does not match the regex, I suppose it is a custom image.
like customer.io.cr/customerRepo/customAgent:1.0-test
I'll rather be conservative here.

edgelet/iotedge/src/check/checks/check_agent_image.rs
huguesBouvier: thanks, old C programmer habit. Rust has true  and false :).

edgelet/iotedge/src/check/checks/check_agent_image.rs
huguesBouvier: Fixed too

arsing (approved):

PR #4630, opened 2021-03-19
[MQTT bridge] fix topic mapping

ancaantochi:
- fixing issue when topic is empty, like is in the doc example: https://docs.microsoft.com/en-us/azure/iot-edge/how-to-publish-subscribe?view=iotedge-2020-11#publish-and-subscribe-between-mqtt-brokers
- fixed other issue discovered while testing empty topic: topic as #, inPrefix: local and send publication on local. It was failing because it matched with topic_name = "local", local_prefix = "local" and it tried to strip prefix as format!("{}/", local_prefix)

- fixed another bug to set retry subscription as warning and add some details in the message. Subscription retry happens when it was rejected by upstream broker because was unauthorized, it would not happen when there is network issue with the upstream broker


vadim-kovalyov:
>     * fixed issue with topic and prefixes concatenation with/without separator

@ancaantochi can you please clarify this one? maybe with example?

vadim-kovalyov (commented):
>     * fixed issue with topic and prefixes concatenation with/without separator

@ancaantochi can you please clarify this one? maybe with example?

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -81,20 +83,38 @@ impl<S> StoreMqttEventHandler<S> {
...
+                            topic_name.strip_prefix::<&str>(
+                                format!("{}{}", local_prefix, TOPIC_SEPARATOR).as_ref(),
+                            )
+                        } else {
+                            topic_name.strip_prefix::<&str>(local_prefix)
+                        }
+                    })
+                    .map(|stripped_topic| match mapper.topic_settings.out_prefix() {
+                        Some(remote_prefix) => {
+                            if mapper.topic_settings.in_prefix().is_none() {
```
varunpuranik, line 105: I am confused - why do you need this logic to strip/add the topic separator? 
Why can't you always strip the '/' and add it back?

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -81,20 +83,38 @@ impl<S> StoreMqttEventHandler<S> {
...
-                .map_or(Some(topic_name), |local_prefix| {
-                    topic_name.strip_prefix(format!("{}/", local_prefix).as_str())
-                })
-                // match topic without local prefix with the topic filter pattern
-                .filter(|stripped_topic| mapper.topic_filter.matches(stripped_topic))
-                .map(|stripped_topic| match mapper.topic_settings.out_prefix() {
-                    Some(remote_prefix) => format!("{}/{}", remote_prefix, stripped_topic),
-                    None => stripped_topic.to_string(),
-                })
+            if mapper.topic_filter.matches(topic_name) {
```
varunpuranik, line 86: Can you add comments to explain the changes (liek this is needed if topicFilter is empty).

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -169,7 +187,10 @@ where
                             self.handle_unsubscribed(&unsub);
                         }
                         SubscriptionUpdateEvent::RejectedByServer(sub) => {
-                            debug!("received subscription rejected: {}", sub.topic_filter);
+                            warn!(
```
varunpuranik, line 190: Can we also add which server?

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -194,10 +215,13 @@ pub async fn retry_subscriptions(
 
     while let Some(subs) = retries.next().await {
         if !subs.is_empty() {
-            info!("try to re-subscribe to {} topics", subs.len());
+            warn!("try to re-subscribe to {} topics", subs.len());
```
varunpuranik, line 218: Nit: Trying to re-subscribing to xx topics.

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -194,10 +215,13 @@ pub async fn retry_subscriptions(
...
             for sub in subs {
                 if topic_mappers_updates.contains_key(&sub.topic_filter) {
-                    debug!("re-subscribe to {} qos {:?}", sub.topic_filter, sub.qos);
+                    warn!(
+                        "subscription was rejected by server, re-subscribe to {} qos {:?}",
```
varunpuranik, line 222: Do we know here that the sub was rejected by the server? Should we log it when it happens instead of here?
Nit - suggested text: Re-subscribing to topic {} with qos {}

varunpuranik (approved):

mqtt/mqtt-bridge/src/messages.rs
ancaantochi: it is because it was rejected because the way is retried now, but removed that part because it makes more sense to log when the reject is received

mqtt/mqtt-bridge/src/messages.rs
ancaantochi: doesn't have here the details about the server, changed to broker which now it must be upstream

mqtt/mqtt-bridge/src/messages.rs
ancaantochi: I changed the code to make it more clear. It can't strip / all the time because if it was not added then strip_prefix will return None, also there is a case when topic is /temp and inPrefix and outPrefix are empty so it should not be stripped.

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -81,20 +83,53 @@ impl<S> StoreMqttEventHandler<S> {
...
-                    Some(remote_prefix) => format!("{}/{}", remote_prefix, stripped_topic),
-                    None => stripped_topic.to_string(),
-                })
+            if mapper.topic_filter.matches(topic_name) {
+                mapper
+                    .topic_settings
+                    .in_prefix()
+                    // maps if local does not have a value it uses the topic that was received,
+                    // else it checks that the received topic starts with local prefix and removes the local prefix
+                    .map_or(Some(topic_name), |local_prefix| {
```
varunpuranik, line 92: Nit: should we call this in_prefix and out_prefix? The local_prefix here can be actually remote prefix for remote subscriptions, right?

mqtt/mqtt-bridge/src/messages.rs
varunpuranik: Thanks Anca.. but I am not sure I still follow. 
What happens in this case - 
InPrefix: ""
OutPrefix: "/foo"
Topic: "/telemetry"
Will you end up with an extra '/' in the output topic?

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -194,10 +230,13 @@ pub async fn retry_subscriptions(
 
     while let Some(subs) = retries.next().await {
         if !subs.is_empty() {
-            info!("try to re-subscribe to {} topics", subs.len());
+            warn!("trying to re-subscribe to {} topics", subs.len());
```
varunpuranik, line 233: This casing looks odd. Do we not use sentence casing anywhere?

mqtt/mqtt-bridge/tests/config.json
```diff
@@ -27,6 +27,43 @@
                 "topic": "floor2/#",
                 "inPrefix": "",
                 "outPrefix": ""
+            },
```
varunpuranik, line 30: This will conflict with this PR I suppose? https://github.com/Azure/iotedge/pull/4559/files

varunpuranik (approved):

mqtt/mqtt-bridge/src/messages.rs
ancaantochi: yes, it will set topic to /foo//telemetry
same for:
if InPrefix: "/in"
outPrefix: "/foo"
topic: "/telemetry"
it subscribes to /in//telemetry and outputs to /foo//telemetry because it doesn't strip / if it is included in topic

mqtt/mqtt-bridge/src/messages.rs
ancaantochi: broker is using lowercase for logging

mqtt/mqtt-bridge/src/messages.rs
varunpuranik: But that is not correct, right?
For this - 
if InPrefix: "/in"
outPrefix: "/foo"
topic: "/telemetry"
I expect the sub to be on /in/telemetry and pub on /foo/telemetry. Otherwise we should explicitly say that we will add a '/' in the middle.

mqtt/mqtt-bridge/tests/config.json
ancaantochi: yeap, conflict

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -81,20 +83,53 @@ impl<S> StoreMqttEventHandler<S> {
 
     fn transform(&self, topic_name: &str) -> Option<String> {
```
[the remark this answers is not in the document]
ancaantochi: I created a task to refactor it

vadim-kovalyov (commented):
There are some conflicts with my recent changes in Bridge config. Feel free to ping me for help.

mqtt/mqtt-bridge/tests/config.json
```diff
@@ -0,0 +1,94 @@
+{
```
vadim-kovalyov, line 1: this file is no longer needed after my changes, right?

vadim-kovalyov (commented):
LGTM, just need to remove old .json test file.

mqtt/mqtt-bridge/src/messages.rs
```diff
@@ -81,20 +83,53 @@ impl<S> StoreMqttEventHandler<S> {
...
+                                "topic {} was matched with {:#?}, but remote topic is not valid",
+                                topic_name, mapper.topic_settings
+                            );
+                            None
+                        } else {
+                            Some(transformed_topic)
+                        }
+                    })
+            } else {
+                None
```
dmolokanov, line 131: Code is rather complicated here we should probably refactor it. Maybe split in multiple functions will help.

dmolokanov (approved):
Overall looks good. Have some suggestions though

mqtt/mqtt-bridge/src/messages.rs
ancaantochi: I removed the default topic separator, now it expects to have:

topic: "bla"
inPrefix: "foo/"
outPrefix: "bar/"

subscribes to "foo/bla" and sends to "bar/bla"

mqtt/mqtt-bridge/src/messages.rs
ancaantochi: After Varun comments and discussion I changed the code so it doesn't add / by default. This made the code simple and also less cases to explain in doc for customers.

mqtt/mqtt-bridge/src/messages.rs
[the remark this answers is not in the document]
ancaantochi: I added more tests for new cases.

mqtt/mqtt-bridge/tests/config.json
ancaantochi: yeap, I added during merge and then I forgot to remove

vadim-kovalyov (approved):
🚢

PR #4631, opened 2021-03-19
Fix RingBuffer initialization logic

nyanzebra:
The Ring buffer currently doesn't read the whole file for determining where write and read pointers should be. It actually short-circuits out when write pointer is determined. This can allow for a case where messages cannot be sent and are instead written over. 

The scenario is as follows:
- RingBuffer is filled and write pointer wraps around
- Data is being batched and read pointer progresses beyond initial point
- A restart occurs which means RingBuffer needs to re-initialize
- RingBuffer finds write pointer before read pointer and doesn't continue to find read pointer
- More writes come in and write over data that was not sent as read pointer is behind write pointer after re-init.

### DIFF
--- a/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
+++ b/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
@@ -425,7 +425,8 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
                 },
                 order,
                 // Check to see if wrap around case.
-                can_read_from_wrap_around_when_write_full: (write == read && write_updates > read_updates),
+                can_read_from_wrap_around_when_write_full: (write == read
+                    && write_updates > read_updates),
             };
         }
 
@@ -464,7 +465,8 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
                 },
                 order,
                 // Check to see if wrap around case.
-                can_read_from_wrap_around_when_write_full: (write == read && write_updates > read_updates),
+                can_read_from_wrap_around_when_write_full: (write == read
+                    && write_updates > read_updates),
             };
         }
 
@@ -479,7 +481,8 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
                     },
                     order,
                     // Check to see if wrap around case.
-                    can_read_from_wrap_around_when_write_full: (write == read && write_updates > read_updates),
+                    can_read_from_wrap_around_when_write_full: (write == read
+                        && write_updates > read_updates),
                 };
             }
         };
@@ -860,7 +863,8 @@ mod tests {
 
             let loaded_read = rb.metadata.file_pointers.read_begin;
             let loaded_write = rb.metadata.file_pointers.write;
-            let loaded_can_read_from_wrap_around_when_write_full = rb.metadata.can_read_from_wrap_around_when_write_full;
+            let loaded_can_read_from_wrap_around_when_write_full =
+                rb.metadata.can_read_from_wrap_around_when_write_full;
             assert_eq!(read, loaded_read);
             assert_eq!(write, loaded_write);
             assert!(!loaded_can_read_from_wrap_around_when_write_full);
@@ -1367,7 +1371,7 @@ mod tests {
             assert_matches!(result, Ok(_));
             let mut batch = result.unwrap();
             while !batch.is_empty() {
-                for (key , publication) in entries.drain(..) {
+                for (key, publication) in entries.drain(..) {
                     let maybe_entry = batch.remove(0);
                     assert!(maybe_entry.is_some());
                     let entry = maybe_entry.unwrap();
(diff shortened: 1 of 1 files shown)


mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -779,6 +790,83 @@ mod tests {
...
+            let mut batch = result.unwrap();
+            assert!(!batch.is_empty());
+
+            for (key, _) in batch.drain(..) {
+                let result = rb.remove(key);
+                assert_matches!(result, Ok(_));
+            }
+
+            // write till wrap around
+            // this will put write ahead of read
```
and-rewsmith, line 834: If we read some values then write until the wraparound, won't write be _before_ read?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -1191,6 +1279,94 @@ mod tests {
         }
     }
 
+    #[test]
+    fn it_batches_ok_when_leftover_from_file_and_write_only_wrap_around() {
```
and-rewsmith, line 1283: Just to confirm I am understanding this, this is doing the same thing to the ring buffer as the other test you added. But it is testing the batching?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Yeah, I can fix the language here

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -436,39 +438,48 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
...
                 file_pointers: FilePointers {
-                    write: start,
+                    write,
                     read_begin: read,
                     read_end: read,
                 },
                 order,
                 // Check to see if wrap around case.
-                can_read_from_wrap_around: (start == read && write_updates > read_updates),
+                can_read_from_wrap_around: (write == read && write_updates > read_updates),
```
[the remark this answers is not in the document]
nyanzebra: Maybe the variable is poorly named... the particular case is for when write wraps around is at the 'exact' position that read is at. Because I don't allow for read or write pointers to pass one another, this is for allowing a special case where it is actually okay because we wrote a full buffer and can read.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -436,39 +438,48 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
             read = end + data_size;
         }
 
-        // Found the last write, take whatever we got for read and write
-        // and return the pointers.
+        // Check next block
+        write_updates += 1;
+        order += 1; // We want to start with the next order number.
+        start = end + data_size;
+        end = start + block_size;
```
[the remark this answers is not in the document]
nyanzebra: Previous pr was off a bit. After getting home yesterday and running against the tests I found that the logic needed to be moved down to point to the correct block. Otherwise it points to the block before the one it should. So, moved it below the update of start.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -436,39 +438,48 @@ fn find_pointers_and_order_post_crash(file: &mut File, max_file_size: u64) -> Ri
...
+
+        // Found the last write, can stop updating write.
         if inner.order() < order {
+            should_update_write = false;
+            // Update write one last time to point at block
+            // where order is less.
+            write = start % max_file_size;
```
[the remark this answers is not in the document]
nyanzebra: So, this particular case is as follows:
1. Fill RingBuffer entirely, except small portion that cannot fit next amount of data
2. Read and delete first entry so that new entry will fit
3. Write entry, where write pointer is now at beginning area of the buffer and 'before' read pointer
4. crash
5. Initialization will read the whole file and find the last write at the end of the buffer, but that the write pointer should wrap around and start after that last entry

Can also draw out any of these cases if they help :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Will fix the language here :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: It is a very similar test, but instead of testing that the pointers are where they should be I am verifying can still read. Will update this further with different data inserted to verify data is correct.

and-rewsmith (approved):

PR #4636, opened 2021-03-19
Simplify insert check logic in RingBuffer

nyanzebra:
The logic in RingBuffer for validating if an insert into buffer can take place can be simplified. We basically need to check two things:
1. That if write has wrapped around and we have a full buffer but both read and write pointers are equal. This is done with a variable that is already tracking this.
2. If read and write pointer are not equal, then we should check that we have enough free space available for the insert.

### DIFF
--- a/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
+++ b/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
@@ -200,21 +200,15 @@ impl StreamWakeableState for RingBuffer {
 
         let block_size = *SERIALIZED_BLOCK_SIZE;
         let total_size = block_size + data_size;
-        // Check to see if we might corrupt data if we write, if so, return err of full.
-        // There are two cases for this:
-        // 1. write has wrapped around and is now behind read
-        if write_index < read_index && write_index + total_size > read_index {
+
+        // Check that we have enough space to insert data.
+        // If we have set can_read_from_wrap_around_when_write_full
+        // then we must also be full.
+        let free_space = (read_index + self.max_file_size - write_index) % self.max_file_size;
+        if self.metadata.can_read_from_wrap_around_when_write_full {
             return Err(PersistError::RingBuffer(RingBufferError::Full));
         }
-        // 2. write has reached the end (or close enough) but read has not moved
-        if read_index == 0 && write_index + total_size > self.max_file_size {
-            return Err(PersistError::RingBuffer(RingBufferError::Full));
-        }
-        // 3. check if write would wrap around end and corrupt
-        if write_index > read_index
-            && write_index + total_size > self.max_file_size
-            && (write_index + total_size) % self.max_file_size > read_index
-        {
+        if read_index != write_index && free_space < total_size {
             return Err(PersistError::RingBuffer(RingBufferError::Full));
         }
 


mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -200,21 +200,15 @@ impl StreamWakeableState for RingBuffer {
...
-            return Err(PersistError::RingBuffer(RingBufferError::Full));
-        }
-        // 2. write has reached the end (or close enough) but read has not moved
-        if read_index == 0 && write_index + total_size > self.max_file_size {
+
+        // Check that we have enough space to insert data.
+        // If we have set can_read_from_wrap_around_when_write_full
+        // then we must also be full.
+        let free_space = (read_index + self.max_file_size - write_index) % self.max_file_size;
+        if self.metadata.can_read_from_wrap_around_when_write_full {
```
varunpuranik, line 208: What does this flag indicate? Storage is full? If so, then you can maybe check that first thing.

nyanzebra: Yeah, the flag gets set when write pointer reaches read pointer and we are full. :)

nyanzebra: Just to be clear you mean this correct?

```rust
if self.metadata.can_read_from_wrap_around_when_write_full || 
   (read_index != write_index && free_space < total_size) {
    return Err(PersistError::RingBuffer(RingBufferError::Full));
}
```

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -200,21 +200,15 @@ impl StreamWakeableState for RingBuffer {
...
             return Err(PersistError::RingBuffer(RingBufferError::Full));
         }
-        // 3. check if write would wrap around end and corrupt
-        if write_index > read_index
-            && write_index + total_size > self.max_file_size
-            && (write_index + total_size) % self.max_file_size > read_index
-        {
+        if read_index != write_index && free_space < total_size {
```
varunpuranik, line 211: read_index should always be == write_index when storage is full, right? So it should never be true here?

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
varunpuranik: No, I just meant before the storage full check..  it was a nit

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: The problem is when read == write and we are not full. There are two conditions where they are equal:
1. Init of a clean slate RB where both read and write are at 0 (this was original reason why there was read_index == 0)
2. When buffer is full and write caught back up to read.

So, I think we need to have the check here to allow for initial clean-slate RB to write. We could have some other boolean flag for this if that seems cleaner, but this is the problem basically.

varunpuranik (approved):

PR #4638, opened 2021-03-19
Remove case sensitivity in hostname check (#4611)

huguesBouvier:
Cherry pick https://github.com/Azure/iotedge/pull/4611


arsing:
I've already deleted this code in https://github.com/Azure/iotedge/pull/4643 so you could just not bother cherry-picking this.

PR #4642, opened 2021-03-20
Longhaul: Add threshold for device not found

and-rewsmith:
Longhaul is seeing DeviceNotFound exceptions. We didn't used to see this. Bear and I think that this is because either:
- EdgeHub is taking a long time to restart
- The exception's transient field was flipped from true to false by the sdk

The fix here is to implement a tolerance to these exceptions, similarly to how we do with status code 0.


test/modules/DirectMethodSender/DirectMethodSenderBase.cs
```diff
@@ -54,7 +54,7 @@ abstract class DirectMethodSenderBase : IDisposable
             }
             catch (DeviceNotFoundException e)
             {
-                logger.LogInformation(e, $"Transient exception caught with count {this.directMethodCount}");
+                logger.LogInformation(e, $"DeviceNotFound exception caught with count {this.directMethodCount}");
```
and-rewsmith, line 57: This exception isn't transient according to the sdk so I am clarifying this.

test/modules/TestResultCoordinator/Reports/DirectMethod/LongHaul/DirectMethodLongHaulReport.cs
```diff
@@ -54,10 +55,11 @@ public bool IsPassedHelper()
             // We expect to get this status sometimes because of edgehub restarts, but if we receive too many we should fail the tests.
             // TODO: When the SDK allows edgehub to de-register from subscriptions and we make the fix in edgehub, then we can fail tests for any status code 0.
             long allStatusCount = this.SenderSuccesses + this.StatusCodeZero + this.Other.Sum(x => x.Value);
-            bool statusCodeZeroBelowThreshold = (this.StatusCodeZero == 0) || (this.StatusCodeZero < ((double)allStatusCount / 100));
```
and-rewsmith, line 57: We don't see status code 0 nearly ever anymore, so I am lowering the tolerance to 0.1%.

yophilav (approved):

PR #4649, opened 2021-03-22
Add test for rare RB init edge case

nyanzebra:
It is possible for a crash to occur where there are no deleted block but reads have occurred. In this case we would want to have read at write position on init as this should only happen if the buffer is full.

### DIFF
--- a/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
+++ b/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
@@ -908,7 +908,7 @@ mod tests {
 
             for (key, _) in batch.drain(..) {
                 rb.remove(key)
-                    .expect(&format!("unable to remove pub with key {:?}", key).to_owned());
+                    .unwrap_or_else(|_| panic!(format!("unable to remove pub with key {:?}", key)));
             }
 
             read = rb.metadata.file_pointers.read_begin;
--- a/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
+++ b/mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
@@ -872,9 +872,7 @@ mod tests {
 
     #[test]
     fn it_inits_ok_with_previous_data_and_write_pointer_is_reaches_read_pointer_after_read() {
-        let result = tempfile::NamedTempFile::new();
-        assert_matches!(result, Ok(_));
-        let file = result.unwrap();
+        let file = tempfile::NamedTempFile::new().unwrap();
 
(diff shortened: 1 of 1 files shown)


mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -865,6 +870,85 @@ mod tests {
         }
     }
 
+    #[test]
+    fn it_inits_ok_with_previous_data_and_write_pointer_is_reaches_read_pointer_after_read() {
+        let result = tempfile::NamedTempFile::new();
+        assert_matches!(result, Ok(_));
```
vadim-kovalyov, line 876: I'm not sure why do you need this? why not just use `.expect("...")` or `.unwrap()` directly? I believe asserts are better for actual test results you want to verify, not for intermediate/setup code.

nyanzebra: Yeah will change :)

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -865,6 +870,80 @@ mod tests {
...
+                RingBuffer::new(&file.path().to_path_buf(), max_file_size, FLUSH_OPTIONS).unwrap();
+
+            // write some
+            for _ in 0..10 {
+                let result = rb.insert(&publication);
+                assert_matches!(result, Ok(_));
+            }
+
+            let result = rb.batch(10);
+            assert_matches!(result, Ok(_));
```
vadim-kovalyov, line 907: I think this can also be replaced with `.expect(..)` or `.unwrap()`.

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
```diff
@@ -865,6 +870,80 @@ mod tests {
...
+            let mut batch = result.unwrap();
+            assert!(!batch.is_empty());
+
+            for (key, _) in batch.drain(..) {
+                let result = rb.remove(key);
```
vadim-kovalyov, line 912: instead of `assert_matches!`
```suggestion
                let result = rb.remove(key).expect(format!("unable to remove pub with key {}", key));
```

mqtt/mqtt-bridge/src/persist/waking_state/ring_buffer/mod.rs
nyanzebra: Done :)

vadim-kovalyov (approved):

PR #4650, opened 2021-03-22
Fix versioning scheme

huguesBouvier:
This PR fix versioning constraints so versioning is more flexible for development.


gordonwang0:
It should be `1.2.0~rc4-1`, not `1.2.0~rc4.1`.

gordonwang0 (approved):

PR #4654, opened 2021-03-22
Update ASP.NET Core Runtime base images

yophilav:
- Linux AMD64: `3.1.12-alpine3.12` --> `3.1.13-alpine3.13`
- Linux ARM32: `3.1.12-bionic-arm*` --> `3.1.13-bionic-arm*` (1.0.6.8-linux-arm*)


test/modules/TemperatureFilter/docker/linux/amd64/Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG base_tag=3.1.12-alpine3.12
-FROM mcr.microsoft.com/dotnet/core/runtime:${base_tag}
```
and-rewsmith, line 2: Why are we removing the ```core``` path?

and-rewsmith (approved):

PR #4659, opened 2021-03-23
[TEST-ONLY] Demo version

yophilav:
Demo version


micahl:
@yophilav Per internal conversation let's use `1.2.0~rc4~device-update-tutorial-1` instead of `1.2.0~pre-rc4-1`.

PR #4660, opened 2021-03-23
Add iotedge system reprovision

gordonwang0:
Adds a command to reprovision device with IoT Hub. Also restarts edged so that the new provisioning info is used.


edgelet/iotedge/src/system.rs
```diff
@@ -71,4 +74,26 @@ impl System {
             Error::from(ErrorKind::System)
         })
     }
+
+    pub fn reprovision(runtime: &mut tokio::runtime::Runtime) -> Result<(), Error> {
+        let uri = url::Url::parse("unix:///run/aziot/identityd.sock")
+            .expect("hard-coded URI should parse");
+        let client = IdentityClient::new(ApiVersion::V2020_09_01, &uri);
```
arsing, line 81: Do you think it would be better to shell out to `aziotctl reprovision` instead? On the one hand it reduces duplication. On the other hand it's a very small piece of code either way. Your choice.

arsing (approved):

edgelet/iotedge/src/system.rs
gordonwang0: I thought about having it shell out, but this way was easier in my mind.

PR #4668, opened 2021-03-24
Add sas token to API proxy

huguesBouvier:
The API proxy will now get a token SAS before starting and use it to connect with edgeHub for all https connections

Test:
1. Tested reload
Hacked to request new key every 5s and changed  the configuration to display the key:
        location / {
            return 200 "${SAS_TOKEN}";
        }

And checked the signature was updating:
 curl http://127.0.0.1:443
SharedAccessSignature sr=20.83.177.78/devices/huguesChild/modules/debug_rust&sig=xxxxx
curl http://127.0.0.1:443
SharedAccessSignature sr=20.83.177.78/devices/huguesChild/modules/debug_rust&sig=xxxxx

To test for stability, left that code reloading 100 times

2. tested restart:
Same code as above, but instead of reload, requests a restart.
Using PS aux, check nginx: master process /usr/sbin/nginx -c /app/nginx_con is changing PID every time.
To test for stability, left that code restarting 100 times

3. Test that reload does not interrupt download:
Seems that docker download is robust. I was not able to break it by restarting the proxy or by reloading the config.

4. Stability test:
Booted nginx 100 times.
+See test 1 and 2


edge-modules/api-proxy-module/src/main.rs
```diff
@@ -101,60 +112,80 @@ pub fn nginx_controller_start(
...
+                );
+
+                // Bug in clippy, not using mut mut here
+                #[allow(clippy::mut_mut)]
+                let proxy_command = select! {
+                    _ = wait_shutdown_ctrl_c => NginxCommands::Stop,
+                    _ = wait_shutdown_signal => NginxCommands::Stop,
+                    _ = cert_reload => NginxCommands::Reload,
+                    _ = config_reload => NginxCommands::Reload,
+                    _ = nginx_start => NginxCommands::Restart,
```
varunpuranik, line 145: When does this return? When nginx crashes?

edge-modules/api-proxy-module/src/monitors/certs_monitor.rs
```diff
@@ -16,12 +16,13 @@ const PROXY_SERVER_CERT_PATH: &str = "/app/server.crt";
 const PROXY_SERVER_PRIVATE_KEY_PATH: &str = "/app/private_key_server.pem";
 
 const PROXY_SERVER_VALIDITY_DAYS: i64 = 90;
+const CERTIFICATE_POLL_INTERVAL: tokio::time::Duration = tokio::time::Duration::from_secs(1);
```
varunpuranik, line 19: is tokio:: needed here?

edge-modules/api-proxy-module/src/monitors/token_client.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    pub async fn get_new_sas_token(
+        &mut self,
+        current_date: DateTime<Utc>,
+    ) -> Result<Option<String>, Error> {
+        //If there is no previous token request a new one right away
+        if let Some(expiration_date) = self.token_expiration_date {
+            //If Token is not expired, sleep until it is.
+            use std::convert::TryInto;
```
varunpuranik, line 91: Can you pls explain what the logic here is doing?

edge-modules/api-proxy-module/src/monitors/token_client.rs
```diff
@@ -0,0 +1,142 @@
...
+                error!("Token has expired before API proxy was able to request a new one. Requesting a new SAS token now");
+            }
+        }
+
+        info!("Generating new token");
+        // Add duration time and remove the margin.
+        let new_expiration_date = Utc::now()
+            .checked_add_signed(self.validity_duration)
+            .context("Could not compute new expiration date for certificate")?
+            .checked_sub_signed(Duration::seconds(TOKEN_EXPIRY_SECONDS_MARGIN))
```
varunpuranik, line 111: Why are you subtracting the margin secs here?

edge-modules/api-proxy-module/src/monitors/config_monitor.rs
```diff
@@ -56,51 +65,66 @@ pub fn start(
...
+            let parse_config_request =
+                match futures::future::select(wait_shutdown, reload_config).await {
+                    Either::Left(_) => {
+                        warn!("Shutting down config monitor!");
+                        return Ok(());
+                    }
+                    Either::Right((reload_config, _)) => match reload_config {
+                        Either::Left((Ok(Some(token)), _)) => {
+                            info!("New SAS token received, reloading the config");
+                            std::env::set_var("SAS_TOKEN", token);
```
varunpuranik, line 87: Emm, i thought you were going to write this to a file?

varunpuranik: Env var is visible from outside the container as well.. (docker inspect..) so not sure if this is better.

edge-modules/api-proxy-module/src/monitors/token_client.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    pub async fn get_new_sas_token(
+        &mut self,
+        current_date: DateTime<Utc>,
+    ) -> Result<Option<String>, Error> {
+        //If there is no previous token request a new one right away
+        if let Some(expiration_date) = self.token_expiration_date {
+            //If Token is not expired, sleep until it is.
+            use std::convert::TryInto;
+            if current_date < expiration_date {
```
varunpuranik, line 92: I think we should simplify the logic here - 
TokenClient should be just responsible for getting the token with the specified duration.
Have an outside loop that gets a token every x mins (say 5) of duration x + y (say 10), passes it back to Nginx and reloads it. 
The code that connects to EdgeHub and maintains a connection gets a token independently from TokenClient (of say 1 hr duration).

edge-modules/api-proxy-module/src/main.rs
huguesBouvier: No when nginx crashes it just restarts it.
It returns on  NginxCommands::Stop. During a signal or ctrl+c

edge-modules/api-proxy-module/src/monitors/config_monitor.rs
huguesBouvier: We write it into a file. That env var is not accessible upon docker inspect. I'll check but it should be local to the process

edge-modules/api-proxy-module/src/monitors/token_client.rs
huguesBouvier: Let's chat about it.

edge-modules/api-proxy-module/src/monitors/certs_monitor.rs
huguesBouvier: It is necessary here or inside the code because we need a duration. I can move it inside the code if it is ugly here

edge-modules/api-proxy-module/src/monitors/token_client.rs
huguesBouvier: This is the same as above, let's chat about it

edge-modules/api-proxy-module/src/monitors/token_client.rs
huguesBouvier: Let's discuss it but in a nutshell:
TokenClient is handling getting the token and its expiration. It make sense that it handles the margin too because margin is part of making sure that we are correctly handling expiry times.

edge-modules/api-proxy-module/src/main.rs
varunpuranik: I am not sure I follow. My questions is - when does nginx_start future return, resulting in NginxCommands;:Restart being executed

edge-modules/api-proxy-module/src/monitors/certs_monitor.rs
varunpuranik: no, just a nit - I mean can you do time::Duration instead of tokio::time::Duration

edge-modules/api-proxy-module/src/monitors/token_client.rs
```diff
@@ -0,0 +1,70 @@
+use anyhow::{Error, Result};
+use log::info;
+use url::form_urlencoded::Serializer as UrlSerializer;
+
+pub const IOTHUB_ENCODE_SET: &percent_encoding::AsciiSet = &percent_encoding::CONTROLS // C0 control
```
varunpuranik, line 5: Can we use this instead? 
https://github.com/Azure/iotedge/blob/master/mqtt/edgelet-client/src/workload/client.rs#L15

edge-modules/api-proxy-module/src/monitors/token_manager.rs
```diff
@@ -0,0 +1,96 @@
...
+            .checked_add_signed(validity_duration)
+            .context("Could not compute new expiration date for certificate")?;
+
+        let token = self
+            .token_client
+            .get_new_sas_token(&new_expiration_date.timestamp().to_string())
+            .await?;
+
+        //Set a "conservative" expiration date, the actual expiration date, minus a margin
+        let new_expiration_date = new_expiration_date
```
varunpuranik, line 86: Nit: suggest calling it something like token_renewal_time or something like that.

edge-modules/api-proxy-module/src/monitors/token_manager.rs
```diff
@@ -0,0 +1,96 @@
...
+            .checked_sub_signed(margin_before_expiry)
+            .context(
+                "When removing the margin, could not compute new expiration date for certificate",
+            )?;
+
+        self.token_expiration_date = Some(new_expiration_date);
```
varunpuranik, line 92: this too

edge-modules/api-proxy-module/src/monitors/config_monitor.rs
huguesBouvier: 1. I pass it as an env VAR because currently API proxy is meant to be configured by user through environment variable. User set env var, API proxy parse the default config and replace place holder with user env var. I set env var here, instead of passing a parameter to keep the previous logic and avoid a refactoring of the function.
2. Now that we have 2 means of configuring that API proxy, will need to refactor that config.

edge-modules/api-proxy-module/src/main.rs
huguesBouvier: Ah I see. Yes, what you said first is correct. The future returns what nginx crashes. I though you meant return from the function.

edge-modules/api-proxy-module/src/monitors/config_monitor.rs
varunpuranik: Okay, can you pls add a comment to explain 
1. Why you are setting the env var
2. TODO to refactor this in the future

varunpuranik (approved):

edge-modules/api-proxy-module/src/monitors/token_client.rs
huguesBouvier: Fixed

edge-modules/api-proxy-module/src/monitors/token_manager.rs
huguesBouvier: Fixed

edge-modules/api-proxy-module/src/monitors/token_manager.rs
huguesBouvier: Fixed

edge-modules/api-proxy-module/src/monitors/config_monitor.rs
huguesBouvier: Explained above.

PR #4670, opened 2021-03-24
Nested Longhaul: Add debug logging for non-broker case (#4607)

and-rewsmith:
Nested Longhaul: Add debug logging for non-broker case.

This was missed when the deployment was created. Adding it now.


yophilav (approved):
Please update the PR description to be more explicit

PR #4671, opened 2021-03-24
Longhaul: Make log rotation less restrictive on amd64 (#4604)

and-rewsmith:
Current log rotation settings for longhaul are configured for pis. We need to have ample logs, especially on amd64 where we do most of our testing. I have added logic to the trcE2Escript.sh to set the log size differently depending on platform.

### DIFF
--- a/scripts/linux/trcE2ETest.sh
+++ b/scripts/linux/trcE2ETest.sh
@@ -216,6 +216,7 @@ function prepare_test_from_artifacts() {
         sed -i -e "s@<DesiredModulesToRestartCSV>@$DESIRED_MODULES_TO_RESTART_CSV@g" "$deployment_working_file"
         sed -i -e "s@<RestartIntervalInMins>@$RESTART_INTERVAL_IN_MINS@g" "$deployment_working_file"
         sed -i -e "s@<SendReportFrequency>@$SEND_REPORT_FREQUENCY@g" "$deployment_working_file"
+        sed -i -e "s@<LogRotationMaxFile>@$log_rotation_max_file@g" "$deployment_working_file"
     fi
 
     if [[ "${TEST_NAME,,}" == "${CONNECTIVITY_TEST_NAME,,}" ]]; then
@@ -866,10 +867,12 @@ quickstart_working_folder="$working_folder/quickstart"
 image_architecture_label=$(get_image_architecture_label)
 optimize_for_performance=true
 log_upload_enabled=true
+log_rotation_max_file="125"
 if [ "$image_architecture_label" = 'arm32v7' ] ||
-   [ "$image_architecture_label" = 'arm64v8' ]; then
+    [ "$image_architecture_label" = 'arm64v8' ]; then
     optimize_for_performance=false
     log_upload_enabled=false
+    log_rotation_max_file="7"
 fi
 
 deployment_working_file="$working_folder/deployment.json"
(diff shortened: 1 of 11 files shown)


scripts/linux/trcE2ETest.sh
```diff
@@ -866,10 +867,12 @@ quickstart_working_folder="$working_folder/quickstart"
 image_architecture_label=$(get_image_architecture_label)
 optimize_for_performance=true
 log_upload_enabled=true
+log_rotation_max_file="125"
```
yophilav, line 870: 125 files * 4MB each * ~16 modules = ~8GB of disk. 

Per https://docs.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series, Standard_D4s_v3 support up to 32GB disk. I think we should be safe to bump the disk usage up to ~30GB i.e. 480 files?

test/modules/TestResultCoordinator/TestReportUtil.cs
```diff
@@ -156,7 +156,7 @@ internal static async Task UploadLogsAsync(string iotHubConnectionString, Uri bl
             ServiceClient serviceClient = ServiceClient.CreateFromConnectionString(iotHubConnectionString);
             CloudToDeviceMethod uploadLogRequest =
                 new CloudToDeviceMethod("UploadModuleLogs")
-                    .SetPayloadJson($"{{ \"schemaVersion\": \"1.0\", \"sasUrl\": \"{blobContainerWriteUri.AbsoluteUri}\", \"items\": [{{ \"id\": \".*\", \"filter\": {{}} }}], \"encoding\": \"gzip\" }}");
+                    .SetPayloadJson($"{{ \"schemaVersion\": \"1.0\", \"sasUrl\": \"{blobContainerWriteUri.AbsoluteUri}\", \"items\": [{{ \"id\": \".*\", \"filter\": {{}} }}], \"encoding\": \"none\",\"content-type\": \"text\" }}");
```
yophilav, line 159: so... we compressed each log and upload each log as a text stream?! I think we could compress it even more? (ps. I'm not sure how much memory the 2nd compression will help in term of size).

scripts/linux/trcE2ETest.sh
and-rewsmith: Some test agents don't have nearly that amount of space. It might be safer to have some agents fail due to disk space so we can see longer logs some of the time. We should handle in a separate PR.

test/modules/TestResultCoordinator/TestReportUtil.cs
and-rewsmith: The reason I went with text stream here is because the compression we were using previously was broken. Now we can at least see the logs.

and-rewsmith: Before it was uploading a .gz that was not in a readable format.

scripts/linux/trcE2ETest.sh
and-rewsmith: When I checked some agents only had ~6 GB of free space, so I thought this was already pushing it. But we can adjust further per logic above.

test/modules/TestResultCoordinator/TestReportUtil.cs
and-rewsmith: I also have separate bug logged to find out if this log upload is a product issue

yophilav (approved):

PR #4672, opened 2021-03-24
Nested Tests: Default dynamic reprovisioning to false (#4622)

and-rewsmith:
For our nested longhaul tests, dynamic provisioning is set to true when it is not using dps. I don't think this config is used with any dps tests as the same config sets up manual provisioning. So it should be safe to update.


yophilav (approved):
Good catch!

PR #4673, opened 2021-03-24
Longhaul: Add threshold for device not found (#4642)

and-rewsmith:
Longhaul is seeing DeviceNotFound exceptions. We didn't used to see this. Bear and I think that this is because either:
- EdgeHub is taking a long time to restart
- The exception's transient field was flipped from true to false by the sdk

The fix here is to implement a tolerance to these exceptions, similarly to how we do with status code 0.


dylanbronson (approved):
LGTM

PR #4674, opened 2021-03-24
[k8s] Throw non-fatal exception if Mounts missing critical data.

darobs:
Most data coming from createOptions is applied to the PodSpec if set, but Mounts are a slightly different story, it has three critical fields, and the Type and Source must be set, or else we don't know how to handle the mount.

Creating an invalid mount in the Mounts section of createOptions was throwing a Null Reference exception, which appears to be marked as a fatal exception, and it causes edgeAgent to crash.  A better experience is to check the fields we need and throw a non-fatal exception.  This also lets us report back to the user what went wrong more clearly, doesn't crash edgeAgent, and allows for a graceful recovery once the problem is fixed.

I did do a check of the other mappers, and I am reasonably certain the mappers changed in this PR are the ones which will most likely cause NullReference exception.

test:
Create an edge deployment with modules using HostConfig.Mounts, but drop critical fields ("Type", "Source", "Target")
- Hub will show a status of "500" and the reported properties will point out the failed module&Mount
Fix the Mount,
- Hub will show 200 OK.

### DIFF
--- a/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
+++ b/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
@@ -334,8 +334,21 @@ namespace Microsoft.Azure.Devices.Edge.Agent.Kubernetes.EdgeDeployment.Deploymen
                 .ForEach(mounts => volumeMountList.AddRange(mounts));
 
             // collect volumes and volume mounts from HostConfig.Mounts section for binds to host path
-            var bindMounts = module.Config.CreateOptions.HostConfig
-                .FlatMap(config => Option.Maybe(config.Mounts))
+            var mountListOption = module.Config.CreateOptions.HostConfig
+                .FlatMap(config => Option.Maybe(config.Mounts));
+
+            mountListOption.ForEach(mounts =>
+                {
+                    foreach (var mount in mounts)
+                    {
+                        if (mount.Type is null || mount.Source is null || mount.Target is null)
+                        {
+                            throw new InvalidMountException($"Type, Source, and Target required for all mounts ({mount.Type}, {mount.Source}, {mount.Target})");
+                        }
+                    }
+                });
+
+            var bindMounts = mountListOption
                 .Map(mounts => mounts.Where(mount => mount.Type.Equals("bind", StringComparison.InvariantCultureIgnoreCase)).ToList());
 
             bindMounts.Map(mounts => mounts.Select(mount => new V1Volume(KubeUtils.SanitizeDNSLabel(mount.Source), hostPath: new V1HostPathVolumeSource(mount.Source, "DirectoryOrCreate"))))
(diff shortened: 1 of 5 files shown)


edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
```diff
@@ -32,7 +32,7 @@ public class KubernetesPvcMapper : IKubernetesPvcMapper
 
         bool ShouldCreatePvc(Mount mount)
```
yophilav, line 33: Should the type be `Mount?` instead of `Mount` ?

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesDeploymentMapperTest.cs
```diff
@@ -745,6 +784,45 @@ public void PassImagePullSecretsInPodSpecForProxyAndModuleContainers()
             Assert.Contains(deployment.Spec.Template.Spec.ImagePullSecrets, secret => secret.Name == "user-registry2");
         }
 
+        [Fact]
+        public void FailMapWhenMountWhenMountTypeIsMissing()
+        {
+            var identity = new ModuleIdentity("hostname", "gatewayhost", "deviceid", "Module1", Mock.Of<ICredentials>());
+            var hostConfig = NoTypeVolumeMountHostConfig;
```
yophilav, line 791: Do we have a test case in which `hostConfig = null`?

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/pvc/KubernetesPvcMapper.cs
darobs: Since we are not currently using [Nullable Reference Types](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types), it is not necessary, since objects can be null.

edge-agent/test/Microsoft.Azure.Devices.Edge.Agent.Kubernetes.Test/KubernetesDeploymentMapperTest.cs
darobs: The `NoMountsNoClaims` test uses default (=null) `HostConfig`.  The `CreatePodParameters` class keeps `HostConfig` as an `Option`.

edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Kubernetes/edgedeployment/deployment/KubernetesDeploymentMapper.cs
```diff
@@ -336,8 +336,21 @@ List<V1EnvVar> CollectModuleEnv(KubernetesModule module, IModuleIdentity identit
...
-            var bindMounts = module.Config.CreateOptions.HostConfig
-                .FlatMap(config => Option.Maybe(config.Mounts))
+            var mountListOption = module.Config.CreateOptions.HostConfig
+                .FlatMap(config => Option.Maybe(config.Mounts));
+
+            mountListOption.ForEach(mounts =>
+                {
+                    foreach (var mount in mounts)
+                    {
+                        if (mount.Type is null || mount.Source is null || mount.Target is null)
```
dmolokanov, line 346: well, my csharp became a bit rusty, what does this mean?
`mount.Type is null` - it looks sql to me :D

dmolokanov: https://devblogs.microsoft.com/dotnet/new-features-in-c-7-0/#is-expressions-with-patterns

dmolokanov (approved):
LGTM!

PR #4679, opened 2021-03-24
Read `parent_hostname` configuration from aziot

daprilik:
Builds on top of https://github.com/Azure/iot-identity-service/pull/175, which moved parent_hostname configuration down into aziot.

Outstanding TODOs:
 
- [x] fixup `iotedge check` and `iotedge config`
- [x] figure out how to wire up the agent-image parent_hostname resolution logic
- [x] remove local aziot crate patches + bump aziot version


edgelet/aziot-edged/src/lib.rs
```diff
@@ -499,12 +490,10 @@ where
         settings.hostname().to_string().to_lowercase(),
     );
 
-    if let Some(parent_hostname) = settings.parent_hostname() {
```
massand, line 502: We want to modify (not remove) the check to compare the parent_hostname to hostname. If they are the same, we should skip inserting the `GATEWAY_HOSTNAME_KEY` environment var.

edgelet/edgelet-docker/src/settings.rs
```diff
@@ -168,28 +164,33 @@ fn init_agent_spec(settings: &mut Settings) -> Result<(), LoadSettingsError> {
...
-    }
-
-    Ok(())
-}
+// TODO: wiring this up will be tricky, since Settings is loaded before the
+// parent_hostname is fetched from aziot...
+//
+// temporarily commenting this out to make forwards progress.
+
+// fn agent_image_resolve(settings: &mut Settings) -> Result<(), LoadSettingsError> {
```
massand, line 177: (I think you might be looking into this already - consider my suggestion if you haven't implemented)
The late binding for `$upstream` should be happening in docker config instead of edged - 
https://github.com/Azure/iotedge/blob/master/edgelet/edgelet-docker/src/config.rs#L12

This way, in the short to mid term future, when edge-agent logic makes its way down to edged, there won't be two different sources of truth to worry about. Current, there is coupling between the edged setting and the gatewayhostname environment variable - https://github.com/Azure/iotedge/blob/master/edge-agent/src/Microsoft.Azure.Devices.Edge.Agent.Core/NestedEdgeParentUrlParser.cs#L8-L34

edgelet/aziot-edged/src/lib.rs
huguesBouvier: parent_hostname == hostname is a possible use case?

huguesBouvier (approved):

daprilik:
I'm really not proud of the code I just wrote to handle `$upstream` resolution in the agent image. I'm going to cross my fingers and hope that someone can offer a better suggestion on how to pull this off, because this code is _really_ gnarly.

edgelet/edgelet-docker/src/settings.rs
```diff
@@ -494,18 +466,18 @@ mod tests {
         );
     }
 
-    #[test]
-    fn agent_image_is_resolved() {
-        let _env_lock = ENV_LOCK.lock().expect("env lock poisoned");
-        std::env::set_var("AZIOT_EDGED_CONFIG", GOOD_SETTINGS_IMAGE_RESOLVE);
-        let mut settings = Settings::new().unwrap();
-        agent_image_resolve(&mut settings).unwrap();
+    // #[test]
```
massand, line 469: Can we relocate the test?

edgelet/aziot-edged/src/lib.rs
```diff
@@ -490,10 +494,12 @@ where
         settings.hostname().to_string().to_lowercase(),
     );
 
-    env.insert(
-        GATEWAY_HOSTNAME_KEY.to_string(),
-        parent_hostname.to_string(),
-    );
+    if parent_hostname != hostname {
```
massand, line 497: Use `to_lowercase()` on both just in case

edgelet/edgelet-docker/src/settings.rs
massand: And remove the commented statements above as well?

massand (changes requested):
Looks good mostly. Few fixes needed

edgelet/aziot-edged/src/lib.rs
daprilik: do we treat these as case invariant anywhere else in the codebase? I'd rather not introduce behavior "just in case"

edgelet/edgelet-docker/src/settings.rs
daprilik: so, the tricky thing with this test is that it relies on the late binding of the `parent_hostname` value, which isn't something that is easily mocked...

Given just how much of a bodge this functionality is now, I think I'm just going to remove the test entirely.

edgelet/aziot-edged/src/lib.rs
massand: Yes - in iotedge check, we perform similar case invariance checks.

massand (approved):
LGTM

edgelet/edgelet-docker/src/config.rs
```diff
@@ -90,6 +91,20 @@ impl DockerConfig {
     }
 }
 
+pub const UPSTREAM_PARENT_KEYWORD: &str = "$upstream";
```
arsing, line 94: `"$upstream/"` ? So as to not match `"$upstreamhahaprankd/..."`

But I see the original code also had this, so if you want to leave it as it is that's fine.

edgelet/iotedge/src/check/checks/parent_hostname.rs
```diff
@@ -29,14 +27,8 @@ impl Checker for ParentHostname {
...
     fn inner_execute(&mut self, check: &mut Check) -> Result<CheckResult, failure::Error> {
-        let settings = if let Some(settings) = &check.settings {
-            settings
-        } else {
-            return Ok(CheckResult::Skipped);
-        };
-
         let config_parent_hostname =
-            if let Some(config_parent_hostname) = settings.parent_hostname() {
+            if let Some(config_parent_hostname) = check.parent_hostname.as_ref() {
```
arsing, line 31: This whole check needs to be moved into `aziotctl check`, right?

daprilik: Yes, it should. It's not pressing, but we should pop open a PBI to track this work. The bulk of the effort will just be coordinating the two PRs

edgelet/edgelet-docker/src/config.rs
daprilik: @huguesBouvier wrote it originally. I'm just preserving functionality :)

arsing (approved):

arsing (dismissed):

arsing (approved):

PR #4681, opened 2021-03-24
SDK references to newest + custom changes

vipeller:
- Changed nuget references to:  
  Microsoft.Azure.Devices -> 1.31.0-NestedEdge"
  Microsoft.Azure.Devices.Client ->1.36.0-NestedEdge"

- Moved over classes from the SDK where it became internal from public


varunpuranik (changes requested):
As discussed, let's go with a patch on 1.28.1-NestedEdge, instead of the latest SDK changes.

PR #4683, opened 2021-03-24
Iotedge system stop

lfitchett:
Pull in the IS changes and add the iotedge system stop command


daprilik:
Oof. There are some non-trivial subtlety around those `local_gateway_hostname` changes that are part of #4679 that haven't landed yet...

If we want to get this up and running first, then please add some very loud `// XXX: temporary until #4679 lands`

lfitchett:
> Oof. There is some non-trivial subtlety around those `local_gateway_hostname` changes that would land as part of #4679...
> 
> If we want to get this up and running first, then please add some very loud `// XXX: temporary until #4679 lands` that I can easily resolve when rebasing on-top of these changes

Will the Default::default() change when your changes go in? B/c all the other fields are set to the default, That's why I felt comfortable adding it.

daprilik:
So, on second glance, the "proper" changes aren't actually too tricky. It's not `Default::default()`-level ezpz, but it's pretty straightforward.

Can you integrate this logic as part of your PR?

https://github.com/Azure/iotedge/pull/4679/files#diff-395b3d6ee402831e0d5e1b0cc0f872e5d0b8385a7ba60bfb9de27d7af83d7c2dL163

lfitchett:
> So, on second glance, the "proper" changes aren't actually too tricky. It's not `Default::default()`-level ezpz, but it's pretty straightforward.
> 
> Can you integrate this logic as part of your PR?
> 
> https://github.com/Azure/iotedge/pull/4679/files#diff-395b3d6ee402831e0d5e1b0cc0f872e5d0b8385a7ba60bfb9de27d7af83d7c2dL163
> 
> It's basically just a bit of extra plumbing

Sure, and then I'll remove the changes when I cherry-pick this commit into the 1.2 release branch

lfitchett:
hmm, on second though, is there a reason for me to remove that stuff in my pr? Won't it get those changes when you merge yours? I won't include them in the cherry-pick to 1.2 anyway

daprilik:
...so that's what I'm saying. if you want to include these stubs, sure, but add a large comment that says these will get removed.

alternatively, port over the minimal subset of functionality from my PR that gets things to compile _and_ do the right thing

lfitchett:
oh, I misunderstood, I thought I needed the stubs and a bit of your code. I'll just add the comments on the stubs so it compiles

daprilik (approved):

varunpuranik (approved):

PR #4684, opened 2021-03-24
TestResultCoordinator: Explicitly fail for duplicate expected results

and-rewsmith:
We need to account for duplicate expected results in our tests. The test report operates with an assumption that expected results are unique and actual results can be duplicated. When duplicate expected results happen (due to product issues), the test report makes it seem like we are missing some actual results. But we need to make it explicit that there are duplicate expected results, which should not be happening in our tests.

Here is what I mean. Imagine the following scenario:
```
Message 0 sent
Message 1 sent
Message 1 sent
Message 2 sent

~ time passes ~

Message 0 received
Message 1 received
Message 1 received
Message 2 received
``` 

In the TRC here are the expected / actual datastructures:
```
Expected: 0->1->1->2
Actual: 0->1->1->->2
```

When the TRC is processing the report it does not account for duplicate actual results. So it is the same scenario as this:
```
Expected: 0->1->1->2
Actual: 0->1->2

```

In processing the counting report, we match results 0 and 1. We see that there are duplicate actual results for 1 so we eliminate them. Then end up in this state which is the problem. Now the counting report will view the situation as we have an unmatched expected result with sequence number 1. This is not the case. 
```
Expected: 1->2
Actual: 2

```

The report looks indicates there is one unmatched result. But really all results are matched. It is just that there is a duplicate expected result. The test report will now reflect this by giving us the number of duplicate expected results that were seen.


and-rewsmith:
I have been trying to run a longhaul test to confirm these changes work, but blocked by unrelated configuration failures. Will update.

and-rewsmith:
This worked in single node and nested longhauls over the 1.2 release

test/modules/Modules.Test/TestResultCoordinator/Reports/CountingReportGeneratorTest.cs
```diff
@@ -19,23 +19,25 @@ public class CountingReportGeneratorTest
...
+                new object[] { Enumerable.Range(1, 7).Select(v => v.ToString()), new[] { "2", "3", "4", "5", "6", "7" }, 4, 7, 6, 0, 0, 1 },
+                new object[] { Enumerable.Range(1, 7).Select(v => v.ToString()), new[] { "1", "3", "4", "5", "6", "7" }, 4, 7, 6, 0, 0, 1 },
+                new object[] { Enumerable.Range(1, 7).Select(v => v.ToString()), new[] { "1", "3", "4", "6", "7" }, 4, 7, 5, 0, 0, 2 },
+                new object[] { Enumerable.Range(1, 7).Select(v => v.ToString()), new[] { "1", "3", "4", "5", "6" }, 4, 7, 5, 0, 0, 2 },
+                new object[] { Enumerable.Range(1, 7).Select(v => v.ToString()), new[] { "2", "3", "4", "5", "7" }, 4, 7, 5, 0, 0, 2 },
+                new object[] { Enumerable.Range(1, 7).Select(v => v.ToString()), new[] { "2", "3", "4", "5", "7", "7" }, 4, 7, 5, 0, 1, 2 },
+                new object[] { new[] { "1", "2", "3", "4", "5", "6", "7", "7" }, Enumerable.Range(1, 7).Select(v => v.ToString()), 10, 7, 7, 1, 0, 0 },
+                new object[] { new[] { "1", "1", "2", "3", "4", "5", "6", "7", "7" }, Enumerable.Range(1, 7).Select(v => v.ToString()), 10, 7, 7, 2, 0, 0 },
```
and-rewsmith, line 40: I wish this structure was defined as a test abstraction with names for the fields, but I would like for this to be out of scope for this PR as I have other changes I have already made and tested to this structure coming in other PRs.

test/modules/TestResultCoordinator/Reports/CountingReport.cs
```diff
@@ -21,7 +21,8 @@ class CountingReport : TestResultReportBase
             string resultType,
             ulong totalExpectCount,
             ulong totalMatchCount,
-            ulong totalDuplicateResultCount,
```
and-rewsmith, line 24: Renaming this field for clarity.

test/modules/TestResultCoordinator/Reports/CountingReportGenerator.cs
```diff
@@ -112,19 +126,28 @@ public async Task<ITestResultReport> CreateReportAsync()
             }
 
             // Check duplicates at the end of actual results
-            while (hasActualResult && this.TestResultComparer.Matches(lastLoadedResult, this.ActualTestResults.Current))
+            while (hasActualResult && this.TestResultComparer.Matches(lastLoadedActualResult, this.ActualTestResults.Current))
```
yophilav, line 129: Please consider the case
Actual: 1 -> 2 -> 3 -> 4
Expect: 1-> 2 -> 3

This should be considered a fail. With the current logic it would be a pass.

and-rewsmith: I disagree because this while loop makes sure the lastLoadedActualResult matches the current.

test/modules/TestResultCoordinator/Reports/CountingReport.cs
```diff
@@ -45,7 +47,9 @@ class CountingReport : TestResultReportBase
 
         public ulong TotalMatchCount { get; }
 
-        public ulong TotalDuplicateResultCount { get; }
+        public ulong TotalDuplicateExpectedResultCount { get; }
```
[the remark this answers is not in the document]
and-rewsmith: I added a lot of context info to the class header docstring. Here is the main reason we need the duplicate result counts though. 
> It tracks a number of result counts in order to give full context of test operation.

yophilav (approved):

PR #4690, opened 2021-03-25
TwinTester: Give more time for updates to come in (#4623)

and-rewsmith:
Even in single node longhaul, sometimes updates arrive late due to transient network conditions. In nested longhaul the updates also have to travel through the whole chain. So bumping this time will help us eliminate false failures in the tests.


dylanbronson (approved):
lgtm cherry pick

PR #4691, opened 2021-03-25
LongHaul: (CherryPick) Adding event hub long haul mode to trc tests (#4507)

dylanbronson:
Add EventHubLongHaul mode to TRC tests so that we can pass on tests that involve using EventHub. We've seen these tests have incredibly long delays. 

This PR makes the change to implement this solution: ignore expected messages (== ones sent successfully from loadGen) that do not match, only match on actual messages (== ones received from eventHub), and also ensure that we are still receiving from EventHub by 1) Using our own timestamps in the TRC for messages we receive and 2) making sure the last message received for the actual store was within an hour (we can define that tolerance however we want)


and-rewsmith (approved):
cherry pick lgtm

